PR #2, opened 2015-08-24
Cookies support

cy6erGn0m:


ktor-core/src/org/jetbrains/ktor/application/ApplicationRequest.kt
```diff
@@ -1,11 +1,12 @@
 package org.jetbrains.ktor.application
 
-import org.jetbrains.ktor.http.*
-import org.jetbrains.ktor.interception.*
+import org.jetbrains.ktor.http.HttpRequestLine
+import org.jetbrains.ktor.http.header
+import org.jetbrains.ktor.interception.Interceptable0
```
orangy: Please use imports with .*

ktor-core/src/org/jetbrains/ktor/application/ApplicationRequest.kt
```diff
@@ -1,11 +1,12 @@
...
 
 /** Established connection with client, encapsulates request and response facilities
  */
-public interface ApplicationRequest {
+public interface ApplicationRequest : HasAttributes {
```
orangy: Why cookies and attributes are intermixed?

ktor-core/src/org/jetbrains/ktor/http/Dates.kt
```diff
@@ -5,14 +5,24 @@ import java.util.*
...
     val calendar = Calendar.getInstance();
-    calendar.setTimeInMillis(this)
-    val dateFormat = SimpleDateFormat("EEE, dd MMM yyyy HH:mm:ss z", Locale.US);
-    dateFormat.setTimeZone(TimeZone.getTimeZone("GMT"));
-    return dateFormat.format(calendar.getTime());
+    calendar.timeInMillis = this
+    return calendar.time.format(httpDateFormat)
 }
+fun Date.toHttpDateString(): String = format(httpDateFormat)
+fun String.fromHttpDateString(): Long = httpDateFormat.parse(this).time
```
orangy: We are targeting JDK 1.8, shouldn't we use java.time instead?

ktor-core/test/org/jetbrains/ktor/tests/http/CookiesTest.kt
```diff
@@ -0,0 +1,118 @@
...
+import org.jetbrains.ktor.testing.TestApplicationHost
+import org.jetbrains.ktor.testing.TestApplicationRequest
+import org.jetbrains.ktor.tests.withTestApplication
+import java.text.SimpleDateFormat
+import kotlin.test.assertEquals
+import org.junit.Test as test
+
+class CookiesTest {
+    test fun `simple cookies`() {
+        withTestApplication {
```
orangy: should we fire up application just to test cookie encoding/decoding?

ktor-core/src/org/jetbrains/ktor/application/ApplicationRequest.kt
orangy: I see, you store parsed cookies there.

ktor-core/src/org/jetbrains/ktor/application/Atributes.kt
```diff
@@ -0,0 +1,14 @@
+package org.jetbrains.ktor.application
+
+interface AttributeKey
+
+interface HasAttributes {
+    val attributes: MutableMap<AttributeKey, Any>
+}
+
+suppress("UNCHECKED_CAST")
+inline fun <T: HasAttributes, R> T.getAttributeOrPut(key: AttributeKey, block: () -> R): R = attributes.getOrPut(key, block) as R
```
orangy: when attributes are backed by concurrenthashmap (often), getOrPut is a wrong thing to do.

cy6erGn0m: Yes, this operation is not intended to be atomic. May be it need to be named to express it

ktor-core/src/org/jetbrains/ktor/http/Dates.kt
cy6erGn0m: Then we need both in spite of `java.util.Date` could be easily converted to `java.time` as there are many libraries still use `Date`

ktor-core/test/org/jetbrains/ktor/tests/http/CookiesTest.kt
cy6erGn0m: `TestApplicationRequest` requires `Application`. Alternatively I can extract parse/render implementations to some public functions but these public functions becomes visible to everyone

ktor-core/src/org/jetbrains/ktor/application/Attributes.kt
```diff
@@ -0,0 +1,24 @@
...
+
+    suppress("UNCHECKED_CAST")
+    public fun <T> get(key: AttributeKey<T>): T = map[key] as T
+
+    public fun <T> put(key: AttributeKey<T>, value: T) {
+        map[key] = value
+    }
+
+    suppress("UNCHECKED_CAST")
+    public fun <T> getOrPut(key: AttributeKey<T>, block: () -> T): T = map.getOrPut(key, block) as T
```
orangy: computeIfAbsent?

ktor-core/src/org/jetbrains/ktor/http/Cookies.kt
```diff
@@ -0,0 +1,150 @@
...
+
+                cookiePartFlag("Secure", secure),
+                cookiePartFlag("HttpOnly", httpOnly)
+        ) + extensions.map {
+            cookiePartExt(it.key.assertCookieName(), it.value, encoding)
+        }).filter { it.isNotEmpty() }
+                .joinToString("; ")
+
+
+public fun ApplicationResponse.setCookie(name: String,
```
orangy: Interceptors should be able to examine cookies set by handlers and other interceptors. With this solution they will have to intercept headers and decode them back.

ktor-core/src/org/jetbrains/ktor/http/Cookies.kt
```diff
@@ -0,0 +1,150 @@
...
+public open class RequestCookies(val request: ApplicationRequest) {
+    private val defaultDecoded: Map<String, String> by lazy { decode(CookieEncoding.URI_ENCODING) }
+
+    public fun get(name: String): String? = defaultDecoded[name]
+    public fun decode(encoding: CookieEncoding): Map<String, String> = request.attributes.getOrPut(CookiesKey(encoding)) {
+        parsedRawCookies.mapValues { decodeCookieValue(it.value, encoding) }
+    }
+
+    public open val parsedRawCookies: Map<String, String> by lazy {
+        request.header("Cookie")?.let { parseClientCookiesHeader(it) } ?: emptyMap()
```
orangy: Is this the only usage of "request" here? May be we should pass list of cookie headers instead?

cy6erGn0m: We need `request` to use `attributes`

cy6erGn0m:
@orangy i think we can merge it too, isn't it?

ktor-core/src/org/jetbrains/ktor/application/ApplicationResponse.kt
```diff
@@ -12,6 +13,20 @@ public interface ApplicationResponse {
     public fun status(value: HttpStatusCode)
     public fun interceptStatus(handler: (value: HttpStatusCode, next: (value: HttpStatusCode) -> Unit) -> Unit)
 
+    public fun cookie(name: String): Cookie? = headers.values("Set-Cookie").map { parseServerSetCookieHeader(it) }.firstOrNull { it.name == name }
+    public fun cookie(item: Cookie): Unit = headers.append("Set-Cookie", renderSetCookieHeader(item))
```
orangy: I really don't like same-looking methods for get and set access.

cy6erGn0m: what if we implement it like we did for headers? (so it will look like `response.cookies.append`)

orangy: Exactly.

cy6erGn0m: Done

ktor-netty/src/org/jetbrains/ktor/netty/NettyApplicationRequest.kt
```diff
@@ -33,4 +36,14 @@ class NettyApplicationRequest(val request: FullHttpRequest) : ApplicationRequest
 
 
     override val attributes = Attributes()
+    override val cookies = NettyCookies(this)
```
orangy: Naming inconsistent. NettyRequestCookies?

ktor-servlet/src/org/jetbrains/ktor/servlet/ServletApplicationResponse.kt
```diff
@@ -21,6 +23,7 @@ public class ServletApplicationResponse(private val servletResponse: HttpServlet
         override fun getHostHeaderNames(): List<String> = servletResponse.headerNames.toList()
         override fun getHostHeaderValues(name: String): List<String> = servletResponse.getHeaders(name).toList()
     }
+    override val cookies = ResponseCookies(this)
```
orangy: Shouldn't we use Servlet's cookies here?

orangy: Ah, probably not, if we want to intercept in headers…

ktor-netty/src/org/jetbrains/ktor/netty/NettyApplicationRequest.kt
cy6erGn0m: Done

ktor-servlet/src/org/jetbrains/ktor/servlet/ServletApplicationResponse.kt
cy6erGn0m: Yes. The other reason to not use host cookie implementation for response is to reduce possible behaviour inconsistence among the hosts

PR #3, opened 2015-08-25
Dates and attributes

cy6erGn0m:


ktor-core/src/org/jetbrains/ktor/application/Attributes.kt
```diff
@@ -0,0 +1,24 @@
...
+
+    suppress("UNCHECKED_CAST")
+    public fun <T> get(key: AttributeKey<T>): T = map[key] as T
+
+    public fun <T> put(key: AttributeKey<T>, value: T) {
+        map[key] = value
+    }
+
+    suppress("UNCHECKED_CAST")
+    public fun <T> computeIfAbsent(key: AttributeKey<T>, block: () -> T): T = map.computeIfAbsent(key) { block() } as T
```
orangy: Do you need to wrap `block` with lambda? Can't we just pass block as parameter?

ktor-core/src/org/jetbrains/ktor/http/Dates.kt
```diff
@@ -1,18 +1,17 @@
...
-fun String.fromHttpDateString(): Long {
-    val dateFormat = SimpleDateFormat("EEE, dd MMM yyyy HH:mm:ss z", Locale.US);
-    dateFormat.setTimeZone(TimeZone.getTimeZone("GMT"));
-    return dateFormat.parse(this).getTime()
-}
\ No newline at end of file
+fun String.fromHttpDateString(): ZonedDateTime = ZonedDateTime.parse(this, httpDateFormat)
+
+val httpDateFormat: DateTimeFormatter
+    get() = DateTimeFormatter
```
orangy: Why get-property? We can initialise it once.

ktor-jetty/src/org/jetbrains/ktor/jetty/CommandLine.kt
```diff
@@ -17,7 +15,9 @@ fun main(args: Array<String>) {
     }
 
     val jar = map["-jar"]?.let { File(it).toURI().toURL() }
-    val combinedConfig = ConfigFactory.load()
+    val argConfig = ConfigFactory.parseMap(map.filterKeys { it.startsWith("-P:") }.mapKeys { it.key.removePrefix("-P:") }, "Command-line options")
+    val combinedConfig = argConfig.withFallback(ConfigFactory.load())
+
```
orangy: Why do we need this?

ktor-core/src/org/jetbrains/ktor/application/Attributes.kt
cy6erGn0m: yes, SAM doesn't couldn't handle it

ktor-core/src/org/jetbrains/ktor/http/Dates.kt
cy6erGn0m: because `SimpleDateFormat` wasn't thread-safe. Will fix it

ktor-jetty/src/org/jetbrains/ktor/jetty/CommandLine.kt
cy6erGn0m: Because it should be way to override any config parameter via command-line. Personally I need it to override the http port as config file is under version control

ktor-core/src/org/jetbrains/ktor/http/Dates.kt
cy6erGn0m: fixed

ktor-core/src/org/jetbrains/ktor/application/Attributes.kt
orangy: It's because you don't pass key to block, as map's own signature does. Why don't we add key to block?

ktor-jetty/src/org/jetbrains/ktor/jetty/CommandLine.kt
orangy: Ok, I see. Then we probably should drop it from this PR, implement reusable command-line configuration facility as we need this in Netty command line and any other host we might have.

cy6erGn0m:
@orangy all done, command line reverted

ktor-core/src/org/jetbrains/ktor/application/Attributes.kt
cy6erGn0m: why do we need it inside the block?

PR #5, opened 2015-09-01
Migrate to append-only response headers to support multiple headers

cy6erGn0m:
`ResponseHeaders` class encapsulates headers response logic including intercepting, reduces code duplication in host implementations

- Functions `header(name)`, `header(name, value)` and `interceptHeaders` are removed and added property `headers: ReponseHeaders` that contains corresponding functions: `get(name)`, `append(name, value)` and `intercept(handler)`

- Notice `ResponseHeaders.append(name, value)` is self-descriptive so it is clear for user that it appends header rather than set/replace existing value(s)

- Function `contians` function useful to test whether we have header set. For example:
```
if (HttpHeaders.ContentType !in response.headers) {
    response.contentType(ContentType.Any)
}
```
- Function `values(name)` provides a list of header values
- Function `allValues()` constructs `ValuesMap` of all headers

### DIFF
--- a/ktor-core/src/org/jetbrains/ktor/application/ApplicationResponse.kt
+++ b/ktor-core/src/org/jetbrains/ktor/application/ApplicationResponse.kt
@@ -1,13 +1,12 @@
 package org.jetbrains.ktor.application
 
 import org.jetbrains.ktor.http.*
+import org.jetbrains.ktor.interception.*
 import java.io.*
 import java.nio.charset.*
 
 public interface ApplicationResponse {
-    public fun header(name: String): String?
-    public fun header(name: String, value: String)
-    public fun interceptHeader(handler: (name: String, value: String, next: (name: String, value: String) -> Unit) -> Unit)
+    public val headers: ResponseHeaders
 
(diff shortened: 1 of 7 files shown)


ktor-core/src/org/jetbrains/ktor/application/ApplicationResponse.kt
```diff
@@ -20,6 +19,31 @@ public interface ApplicationResponse {
     public fun interceptSend(handler: (message: Any, next: (message: Any) -> ApplicationRequestStatus) -> ApplicationRequestStatus)
 }
 
+public abstract class ResponseHeaders {
+    private val headersChain = Interceptable2<String, String, Unit> { name, value ->
+        appendHeaderImpl(name, value)
+    }
+
+    public fun contains(name: String): Boolean = getHeaderValuesImpl(name).isNotEmpty()
+    public fun get(name: String): String? = getHeaderValuesImpl(name).firstOrNull()
```
orangy: singleOrNull, probably? What is intended behaviour when we append multiple headers, and the get one?

PR #6, opened 2015-09-03
Urlencoded content

cy6erGn0m:
- parse urlencoded content function
- content charset shortcut property
- netty should handle urlencoded parameters

### DIFF
--- a/ktor-core/test/org/jetbrains/ktor/tests/http/ContentTypeTest.kt
+++ b/ktor-core/test/org/jetbrains/ktor/tests/http/ContentTypeTest.kt
@@ -89,4 +89,14 @@ class ContentTypeTest {
             }
         }
     }
+
+    Test
+    fun `content type with empty parameters block`() {
+        on("parsing empty parameters block so we have trailing semicolon and possibly whitespaces") {
+            it("shouldn't fail and should pass equality checks") {
+                assertEquals(ContentType.Text.Plain, ContentType.parse("text/plain; "))
+                assertEquals(ContentType.Text.Plain, ContentType.parse("text/plain;"))
+            }
+        }
+    }
 }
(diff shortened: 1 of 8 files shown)


ktor-netty/src/org/jetbrains/ktor/netty/NettyApplicationRequest.kt
```diff
@@ -28,6 +25,10 @@ class NettyApplicationRequest(val request: FullHttpRequest) : ApplicationRequest
             QueryStringDecoder(request.uri).parameters().forEach {
                 appendAll(it.key, it.value)
             }
+            // as far as we have full request we can access request body many times
+            if (contentType().contentType == "application" && contentType().contentSubtype == "x-www-form-urlencoded") {
+                appendAll(parseUrlEncodedParameters())
```
orangy: better add it to ContentType.Application as a standard content-type

cy6erGn0m: yes, we can add it but still can't compare contentType because it could contain "parameters" that prevent it to be equal to our constant

orangy: We probably need to add ContentType.matches(value), so that it takes type/* into account, etc.

cy6erGn0m:
@orangy added `ContentType.withoutParameters` function and `ContentType.Application.FormUrlEncoded` so now I use them in request content type comparison

orangy:
I don't like "withoutParameters" function, what's the generic purpose of this API? It's more likely for users match what they've got to Accept headers, or incoming content-type to what they can process, like you does. So they likely have some contet-type value from external source, and then some template to match against. Template might have parameters, or not, might have Any type or subtype, etc.

cy6erGn0m:
@orangy I agree that match function is better solution. See new version with `ContentType.match` functions and tests

ktor-core/test/org/jetbrains/ktor/tests/http/ContentTypeTest.kt
```diff
@@ -89,4 +89,10 @@ class ContentTypeTest {
             }
         }
     }
+
+    Test
+    fun `content type with empty parameters block`() {
+        ContentType.parse("text/plain; ")
```
orangy: Shouldn't we check for equality to ContentType.Text.Plain?

cy6erGn0m: Added it. But the main goal here is to test that we are not crashing at such string as we did before

PR #10, opened 2015-09-24
Support quoted and escaped values in headers

cy6erGn0m:


orangy:
Copied tests, added own, implemented with simple recursive parser (3 times faster)

PR #11, opened 2015-09-25
Case insensitive ValuesMap, headers map should be case insensitive

cy6erGn0m:


orangy:
Implemented slightly differently

PR #16, opened 2015-11-26
Response compression

cy6erGn0m:
Support for gzip/deflate, configurable compression conditions (to be able to disable compression in cases), simple conditions (by size, compress stream)

### DIFF
--- a/ktor-core/src/org/jetbrains/ktor/http/Compression.kt
+++ b/ktor-core/src/org/jetbrains/ktor/http/Compression.kt
@@ -1,0 +1,81 @@
+package org.jetbrains.ktor.http
+
+import org.jetbrains.ktor.application.*
+import java.io.*
+import java.util.zip.*
+
+data class CompressionOptions(var minSize: Long = 0L,
+                              var compressStream: Boolean = true,
+                              var defaultEncoding: String = "deflate",
+                              val compressorRegistry: MutableMap<String, CompressionEncoder> = hashMapOf(),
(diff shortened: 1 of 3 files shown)


ktor-core/src/org/jetbrains/ktor/http/Compression.kt
```diff
@@ -0,0 +1,81 @@
+package org.jetbrains.ktor.http
+
+import org.jetbrains.ktor.application.*
+import java.io.*
+import java.util.zip.*
+
+data class CompressionOptions(var minSize: Long = 0L,
+                              var compressStream: Boolean = true,
+                              var defaultEncoding: String = "deflate",
```
jkwatson: var?

ktor-core/src/org/jetbrains/ktor/http/Compression.kt
```diff
@@ -0,0 +1,81 @@
...
+                              val compressorRegistry: MutableMap<String, CompressionEncoder> = hashMapOf(),
```
jkwatson: this API bugs me. So, the consumer is expected to mutate the map, before calling setupCompression?  Or am I missing something here?

cy6erGn0m: the `configure` lambda of the `setupCompression` could mutate them

```kotlin
init {
    setupComression {
        defaultEncoding = "gzip"
    }
}
```

PR #36, opened 2016-07-30
Kobalt

cbeust:
Here is a Kobalt build file for ktor.

FYI, this is more for your curiosity, I won't be hurt if you decide to close that PR without merging, but I was curious to see how Kobalt would fare on ktor and I think other people might as well. I'm pretty happy with the result overall.

The current ktor build is about 25 `pom.xml` files adding up to 1000 lines with a lot of repetition in it. Kobalt's `Build.kt` is a single file with around 200 lines and hardly any repetition (and potential to factor it out further).

Build time is about on par with Maven, which is not very surprising since most of the time is spent running `kotlinc` and `JUnit`.

You can build with `./kobaltw assemble` and run tests with `./kobaltw test`.

I also made the `ktor-samples-hello` executable (I generate a fat jar and I declare the correct main class) and you can launch it with `./kobaltw ktor-samples-hello:run`.

Finally, if you don't feel like trying this yourself, I set up a travis-ci job:

https://travis-ci.org/cbeust/ktor

The output is very verbose though (mostly downloading dependencies) so it won't give you a clear idea of what a regular build is like. Also, the build seems to have a couple of errors which I'm not seeing on my local machine (tested on Windows/Cygwin and MacOS so far), so probably a travis-specific thing.

Comments welcome.


orangy:
Close as outdated, sorry @cbeust :)

PR #39, opened 2016-08-24
use more idiomatic name memberLeft

pabl0rg:
memberLeaved is confusing


cy6erGn0m:
@pabl0rg thanks, merged

PR #41, opened 2016-09-12
Bind pool number and thread counter instead of globally for DefaultEx…

spand:
…ecutorServiceBuilder.

Currently both counters have the same value.


cy6erGn0m:
Nice catch, thanks!

PR #51, opened 2016-11-10
Added MapApplicationConfig.putList for ease of use.

spand:


ktor-core/src/org/jetbrains/ktor/config/MapApplicationConfig.kt
```diff
@@ -16,6 +16,12 @@ class MapApplicationConfig : ApplicationConfig {
         map.put(path, value)
     }
 
+    fun putList(path: String, values: List<String>) {
+        this.put(path, "")
```
orangy: `this` is redundant.

ktor-core/src/org/jetbrains/ktor/config/MapApplicationConfig.kt
```diff
@@ -16,6 +16,12 @@ class MapApplicationConfig : ApplicationConfig {
...
+        this.put("$path.size", values.size.toString())
```
orangy: Use `combine(path, "size")` for consistency.

ktor-core/src/org/jetbrains/ktor/config/MapApplicationConfig.kt
```diff
@@ -16,6 +16,12 @@ class MapApplicationConfig : ApplicationConfig {
...
+        values.forEachIndexed { i, value -> put("$path.$i", value)}
```
orangy: Use `combine(path, i.toString())`

ktor-core/src/org/jetbrains/ktor/config/MapApplicationConfig.kt
```diff
@@ -16,6 +16,12 @@ class MapApplicationConfig : ApplicationConfig {
         map.put(path, value)
     }
 
+    fun putList(path: String, values: List<String>) {
```
orangy: Use `Collection` for `values` to allow for more input types.

orangy:
Hmm, I thought `putList` was good for consistency with `getList`, but now you changed it to just `put` :)

Issue #52, opened 2016-11-22
java.lang.IllegalStateException: ReadListener already set


mathiasbn:
I have a long running request (>1min). At some point the server just crashes with the following breakpoint.
Even if I just put a "thread only" suspending breakpoint in the `post` body it crashes
```
java.lang.IllegalStateException: ReadListener already set
    at org.eclipse.jetty.server.HttpInput.setReadListener(HttpInput.java:600) ~[jetty-server-9.3.13.v20161014.jar:9.3.13.v20161014]
    at org.jetbrains.ktor.servlet.ServletReadChannel.read(ServletReadChannel.kt:45) ~[ktor-servlet-0.2.3.jar:0.2.3]
    at org.jetbrains.ktor.nio.AsyncReadChannelAdapterStream.read(Channels.kt:114) ~[ktor-core-0.2.3.jar:0.2.3]
    at sun.nio.cs.StreamDecoder.readBytes(StreamDecoder.java:284) ~[na:1.8.0_60]
    at sun.nio.cs.StreamDecoder.implRead(StreamDecoder.java:326) ~[na:1.8.0_60]
    at sun.nio.cs.StreamDecoder.read(StreamDecoder.java:178) ~[na:1.8.0_60]
    at java.io.InputStreamReader.read(InputStreamReader.java:184) ~[na:1.8.0_60]
    at java.io.Reader.read(Reader.java:140) ~[na:1.8.0_60]
    at kotlin.io.TextStreamsKt.copyTo(ReadWrite.kt:116) ~[kotlin-stdlib-1.0.5.jar:1.0.5]
    at kotlin.io.TextStreamsKt.copyTo$default(ReadWrite.kt:113) ~[kotlin-stdlib-1.0.5.jar:1.0.5]
    at kotlin.io.TextStreamsKt.readText(ReadWrite.kt:100) ~[kotlin-stdlib-1.0.5.jar:1.0.5]
    at org.jetbrains.ktor.request.RequestContent$contentAsString$2.invoke(RequestContent.kt:11) ~[ktor-core-0.2.3.jar:0.2.3]
    at org.jetbrains.ktor.request.RequestContent$contentAsString$2.invoke(RequestContent.kt:10) ~[ktor-core-0.2.3.jar:0.2.3]
    at kotlin.SynchronizedLazyImpl.getValue(Lazy.kt:131) ~[kotlin-stdlib-1.0.5.jar:1.0.5]
[...]

PR #54, opened 2016-11-30
Fix multiple dispatch problem with async servlet timeouts.

spand:
Servlet containers with async context will redispatch after a timeout. In Jettys case the default seems to be 30 secs. I think it is part of the three steps discribed here: https://docs.oracle.com/javaee/6/api/javax/servlet/AsyncContext.html

The provided testcase fails without the fix.

This might also be the cause of what we are seeing in #52

### DIFF
--- a/ktor-hosts/ktor-jetty/test/MultipleDispatchOnTimeout.kt
+++ b/ktor-hosts/ktor-jetty/test/MultipleDispatchOnTimeout.kt
@@ -1,0 +1,60 @@
+import org.jetbrains.ktor.application.BasicApplicationEnvironment
+import org.jetbrains.ktor.application.call
+import org.jetbrains.ktor.application.install
+import org.jetbrains.ktor.application.respondWrite
+import org.jetbrains.ktor.config.MapApplicationConfig
+import org.jetbrains.ktor.host.applicationHostConfig
+import org.jetbrains.ktor.host.connector
+import org.jetbrains.ktor.jetty.embeddedJettyServer
+import org.jetbrains.ktor.logging.SLF4JApplicationLog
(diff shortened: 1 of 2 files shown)


ktor-hosts/ktor-jetty/test/MultipleDispatchOnTimeout.kt
```diff
@@ -0,0 +1,60 @@
...
+import java.net.URL
+import java.util.concurrent.atomic.AtomicInteger
+import kotlin.test.assertEquals
+
+
+class MultipleDispatchOnTimeout {
+
+    @Test
+    fun foo(){
+        val port = 44003
```
cy6erGn0m: fixed port number is not good, see `findFreePort`

ktor-hosts/ktor-jetty/test/MultipleDispatchOnTimeout.kt
```diff
@@ -0,0 +1,60 @@
...
+import org.junit.Test
+import java.net.URL
+import java.util.concurrent.atomic.AtomicInteger
+import kotlin.test.assertEquals
+
+
+class MultipleDispatchOnTimeout {
+
+    @Test
+    fun foo(){
```
cy6erGn0m: test name

spand:
Yes. About 2.8 seconds.

mathiasbn:
A temporary workaround:
System.setProperty("org.eclipse.jetty.server.HttpChannelState.DEFAULT_TIMEOUT","0")

cy6erGn0m:
@spand merged, thanks!

PR #58, opened 2016-12-22
Location classes with List property cannot be routed to with zero url…

spand:
… parameter values.


e5l:

PR #68, opened 2017-01-22
fix HttpStatusCode.allStatusCodes computation

BugsBunnyBR:
The current implementation always return zero values to `allStatusCodes`. I am not quite sure about the fix, but it seems to work.


cy6erGn0m:
is it because of the comparison `it.returnType == HttpStatusCode::class` ?

BugsBunnyBR:
@cy6erGn0m,
As far as I could understand, yes.

orangy:
I would prefer failing test first.

BugsBunnyBR:
I included a simple unit test that fails in the current implementation, but works with the PR suggestion.

orangy:
Merged manually with more tests and slight change in API for clarity.

PR #74, opened 2017-03-10
ktor-samples-httpbin

jmfayard:
Work In Progress but plenty of endpoints implemented already

The most interesting stuff is in [HttpbinApplication.kt](https://github.com/jmfayard/ktor/blob/sample-httpbin/ktor-samples/ktor-samples-httpbin/src/org/jetbrains/ktor/samples/httpbin/HttpbinApplication.kt)

The best way to play with the endpoints is to install https://www.getpostman.com/   and to `import` the file `httpbin.postman_collection.json`


jmfayard:
@orangy  `ktor-samples-httpbin` in Postman!

httpbin

Most important thing missing is to convert `httpbin.1.html` to Freemarker or another engine template so that the links are working when opening http://localhost:8080/

orangy:
I would ask you to either squash all commits into single commit, or update commit message more or less in line with https://chris.beams.io/posts/git-commit/ recommendations. We don't strictly follow these rules, but trying to be descriptive and specific, and use imperative form.

jmfayard:
Commits squashed

jmfayard:
Obsoleted by #76 , I did a git mess in this one

PR #85, opened 2017-04-11
Replace future/async usages with more idiomatic coroutine code

elizarov:


orangy:
Cleaned, fixed, and merged.

PR #94, opened 2017-04-21
Chat sample fix

OliverO2:
Fixes two issues in the Chat (websocket) sample:

1. On startup (browser connecting), the application fails with
```
500 Internal Server Error

IllegalArgumentException: Couldn't instantiate type class org.jetbrains.ktor.samples.chat.Session for parameters [userId]
```

2. The '/who' command displays hash IDs instead of names.

### DIFF
--- a/ktor-samples/ktor-samples-websocket/src/org/jetbrains/ktor/samples/chat/ChatApplication.kt
+++ b/ktor-samples/ktor-samples-websocket/src/org/jetbrains/ktor/samples/chat/ChatApplication.kt
@@ -35,16 +35,16 @@ fun Application.main() {
 
             pingInterval = Duration.ofMinutes(1)
 
-            server.memberJoin(session.userId, this)
+            server.memberJoin(session.id, this)
 
             handle { frame ->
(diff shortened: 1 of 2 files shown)


cy6erGn0m:
@OliverO2 where `userId` comes from? Isn't it caused by old session cookies?

OliverO2:
@cy6erGn0m You are probably right. As I was playing around with several samples, an old session cookie might have been around. I have just reverted my code changes and tried the original code on a clean browser. Now all is well, my patch is not needed.

Maybe it would be a good idea to make samples more robust to deal with session cookie re-use. Anyway, thanks for looking at it!

cy6erGn0m:
@OliverO2 you could simply skip rename commit but keep keys -> values change

OliverO2:
@cy6erGn0m Done.

cy6erGn0m:
thanks!

fyn-dev:
I have same issue with my browser. Once I changed it to userId everything start to work right. Maybe you at least provide in this example how to use websockets without cookies or sessions - more simplified alternative?

PR #104, opened 2017-05-13
Modifying test to address property sslPort

jmcshane:
The sslPort property on the HttpsRedirect feature is not currently tested in this feature test. The modified test confirms the new functionality.


cy6erGn0m:
@jmcshane I believe both cases need to be tested: default and custom ports

cy6erGn0m:
Merged manually

PR #114, opened 2017-05-24
Failing testcase with empty string valued path parameter.

spand:


orangy:
This one is unclear. With one last empty string parameter it's fine, but what if it's not last? What if there are two empty strings? It would become ambiguous. I think a better approach would be to use optional path segment with a default value.

spand:
If it is not last then I would expect it to produce the string for ie. `/user//foo`. Two empty: `/user//`. Maybe I havent thought this through but it seems very weird that the string <-> location mapping is not one to one. Generating hrefs that cannot roundtrip into the equivalent location seems very unsound. What other criteria do you weigh higher ?

e5l:

PR #115, opened 2017-05-24
Failing testcase with too much escaping in location values.

spand:


orangy:
From JDK's UrlEncoder:
```
         * It appears that both Netscape and Internet Explorer escape
         * all special characters from this list with the exception
         * of "-", "_", ".", "*". While it is not clear why they are
         * escaping the other characters, perhaps it is safest to
         * assume that there might be contexts in which the others
         * are unsafe if not escaped. Therefore, we will use the same
         * list. It is also noteworthy that this is consistent with
         * O'Reilly's "HTML: The Definitive Guide" (page 164).
         *
         * As a last note, Intenet Explorer does not encode the "@"
         * character which is clearly not unreserved according to the
         * RFC. We are being consistent with the RFC in this matter,
         * as is Netscape.
```

spand:
And for that we must be punished forever. Any chance of Ktor not following the same path and implementing the RFC (perhaps with a compat escape hatch) instead of a the conservative "might break crappy software" route ?

orangy:
Sure, it's not for an excuse :) Just checked what happened and why, and drop a notice here so I don't forget when someone comes to actually fix it.

cy6erGn0m:
Btw, why do you think that `;` should be allowed? For sure it has special meaning in URL so need to be escaped anyway at least in query parameters and values

spand:
There is a bug in the code. It should be the unreserved val instead of segment... so I don’t. I had a look at fixing this myself the other day but must have been distracted. I can update the PR tomorrow when at work.

spand:
I remember now. So it turns out I looked in rfc2396, but it has been obsoleted by rfc3986. In the new rfc the unreserved set has been reduced to `unreserved    = ALPHA / DIGIT / "-" / "." / "_" / "~"` versus 
```
unreserved    = alphanum | mark
      mark          = "-" | "_" | "." | "!" | "~" | "*" | "'" |
                      "(" | ")"
```
I guess Ktor should follow the newer rfc. I dont know how you would prefer to do this. Continue with the `.replace()` calls in `encodeURLPart` in Codecs.kt or implement/use an encoder that supports this natively.

e5l:

PR #116, opened 2017-05-24
ktor-freemarker: Rendering freemarker template without etag

tokuhirom:
In a typical use case, users don't want to generate ETag for HTML templates.


cy6erGn0m:
Thanks!

PR #121, opened 2017-06-03
Compression: preserve HTTP status

diego-rocha-movile:
Preserve HTTP status of compressed responses by passing the original response status to the new one.
Before this changes, every CompressedResponse and CompressedWriteResponse would return a 200 Ok status because their status was never defined.

A new test case has been added to avoid regressions.


cy6erGn0m:
@diego-rocha-movile thanks!

PR #138, opened 2017-07-08
Added a JacksonSupport feature.

magnuschatt:
Added a module very similar to GsonSupport, which uses Jackson instead of GSON. I really don't like the fact that GSON breaks Kotlins null safety by inserting nulls into non-nullable fields, without throwing an exception!


magnuschatt:
Oh wow! Didn't see the other Jackson pull request. You may reject this.

magnuschatt:
Oh wow! Didn't notice the other Jackson pull request... You may do with this pull request as you like, as long some sort of Jackson support is merged :)

orangy:
That one is really old and probably doesn't match current architecture. Thank you!

dinomite:
I like it and it works with my cursory testing.

orangy:
Will be obsoleted by #167, but we will still need Jackson integration, so I'm leaving it open for future reference.

rocketraman:
Works for me out of the box as well, whereas GsonSupport did not.

orangy:
Will merge another PR with current architecture. Thank you for your effort! I was too slow to merge it working on ContentNegotiation feature.

PR #150, opened 2017-07-16
Added sample for Static Content

hhariri:


ktor-features/pom.xml
```diff
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
```
orangy: Do not change pom.xml without a reason

ktor-samples/ktor-samples-static-content/pom.xml
```diff
@@ -0,0 +1,34 @@
...
+    <artifactId>ktor-samples-static-content</artifactId>
+
+    <build>
+        <sourceDirectory>src</sourceDirectory>
+        <resources>
+            <resource>
+                <directory>resources</directory>
+            </resource>
+        </resources>
+    </build>
```
orangy: This shouldn't be needed, it's specified in root pom

ktor-samples/ktor-samples-static-content/src/org/jetbrains/ktor/samples/staticcontent/StaticContent.kt
```diff
@@ -0,0 +1,27 @@
+package org.jetbrains.ktor.samples.staticcontent
+
+import org.jetbrains.ktor.content.*
+import org.jetbrains.ktor.host.embeddedServer
```
orangy: Please use `*` imports

ktor-features/pom.xml
hhariri: Hmm. Wrong pom, sorry.

ktor-samples/ktor-samples-static-content/pom.xml
hhariri: It's essentially copy/paste of other sample, so I guess those need to be cleaned up too.

PR #177, opened 2017-08-09
Add support for UUIDs in session serialization

schleinzer:
In my current project, I want to store a `UUID` within a session object, finding that those are not supported during serialization. Instead of fiddling with `String`s, I would rather be very happy to have `UUID`s supported.

So I added support for `UUID`s to `SessionSerializerReflection`, quite along the lines the recently added support for enums.

I hope you like the change and accept it.

Thank you!


cy6erGn0m:
Thanks!

PR #179, opened 2017-08-11
Use MessageDigest.isEqual to compare hashes

kosiakk:
Timing attack on HMAC is possible otherwise

### DIFF
--- a/ktor-core-tests/test/org/jetbrains/ktor/tests/session/SessionTest.kt
+++ b/ktor-core-tests/test/org/jetbrains/ktor/tests/session/SessionTest.kt
@@ -135,7 +135,7 @@ class SessionTest {
             }
 
             handleRequest(HttpMethod.Get, "/2") {
-                val brokenSession = sessionId.mapIndexed { i, c -> if (i == sessionId.lastIndex) 'x' else c }.joinToString("")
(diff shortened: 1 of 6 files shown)


ktor-core-tests/test/org/jetbrains/ktor/tests/session/SessionTest.kt
```diff
@@ -479,13 +487,21 @@ class SessionTest {
...
-                val brokenSession = sessionId.mapIndexed { i, c -> if (i == sessionId.lastIndex) 'x' else c }.joinToString("")
+                val brokenSession = flipLastHexDigit(sessionId)
                 addHeader(cookieName, brokenSession)
             }.let { call ->
                 assertEquals("ok, null, false", call.response.content)
             }
         }
     }
+
+    private fun flipLastHexDigit(sessionId: String) = sessionId.mapIndexed { index, letter ->
```
kosiakk: 'x' is not a valid HEX digit -> NumberFormatException later, when hex() is called
Still, one old test for values with 'x' is preserved

cy6erGn0m:
Nice catch! In spite of that it is unlikely that one can actually do time-attack over real network it is better to eliminate it

PR #182, opened 2017-08-14
Default Accept-Charset from ISO-8859-1 to UTF-8

bluemix:
Ktor responds with UTF-8 by [default](https://github.com/Kotlin/ktor/issues/80), then why not accepts default content charset as UTF-8 too?

### DIFF
--- a/ktor-core/src/org/jetbrains/ktor/content/IncomingContent.kt
+++ b/ktor-core/src/org/jetbrains/ktor/content/IncomingContent.kt
@@ -20,5 +20,5 @@ suspend fun IncomingContent.readText(): String {
     }
 
     readChannel().copyTo(buffer) // TODO provide buffer pool to copyTo function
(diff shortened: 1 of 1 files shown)


cy6erGn0m:
The default ISO-8859-1 encoding was initially specified in RFC 2616. However RFC 7231 obsoletes it (see Appendix B):

```
   The default charset of ISO-8859-1 for text media types has been
   removed; the default is now whatever the media type definition says.
   Likewise, special treatment of ISO-8859-1 has been removed from the
   Accept-Charset header field.
```

New RFC clearly states that the default character encoding is a part of MIME therefore nowadays a web application has to decide which character encoding it should use during **content negotiation**

So both request and response body default character encoding should be *negotiated*.

orangy:
`ContentNegotiation` feature is in the works, however according to many sources I looked through, it is still not employed everywhere and many clients just assume UTF-8 without proper protocol. More over, only server-side content negotiation is more or less specified. Client-side content negotiation is not specified in any way and thus cannot be implemented in the framework.

PR #183, opened 2017-08-15
add unprocessable entity status code

ayrat555:


orangy:
This is part of the WebDAV standard. We should either implement [RFC 4918](https://tools.ietf.org/html/rfc4918) completely, or do not add things from it at all.

PR #189, opened 2017-08-22
Default error (status) pages should properly show HTML page, not plaintext

kosiakk:
Default 404 page is showing html-markup as plain text. 
The response being sent is predefined and is clearly html in `ApplicationSendPipeline.installDefaultTransformations`:

- there's no need to favor any user-provided ContentType
- defaultTextContentType is `text/plain` which is not suitable for this case

To test: open any example like `Ktor :: Hello` and try to open any wrong url. 
Before commit:
```
<H1>404 Not Found</H1><P>Cannot find resource with the requested URI: /x</P>
```
After correction:

> ## 404 Not Found
> ### Cannot find resource with the requested URI: /x

### DIFF
--- a/ktor-core-tests/test/org/jetbrains/ktor/tests/http/StatusPageTest.kt
+++ b/ktor-core-tests/test/org/jetbrains/ktor/tests/http/StatusPageTest.kt
@@ -23,7 +23,9 @@ class StatusPageTest {
                 call.respond(HttpStatusCode.NotFound)
             }
             handleRequest(HttpMethod.Get, "/foo").let { call ->
+                assertEquals(HttpStatusCode.NotFound, call.response.status(), "Actual status must be kept")
                 assertEquals("<html><body>error 404</body></html>", call.response.content)
+                assertEquals(ContentType.Text.Html, call.response.contentType)
             }
         }
     }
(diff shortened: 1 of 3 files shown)


ktor-hosts/ktor-hosts-common/src/org/jetbrains/ktor/host/DefaultTransform.kt
```diff
@@ -19,8 +19,7 @@ fun ApplicationSendPipeline.installDefaultTransformations() {
                 ByteArrayContent(value)
             }
             is HttpStatusContent -> {
-                val contentType = call.defaultTextContentType(null)
-                TextContent("<H1>${value.code}</H1><P>${value.message.escapeHTML()}</P>", contentType, value.code)
+                TextContent("<H1>${value.code}</H1><P>${value.message.escapeHTML()}</P>", ContentType.Text.Html, value.code)
```
orangy: Should probably be `ContentType.Text.Html.withCharset(Charsets.UTF8)`

orangy: It would also be nice to add a regression test. It was broken during introducing processing of charsets/contenttypes with `defaultTextContentType` and wouldn't be broken if there were a test :)

kosiakk:
I failed to write tests for `defaultTextContentType`. Somehow, `defaultHostPipeline` must be included into `TestApplicationHost`.

ktor-core-tests/test/org/jetbrains/ktor/tests/http/StatusPageTest.kt
```diff
@@ -23,7 +23,9 @@ class StatusPageTest {
                 call.respond(HttpStatusCode.NotFound)
             }
             handleRequest(HttpMethod.Get, "/foo").let { call ->
+                assertEquals(HttpStatusCode.NotFound, call.response.status(), "Actual status must be kept")
```
kosiakk: This test failed (handler of `statusFile` responded with `200 Ok`) and caused the changes in `StatusPages.Configuration.statusFile`

orangy:
Oh, I've included status preserving code in StatusPages for the different issue before looking into this. Could you please rebase the PR?

orangy:
Merged manually with some cleanup

Issue #165, opened 2017-07-30
Allow objects in locations for parameterless paths.


spand:
ie. `@location("/") object Index`

PR #191, opened 2017-08-24
Allow objects in locations for parameterless paths

t-kameyama:
Fix for #165

### DIFF
--- a/ktor-features/ktor-locations/test/org/jetbrains/ktor/tests/locations/LocationsTest.kt
+++ b/ktor-features/ktor-locations/test/org/jetbrains/ktor/tests/locations/LocationsTest.kt
@@ -410,5 +410,13 @@ class LocationsTest {
         urlShouldBeUnhandled("/users/me")
     }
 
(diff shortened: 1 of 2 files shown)


ktor-features/ktor-locations/test/org/jetbrains/ktor/tests/locations/LocationsTest.kt
```diff
@@ -341,5 +341,74 @@ class LocationsTest {
...
+            get<help> {
+                call.respond(HttpStatusCode.OK)
+            }
+        }
+        urlShouldBeHandled(href)
+        urlShouldBeUnhandled("/help/123")
+    }
+
+    @location("/users")
+    object users {
```
kosiakk: A very realistic example! 👍

spand:
Does it throw an error on ie `@location("/users/{id}") ? Should it ?

orangy:
It should. There is no place to bind `{id}`

orangy:
Thank you!

PR #198, opened 2017-09-07
Support Configurable Log Levels for Call Logging Feature

ddsoyka:
This PR allows the `CallLogging` feature to be configured to use any log level, while defaulting to the previous behaviour of logging at `trace`.

### DIFF
--- a/ktor-core/src/org/jetbrains/ktor/features/CallLogging.kt
+++ b/ktor-core/src/org/jetbrains/ktor/features/CallLogging.kt
@@ -54,7 +54,7 @@ class CallLogging(private val log: Logger, private val monitor: ApplicationMonit
             level == Level.INFO && log.isInfoEnabled -> log.info(message)
             level == Level.DEBUG && log.isDebugEnabled -> log.debug(message)
             level == Level.TRACE && log.isTraceEnabled -> log.trace(message)
-            else -> throw IllegalArgumentException("Call logging is not supported for levels higher than INFO.")
+            else -> throw IllegalArgumentException("Call logging is not supported for the $level log level.")
         }
     }
 
--- a/ktor-core/src/org/jetbrains/ktor/features/CallLogging.kt
+++ b/ktor-core/src/org/jetbrains/ktor/features/CallLogging.kt
@@ -6,19 +6,22 @@ import org.jetbrains.ktor.pipeline.*
 import org.jetbrains.ktor.request.*
 import org.jetbrains.ktor.util.*
 import org.slf4j.*
+import org.slf4j.event.*
 
-class CallLogging(private val log: Logger, private val monitor: ApplicationMonitor) {
+class CallLogging(private val log: Logger, private val monitor: ApplicationMonitor, private val configuration: Configuration) {
 
-    class Configuration
+    class Configuration {
+        var level: Level = Level.TRACE
+    }
 
-    private val starting: (Application) -> Unit = { it.log.trace("Application starting: $it") }
-    private val started: (Application) -> Unit = { it.log.trace("Application started: $it") }
-    private val stopping: (Application) -> Unit = { it.log.trace("Application stopping: $it") }
+    private val starting: (Application) -> Unit = { log("Application starting: $it") }
+    private val started: (Application) -> Unit = { log("Application started: $it") }
+    private val stopping: (Application) -> Unit = { log("Application stopping: $it") }
     private var stopped: (Application) -> Unit = {}
 
     init {
         stopped = {
-            it.log.trace("Application stopped: $it")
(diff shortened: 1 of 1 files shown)


ktor-core/src/org/jetbrains/ktor/features/CallLogging.kt
```diff
@@ -6,19 +6,22 @@ import org.jetbrains.ktor.pipeline.*
 import org.jetbrains.ktor.request.*
 import org.jetbrains.ktor.util.*
 import org.slf4j.*
+import org.slf4j.event.*
 
-class CallLogging(private val log: Logger, private val monitor: ApplicationMonitor) {
+class CallLogging(private val log: Logger, private val monitor: ApplicationMonitor, private val configuration: Configuration) {
```
orangy: Do not store configuration in a feature instance. It is a mutable class that could potentially be captured from a config script and modified concurrently. Get a `level` property from it and store it instead.

ktor-core/src/org/jetbrains/ktor/features/CallLogging.kt
```diff
@@ -35,8 +38,9 @@ class CallLogging(private val log: Logger, private val monitor: ApplicationMonit
         override val key: AttributeKey<CallLogging> = AttributeKey("Call Logging")
         override fun install(pipeline: Application, configure: Configuration.() -> Unit): CallLogging {
             val loggingPhase = PipelinePhase("Logging")
-            Configuration().apply(configure)
-            val feature = CallLogging(pipeline.log, pipeline.environment.monitor)
+            val configuration = Configuration()
+            configuration.apply(configure)
```
orangy: Minor: could be one-liner `val configuration = Configuration().apply(configure)`

ktor-core/src/org/jetbrains/ktor/features/CallLogging.kt
```diff
@@ -47,12 +51,21 @@ class CallLogging(private val log: Logger, private val monitor: ApplicationMonit
 
     }
 
+    private fun log(message: String) {
+        when (configuration.level) {
+            Level.ERROR -> if (log.isErrorEnabled) log.error(message)
```
orangy: Looks like it doesn't make any sense to allow `ERROR` or `WARNING` level for call logging. Not sure we should limit it here, but that looks weird.

ddsoyka:
What type of exception should be thrown here?

orangy:
`IllegalArgumentException`?

ktor-core/src/org/jetbrains/ktor/features/CallLogging.kt
```diff
@@ -6,19 +6,23 @@ import org.jetbrains.ktor.pipeline.*
 import org.jetbrains.ktor.request.*
 import org.jetbrains.ktor.util.*
 import org.slf4j.*
+import org.slf4j.event.*
 
-class CallLogging(private val log: Logger, private val monitor: ApplicationMonitor) {
-
-    class Configuration
+class CallLogging(private val log: Logger, private val monitor: ApplicationMonitor, configuration: Configuration) {
+    private val level = configuration.level
```
orangy: I would just pass level as a constructor parameter.

ktor-core/src/org/jetbrains/ktor/features/CallLogging.kt
```diff
@@ -35,24 +37,34 @@ class CallLogging(private val log: Logger, private val monitor: ApplicationMonit
         override val key: AttributeKey<CallLogging> = AttributeKey("Call Logging")
         override fun install(pipeline: Application, configure: Configuration.() -> Unit): CallLogging {
             val loggingPhase = PipelinePhase("Logging")
-            Configuration().apply(configure)
-            val feature = CallLogging(pipeline.log, pipeline.environment.monitor)
+            val configuration = Configuration().apply(configure)
+            val level = when {
+                configuration.level == Level.TRACE && pipeline.log.isTraceEnabled -> Level.TRACE
+                configuration.level == Level.DEBUG && pipeline.log.isDebugEnabled -> Level.DEBUG
+                configuration.level == Level.INFO && pipeline.log.isInfoEnabled -> Level.INFO
```
orangy: Disagree with the code. As it is written, if I set logging level in code to TRACE, but it will be disabled in logging config (say, for production deployment), it will suddenly throw and fail to start application. This is not what is expected.

ddsoyka:
Is this PR still relevant?

orangy:
Yes, but currently more as an issue than as actual code. I will implement it soon.

orangy:
Implemented in `calllogging` branch, closing this. Thanks for the input, I copied most of the code :)

PR #221, opened 2017-10-18
Add ktor-jackson support, based on contentNegotiation feature

triplem:
Since we do not need to add a new "feature" and there is no pull-request for this support thingy, I created this PR.


ktor-features/ktor-jackson/src/io/ktor/jackson/JacksonConverter.kt
```diff
@@ -0,0 +1,30 @@
...
+import io.ktor.application.*
+import io.ktor.content.*
+import io.ktor.features.*
+import io.ktor.http.*
+import io.ktor.pipeline.*
+import io.ktor.request.*
+
+/**
+ *    install(ContentNegotiation) {
+ *       register(ContentType.Application.Json, jacksonObjectMapper())
```
orangy: Shouldn't it create JacksonConverter with jacksonObjectMapper?

triplem:
Adopted the documentation to make the configuration clearer (as seen in the ktor-samples added by this PR as well).

orangy:
Thank you very much!

PR #224, opened 2017-10-20
Metrics improvement

### LINKED ISSUE #223: Metrics Improvements
Right now there are the following metrics available:

```
/ENDPOINT/(method:GET).200 OK
/ENDPOINT/(method:GET).meter
/ENDPOINT/(method:GET).timer
```

I would suggest to refactor these slightly to have the following names, to be able to parse those in an easy and nice way:

```
/method:GET/ENDPOINT/status.200 (so without OK)
/method:GET/ENDPOINT/meter
/method:GET/ENDPOINT/timer
```
This would allow to query for methods vs by endpoint as implemented right now and equals the defined ktor.calls.status metrics.

```
ktor.calls.status.200
```

triplem:
should fix #223. Furthermore it adds additional jvm specific metrics.

### DIFF
--- a/ktor-features/ktor-metrics/src/io/ktor/metrics/Metrics.kt
+++ b/ktor-features/ktor-metrics/src/io/ktor/metrics/Metrics.kt
@@ -1,22 +1,12 @@
 package io.ktor.metrics
 
-import com.codahale.metrics.JvmAttributeGaugeSet
-import com.codahale.metrics.Meter
-import com.codahale.metrics.MetricRegistry
-import com.codahale.metrics.Timer
-import com.codahale.metrics.jvm.FileDescriptorRatioGauge
-import com.codahale.metrics.jvm.GarbageCollectorMetricSet
-import com.codahale.metrics.jvm.MemoryUsageGaugeSet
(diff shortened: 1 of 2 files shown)


ktor-features/ktor-metrics/src/io/ktor/metrics/Metrics.kt
```diff
@@ -1,11 +1,22 @@
...
+import com.codahale.metrics.Timer
+import com.codahale.metrics.jvm.FileDescriptorRatioGauge
+import com.codahale.metrics.jvm.GarbageCollectorMetricSet
+import com.codahale.metrics.jvm.MemoryUsageGaugeSet
+import com.codahale.metrics.jvm.ThreadStatesGaugeSet
+import io.ktor.application.Application
+import io.ktor.application.ApplicationCall
+import io.ktor.application.ApplicationCallPipeline
+import io.ktor.application.ApplicationFeature
+import io.ktor.pipeline.PipelinePhase
```
orangy: please use star-imports

triplem:
Adopted. Now using star-imports.

PR #235, opened 2017-10-24
Rename annotation  ->  to keep up with capitalized @Annotation convention

### LINKED ISSUE #227: Rename Annotation @location to @Location
To following Coding Conventions, types should be UpperCase.

kimar:
Fixes #227 

I know this will introduce breaking changes to existing users but better break 'em know than later.


cy6erGn0m:
I'd avoid changes in run configurations. Also it would be nice to have deprecated lower-case annotation typealias for better migration (with the corresponding `ReplaceWith`)

orangy:
Merged manually with a typealias

PR #239, opened 2017-10-28
Extract GSON serializer from ktor-client-json as ktor-client-gson

Prototik:
1. GSON serializer extracted as ktor-client-gson
2. Added ktor-client-jackson with jackson-based serializer


Prototik:
@e5l any progress?

e5l:
We plan to make a common content negotiation feature with shared content converters. Suspended for now.

PR #240, opened 2017-10-28
Honor Accept-Charset during content negotiation

Prototik:


ktor-server/ktor-server-core/src/io/ktor/features/ContentNegotiation.kt
```diff
@@ -53,11 +44,22 @@ class ContentNegotiation(val converters: List<ContentConverterRegistration>) {
             pipeline.sendPipeline.intercept(ApplicationSendPipeline.Render) {
                 if (subject is OutgoingContent) return@intercept
 
-                val suitableConverters = acceptedTypes().mapNotNull { (contentType, _) ->
+                val suitableConverters = call.request.acceptItems().mapNotNull { (contentType, _) ->
```
cy6erGn0m: nice catch

ktor-server/ktor-server-core/src/io/ktor/features/ContentNegotiation.kt
```diff
@@ -53,11 +44,22 @@ class ContentNegotiation(val converters: List<ContentConverterRegistration>) {
...
                     feature.converters.firstOrNull { it.contentType.match(contentType) }
                 }
 
-                val converted = suitableConverters.mapNotNull { it.converter.convertForSend(this, subject) }.firstOrNull()
+                val charset = call.request.acceptCharsetItems().mapNotNull { (charset, _) ->
+                    when {
+                        charset == "*" -> Charsets.UTF_8
+                        Charset.isSupported(charset) -> Charset.forName(charset)
```
cy6erGn0m: too expensive, need to find first supported, don't need to map every provided charset, consider use `firstOrNull { }` and then convert single value

ktor-server/ktor-server-core/src/io/ktor/features/ContentNegotiation.kt
```diff
@@ -53,11 +44,22 @@ class ContentNegotiation(val converters: List<ContentConverterRegistration>) {
...
+                        else -> null
+                    }
+                }.firstOrNull() ?: Charsets.UTF_8
+
+                val converted = suitableConverters.mapNotNull {
+                    val contentType = it.contentType.withCharset(charset)
+                    it.converter.convertForSend(this, contentType, subject)
```
cy6erGn0m: Same here: running every suitable converter could be too slow, need to find first, most likely we need old-school for loop here

ktor-server/ktor-server-core/src/io/ktor/features/ContentNegotiation.kt
```diff
@@ -73,32 +75,24 @@ class ContentNegotiation(val converters: List<ContentConverterRegistration>) {
...
 
 interface ContentConverter {
-    suspend fun convertForSend(context: PipelineContext<Any, ApplicationCall>, value: Any): Any?
+    suspend fun convertForSend(context: PipelineContext<Any, ApplicationCall>, contentType: ContentType, value: Any): Any?
     suspend fun convertForReceive(context: PipelineContext<ApplicationReceiveRequest, ApplicationCall>): Any?
 }
 
-class ConvertedContent(text: String, val contentType: ContentType) : OutgoingContent.ByteArrayContent() {
-    private val bytes = text.toByteArray(Charsets.UTF_8)
+class ConvertedContent(val text: String, val contentType: ContentType, override val status: HttpStatusCode? = null) : OutgoingContent.ByteArrayContent() {
```
cy6erGn0m: It is not good at all: we should keep only one of them at a time: only text or only bytes: we shouldn't keep both as it wastes memory

cy6erGn0m: Status code is not a part of content conversion logic, it shouldn't be here

ktor-server/ktor-server-core/src/io/ktor/features/ContentNegotiation.kt
```diff
@@ -73,32 +75,24 @@ class ContentNegotiation(val converters: List<ContentConverterRegistration>) {
...
+    private val bytes by lazy { text.toByteArray(contentType.charset() ?: Charsets.UTF_8) }
```
cy6erGn0m: text will be converted to bytes anyway (unless gets filtered out) so I see no reason to use lazy delegation here as it costs few more object instances

ktor-server/ktor-server-core/src/io/ktor/features/ContentNegotiation.kt
```diff
@@ -53,11 +44,22 @@ class ContentNegotiation(val converters: List<ContentConverterRegistration>) {
             pipeline.sendPipeline.intercept(ApplicationSendPipeline.Render) {
                 if (subject is OutgoingContent) return@intercept
 
-                val suitableConverters = acceptedTypes().mapNotNull { (contentType, _) ->
+                val suitableConverters = call.request.acceptItems().mapNotNull { (contentType, _) ->
                     feature.converters.firstOrNull { it.contentType.match(contentType) }
                 }
 
-                val converted = suitableConverters.mapNotNull { it.converter.convertForSend(this, subject) }.firstOrNull()
+                val charset = call.request.acceptCharsetItems().mapNotNull { (charset, _) ->
```
cy6erGn0m: The main issue here is that `charset` is only relevant for text-producing content converters (plain-text, json) but it could be completely irrelevant for others (for example Java Serialization). Parsing, sorting and enumerating character encodings is releatively expensive and most likely need to be done in content converters rather than here

ktor-server/ktor-server-core/src/io/ktor/features/ContentNegotiation.kt
Prototik: It may be a part of conversion logic, for example converter may reply with 204 No Content status

ktor-server/ktor-server-core/src/io/ktor/features/ContentNegotiation.kt
Prototik: It's almost direct copy of TextContent. Although I don't see any reason for introducing new OutgoingContent classes, especially tuned for text-only representation. May be better solution is completely remove ConvertedContent and delegate choosing correct classes to converters?

Prototik:
@cy6erGn0m most issues are fixed

cy6erGn0m:
Thanks! Merged manyally

PR #273, opened 2017-12-11
A bit of top-level docs for CallLogging & DefaultHeaders features

elizarov:
(they are used in HelloApplication sample)

### DIFF
--- a/ktor-server/ktor-server-core/src/io/ktor/features/DefaultHeaders.kt
+++ b/ktor-server/ktor-server-core/src/io/ktor/features/DefaultHeaders.kt
@@ -1,11 +1,17 @@
 package io.ktor.features
 
 import io.ktor.application.*
-import io.ktor.http.*
-import io.ktor.response.*
-import io.ktor.util.*
+import io.ktor.http.HttpHeaders
+import io.ktor.http.toHttpDateString
+import io.ktor.response.header
+import io.ktor.util.AttributeKey
+import io.ktor.util.ValuesMapBuilder
 import java.time.*
 
+/**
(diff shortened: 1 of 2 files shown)


ktor-server/ktor-server-core/src/io/ktor/features/DefaultHeaders.kt
```diff
@@ -15,8 +19,15 @@ class DefaultHeaders(config: Configuration) {
...
     @Volatile private var cachedDateText: String = ZonedDateTime.now(ZoneOffset.UTC).toHttpDateString()
 
+    /**
+     * Configuration for [DefaultHeaders] feature.
+     */
     class Configuration {
         val headers = ValuesMapBuilder()
+
+        /**
+         * Adds standard header property [name] with the specified[value]
```
orangy: Space before `[value]`

orangy: Also inconsistent use of _period_ at the end of the text, other docs in this PR have them. I wonder what should be the global style on this. Should we end all API docs with period or not?

elizarov: Yes. IMHO, we should use period.

elizarov:
Updated

orangy:
Now, on the commit message. See https://chris.beams.io/posts/git-commit/ for the overall idea about how we try (but not always follow) to phrase commit messages. It should be something like "Document CallLogging & DefaultHeaders features".

PR #276, opened 2017-12-18
(fix) Incorrect credential check in ktor-sample-kweet app

spitty:
Without this patch any credential check fails.


cy6erGn0m:
Thanks!

Issue #206, opened 2017-09-15
418, I'm a teapot not supported


radekkut:
org.jetbrains.ktor.http.HttpStatusCode does not contain the 418 I'm a teapot status code.

hhariri:
To be honest, this is not part of the specification, and recently there has been concerns about people actually using this code for the wrong reason. It was merely a joke when it came out. I'd be hesitant in adding it. @orangy ?

cy6erGn0m:
418 status code is described in RFC 2324 (Hyper Text Coffee Pot Control Protocol) and it's April Fools joke.

orangy:
Also, you can add it as an extension property

    val HttpStatusCode.Companion.IAmATeaPot get() = HttpStatusCode(418, "I'm a tea pot")

deviant-studio:
somebody should publish the lib with such extension!

radekkut:
I dont think one line of code is worth publishing a lib

y9san9:
Sorry for nekroposting, but 418 was formally reserved to have no other use in foreseeable future, so it is definitely safe to include that to a library (https://www.rfc-editor.org/rfc/rfc9110#name-418-unused). I believe it should be up to user whether to use that code actually or not. A lot of http libraries include that code. But I don't care much, since, again, you can define this as an extension function.

PR #297, opened 2018-01-24
Add missing 418 status code

Juici:
Add missing 418 status code to match [RFC 2324](https://tools.ietf.org/html/rfc2324#section-2.3.2).


cy6erGn0m:
Hm, it was already rejected before. There are two reasons for that:
- RFC2324 is a Fool's day joke 
- You can always declare this StatusCode via extension property in your code

Juici:
It would be nice to have as a little Easter egg though.

kkriske:
#206

PR #303, opened 2018-01-29
Mark dsl functions with annotations

pTalanov:
This is a part of proposal to provide custom highlighting and improved completion in the IDE
Those annotations are needed to distinguish between 'dsl' and 'non-dsl' functions


orangy:
Merged manually

PR #309, opened 2018-01-31
Fix feature Compression always set Transfer-Encoding as chunked

xxxifan:
Compression feature did not implement `contentLength` so each response will set `Transfer-Encoding=chunked` instead of correct `Content-Length`.

### DIFF
--- a/ktor-server/ktor-server-core/src/io/ktor/features/Compression.kt
+++ b/ktor-server/ktor-server-core/src/io/ktor/features/Compression.kt
@@ -116,6 +116,7 @@ class Compression(compression: Configuration) {
         }
 
         override val contentType: ContentType? get() = original.contentType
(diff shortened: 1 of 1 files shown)


cy6erGn0m:
This is not correct as Content-Length should be set to compressed content size that is not known before content compressed and for sure it is not equal to origin's content length

xxxifan:
Sorry I have some misunderstanding with http compresssion . But it's weird. OkHttp add `Accept-Encoding: gzip` by default, so it's expecting a compressed response, but for some reason I need response content length and visible content, So it should not use compression in api requests and enable in partial contents.

cy6erGn0m:
You can configure predicates for compression feature to disable it for api (there are filters by content-type)

xxxifan:
Thank you, then I'll close this pr

Issue #307, opened 2018-01-31
Allow ContentNegotiation to match against missing Accept header


remen:
Given the following snippet:

```kotlin
embeddedServer(Netty, 8080) {
    install(ContentNegotiation) {
        register(ContentType.Any, JacksonConverter())
    }

    routing {
        get("/") {
            call.respond(mapOf("message" to "Hello World"))
        }
    }
}.start(wait = true)
```

It would be convenient to be able to send requests without an `Accept` header and have it match against `ContentType.Any`.

Currently, I get the following:

```bash
❯ curl -i -H 'Accept:' localhost:8080
HTTP/1.1 406 Not Acceptable
Content-Length: 0
```

Would it make sense to change this to support missing Accept headers? Would you appreciate a PR?

orangy:
Empty `Accept` is not any. Any is `*/*`.

remen:
I see.

How would you recommend I deal with the example above? I want to serve `application/json` by default if the client does not supply a `Accept` header.

orangy:
Hmm, looks like it's not possible right now. But that's good suggestion, I think `Any` can be handled without `Accept` header indeed.

remen:
I can send you a PR if you'd like, I think a reasonable place to change is at https://github.com/ktorio/ktor/blob/8fbae7d508660538ebd3dc7cb139217c9b1ceaad/ktor-server/ktor-server-core/src/io/ktor/features/ContentNegotiation.kt#L68
```kotlin
                val suitableConverters = call.request.acceptItems().mapNotNull { (contentType, _) ->
                    feature.registrations.firstOrNull {
                        it.contentType.match(contentType)
                    }
                }
```

where we could choose to have a "special" path if acceptItems is an empty list.

orangy:
Exactly, PRs are very welcome!

remen:
See PR above.

orangy:
Fixed

PR #312, opened 2018-02-02
Allow missing Accept header in ContentNegotiation

remen:
According to https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html:

> If no Accept header field is present, then it is assumed that the client accepts all media types.

This commit makes ContentNegotiation treat a missing `Accept` header as if the client requested `Accept: */*`

This fixes: #307

### DIFF
--- a/ktor-server/ktor-server-core/src/io/ktor/features/ContentNegotiation.kt
+++ b/ktor-server/ktor-server-core/src/io/ktor/features/ContentNegotiation.kt
@@ -65,15 +65,16 @@ class ContentNegotiation(val registrations: List<ConverterRegistration>) {
             pipeline.sendPipeline.intercept(ApplicationSendPipeline.Render) {
                 if (subject is OutgoingContent) return@intercept
 
-                var acceptContentTypes = call.request.acceptItems().map { it.value }
-                // From https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html
-                // "If no Accept header field is present, then it is assumed that the client accepts all media types."
-                if(acceptContentTypes.isEmpty()) {
-                    acceptContentTypes = listOf("*/*")
-                }
-                val suitableConverters = acceptContentTypes.mapNotNull { contentType ->
-                    feature.registrations.firstOrNull {
-                        it.contentType.match(contentType)
+                val acceptItems = call.request.acceptItems()
+                val suitableConverters = if (acceptItems.isEmpty()) {
+                    // From https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html
+                    // "If no Accept header field is present, then it is assumed that the client accepts all media types."
+                    feature.registrations
+                } else {
+                    acceptItems.mapNotNull { (contentType, _) ->
+                        feature.registrations.firstOrNull {
+                            it.contentType.match(contentType)
+                        }
                     }
                 }
 
--- a/ktor-server/ktor-server-core/src/io/ktor/features/ContentNegotiation.kt
+++ b/ktor-server/ktor-server-core/src/io/ktor/features/ContentNegotiation.kt
@@ -65,7 +65,13 @@ class ContentNegotiation(val registrations: List<ConverterRegistration>) {
             pipeline.sendPipeline.intercept(ApplicationSendPipeline.Render) {
                 if (subject is OutgoingContent) return@intercept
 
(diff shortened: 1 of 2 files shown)


ktor-server/ktor-server-core/src/io/ktor/features/ContentNegotiation.kt
```diff
@@ -65,7 +65,13 @@ class ContentNegotiation(val registrations: List<ConverterRegistration>) {
             pipeline.sendPipeline.intercept(ApplicationSendPipeline.Render) {
                 if (subject is OutgoingContent) return@intercept
 
-                val suitableConverters = call.request.acceptItems().mapNotNull { (contentType, _) ->
+                var acceptContentTypes = call.request.acceptItems().map { it.value }
+                // From https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html
+                // "If no Accept header field is present, then it is assumed that the client accepts all media types."
+                if(acceptContentTypes.isEmpty()) {
```
orangy: I think it would be better to just use `firstOrNull` on registrations instead of going through all the hoops to match `*/*`

remen: Makes sense. See new commit.

remen: Note that we still need to run every suitableRegistration until we've found a corresponding converter that can handle the value.

remen: Hmm... nevermind, I think I need to look at the logic more carefully. I'll get back.

remen: Ok, I had another look. Which do you prefer from the following two alternatives?

```kotlin
val acceptItems = call.request.acceptItems()
val suitableConverters = if (acceptItems.isEmpty()) {
    feature.registrations.take(1)
} else {
    acceptItems.mapNotNull { (contentType, _) ->
        feature.registrations.firstOrNull {
            it.contentType.match(contentType)
        }
    }
}

// Pick the first one that can convert the subject successfully
val converted = suitableConverters.mapUntilNotNull {
    it.converter.convertForSend(this, it.contentType, subject)
}
```

```kotlin
val acceptItems = call.request.acceptItems()
val suitableConverters = if (acceptItems.isEmpty()) {
    feature.registrations
} else {
    acceptItems.flatMap { (contentType, _) ->  // flatMap here
        feature.registrations.filter {  // filter here
            it.contentType.match(contentType)
        }
    }
}

// Pick the first one that can convert the subject successfully
val converted = suitableConverters.mapUntilNotNull {
    it.converter.convertForSend(this, it.contentType, subject)
}
```

The second alternative seems more general to me. We will try all registered converters that matches the Accept field, and pick the first one that succeeds.

In the first case, we will only try to convert with the *first* converter (for each Accept-statement).

What do you think?

remen:
Have you had a chance to have look at the change above? Also, repeating a comment here that is no longer visible due to outdated changes:

Which do you prefer from the following two alternatives?

```kotlin
val acceptItems = call.request.acceptItems()
val suitableConverters = if (acceptItems.isEmpty()) {
    feature.registrations.take(1)
} else {
    acceptItems.mapNotNull { (contentType, _) ->
        feature.registrations.firstOrNull {
            it.contentType.match(contentType)
        }
    }
}

// Pick the first one that can convert the subject successfully
val converted = suitableConverters.mapUntilNotNull {
    it.converter.convertForSend(this, it.contentType, subject)
}
```

```kotlin
val acceptItems = call.request.acceptItems()
val suitableConverters = if (acceptItems.isEmpty()) {
    feature.registrations
} else {
    acceptItems.flatMap { (contentType, _) ->  // flatMap here
        feature.registrations.filter {  // filter here
            it.contentType.match(contentType)
        }
    }
}

// Pick the first one that can convert the subject successfully
val converted = suitableConverters.mapUntilNotNull {
    it.converter.convertForSend(this, it.contentType, subject)
}
```

The second alternative seems more general to me. We will try all registered converters that matches the Accept field, and pick the first one that succeeds.

In the first case, we will only try to convert with the *first* converter (for each Accept-statement).

What do you think?

orangy:
I've been traveling last week, I will review everything during coming week.

orangy:
Merged it manually with slight modification (distinct) to the second variant. Thanks a lot!

PR #176, opened 2017-08-09
Adding readme file with usage instructions

McPringle:
Yesterday at the Hackergarten in Bern it took us a while to figure out how to use `ktor` with `GsonSupport` in an embedded server. So I decided to start writing a readme file with basic usage instructions. Hope you'll find it useful…

---
This change is [Reviewable](https://reviewable.io/reviews/kotlin/ktor/176)

PR #316, opened 2018-02-05
Adding readme file with gson usage instructions

soywiz:
This continues #176 since it was outdated


soywiz:
Squashed and rebased. Thanks @McPringle !

PR #331, opened 2018-02-21
Fixes imports in the example on README

sandeepraju:
The example on the README no longer works. This commit updates the imports by referring to the example here: http://ktor.io/quickstart/application.html

With this fix, the example on the README should work now!


soywiz:
Confirmed. Thanks!

PR #332, opened 2018-02-22
Added Kdoc for experimental locations feature

soywiz:


cy6erGn0m:
`LocationSamples.kt` shouldn't be in the production sources root. Is there way to move it to test that wouldn't break `@sample` kdoc tag?

soywiz:
I have deleted the `@sample` tag for now –squashing the changes in this branch–

Once I discover how to properly add sample sources without actually including them in the main artifact, I will create a separate PR with that.

PR #356, opened 2018-03-10
Set cookie's domain and path on sessions.clear()

### LINKED ISSUE #259: Cookie's path can be controlled for creation, but not for deletion
In order to delete a cookie you've to use the same path that is stored with the cookie on the client browser. Ktor let us control the path with cookie creation. Here's an example configuration:

    install(Sessions) {
        cookie<MySession>(sessionMarker) {
            cookie.duration = cookieDuration
            cookie.path = "/myWebapp"
            transform(SessionTransportTransformerMessageAuthentication(sessionKey))
        }
    }

On the login page `/myWebapp/page/login`, the cookie is created:

    call.sessions.set(MySession([some arguments]))

The cookie shows the path `/myWebapp` when inspected on the client. The default would be `/myWebapp/page` because that's the path of the login page, but the path is taken from the cookie configuration shown above.

Eventually the user wants to log out via a request to the login page:

    call.sessions.clear<MySession>()

There's no way of controlling the cookie's path here. Because the login page `/myWebapp/page/login` is sending the expired cookie back to the client for another path (its own `/myWebapp/page`), the original cookie is not deleted (`/myWebapp/page` != `/myWebapp`).

**Edit**: I would expect that the call to `clear()` uses the same path that was used for the cookie's creation.

andgein:
Fixes #259

### DIFF
--- a/ktor-server/ktor-server-core/src/io/ktor/sessions/SessionTransportCookie.kt
+++ b/ktor-server/ktor-server-core/src/io/ktor/sessions/SessionTransportCookie.kt
@@ -39,7 +39,7 @@ class SessionTransportCookie(val name: String,
     }
 
     override fun clear(call: ApplicationCall) {
-        call.response.cookies.appendExpired(name)
+        call.response.cookies.appendExpired(name, configuration.domain, configuration.path)
     }
 }
 


cy6erGn0m:
Would be nice to add a test to ensure that it generates proper `Set-Cookie` header. See `SessionTest.kt`

cy6erGn0m:
Ok, will add it myself, thanks!

PR #370, opened 2018-03-27
Fix bug with double encoding of OAuth2 scopes

burkov:
Fixes a bug with double encoding of OAuth2 scope parameter


cy6erGn0m:
Thanks!

PR #380, opened 2018-04-05
Digest Authentication not working in Chrome due to not following RFC 2617

petertulala:
According to **RFC 2617 (Section 3.2.2.1)**, if `qop` is not present, it should follow the older **RFC 2069** and the response should be in form of:

`MD5(HA1:nonce:HA2)`

 instead of:

`MD5(HA1:nonce:nonceCount:cnonce:qop:HA2)`

The current implementation of Digest Authorization does not work in Chrome 65.0.3325.181 due to not following the RFC.


cy6erGn0m:
Thanks!

PR #398, opened 2018-05-07
JWT: Support multiple schemes in addition to Bearer

### LINKED ISSUE #397: JWT custom Authentication Scheme Name
Right now JWT authentication only handles "Bearer" authentication types:
https://github.com/ktorio/ktor/blob/4196d2818ccd5cdbafbd412f414c642b64ddc5fd/ktor-features/ktor-auth-jwt/src/io/ktor/auth/jwt/JWTAuth.kt#L111

It happens that instead of Bearer, in the realworld.io sample, it uses `Authentication: Token ...`
https://github.com/gothinkster/realworld/blob/master/api/Conduit.postman_collection.json#L203

I'm not sure if it is supported:
http://www.iana.org/assignments/http-authschemes/http-authschemes.xhtml

Still. Would be nice to allow to change the supported auth scheme or just load them all.

soywiz:
This should fix #397 required for supporting realworld with ktor https://github.com/gothinkster/node-express-realworld-example-app/blob/ba04b70c31af81ca7935096740a6e083563b3a4a/routes/auth.js#L5

### DIFF
--- a/ktor-features/ktor-auth-jwt/src/io/ktor/auth/jwt/JWTAuth.kt
+++ b/ktor-features/ktor-auth-jwt/src/io/ktor/auth/jwt/JWTAuth.kt
@@ -21,16 +21,16 @@ class JWTPrincipal(val payload: Payload) : Principal
 class JWTAuthenticationProvider(name: String?) : AuthenticationProvider(name) {
     internal var authenticationFunction: suspend ApplicationCall.(JWTCredential) -> Principal? = { null }
 
-    internal var supportedAuthSchemes = setOf("Bearer")
+    internal var schemes = JWTAuthSchemes("Bearer")
     var realm: String = "Ktor Server"
     internal var verifier: ((HttpAuthHeader?) -> JWTVerifier?) = { null }
 
     /**
-     * @param [schemes] list of supported authentication schemes for JWT (by default only "Bearer")
+     * @param [defaultScheme] default scheme that will be used to challenge the client when no valid auth is provided
+     * @param [additionalSchemes] additional schemes that will be accepted when validating the authentication
      */
-    fun authSchemes(vararg schemes: String = arrayOf("Bearer")) {
-        require(schemes.isNotEmpty()) { "At least one scheme should be provided" }
-        supportedAuthSchemes = schemes.toSet()
+    fun authSchemes(defaultScheme: String = "Bearer", vararg additionalSchemes: String) {
+        schemes = JWTAuthSchemes(defaultScheme, *additionalSchemes)
     }
 
     /**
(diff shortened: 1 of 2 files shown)


ktor-features/ktor-auth-jwt/src/io/ktor/auth/jwt/JWTAuth.kt
```diff
@@ -107,8 +116,8 @@ private suspend fun verifyAndValidate(call: ApplicationCall, jwtVerifier: JWTVer
     return validate(call, credentials)
 }
 
-private fun HttpAuthHeader?.getBlob() = when {
-    this is HttpAuthHeader.Single && authScheme == "Bearer" -> blob
+private fun HttpAuthHeader?.getBlob(supportedAuthSchemes: Set<String>) = when {
+    this is HttpAuthHeader.Single && authScheme in supportedAuthSchemes -> blob
```
cy6erGn0m: should it be case sensitive?

ktor-features/ktor-auth-jwt/src/io/ktor/auth/jwt/JWTAuth.kt
```diff
@@ -21,9 +21,17 @@ class JWTPrincipal(val payload: Payload) : Principal
...
     internal var authenticationFunction: suspend ApplicationCall.(JWTCredential) -> Principal? = { null }
 
+    internal var supportedAuthSchemes = setOf("Bearer")
     var realm: String = "Ktor Server"
     internal var verifier: ((HttpAuthHeader?) -> JWTVerifier?) = { null }
 
+    /**
+     * @param [supported] list of supported authentication schemes for JWT (by default only "Bearer")
+     */
+    fun authSchemes(vararg schemes: String = arrayOf("Bearer")) {
```
cy6erGn0m: `require(schemes.isNotEmpty()) { " .... " }`

ktor-features/ktor-auth-jwt/src/io/ktor/auth/jwt/JWTAuth.kt
```diff
@@ -118,8 +127,8 @@ private fun ApplicationRequest.parseAuthorizationHeaderOrNull() = try {
     null
 }
 
-private fun HttpAuthHeader.Companion.bearerAuthChallenge(realm: String): HttpAuthHeader =
-        HttpAuthHeader.Parameterized("Bearer", mapOf(HttpAuthHeader.Parameters.Realm to realm))
+private fun HttpAuthHeader.Companion.bearerAuthChallenge(realm: String, supportedAuthSchemes: Set<String>): HttpAuthHeader =
+        HttpAuthHeader.Parameterized(supportedAuthSchemes.first(), mapOf(HttpAuthHeader.Parameters.Realm to realm))
```
cy6erGn0m: it looks like we need to produce multiple auth challenges, not just for the first one

ktor-features/ktor-auth-jwt/src/io/ktor/auth/jwt/JWTAuth.kt
soywiz: Yes. It seems so: https://tools.ietf.org/html/rfc7235#section-2.1

soywiz:
I have done some changes. Please re-review. 

> it looks like we need to produce multiple auth challenges, not just for the first one

I don't know about a reasonable way for providing different schemes there. So what I have done in the changes is to make one scheme the default one (the expected for this authentication instead of arbitrarily getting the first one of the set).

Issue #414, opened 2018-05-17
Basic Authentication provider fails on invalid Authorization header


fabianishere:
Routes protected behind the Basic Authentication Provider can fail when an invalid Authorization header is sent with the request. 

For example, sending a request with the header `Authorization: Bearer some-token` to a Basic-protected endpoint will crash the request handler, respond with a 500 Internal Server Error and spit the following error in the console:

> java.lang.IllegalArgumentException: Illegal base64 character 2e
	at java.util.Base64$Decoder.decode0(Base64.java:714)
	at java.util.Base64$Decoder.decode(Base64.java:526)
	at java.util.Base64$Decoder.decode(Base64.java:549)
	at io.ktor.auth.BasicAuthKt.basicAuthenticationCredentials(BasicAuth.kt:75)
       ...

Is this expected behaviour? I have added two test cases and a possible fix at 1d56c52  on the `bug/auth-basic-invalid-scheme` branch of my fork.

fabianishere:
@cy6erGn0m Do you want me to split the changes into two commits perhaps?

fabianishere:
See #415 for the pull request.

PR #415, opened 2018-05-18
Validate scheme and contents in authorization header

fabianishere:
This change adds validation for the scheme and contents when trying to
parse the HTTP Basic credentials from the Authorization HTTP header.
This prevents the HTTP Basic protected routes from crashing when the
user sends an Authorization header with a scheme other than "Basic" or
when the user sends a malformed Base64 string.

See #414 for the responsible issue.


cy6erGn0m:
Thanks!

PR #416, opened 2018-05-18
Allow access to more `get` functions in configs

FredDeschenes:
Backing library allows getting more than `String`s, so should ktor.

I skipped `getBytes` because it's basically the same as `getMemorySize`, I can add it if you want.

This also removes the need for the `ApplicationConfigValue` interface as both the Hocon and map implementations use `HoconApplicationConfigValue` directly (Hocon's `ConfigFactory.parseMap` is used for the map implementation so the parsing of the config values should be exactly the same regardless of which implementation is used).

This should be a drop-in replacement but I might have missed something (I'm new to ktor so I might have missed some use cases).

See #422


FredDeschenes:
FYI I've updated the PR to fit with the latest Ktor version.

FredDeschenes:
Bump... Is this still something you might be interested in merging? I can fix the PR (again) if so, otherwise I'll just close it.

ktor-server/ktor-server-core/jvm/src/io/ktor/config/ApplicationConfig.kt
```diff
@@ -15,12 +20,12 @@ interface ApplicationConfig {
      * Get config property with [path] or fail
      * @throws ApplicationConfigurationException
      */
-    fun property(path: String): ApplicationConfigValue
+    fun property(path: String): HoconApplicationConfigValue
```
cy6erGn0m: Here we have abstraction so HOCON is just one of the implementations. We potentially can migrate to something else, for example, kotlinx.serialization. So we can't use this type here.

ktor-server/ktor-server-core/jvm/src/io/ktor/config/ApplicationConfig.kt
```diff
@@ -39,18 +44,160 @@ interface ApplicationConfig {
...
+     * Get property [Number] value
+     * @see [Config.getNumber]
+     */
+    fun getNumber(): Number = config.getNumber(path)
+
+    /**
+     * Get property [Int] value
+     * @see [Config.getInt]
+     */
+    fun getInt(): Int = config.getInt(path)
```
cy6erGn0m: I believe these functions would be just extension functions like `getString().toInt()`. For sure there is no need to have members like that unless there is something very special in number representation in configs, right?

ktor-server/ktor-server-core/jvm/src/io/ktor/config/ApplicationConfig.kt
```diff
@@ -39,18 +44,160 @@ interface ApplicationConfig {
...
+     * @param timeUnit The [TimeUnit] to get the duration as
+     * @see [Config.getDuration]
+     */
+    fun getDuration(timeUnit: TimeUnit): Long = config.getDuration(path, timeUnit)
+
+    /**
+     * Get property [Duration] value
+     * @see [Config.getDuration]
+     */
+    fun getDuration(): Duration = config.getDuration(path)
```
cy6erGn0m: Even more: `Duration` requires Android API level 26. This is quite problematic, unfortunately.

ktor-server/ktor-server-core/jvm/src/io/ktor/config/ApplicationConfig.kt
```diff
@@ -39,18 +44,160 @@ interface ApplicationConfig {
...
+
+    /**
+     * Get property [Boolean] list value
+     * @see [Config.getBooleanList]
+     */
+    fun getBooleanList(): List<Boolean> = config.getBooleanList(path)
```
cy6erGn0m: Not sure if it is commonly used. A configuration parameter with a boolean list would look like this: `param: true, false, false, true`. Confusing, isn't it?

ktor-server/ktor-server-core/jvm/src/io/ktor/config/ApplicationConfig.kt
FredDeschenes: Alright, I'll remove the deprecated `ApplicationConfig` typealias and reset these to it.

ktor-server/ktor-server-core/jvm/src/io/ktor/config/ApplicationConfig.kt
FredDeschenes: HOCON actually loads the configurations with the correct types, so an integer is loaded as an integer and kept as such. Currently every time you get an `int` in a config, it gets converted to a `String`, and then you do `toInt()` to convert it back into an integer (or whatever other type).

You can test this by putting a breakpoint in `SimpleConfig::findKeyOrNull`. `v` will be of type `ConfigInt`, which then gets converted to a `Quoted` (which is a wrapper around a string) in `DefaultTransformer.transform`.

ktor-server/ktor-server-core/jvm/src/io/ktor/config/ApplicationConfig.kt
FredDeschenes: Wow, I'm glad I don't have to deal with mobile stuff when I read stuff like this! I guess this one could be removed.

ktor-server/ktor-server-core/jvm/src/io/ktor/config/ApplicationConfig.kt
FredDeschenes: ¯\_(ツ)_/¯, it's there in HOCON so I put it there as well. I don't think anyone would miss it so I'll just remove it.

PR #420, opened 2018-05-21
Defaults charset for BasicAuth to UTF-8 and allows to customize it

### LINKED ISSUE #418: Wrong user credential for Basic Auth using Cyrillic
My client login is "Лира" and I get incorrect decoded name using a basicAuthenticationCredentials method. It leads to very strange behaviour in my app flow. Also it is not really obviously to understand what happens until I saw the Base64.getDecoder().decode(parsed.blob).toString(**Charsets.ISO_8859_1**). It broke my mind and I got disappointed.

Can we get any little fix for user name decoding?

soywiz:
This should fix #418

### DIFF
--- a/ktor-features/ktor-auth/src/io/ktor/auth/BasicAuth.kt
+++ b/ktor-features/ktor-auth/src/io/ktor/auth/BasicAuth.kt
@@ -3,6 +3,7 @@ package io.ktor.auth
 import io.ktor.application.*
 import io.ktor.request.*
 import io.ktor.response.*
+import java.nio.charset.*
 import java.util.*
 
 
@@ -19,6 +20,11 @@ class BasicAuthenticationProvider(name: String?) : AuthenticationProvider(name) 
     var realm: String = "Ktor Server"
 
     /**
+     * Specifies the charset to be used.
+     */
+    var charset: Charset = Charsets.UTF_8
+
+    /**
      * Sets a validation function that will check given [UserPasswordCredential] instance and return [Principal],
      * or null if credential does not correspond to an authenticated principal
      */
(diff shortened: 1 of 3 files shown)


cy6erGn0m:
Does it work with IE/Edge ? Does it support additional parameter?

soywiz:
Going to check Internet Explorer & Edge. My guess is that it won't work as expected in IE. For what I have read it sends one byte per character.
Not sure if the charset parameter in `WWW-Authentication` will cause problems with it. If that's the case we can make it nullable to allow people to not send it.

At any rate, for people wanting to support IE, they can set the charset to latin1 or US-ASCII. UTF-8 should work as well as long as people stick to 7-bit ASCII characters when using IE.

Going to check it in IE and Edge and will provide the results once tested.

soywiz:
Tried:
**IE6, IE11 and Edge** - Extra param doesn't cause issues, but it is ignored, those browsers seems to use latin1 in all cases (not sure if language/the default charset of the OS affect here; didn't tried with windows in other languages, only in english).
**Chrome, Firefox and Opera (latest version)** - Respect the charset set

Nullable charset to not emit the extra param could not be necessary since IE doesn't have issues with it.
It seems that you cannot put any non-latin characters with basic encoding with IE or Edge.

So my proposal here is to default it to null or to latin1, but still allow to set UTF8 for example in the case you don't need to support IE/Edge or it is good enough to only support US-ASCII in IE/Edge.

My personal vote goes for default charset=null to produce the previous result (do not add additional parameter but still use latin1 when decoding). What do you think?

cy6erGn0m:
Doesn't the new RFC define what we should do in this case? Perhaps we can force UTF-8 by default so non-latin logins will work if possible out of the box but provide ability to not send via null charset specified.

soywiz:
I have changed the charset property to be nullable, and restricted the value to UTF-8 or null (the spec says that the only acceptable value is UTF-8). Since it is unlikely to happen, maybe we can change the charset property to a mere boolean stating whether we are including the charset="UTF-8" or not.
I couldn't find any suggestion other than use UTF-8 that is a super-set of US-ASCII. But definitely it would break if some people are using extended characters from latin1 in IE/Edge like "á" or things like that.
The spec also says that some servers checks the user-agent to determine which charset to use.

PR #430, opened 2018-06-05
Don't hardcode url schemes in URLBuilder.takeFrom()

Prototik:
It leads to curious errors, e.g.:
```kotlin
URLBuilder().takeFrom("wss://localhost/path").buildString() == "wss://localhost:80/path"
```


cy6erGn0m:
thanks!

PR #433, opened 2018-06-06
Properly parse the multipart content type including its parameters

lion7:
If the boundary parameter in the content type header is quoted an exception is thrown when parsing the preamble. An example: `ContentType: multipart/form-data; boundary="some-boundary"`.

Parsing the content type header using the existing `ContentType.parse(String)` method properly handles quoted values and multiple parameters.

### DIFF
--- a/ktor-http-cio/src/io/ktor/http/cio/Multipart.kt
+++ b/ktor-http-cio/src/io/ktor/http/cio/Multipart.kt
@@ -1,5 +1,6 @@
 package io.ktor.http.cio
 
+import io.ktor.http.ContentType
 import io.ktor.http.cio.internals.*
 import kotlinx.coroutines.experimental.*
 import kotlinx.coroutines.experimental.channels.*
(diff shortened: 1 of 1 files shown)


lion7:
@e5l @cy6erGn0m Could 1 of you take a look at this PR?

ktor-http-cio/src/io/ktor/http/cio/Multipart.kt
```diff
@@ -148,10 +147,10 @@ fun parseMultipart(
     boundaryBytes.put(pch)
     boundaryBytes.put(pch)
 
-    for (i in 0 until boundaryLength) {
-        val ch = contentType[boundaryStart + i].toInt() and 0xffff
-        if (ch > 0x7f) throw IOException("Failed to parse multipart: wrong boundary byte 0x${ch.toString(16)} - should be 7bit character")
-        boundaryBytes.put(ch.toByte())
+    for (ch in boundaryParameter) {
+        val b =  ch.toByte()
+        if (b > 0x7f) throw IOException("Failed to parse multipart: wrong boundary byte 0x${b.toString(16)} - should be 7bit character")
```
soywiz: Byte is signed `[-128, 127]` -- `0x7f == 127`. So I guess it doesn't work as expected. Maybe a test verifying that this exception is thrown if not there already?

`if (b > 0x7f)` -> `if ((b.toInt() and 0xFF) > 0x7f)`
or
`if (b > 0x7f)` -> `if (b < 0)`

cy6erGn0m:
Parsing is improved in master. It doesn't use `ContentType.parse` because `parseMultipart` is quite low-level and I'd like to avoid making so much garbage at this level.

lion7:
I just checked your commit and it ~~looks~~ works great. It's indeed way more low-level, I guess that's needed for security and/or performance reasons?

Anyway, since my issue is now solved I will close this PR :smile:

Issue #434, opened 2018-06-07
when building a JWK verifier, the alg field should be optional.


mkporwit:
The JWK RFC (https://tools.ietf.org/html/rfc7517#section-4.4) specifies that the alg parameter is optional. However, JWTAuth.kt:getVerifier() throws a "java.lang.IllegalArgumentException: Unsupported algorithm null" when it tries to parse JWKS that do not define one (such as those at https://login.microsoftonline.com/common/discovery/v2.0/keys)

A proposed fix would be to either peek at the JWT and use the same algorithm as the JWT is signed with, or just default to RS256 if nothing is defined. That latter approach was taken by Spring, here: https://github.com/spring-projects/spring-security-oauth/commit/9b7d5767cb07a4e9d643d2d59843821a252f6fd1

mkporwit:
Fix has been merged into master and released in 0.9.3. Closing this issue.

PR #435, opened 2018-06-07
default to RS256 as algorithm if JWK does not optionally provide one (#434)

mkporwit:
Per RFC7517, alg should be optional. Spring fixes this by defaulting to RS256 if no alg is given.


cy6erGn0m:
Thanks!

PR #452, opened 2018-06-25
 Jackson uses `line.separator` by default causing issues in windows

soywiz:
Proposal: for a web server I think it is better to provide consistency by default. So this proposal overrides the Jackson default to use `LF`(`\n`) instead of platform dependant.


cy6erGn0m:
Thanks!

PR #457, opened 2018-07-01
Cleanup settings.gradle

soywiz:


e5l:
LGTM, @cy6erGn0m could you take a look?

cy6erGn0m:
The only possible issue is that IDEA's module creation wizard could get crazy because of this

Issue #466, opened 2018-07-05
Add support for the SameSite cookie attribute


mdr:
It would be useful if ktor was able to set the `SameSite` attribute on a cookie, including as part of the sessions support (`CookieIdSessionBuilder`).

https://www.owasp.org/index.php/SameSite

marshallpierce:
Since this is a normal "key-value" cookie attribute (unlike, say, httponly), you can do this with the existing `extensions` mechanism:

```
cookie.extensions["SameSite"] = "lax"
```

I wouldn't object if there was a type safe way to do that, but I don't have a problem with the above either.

oleg-larshin:
Please check the <a href="https://youtrack.jetbrains.com/issue/KTOR-770">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

PR #470, opened 2018-07-06
SameSite cookie extension. Implements #466

soywiz:


cy6erGn0m:
Shouldn't it be on cookie itself rather than sessions?

soywiz:
Sorry for the delay. It seems that I missed this.

Even if that code is inside `SessionTransportCookie.kt` it is in the class `CookieConfiguration`, that is a builder for cookies. Since that property is mutable, it has to be in t he builder since `Cookie` is immutable.

I still kept the `sameSite` there but moved the enum outside so can be used by `Cookie` too.

Also I have done a refactor to move the building cookie from configuration a proper method inside `CookieConfiguration` so it is reusable.

PR #473, opened 2018-07-09
RateLimits Feature

Shengaero:
Adds a new ktor server feature module with support for rate-limiting requests, with exposed support for things like storing rate limit data and response customization.

This feature also comes with a set of extensions for `Route` instances that allow easy appending of rate-limited routes:
```kotlin
install(RateLimits)
routing {
    rateLimit("/hello", limit = 5, seconds = 5) {
        get {
            // ratelimited GET - "/hello"
        }
        post {
            // ratelimited POST- "/hello"
        }
    }
    route("/goodbye") {
        rateLimit(limit = 5, seconds = 5) {
            get {
                // ratelimited GET - "/goodbye"
            }
        }
        post {
            // not ratelimited POST - "/goodbye"
        }
    }
}
```

The only unfortunate circumstance of my implementation is that I couldn't figure out a way to override parent RateLimits, and as such the following is invalid and will throw an `IllegalArgumentException`:
```kotlin
routing {
    rateLimit("/foo", limit = 5, seconds = 5) {
        // inner rateLimit, IllegalArgumentException will be thrown.
        rateLimit("/bar", limit = 10, seconds = 10) {
            // ...
        }
    }
}
```


cy6erGn0m:
Looks promising. There are open questions:
- concurrency: for sure it will not work under load as controller's map is not concurrent-safe
- performance: we instantiate a lot of stuff for every request, many attributes and so on so it will significantly slow-down execution. So is there way to reduce the cost?

qwwdfsad:
The whole change looks like an attempt to make Ktor-specific Guava-like rate limiter.

Maybe it's better to implement `suspend` based `RateLimiter` with similar to Guava API?
Then `rateLimit {}` DSL could be just a handy wrapper around limiter.

Later `RateLimiter` could be extracted to `kotlinx.coroutines`as it's generally useful.

Shengaero:
@cy6erGn0m
When you say concurrency, I assume you're referring to the `MapRateLimitController`. If you are, I should probably have left a more detailed comment above there, because the `MapRateLimitController` is just a "default", really only useful in testing scenarios. I think you're correct it should use a `ConcurrentHashMap` instead of a `mutableMapOf`, but I wouldn't advise anyone use it in production, as library users should implement their own `RateLimitController` to manage stateless storage in a database for example.
I was more concerned with not imposing a bias on the feature user, it's one of the reasons why an implementable `RateLimitController` exists in the first place. Maybe it might be better to rename it? Possibly `RateLimitManager` or `RateLimitTransport`?

With regards to creating many `RateLimit` instances, when I was designing this feature, I assumed it would be more effective in achieving concurrency-safety when objects are not changing state per request, as mutating `RateLimit` instances in multiple places at the same time would already present some issues, however I'd be up for changing some things if anyone has ideas on how to do it! 😃 

As for the attribute storage, I suppose I could try and figure out a way to remove the addition of the `Instant` attribute stored [here](https://github.com/Shengaero/ktor/blob/feature/ratelimits/ktor-features/ktor-server-ratelimits/src/io/ktor/ratelimits/RateLimits.kt#L107), but removing the `RateLimit` stored in the call attributes would mean remodeling the extension property [here](https://github.com/Shengaero/ktor/blob/feature/ratelimits/ktor-features/ktor-server-ratelimits/src/io/ktor/ratelimits/extensions.kt#L162), and I can't think of how I'd do that.

@qwwdfsad 
I had not considered guava, mostly because I'm not familiar with it as of right now, but guava's [RateLimiter](https://github.com/google/guava/blob/master/guava/src/com/google/common/util/concurrent/RateLimiter.java) has some downsides that might make it less safe to use. For one, it is still marked with `@Beta`, meaning it's subject to change at any moment. Also I am would be hesitant about using any APIs built for threads and thread blocking considering that blocking coroutines is resource consuming.

Not to mention I am led to believe from the class documentation of the guava `RateLimiter` that the API is more geared towards client usage. Servers should not delay their responses to an exceeded RateLimit to their clients, and should immediately send back a `429 - Too Many Requests`. This is the reason why the feature intercepts the ApplicationCallPipeline *before* the actual call logic.

qwwdfsad:
@Shengaero yes, it's `@Beta` for six years already. `RateLimiter` is battle-tested in a lot of projects including Spark and Cassandra. It's not something they can easily change


>Also I am would be hesitant about using any APIs built for threads and thread blocking considering that blocking coroutines is resource consuming

That's not what I've proposed. I think `suspend` (as opposed to blocking) based `RateLimiter` should be implemented, which is exact the opposite of thread blocking.

>Servers should not delay their responses to an exceeded RateLimit to their clients

It's implementable by replacing `acquire` with `tryAcquire`:
```
if (limiter.tryAcquire()) respond()
else respondTooManyRequests()
```

Shengaero:
@qwwdfsad 
I apologize, again, I am not at all familiar with guava, but I appreciate the elaboration.
As for the implementation, if kotlinx.coroutines were to implement their own rate-limiter, I would jump on that in a heartbeat as opposed to my own implementation. 😄

qwwdfsad:
@sandeepraju 
We definitely will implement `RateLimiter` at some point, so it's worth keeping in mind while designing API for Ktor, so our implementation could be integrated with import replacement.

Note that RateLimiter is tricky to implement. E.g. current implementation maintains "discrete" rate limit, allowing to have some intervals with a rate higher than declared, maintaining declared rate **in average**. It's worth to reflect that fact in KDoc

TheBestPessimist:
When talking about a rate limiter i would really like something like what nginx offers: the base rate-limit value, but an extra `burst` value.

Reference: http://nginx.org/en/docs/http/ngx_http_limit_req_module.html#limit_req

PR #480, opened 2018-07-16
Provide metrics with Micrometer.io -- Initial commit for discussion NOT TO MERGE!

dave08:
The changes don't keep the old metrics names due to the difference of how Micrometer works.

Next steps:
- How should different deps for each framework be handled?
- Should default histogram buckets be specified, or should the user be able to configure them according to their own case?
- Should any more types of metrics be introduced? As in https://vertx.io/docs/vertx-micrometer-metrics/java/ Http server section or https://github.com/nolequen/vertx-prometheus-metrics?


hpoul:
Hi,
just wondering about the status of this? is micrometer the future for ktor, or just an experiment? 

thanks,
  herbert

dave08:
@hpoul You could just copy the code and use it in your project. I started this PR since I understood there was some interest in expanding metrics support. I'm just waiting for review if this is what they expected, and then I was ready to add things like DSL for common configs, and a builder function at least for Prometheus (others could be added later...), feedback on extra metrics needed would be nice! I already have this in Production and in it's basic current form it's already very helpful!

dave08:
@hpoul What is your use case? Will you be using some kind of Orchestration (Docker Swarm, Kubernetes, Openshift)? The truth is that whether Dropwizard will be replaced or not, your code shouldn't change too much unless you're including custom metrics in your code. And the metric names/types will be different...

dave08:
@hpoul Then I really think you should try copying this PR over (even if it doesn't get merged, you could always use it...), since Prometheus is the most popular monitoring system for Docker, Docker Swarm and Kubernetes and Dropwizard doesn't support it. Good luck! You can even keep a reference of Micrometer before you pass it to the `Feature`. If you have any suggestions for improvements, I'd like to hear!

orangy:
I'm not sure why do you want to replace metrics support? They could be two different features in two different modules so that user can choose which metrics library to use.

orangy:
I would go with 
> install(DropwizardMetrics) or install(MicrometerMetrics)

for a start. I don't think a lot of people would want to install both at the same time. They should be in different modules, since they have different dependencies, obviously.

dave08:
@orangy How could I handle the dependencies? In Micrometer, they have one for each metrics system. Like `compile 'io.micrometer:micrometer-registry-jmx:latest.release'` or `compile 'io.micrometer:micrometer-registry-prometheus:latest.release'`. Also, now there's the Dropwizard deps.

dave08:
Ok, that seems solved. What about `AuthenticationRouteSelector` check causing a dependency on auth module? Should I add the dependency, or is there a better way of ignoring those kind of route parts?

dave08:
So there isn't any advantage to providing both @jkschneider? Micrometer provides all the features of Dropwizard, and is even potentially backwards-compatible if the user is already using Dropwizard api in their own code?

michael-barker:
Any updates on this?

dave08:
I could, we are using this internally, but I'm just not sure how to split this into two modules in gradle...

jkschneider:
@dave08 Can you describe how you are trying to divide the code into two modules in Gradle? I could probably help you with that.

orangy:
@dave08 looks like something is wrong with the PR, lots of unrelated commits…

dave08:
Replaced by https://github.com/ktorio/ktor/pull/718

Issue #485, opened 2018-07-18
Test engine mangles PartData.FileItem contents


marshallpierce:
Demo project: https://bitbucket.org/marshallpierce/ktor-multipart-upload-test-repro/src/master/

It seems like maybe bytes with the high bit set confuse things.

With a real service file uploads seem to be working ok (I tested uploading with a file with random bytes and I was able to download the right bytes again).

marshallpierce:
Any idea when the release incorporating #486 will happen?

marshallpierce:
I took a minute to try, but had various API compatibility related problems. Are the API changes for 0.9.4 going to change further? If so I'll make the necessary changes, but I'd rather not do it twice for 0.9.4. :)

marshallpierce:
Looks good in 0.9.4, thanks.

PR #486, opened 2018-07-18
Fixes binary data in multipart tests `TestApplicationRequest.setBody`

soywiz:
The problem was that File input streams read as `ISO-8859-1`, but later the whole content was encoded as UTF-8.
This PR tries to handle it as a binary stream so no decoding+encoding is required, and calculates it with a writer and pipes the content, so if the input comes from a large File for testing, no need to read the whole content in memory.


soywiz:
Fixes #485

e5l:
Hi, @soywiz. Thanks for the request.
Could you make the test separate?

soywiz:
Done

PR #488, opened 2018-07-19
resovle config file in command line env

### LINKED ISSUE #374: When config is passed via -config=<filename>, environment variables are not resolved
I've discovered that the config parser doesn't resolve environment variables if the config is passed via the `-config=` command line option instead of using the default `application.conf`.

To test:
- open the `ktor-samples-hello` project
- create a test config where `port = ${PORT}`, see [test.conf.txt](https://github.com/ktorio/ktor/files/1863449/test.conf.txt)
- build as an application, e.g. `gradle distTar`
- run `PORT=8081 bin/ktor-samples-hello -config=../../../test.conf`
You should get the error: `Exception in thread "main" com.typesafe.config.ConfigException$NotResolved: need to Config#resolve(), see the API docs for Config#resolve(); substitution not resolved: ConfigReference(${PORT})`

I would expect environment variables to still be resolved as with the default `application.conf`

trevorwhitney:
fixes #374

It was hard to test that the resolution actually pulls in env vars. Without the `.resolve()` the test blows up, so I still think it's a worthwhile test, just not the most thorough. Open to suggestions.

### DIFF
--- a/ktor-server/ktor-server-host-common/src/io/ktor/server/engine/CommandLine.kt
+++ b/ktor-server/ktor-server-host-common/src/io/ktor/server/engine/CommandLine.kt
@@ -24,7 +24,7 @@ fun commandLineEnvironment(args: Array<String>): ApplicationEngineEnvironment {
     val commandLineMap = argsMap.filterKeys { it.startsWith("-P:") }.mapKeys { it.key.removePrefix("-P:") }
 
     val environmentConfig = ConfigFactory.systemProperties().withOnlyPath("ktor")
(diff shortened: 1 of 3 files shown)


cy6erGn0m:
Shouldn't it be resolved later? At least after combining it with fallbacks?

trevorwhitney:
@cy6erGn0m I was assuming that env vars would only be provided via a file config, but I that's a good suggestion to delay resolution until after the fallbacks are merged. I've updated the PR. Thanks!

cy6erGn0m:
thanks!

Issue #490, opened 2018-07-20
Error with static content resolution when running from IDE


FredDeschenes:
Test case here : https://github.com/FredDeschenes/ktor-static-content-resolution-bugs

This is the issue described as "First issue, running from an IDE" in the repo.

Basically no check is done to validate that a file is being accessed instead of a directory, so `RandomAccessFile.open` explodes.

PR incoming.

PR #491, opened 2018-07-20
Fix for #490

FredDeschenes:
Added a check in `ApplicationCall.resourceResource` to validate that we
are accessing a file and not a directory before attempting to open it.

Not sure if this can be unit-tested, if you have any idea how to achieve this I'll gladly add the test case(s).

### DIFF
--- a/ktor-server/ktor-server-core/src/io/ktor/content/StaticContentResolution.kt
+++ b/ktor-server/ktor-server-core/src/io/ktor/content/StaticContentResolution.kt
@@ -27,7 +27,7 @@ fun ApplicationCall.resolveResource(path: String,
         when (url.protocol) {
             "file" -> {
                 val file = File(decodeURLPart(url.path))
-                return if (file.exists()) LocalFileContent(file, mimeResolve(file.extension)) else null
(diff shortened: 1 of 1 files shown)


ktor-server/ktor-server-core/src/io/ktor/content/StaticContentResolution.kt
```diff
@@ -27,7 +27,7 @@ fun ApplicationCall.resolveResource(path: String,
         when (url.protocol) {
             "file" -> {
                 val file = File(decodeURLPart(url.path))
-                return if (file.exists()) LocalFileContent(file, mimeResolve(file.extension)) else null
+                return if (file.exists() && file.isFile) LocalFileContent(file, mimeResolve(file.extension)) else null
```
cy6erGn0m: we can replace exists check with `isFile` check as it will return false for non-existing path as well. It will avoid double native function invocation inside of `java.io.File`

FredDeschenes: Good catch, I'll remove the `.exists()` call.

PR #493, opened 2018-07-20
Fix for #492

### LINKED ISSUE #492: Error with static content resolution when running from a JAR
Test case here : https://github.com/FredDeschenes/ktor-static-content-resolution-bugs

This is the issue described as "Second issue, running from a JAR" in the repo.

Basically no range check is done to validate that the path we are attempting to load actually has an extension before calling `substring` to extract the extension.

PR incoming.

FredDeschenes:
Fixes #492 

Added range check before calling `substring` to extract the extension
of a file path.

As with #491, not sure if/how this can be unit-tested, if you have any idea how to achieve this I'll gladly add the test case(s).

### DIFF
--- a/ktor-server/ktor-server-core/src/io/ktor/content/StaticContentResolution.kt
+++ b/ktor-server/ktor-server-core/src/io/ktor/content/StaticContentResolution.kt
@@ -30,15 +30,15 @@ fun ApplicationCall.resolveResource(path: String,
                 return if (file.exists()) LocalFileContent(file, mimeResolve(file.extension)) else null
             }
             "jar" -> {
-                val zipFile = findContainingJarFile(url.toString())
-                return url.path.extension()?.let { extension ->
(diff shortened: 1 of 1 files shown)


cy6erGn0m:
Why should we fail to resolve if there is no extension?

FredDeschenes:
So should I return the file anyway but with the `ContentType.Application.OctetStream` content type (which appears to be the default in the `ContentType` extension functions)? This would break the `mimeResolve` lambda though as it works on file extensions.

cy6erGn0m:
I'd say it's better to pass an empty string to `mimeResolve` (it should fallback to octet stream by default). Also one can provide a custom `mimeResolve` function that could handle this case particularly

FredDeschenes:
Mmm this is giving me weird results. Attempting to GET "/" now downloads an empty file since there's no way to check if the content we are trying to load is a file without an extension or a directory since they's inside the JAR (can't use the `.isFile` like for #491).

PR #498, opened 2018-07-25
Added initial webjars feature

viniciusccarvalho:
Adds a new feature to allow one to consume static assets from webjars.

```kotlin
install(Webjars) {
path = '/assets' //defaults to /webjars
}
```

So, if you have `org.webjars:jquery:3.3.1` jar on your class path, either of the following scripts would resolve to the proper file

`<script src="/webjars/jquery/jquery.js">` -> Version agnostic
`<script src="/webjars/jquery/3.3.1/jquery.js">` -> Using version

Not sure how to contribute to the documentation if this PR is accepted though


ktor-features/ktor-webjars/src/io/ktor/webjars/Webjars.kt
```diff
@@ -0,0 +1,85 @@
...
+
+    private fun extractWebJar(path : String) : String {
+        val firstSlash = if (path.startsWith("/")) 1 else 0
+        val nextSlash = path.indexOf("/", 1)
+        val webjar = if(nextSlash > -1) path.substring(firstSlash,  nextSlash) else ""
+        val partialPath = path.substring(nextSlash + 1)
+        return locator.getFullPath(webjar, partialPath)
+    }
+
+    private val locator = WebJarAssetLocator()
```
cy6erGn0m: Is is stateless? Is is thread-safe? Is it safe to call `getFullPath` concurrently?

ktor-features/ktor-webjars/src/io/ktor/webjars/Webjars.kt
```diff
@@ -0,0 +1,85 @@
...
+
+    class Configuration {
+        var path = "/webjars"
+        set(value) {
+            var buffer = StringBuilder(value)
```
cy6erGn0m: Not sure about it. Most likely it's better to normalize it later. It looks strange if a property modifies values that has been written.

ktor-features/ktor-webjars/src/io/ktor/webjars/Webjars.kt
```diff
@@ -0,0 +1,85 @@
...
+            if(value.endsWith("/")){
+                buffer.deleteCharAt(buffer.length - 1)
+            }
+            field = buffer.toString()
+        }
+    }
+
+    private suspend fun intercept(context: PipelineContext<Unit, ApplicationCall>){
+        val fullPath = context.call.request.uri
+        if(fullPath.startsWith(configuration.path) && context.call.request.httpMethod == HttpMethod.Get){
```
cy6erGn0m: What if one makes call to `/webjars-something` ? This need to be checked more precisely. Also need to be checked in tests that we don't handle irrelevant requests.

ktor-features/ktor-webjars/src/io/ktor/webjars/Webjars.kt
```diff
@@ -0,0 +1,85 @@
...
+            val fileName = fileName(context.call.request.uri)
+            val partialPath = fullPath.replace(configuration.path, "")
```
cy6erGn0m: We need to remove only prefix instead of all occurrences. I'd prefer 'removePrefix` here instead as it works faster and does exactly what we need.

ktor-features/ktor-webjars/src/io/ktor/webjars/Webjars.kt
```diff
@@ -0,0 +1,85 @@
...
+            try {
+                val location = extractWebJar(partialPath)
+                context.call.respondBytes(ContentType.defaultForFilePath(fileName), HttpStatusCode.OK) {
+                    Webjars::class.java.classLoader.getResourceAsStream(location).readBytes()
+                }
+            }catch (multipleFiles: MultipleMatchesException){
+                context.call.respond(HttpStatusCode.BadRequest)
```
cy6erGn0m: 400 Bad request means that the HTTP request has illegal syntax or broken. For sure it should be 500 Internal Server Error as it is caused by wrong server packaging.

ktor-features/ktor-webjars/src/io/ktor/webjars/Webjars.kt
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    private suspend fun intercept(context: PipelineContext<Unit, ApplicationCall>){
+        val fullPath = context.call.request.uri
+        if(fullPath.startsWith(configuration.path) && context.call.request.httpMethod == HttpMethod.Get){
+            val fileName = fileName(context.call.request.uri)
+            val partialPath = fullPath.replace(configuration.path, "")
+            try {
+                val location = extractWebJar(partialPath)
+                context.call.respondBytes(ContentType.defaultForFilePath(fileName), HttpStatusCode.OK) {
```
cy6erGn0m: This will not provide versions tag so Condition headers support will not work that is VERY important for static files. You can try `JarFileContent` instead or provide etag based on webjar's version/checksum.

ktor-features/ktor-webjars/src/io/ktor/webjars/Webjars.kt
```diff
@@ -0,0 +1,85 @@
...
+                    Webjars::class.java.classLoader.getResourceAsStream(location).readBytes()
```
cy6erGn0m: NullPointerException?

cy6erGn0m: Shouldn't we try to stream it instead otherwise server will go down under load with OOM because of too many byte arrays

ktor-features/ktor-webjars/src/io/ktor/webjars/Webjars.kt
```diff
@@ -0,0 +1,85 @@
...
+                }
+            }catch (multipleFiles: MultipleMatchesException){
+                context.call.respond(HttpStatusCode.BadRequest)
+            }
+            catch (notFound: IllegalArgumentException){
+                context.call.respond(HttpStatusCode.NotFound)
```
cy6erGn0m: It's fine to respond 404 here however we actually can simply by-pass so anybody else could try to handle the request. Also consider usecase when `configuration.path` = `"/"` so any other paths and routing should work that is not the case while we are always reply 404 here

ktor-features/ktor-webjars/src/io/ktor/webjars/Webjars.kt
```diff
@@ -0,0 +1,85 @@
...
+            val partialPath = fullPath.replace(configuration.path, "")
+            try {
+                val location = extractWebJar(partialPath)
+                context.call.respondBytes(ContentType.defaultForFilePath(fileName), HttpStatusCode.OK) {
+                    Webjars::class.java.classLoader.getResourceAsStream(location).readBytes()
+                }
+            }catch (multipleFiles: MultipleMatchesException){
+                context.call.respond(HttpStatusCode.BadRequest)
+            }
+            catch (notFound: IllegalArgumentException){
```
cy6erGn0m: Can't we check exactly this case? At least we should't catch IAE from call.respond as for sure it will be some irrelevant IAE

ktor-features/ktor-webjars/src/io/ktor/webjars/Webjars.kt
```diff
@@ -0,0 +1,85 @@
...
+
+class Webjars(val configuration: Configuration) {
+
+    private fun fileName(path: String) : String = Paths.get(path).fileName.toString()
+
+    private fun extractWebJar(path : String) : String {
+        val firstSlash = if (path.startsWith("/")) 1 else 0
+        val nextSlash = path.indexOf("/", 1)
+        val webjar = if(nextSlash > -1) path.substring(firstSlash,  nextSlash) else ""
+        val partialPath = path.substring(nextSlash + 1)
```
cy6erGn0m: `"some-path"` -> `webjar` = `""` and `partialPath` = `"some-path"` Looks strange.. Shouldn't we simply return immediately (for example, `null` and then by-pass it or respond 404)

ktor-features/ktor-webjars/src/io/ktor/webjars/Webjars.kt
```diff
@@ -0,0 +1,85 @@
...
+
+    class Configuration {
+        var path = "/webjars"
+        set(value) {
+            var buffer = StringBuilder(value)
+            if(!value.startsWith("/")){
+                buffer.insert(0, "/")
+            }
+            if(value.endsWith("/")){
+                buffer.deleteCharAt(buffer.length - 1)
```
cy6erGn0m: What if `value` = `"/"` ?

viniciusccarvalho:
I don't know why so many .idea files were submitted, they were supposed to be ignored by .gitignore :(

ktor-features/ktor-webjars/src/io/ktor/webjars/Webjars.kt
viniciusccarvalho: From AFAIK yes, it's thread safe, and can be called concurrently. Spring would be on pretty bad grounds if this were not thread-safe ;)

ktor-features/ktor-webjars/src/io/ktor/webjars/Webjars.kt
viniciusccarvalho: It can be changed for sure, but I usually leverage setters for this purpose, to make sure that any incoming change conforms to the expected value of the parent class. Not unusual to see on many places this pattern. But if this is not something kotlin or ktor likes (to me is a coding style more than anything) it can indeed be done somewhere else, perhaps on a init{} block to normalize the path

ktor-features/ktor-webjars/src/io/ktor/webjars/Webjars.kt
viniciusccarvalho: Ideally I would like to use routing to handle this, but I did not know how to leverage routing from the pipeline.intercept chain. I followed the similar pattern used by the shutdown feature, but I see your point, will check for a path part now /\w/* first occurrence and see if it's an exact match.

ktor-features/ktor-webjars/src/io/ktor/webjars/Webjars.kt
viniciusccarvalho: The issue is that the webjars locator will throw IAE when it finds multiple files, and I believe those are different errors

ktor-features/ktor-webjars/src/io/ktor/webjars/Webjars.kt
viniciusccarvalho: It was my understanding that the fat jar would explode all jars inside the fat jar. This is why I did not use a `JarFileContent` as it would not find the jar inside the fat jar. Using `getResourceAsStream` would at least give me a inputStream to the contents. I need to figure out a way to associate a version now

ktor-features/ktor-webjars/src/io/ktor/webjars/Webjars.kt
viniciusccarvalho: Do I need to create my own `WriteChannelContent` subclass for that? Not sure how to proceed here

cy6erGn0m:
```
    @Test
    fun rootPath2() {
        withTestApplication {
            application.install(Webjars){
                path = "/"
            }
            application.routing {
                get("/") { call.respondText("Hello, World") }
            }
            handleRequest(HttpMethod.Get, "/jquery/jquery.js").let { call ->
                assertEquals(HttpStatusCode.OK, call.response.status())
                assertEquals("application/javascript", call.response.headers["Content-Type"])
            }
            handleRequest(HttpMethod.Get, "/").let { call ->
                assertEquals(HttpStatusCode.OK, call.response.status())
                assertEquals("Hello, World", call.response.content)
            }
        }
    }
```

ktor-features/ktor-webjars/src/io/ktor/webjars/Webjars.kt
viniciusccarvalho: Thanks I'm really bad in naming things. So `rootPath` is the configuration.path you set (for instance `/webjars/`), `resourcePath` is whatever is left after that (`/jquery/js/jquery.js`), the `webjar` is the first part of that path, could be `jquery`, `bootstrap` ... and finally the `partialPath` is `/js/jquery.js` I hope this makes thing clearer

viniciusccarvalho:
@cy6erGn0m Question for you: Trying to adapt to a streaming model here, which I think will come down to something like  overly simplified snippet
```kotlin
class InputStreamContent(val input: InputStream) :  OutgoingContent.ReadChannelContent() {
    override fun readFrom(): ByteReadChannel {
        return input.toByteReadChannel()
    }

}
```
And the use it via `context.call.respond(InputStreamContent(Webjars::class.java.classLoader.getResourceAsStream(location)))`

Is this what you meant by your comment on streaming, and if so, wouldn't a `InputStreamContent` class make sense to be reusable?

Regards

cy6erGn0m:
Yes, this is what I talked about. However we wouldn't like to have `InputStreamContent` as it is blocking API

cy6erGn0m:
@viniciusccarvalho Is it ready to be reviewed again?

viniciusccarvalho:
@cy6erGn0m I believe so, I've addressed all of the comments on either code or feedback.

The Last-Modified is the only piece dangling, I'm assuming that for now, we set it to the time the server started, given that replacing a webjar will probably mean a server restart. Not dealing with etag yet, but can be done via a checksum of the bytes from the inputstream, if Last-Modified is not enough.

Thanks for the review

cy6erGn0m:
Ok, looks better. So the only thing need to be done: formatting. Too many spaces need to be fixed.

cy6erGn0m:
Thanks!

PR #504, opened 2018-08-01
Added RedisSessionStorage

soywiz:
Let's wait until Redis client is published. But we can start commenting and changing things already.


ktor-features/ktor-server-sessions-redis/src/io/ktor/sessions/RedisSessionStorage.kt
```diff
@@ -0,0 +1,33 @@
...
+        }
+    }
+
+    override suspend fun write(id: String, data: ByteArray?) {
+        val key = buildKey(id)
+        if (data == null) {
+            redis.del(buildKey(id))
+        } else {
+            redis.set(key, hex(data))
+            redis.expire(key, ttlSeconds)
```
sh-cho, line 24:
```suggestion
            redis.setex(key, ttlSeconds, hex(data))
```

:)

PR #506, opened 2018-08-02
Handling jwt exceptions for missing keyId and failure to decode token

Zddie:
Found and fixed a little bug when kid not found in jwks and when invalid bearer token used on verification of jwt (throws and causes 500 instead of 401)


Zddie:
@cy6erGn0m changes made.

cy6erGn0m:
Thanks!

Zddie:
No worries, glad I could help.

PR #509, opened 2018-08-03
Add HttpStatusCode constants for the status codes defined in RFC 2518 (WebDAV)

mdr:
This adds 5 additional `HttpStatusCode` constants specified in [WebDAV](https://tools.ietf.org/html/rfc2518#section-10) (see also the [HTTP status registry](https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml)). One, "102 Processing", was already present.

Some of these status codes are in common use in REST services, such as "422 Unprocessable Entity" (e.g. https://www.yiiframework.com/doc/guide/2.0/en/rest-error-handling).


cy6erGn0m:
Thanks!

PR #524, opened 2018-08-16
Makes SessionTransportTransformerEncrypt return null on failure

soywiz:


cy6erGn0m:
Would be nice to add logger.debug to be able to see exact error in the log

soywiz:
Done

PR #547, opened 2018-08-28
Rename `XForwardedHeaderSupport` to `XForwardedHeadersSupport`

soywiz:
To be consistent with `ForwardedHeaderSupport`.

The alternative is to rename `ForwardedHeaderSupport` to `ForwardedHeadersSupport`.

Documentation was wrong because of copy-pasting `ForwardedHeaderSupport` adding an `X`: https://github.com/ktorio/ktorio.github.io/commit/3985523155489d17c8ac623595629e2e1ef5c653

But would be nice if they are consistent.


cy6erGn0m:
Merged with the only change 119841c66cc26596d497d4785b8575705c6f9352

PR #554, opened 2018-08-31
[ktor-client] feature for set default url parameters for all requests

fegol:
In many cases we need set host, basePath etc once for all requests. This is an example of how it can be make. Feature is raw and just demonstrates what i want see in ktor client. If it can be make in another way then just show me it :)


e5l:
Could you try to use the `defaultRequest {}`?

fegol:
Thank you, this is what I need! But I would like to be able to specify the base path for requests. It's would be nice feature.

e5l:
https://github.com/ktorio/ktor/issues/555

PR #559, opened 2018-09-03
ktor-client-core: Copy existing config in HttpClient#config

kenzierocks:
Copy the existing config in HttpClient#config, as the docs specify.
Prior to this all features and interceptors were lost.


e5l:
The test class locates in `src` directory and can't be recognized as test.
Typically we use such classes as the `test template`.

kenzierocks:
Added engine test runners.

e5l:
LGTM.

Awesome, thanks for the request.

PR #565, opened 2018-09-06
Introduce CallId feature

cy6erGn0m:
* Introduced `CallId` feature
* Introduced `CallLogging` MDC support (via [ThreadContextElement](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.experimental/-thread-context-element/index.html))
* Two more call phases added: `CallPreparation` and `CallLifecycle`
    * `CallPreparation` is intended for call initial preparation, attributes setup and so on
    * `CallLifecycle` is purposed for logging, metrics, error handling and other call-tracking features that usually does `try-finally` around the whole call 
* `Metrics`, `StatusPages` and `CallLogging` features moved to new phases
* Added tests to ensure that moving `StatusPages` feature doesn't break `AutoHeadResponse` feature in spite of that they are installed into different phases

Open questions:
* `CallLifecycle` [naming]
* `CallId`: `fetch` vs `extract` [naming]
* `callIdMdc` extension function naming
    * `callIdMdc`
    * `mdcCallId`
    * `mdcForCallId`, `mdcFromCallId`


cy6erGn0m:
Merged.

PR #572, opened 2018-09-08
ktor-http: URLBuilder isolate port and the protocol's defaultPort

### LINKED ISSUE #561: ktor-http: [URLBuilder] Specifying a protocol with a default port value overwrites an already specified port
## Overview

| Library | Version |
|-|-|
| ktor-http | `0.9.4` |

Given the following execution,
```kotlin
val url = URLBuilder().apply {
    takeFrom(getenv("MYSQL_URL")) // import from "mysql://user:password@localhost:34567/myDb"
    protocol = URLProtocol("mysql", 3306)
}.build()
```

## Expected Outcome

I expected the default port to be used on the protocol only if a port was not explicitly specified

```kotlin
println(url) // "mysql://user:password@localhost:34567/myDb"
```

## Actual Outcome

The port is always overwritten when a protocol is set

```kotlin
println(url) // "mysql://user:password@localhost/myDb" (implying port `3306`)
```

https://github.com/ktorio/ktor/blob/a65667a6f0ca98128c38025f989bad7cb6da4576/ktor-http/src/io/ktor/http/URLBuilder.kt#L16-L18

## Suggested Resolution

https://github.com/ktorio/ktor/blob/a65667a6f0ca98128c38025f989bad7cb6da4576/ktor-http/src/io/ktor/http/URLProtocol.kt#L19

The default port on a URLProtocol that has not already been mapped is `-1`, represented by the following code:
```kotlin
val url = URLBuilder().apply {
    takeFrom(getenv("MYSQL_URL")) // import from "mysql://user:password@hostUsingDefaultPort/myDb"
}.build()
```

the following is built

```kotlin
println(url.port) // "-1"
```

Since this is the case, I propose that the `port` is only overwritten when the protocol would overwrite a port of `-1`

```kotlin
var protocol: URLProtocol by observable(protocol) { _, _, value -> 
    if (port == -1) {
        port = value.defaultPort 
    }
} 
```

## Current Workaround

Currently, my solution is to build the `Url` beforehand and check if a port was specified after applying the protocol
```kotlin
val url = URLBuilder().apply {
    val envUrl = Url(getenv("MYSQL_URL")) // import from "mysql://user:password@localhost:34567/myDb"
    takeFrom(envUrl)
    protocol = URLProtocol("mysql", 3306)
    if (envUrl.port > 0) {
        port = envUrl.port
    }
}.build()
```

bradynpoulsen:
See #561 for the original issue discussion

This modification removes the behavior on the [URLBuilder] to no longer set the `port` when explicitly specifying a `protocol`.

#### Before

```kotlin
val url = URLBuilder().apply {
    port = 8080
    protocol = URLProtocol.HTTPS
}.build()
println(url.port) // 443
```

#### After

```kotlin
val url = URLBuilder().apply {
    port = 8080
    protocol = URLProtocol.HTTPS
}.build()
println(url.port) // 8080
```

#### Before

```kotlin
val url = URLBuilder().apply {
    protocol = URLProtocol.HTTPS
    println(port) // 443, changed by `protocol =`
}.build()
println(url.port) // 443
```

#### After

```kotlin
val url = URLBuilder().apply {
    protocol = URLProtocol.HTTPS
    println(port) // -1, no longer changed by `protocol =`
}.build()
println(url.port) // 443
```

#### Before

```kotlin
val url = URLBuilder().apply {
    takeFrom("http://localhost/path")
    protocol = URLProtocol("custom", 8080)

    println(port) // 8080, changed by `protocol =`
}.build()
println(url.protocol) // URLProtocol(name=custom, defaultPort=8080)
println(url.port) // 8080
```

#### After

```kotlin
val url = URLBuilder().apply {
    takeFrom("http://localhost/path")
    protocol = URLProtocol("custom", 8080)

    println(port) // -1, no port was specified in URL imported by `takeFrom` and not set by `protocol =`
}.build()
println(url.protocol) // URLProtocol(name=custom, defaultPort=8080)
println(url.port) // 8080
```

### DIFF
--- a/ktor-http/src/io/ktor/http/URLBuilder.kt
+++ b/ktor-http/src/io/ktor/http/URLBuilder.kt
@@ -1,11 +1,11 @@
 package io.ktor.http
 
-import kotlin.properties.Delegates.observable
+const val UNKNOWN_PORT = 0
 
 class URLBuilder(
-    protocol: URLProtocol = URLProtocol.HTTP,
+    var protocol: URLProtocol = URLProtocol.HTTP,
     var host: String = "localhost",
-    var port: Int = protocol.defaultPort,
+    var port: Int = UNKNOWN_PORT,
     var user: String? = null,
     var password: String? = null,
     var encodedPath: String = "/",
@@ -13,9 +13,7 @@ class URLBuilder(
     var fragment: String = "",
     var trailingQuery: Boolean = false
 ) {
-    var protocol: URLProtocol by observable(protocol) { _, _, value ->
-        port = value.defaultPort
-    }
+    private inline val portOrProtocolPort get() = port.takeUnless { it == UNKNOWN_PORT } ?: protocol.defaultPort
 
     fun path(vararg components: String) {
         path(components.asList())
@@ -38,7 +36,7 @@ class URLBuilder(
         }
         out.append(host)
 
-        if (port != protocol.defaultPort) {
+        if (port != UNKNOWN_PORT && port != protocol.defaultPort) {
             out.append(":")
             out.append(port.toString())
         }
@@ -57,7 +55,7 @@ class URLBuilder(
     fun buildString(): String = appendTo(StringBuilder(256)).toString()
 
     fun build(): Url = Url(
-        protocol, host, port, encodedPath, parameters.build(), fragment, user, password, trailingQuery
+        protocol, host, portOrProtocolPort, encodedPath, parameters.build(), fragment, user, password, trailingQuery
     )
 
     // Required to write external extension function
@@ -77,6 +75,10 @@ data class Url(
     val password: String?,
     val trailingQuery: Boolean
 ) {
+    init {
+        require(port in 1..65536 || port == UNKNOWN_PORT) { "port must be between 1 and 65536, or $UNKNOWN_PORT if not set" }
+    }
+
     override fun toString(): String = buildString {
         append(protocol.name)
         append("://")
@@ -88,7 +90,7 @@ data class Url(
             }
             append('@')
         }
-        if (port == protocol.defaultPort) {
+        if (port == UNKNOWN_PORT || port == protocol.defaultPort) {
             append(host)
         } else {
             append(hostWithPort)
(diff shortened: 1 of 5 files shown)


ktor-http/src/io/ktor/http/URLBuilder.kt
```diff
@@ -86,6 +84,10 @@ data class Url(
     val password: String?,
     val trailingQuery: Boolean
 ) {
+    init {
+        require(port >= 0 || port == UNKNOWN_PORT) { "port must be >= 0 or $UNKNOWN_PORT if not set" }
```
bradynpoulsen: Should `0` be allowed as a port? Perhaps this should actually validate a legitimate range
```kotlin
require(port in 1..65535 || port == UNKNOWN_PORT)
```

e5l:
Hi, @bradynpoulsen. Thanks for the PR. 
I'll take a look in a few days.

ktor-http/src/io/ktor/http/URLBuilder.kt
```diff
@@ -1,21 +1,19 @@
 package io.ktor.http
 
-import kotlin.properties.Delegates.observable
+const val UNKNOWN_PORT = -1
```
e5l: Let use 0 value as default.

ktor-http/src/io/ktor/http/URLBuilder.kt
```diff
@@ -38,9 +36,9 @@ class URLBuilder(
         }
         out.append(host)
 
-        if (port != protocol.defaultPort) {
+        if (portOrDefault != protocol.defaultPort) {
             out.append(":")
-            out.append(port.toString())
+            out.append(portOrDefault.toString())
```
e5l: We could try switch to the different logic here:
   - if the port is default: don't print anything.
   - if the user specifies port by hand: print it(even it equals the default value)

ktor-http/src/io/ktor/http/URLBuilder.kt
```diff
@@ -38,7 +36,7 @@ class URLBuilder(
         }
         out.append(host)
 
-        if (port != protocol.defaultPort) {
```
bradynpoulsen: To remain as backwards compatible with previous behavior as possible, I left the check to only add if the port is not the protocol's default port

bradynpoulsen:
@e5l Let me know what you think of these modifications. I'll rebase to import the new UrlTest.kt changes if these modifications look good or need more work

e5l:
Could you rebase branch on master to run integration tests?

PR #573, opened 2018-09-08
Adds API documentation to the auth Headers

soywiz:


ktor-features/ktor-auth/src/io/ktor/auth/Headers.kt
```diff
@@ -78,8 +147,16 @@ sealed class HttpAuthHeader(val authScheme: String) {
             }
         }
 
+        /**
+         * Copies this [Parameterized] appending a new parameter [name] [value].
+         */
         fun withParameter(name: String, value: String) = Parameterized(authScheme, this.parameters + HeaderValueParam(name, value))
```
soywiz: I guess this should preserve the `encoding`. I can add another commit fixing this + test if confirmed.

cy6erGn0m: Yes, you are right, encoding is lost here

ktor-features/ktor-auth/test/io/ktor/tests/auth/AuthHeadersTest.kt
```diff
@@ -5,6 +5,15 @@ import org.junit.Test
 import kotlin.test.*
 
 class AuthHeadersTest {
+    @Test
+    fun testWithParameterParameterBehaviour() {
+        val header = HttpAuthHeader.Parameterized("authScheme", mapOf("a" to "a"))
+        // It uses singleOrNull, and since withParameter appends, it would be twice and would return null.
+        // Verify if this is the expected behaviour.
+        assertEquals(null, header.withParameter("a", "A").parameter("a"))
```
soywiz: This doesn't look like the expected behaviour. I can make it replace the old parameter, or that the `parameter` method returns the last matching parameter or a list instead.

cy6erGn0m: That's serious question. Returning first or last looks more natural to end-users. As far as I remember we have `header()` and `headers[name]` returning `firstOrNull`. Same for request parameters.
Unfortunately `withParameter` name doesn't describe what exactly it does (append or replace)

cy6erGn0m: After all I think we better to not change it's behaviour. Currently it is always appending new parameter and it was always fine for everyone. So let's keep it appending rather than replaicing.

soywiz: Ok. Then what I can do is to continue appending, but change the `singleOrNull` with `lastOrNull`?

cy6erGn0m: Well, we have it `firstOrNull` everywhere and this is what the most apis do in such cases

PR #575, opened 2018-09-11
improve kotlinxSerializer

NikkyAI:
this allows for serializing and deserializing of any types and classes without generic arguments

with expect / actual this reuses `serializerByTypeToken` on the jvm and uses a naive implementation on IOS because (afaik) there is no reified type available there, i cannot test this

it would be futureproofing to expose a TypeInfo to the write frunction (and the HttpRequestPipeline in general i guess)

### DIFF
--- a/ktor-client/ktor-client-features/ktor-client-json/ktor-client-json-ios/src/io/ktor/client/features/json/serializer/SerialHelperIos.kt
+++ b/ktor-client/ktor-client-features/ktor-client-json/ktor-client-json-ios/src/io/ktor/client/features/json/serializer/SerialHelperIos.kt
@@ -1,0 +1,11 @@
+package io.ktor.client.features.json.serializer
+
+import io.ktor.client.call.TypeInfo
+import kotlinx.serialization.KSerializer
+import kotlinx.serialization.list
+import kotlinx.serialization.serializer
+import kotlinx.serialization.set
+import java.lang.reflect.ParameterizedType
+import java.lang.reflect.WildcardType
+
+internal actual fun serializerForPlatform(type: TypeInfo): KSerializer<*> = type.type.serializer()
--- a/ktor-client/ktor-client-features/ktor-client-json/ktor-client-json-ios/src/io/ktor/client/features/json/serializer/SerialHelperIos.kt
+++ b/ktor-client/ktor-client-features/ktor-client-json/ktor-client-json-ios/src/io/ktor/client/features/json/serializer/SerialHelperIos.kt
@@ -9,3 +9,5 @@ import java.lang.reflect.ParameterizedType
 import java.lang.reflect.WildcardType
 
 internal actual fun serializerForPlatform(type: TypeInfo): KSerializer<*> = type.type.serializer()
+
+internal actual fun serializerForObject(element: Any): KSerializer<Any>? = null
--- a/ktor-client/ktor-client-features/ktor-client-json/ktor-client-json-jvm/src/io/ktor/client/features/json/serializer/SerialHelperJvm.kt
+++ b/ktor-client/ktor-client-features/ktor-client-json/ktor-client-json-jvm/src/io/ktor/client/features/json/serializer/SerialHelperJvm.kt
@@ -1,0 +1,56 @@
+package io.ktor.client.features.json.serializer
+
+import io.ktor.client.call.TypeInfo
+import kotlinx.serialization.*
+import kotlinx.serialization.internal.*
+import java.lang.reflect.ParameterizedType
+import java.lang.reflect.Type
+import java.lang.reflect.WildcardType
+import kotlin.reflect.KClass
+
+internal actual fun serializerForPlatform(type: TypeInfo): KSerializer<*> = serializerByTypeToken(type.reifiedType)
+
+@Suppress("UNCHECKED_CAST")
+fun serializerByTypeToken(type: Type): KSerializer<Any> = when (type) {
+    is Class<*> -> if (!type.isArray) {
+        serializerByClass(type.kotlin)
+    } else {
+        val componentType: Class<*> = type.componentType
+        val componentSerializer = kotlinx.serialization.serializerByTypeToken(componentType)
+        ReferenceArraySerializer(componentType.kotlin as KClass<Any>, componentSerializer) as KSerializer<Any>
+    }
+    is ParameterizedType -> {
+        val rootClass = (type.rawType as Class<*>)
+        val args = (type.actualTypeArguments).map { argument ->
+            when (argument) {
+                is WildcardType -> {
+                    argument.upperBounds.first().let { firstUpperBound ->
+                        when (firstUpperBound) {
+                            is Class<*> -> firstUpperBound
+                            else -> null
+                        }
+                    }
+                }
+                else -> null
+            } ?: argument
+        }
+        when {
+            List::class.java.isAssignableFrom(rootClass) -> ArrayListSerializer(serializerByTypeToken(args[0])) as KSerializer<Any>
+            Set::class.java.isAssignableFrom(rootClass) -> HashSetSerializer(serializerByTypeToken(args[0])) as KSerializer<Any>
+            Map::class.java.isAssignableFrom(rootClass) -> HashMapSerializer(
+                serializerByTypeToken(args[0]), serializerByTypeToken(args[1])
+            ) as KSerializer<Any>
+            Map.Entry::class.java.isAssignableFrom(rootClass) -> MapEntrySerializer(
+                serializerByTypeToken(args[0]), serializerByTypeToken(args[1])
+            ) as KSerializer<Any>
+
+            else -> {
+                throw IllegalStateException("ParameterizedType '${type.rawType}' is not implemented")
+                // Cannot access 'invokeSerializerGetter': it is internal in 'kotlinx.serialization'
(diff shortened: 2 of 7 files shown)


NikkyAI:
one thing i am still unhappy with is that i need to compile against `ktor-client-json-jvm` because the defaultSerializer is actualized there even though i do not care for any of its dependencies and should not need it

NikkyAI:
maybe the kotlinx-coroutines plugin should be applied to be able to add tests for this

e5l:
Hi, @NikkyAI. Thanks for the PR, I'll take a look in a few days.
No problem with plugin :)  Could you apply it and add some tests?

NikkyAI:
i added tests for jvm, in the process i discovered bugs in kotlinx.serialization for jvm, i made a copy of the method that i fixed and a PR https://github.com/Kotlin/kotlinx.serialization/pull/212

ktor-client/ktor-client-features/ktor-client-json/ktor-client-json-jvm/test/io/ktor/client/features/json/test/JsonTest.kt
```diff
@@ -3,6 +3,7 @@ package io.ktor.client.features.json.test
 import io.ktor.application.*
 import io.ktor.client.engine.cio.*
 import io.ktor.client.features.json.*
+import io.ktor.client.features.json.serializer.KotlinxSerializer
```
e5l: Please consider using star(`*`) imports

ktor-client/ktor-client-features/ktor-client-json/ktor-client-json-ios/src/io/ktor/client/features/json/serializer/SerialHelperIos.kt
```diff
@@ -0,0 +1,11 @@
...
+
+import io.ktor.client.call.TypeInfo
+import kotlinx.serialization.KSerializer
+import kotlinx.serialization.list
+import kotlinx.serialization.serializer
+import kotlinx.serialization.set
+import java.lang.reflect.ParameterizedType
+import java.lang.reflect.WildcardType
+
+actual fun serializerForPlatform(type: TypeInfo): KSerializer<*> = type.type.serializer()
```
e5l: Could it be internal?

ktor-client/ktor-client-features/ktor-client-json/ktor-client-json-jvm/src/io/ktor/client/features/json/serializer/SerialHelperJvm.kt
```diff
@@ -0,0 +1,45 @@
...
+
+actual fun serializerForPlatform(type: TypeInfo): KSerializer<*> = serializerByTypeToken(type.reifiedType)
+
+@Suppress("UNCHECKED_CAST")
+fun serializerByTypeToken(type: Type): KSerializer<Any> = when (type) {
+    is Class<*> -> if (!type.isArray) {
+        serializerByClass(type.kotlin)
+    } else {
+        val eType: Class<*> = type.componentType
+        val s = kotlinx.serialization.serializerByTypeToken(eType)
```
e5l: Please consider using meaningful names for identifiers

ktor-client/ktor-client-features/ktor-client-json/ktor-client-json-jvm/src/io/ktor/client/features/json/serializer/SerialHelperJvm.kt
```diff
@@ -0,0 +1,45 @@
...
+    is ParameterizedType -> {
+        val rootClass = (type.rawType as Class<*>)
+        val args = type.actualTypeArguments.map { argument ->
+            when (argument) {
+                is WildcardType -> argument.upperBounds.first() as Class<*>
+                else -> argument
+            }
+        }
+        when {
+            List::class.java.isAssignableFrom(rootClass) -> ArrayListSerializer(serializerByTypeToken(args[0])) as KSerializer<Any>
```
e5l: Could we introduce the extension point here to add user mappers?

ktor-client/ktor-client-features/ktor-client-json/ktor-client-json-jvm/test/io/ktor/client/features/json/test/JsonTest.kt
```diff
@@ -13,16 +14,29 @@ import io.ktor.response.*
...
+import kotlinx.serialization.Serializable
+import java.util.*
 import kotlin.test.*
 
+@Serializable
 data class Widget(
-    val name: String,
-    val value: Int,
-    val tags: List<String> = emptyList()
+        val name: String,
```
e5l: Could you use the kotlin style guide(https://kotlinlang.org/docs/reference/coding-conventions.html)?
It could be configured as default style guide in IDEA `Code Style` section for kotlin.

ktor-client/ktor-client-features/ktor-client-json/src/io/ktor/client/features/json/serializer/KotlinxSerializer.kt
```diff
@@ -1,39 +1,42 @@
...
     }
 
     override fun write(data: Any): OutgoingContent {
         @Suppress("UNCHECKED_CAST")
-        val mapper = (writeMappers[data::class] as? (Any) -> JsonElement)
-            ?: error("There is no mapper for $data")
-
-        return TextContent(mapper(data).toString(), ContentType.Application.Json)
+        val clazz: KClass<Any> = data::class as KClass<Any>
+        val serializer = clazz.serializer()
```
e5l: Unfortunately, `KClass<*>.serializer()` is not supported in the native plugin for now.
You could consult about it there https://github.com/Kotlin/kotlinx.serialization

NikkyAI: well it seems to work, for now, only non non-parameterized types though .. for more some typeinfo would do the trick
i did not want to do such a invasive change though

ktor-client/ktor-client-features/ktor-client-json/ktor-client-json-jvm/src/io/ktor/client/features/json/serializer/SerialHelperJvm.kt
NikkyAI: as a user i would register custom serializers on the SerialContext when setting
```
serializer = kotlinxSerializer() { 
    registerSerializer(...)
}
```
or by passing it a existing `JSON`
mappers as they were used were quite limited since they only mapped based on class and generics were erased
is there any usecase where mappers would be prefered ?

NikkyAI: we could make a extension point to allow users to map custom serializers though
backed by a `HasMap<Class, (List<Type>) -> KSerializer<Any>>` maybe ?

ktor-client/ktor-client-features/ktor-client-json/ktor-client-json-jvm/test/io/ktor/client/features/json/test/JsonTest.kt
e5l: Thanks

ktor-client/ktor-client-features/ktor-client-json/ktor-client-json-jvm/src/io/ktor/client/features/json/serializer/SerialHelperJvm.kt
e5l: Something that looks like:
`registerSerializer(Type) { /* Serializer */ } `

And we could store serializers in `List` instead of the hashmap to register multiple serializers and find first `not failing`.

e5l: Sure. Let's wait for the PR and just upgrade serialization version.

NikkyAI: with `not failing`do you meant to try all serializers for a class and use the result of the first one that does not throw ?

NikkyAI:
now `clazz.serializer()` will only be called in case no matching serializer was found, on read and write, this should fix it for native i hope ?

NikkyAI:
next step would be to add tests for IOS, which i cannot do
you should be able to register serializers with
```
serializer = KotlinxSerializer().apply {
    serializer(Widget::class) {
       ...
    }
}
```

e5l:
Hi, @NikkyAI. Please ping me if the PR is ready for review.

NikkyAI:
@e5l 
I wonder if we should wait for https://github.com/Kotlin/kotlinx.serialization/commit/2937af7e6065553ca3628cef0084c3bfe0882036
to be in a release or copy the relevant parts and remove the duplicate code later ?

NikkyAI:
i think i implemented all features as good as i could given the limitations with the RequestPipeline

this could be merged now and be improved once #591 is adrdessed
or kept open

PR #576, opened 2018-09-12
[ktor-client] Pass AttributesBuilder through interceptor pipelines

enleur:
In my case I need to pass some attributes from call and get them after receive response in order to add logs and record metrics

### DIFF
--- a/ktor-client/ktor-client-core/src/io/ktor/client/request/HttpRequest.kt
+++ b/ktor-client/ktor-client-core/src/io/ktor/client/request/HttpRequest.kt
@@ -107,6 +107,19 @@ class HttpRequestBuilder : HttpMessageBuilder {
         attributesBuilder = { old(); block() }
     }
 
+    /**
+     * Mutates [this] copying all the data from another [builder] using it as base.
+     */
+    fun takeFrom(builder: HttpRequestBuilder): HttpRequestBuilder {
+        method = builder.method
+        body = builder.body
+        url.takeFrom(builder.url)
+        headers.appendAll(builder.headers)
+        attributesBuilder = builder.attributesBuilder
+
+        return this
+    }
+
     companion object
 }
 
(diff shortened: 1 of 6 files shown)


ktor-client/ktor-client-core/src/io/ktor/client/request/HttpRequest.kt
```diff
@@ -88,7 +88,7 @@ class HttpRequestBuilder : HttpMessageBuilder {
      */
     val executionContext: CompletableDeferred<Unit> = CompletableDeferred()
 
-    private var attributesBuilder: Attributes.() -> Unit = {}
+    var attributesBuilder: Attributes.() -> Unit = {}
```
enleur: It would be better to have `takeFrom` method inside the class so we do not expose `attributesBuilder`. What to you think?

e5l: Yep. Let's add the method.

ktor-client/ktor-client-tests/src/io/ktor/client/tests/AttributesTest.kt
```diff
@@ -0,0 +1,56 @@
...
+import io.ktor.client.tests.utils.test
+import io.ktor.response.respondText
+import io.ktor.routing.get
+import io.ktor.routing.routing
+import io.ktor.server.engine.ApplicationEngine
+import io.ktor.server.engine.embeddedServer
+import io.ktor.server.netty.Netty
+import io.ktor.util.AttributeKey
+import kotlin.test.Test
+import kotlin.test.assertEquals
```
e5l: Could you use `star` imports here?

e5l:
Hi, @enleur. Thanks for the PR.

LGTM. I could merge it after the imports fix.

enleur:
@e5l done

PR #585, opened 2018-09-14
 Added `TestApplicationEngine.cookiesSession` to keep cookies between test requests

soywiz:
* Added `TestApplicationEngine.cookiesSession` feature to keep cookie
* Added a `TestApplicationEngine.hookRequests` mechanism to hook requests in a scope
* Added a CurrentSession.getOrSet convenience method.

### DIFF
--- a/ktor-server/ktor-server-test-host/src/io/ktor/server/testing/TestApplicationEngine.kt
+++ b/ktor-server/ktor-server-test-host/src/io/ktor/server/testing/TestApplicationEngine.kt
@@ -142,12 +142,12 @@ fun TestApplicationEngine.cookiesSession(callback: () -> Unit) {
 
     hookRequests(
         processRequest = { setup ->
-            val cookieValue = trackedCookies.map { (it.name).encodeURLParameter() + "=" + (it.value).encodeURLParameter() }.joinToString("; ")
-            addHeader("Cookie", cookieValue)
+            val cookieValue = trackedCookies.joinToString("; ") { (it.name).encodeURLParameter() + "=" + (it.value).encodeURLParameter() }
+            addHeader(HttpHeaders.Cookie, cookieValue)
             setup() // setup after setting the cookie so the user can override cookies
         },
         processResponse = {
-            trackedCookies = response.headers.values("Set-Cookie").map { parseServerSetCookieHeader(it) }
+            trackedCookies = response.headers.values(HttpHeaders.SetCookie).map { parseServerSetCookieHeader(it) }
         }
     ) {
         callback()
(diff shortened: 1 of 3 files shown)


ktor-server/ktor-server-test-host/src/io/ktor/server/testing/TestApplicationEngine.kt
```diff
@@ -41,8 +41,35 @@ class TestApplicationEngine(
         environment.stop()
     }
 
+    private var processRequest: TestApplicationRequest.(setup: TestApplicationRequest.() -> Unit) -> Unit = { it() }
+    private var processResponse: TestApplicationCall.() -> Unit = { }
+
+    fun hookRequests(
```
cy6erGn0m: This looks like just interceptors. Why do we need specialized hooks here?

soywiz: I'm doing some things before the `setup: TestApplicationRequest.() -> Unit` provided by the user is executed.

I'm not aware of interceptors that allows to do things before constructing the test request.

The idea is that you set cookies there, but allows the user to override or remove them since their code is executed later.

The response part could be done with an interceptor indeed: installing once the interceptor, and then changing a flag in the attributes to see if it is enabled or a similar mechanism since this is scoped.

But since I needed the `processRequest` part I decided to do the two scoped hooks for simplicity.
At least this code is contained in the TestApplicationEngine part, without affecting the core of ktor.

Still, if you have a suggestion for intercepting the request building or another way of doing it, while allowing the user to control the cookies in a scoped way.

I can try to put an interceptor before the request, and check the set cookies from the user and do not re-set if they set them before. But that wouldn't allow to remove just one cookie for a specific request while preserving potentially another one.

ktor-server/ktor-server-test-host/src/io/ktor/server/testing/TestApplicationEngine.kt
```diff
@@ -98,9 +127,29 @@ class TestApplicationEngine(
...
         return call
     }
 
     fun createCall(readResponse: Boolean = false, setup: TestApplicationRequest.() -> Unit): TestApplicationCall =
         TestApplicationCall(application, readResponse).apply { setup(request) }
-}
\ No newline at end of file
+}
+
+fun TestApplicationEngine.cookiesSession(callback: () -> Unit) {
```
cy6erGn0m: Would you mind to add a short kdoc here

ktor-server/ktor-server-test-host/src/io/ktor/server/testing/TestApplicationEngine.kt
cy6erGn0m: ok, go ahead with hooks

PR #587, opened 2018-09-17
Ktor-JRPC support

Zeddushka:
Good day! 
I want to add JRPC support in easy-way with ktor, with DSL-like method adding.
It will be very interesting to hear critics and recommendations from your team :)

Example (model User is defined in other source. Dosc, Test included):

    
    fun Application.main() {
        install(DefaultHeaders)
        install(Compression) //not needed if deployed on tomcat with compression and HTTPS
        install(CallLogging) {
            level = Level.DEBUG
        }
    
        install(JrpcSupport) {
            gson {
                setDateFormat(DateFormat.LONG)
            }
            printStackTraces = true
        }
    
        val contextPath = environment.config.propertyOrNull("ktor.deployment.contextpath")?.getString() ?: "/"
    
        routing {
    
            route(contextPath) {
    
                get("/") { rootHandler() }
    
                jrpc("/json-rpc") {
    
                    method<User>("hello") { user ->
                        val token = call.request.headers["Session-Token"]
                        "Hello, ${user.name}${if (token == null) " you must authorize" else " your token is $token"}"
                    }
    
                    beforeJrpc { jrpcRequest ->
                        call.response.header("beforeJrpc", jrpcRequest.method)
                        null
                    }
    
                    afterJrpc { jrpcRequest, jrpcResponse ->
                        call.response.header("afterJrpc", jrpcRequest.method)
                        jrpcResponse
                    }
                }
    
                jrpc("/json-rpc-echo", echoRouter)
            }
        }
    }


ktor-features/ktor-jrpc/src/io/ktor/jrpc/extensions/Route.kt
```diff
@@ -0,0 +1,50 @@
...
+ * Builds a route to match specified path and handle with JSON-RPC handler
+ * @param path is a path for JRPC handler
+ * @param router is a already configured JRPC Router
+ */
+@ContextDsl
+fun Route.jrpc(path: String, router: JrpcRouter): Route {
+    return route(path) {
+        val allowedContentType = ContentType.Application.Json.toString()
+        handle({
+            if (context.request.httpMethod == HttpMethod.Post) {
```
cy6erGn0m: What about HEAD/OPTIONS? CORS?

ktor-features/ktor-jrpc/src/io/ktor/jrpc/extensions/Route.kt
```diff
@@ -0,0 +1,50 @@
...
+                val contentType = context.request.headers[HttpHeaders.ContentType]
+                if (contentType == allowedContentType) {
```
cy6erGn0m: Use `ContentType.match` instead of equals check as parameters such as character encoding will break it

ktor-features/ktor-jrpc/src/io/ktor/jrpc/routers/JrpcRouter.kt
```diff
@@ -0,0 +1,191 @@
...
+                        }
+                        if (methodParams == null) {
+                            JrpcResponse(requestId, error = JrpcError.INVALID_PARAMS)
+                        } else {
+                            JrpcResponse(requestId, handler(context, methodParams))
+                        }
+                    }
+                }
+            } catch (je: JrpcException) {
+                JrpcResponse(requestId, error = JrpcError("${je.message}:\r\n${
```
cy6erGn0m: Can't we handle it in `jrpcPostHandler` ?

ktor-features/ktor-jrpc/src/io/ktor/jrpc/routers/JrpcRouter.kt
```diff
@@ -0,0 +1,191 @@
...
+ */
+typealias JrpcMethodRouter = HashMap<String, JrpcRouteItem<*>>
+
+/**
+ * JSON-RPC router class
+ * Use GSON from [JrpcSupport] for converting JRPC body
+ * Handle JRPC-Requests using configured handlers
+ * Before JRPC and After JRPC hooks can be set
+ */
+open class JrpcRouter {
```
cy6erGn0m: Why is it open?

ktor-features/ktor-jrpc/src/io/ktor/jrpc/routers/JrpcRouter.kt
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    /**
+     * If handler parse JrpcRequest properly, will called before handle any JRPC-method
+     * Usable for checks, headers, etc
+     * @param block - block of code that will be executed. If block return JrpcResponse, the other JRPC methods will not be handled, the method returns immediately with that JrpcResponse. Note that params will be not parsed at the moment.
+     * @throws RuntimeException if called twice
+     */
+    @ContextDsl
+    fun beforeJrpc(block: PipelineContext<Unit, ApplicationCall>.(JrpcRequest) -> JrpcResponse?) {
```
cy6erGn0m: It looks like we need a pipeline here instead of before/after hooks. Similar to `ApplicationCallPipeline` we can use `JrpcCallPipeline` with appropriate phases

ktor-features/ktor-jrpc/src/io/ktor/jrpc/routers/JrpcRouter.kt
Zeddushka: For constructions like this 

    jrpc("/json-rpc", MainRouter(environment.config))

and the MainRouter is defined in other file and use some internal logic (lacal variables, functions, etc.). In this case, kotlin file with routing will more readable without lists of methods implementations.
Or this is not a Ktor-way?

Zeddushka:
Thank you for your review, @cy6erGn0m , I will work on suggested solutions!

ktor-features/ktor-jrpc/src/io/ktor/jrpc/routers/JrpcRouter.kt
cy6erGn0m: Well, you can always extract your builder function to a separate file

cy6erGn0m: `jrpc("/json-rpc", ::mainRouter)` or `jrpc("/json-rpc") { mainRouter() }`

Zeddushka: That's right, so you recommend to change JrpcRouter to default (final) class?

cy6erGn0m: Yes, exactly. Users can extract function/class anyway so there is no need to make router open for inheritance. Also it could be internal (? not sure)

Issue #588, opened 2018-09-18
Client does not support sending or receive json null value


spand:
How does one send and receive the json `null` value ?

I had a look at implementing the change myself but it seems the non-nullability of the various content properties are quite pervasive so I am not sure how such a change would be most desirable to implement. I get the feeling that making them all nullable isnt appropriate.

According to https://tools.ietf.org/html/rfc8259#section-2 it seems to be valid.
```
2.  JSON Grammar

   A JSON text is a sequence of tokens.  The set of tokens includes six
   structural characters, strings, numbers, and three literal names.

   A JSON text is a serialized value.  Note that certain previous
   specifications of JSON constrained a JSON text to be an object or an
   array.  Implementations that generate only objects or arrays where a
   JSON text is called for will be interoperable in the sense that all
   implementations will accept these as conforming JSON texts.
```

spand:
Could you elaborate on my question if you are not working on this yourself @e5l ?

oleg-larshin:
Please check the <a href="https://youtrack.jetbrains.com/issue/KTOR-745">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

PR #594, opened 2018-09-20
Initial client-json support of json values (null, string, ...)

spand:
Not tested on other platforms and serverside json support is still lacking.

First step at fixing #588 by wrapping object/value meant for json (de)serialization. This seems like the best solution without touching the core machinery.

Can you provide some feedback ?

### DIFF
--- a/ktor-client/ktor-client-features/ktor-client-json/ktor-client-json-jvm/src/io/ktor/client/features/json/DefaultJvm.kt
+++ b/ktor-client/ktor-client-features/ktor-client-json/ktor-client-json-jvm/src/io/ktor/client/features/json/DefaultJvm.kt
@@ -1,6 +1,9 @@
 package io.ktor.client.features.json
 
+import io.ktor.client.call.TypeInfo
+import java.lang.reflect.ParameterizedType
 import java.util.*
+import kotlin.reflect.full.isSubclassOf
 
 actual fun defaultSerializer(): JsonSerializer {
     val serializers = ServiceLoader.load(JsonSerializer::class.java).toList()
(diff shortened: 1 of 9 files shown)


ktor-client/ktor-client-features/ktor-client-json/ktor-client-json-jvm/src/io/ktor/client/features/json/DefaultJvm.kt
```diff
@@ -1,6 +1,9 @@
 package io.ktor.client.features.json
 
+import io.ktor.client.call.TypeInfo
```
e5l: Please consider using `*` imports like everywhere in ktor. It could be configured in IDEA kotlin code style

ktor-client/ktor-client-features/ktor-client-json/src/io/ktor/client/features/json/JsonFeature.kt
```diff
@@ -48,21 +56,34 @@ class JsonFeature(val serializer: JsonSerializer) {
...
-                ) return@intercept
-
-                proceedWith(HttpResponseContainer(info, feature.serializer.read(info, response)))
+                if (context.response.contentType()?.match(ContentType.Application.Json) == true){
+                    val responseToParse = when (response){
+                        is HttpResponse -> response
+                        is String -> ResponseWithContent(context.response, response) // Undo HttpPlainText feature
+                        else -> return@intercept
+                    }
+                    val unwrappedInfo = unwrapJsonContent(info)
```
e5l: Could you add test for not `JsonContent` response in info? 
Should we really fail here?

e5l:
Hi, @span. Thanks for the PR. 
Lgtm for the most stuff, but there are no tests.

Could you add some?

ktor-client/ktor-client-features/ktor-client-json/src/io/ktor/client/features/json/JsonFeature.kt
spand: No. It seems more symmetric to pass on it here also.

spand:
> 
> 
> Hi, @span. Thanks for the PR.
> Lgtm for the most stuff, but there are no tests.
> 
> Could you add some?

I am not sure what other tests to add ? I have added six in JsonTest.

e5l:
It looks like some json tests are broken. Could you take a look?

spand:
Should work now.

A nice to have would be to use the same tests for all implementations as a contract test but I dont have the time to it.

Also, since KotlinxSerializer.kt is unused my modifications to that have not been tested.

PR #613, opened 2018-09-28
RFC4180: Add ContentType for text/csv

JLLeitschuh:
This Content Type is part of this RFC:
https://tools.ietf.org/html/rfc4180

### DIFF
--- a/ktor-http/src/io/ktor/http/ContentTypes.kt
+++ b/ktor-http/src/io/ktor/http/ContentTypes.kt
@@ -184,7 +184,7 @@ class ContentType private constructor(val contentType: String, val contentSubtyp
         val Any = ContentType("text", "*")
         val Plain = ContentType("text", "plain")
         val CSS = ContentType("text", "css")
-        val Csv = ContentType("text", "csv")
+        val CSV = ContentType("text", "csv")
         val Html = ContentType("text", "html")
         val JavaScript = ContentType("text", "javascript")
         val VCard = ContentType("text", "vcard")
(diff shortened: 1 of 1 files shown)


ktor-http/src/io/ktor/http/ContentTypes.kt
```diff
@@ -216,4 +217,4 @@ fun ContentType.withCharset(charset: Charset) = withParameter("charset", charset
 /**
  * Extracts a [Charset] value from the given `Content-Type`, `Content-Disposition` or similar header value.  
  */
-fun HeaderValueWithParameters.charset() = parameter("charset")?.let { Charset.forName(it) }
\ No newline at end of file
+fun HeaderValueWithParameters.charset() = parameter("charset")?.let { Charset.forName(it) }
```
JLLeitschuh: I created this PR with the github UI (didn't even check out the code locally). It seems to automatically add these when you do that. Sorry if this is an annoyance.

ktor-http/src/io/ktor/http/ContentTypes.kt
```diff
@@ -184,6 +184,7 @@ class ContentType private constructor(val contentType: String, val contentSubtyp
         val Any = ContentType("text", "*")
         val Plain = ContentType("text", "plain")
         val CSS = ContentType("text", "css")
+        val Csv = ContentType("text", "csv")
```
cy6erGn0m: I'd say this need to be `CSV` as we already have abbreviations uppercased

JLLeitschuh: Fixed

cy6erGn0m:
Thanks!

PR #621, opened 2018-10-02
Add content types for csv and pdf

marshallpierce:
I'm not sure what the criteria are for adding content types to this list, so feel free to reject if these aren't what you're after, but csv and pdf seemed at least no more obscure than vcard or ogg. :)

On my system (x64 linux, Java 10) ktor's test suite has a number of timed out tests and `UnknownHostException` so I can't say that this passes the tests, but I'm guessing that adding two content types doesn't break much.


cy6erGn0m:
CSV is already added as per another PR, sorry

PR #628, opened 2018-10-07
Introduce HttpResponseConfig for client & enable to set custom default response charset rather than ISO-8859-1

hltj:
Some HTTP servers maybe send a Content-Type header with MIME-type only( i.e. without charset), e.g. `text/plain`, `application/json`, etc. And if the real encoding of the response content is ISO-8859-1, the HTTP client now can handle the text/JSON correctly. But if the content is other encoding (e.g. UTF-8), the result of `HttpClient.get<String>` or `HttpClient.get<AComplexObject>` may contains messy code.

This change is for solving the above problem manually. Added a `HttpResponseConfig` property to the `HttpClientEngineConfig` and this enabled to set a custom default charset of response content rather than the hard coded ISO-8859-1.

e.g. To request a server URL which respond a json contains **UTF-8** strings with the Content-Type `application/json`, we can configure the HTTP client as followed:

``` kotlin
val httpClient = HttpClient() {
    install(JsonFeature) {
        serializer = JacksonSerializer  {
        }
    }
    engine {
        response.apply {
            defaultCharset = Charsets.UTF_8
        }
        // ...
    }
}
```

### DIFF
--- a/ktor-client/ktor-client-core/src/io/ktor/client/call/HttpClientCall.kt
+++ b/ktor-client/ktor-client-core/src/io/ktor/client/call/HttpClientCall.kt
@@ -28,6 +28,11 @@ class HttpClientCall internal constructor(
         internal set
 
     /**
+     * Configuration for the [response].
+     */
+    val responseConfig: HttpResponseConfig = client.engineConfig.response
+
+    /**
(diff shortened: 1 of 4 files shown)


e5l:
Hi, @hltj. Thanks for the PR :)

Lgtm

PR #629, opened 2018-10-07
Provide user-defined function to configure Netty's HttpServerCodec

### LINKED ISSUE #580: Netty configuration doesn't allow configuring request max length
We need to be able to support urls of length 8kb, however, NettyChannelInitializer has no way of customizing the HttpServerCodec added to the pipeline.

https://github.com/ktorio/ktor/blob/master/ktor-server/ktor-server-netty/src/io/ktor/server/netty/NettyChannelInitializer.kt#L101

HttpServerCodec (maxInitialLineLength is relevant argument):
https://netty.io/4.0/api/io/netty/handler/codec/http/HttpServerCodec.html

renannprado:
fixes #580


cy6erGn0m:
Thanks!

PR #639, opened 2018-10-12
Js engine

e5l:


ktor-client/ktor-client-core/ktor-client-core-js/src/io/ktor/client/engine/js/ReadableStream.kt
```diff
@@ -0,0 +1,42 @@
...
+        val chunk = reader.readChunk() ?: break
+        channel.writeFully(chunk.toByteArray())
+    }
+}.channel
+
+internal suspend fun ReadableStreamReader.readChunk(): Uint8Array? = suspendCancellableCoroutine { continuation ->
+    read().then {
+        val chunk = it.value
+        val result = Result.success(if (it.done || chunk == null) null else chunk)
+        continuation.resumeWith(result)
```
cy6erGn0m: Any possible unsuccessful results?

ktor-client/ktor-client-core/ktor-client-core-js/src/io/ktor/client/engine/js/ReadableStream.kt
```diff
@@ -0,0 +1,42 @@
...
+    }
+}
+
+internal fun Uint8Array.toByteArray(): ByteArray = ByteArray(length) { get(it) }
```
cy6erGn0m: Something like 

```
    @Suppress("UnsafeCastFromDynamic")
    private fun Uint8Array.toByteArray(): ByteArray {
        return Int8Array(buffer, byteOffset, length).asDynamic()
    }
```

ktor-client/ktor-client-core/ktor-client-core-js/src/io/ktor/client/features/DefaultTransformJs.kt
```diff
@@ -0,0 +1,5 @@
+package io.ktor.client.features
+
+import io.ktor.client.*
+
+internal actual fun HttpClient.platformDefaultTransformers() {}
```
cy6erGn0m: `Blob`? (FYI, ok to not support in the first implementation)

ktor-client/ktor-client-core/ktor-client-core-js/src/io/ktor/client/utils/CIOJs.kt
```diff
@@ -0,0 +1,13 @@
...
+
+/**
+ * Ios http client use the main queue to execute request.
+ */
+actual val HTTP_CLIENT_THREAD_COUNT: Int = 0
+
+/**
+ * Default [IOCoroutineDispatcher] that uses [HTTP_CLIENT_THREAD_COUNT] as the number of threads.
+ */
+actual val HTTP_CLIENT_DEFAULT_DISPATCHER: CoroutineDispatcher = Dispatchers.Default
```
cy6erGn0m: This need to be deprecated with error

ktor-client/ktor-client-js/src/Js.kt
```diff
@@ -0,0 +1 @@
+
```
cy6erGn0m: ?

ktor-utils/ktor-utils-js/src/io/ktor/util/date/DateJs.kt
```diff
@@ -0,0 +1,34 @@
+package io.ktor.util.date
+
+import kotlin.js.*
+
+actual fun GMTDate(timestamp: Long?): GMTDate {
+    val date = timestamp?.let { Date(it) } ?: Date()
```
cy6erGn0m: How does it work? `Long` is not JS native type so how can we pass it to JS native `Date`?

soywiz: I guess it misses a `toDouble`. But I wonder about the signature. Is it a Number or dynamic? Maybe it works converting to string then to double?

ktor-http/test/io/ktor/tests/http/HeadersTest.kt
```diff
@@ -181,7 +181,7 @@ class HeadersTest {
     }
 
     @Test
-    fun `headersOf should be case insensitive`() {
+    fun headersOfShouldBeCaseInsensitive() {
```
soywiz: Hopefully this would be supported in the future in all the backends or at least in none. But in js it is possible using `obj[“my property”]`

ktor-client/ktor-client-core/ktor-client-core-js/src/io/ktor/client/engine/js/ReadableStream.kt
soywiz: I would use `asByteArray` since it shares the buffer and arrays are mutable is more like a cast instead of a conversion.

ktor-client/ktor-client-core/ktor-client-core-js/src/io/ktor/client/engine/js/JsClientEngine.kt
```diff
@@ -0,0 +1,42 @@
...
+    override val coroutineContext: CoroutineContext = dispatcher + SupervisorJob()
+
+    override suspend fun execute(
+        call: HttpClientCall, data: HttpRequestData
+    ): HttpEngineCall = withContext(dispatcher) {
+        val callContext = CompletableDeferred<Unit>(this@JsClientEngine.coroutineContext[Job]) + dispatcher
+
+        val requestTime = GMTDate()
+        val request = DefaultHttpRequest(call, data)
+        val rawResponse = fetch(request.url, request.toRaw())
```
soywiz: I guess that fetch is already pretty compatible.
Would be nice if a later version supports nodejs too.
Maybe two engines (JsBrowserEngine, NodeJsEngine) and this one detecting if running on the browser or nodejs and calling the appropiate subengine?

ktor-utils/ktor-utils-js/src/io/ktor/util/date/DateJs.kt
e5l: Let's add explicit `toDouble` call here

ktor-client/ktor-client-core/ktor-client-core-js/src/io/ktor/client/engine/js/JsClientEngine.kt
e5l: There is an issue with service loader: I can't extract the js engine into a separate dependency.
So we could have only 1 js engine for now. :(

soywiz: Well, maybe an if and couple of methods would do the job. Shouldt be too large.

PR #642, opened 2018-10-15
Prepare to migration to 1.0.0-RC

qwwdfsad:


ktor-http/ktor-http-cio/src/io/ktor/http/cio/internals/WeakTimeoutQueue.kt
```diff
@@ -142,6 +142,7 @@ class WeakTimeoutQueue(
         }
     }
 
+    @UseExperimental(InternalCoroutinesApi::class)
     private class WeakTimeoutCoroutine<in T>(context: CoroutineContext, val delegate: Continuation<T>) :
```
qwwdfsad: `WeakTimeoutCoroutine` has probably the same bugs as `TimeoutCoroutine` had.
I've tried to comment some parts of cancellation machinery here and there, but none of `cio` tests have failed, so I decided not to touch this code.

ktor-server/ktor-server-servlet/src/io/ktor/server/servlet/ServletUpgrade.kt
```diff
@@ -94,6 +94,8 @@ class ServletUpgradeHandler : HttpUpgradeHandler, CoroutineScope {
         launch(up.userContext, start = CoroutineStart.UNDISPATCHED) {
             val job = up.upgradeMessage.upgrade(inputChannel, outputChannel, up.engineContext, up.userContext)
 
+            // TODO it's parent-child?
```
qwwdfsad: Why are you not using attachChild API here?

cy6erGn0m: After discussion with Roman I'd remove it after beta release

ktor-http/ktor-http-cio/src/io/ktor/http/cio/internals/WeakTimeoutQueue.kt
cy6erGn0m: Yes, this need to be updated to be in-sync with `TimeoutCoroutine`

PR #650, opened 2018-10-16
Improve KotlinxSerializer

sommd:
Not sure if about the name of `register()`. Might be more consistent to call it `setMapper()` as well, but that name makes less sense since there's only one parameter.


e5l:
Hi, @sommd. Thanks for the PR. 
Could you add some serialization tests?

sommd:
Refactored tests from ktor-client-gson/jackson/json/json-jvm to share a common base class (`JsonTest`). Currently the tests for `KotlinxSerializer` don't pass, since it doesn't support parameterized types. I'll look into getting it working.

e5l:
It looks like the PR run issue with `KotlinxJsonTest.testSerializeNested` test. Could you take a look?

sommd:
Fixed now, I made a mistake writing the test.

e5l:
Lgtm. Thanks for the PR :)

PR #651, opened 2018-10-16
Cy/cio non parallel fix

cy6erGn0m:
This is required to break unconfined loop in the copy pipe that blocks
selector's loop forever and cause the whole server to stuck


cy6erGn0m:
Merged manually

PR #663, opened 2018-10-19
 Add kotlinx-serialization ContentConverters for JSON, CBOR and ProtoBuf

sommd:


PPACI:
This PR would be great !

felipehjcosta:
I agree! @PPACI

MrChoclate:
This is great !
I have two questions on this converters. Why register serializers and not use type.serializer() and how do you handle generic types like `call.receive<List<MyEntity>>()` ?

cy6erGn0m:
It looks like we have `ktor-serialization` module, so we need to generalize it.

PR #665, opened 2018-10-20
Properly construct request headers in JS engine

DrewCarlson:
Requests do not currently send content headers.


cy6erGn0m:
Thanks

PR #695, opened 2018-11-02
Allows customizing JSON instance

Syhids:
This permits setting non-strict mode, for instance.


e5l:
Hi, @Syhids.
Thanks for the PR. LGTM

PR #706, opened 2018-11-06
Annotation base Controllers feature

blurkidi:
The Controllers feature allows annotating a class and its methods to define routes. For example, the following class defines a simple Get mapping:

```
    @RouteController
    class MyController {
        @Get("/hello/{name}")
        fun hello(@PathParam name: String): String = "Hello $name!"
    }
```

The feature needs to be installed as usual. Then each controller must be set up in routing feature:

```
    install(Controllers)
    routing {
        setupController(MyController())
    }

```
**Motivation**

This is a feature similar to what we find in frameworks like Spring. Such code is easy to read and maintain. I personally think that it would also improve the adoption of Ktor by people that are used to this coding style.


orangy:
I believe controller's functions should at least be `suspend`?

forseti:
Hi,

It would be cool if we could have ```setupControllers``` that register
* ```vararg``` of ```Controller()```
* ```vararg``` of ```Pair(path, Controller())```

Thanks!!!

blurkidi:
@forseti It is a nice addition, indeed. I personally find separate entries more readable, but I understand other might prefer this coding style. In that case I would include "setupControllers(path, vararg controllers)" as well.

blurkidi:
Some thoughts about the feature:

**NAMING**
- The feature is named _controllers_ while the annotation is @RouteController. I found that @Controller would be too generic.
- I settled with 'setupController' for the method name. Other names I thought of are 'install' (this is used for features thought), 'map', 'route' and 'config'
- Annotation naming is similar to the one in Spring framework, but I decided to respect the Ktor naming. [Spring @PathVariable] -> [Ktor @PathParam], because it is called ApplicationCall.parameters in Ktor.

**FEATURES**
- I would still like to add a @AuthPrincipal param detector. I think it is a must.
- Spring can scan all controllers and integrate the automatically while in Ktor Controllers, we need to set them up 'manually'. I think this fits better with the Ktor philosophy, where everything is more intentional, less magical. Nevertheless, I thought of implementing an 'autoSetupControllers' feature that would scan all @RouteController annotations and set them up. This is no trivial task but it would be nice to look into it in the future.

**CORE IMPLEMENTATION**
The current implementation scans annotations at runtime and performs the calls using reflection. This is the same approach as the Locations feature.
The other option was to implement a gradle plugin that would generate the code that would set up the controllers. It is a more complex implementation but it should be more performant in theory. It could be nice to do some performance tests.
I think that both approaches are compatible. In the future, the same @RouteController annotated class might be set up using reflection or by using the code generated by the plugin. The resulting functionality should be the same.

Now, I would love to know your opinion about all these subjects. I will be happy adapt the pull request based on the feedback.

forseti:
Thanks @blurkidi

```@RouteController``` is a cool name.

I think for auto setup, scanning list of defined packages would be better? (non-recursive)

```autoSetupControllers(vararg packages)```

Thanks again!!!

cy6erGn0m:
The main disadvantage of annotation scanning is poor startup time. Perhaps we need to build a list controller classes at build time for every module and configure include/exclude list at runtime. Similar to what JMH does for benchmark classes.

ssijak:
Just my 2 cents. The reason why I will try ktor for one of my pet projects instead of Spring which I use for years is lack (or minimal) magic and annotations. I like Spring, but "magic" and slow startup times (caused in part because of reflection/annotations/classpath scanning), can really get on your nerves when you want to iterate quickly on some feature and when you really want to know and control what the application is doing.

blurkidi:
Thaks for the feedback. It really confirms the assumptions I made in the first place. On one hand, explicitly declaring each contoller is more Ktor-like and avoids the "magic" of other frameworks. On the other hand, scanning for controllers may lead to startup/performance issues.
In conclusion, we should keep the current strategy to declare each controller explicitly.
Thanks again for the feedback and don't hesitate to comment on naming, implementation, functionally...

Issue #713, opened 2018-11-09
ktor-features: Add Mustache feature


erdlet:
To have a more lightweight templating engine compared to Freemarker or Velocity, it'd be nice to have
a Mustache feature.

Referenced implementation by [Mustache](https://mustache.github.io/): https://github.com/spullara/mustache.java

e5l:
Released in `1.1.1`

PR #714, opened 2018-11-09
Add Mustache templating feature (#713)

erdlet:


erdlet:
Hey team,
is there any progress on this PR? Or at least some plan from your side if it has a chance to get part of ktor or not?

ktor-features/ktor-mustache/test/io/ktor/mustache/MustacheTest.kt
```diff
@@ -0,0 +1,161 @@
...
+                }
+            }
+
+            assertEquals(null, handleRequest(HttpMethod.Get, "/").response.headers[HttpHeaders.ETag])
+        }
+    }
+
+    private fun Application.setupMustache() {
+        install(Mustache) {
+            DefaultMustacheFactory()
```
cy6erGn0m: This seems to be redundant isn't it?

erdlet: Yeah that could be possible. I'll have a look on it

erdlet: You were right, I implemented the feature configuration wrong there. 
Now I added an explicit configuration, so the usage of other Factories is possible, but default is DefaultMustacheFactory

ktor-features/ktor-mustache/src/io/ktor/mustache/Mustache.kt
```diff
@@ -0,0 +1,90 @@
...
+
+/**
+ * Feature for providing Mustache templates as [MustacheContent]
+ */
+class Mustache(configuration: Configuration) {
+
+    private val mustacheFactory = configuration.mustacheFactory
+
+    class Configuration {
+        var mustacheFactory = DefaultMustacheFactory()
```
cy6erGn0m: Shouldn't it have type `MustacheFactory` instead of `DefaultMustacheFactory` so user's will be able to use other implementation classes?

erdlet: I shouldn't code after a whole day of meetings... yes that is completely right, I forgot that mustacheFactory gets the type `DefaultMustacheFactory`

erdlet: OK the configuration's `mustacheFactory` is now typed with the interface so other
Factory implementations are applicable.

cy6erGn0m:
Merged manually

PR #718, opened 2018-11-11
Add micrometer metrics

dave08:
Continuation of https://github.com/ktorio/ktor/pull/480


dave08:
@orangy Is this what you meant by two modules?

oripwk:
Hi,

What's the status of this?

dave08:
I just wanted a confirmation if I set up the modules the way @orangy wanted. Now I'm in the middle of a release on an internal project, at the end of it I'll need to update this Feature again, hopefully within the next two weeks.

I still think that micrometer can completely replace the existing implementation since it also supports Dropwizard, but if they want to support both, that's also fine.

You should be able to copy the code into your project almost mostly as is and just add a few buckets for the timing meter (I didn't add them because they depend on your SLA, but was planning to add an option for such a list in the config block of the Feature).

JotBePunkt:
replaced by #1037

cy6erGn0m:
@dave08 would you mind if I close it since we are going to introduce 1037?

dave08:
Closed :smile:

Issue #696, opened 2018-11-02
Support RFC 850 date format


Tea-Ayataka:
Currently, the http client doesn't support the date format of RFC 850.
But actually, some websites are still using it. including Cloudflare.
I think it wouldn't be a bad idea to support the old format.

e5l:
Hi, @Tea-Ayataka. 
Thanks for the report. It's really good idea :)

PR #719, opened 2018-11-12
Support RFC 850 date format #696

lempere:


ktor-utils/src/io/ktor/util/date/ParseDate.kt
```diff
@@ -0,0 +1,18 @@
...
+
+    return try {
+        val subDate = rfc850.subSequence(rfc850.indexOf(',')+2, rfc850.length)
+        val parts = subDate.split(' ')
+        val date = parts[0].split('-')
+        val time = parts[1].split(':')
+
+        GMTDate(time[2].toInt(),time[1].toInt(),time[0].toInt(), date[0].toInt(), Month.from(date[1]), yearCommonEra( date[2].toInt() ) )
+    }catch (ex:Exception){
+        GMTDate.START
```
e5l: Maybe we should throw `FailToParseDateException` here?

ktor-utils/src/io/ktor/util/date/ParseDate.kt
```diff
@@ -0,0 +1,18 @@
...
+    }
+
+}
+
+fun yearCommonEra(year: Int) = if (year<70) year + 2000 else year + 1900
```
e5l: private or internal

e5l:
Hi @lempere, thanks for the PR.
Could you use `*` imports everywhere as in ktor code style?
And probably enable autoformatting to fix minor formatting issues.

ktor-utils/src/io/ktor/util/date/ParseDate.kt
lempere: Or maybe null could do the job

e5l: `null` is for missing

PR #724, opened 2018-11-14
Reduce number of internal kotlinx.coroutines dependencies

cy6erGn0m:
The only remaining dependencies are:
- `invokeOnCompletion(cancelling = true) { }`
- `ExperimentalCoroutinesDispatcher` and related functions

### DIFF
--- a/ktor-http/ktor-http-cio/src/io/ktor/http/cio/internals/WeakTimeoutQueue.kt
+++ b/ktor-http/ktor-http-cio/src/io/ktor/http/cio/internals/WeakTimeoutQueue.kt
@@ -1,6 +1,7 @@
 package io.ktor.http.cio.internals
 
 import io.ktor.util.*
+import kotlinx.atomicfu.*
 import kotlinx.coroutines.*
 import kotlinx.coroutines.internal.*
 import java.time.*
@@ -21,7 +22,6 @@ import kotlin.coroutines.intrinsics.*
  *  as we really don't care about stalling IDLE connections if there are no more incoming
  */
 @InternalAPI
-@Suppress("INVISIBLE_REFERENCE", "INVISIBLE_MEMBER")
 class WeakTimeoutQueue(
     private val timeoutMillis: Long,
     private val clock: Clock = Clock.systemUTC(),
@@ -35,7 +35,7 @@ class WeakTimeoutQueue(
     /**
      * Register [job] in this queue. It will be cancelled if doesn't complete in time.
      */
-    fun register(job: Job): DisposableHandle {
+    fun register(job: Job): Registration {
         val now = clock.millis()
         val head = head
         if (cancelled) throw cancellationException()
@@ -74,34 +74,29 @@ class WeakTimeoutQueue(
     suspend fun <T> withTimeout(block: suspend CoroutineScope.() -> T): T {
         return suspendCoroutineUninterceptedOrReturn { rawContinuation ->
             val continuation = rawContinuation.intercepted()
+
             val wrapped = WeakTimeoutCoroutine(continuation.context, continuation)
             val handle = register(wrapped)
-            wrapped.disposeOnCompletion(handle)
+            wrapped.invokeOnCompletion(handle)
 
             val result = try {
-                if (wrapped.isCancelled) throw wrapped.getCancellationException()
-                block.startCoroutineUninterceptedOrReturn(receiver = wrapped, completion = wrapped)
+                if (wrapped.isCancelled) COROUTINE_SUSPENDED
+                else block.startCoroutineUninterceptedOrReturn(receiver = wrapped, completion = wrapped)
             } catch (t: Throwable) {
-                CompletedExceptionally(t)
+                if (wrapped.tryComplete()) {
+                    handle.dispose()
+                    throw t
+                }
+                else COROUTINE_SUSPENDED
             }
 
-            unwrapResult(wrapped, handle, result)
-        }
-    }
-
-    private fun unwrapResult(c: WeakTimeoutCoroutine<*>, handle: DisposableHandle, result: Any?): Any? {
-        val suspended = COROUTINE_SUSPENDED
-        return when {
-            result === suspended -> result
-            c.isCompleted -> suspended
-            result is CompletedExceptionally -> {
-                handle.dispose()
-                throw result.cause
-            }
-            else -> {
-                handle.dispose()
-                result
-            }
+            if (result !== COROUTINE_SUSPENDED) {
+                if (wrapped.tryComplete()) {
+                    handle.dispose()
+                    result
+                }
+                else COROUTINE_SUSPENDED
+            } else COROUTINE_SUSPENDED
         }
     }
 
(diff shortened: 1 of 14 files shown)


ktor-http/ktor-http-cio/src/io/ktor/http/cio/internals/WeakTimeoutQueue.kt
```diff
@@ -74,34 +74,29 @@ class WeakTimeoutQueue(
...
-            c.isCompleted -> suspended
-            result is CompletedExceptionally -> {
-                handle.dispose()
-                throw result.cause
-            }
-            else -> {
-                handle.dispose()
-                result
-            }
+            if (result !== COROUTINE_SUSPENDED) {
```
e5l: &&

ktor-http/ktor-http-cio/src/io/ktor/http/cio/internals/WeakTimeoutQueue.kt
```diff
@@ -74,34 +74,29 @@ class WeakTimeoutQueue(
     suspend fun <T> withTimeout(block: suspend CoroutineScope.() -> T): T {
```
qwwdfsad: KDoc to with timeout: "Execute [block] and cancel if doesn't complete it time." 
it time -> in time

ktor-http/ktor-http-cio/src/io/ktor/http/cio/internals/WeakTimeoutQueue.kt
```diff
@@ -74,34 +74,29 @@ class WeakTimeoutQueue(
...
         return suspendCoroutineUninterceptedOrReturn { rawContinuation ->
             val continuation = rawContinuation.intercepted()
+
             val wrapped = WeakTimeoutCoroutine(continuation.context, continuation)
             val handle = register(wrapped)
-            wrapped.disposeOnCompletion(handle)
+            wrapped.invokeOnCompletion(handle)
```
qwwdfsad: If `withTimeout` provides a scope, then it should wait for all children (and aggregate their exceptions if there are some) before exiting that scope.
E.g.
```
q.withTimeout {
                launch(start = CoroutineStart.ATOMIC) {
                    Thread.sleep(2000)
                    println("Child finished")
                }
                throw MyException()
            }
```

what is expected behaviour here?
In kx.coroutines there it will wait for the child before throwing exception

ktor-http/ktor-http-cio/src/io/ktor/http/cio/internals/WeakTimeoutQueue.kt
```diff
@@ -74,34 +74,29 @@ class WeakTimeoutQueue(
...
 
             val result = try {
-                if (wrapped.isCancelled) throw wrapped.getCancellationException()
-                block.startCoroutineUninterceptedOrReturn(receiver = wrapped, completion = wrapped)
+                if (wrapped.isCancelled) COROUTINE_SUSPENDED
```
qwwdfsad: `if (wrapped.isCancelled) COROUTINE_SUSPENDED`
But cancellation exception should be thrown instead immediately?

ktor-http/ktor-http-cio/src/io/ktor/http/cio/internals/WeakTimeoutQueue.kt
```diff
@@ -74,34 +74,29 @@ class WeakTimeoutQueue(
     suspend fun <T> withTimeout(block: suspend CoroutineScope.() -> T): T {
         return suspendCoroutineUninterceptedOrReturn { rawContinuation ->
             val continuation = rawContinuation.intercepted()
+
             val wrapped = WeakTimeoutCoroutine(continuation.context, continuation)
             val handle = register(wrapped)
```
qwwdfsad: Performance optimization: you can first invoke the block and add it to the queue if and only if `COROUTINE_SUSPENDED` was a result of the block

ktor-http/ktor-http-cio/src/io/ktor/http/cio/internals/WeakTimeoutQueue.kt
```diff
@@ -74,34 +74,29 @@ class WeakTimeoutQueue(
...
-                block.startCoroutineUninterceptedOrReturn(receiver = wrapped, completion = wrapped)
+                if (wrapped.isCancelled) COROUTINE_SUSPENDED
+                else block.startCoroutineUninterceptedOrReturn(receiver = wrapped, completion = wrapped)
             } catch (t: Throwable) {
-                CompletedExceptionally(t)
+                if (wrapped.tryComplete()) {
+                    handle.dispose()
+                    throw t
+                }
+                else COROUTINE_SUSPENDED
```
qwwdfsad: This code path is not covered by tests and it's unclear why is it here.
Is it protection against concurrent cancel + completion race? If so, it should be outcome of the coroutine, not `COROUTINE_SUSPENDED `

ktor-http/ktor-http-cio/src/io/ktor/http/cio/internals/WeakTimeoutQueue.kt
```diff
@@ -74,34 +74,29 @@ class WeakTimeoutQueue(
...
-            result is CompletedExceptionally -> {
-                handle.dispose()
-                throw result.cause
-            }
-            else -> {
-                handle.dispose()
-                result
-            }
+            if (result !== COROUTINE_SUSPENDED) {
+                if (wrapped.tryComplete()) {
```
qwwdfsad: Same here about `COROUTINE_SUSPENDED`

ktor-http/ktor-http-cio/src/io/ktor/http/cio/internals/WeakTimeoutQueue.kt
```diff
@@ -142,15 +145,48 @@ class WeakTimeoutQueue(
...
+        override val coroutineContext: CoroutineContext get() = context
+
+        private val state = atomic<Continuation<T>?>(delegate)
+
+        init {
+            context[Job]?.let { parent ->
+                @UseExperimental(InternalCoroutinesApi::class)
+                parent.invokeOnCompletion(onCancelling = true) {
+                    if (it != null) {
+                        resumeWithException(it)
```
qwwdfsad: You are invoking `resumeWithException` twice.
Once here and once in `job.invokeOnCompletion` below

ktor-http/ktor-http-cio/src/io/ktor/http/cio/internals/WeakTimeoutQueue.kt
```diff
@@ -142,15 +145,48 @@ class WeakTimeoutQueue(
...
+                null
+            }?.let {
+                it.resumeWith(result)
+                job.cancel()
+            }
         }
 
-        override fun onCompletedExceptionally(exception: Throwable) {
-            delegate.resumeWithException(exception)
+        fun tryComplete(): Boolean {
```
qwwdfsad: Semantics around `tryComplete` is a bit unclear, especially with`COROUTINE_SUSPENDED` part and it is very similar to our internal `tryResume`.

What you can do is to introduce your own method `tryResume + completeResume` (or indicate success of `tryResume` with nullability) and make all actions atomic instead of `tryComplete` + suspended

ktor-http/ktor-http-cio/src/io/ktor/http/cio/internals/WeakTimeoutQueue.kt
cy6erGn0m: If it is already cancelled then it was already resumed this is why we should suspend here. However it could be improved for sure, just not priority for now

ktor-http/ktor-http-cio/src/io/ktor/http/cio/internals/WeakTimeoutQueue.kt
cy6erGn0m: Similar to the pervious case, if it is already cancelled concurrently we need to return `COROUTINE_SUSPENDED` since the continuation is already resumed. Passing it to `result` will cause `suspendCoroutineUninterceptedOrReturn` return `COROUTINE_SUSPENDED` as well. However with the propsed optimization it becomes impossible so I'll simplify it.

PR #730, opened 2018-11-15
Fix documentation inaccuracy in stop()

magott:
Fix error in documentation which wrongly claims that the gradePeriod is in millis. In reality both the gradeperiod and the timeout is applied to the timeUnit specified as the third argument.
If one specifies `stop(1_000, 20, TimeUnit.SECONDS)`an `java.lang.IllegalArgumentException: timeout: 20 (expected >= quietPeriod (1000))` is thrown.


cy6erGn0m:
Thanx

PR #735, opened 2018-11-18
ApplicationConfig: Property or default

dmitvitalii:
Make `ApplicationConfig.property()` method have a default parameter: `default: ApplicationConfigValue? = null`.
For some cases when we don't want to handle either one extra exception or `null`.
Having a default property we will be able to have some object, which implements `ApplicationConfigValue` and use it, e.g., as a null-object.

Some other cases:
1. Extra safety for a config. For example, for a database: if we store user/pass in a Hocon configuration file, but we want to have user=root and password=root after deployment as a failsafe, we could use the default value:
```
class DatabaseRootCredentials : ApplicationConfigValue {
    override fun getList() = emptyList<String>()

    override fun getString() = "root"
}
```
This could be also useful for a deployment configurations, when we have default port, url, protocol etc.

2. Testing. If we somehow want to have tests, which still might be changed using configurations. We can provide some test object as a default if we have no test configuration.

If a developer needs the default value, this PR will allow one to use it. If a developer doesn't need it, one can still use `ApplicatoinConfig.property()` and `ApplicatoinConfig.propertyOrNull()` without specifying any defaults and the methods will have an old behaviour, throw an exception or return null respectively.

### Changes
- Add a `default` parameter to `ApplicationConfig.property*` methods, which is `null` by default;
- Implement in MapApplicationConfig and HoconApplicationConfig;
- Add tests for `MapApplicationConfig.property*` with default values.

### DIFF
--- a/ktor-server/ktor-server-core/src/io/ktor/config/ApplicationConfig.kt
+++ b/ktor-server/ktor-server-core/src/io/ktor/config/ApplicationConfig.kt
@@ -8,15 +8,15 @@ import io.ktor.util.*
 @KtorExperimentalAPI
 interface ApplicationConfig {
     /**
-     * Get config property with [path] or fail
+     * Get config property with [path], return [default] or fail if no [default] provided
      * @throws ApplicationConfigurationException
      */
-    fun property(path: String): ApplicationConfigValue
+    fun property(path: String, default: ApplicationConfigValue? = null): ApplicationConfigValue
 
     /**
-     * Get config property value for [path] or return `null`
+     * Get config property value for [path], return [default] or return `null`
      */
-    fun propertyOrNull(path: String): ApplicationConfigValue?
+    fun propertyOrNull(path: String, default: ApplicationConfigValue? = null): ApplicationConfigValue?
 
     /**
      * Get config child node or fail
--- a/ktor-server/ktor-server-core/src/io/ktor/config/ApplicationConfig.kt
+++ b/ktor-server/ktor-server-core/src/io/ktor/config/ApplicationConfig.kt
@@ -19,6 +19,11 @@ interface ApplicationConfig {
     fun propertyOrNull(path: String): ApplicationConfigValue?
 
     /**
+     * Get config property value for [path] or return [default]
+     */
+    fun propertyOrDefault(path: String, default: ApplicationConfigValue) = propertyOrNull(path) ?: default
+
+    /**
      * Get config child node or fail
      * @throws ApplicationConfigurationException
      */
--- a/ktor-server/ktor-server-core/src/io/ktor/config/ApplicationConfig.kt
+++ b/ktor-server/ktor-server-core/src/io/ktor/config/ApplicationConfig.kt
@@ -8,15 +8,15 @@ import io.ktor.util.*
 @KtorExperimentalAPI
 interface ApplicationConfig {
     /**
-     * Get config property with [path], return [default] or fail if no [default] provided
+     * Get config property with [path] or fail
      * @throws ApplicationConfigurationException
      */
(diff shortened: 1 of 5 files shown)


ktor-server/ktor-server-core/src/io/ktor/config/ApplicationConfig.kt
```diff
@@ -8,15 +8,15 @@ import io.ktor.util.*
...
      */
-    fun property(path: String): ApplicationConfigValue
+    fun property(path: String, default: ApplicationConfigValue? = null): ApplicationConfigValue
 
     /**
-     * Get config property value for [path] or return `null`
+     * Get config property value for [path], return [default] or return `null`
      */
-    fun propertyOrNull(path: String): ApplicationConfigValue?
+    fun propertyOrNull(path: String, default: ApplicationConfigValue? = null): ApplicationConfigValue?
```
cy6erGn0m: Then it should be named `propertyOrDefault`. However I'd keep `propertyOrNull` as well with the default implementation like this: `fun propertyOrNull(path: String) = propertyOrDefault(path, null)`

ktor-server/ktor-server-core/src/io/ktor/config/ApplicationConfig.kt
```diff
@@ -8,15 +8,15 @@ import io.ktor.util.*
 @KtorExperimentalAPI
 interface ApplicationConfig {
     /**
-     * Get config property with [path] or fail
+     * Get config property with [path], return [default] or fail if no [default] provided
      * @throws ApplicationConfigurationException
      */
-    fun property(path: String): ApplicationConfigValue
+    fun property(path: String, default: ApplicationConfigValue? = null): ApplicationConfigValue
```
cy6erGn0m: There is no need to change it since we already have `propertyOrNull/Default`

ktor-server/ktor-server-tests/test/io/ktor/tests/server/application/MockConfigValue.kt
```diff
@@ -0,0 +1,9 @@
+package io.ktor.tests.server.application
+
+import io.ktor.config.ApplicationConfigValue
+
+class MockConfigValue(private val stringValue: String, private val listValue: List<String>) : ApplicationConfigValue {
```
cy6erGn0m: I'd suggest a secondary constructor with varargs like this

```
constructor(stringValue: String, vararg listValues: String) : this(stringValue, listValues.toList())
```

ktor-server/ktor-server-core/src/io/ktor/config/ApplicationConfig.kt
dmitvitalii: One of the ideas was to provide a method, which returns a null-safe object.
Please, consider this commit: https://github.com/ktorio/ktor/pull/735/commits/a505f85a75182f60133d0b404618e25f6423c682.

cy6erGn0m: The main question is remaining: why one couldn't simply always use elvis instead?

dmitvitalii: The same reason Kotlin collections includes symmetric method, like `isEmpty()`/`isNotEmpty()`: convenience and ability to write a bit more clean code.
Using of `propertyOrNull()` with elvis operator to use some default looks like a tricky solution and complicates a code a bit.

However, if we could still consider the first commit: it would make more sense to have an existing methods `property()`/`propertyOrNull()`, which can be used "as is" or with a default value. There is a good reason to have separate methods with two different reactions: throw an expection, or return null/error code, but `propertyOrDefault()` seems too much. Defaults are convenient as optional parameters.

Issue #745, opened 2018-11-21
Conflict between ContentNegotiation and Websockets


laurencegw:
### Ktor Version 
"1.0.0"

### Ktor Engine Used(client or server and name)
ktor-server-netty

### JVM Version, Operating System and Relevant Context
openjdk version "1.8.0_181"
ubuntu

### Feedback

At the moment it seems that `ContentNegotiation` plugin is incompatible with the `Websockets` plugin.
Possibly a feature/config is needed to restrict the routes that content negotiation occurs on?

When both are installed, any calls to the websocket endpoint just hangs.

#### Code gist - slightly modified from the websocket testing example.

    import io.ktor.application.install
    import io.ktor.gson.gson
    import io.ktor.http.cio.websocket.Frame
    import io.ktor.http.cio.websocket.readText
    import io.ktor.routing.routing
    import io.ktor.server.testing.withTestApplication
    import io.ktor.websocket.WebSockets
    import io.ktor.websocket.webSocket
    import kotlinx.coroutines.channels.ClosedReceiveChannelException
    import org.junit.jupiter.api.Assertions.assertEquals
    import org.junit.jupiter.api.Test
    
    class MyAppTest {
        @Test
        fun testConversation() {
            withTestApplication {
                application.install(WebSockets)
                application.install(io.ktor.features.ContentNegotiation) {
                    gson {}
                }
    
                val received = arrayListOf<String>()
                application.routing {
                    webSocket("/echo") {
                        try {
                            while (true) {
                                val text = (incoming.receive() as Frame.Text).readText()
                                received += text
                                outgoing.send(Frame.Text(text))
                            }
                        } catch (e: ClosedReceiveChannelException) {
                            // Do nothing!
                        } catch (e: Throwable) {
                            e.printStackTrace()
                        }
                    }
                }
    
                handleWebSocketConversation("/echo") { incoming, outgoing ->
                    val textMessages = listOf("[]", "[]")
                    for (msg in textMessages) {
                        outgoing.send(Frame.Text(msg))
                        assertEquals(msg, (incoming.receive() as Frame.Text).readText())
                    }
                    assertEquals(textMessages, received)
                }
            }
        }
    }

#### Gradle

    import org.jetbrains.kotlin.gradle.tasks.KotlinCompile

    val kotlinVersion = "1.3.0"
    val kotlinXVersion = "1.0.0"
    val junitJupiterVersion = "5.2.0"
    val ktorVersion = "1.0.0"
[...]

PR #750, opened 2018-11-22
Prevent ContentNegotiation from handling byte channel pipeline (#745)

cy6erGn0m:
Fix for ContentNegotiation bug. The reason why it did fail is that ContentNegotation shouldn't negotiate already transformed values.


cy6erGn0m:
Merged

Issue #736, opened 2018-11-18
Support Session Cookie


doyaaaaaken:
### Ktor Engine Used(client or server and name) server

Hi, it seems we can't set cookie's duration. Is it right?
https://github.com/ktorio/ktor/blob/master/ktor-server/ktor-server-core/src/io/ktor/sessions/SessionTransportCookie.kt#L61

I want to set null on `Expires` attribute of `Set-Cookie` header, and implement session cookie.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie
>Expires=<date> Optional
>The maximum lifetime of the cookie as an HTTP-date timestamp. See Date for the detailed format. If not specified, the cookie will have the lifetime of a session cookie. 

Is there any way to implement this?

doyaaaaaken:
Sorry, my explanation was bad.

I mean I want to set null on cookie's duration.
Like below code.
```
install(Sessions) {
    cookie<MySessionDataclass>("MySessionToken", storage = MySessionStorage) {
        cookie.path = "/"
        cookie.duration = null // set null data
    }
}
```

This is because I want to get `session cookie` (This session is finished when the browser is shut down).

Bluexin:
Oh, I totally misunderstood, my bad.

cy6erGn0m:
Fixed in 1.0.1

PR #751, opened 2018-11-22
Make session cookie duration optional (#736)

cy6erGn0m:
Provide ability to not send `max-age` and `expires` in session cookies


cy6erGn0m:
Merged

Issue #753, opened 2018-11-22
Autoreload is broken


cy6erGn0m:
### Ktor Version
1.0.0

### Ktor Engine Used(client or server and name)
Netty (perhaps, any other)

After reload response is always empty in spite of that a handler and call pipeline are actually invoked

abangkis:
Hi. Just trying out ktor today. And faced with this problem. Is there a date where this fix is available for general public? Thanks

cy6erGn0m:
Planned release date is December 4th

abangkis:
Okay. Thank you very much for the information. Will be waiting for the release 😊

cy6erGn0m:
Fixed in 1.0.1

qihigh:
it works on 1.0.1

chrisjenx:
Still broken for me:

```
2018-12-16 10:02:22.779 [nettyCallPool-4-5] TRACE Application - Unhandled: GET - /assets/images/favicon.ico
2018-12-16 10:02:36.858 [nettyWorkerPool-3-5] INFO  Application - Changes in application detected.
2018-12-16 10:02:37.060 [nettyWorkerPool-3-5] DEBUG Application - Changes to 9 files caused application restart.
2018-12-16 10:02:37.060 [nettyWorkerPool-3-5] DEBUG Application - ...  MainKt$module$1$2$1.class
2018-12-16 10:02:37.060 [nettyWorkerPool-3-5] DEBUG Application - ...  MainKt$module$1$2.class
2018-12-16 10:02:37.060 [nettyWorkerPool-3-5] DEBUG Application - ...  MainKt$module$1$2$1$1.class
2018-12-16 10:02:37.060 [nettyWorkerPool-3-5] DEBUG Application - ...  MainKt.class
2018-12-16 10:02:37.060 [nettyWorkerPool-3-5] DEBUG Application - ...  MainKt$module$1.class
2018-12-16 10:02:37.060 [nettyWorkerPool-3-5] TRACE Application - Application stopping: io.ktor.application.Application@6a66a204
2018-12-16 10:02:37.062 [nettyWorkerPool-3-5] TRACE Application - Application stopped: io.ktor.application.Application@6a66a204
2018-12-16 10:02:37.062 [nettyWorkerPool-3-5] DEBUG Application - Java Home: /Library/Java/JavaVirtualMachines/jdk1.8.0_192.jdk/Contents/Home
[...]

PR #759, opened 2018-11-26
Fix broken autoreload feature (#753)

cy6erGn0m:
Introduce engine's receive and send pipeline. Merge them into the application send/receive pipelines at startup (including reloading).


cy6erGn0m:
Merged

PR #776, opened 2018-12-05
Clarify usage of HttpRequestBuilder.parameter() function

MariusVolkhart:


e5l:
Hi, @MariusVolkhart. Thanks for the notice :) 
Probably we could set form parameter in a URL but can't in a body. 
Could you clarify that in the docstring?

e5l:
LGTM

PR #777, opened 2018-12-05
`kotlin-multiplatform` + CURL engine

e5l:


ktor-client/ktor-client-curl/build.gradle
```diff
@@ -0,0 +1,35 @@
+kotlin {
+    targets {
+        List<org.jetbrains.kotlin.gradle.plugin.KotlinTarget> current = []
+
+        if (project.ext.ideaActive) {
+            current.add(posix)
+        } else {
+            current = [iosX64, iosArm64, macosX64]
+            if (project.ext.isLinuxHost) current = [linuxX64, mingwX64]
```
alexander-gorshenev: doesn't look right. We don't cross compile from linux to mingw.

PR #784, opened 2018-12-11
Introduce CoroutineStackFrame and implement it in SuspendFunctionGun …

qwwdfsad:
…reusable completion in order to properly walk through stackframes of interceptors

### DIFF
--- a/ktor-utils/ktor-utils-ios/src/io/ktor/util/StackFramesIos.kt
+++ b/ktor-utils/ktor-utils-ios/src/io/ktor/util/StackFramesIos.kt
@@ -1,0 +1,10 @@
+package io.ktor.util
+
+@Suppress("UNUSED")
+internal actual interface CoroutineStackFrame {
+    public actual val callerFrame: CoroutineStackFrame?
+    public actual fun getStackTraceElement(): StackTraceElement?
+}
+
+@Suppress("ACTUAL_WITHOUT_EXPECT")
+actual typealias StackTraceElement = Any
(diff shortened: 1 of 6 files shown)


ktor-utils/ktor-utils-jvm/src/io/ktor/util/StackFramesJvm.kt
```diff
@@ -0,0 +1,7 @@
+package io.ktor.util
+
+@Suppress("ACTUAL_WITHOUT_EXPECT")
+actual typealias CoroutineStackFrame = kotlin.coroutines.jvm.internal.CoroutineStackFrame
```
cy6erGn0m: internal ?

ktor-utils/ktor-utils-ios/src/io/ktor/util/StackFramesIos.kt
```diff
@@ -0,0 +1,10 @@
+package io.ktor.util
+
+@Suppress("UNUSED")
+internal actual interface CoroutineStackFrame {
+    public actual val callerFrame: CoroutineStackFrame?
+    public actual fun getStackTraceElement(): StackTraceElement?
+}
+
+@Suppress("ACTUAL_WITHOUT_EXPECT")
+actual typealias StackTraceElement = Any
```
cy6erGn0m: Should it be public?

ktor-utils/src/io/ktor/util/pipeline/PipelineContext.kt
```diff
@@ -97,13 +102,16 @@ private class SuspendFunctionGun<TSubject : Any, TContext : Any>(
...
+    // Continuation where 'proceed' was called in order to properly walk over stacktrace
+    private var proceedContinuation: Continuation<*>? = null
     private var index = 0
 
     override fun finish() {
         index = blocks.size
     }
 
     override suspend fun proceed(): TSubject = suspendCoroutineUninterceptedOrReturn { continuation ->
+        proceedContinuation = continuation
```
cy6erGn0m: proceed()` could be invoked multiple times: this is why continuations are stacked in this case. Shouldn't `proceedContinuation` be restored back when leaving a proceed level?

PR #785, opened 2018-12-11
okhttp-client: pass contentLength for multipart form data to StreamRe…

jonasbark:
…questBody

The existing implementation caused okhttp to never send the correct `Content-Length` and instead sending `Transfer-Encoding: chunked`, even though one might have set the `contentLength` in a `OutgoingContent.WriteChannelContent` subclass.

Note: the MultiPartFormDataContent from ktor should also support calculating contentLength


e5l:
Hi, @jonasbark. Thanks for the request. LGTM

PR #786, opened 2018-12-11
Cy/windows websocket

cy6erGn0m:
Fix network errors on Windows: the main idea is to not close actual socket at server side as it usually should do a client (to avoid TIME_WAIT). In fact we probably need to do this for any requests but for now we can fix for websockets only as it caused errors

### DIFF
--- a/ktor-server/ktor-server-test-host/src/io/ktor/server/testing/EngineTestSuite.kt
+++ b/ktor-server/ktor-server-test-host/src/io/ktor/server/testing/EngineTestSuite.kt
@@ -1361,6 +1361,8 @@ abstract class EngineTestSuite<TEngine : ApplicationEngine, TConfiguration : App
 
     @Test
     open fun testUpgrade() {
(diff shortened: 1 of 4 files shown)


ktor-server/ktor-server-test-host/src/io/ktor/server/testing/EngineTestSuite.kt
```diff
@@ -1361,6 +1361,8 @@ abstract class EngineTestSuite<TEngine : ApplicationEngine, TConfiguration : App
 
     @Test
     open fun testUpgrade() {
+        var completed = CompletableDeferred<Unit>()
```
e5l: var?

cy6erGn0m: Nice catch :)

cy6erGn0m:
Merged

PR #794, opened 2018-12-14
Switch default algorithm to SHA-256 from SHA-1 for signed cookies

veeti:


cy6erGn0m:
Unlike usual hashes, with HMAC it doesn't make notable difference to switch from SHA1 to SHA256 except that HmacSHA256 is slower. Is there any reason to do so here?

PR #805, opened 2018-12-18
Add UrlFetch Engine for Requests from Google App Engine environment

Meosit:
UrlFetch Engine created on top of standard java.net.URL class which allow to use Ktor with [Google App Engine](https://cloud.google.com/appengine/) on Java 8 Runtime since it require to [issue HTTP requests by using `java.net` package classes](https://cloud.google.com/appengine/docs/standard/java/issue-requests) in case of disabled billing.


e5l:
Hi, @Meosit. Thanks for the PR :)

I'll take a look on days.

Meosit:
Hi, @cy6erGn0m. Didn't noticed that Android client already built using `java.net` classes.

So I recon it will be good enough to generalize the existing engine. Does it should be renamed?

Thanks!

e5l:
We could make an alias.

cy6erGn0m:
The main disadvantage of HttpUrlConnection is that it is synchronous and blocking while `URLFetchService` provides async API (based on `j.u.c.Future`). Not sure if it is truly asynchronous (need to be investigated) and if there are any advantages of using it.

Meosit:
Hi @e5l, @cy6erGn0m,

Sorry for the late reply. After investigation on the subject of this ticket I came up with the following results:
1. GAE applications (as well as [URLFetchService](https://cloud.google.com/appengine/docs/standard/java/javadoc/com/google/appengine/api/urlfetch/URLFetchService)) require [ThreadManager](https://cloud.google.com/appengine/docs/standard/java/javadoc/com/google/appengine/api/ThreadManager) class for creating new threads.
1. In order to fit with the existing Engine-based structure, the following construction can be used for corutineDispatcher (also since the current implementation do not allow to configure custom dispatcher for existing engines (deprecated), only new ones can be created):  
    ```kotlin
    Executors.newFixedThreadPool(threadCount, ThreadManager.backgroundThreadFactory())
            .asCoroutineDispatcher()
    ```
    But this working only with [manual scaling](https://cloud.google.com/appengine/docs/standard/java/config/appref#scaling_elements_manual_scaling), which is not a best option for free GAE applications due to 9 free instance hours per day limitation.
1. `AndroidClientEngine` also affected by the threads limitation and cannot be used.

Taking into account the points above there is no generic solution can be provided other than simple blocking one. So closing the pull request.

e5l:
Probably you could override the dispatcher with a custom static instance.

Issue #808, opened 2018-12-19
HTTPClient Logging feature depends on an SLF4J implementation


EdwardDrapkin:
### Ktor Version
1.0.1

### Ktor Engine Used(client or server and name)
JVM HttpClient

### Feedback
ktor-client-logging-jvm has a dependency on slf4j-simple instead of slf4j-api:

```

org.slf4j:slf4j-simple:1.6.1
\--- io.ktor:ktor-client-logging-jvm:1.0.1
     \--- compileClasspath

```

When you have a dependency on another adapter, e.g. logback, you get the following warning:

```
SLF4J: Class path contains multiple SLF4J bindings.
SLF4J: Found binding in [jar:file:/Users/edwarddrapkin/.gradle/caches/modules-2/files-2.1/ch.qos.logback/logback-classic/1.2.1/706a8b8206ead3683ec639dd270d11fd948fbb0e/logback-classic-1.2.1.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Found binding in [jar:file:/Users/edwarddrapkin/.gradle/caches/modules-2/files-2.1/org.slf4j/slf4j-simple/1.6.1/58e59bfb3e247097b8122243b3bfe0049c8cfae8/slf4j-simple-1.6.1.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: See http://www.slf4j.org/codes.html#multiple_bindings for an explanation.
SLF4J: Actual binding is of type [ch.qos.logback.classic.util.ContextSelectorStaticBinder]
```

Excluding it from the dependency resolution in gradle fixes the issue:

```
compile "io.ktor:ktor-client-logging-jvm:$ktor_version", {
        exclude group: "org.slf4j", module: "slf4j-simple"
    }
```

e5l:
Hi, @EdwardDrapkin. Thanks for the report.

PR #809, opened 2018-12-19
Update slf4j dependency in JVM HTTP Client Logging Feature

EdwardDrapkin:
Simple change to swap the slf4j implementation with the API.  I used the same slf4j-api version from jvm.gradle.


EdwardDrapkin:
This should fix #808

e5l:
LGTM. Merged

Issue #429, opened 2018-06-04
correctly encoded query string has been altered after decode and re-encode process


zdevwu:
https://github.com/ktorio/ktor/blob/7092c30706f34242419a5f0c81445dca06461dc3/ktor-http/src/io/ktor/http/URLUtils.kt#L28

should the query string from URI be considered encoded already and skip the decode and re-encode process?

cy6erGn0m:
Is it performance concern or there is something that doesn't work as expected?

zdevwu:
I've got a pre-signed url from aws s3, the query parameters are altered after the decode and re-encode

oleg-larshin:
Please check the <a href="https://youtrack.jetbrains.com/issue/KTOR-778">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

PR #819, opened 2018-12-23
http-client UrlBuilder adds = to query parameters without value

### LINKED ISSUE #818: http-client UrlBuilder adds = to query parameters without value
Simple test to show issue:
```
    @Test
    fun queryParamsWithNoValue() {
        val url = URLBuilder("https://httpstat.us/?novalue")
        assertEquals("https://httpstat.us/?novalue", url.buildString())
        //Fails -> actual: https://httpstat.us/?novalue=
    }
```

sjurba:
I noticed that you could add a key with an empty list to the `StringValues` class. I used this to convey that the key has no value and should not append the =. This means you can still get the = by setting the value to the empty string. An alternative solution would be to allow null values, and use that to convey no value. But that would require a major refactoring of the `StringValues` class and its  dependencies.

Fixes #818


sjurba:
A third option would be to make it possible to create URLs without re-encoding the query parameters as suggested in #429

e5l:
Hi @sjurba, thanks for the PR :)
Looks nice, we could merge it in `1.1.2`.

PR #823, opened 2018-12-24
Reduce JDK7/8 dependencies

cy6erGn0m:
- several modules are compiled with JDK8, remaining with 6
- rewritten several parts from NIO to regular java.io.File
- several top-level extension functions were moved to a separate files so `NoClassDefFound` will only fail when a particular function is touched
- socket builders were rewritten to not rely on JDK
- socket implementation fixed to prefer JDK's socket API over channels for setting options to work on JDK6


cy6erGn0m:
Merged

Issue #800, opened 2018-12-15
[BUG] HttpClient(CIOEngine)


revintec:
1) aSocket(...) may throw IOExceptions and causing semaphore to leak, related code:
https://github.com/ktorio/ktor/blob/f1a063a3cdcb4789534b9898058fd00fad0aab3f/ktor-client/ktor-client-cio/src/io/ktor/client/engine/cio/Endpoint.kt#L175
https://github.com/ktorio/ktor/blob/f1a063a3cdcb4789534b9898058fd00fad0aab3f/ktor-client/ktor-client-cio/src/io/ktor/client/engine/cio/ConnectionFactory.kt#L14
2) `retryAttempts` should be named `connectionAttempts` or we should be using `repeat(retryAttempts+1)`
https://github.com/ktorio/ktor/blob/f1a063a3cdcb4789534b9898058fd00fad0aab3f/ktor-client/ktor-client-cio/src/io/ktor/client/engine/cio/Endpoint.kt#L174

revintec:
any update on this?

revintec:
Thanks, I'll try it when the new version is released

Issue #806, opened 2018-12-19
ktor-client with CIO engine stops perfoming requests


ruXlab:
Hello here

It seems that ktor-client with CIO module doesn't handle concurrent requests nicely. In my handcrafted test ktor client loop simply stops working

I wrote a simple manual test to reproduce this situation:
```kotlin

import io.ktor.client.HttpClient
import io.ktor.client.engine.cio.CIO
import io.ktor.client.request.head
import kotlinx.coroutines.delay
import kotlinx.coroutines.isActive
import kotlinx.coroutines.launch
import kotlinx.coroutines.runBlocking
import java.util.concurrent.atomic.AtomicInteger

fun main(args: Array<String>) {

    val client = HttpClient(CIO) {
        engine {
            maxConnectionsCount = 1000000
            endpoint.maxConnectionsPerRoute = 1000000
            endpoint.connectRetryAttempts = 1
            endpoint.pipelineMaxSize = 1000000
        }
    }

    val reqCounter = AtomicInteger(0)
    var oldCounter = 0
    runBlocking {
        launch {
            0.rangeTo(10000).map { i ->
                delay(1000L)
                val current = reqCounter.get()
                val diff = current - oldCounter
                oldCounter = current
                println("#$i. Speed: ${diff}req/s, total requests: $current")
            }
        }
        for(i in 0..100) {
            launch {
                while(isActive) {
                    try {
                        client.head<String>("https://facebook.com")
                    } catch (e: Exception) {
                        e.printStackTrace()
                    }
                    reqCounter.incrementAndGet()
                }
            }
        }
    }
}
```
Sample output
```bash
#0. Speed: 0req/s, total requests: 0
#1. Speed: 0req/s, total requests: 0
#2. Speed: 21req/s, total requests: 21
#3. Speed: 70req/s, total requests: 91
#4. Speed: 27req/s, total requests: 118
#5. Speed: 84req/s, total requests: 202
#6. Speed: 96req/s, total requests: 298
#7. Speed: 77req/s, total requests: 375
#8. Speed: 93req/s, total requests: 468
#9. Speed: 39req/s, total requests: 507
#10. Speed: 0req/s, total requests: 507
. . . . . 
#362. Speed: 0req/s, total requests: 507
#404. Speed: 0req/s, total requests: 507
```

**Env info:**
Ktor Version: 1.0.1
Ktor Engine Used: CIO
JVM Version: 1.8.0_191-b12
Operating System: Ubuntu 18.04

[...]

PR #838, opened 2018-12-28
Fix leaking connections count and semaphore (#800)

cy6erGn0m:
In case of error opening a socket the semaphore inside of `connectionFactory` is not decremented


e5l:
Btw, could you add the test?

cy6erGn0m:
It also could be related to #806

PR #849, opened 2019-01-04
Client: Support relative redirects by making #takeFrom() resolve relative urls.

hpoul:
Currently the HttpRedirect feature of the client cannot handle relative redirects.
I discovered this when requesting a Wikipedia URL like `https://en.wikipedia.org/api/rest_v1/page/summary/Amazon_Inc` which returns: `location: Amazon_(company)`

I created a simple fix directly in `URLBuilder.takeFrom`, which imo makes sense that it also supports resolving relative URLs? (It simply makes sure the old path is not overwritten by the new one, if the new one is relative - but as before it won't normalize paths like `../newpath`)

### DIFF
--- a/ktor-client/ktor-client-tests/jvm/src/io/ktor/client/tests/HttpRedirectTest.kt
+++ b/ktor-client/ktor-client-tests/jvm/src/io/ktor/client/tests/HttpRedirectTest.kt
@@ -39,6 +39,12 @@ abstract class HttpRedirectTest(private val factory: HttpClientEngineFactory<*>)
 
                 call.respondText("OK")
             }
+            get("/directory/redirectFile") {
+                call.respondRedirect("targetFile")
+            }
+            get("/directory/targetFile") {
+                call.respondText("targetFile")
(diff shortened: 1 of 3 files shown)


cy6erGn0m:
Could you please add test for absolute path redirect (redirect path staring with slash character)?

hpoul:
@cy6erGn0m sure, see https://github.com/ktorio/ktor/pull/849/commits/c7d5d6334ded0c9c7497f882ce7922da64437438 - i have also added a test for host absolute redirects.. which made me discover another potential problem: a absolute redirect does not overwrite a customized port, which wouldn't make sense for the new host. So I have also changed that.

hpoul:
@cy6erGn0m nope, because the port is only replaced at the same time the host is replaced. ie. if the host is not replaced the port is not as well.. all tests actually use a custom port, that is why i discovered the problem when redirecting to `https://httpstat.us/200`

cy6erGn0m:
Ok, LGTM

PR #857, opened 2019-01-08
On jvm, break up long logs that would be truncated by android's max length of 4068 characters

luca992:
Long logs on android are currently truncated at [4068](https://stackoverflow.com/a/8899735/1363742) bytes 

Recursive log function adapted from [here](https://stackoverflow.com/a/48284047/1363742)


cy6erGn0m:
Shouldn't it be done by configuring logger itself instead of applying this to ktor? Doesn't logback/log4j provide a workaround?

luca992:
@cy6erGn0m I am not seeing any workarounds provided by log4j.

okhttp splits up their logs in a similar way found [here](https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/internal/platform/AndroidPlatform.java) 


    @Override public void log(int level, String message, @Nullable Throwable t) {
        int logLevel = level == WARN ? Log.WARN : Log.DEBUG;
        if (t != null) message = message + '\n' + Log.getStackTraceString(t);

        // Split by line, then ensure each line can fit into Log's maximum length.
        for (int i = 0, length = message.length(); i < length; i++) {
           int newline = message.indexOf('\n', i);
          newline = newline != -1 ? newline : length;
          do {
            int end = Math.min(newline, i + MAX_LOG_LENGTH);
            Log.println(logLevel, "OkHttp", message.substring(i, end));
            i = end;
          } while (i < newline);
        }
    }


How about if I left the default logger on jvm as is and just added a Logger for android to LoggerJvm.kt

Like so:

    package io.ktor.client.features.logging
    
    import io.ktor.client.*
    import org.slf4j.*
    
    actual val Logger.Companion.DEFAULT: Logger
        get() = object : Logger {
            private val delegate = LoggerFactory.getLogger(HttpClient::class.java)!!
            override fun log(message: String) {
                delegate.info(message)
            }
        }
    
    
    /***
     * Workaround Logger to log long messages that would otherwise be truncated in Android.
     */
    val Logger.Companion.ANDROID: Logger
        get() = MaxLengthLogger()
    
    
    class MaxLengthLogger(val MAX_INDEX : Int = 4000) : Logger {
        private val delegate = LoggerFactory.getLogger(HttpClient::class.java)!!
        override fun log(message: String) {
            logLong(message)
        }
    
        fun logLong(message: String) {
    
            val MIN_INDEX = 3000
    
            // String to be logged is longer than the max...
            if (message.length > MAX_INDEX) {
                var theSubstring = message.substring(0, MAX_INDEX)
                var theIndex = MAX_INDEX
    
                // Try to find a substring break at a line end.
                theIndex = theSubstring.lastIndexOf('\n')
                if (theIndex >= MIN_INDEX) {
                    theSubstring = theSubstring.substring(0, theIndex)
                } else {
                    theIndex = MAX_INDEX
                }
    
                // Log the substring.
                delegate.info(theSubstring)
    
                // Recursively log the remainder.
                logLong(message.substring(theIndex))
            } else {
                delegate.info(message)
            }// String to be logged is shorter than the max...
        }
    }

ktor-client/ktor-client-features/ktor-client-logging/jvm/src/io/ktor/client/features/logging/LoggerJvm.kt
```diff
@@ -10,3 +10,45 @@ actual val Logger.Companion.DEFAULT: Logger
...
+    get() = MaxLengthLogger()
+
+
+class MaxLengthLogger(val MAX_INDEX : Int = 4000) : Logger {
+    private val delegate = LoggerFactory.getLogger(HttpClient::class.java)!!
+    override fun log(message: String) {
+        logLong(message)
+    }
+
+    fun logLong(message: String) {
```
cy6erGn0m: `private tailrec` looks reasonable here

ktor-client/ktor-client-features/ktor-client-logging/jvm/src/io/ktor/client/features/logging/LoggerJvm.kt
```diff
@@ -10,3 +10,45 @@ actual val Logger.Companion.DEFAULT: Logger
...
+
+        val MIN_INDEX = 3000
+
+        // String to be logged is longer than the max...
+        if (message.length > MAX_INDEX) {
+            var theSubstring = message.substring(0, MAX_INDEX)
```
cy6erGn0m: `the` suffix makes the code more difficult to read. It would be better to rename.

ktor-client/ktor-client-features/ktor-client-logging/jvm/src/io/ktor/client/features/logging/LoggerJvm.kt
```diff
@@ -10,3 +10,45 @@ actual val Logger.Companion.DEFAULT: Logger
...
+                theSubstring = theSubstring.substring(0, theIndex)
+            } else {
+                theIndex = MAX_INDEX
+            }
+
+            // Log the substring.
+            delegate.info(theSubstring)
+
+            // Recursively log the remainder.
+            logLong(message.substring(theIndex))
```
cy6erGn0m: Shouldn't we discard leading newline character in the first case when it is found?

ktor-client/ktor-client-features/ktor-client-logging/jvm/src/io/ktor/client/features/logging/LoggerJvm.kt
```diff
@@ -10,3 +10,45 @@ actual val Logger.Companion.DEFAULT: Logger
...
+
+
+/***
+ * Workaround Logger to log long messages that would otherwise be truncated in Android.
+ */
+val Logger.Companion.ANDROID: Logger
+    get() = MaxLengthLogger()
+
+
+class MaxLengthLogger(val MAX_INDEX : Int = 4000) : Logger {
```
cy6erGn0m: It should be either documented or internal/private instead.

ktor-client/ktor-client-features/ktor-client-logging/jvm/src/io/ktor/client/features/logging/LoggerJvm.kt
```diff
@@ -10,3 +10,45 @@ actual val Logger.Companion.DEFAULT: Logger
...
+    private val delegate = LoggerFactory.getLogger(HttpClient::class.java)!!
+    override fun log(message: String) {
+        logLong(message)
+    }
+
+    fun logLong(message: String) {
+
+        val MIN_INDEX = 3000
```
cy6erGn0m: This could be configurable just like `MAX_INDEX`

ktor-client/ktor-client-features/ktor-client-logging/jvm/src/io/ktor/client/features/logging/LoggerJvm.kt
luca992: I documented it

ktor-client/ktor-client-features/ktor-client-logging/jvm/src/io/ktor/client/features/logging/LoggerJvm.kt
luca992: Sounds like a good idea

ktor-client/ktor-client-features/ktor-client-logging/jvm/src/io/ktor/client/features/logging/LoggerJvm.kt
luca992: yeah, didn't catch that

cy6erGn0m:
Merged manually, renamed to `MessageLengthLimitingLogger`

PR #870, opened 2019-01-14
Introduce standard ktor exceptions

cy6erGn0m:
- Introduce standard ktor exceptions:
  - `BadRequestException`  (leads to `400 Bad Request`)
    - `MissingRequestParameterException`
    - `ParameterConversionException`
  - `NotFoundException` (leads to `404 Not Found`)
  - handle `TimeoutException` and `TimeoutCancellationException` (leads to `504 Gateway Timeout`)
  - `UnsupportedMediaTypeException` is handled without `ContentNegotiation` feature
- Locations feature supports these exceptions out of the box
- Introduce initial parameters delegation support, allows the following

```kotlin
get("/") {
    val name: String by call.request.queryParameters
    val page: Int by call.request.queryParameters
}
```

Open questions:
- There are `ParameterConversionException` (when request parameter conversion failed, e.g. `toInt()`) and `ContentTransformationException` (when request body fails) but the first one inherits `BadRequestException` while the second doesn't. Should it? Note that `UnsupportedMediaTypeException` does inherit `ContentTransformationException`.
- Should we handle timeout exceptions? `NotImplementedError`? (such exception is thrown by `kotlin.TODO()`
- Parameters delegation issues:
  - `DefaultConversionService` is always used, there is no way to capture application's service
  - operator function `getValue` doesn't support overloading so the leads to:
    - local variable type specification is required
    - relatively slow due to KClass reflection magic 
- test engine usually by-passed all exceptions from request handlers to test level however not these new kotlin exceptions are handled so they are not by-passed but converted to the corresponding status codes

### DIFF
--- a/ktor-server/ktor-server-core/jvm/src/io/ktor/features/ContentNegotiation.kt
+++ b/ktor-server/ktor-server-core/jvm/src/io/ktor/features/ContentNegotiation.kt
@@ -48,7 +48,7 @@ class ContentNegotiation(val registrations: List<ConverterRegistration>) {
      * Implementation of an [ApplicationFeature] for the [ContentNegotiation]
      */
     companion object Feature : ApplicationFeature<ApplicationCallPipeline, Configuration, ContentNegotiation> {
-        override val key = AttributeKey<ContentNegotiation>("ContentNegotiation")
+        override val key: AttributeKey<ContentNegotiation> = AttributeKey("ContentNegotiation")
 
         override fun install(pipeline: ApplicationCallPipeline, configure: Configuration.() -> Unit): ContentNegotiation {
             val configuration = Configuration().apply(configure)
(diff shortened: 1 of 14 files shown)


cy6erGn0m:
```kotlin
val x: String by call.request.queryParameters.required()
val y: String? by call.request.queryParameters.optional()
```

where `required()` and `optional()` are extension functions inside of `PipelineContext`. It works but there is the serious disadvantage because it is not extendable and even worse it affects ABI since they are interface member functions.

ktor-server/ktor-server-core/jvm/src/io/ktor/features/ContentNegotiation.kt
```diff
@@ -105,12 +105,6 @@ class ContentNegotiation(val registrations: List<ConverterRegistration>) {
     }
 }
 
-/**
- * Thrown when there is no conversion for a content type configured
- */
-class UnsupportedMediaTypeException(contentType: ContentType) :
```
e5l: Deprecation

cy6erGn0m:
Merged. Resolution:
- everything experimental
- delegation is limited to `DefaultConversionService`
- no extra exceptions handled
- request body conversion -> 500 Internal Server Error, parameters conversion -> 400 Bad Request

Issue #876, opened 2019-01-15
Invalid routing on multiple tailcards


vonox7:
A request to `http://localhost:8090/asdf` should not be handled by any route but is in Ktor 1.1.1 handled by the "ADMIN" route.
```
embeddedServer(Netty, port = 8090) {
      // http://localhost:8090/admin-prefix-hello --> "ADMIN" (OK)
      // http://localhost:8090/subdirectory/hello --> "Wildcard 404" (OK)
      // http://localhost:8090/asdf --> "ADMIN" (WHAT? Why did it take that route?)
      install(Routing) {
        route("admin-prefix-{...}", HttpMethod.Get) {
          handle {
            call.respond("ADMIN")
          }
        }
        route("subdirectory") {
          route("{...}", HttpMethod.Get) {
            handle {
              call.respond("Wildcard 404")
            }
          }
        }
      }
    }.start()
```

orangy:
Thanks for a report, we'll investigate. 

PS: As a side question, is there a reason you don't use `get("…") { }` instead of route/handle?

vonox7:
Because I copy pasted the example from various sources, tried to find the minimum prototype to reproduce this error ;)

cy6erGn0m:
Fixed

vonox7:
That was prompt! Thanks!

Issue #526, opened 2018-08-17
Tailcard parameter matcher capturest rest of its segment 


NekoiNemo:
Example:  
Using `get("explore{path...}")` handler, url ending with `/explore/log/server.log` will result in `call.parameters.getAll("path")` returning list containing: "explore, log, server.log", rather than just last 2

Using `get("explore/{path...}")`, however, returns only last 2 as expected

See https://kotlinlang.slack.com/archives/C0A974TJ9/p1534363258000100

cy6erGn0m:
Why 2 components are expected? I'd expect it to be `"explore", "log", "server.log"` or `"", "log", "server.log"`. For sure `"explore{path...}"` and `"explore/{path...}"` are not the same patterns

NekoiNemo:
Well, i guess 3 starting with "" would be most logical one. Then again, empty segments are a bit pointless. But in any case, i don't think "explore", located outside of part captured by parameter, should be included

cy6erGn0m:
Fixed ignoring prefix

PR #879, opened 2019-01-16
Add tailcard prefix support (#876, #526)

cy6erGn0m:
Currently the following tailcard `prefix-{param...}` route does actually handle all requests in spite of `prefix-`. This commit introduce prefix handling. Also it now cuts prefix from the first matched segment similar to `PathSegmentOptionalParameterRouteSelector` when capturing parameter values:
- `prefix-value` -> `param = ['value']`
- `prefix-a/b/c` -> `param = ['a', 'b', 'c']`


cy6erGn0m:
Merged.

Issue #872, opened 2019-01-15
ContentNegotiation register content type "text/*" can not match "text/plain"


Ccixyj:
### Ktor Version
1.1.1

### Ktor Engine Used(client or server and name)
server netty

### JVM Version, Operating System and Relevant Context
windows

### Feedback

```kotlin
install(ContentNegotiation) {
        gson(block = buildBlock)
        register(contentType = ContentType.Text.Any,converter =  GsonConverter(GSON))
    }
```
for me , this code looks like all subtype  will match,but failed.

https://github.com/ktorio/ktor/blob/324872f05c4934747bc9a226bdb8a50d9a10da26/ktor-server/ktor-server-core/jvm/src/io/ktor/features/ContentNegotiation.kt#L97

`it.contentType.match(contentType)`  looks like ` "text/*".match("text/plain") ` reutrn false

cy6erGn0m:
The following works:

```kotlin
    @Test
    fun testGsonOnTextAny(): Unit = withTestApplication {
        data class C(val x: Int)

        application.install(ContentNegotiation) {
            gson()
            register(contentType = ContentType.Text.Any, converter = GsonConverter())
        }

        application.routing {
            post("/") {
                val instance = call.receive<C>()
                assertEquals(C(777), instance)
                call.respondText("OK")
            }
        }

        handleRequest(HttpMethod.Post, "/") {
            addHeader(HttpHeaders.ContentType, "text/plain")
            setBody("{\"x\": 777}")
        }
        handleRequest(HttpMethod.Post, "/") {
            addHeader(HttpHeaders.ContentType, "application/json")
            setBody("{\"x\": 777}")
        }
    }
```

cy6erGn0m:
Fixed

Issue #811, opened 2018-12-20
Got "Content type application/json is not supported" exception when the class to serialize/unserialize is in the function


scorsi:
Hello,

I don't know if it is an expected result but here is something that didn't work and how to solve it.
I spent a lot of time to fix that.
Did the same with `ktor-gson` or `ktor-jackson`.

### Ktor Version
`1.0.1`
### Ktor Engine Used(client or server and name)
`ktor-server-netty`
### JVM Version, Operating System and Relevant Context
`jdk1.8.0_17` on `macOS Mojave 10.14.1`
### Feedback
I'm getting
```
io.ktor.features.UnsupportedMediaTypeException: Content type application/json is not supported
	at io.ktor.features.ContentNegotiation$Feature$install$3.invokeSuspend(ContentNegotiation.kt:100)
	at io.ktor.features.ContentNegotiation$Feature$install$3.invoke(ContentNegotiation.kt)
	at io.ktor.util.pipeline.SuspendFunctionGun.loop(PipelineContext.kt:248)
	at io.ktor.util.pipeline.SuspendFunctionGun.access$loop(PipelineContext.kt:63)
	at io.ktor.util.pipeline.SuspendFunctionGun.proceed(PipelineContext.kt:111)
	at io.ktor.util.pipeline.SuspendFunctionGun.execute(PipelineContext.kt:131)
	at io.ktor.util.pipeline.Pipeline.execute(Pipeline.kt:24)
	at io.ktor.request.ApplicationReceiveFunctionsKt.receive(ApplicationReceiveFunctions.kt:68)
	at something.SomethingKt.notes(something.kt:176)
	at something.SomethingKt$api$1.invokeSuspend(something.kt:8)
	at something.SomethingKt$api$1.invoke(something.kt)
	at io.ktor.util.pipeline.SuspendFunctionGun.loop(PipelineContext.kt:248)
	at io.ktor.util.pipeline.SuspendFunctionGun.access$loop(PipelineContext.kt:63)
	at io.ktor.util.pipeline.SuspendFunctionGun.proceed(PipelineContext.kt:111)
	at io.ktor.util.pipeline.SuspendFunctionGun.execute(PipelineContext.kt:131)
	at io.ktor.util.pipeline.Pipeline.execute(Pipeline.kt:24)
	at io.ktor.routing.Routing.executeResult(Routing.kt:148)
	at io.ktor.routing.Routing.interceptor(Routing.kt:29)
[...]

PR #880, opened 2019-01-16
GSON and content negotiation bugfiexes

cy6erGn0m:
- Fix for #872 (Registering content converter with custom content type pattern)
- Fix for #811 (Deserializing null values and excluded types)


cy6erGn0m:
Merged.

Issue #889, opened 2019-01-20
Conflict between WebSockets and StatusPages


neworld:
Following test just hangs. This issue could be similar to #745.
```kotlin
    @Test
    fun testConversation() {
        withTestApplication {
            application.install(WebSockets)
            application.install(StatusPages) // <-- I added only this

            val received = arrayListOf<String>()
            application.routing {
                webSocket("/echo") {
                    try {
                        while (true) {
                            val text = (incoming.receive() as Frame.Text).readText()
                            received += text
                            outgoing.send(Frame.Text(text))
                        }
                    } catch (e: ClosedReceiveChannelException) {
                        // Do nothing!
                    } catch (e: Throwable) {
                        e.printStackTrace()
                    }
                }
            }

            handleWebSocketConversation("/echo") { incoming, outgoing ->
                val textMessages = listOf("HELLO", "WORLD")
                for (msg in textMessages) {
                    outgoing.send(Frame.Text(msg))
                    assertEquals(msg, (incoming.receive() as Frame.Text).readText())
                }
                assertEquals(textMessages, received)
            }
        }
    }
```

### Versions
- Ktor: 1.1.1
- Java: openJDK 11.0.1
- Kotlin: 1.3.11
- Coroutines: 1.1.0

### Ktor Engine Used(client or server and name)

Test engine.

### Thread dump

```
"Test worker" #13 prio=5 os_prio=0 cpu=355.78ms elapsed=20.86s tid=0x00007f9344b56000 nid=0x5c4c waiting on condition  [0x00007f92ff095000]
   java.lang.Thread.State: WAITING (parking)
        at jdk.internal.misc.Unsafe.park(java.base@11.0.1/Native Method)
        - parking to wait for  <0x000000062dd27610> (a java.util.concurrent.CompletableFuture$Signaller)
        at java.util.concurrent.locks.LockSupport.park(java.base@11.0.1/LockSupport.java:194)
        at java.util.concurrent.CompletableFuture$Signaller.block(java.base@11.0.1/CompletableFuture.java:1796)
        at java.util.concurrent.ForkJoinPool.managedBlock(java.base@11.0.1/ForkJoinPool.java:3128)
        at java.util.concurrent.CompletableFuture.waitingGet(java.base@11.0.1/CompletableFuture.java:1823)
        at java.util.concurrent.CompletableFuture.join(java.base@11.0.1/CompletableFuture.java:2043)
[...]

PR #890, opened 2019-01-21
Fix web socket conversation

cy6erGn0m:
The old `handleWebSocketConversation` implementation was relying on the fact that the call pipeline is entirely executed and a client's conversation block was launched only after pipeline completion. However we actually can't rely on this behavior: `StatusPages` is a typical case (#889) : call pipeline is not completed because or async child jobs so it causes suspending _deadlock_ since client conversation block is not started at all if case when the server conversation block is waiting for client's frames.


cy6erGn0m:
Merged manually.

PR #910, opened 2019-01-31
Add optional contentType to formDsl

### LINKED ISSUE #766: FormBuilder lacks API constructing file upload request with custom body header
### Ktor Version

1.0.0

OkHttp

### JVM Version, Operating System and Relevant Context

Android 28

### Feedback

With ktor, it seems that we can't post `Multipart/form-data` that has both `Content-Type` and `Content-Disposition` with `FormBuilder`.

Currently the following APIs are available:

```kotlin
fun FormBuilder.append(key: String, headers: Headers = Headers.Empty, bodyBuilder: BytePacketBuilder.() -> Unit)
fun FormBuilder.append(key: String, filename: String, bodyBuilder: BytePacketBuilder.() -> Unit) 
```

But we can't use filename and headers together to form body likes(copied from https://stackoverflow.com/a/28380690/3562486):

```plaintext
POST / HTTP/1.1
[[ Less interesting headers ... ]]
Content-Type: multipart/form-data; boundary=---------------------------735323031399963166993862150
Content-Length: 834
-----------------------------735323031399963166993862150
Content-Disposition: form-data; name="file1"; filename="a.txt"
Content-Type: text/plain

Content of a.txt.

-----------------------------735323031399963166993862150
Content-Disposition: form-data; name="file2"; filename="a.html"
Content-Type: text/html

<!DOCTYPE html><title>Content of a.html.</title>
```

Inside the body both `Content-Type` header and `filename` are used.

wollnyst:
Fixes #766 

IMO it's enough to add `ContentType`, that covers like 99% of upload scenarios now. For all other scenarios we can use `append(key, headers,, bodyBuilder)`

### DIFF
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/request/forms/formDsl.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/request/forms/formDsl.kt
@@ -80,15 +80,18 @@ inline fun FormBuilder.append(key: String, headers: Headers = Headers.Empty, bod
 }
 
 /**
- * Append a form part with the specified [key] and [filename] using [bodyBuilder] for it's body
+ * Append a form part with the specified [key], [filename] and optional [contentType] using [bodyBuilder] for it's body
  */
 @UseExperimental(ExperimentalContracts::class)
-fun FormBuilder.append(key: String, filename: String, bodyBuilder: BytePacketBuilder.() -> Unit) {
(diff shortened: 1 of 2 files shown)


e5l:
Hi @wollnyst, thanks for the PR.
Could you add the test with `contentType`?

wollnyst:
Done.

e5l:
Merged into `develop` to verify binary compatibility.

PR #919, opened 2019-02-01
Adds a hook to allows options for the JWTVerifier when using a JWKProvider

jeancharles-roger:
Hi,

I'm using a `JWKProvider` and I need to configure the `JWTVerifier` to add a leeway (see https://github.com/auth0/java-jwt/issues/268). Here is my proposition by adding a `verificationHook` to provide access to the `JWTVerifier` builder for configuration.

I also simplified the `getVerifier()` function, it seemed relevant.

Sorry for the imports, I only did see it on the pull request (I can probably can revert this).

Best regards


cy6erGn0m:
Thanks for point out the problem. This will be done via PR #1104

Issue #913, opened 2019-01-31
Extend MockEngine


sevichAk:
Hi! 

I stumbled upon a problem when writing client tests in Ktor. There is MockEngine which you pass to a HttpClient so you can mock http responses. The documentation uses matching on a url when http call is processed. But what if I want to give different responses to the same url called? Like requesting authorisation token with a different expiration date. So I can give responses based on order, in which they were added to a MockEngine. 

So I was thinking about how MockWebServer works and came up with some MockEngine implementation, where one can just enqueue MockResponses to a MockEngine and they will be given back in the order they are added. I also keep requests that are passed to MockEngine, so I can check what is actually packed into the request. 

Then I can define one MockEngine instance for client test and just reset it before every test:
```kotlin
val mockEngine = MockEngineExtended(HttpClientEngineConfig())

fun test() {
    mockEngine.enqueueResponses(
        respondOkFile("first.json"),
        respondOkFile("second.json")
    )

    val responseOne = service.fireRequest()
    val responseTwo = service.fireRequest()
    // assertions
}

fun reset() {
  mockEngine.reset()
}
```

Here is the implementation

```kotlin
class MockEngineExtended(
    override val config: HttpClientEngineConfig
) : HttpClientEngine {

    private var invocationCount = 0

    private val responses: MutableMap<Int, HttpClientCall.() -> MockHttpResponse> = mutableMapOf()

    private val processedRequests: MutableList<MockHttpRequest> = mutableListOf()

    fun enqueueResponses(vararg mockResponses: HttpClientCall.() -> MockHttpResponse) {
        mockResponses.forEachIndexed { index, function ->
            responses[index] = function
        }
    }

    fun reset() {
        responses.clear()
        processedRequests.clear()
    }

    fun takeRequest(order: Int = processedRequests.size - 1) = processedRequests[order]

    override val dispatcher: CoroutineDispatcher = Dispatchers.Unconfined

    override val coroutineContext: CoroutineContext = dispatcher + CompletableDeferred<Unit>()

    override suspend fun execute(call: HttpClientCall, data: HttpRequestData): HttpEngineCall {
        val request = data.toRequest(call)
        if (invocationCount == responses.size) {
            error("Unhandled ${request.url} on invocationCount=$invocationCount")
        } else {
            processedRequests.add(request)
            val response = responses[invocationCount++]!!.invoke(call)
            return HttpEngineCall(request, response)
        }
    }

    override fun close() {}

[...]

PR #924, opened 2019-02-02
MockEngine version which favors execution order and returns processed requests

sevichAk:
Related issue - https://github.com/ktorio/ktor/issues/913


e5l:
Hi @sevichAk, thanks for the report.
LGTM

PR #937, opened 2019-02-08
Allow header versioning

pkulak:
The definition of "json" is currently a bit too restrictive to allow GItHub-style header versioning without these changes.


pkulak:
Sorry, I changed my mind about this strategy. I'll open up a new PR later.

pkulak:
Okay, I spent some more time running this locally and think it probably is the right way to go. Open to feedback, however!

Issue #671, opened 2018-10-24
Support native windows HTTP client 


dtretyakov:
To create native windows applications could be useful to have native HTTP client which supports using Windows certificates store for SSL connections, proxy settings configuration on the system level, etc.

Microsoft recommends using [WinHTTP](https://docs.microsoft.com/en-us/windows/desktop/winhttp/about-winhttp) for server-side applications and services. WinHTTP is used in .NET [windows http client](https://github.com/dotnet/corefx/tree/master/src/System.Net.Http.WinHttpHandler/src/System/Net/Http) so it's a good candidate to use it on Windows. WinHTTP could work in sync and [async mode, which is recommended to use](https://docs.microsoft.com/en-us/windows/desktop/api/winhttp/nf-winhttp-winhttpopen#remarks) and async calls could be wrapped by using Kotlin coroutines.

dtretyakov:
The sample kotlin native application which provides WinHTTP-based client is available here: https://github.com/dtretyakov/kotlin-winhttp

Currently it has the following problems:
* To use suspend functions in kotlin native coroutines should be supported multi-threading working mode, since WinHTTP uses pool of worker threads and completion callback could be executed on any of them: (https://github.com/Kotlin/kotlinx.coroutines/issues/462)
* WinHTTP headers are not included as Windows platform library in kotlin native, so [it requires having cintorop library](https://github.com/dtretyakov/kotlin-winhttp/blob/master/src/mingwMain/c_interop/winhttp.def). I'm not sure where is a better place to have winhttp interop in the [windows platform](https://github.com/JetBrains/kotlin-native/tree/master/platformLibs/src/platform/mingw) or in this repository.
* In WinAPI most of calls returns string as UTF16 and I it looks like kotlin native library is missing helper functions to create Kotlin strings from `ShortArray`s and it requires constructing string [like that](https://github.com/dtretyakov/kotlin-winhttp/blob/63d83476219aaa986cd21b421d995bf169505a49/src/mingwMain/kotlin/sample/RequestContext.kt#L119-L121):
```
val buffer = allocArray<ShortVar>(size)
// Call to fill in the buffer
String(CharArray(size) {
    buffer[it].toChar()
})
```

cy6erGn0m:
Kotlinx I/O supports string constructor  `String(ByteArray, Charset)`, perhaps I can support construction from any CPointer as well

cy6erGn0m:
I believe it's ok to have winhttp cinterop in the client

[...]

PR #957, opened 2019-02-18
Add WinHTTP-based HTTP client

dtretyakov:
This change brings WinHTTP-based HTTP client on Windows platform. [WinHTTP library](https://docs.microsoft.com/en-us/windows/desktop/WinHttp/about-winhttp) is available since Windows XP, Windows 2000 and is used in .NET framework. 

WinHTTP provides async working mode where via request function could be trigger some operation and result of it will be returned in the [callback function](https://github.com/ktorio/ktor/blob/583d3165db23f8064b8374a1a71e9e6880f11a63/ktor-client/ktor-client-winhttp/windows/src/io/ktor/client/engine/winhttp/internal/WinHttpCallbacks.kt) which is executed on any thread from WinHTTP thread pool.

This PR addresses #671

**Note**: The latest commit (https://github.com/ktorio/ktor/commit/583d3165db23f8064b8374a1a71e9e6880f11a63) adds ability to stream request/response body by using deferred calls, but it leads to deadlocks during continuation after `.await()`, so it requires help to resolve this problem.


dtretyakov:
@e5l, in the recent commit https://github.com/ktorio/ktor/pull/957/commits/17da3d2494f065d05082e43d74f97d97a88df199 was introduced synchronous working mode enabled by default where continuations are not used and as result it is not affected by deadlocks.

e5l:
Hi @dtretyakov, thanks for the pr 
Is the implementation ready for review?

dtretyakov:
@e5l, yes, it's ready for review and contribution. Here is the following highlights:
* The synchronous mode for WinHTTP works just fine.
* The asynchronous mode has problems with deadlocks during continuation after `await()` calls, so if you know how it could be solved please feel free to improve it. Otherwise it could be disabled for now.
* Currently request body is [loaded as `ByteArray` into memory](https://github.com/ktorio/ktor/blob/2c31d1462deab23e49446ba0b890aff0458fd42f/ktor-client/ktor-client-winhttp/windows/src/io/ktor/client/engine/winhttp/internal/WinHttpRaw.kt#L19-L27) the same way as in Curl HttpClient, but as an enhancement [it could be written as chunks](https://github.com/ktorio/ktor/blob/2c31d1462deab23e49446ba0b890aff0458fd42f/ktor-client/ktor-client-winhttp/windows/src/io/ktor/client/engine/winhttp/WinHttpClientEngine.kt#L54-L62).

e5l:
Ok. I accept and merge the PR into `e5l/develop` to continue work on it.

vehovsky:
Hi @e5l, any news on this one?

PR #983, opened 2019-02-27
Move client auth to common

e5l:


ktor-client/ktor-client-features/ktor-client-auth/common/src/io/ktor/client/features/auth/providers/BasicAuthProvider.kt
```diff
@@ -43,6 +45,7 @@ class BasicAuthProvider(
     private val password: String,
     private val realm: String?
 ) : AuthProvider {
+    private val defaultCharset = Charset.forName("ISO_8859_1")
```
cy6erGn0m: This is not going to work on JS. In spite of that encoding is supported on the platform, kotlinx-io doesn't allow to get this charset at all :(

cy6erGn0m: Perhaps, we need to support ISO in kxio js first

PR #993, opened 2019-03-01
Fixes SessionTransportTransformerEncrypt performance

soywiz:
SecureRandom.generateSeed might be extremely slow on linux since it requires real random data.
This patch uses a different approach for the default ivGenerator.


ktor-server/ktor-server-core/jvm/src/io/ktor/sessions/SessionTransportTransformerEncrypt.kt
```diff
@@ -29,14 +29,18 @@ import javax.crypto.spec.*
...
-    val ivGenerator: (size: Int) -> ByteArray = { size -> SecureRandom().generateSeed(size) },
+    ivGenerator: ((size: Int) -> ByteArray)? = null,
     val encryptAlgorithm: String = encryptionKeySpec.algorithm,
     val signAlgorithm: String = signKeySpec.algorithm
 ) : SessionTransportTransformer {
     companion object {
         private val log = LoggerFactory.getLogger(SessionTransportTransformerEncrypt::class.qualifiedName)
     }
 
+    private val random by lazy { SecureRandom.getInstance("SHA1PRNG").apply { generateSeed(16) } }
```
cy6erGn0m: Is it concurrent safe? If not then it can't be used like this. Perhaps, an object pool is required here.

ktor-server/ktor-server-core/jvm/src/io/ktor/sessions/SessionTransportTransformerEncrypt.kt
```diff
@@ -29,14 +29,18 @@ import javax.crypto.spec.*
 class SessionTransportTransformerEncrypt(
     val encryptionKeySpec: SecretKeySpec,
     val signKeySpec: SecretKeySpec,
-    val ivGenerator: (size: Int) -> ByteArray = { size -> SecureRandom().generateSeed(size) },
```
cy6erGn0m: Why nullable? Can't it simply use the defauilt in the default lambda argument?

ktor-server/ktor-server-core/jvm/src/io/ktor/sessions/SessionTransportTransformerEncrypt.kt
JLLeitschuh: > Instances of java.util.Random are threadsafe. However, the concurrent use of the same java.util.Random instance across threads may encounter contention and consequent poor performance. Consider instead using ThreadLocalRandom in multithreaded designs.
> \- https://docs.oracle.com/javase/7/docs/api/java/util/Random.html

However, using `ThreadLocalRandom` is not cryptographically secure, so don't use that.

JLLeitschuh: SHA-1 is considered cryptographically broken by NIST as of 2011. Does that same concern apply here?

JLLeitschuh: I'd just recommend using `new SecureRandom()`. It will be secure enough.

PR #998, opened 2019-03-06
Added ReadTimeoutHandler to Netty pipeline.

sizeak:
I encountered an issue with the Netty pipeline where it's possible to keep a connection open indefinitely when sending a request with a JSON body (parsed by Jackson) that doesn't complete, for example by using Telnet to send a request but never closing it (I have hit this issue in real world use too though).

When using ktor in an environment where it will only handle a single connection at a time, this causes it to entirely stop responding to new requests because the stream is kept open waiting for the partial request to complete. 

This PR adds a `ReadTimeoutHandler` to Netty which terminates the connection after a configurable timeout, as the pipeline already does for writes. I'm happy to make changes to the PR if I've done anything nasty  :)

Thanks!

### DIFF
--- a/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationEngine.kt
+++ b/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationEngine.kt
@@ -48,6 +48,11 @@ class NettyApplicationEngine(environment: ApplicationEngineEnvironment, configur
         var responseWriteTimeoutSeconds: Int = 10
 
         /**
+         * Timeout in seconds for reading requests from client.
+         */
+        var requestReadTimeoutSeconds: Int = 10
+
+        /**
          * User-provided function to configure Netty's [HttpServerCodec]
          */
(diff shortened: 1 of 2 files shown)


cy6erGn0m:
Well, at least this should prevent websockets from working: will always disconnect after 10 seconds.

cy6erGn0m:
Also file upload through a slow connection will always get aborted quite fast. Not sure if it is somehow different from response write timeout that will abort response transfer to a slow client.

sizeak:
Yes that's true, 10 seconds is good for our use case, maybe not so good for others. I could update the PR so the default timeout is infinite, which would maintain the old behaviour unless the explicitly changed, would that be OK?

cy6erGn0m:
Yes, and in the infinite case it shouldn't install the handler at all since it has significant memory traffic cost.

cy6erGn0m:
The other problem with finite small timeout is that it prevents HTTP pipelining and does always discard client's connection pools since the handler is installed before codec so it causes connection close at IDLE between requests.

sizeak:
OK thanks for the tips, will update.

sizeak:
@cy6erGn0m is this ok now?

PR #1014, opened 2019-03-13
Workaround variadic C function arguments

cy6erGn0m:
Since variadic C function arguments is no longer supported in 1.3.30 we need to specify exact list. Passing `Any` arguments is restricted as well.

See [KT-30386](https://youtrack.jetbrains.com/issue/KT-30386) for more the explanation.

### DIFF
--- a/ktor-client/ktor-client-curl/posix/src/io/ktor/client/engine/curl/internal/CurlAdapters.kt
+++ b/ktor-client/ktor-client-curl/posix/src/io/ktor/client/engine/curl/internal/CurlAdapters.kt
@@ -25,12 +25,28 @@ internal fun CURLcode.verify() {
     }
 }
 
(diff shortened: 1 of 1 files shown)


ktor-client/ktor-client-curl/posix/src/io/ktor/client/engine/curl/internal/CurlAdapters.kt
```diff
@@ -25,12 +25,28 @@ internal fun CURLcode.verify() {
     }
 }
 
-internal fun EasyHandle.option(option: CURLoption, vararg variadicArguments: Any) {
```
e5l: Please consider using `optionValue` name instead.

cy6erGn0m:
Merged.

PR #1036, opened 2019-03-22
allow many ContentTypes when use JsonFeature

gkm2164:
Hi, while I use the JsonFeature to ktor-client, I wanted to allow more ContentTypes (from other server, they send json data with not a 'application/json' one, like, 'application/x-json')

My recommendation is as follows.
At the request and response process side,(line 58 ~ 61 and 75 ~ 76 at my new code) beside comparing with only "application/json", by comparing with list of allowed content types with list operation, like ```forEach``` and ```none``` function, this would be accomplished easily.
And for the configuration part, I think by adding more config option, (I named allowedContentTypes), and set default value to the "application/json", changing other codes would not be required.

Thanks.
p.s. Sorry for poor writing, naming or else.

### DIFF
--- a/ktor-client/ktor-client-features/ktor-client-json/common/src/io/ktor/client/features/json/JsonFeature.kt
+++ b/ktor-client/ktor-client-features/ktor-client-json/common/src/io/ktor/client/features/json/JsonFeature.kt
@@ -29,7 +29,10 @@ expect fun defaultSerializer(): JsonSerializer
  * Note: It will de-serialize the body response if the specified type is a public accessible class
  *       and the Content-Type is `application/json`
  */
-class JsonFeature(val serializer: JsonSerializer) {
+class JsonFeature(
+    val serializer: JsonSerializer,
+    val allowedContentTypes: List<ContentType>
+) {
     class Config {
         /**
          * Serialized that will be used for serializing requests bodies,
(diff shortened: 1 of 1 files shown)


gkm2164:
I've just rebased to current master.

e5l:
Hi @gkm2164, could you recheck and squash history?

gkm2164:
> Hi @gkm2164, could you recheck and squash history?

Did I do it right?

e5l:
It looks like you accidentally grab some side changes :).
Please make sure that changelog contains files modified only by you.

gkm2164:
> It looks like you accidentally grab some side changes :).
> Please make sure that changelog contains files modified only by you.

Ah.. got it.  I think the rebasing that I used was not proper to use in this case. I (forcefully) committed again over current ktor master branch. Thanks.

rumatoest:
Please release with feature ASAP.
There are many dumb API that returns JSON with any other content types like "document/html"
And all what I can get is NoTransformationFoundException  without any reasoning message about actual cause, there is no mention about wrong content type in response.

ktor-client/ktor-client-features/ktor-client-json/common/src/io/ktor/client/features/json/JsonFeature.kt
```diff
@@ -38,19 +41,25 @@ class JsonFeature(val serializer: JsonSerializer) {
          * Default value is [defultSerializer]
          */
         var serializer: JsonSerializer? = null
+        var allowedContentTypes: List<ContentType>? = null
```
cy6erGn0m: Can't it be simply `listOf(ContentType.Application.Json)` instead of nullable so there will be no need to have elvis in `prepare`

cy6erGn0m: Also it should be prohibited to set an empty list of content types

cy6erGn0m:
Also at least one test with a custom content type is required

ktor-client/ktor-client-features/ktor-client-json/common/src/io/ktor/client/features/json/JsonFeature.kt
cy6erGn0m: Also this property requires kdoc and `@KtorExperimentalAPI` since it will be deprecated in the future once client-side content negotiation will be introduced.

gkm2164: As you pointed out, 
- changed ```allowedContentTypes``` variable to non-nullable type.
- prohibited initialization with empty list of content type, by throwing a proper exception.
- added some documentation for explaining of this config variable.

Thanks for your advice!

PR #1044, opened 2019-03-27
Eliminate byteOrder usage + fix

cy6erGn0m:
- Eliminate using mutable channel's byteOrder, using `reverseByteOrder` instead.
- Fix random gzip/deflate corruption on JDK11


cy6erGn0m:
Merged.

PR #1062, opened 2019-04-06
Replace text with reader in JsonSerializer.

robxyy:


JLLeitschuh:
This was actually a valuable PR that would have improved performance. What kept it from being merged?

PR #1073, opened 2019-04-16
add missing quotes

XGFan:
when use multipart/form-data 
the name of item should be wrapped in quotes

ref:https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition


cy6erGn0m:
At least `String.quote()` or `String.escapeIfNeeded` should be used

cy6erGn0m:
Or `ContentDisposition` class should be used instead

cy6erGn0m:
Will be fixed manually. Thanks for pointing out this.

drampelt:
@cy6erGn0m did this ever get fixed? I'm not seeing any changes related to this in the code

PR #1079, opened 2019-04-18
Infrastructure to build Ktor against Kotlin snapshots

qwwdfsad:


build.gradle
```diff
@@ -115,6 +158,31 @@ allprojects {
     }
 }
 
+println("Using Kotlin compiler version: $org.jetbrains.kotlin.config.KotlinCompilerVersion.VERSION")
+if (build_snapshot_train) {
+    println "Hacking test tasks, removing stress and flaky tests"
+    allprojects {
+        tasks.withType(Test).all {
+            exclude '**/*io.ktor.tests.server*'
```
qwwdfsad: probably should be `io/ktor/server/tests`

gradle.properties
```diff
@@ -50,4 +50,4 @@ text_encoding_version=0.7.0
 node_fetch_version=2.3.0
 ws_version=6.2.1
 
-org.gradle.parallel=true
+#org.gradle.parallel=true
```
cy6erGn0m: Disabling parallel may seriously slowdown the build. Is it required for train? Why?

PR #1080, opened 2019-04-22
Bugfixes and improvements

cy6erGn0m:


ktor-utils/jvm/src/io/ktor/util/CryptoJvm.kt
```diff
@@ -55,7 +67,7 @@ private inline class DigestImpl(val delegate: MessageDigest) : Digest {
     ReplaceWith("s.toByteArray(Charsets.UTF_8)"),
     level = DeprecationLevel.ERROR
 )
-fun raw(s: String) = s.toByteArray(Charsets.UTF_8)
+fun raw(s: String): ByteArray = s.toByteArray(Charsets.UTF_8)
```
e5l: Could it be private or should we rename the parameter and add the doc?

ktor-server/ktor-server-core/jvm/src/io/ktor/util/Paths.kt
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    return result
+}
+
+private fun MutableList<String>.addReplacedComponent(component: String) {
+    if (component.isEmpty() || component == "." || component == "~" || component.toUpperCase() in ReservedWords) return
+    if (component == "..") {
+        if (isNotEmpty()) {
+            removeAt(lastIndex)
```
e5l: It's quite ambiguous to call remove in add method. Could we avoid that?

ktor-server/ktor-server-core/jvm/src/io/ktor/util/Paths.kt
```diff
@@ -0,0 +1,98 @@
...
+            result.addReplacedComponent(component)
+        } else {
+            result.add(component)
+        }
+    }
+
+    return result
+}
+
+private fun MutableList<String>.addReplacedComponent(component: String) {
```
e5l: Can we have trailing spaces in the component?

ktor-features/ktor-auth-ldap/jvm/test/io/ktor/tests/auth/ldap/LdapEscapeTest.kt
```diff
@@ -0,0 +1,33 @@
...
+
+    @Test
+    fun testUnicodeAndSpecials() {
+        assertEquals("\\00", ldapEscape("\u0000"))
+        assertEquals("\\d0\\96", ldapEscape("\u0416"))
+    }
+
+    @Test
+    fun integration() {
+
```
e5l: empty

ktor-utils/jvm/src/io/ktor/util/CryptoJvm.kt
cy6erGn0m: It is already deprecated with error, going to be dropped

ktor-server/ktor-server-core/jvm/src/io/ktor/util/Paths.kt
cy6erGn0m: Yes, rename to processComponent to avoid add prefix

ktor-server/ktor-server-core/jvm/src/io/ktor/util/Paths.kt
cy6erGn0m: No, trailing spaces will be discarded: they are not allowed in Windows

cy6erGn0m:
Merged

PR #1092, opened 2019-04-29
 Improve session provider registration diagnostics 

cy6erGn0m:


ktor-server/ktor-server-core/jvm/src/io/ktor/sessions/SessionProvider.kt
```diff
@@ -17,4 +17,8 @@ import kotlin.reflect.*
 class SessionProvider(val name: String,
```
e5l: Codestyle

ktor-server/ktor-server-core/jvm/src/io/ktor/sessions/SessionTrackerByValue.kt
```diff
@@ -16,7 +16,14 @@ import kotlin.reflect.*
  */
 class SessionTrackerByValue(val type: KClass<*>, val serializer: SessionSerializer) : SessionTracker {
     override suspend fun load(call: ApplicationCall, transport: String?): Any? {
-        return transport?.let { serializer.deserialize(it) }
+        return transport?.let { serialized ->
+            try {
+                serializer.deserialize(serialized)
+            } catch (t: Throwable) {
+                call.application.log.debug("Failed to deserialize session: $serialized", serialized)
```
e5l: Should we also log cause here?

cy6erGn0m:
Fixed and merged

PR #1096, opened 2019-04-30
Make hot reload work with JDK9+

cy6erGn0m:
Currently it requires an additional VM option to be passed:
--add-opens java.base/jdk.internal.loader=ALL-UNNAMED


cy6erGn0m:
Merged.

PR #1100, opened 2019-05-06
Retain empty url path (ktorio#958)

### LINKED ISSUE #958: URLBuilder: extra '/' at the end, when parameters are used
### Ktor Version: 
1.1.2

### Ktor Engine Used(client or server and name): 
Multiplatform/client (io.ktor:ktor-client-core, io.ktor:ktor-client-core-jvm)

### JVM Version, Operating System and Relevant Context: 
Tested on Android 8.1.0

### Feedback

```url {
                    protocol = URLProtocol.HTTPS
                    host = endpoint.url // String: "url.com"
                    parameters.append(name = "before", value = endpoint.parameters.get("before"))
                    parameters.append(name = "limit", value = (endpoint.parameters.get("limit") as? Int).toString()
                }
                println(url.buildString().toString())```

Prints:

```https://url.com/?before=now&limit=50```

Expected:

```https://url.com?before=now&limit=50```

doyaaaaaken:
Fixes #958 .

Before this change, `URLBuilder` convert "http://localhost" into "http://localhost/" (automatically append root path `/`).
But described in [RFC 3986 section-3.3](https://tools.ietf.org/html/rfc3986#section-3.3), empty path is allowed in URI string.

>path          = path-abempty    ; begins with "/" or is empty
                    / path-absolute   ; begins with "/" but not "//"
                    / path-noscheme   ; begins with a non-colon segment
                    / path-rootless   ; begins with a segment
                    / path-empty      ; zero characters

### DIFF
--- a/ktor-http/common/src/io/ktor/http/URLUtils.kt
+++ b/ktor-http/common/src/io/ktor/http/URLUtils.kt
@@ -86,7 +86,7 @@ internal fun Appendable.appendUrlFullPath(
     queryParameters: Parameters,
     trailingQuery: Boolean
 ) {
-    if (!encodedPath.startsWith("/")) {
+    if (encodedPath.isNotBlank() && !encodedPath.startsWith("/")) {
         append('/')
     }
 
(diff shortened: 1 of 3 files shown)


e5l:
Hi, @doyaaaaaken. Thanks for the PR. 
It looks like some tests are broken. Could you check them?

doyaaaaaken:
@e5l 
I'm sorry! Fixes failed test in 831c679.

In 831c679 , I change test `testMissingSlash` into `testRetainEmptyPath`.
But they assert opposite result. (because of specification change by this PR)

PR #1105, opened 2019-05-07
abstracted micrometer feature so we can provide provider specific

JotBePunkt:
as wished by @dave08 in #1037 In also added an prometheus specific implementation that exposes the metrics endpoint with an optional authentication

### DIFF
--- a/ktor-features/ktor-metrics-micrometer-prometheus/jvm/src/io/ktor/metrics/micrometer/prometheus/MicrometerMetricsPrometheus.kt
+++ b/ktor-features/ktor-metrics-micrometer-prometheus/jvm/src/io/ktor/metrics/micrometer/prometheus/MicrometerMetricsPrometheus.kt
@@ -1,0 +1,149 @@
+/*
+ * Copyright 2014-2019 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.metrics.micrometer.prometheus
+
+import io.ktor.application.Application
+import io.ktor.application.ApplicationCall
+import io.ktor.application.call
+import io.ktor.application.feature
+import io.ktor.http.ContentType
+import io.ktor.http.charset
+import io.ktor.http.content.OutgoingContent
+import io.ktor.metrics.micrometer.AbstractMicrometerMetrics
+import io.ktor.response.ApplicationSendPipeline
+import io.ktor.response.respond
+import io.ktor.routing.get
+import io.ktor.routing.routing
+import io.ktor.util.cio.bufferedWriter
+import io.micrometer.core.instrument.distribution.DistributionStatisticConfig
+import io.micrometer.prometheus.PrometheusConfig
+import io.micrometer.prometheus.PrometheusMeterRegistry
+import io.prometheus.client.Collector
+import io.prometheus.client.exporter.common.TextFormat
+import kotlinx.coroutines.io.ByteWriteChannel
+import java.util.*
+
+class MicrometerMetricsPrometheus(config: Configuration) :
+    AbstractMicrometerMetrics<PrometheusMeterRegistry, MicrometerMetricsPrometheus.Configuration>(config) {
+
+    internal fun process(value: MetricsContent) = object : OutgoingContent.WriteChannelContent() {
+
+        override val contentType = ContentType.parse(TextFormat.CONTENT_TYPE_004)
+
+        override suspend fun writeTo(channel: ByteWriteChannel) {
+            val metricSamplesToScrape = value.filteredMetricFamilySamples()
+
+            channel.bufferedWriter(contentType.charset() ?: Charsets.UTF_8).use {
+                TextFormat.write004(it, metricSamplesToScrape)
+            }
+        }
+    }
+
+    class Configuration : AbstractMicrometerMetrics.AbstractConfiguration<PrometheusMeterRegistry>() {
+
+        /**
+         * The meter registry
+         */
+        override var registry = PrometheusMeterRegistry(PrometheusConfig.DEFAULT)
+
+        override var distributionStatisticConfig: DistributionStatisticConfig = DistributionStatisticConfig.Builder()
+            .percentilesHistogram(true)
+            .build()
+
+        /**
+         * Defines if a metrics endpoint is installed under "/metrics" for all metrics registered at [registry].
+         * Set to false if you want to declare a custom metric endpoint instead.
+         */
+        var installDefaultEndpoint = true
+    }
+
+
+    companion object MetricsFeature :
+        AbstractMetricsFeature<PrometheusMeterRegistry, Configuration>(
+            ::MicrometerMetricsPrometheus,
+            MicrometerMetricsPrometheus::Configuration
+        ) {
+
+        override fun install(
+            pipeline: Application,
+            configure: Configuration.() -> Unit
+        ): AbstractMicrometerMetrics<PrometheusMeterRegistry, Configuration> {
+            val feature = super.install(pipeline, configure) as MicrometerMetricsPrometheus
+
+            pipeline.sendPipeline.intercept(ApplicationSendPipeline.Transform) { value ->
+                if (value is MetricsContent) {
+                    val response = feature.process(value)
+                    proceedWith(response)
+                }
+            }
+
+            if (feature.config.installDefaultEndpoint)
+                pipeline.routing {
+                    get("/metrics") {
+                        call.respondMetrics()
+                    }
+                }
+
+            return feature
+        }
+    }
+}
+
+
+/**
+ * the meter registry registered for this feature
+ */
+val ApplicationCall.meterRegistry
+    get() = application.feature(MicrometerMetricsPrometheus).config.registry
+
+/**
+ * responds with scraped metrics of this registry, filtered by the specified filter. The filter works after the names
+ * of the metrics have been "prometheusized" meaning that the names and tags
+ */
+suspend fun ApplicationCall.respondMetrics(
+    registry: PrometheusMeterRegistry = meterRegistry,
+    filter: Collector.MetricFamilySamples.() -> Boolean = { true }
+) =
+    respond(MetricsContent(registry, filter))
+
+
+internal class MetricsContent(
+    val registry: PrometheusMeterRegistry,
+    val filter: Collector.MetricFamilySamples.() -> Boolean
+) {
+    fun filteredMetricFamilySamples(): Enumeration<Collector.MetricFamilySamples> =
+        object : Enumeration<Collector.MetricFamilySamples> {
+
+            private val internalCollection =
+                registry.prometheusRegistry.metricFamilySamples()
+
+            private var currentElement: Collector.MetricFamilySamples? = findNext()
+
+            private fun findNext(): Collector.MetricFamilySamples? {
+                var next: Collector.MetricFamilySamples?
+                do {
+                    if (internalCollection.hasMoreElements())
+                        next = internalCollection.nextElement()
+                    else
+                        next = null
+                } while (next != null && !acceptedByFilter(next))
+
+                return next
+            }
+
+            private fun acceptedByFilter(samples: Collector.MetricFamilySamples) =
+                filter(samples)
+
+            override fun hasMoreElements(): Boolean =
+                currentElement != null
+
+            override fun nextElement(): Collector.MetricFamilySamples {
+                val result = currentElement
+                currentElement = findNext()
+                return result ?: throw NoSuchElementException()
+            }
+        }
+}
+
--- a/ktor-features/ktor-metrics-micrometer/jvm/src/io/ktor/metrics/micrometer/AbstractMicrometerMetrics.kt
+++ b/ktor-features/ktor-metrics-micrometer/jvm/src/io/ktor/metrics/micrometer/AbstractMicrometerMetrics.kt
@@ -1,0 +1,237 @@
+/*
+ * Copyright 2014-2019 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.metrics.micrometer
+
+import io.ktor.application.*
+import io.ktor.request.httpMethod
+import io.ktor.request.path
+import io.ktor.response.ApplicationSendPipeline
+import io.ktor.routing.Routing
+import io.ktor.util.AttributeKey
+import io.ktor.util.pipeline.PipelinePhase
+import io.micrometer.core.instrument.Meter
+import io.micrometer.core.instrument.MeterRegistry
+import io.micrometer.core.instrument.Tag
+import io.micrometer.core.instrument.Timer
+import io.micrometer.core.instrument.binder.MeterBinder
+import io.micrometer.core.instrument.binder.jvm.ClassLoaderMetrics
+import io.micrometer.core.instrument.binder.jvm.JvmGcMetrics
+import io.micrometer.core.instrument.binder.jvm.JvmMemoryMetrics
+import io.micrometer.core.instrument.binder.jvm.JvmThreadMetrics
+import io.micrometer.core.instrument.binder.system.FileDescriptorMetrics
+import io.micrometer.core.instrument.binder.system.ProcessorMetrics
+import io.micrometer.core.instrument.config.MeterFilter
+import io.micrometer.core.instrument.distribution.DistributionStatisticConfig
+import java.util.concurrent.atomic.AtomicInteger
+
+
+/**
+ * Abstract class for several micrometer features.
+ */
+abstract class AbstractMicrometerMetrics<TMeterRegistry : MeterRegistry,
+    TConfiguration : AbstractMicrometerMetrics.AbstractConfiguration<TMeterRegistry>>(
+    val config: TConfiguration
+) {
+
+    private val active = config.registry.gauge(activeGaugeName, AtomicInteger(0))
+
+    init {
+        enableTimerDistributionConfig(config.distributionStatisticConfig)
+    }
+
+    private fun enableTimerDistributionConfig(timerDistributionConfig: DistributionStatisticConfig) {
+        config.registry.config().meterFilter(object : MeterFilter {
+            override fun configure(id: Meter.Id, config: DistributionStatisticConfig): DistributionStatisticConfig =
+                if (id.name == requestTimerName)
+                    timerDistributionConfig.merge(config)
+                else
+                    config
+        })
+    }
+
+    private fun CallMeasure.recordDuration(call: ApplicationCall) {
+        timer.stop(
+            Timer.builder(requestTimerName)
+                .addDefaultTags(call, throwable)
+                .customize(call, throwable)
+                .register(config.registry)
+        )
+    }
+
+    private fun Timer.Builder.customize(call: ApplicationCall, throwable: Throwable?) =
+        this.apply {
+            config.timerBuilder.invoke(this, call, throwable)
+        }
+
+    private fun Timer.Builder.addDefaultTags(call: ApplicationCall, throwable: Throwable?): Timer.Builder {
+        tags(
+            listOf(
+                Tag.of("address", call.request.local.let { "${it.host}:${it.port}" }),
+                Tag.of("method", call.request.httpMethod.value),
+                Tag.of("route", call.attributes[measureKey].route ?: call.request.path()),
+                Tag.of("status", call.response.status()?.value?.toString() ?: "n/a"),
+                Tag.of("throwable", throwable?.let { it::class.qualifiedName } ?: "n/a")
+            )
+        )
+        return this
+    }
+
+    private fun before(call: ApplicationCall) {
+        active?.incrementAndGet()
+
+        call.attributes.put(
+            measureKey,
+            CallMeasure(Timer.start(config.registry))
+        )
+    }
+
+    private fun after(call: ApplicationCall) {
+        active?.decrementAndGet()
+
+        call.attributes.getOrNull(measureKey)?.recordDuration(call)
+
+    }
+
+    private fun throwable(call: ApplicationCall, t: Throwable) {
+        call.attributes.getOrNull(measureKey)?.apply {
+            throwable = t
+        }
+    }
+
+    /**
+     * Configures this MetricsFeature
+     * @property registry The meter registry where the meters are registered. Mandatory
+     * @property meterBinders The binders that are automatically bound to the registry. Default: [ClassLoaderMetrics],
+     * [JvmMemoryMetrics], [ProcessorMetrics], [JvmGcMetrics], [ProcessorMetrics], [JvmThreadMetrics], [FileDescriptorMetrics]
+     * @property distributionStatisticConfig configures the histogram and/or percentiles for all request timers. By
+     * default 50%, 90% , 95% and 99% percentiles are configured. If your backend supports server side histograms you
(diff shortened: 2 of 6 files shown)


ktor-server/ktor-server-jetty/ktor-server-jetty-test-http2/build.gradle
```diff
@@ -18,9 +18,9 @@ configure(tasks.jvmTest) {
     exclude("**/*StressTest*")
 
     if (JavaVersion.current() == JavaVersion.VERSION_1_8) {
-        jvmArgs {
-            def bootClasspath = configurations.boot.files
-            bootClasspath.collect { "-Xbootclasspath/p:${it.absolutePath}" }.iterator()
-        }
+       // jvmArgs {
```
cy6erGn0m: ?

ktor-features/ktor-metrics-micrometer/jvm/src/io/ktor/metrics/micrometer/MicrometerMetrics.kt
```diff
@@ -42,177 +31,24 @@ import java.util.concurrent.atomic.*
...
-
-            return feature
-        }
-    }
-
-
 }
 
+/**
+ * the meter registry registered for this feature
```
cy6erGn0m: Capitalize, same in other kdocs

ktor-features/ktor-metrics-micrometer/jvm/src/io/ktor/metrics/micrometer/AbstractMicrometerMetrics.kt
```diff
@@ -0,0 +1,212 @@
...
+
+    }
+
+
+    companion object {
+
+        private val measureKey = AttributeKey<CallMeasure>("metrics")
+
+        private const val baseName: String = "ktor.http.server"
+        const val requestTimerName = "$baseName.requests"
```
cy6erGn0m: kdoc or internal

ktor-features/ktor-metrics-micrometer/jvm/src/io/ktor/metrics/micrometer/AbstractMicrometerMetrics.kt
```diff
@@ -0,0 +1,212 @@
...
+                throw IllegalArgumentException(
+                    "Meter registry is missing. Please initialize the field 'registry'"
+                )
+            }
+
+            val feature = constructor(configuration)
+
+            configuration.meterBinders.forEach { it.bindTo(configuration.registry) }
+
+            val phase = PipelinePhase("Metrics")
```
cy6erGn0m: Please note that only one metrics feature could be installed since double phase insertion is not allowed. Phase need to be renamed or installed only if not yet present.

ktor-features/ktor-metrics-micrometer/jvm/src/io/ktor/metrics/micrometer/AbstractMicrometerMetrics.kt
```diff
@@ -0,0 +1,212 @@
...
+                feature.before(call)
+                try {
+                    proceed()
+                } catch (e: Throwable) {
+                    feature.throwable(call, e)
+                    throw e
+                }
+            }
+
+            val postSendPhase = PipelinePhase("MetricsPostSend")
```
cy6erGn0m: Similar to `Metrics` phase

ktor-features/ktor-metrics-micrometer-prometheus/jvm/src/io/ktor/metrics/micrometer/prometheus/installFeatureWithAuth.kt
```diff
@@ -0,0 +1,32 @@
...
+            metricsEndpoint(config)
+        }
+    }
+}
+
+private fun Application.checkIfAuthenticationFeatureIsInstalled() {
+    try {
+        feature(Authentication)
+    } catch (ex: MissingApplicationFeatureException) {
+        throw Exception(
```
cy6erGn0m: Why `Exception`? I'd prefer to see some meaningful type

ktor-features/ktor-metrics-micrometer-prometheus/jvm/src/io/ktor/metrics/micrometer/prometheus/installFeatureWithAuth.kt
```diff
@@ -0,0 +1,32 @@
...
+ * Copyright 2014-2019 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.metrics.micrometer.prometheus
+
+import io.ktor.application.*
+import io.ktor.auth.*
+import io.ktor.routing.*
+
+fun Application.installMetricsEndpointWithAuth(config: MicrometerMetricsPrometheus.Configuration) {
```
cy6erGn0m: We usually call such functions in declarative manner. For example, `routing { }` instead of `installRouting`, `authentication` instead of `installAuthentication` and so on. Also it makes sense to have micrometer in the name. Possible names would be `prometheusEndpoint` or `micrometerPrometheusEndpoint`. 

The other problem is that it requires `confi`g as a parameter while we are usually providing a lambda block to configure a feature

ktor-features/ktor-metrics-micrometer-prometheus/jvm/src/io/ktor/metrics/micrometer/prometheus/installFeatureWithAuth.kt
```diff
@@ -0,0 +1,32 @@
...
+
+    routing {
+        authenticate(*config.authenticationConfig) {
+            metricsEndpoint(config)
+        }
+    }
+}
+
+private fun Application.checkIfAuthenticationFeatureIsInstalled() {
+    try {
```
cy6erGn0m: According to your description, auth is optional however we actually always require auth.

ktor-features/ktor-metrics-micrometer-prometheus/jvm/src/io/ktor/metrics/micrometer/prometheus/MicrometerMetricsPrometheus.kt
```diff
@@ -0,0 +1,89 @@
...
+    get() = call.application.feature(MicrometerMetricsPrometheus).config.registry
+
+
+internal fun Route.metricsEndpoint(config: MicrometerMetricsPrometheus.Configuration) =
+    get(config.endpointPath) {
+        prometheusMetrics()
+    }
+
+private suspend fun PipelineContext<Unit, ApplicationCall>.prometheusMetrics() {
+    call.respondText(contentType = ContentType.parse(TextFormat.CONTENT_TYPE_004)) {
```
cy6erGn0m: `CONTENT_TYPE_004` ?

ktor-features/ktor-metrics-micrometer-prometheus/jvm/src/io/ktor/metrics/micrometer/prometheus/MicrometerMetricsPrometheus.kt
```diff
@@ -0,0 +1,89 @@
...
+            MicrometerMetricsPrometheus::Configuration
+        ) {
+
+        override fun install(
+            pipeline: Application,
+            configure: Configuration.() -> Unit
+        ): AbstractMicrometerMetrics<PrometheusMeterRegistry, Configuration> {
+            val feature = super.install(pipeline, configure)
+
+            if (feature.config.authenticationConfig.isEmpty()) {
```
cy6erGn0m: This need to be unified to avoid duplication.

dave08:
Thanks @JotBePunkt,two little questions though:

1. Why do we need to reimplement the feature? Wouldn't it be enough to have an extension function like in `ContentNegotiation`?
2. Can the metrics path be installed without authentication? My use case is that my Prometheus instance scraps our services inside the internal network between our VMs, so there isn't any real need for auth.

ktor-features/ktor-metrics-micrometer-prometheus/jvm/src/io/ktor/metrics/micrometer/prometheus/MicrometerMetricsPrometheus.kt
JotBePunkt: the value behind it is `text/plain; version=0.0.4; charset=utf-8`, I would take it as is so prometheus can easily parse it

ktor-server/ktor-server-jetty/ktor-server-jetty-test-http2/build.gradle
JotBePunkt: that should not have gone into the commit, there was an error preventing me from building.

JotBePunkt: that should not have gone into the commit, there was an error preventing me from building. fixed with the next push

ktor-features/ktor-metrics-micrometer/jvm/src/io/ktor/metrics/micrometer/AbstractMicrometerMetrics.kt
JotBePunkt: I prefixed it with the actual class name, this should make it unique someone needed more than one Metric feature.

ktor-features/ktor-metrics-micrometer-prometheus/jvm/src/io/ktor/metrics/micrometer/prometheus/installFeatureWithAuth.kt
JotBePunkt: fixed (I forgot how easy it is to declare a class in Kotlin...) In general this exception should not be caught during runtime, it shows the developer that they did not configure the feature correctly.

ktor-features/ktor-metrics-micrometer-prometheus/jvm/src/io/ktor/metrics/micrometer/prometheus/installFeatureWithAuth.kt
JotBePunkt: only if authentication() config is set in the configuration, if this is not set, we dont need the auth feature

ktor-features/ktor-metrics-micrometer-prometheus/jvm/src/io/ktor/metrics/micrometer/prometheus/installFeatureWithAuth.kt
JotBePunkt: made it internal and renamed it, its called when the feature is installed

ktor-features/ktor-metrics-micrometer-prometheus/jvm/src/io/ktor/metrics/micrometer/prometheus/MicrometerMetricsPrometheus.kt
JotBePunkt: the value behind the constant is `text/plain; version=0.0.4; charset=utf-8` I would just use it like Prometheus needs it

ktor-features/ktor-metrics-micrometer-prometheus/jvm/src/io/ktor/metrics/micrometer/prometheus/installFeatureWithAuth.kt
dave08: Also, I would need, if possible, to be able to install the endpoint in a specific port (I have an internal port for prometheus scraping, and an external one that our proxy service exposes to the internet).

dave08: Maybe like providing an `var endpointPort: Int? = null`?

cy6erGn0m:
The main problem here is that there is huge abstraction, a lot of type parameters, constructor functions. Do we actually need it? Can't we replace inheritance with delegation/decoration somehow? So Prometheus endpoint feature could simply plug in into the micrometer feature

JotBePunkt:
>     1. Why do we need to reimplement the feature? Wouldn't it be enough to have an extension function like `gson { }` or `jackson { }` in `ContentNegotiation`?
The reusability is not really high, in contrast the json features, there ins only one thing that is encoded in that way (the meters registered in the prometheus meter registry) and there is more or less only one way to implement it so keep it simple here

> 2. Can the metrics path be installed without authentication? 

yes it can, just don't use the authenticate() method.
E.g.: 
https://github.com/ktorio/ktor/blob/159ad3707d9a95b932bdee275f9a2d56689fab45/ktor-features/ktor-metrics-micrometer-prometheus/jvm/test/io/ktor/metrics/micrometer/prometheus/MicrometerMetricsPrometheusTest.kt#L24

vs. (with auth)

https://github.com/ktorio/ktor/blob/159ad3707d9a95b932bdee275f9a2d56689fab45/ktor-features/ktor-metrics-micrometer-prometheus/jvm/test/io/ktor/metrics/micrometer/prometheus/MicrometerMetricsPrometheusTest.kt#L48-L50

dave08:
But there isn't only Prometheus, there's also Dropwizard, and all the other implementations of the Micrometer facade (just like `gson { }` and `jackson { }`. If the feature's code is more or less the same for all those, it might be better to use a `ContentNegotiation`-type implementation... that was my original plan. Also I'd suppose that configs for the other types also vary...

Is there any real big changes in the base-Feature implementation that would make it simpler in the way you're currently doing it?

JotBePunkt:
> The main problem here is that there is huge abstraction, a lot of type parameters, constructor functions. Do we actually need it? Can't we replace inheritance with delegation/decoration somehow? So Prometheus endpoint feature could simply plug in into the micrometer feature

Yes, it looks complicate with all the generics and parameters. My main goals was to keep the usage a simple as possible. The alternatives would that the Prometheus feature only provides the metrics endpoint and
- the user needs to install the micrometer feature manually (which increases the complexity and the feature cannot provide prometheus specific default values for the config)
- the prometheus feature installs automatically the micrometer feature (which is not transparent for the user)

dave08:
That might still be possible by having the Feature run a lambda that installs any endpoints the implementation requires, and have that provided by the `prometheus { }` config block...

JotBePunkt:
> But there isn't only Prometheus, there's also Dropwizard, and all the other implementations of the Micrometer facade (just like `gson { }` and `jackson { }`. If the feature's code is more or less the same for all those, it might be better to use a `ContentNegotiation`-type implementation... that was my original plan. Also I'd suppose that configs for the other types also vary...

As far as I know, Prometheus is the only metrics backend that needs a http endpoint in the application. The other implementation push their metrics to the metrics server or use an own endpoint (like JMX). 

The configs do vary but they are set in the registry instanciation e.g.

https://github.com/micrometer-metrics/micrometer/blob/127fa3265325cc894f368312ed8890b76a055d88/implementations/micrometer-registry-azure-monitor/src/main/java/io/micrometer/azuremonitor/AzureMonitorMeterRegistry.java#L55

https://github.com/micrometer-metrics/micrometer/blob/127fa3265325cc894f368312ed8890b76a055d88/implementations/micrometer-registry-atlas/src/main/java/io/micrometer/atlas/AtlasMeterRegistry.java#L53

In these cases it would be feasable to use the `MicrometerFeature` and provide a specificially configured `MeterRegistry`

> Is there any real big changes in the base-Feature implementation that would make it simpler in the way you're currently doing it?

ktor-features/ktor-metrics-micrometer-prometheus/jvm/src/io/ktor/metrics/micrometer/prometheus/installFeatureWithAuth.kt
JotBePunkt: hm... the next request would to specify the host name and now we start to reimplement the routing declaration. 
Alternative to the current implementation would be: The feature itself declares by default the endpoint under `/metrics` without any authentication, port selection etc. If you need some special routing feature you set a property `declareEndpoint = false` in the config, then you can declare the endpoint like you want it to be:
```
routing {
   port(1234) {
      authentication("superSecret") {
         //...
         get("/myMetricsPath") {
            prometheusMetrics()
          }
    }
}

**Comment by JotBePunkt** [comment]
> That might still be possible by having the Feature run a lambda that installs any endpoints the implementation requires, and have that provided internally by the `prometheus { }` config block...

can you give an example how you would configure it?

File: ktor-features/ktor-metrics-micrometer-prometheus/jvm/src/io/ktor/metrics/micrometer/prometheus/MicrometerMetricsPrometheus.kt (hunk 9 above)
Comment (cy6erGn0m) [reply ktor-features/ktor-metrics-micrometer-prometheus/jvm/src/io/ktor/metrics/micrometer/prometheus/MicrometerMetricsPrometheus.kt]: Can we reassign it to a val-property with self-explaining name?

File: ktor-features/ktor-metrics-micrometer-prometheus/jvm/src/io/ktor/metrics/micrometer/prometheus/installFeatureWithAuth.kt (hunk 8 above)
Comment (cy6erGn0m) [reply ktor-features/ktor-metrics-micrometer-prometheus/jvm/src/io/ktor/metrics/micrometer/prometheus/installFeatureWithAuth.kt]: The other solution is to provide routing entry for prometheus endpoint without any built-in auth. Like this:

```
fun Routing.prometheusEndpoint(...) { ... }

// so one can install it anywhere 
app.routing {
    route("/api") {
        prometheus { enable public metrics }
    }
    route("/admin") {
        authenticate(...) {
            // some admin actions go here
            prometheus { enable all metrics }
        }
    }
}
```

Comment (cy6erGn0m) [reply ktor-features/ktor-metrics-micrometer-prometheus/jvm/src/io/ktor/metrics/micrometer/prometheus/installFeatureWithAuth.kt]: The advantage is that you don't need to care of auth at all. This is much more composable.

**Comment by cy6erGn0m** [comment]
(conflict is caused by metrics phase name clash fix in master)

**Comment by JotBePunkt** [comment]
as discussed above, I removed the support for auth completely. I left default implementation of the metrics endpoint, which can be disabled so that the user can implement its own metricsEndpoints.
I switched the implementaiton to a pattern like the templating engines. In contrast to the Json features there is nothing to negotiate about the content type.

File: ktor-features/ktor-metrics-micrometer-prometheus/jvm/src/io/ktor/metrics/micrometer/prometheus/MicrometerMetricsPrometheus.kt [hunk 11]
```diff
@@ -0,0 +1,149 @@
...
+
+        override var distributionStatisticConfig: DistributionStatisticConfig = DistributionStatisticConfig.Builder()
+            .percentilesHistogram(true)
+            .build()
+
+        /**
+         * Defines if a metrics endpoint is installed under "/metrics" for all metrics registered at [registry].
+         * Set to false if you want to declare a custom metric endpoint instead.
+         */
+        var installDefaultEndpoint = true
```
cy6erGn0m: Instead of the flag something like `defaultEndpointRoute: String? = "/metrics"` can be used so both enable/disable and path customization will be accomplished at once.

ktor-features/ktor-metrics-micrometer-prometheus/jvm/src/io/ktor/metrics/micrometer/prometheus/MicrometerMetricsPrometheus.kt
```diff
@@ -0,0 +1,149 @@
...
+
+
+internal class MetricsContent(
+    val registry: PrometheusMeterRegistry,
+    val filter: Collector.MetricFamilySamples.() -> Boolean
+) {
+    fun filteredMetricFamilySamples(): Enumeration<Collector.MetricFamilySamples> =
+        object : Enumeration<Collector.MetricFamilySamples> {
+
+            private val internalCollection =
```
cy6erGn0m: `registry.prometheusRegistry.metricFamilySamples().iterator().asSequence().filter(filter).iterator()` could replace a lot of the following code

ktor-features/ktor-metrics-micrometer-prometheus/jvm/src/io/ktor/metrics/micrometer/prometheus/MicrometerMetricsPrometheus.kt
```diff
@@ -0,0 +1,149 @@
...
+import io.micrometer.prometheus.PrometheusMeterRegistry
+import io.prometheus.client.Collector
+import io.prometheus.client.exporter.common.TextFormat
+import kotlinx.coroutines.io.ByteWriteChannel
+import java.util.*
+
+class MicrometerMetricsPrometheus(config: Configuration) :
+    AbstractMicrometerMetrics<PrometheusMeterRegistry, MicrometerMetricsPrometheus.Configuration>(config) {
+
+    internal fun process(value: MetricsContent) = object : OutgoingContent.WriteChannelContent() {
```
cy6erGn0m: I'd prefer a separate internal class `PrometheusResponse`

ktor-features/ktor-metrics-micrometer-prometheus/jvm/src/io/ktor/metrics/micrometer/prometheus/MicrometerMetricsPrometheus.kt
```diff
@@ -0,0 +1,149 @@
...
+import io.ktor.util.cio.bufferedWriter
+import io.micrometer.core.instrument.distribution.DistributionStatisticConfig
+import io.micrometer.prometheus.PrometheusConfig
+import io.micrometer.prometheus.PrometheusMeterRegistry
+import io.prometheus.client.Collector
+import io.prometheus.client.exporter.common.TextFormat
+import kotlinx.coroutines.io.ByteWriteChannel
+import java.util.*
+
+class MicrometerMetricsPrometheus(config: Configuration) :
```
cy6erGn0m: After removing unnecessary functionality, the redundant inheritance is even more obvious. I believe prometheus should be installed into the original micrometer metrics functionality. There is no need to inherit it. We need a separate Prometheus config and a routing function to install an endpoint. `meterRegistry` extension property could return it from the original feature as well.

JotBePunkt:
Sorry, I did not have much time to continue here. My nitial idea was to enable a specific configuration option for each of the MetricRegistry implementations. Prometheus is just one of them (but the one I am actively using), other implementation needs other config options. Therefore I used the implementation with the generics, which is not the favorite of @cy6erGn0m.

Oxymoron290:
@JotBePunkt Understood, How can I help move this along. I would really like to see this feature as part of the project.

JotBePunkt:
Let me check, what I did so far. Afaik the Prometheus Implementation is kind of finished (I think we need to finish some details). What is missing are the other registry implementations for micrometer (or possibilities to configure them). I started to implement them, but its currenly in a state where it does not compile yet. What I could do is commit that to another branch and create a PR (as kind of preview, so you see what is planned).

JotBePunkt:
@Oxymoron290: Here is the continuation of the work: https://github.com/JotBePunkt/ktor/pull/2
I also asked the micrometer contributors if they like to see the configuration (Kotlin) classes in their repo, as other projects might want to profit from them too: 

https://github.com/micrometer-metrics/micrometer/issues/1607

PR #1126, opened 2019-05-14
Set timeouts, reduce secure random pressure

e5l:


ktor-client/ktor-client-tests/common/src/io/ktor/client/tests/utils/CommonClientTestUtils.kt
```diff
@@ -53,11 +53,7 @@ fun <T : HttpClientEngineConfig> clientTest(
     }
 
     try {
-        withTimeout(5000) {
```
cy6erGn0m: Why we don't need it anymore?

Issue #798, opened 2018-12-15
Digest challenge invokes browser credentials popup


sethchhim:
### Ktor Version
1.0.1

I would like to retrieve an initial digest response so that I can work with the response headers it gives. This works well but it also throws an `Unauthorized 401` response resulting in an unwanted browser popup requesting credentials. I believe this is because of the specific status of 401. Could we use a generic status instead to prevent this?

[DigestAuth.kt](https://github.com/ktorio/ktor/blob/6c724f804bd6f25158d284d05c49235c67573019/ktor-features/ktor-auth/src/io/ktor/auth/DigestAuth.kt) Line 75-78
```
context.challenge(digestAuthenticationChallengeKey, cause) {
    call.respond(UnauthorizedResponse(HttpAuthHeader.digestAuthChallenge(realm)))
    it.complete()
}
```

Thank you for any consideration.

sethchhim:
I solved this by changing `UnauthorizedResponse` to a `ForbiddenResponse`. I'm sure 401 is correct and standard practice but that browser popup is not needed for my use case. 

```
fun Authentication.Configuration.digestFix(name: String? = null, configure: DigestAuthenticationProvider.() -> Unit) {
    val provider = DigestAuthenticationProvider(name).apply(configure)
    val realm = provider.realm
    val userNameRealmPasswordDigestProvider = provider.userNameRealmPasswordDigestProvider
    val digester = provider.digester

    provider.pipeline.intercept(AuthenticationPipeline.RequestAuthentication) { context ->
        val authorizationHeader = call.request.parseAuthorizationHeader()
        val credentials = authorizationHeader?.let { authHeader ->
            if (authHeader.authScheme == AuthScheme.Digest && authHeader is HttpAuthHeader.Parameterized) {
                authHeader.toDigestCredential()
            } else
                null
        }

        val principal = credentials?.let {
            if ((it.algorithm ?: "MD5") == digester.algorithm
                && it.realm == realm
                && it.verifier(call.request.local.method, digester, userNameRealmPasswordDigestProvider))
                UserIdPrincipal(it.userName)
            else
                null
        }

        when (principal) {
            null -> {
                val cause = when (credentials) {
                    null -> AuthenticationFailedCause.NoCredentials
                    else -> AuthenticationFailedCause.InvalidCredentials
                }

                context.challenge("Digest", cause) {
                    call.respond(ForbiddenResponse(HttpAuthHeader.digestAuthChallenge(realm)))
                    it.complete()
[...]

PR #1130, opened 2019-05-14
add forbidden response (403) alternative to unauthorized response (401)

sethchhim:
I would like to receive `Forbidden Response` instead of `Unauthorized Response` in event of failure.

Reason: Unauthorized response will automatically induce the browser credential popup and in my use case is not desirable because I plan to handle it manually. There is no workaround other than to return a response that is not `401`. 

fstbx
gif courtesy of stackoverflow [issue](https://stackoverflow.com/questions/52818710/google-chrome-unauthorized-response-causes-a-popup)

My previous workaround is blocked by 1.2.0 update, so I will submit this pr.

```
    install(Authentication) {
        digest { 
            useForbiddenResponse = true
            // do stuff
        }
    }
```

Thank you for any consideration.

issue: https://github.com/ktorio/ktor/issues/798

stackoverflow1: https://stackoverflow.com/questions/9859627/how-to-prevent-browser-to-invoke-basic-auth-popup-and-handle-401-error-using-jqu/39936894


ktor-features/ktor-auth/jvm/src/io/ktor/auth/ForbiddenResponse.kt
```diff
@@ -0,0 +1,25 @@
...
+
+import io.ktor.http.content.*
+import io.ktor.http.*
+import io.ktor.http.auth.*
+
+/**
+ * Response content with `403 Forbidden` status code and `WWW-Authenticate` header of supplied [challenges]
+ * @param challenges to be passed with `WWW-Authenticate` header
+ */
+class ForbiddenResponse(vararg val challenges: HttpAuthHeader) : OutgoingContent.NoContent() {
```
cy6erGn0m: Code duplication with `UnauthorizedResponse`. Possible option could be creating 'ChallengeResponse` with status code first parameter, deprecate or inherit `UnauthorizedResponse`.

ktor-features/ktor-auth-jwt/jvm/src/io/ktor/auth/jwt/JWTAuth.kt
```diff
@@ -148,10 +149,11 @@ fun Authentication.Configuration.jwt(
     val authenticate = provider.authenticationFunction
     val verifier = provider.verifier
     val schemes = provider.schemes
+    val useForbiddenResponse = provider.useForbiddenResponse
```
cy6erGn0m: Instead of using switch options that is not composable, we usually use challenge sealed class like this one: https://github.com/ktorio/ktor/blob/9fdc5050396a9785fc39bad9cb5f8e4c0b2567c1/ktor-features/ktor-auth/jvm/src/io/ktor/auth/SessionAuth.kt#L126 
The only problem is that it is done every time for every feature and not generalized yet...

cy6erGn0m:
Also this functionality need to be tested in the corresponding tests of modified features.

cy6erGn0m:
Should be fixed as per #1178

PR #1131, opened 2019-05-15
Complete the basic-auth-test for ktor-client-auth

nilsmagnus:


cy6erGn0m:
This shouldn't be needed since [ResponseValidator](https://ktor.io/clients/http-client/features/response-validation.html) is enabled by default so any non-successful response status should cause `get` to fail without explicit assertion required.

PR #1147, opened 2019-05-22
Fix SessionTransportTransformers application order

Meztihn:
I've found that if you add 2 different transformers, session reading breaks because of incorrect transformations order. Here is simple fix with tests.


cy6erGn0m:
Thanks!

PR #1148, opened 2019-05-23
Fix module function lookup with default args (regression fix)

cy6erGn0m:
Using module functions with default arguments is broken as per 10ae206e196bf944016a3c35a28fb914c8cdfbd5 (#1082)

This is critical since already existing code may stop working.


cy6erGn0m:
Merged manually

PR #1157, opened 2019-05-28
Remove 'development' word in comments of EngineMain

inkyfox:
Remove 'development' word in comments of EngineMain so not to confuse.


cy6erGn0m:
Thanks!

Issue #1168, opened 2019-06-03
MockEngine does not merge HTTP request headers


uzzu:
### Ktor Version
1.2.1

### Ktor Engine Used(client or server and name)
MockEngine

### JVM Version, Operating System and Relevant Context
1.8, MacOSX 10.14.4

### Feedback

Since using 1.2.1 (from 1.1.2), my project's unit test become failed.
Wrote a unit test into MockEngineTests.kt and it reproduced.

```kotlin
    @Test
    fun testRequestContentType() = testBlocking {
        var actualContentType: String? = null
        val client = HttpClient(MockEngine { request ->
            actualContentType = request.headers[HttpHeaders.ContentType]
            respondOk("${request.url}")
        })

        val formDataContent = FormDataContent(Parameters.build { append("myKey", "myValue") })
        client.post<String> {
            url("http://127.0.0.1/formdata-content")
            body = formDataContent
        }

        assertEquals(actualContentType, formDataContent.contentType.toString())
    }
```

Would be better to merge HTTP request headers like other HTTPClient engines.

uzzu:
I'm trying to make this PR.

uzzu:
https://github.com/ktorio/ktor/pull/1169 is ready for review :)

PR #1169, opened 2019-06-03
Fix MockEngine request header merging

uzzu:
Addresses: #1168


e5l:
Hi @uzzu, the headers merge depends on the engine. Mock engine is used for testing, probably we shouldn't merge the headers inside.

PR #1172, opened 2019-06-03
Add a way to make daemon thread pool on jvm

### LINKED ISSUE #1170: [client] Provide a way to make jvm dispatcher is daemon
### Ktor Version
1.2.1

CIO client

Currently all jvm client engines extend HttpClientJvmEngine, which uses fixed thread pool with non-daemon threads. I want my jvm terminates by self when all my stuff done, but if I create any client it creates thread pool and block jvm from terminating.

https://github.com/ktorio/ktor/blob/e0525a274d2c9958778fb649df39d59c44341b2b/ktor-client/ktor-client-core/jvm/src/io/ktor/client/engine/HttpClientJvmEngine.kt#L18

I'm not sure if it should be a switch in a engine config or just make it as default, but it would be very useful.

Similar was reported in #543, but for server engines.

Prototik:
Closes #1170 

Need comments about keeping binary compatibility caused by extracting `threadsCount` from common sources (as it uses only on jvm it makes sense).

### DIFF
--- a/ktor-client/ktor-client-android/jvm/src/io/ktor/client/engine/android/AndroidEngineConfig.kt
+++ b/ktor-client/ktor-client-android/jvm/src/io/ktor/client/engine/android/AndroidEngineConfig.kt
@@ -12,7 +12,7 @@ import javax.net.ssl.*
 /**
  * Configuration for [Android] client engine.
  */
-class AndroidEngineConfig : HttpClientEngineConfig() {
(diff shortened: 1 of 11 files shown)


e5l:
Hi @Prototik, thanks for the PR.
The client dispatcher should be daemon by default and no option needed here.

It will be fixed in next minor release.

PR #1176, opened 2019-06-07
Client bug fixes

e5l:


ktor-utils/common/src/io/ktor/util/Lock.kt
```diff
@@ -6,14 +6,22 @@
...
-expect class Lock() {
+expect class Lock() : Closeable {
     fun lock()
     fun unlock()
 }
 
 @InternalAPI
-inline fun <R> Lock.use(block: () -> R): R {
+expect class ReadWriteLock() : Closeable {
+    fun readLock(): LockTicket
```
cy6erGn0m: These methods would be just fine, but Java's RWL has the same with different behaviour. So I'd prefer to name it differently to focus attention on the fact that it acquires the lock.

ktor-client/ktor-client-core/common/src/io/ktor/client/features/cookies/AcceptAllCookiesStorage.kt
```diff
@@ -18,16 +19,16 @@ class AcceptAllCookiesStorage : CookiesStorage {
     private val oldestCookie: AtomicLong = atomic(0L)
     private val mutex = Lock()
 
-    override suspend fun get(requestUrl: Url): List<Cookie> = mutex.use {
+    override suspend fun get(requestUrl: Url): List<Cookie> = mutex.useLocked {
```
cy6erGn0m: or, probably, `withLocked`

PR #1178, opened 2019-06-07
Several fixes and cleanup

cy6erGn0m:
- deprecations
- JVM shutdown hooks management for all server engines implementations
- custom challenge builder added to server auth providers 
- fixed internal utility function for new IR JS implementation (that breaks JS ABI)


cy6erGn0m:
Merged manually

PR #1179, opened 2019-06-09
Move Frame.Close.readReason() to common

Prototik:


cy6erGn0m:
Oh, binary compatibility is broken. What a shame. So, instead of removing the old function you need to replace it with the following:

```kotlin
@Deprecated("Binary compatibility", level = Deprecated.Level.HIDDEN)
fun Frame.Close.readReason(): CloseReason? = readReason()
```

Then, run the following:
./gradlew clean binary-compatibility-validator:jvmTest -Poverwrite.output=true

then review and commit changed validator's data

Prototik:
Some dark magic required here, but seems abi compatible now

cy6erGn0m:
Merged, thanks!

Issue #649, opened 2018-10-16
Should return 405 when route exists but not for given method instead of 404


sommd:
If you have a route, such as `/foo` with the `GET` method defined, but you make a request with another method, such as `POST`, then 404 is returned instead of [405 (Method Not Allowed)](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/405). 404 isn't really the correct status, since the resource (`/foo`) does exist, but only `GET` is supported, so 405 makes more sense. This is what most other frameworks return. If possible, it'd be nice if Ktor also generated the `Allow` header automatically.

cy6erGn0m:
Well, replying 405 will prevent other handlers/interceptors/features from handling requests since routing is just a feature. Need to think about it.

`Allow` is that for sure we need, we have PR #581 but still there are open questions about it

master-bytes-krafter:
Hello, we are also very interested by this automatic 405 feature, the workaround for now is quite clunky :(
Have you made any decisions yet ?
Regards

oleg-larshin:
Please check the <a href="https://youtrack.jetbrains.com/issue/KTOR-737">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

PR #1192, opened 2019-06-19
Use 404 status code instead of null for unhandled test requests

camhashemi:
There seems to be a discrepancy between how DefaultEngine and TestApplicationEngine handle unhandled requests:

If a call has not been assigned a status code, DefaultEnginePipeline adds 404 to the call, which makes sense, but TestApplicationEngine's pipeline doesn't modify these requests at all, leading `response.statusCode() == null` instead of `response.statusCode() == 404`, as is expected.

This can be worked around, but makes testing less clear, so it would be nice if there was some parity here.

Background: 
[DefaultEnginePipeline](https://github.com/ktorio/ktor/blob/68eef1948ce4c8c11dae794588f8d063afb67d0b/ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/DefaultEnginePipeline.kt#L35)
[TestApplicationEngine](https://github.com/ktorio/ktor/blob/68eef1948ce4c8c11dae794588f8d063afb67d0b/ktor-server/ktor-server-test-host/jvm/src/io/ktor/server/testing/TestApplicationEngine.kt#L56)

### DIFF
--- a/ktor-server/ktor-server-test-host/jvm/src/io/ktor/server/testing/TestApplicationEngine.kt
+++ b/ktor-server/ktor-server-test-host/jvm/src/io/ktor/server/testing/TestApplicationEngine.kt
@@ -54,6 +54,9 @@ class TestApplicationEngine(
         {
             try {
                 call.application.execute(call)
+                if (call.response.status() == null) {
+                    call.respond(HttpStatusCode.NotFound)
+                }
             } catch (cause: Throwable) {
                 handleTestFailure(cause)
             }


cy6erGn0m:
For sure this breaks a lot of tests in the ktor codebase. However, I've heard this request several times, so we probably need to apply it somehow to avoid confusion.

camhashemi:
Yeah, it was hard for me to see what tests failed since the output was full of noise regarding ios. But I figured it'd be easier to show than tell.

jvarela-sr:
My only question is about the use of 404 as the code for an unhandled case.  Shouldn't it be 405 instead?

camhashemi:
I believe if you run the Ktor server outside of a test setting, it will
return 404 for an unsupported path. So it's more about the test server
being consistent with the real thing.

camhashemi:
I believe Spring and Play framework have the same behavior, 404 if you send
a request to a non-existent path.

jvarela-sr:
This kind of deviates from expected REST behavior (regardless of what Spring does).  There seems to be a request to fix Ktor for this very issue.  Might want to fix testing as well. 
https://github.com/ktorio/ktor/issues/649

camhashemi:
I see what you mean. What I'm referring to is 404 behavior (no path found),
but it would apply for 405 behavior as well (no method found for path).

cy6erGn0m:
See https://youtrack.jetbrains.com/issue/KTOR-842

PR #1205, opened 2019-06-26
Bug fixes

e5l:


ktor-client/ktor-client-android/jvm/src/io/ktor/client/engine/android/AndroidClientEngine.kt
```diff
@@ -27,15 +27,15 @@ class AndroidClientEngine(override val config: AndroidEngineConfig) : HttpClient
     override suspend fun execute(
         data: HttpRequestData
     ): HttpResponseData {
-        val callContext = createCallContext()
+        val callContext: CoroutineContext = createCallContext()
         return async(callContext) {
-            val requestTime = GMTDate()
+            val requestTime: GMTDate = GMTDate()
```
cy6erGn0m: Why so many explicit types?

ktor-client/ktor-client-apache/jvm/src/io/ktor/client/engine/apache/ApacheEngine.kt
```diff
@@ -33,7 +33,13 @@ internal class ApacheEngine(override val config: ApacheEngineConfig) : HttpClien
     private fun prepareClient(): CloseableHttpAsyncClient {
         val clientBuilder = HttpAsyncClients.custom()
         with(clientBuilder) {
-            setThreadFactory { Thread(it, "Ktor-client-apache").apply { isDaemon = true } }
+            setThreadFactory {
+                Thread(it, "Ktor-client-apache").apply {
+                    isDaemon = true
+                    setUncaughtExceptionHandler { _, _ -> }
```
cy6erGn0m: Shouldn't we do something instead of ignoring exceptions?

ktor-client/ktor-client-apache/jvm/src/io/ktor/client/engine/apache/ApacheHttpRequest.kt
```diff
@@ -30,39 +30,22 @@ internal suspend fun CloseableHttpAsyncClient.sendRequest(
...
-            response.completeExceptionally(exception)
+            continuation.cancel(exception)
         }
 
         override fun completed(result: Unit) {}
 
         override fun cancelled() {
             callContext.cancel()
-            response.cancel()
+            continuation.cancel()
```
cy6erGn0m: Is that true that only one of these handlers will be called and only once?

ktor-client/ktor-client-apache/jvm/src/io/ktor/client/engine/apache/ApacheResponseConsumer.kt
```diff
@@ -4,93 +4,56 @@
...
-    }.invokeOnCompletion { cause ->
-        channel.close(cause)
+    override fun responseCompleted(context: HttpContext?) {
+        contentReader.close()
     }
 
-    private suspend fun ByteWriteChannel.writeRemaining() {
-        current.flip()
-        if (current.hasRemaining()) writeFully(current)
+    override fun getException(): Exception? = null
```
cy6erGn0m: Should we always return null here?

ktor-client/ktor-client-apache/jvm/src/io/ktor/client/engine/apache/ContentReader.kt
```diff
@@ -0,0 +1,119 @@
...
+    val content: ByteReadChannel get() = _content
+
+    private val events = ConcurrentLinkedQueue<Runnable>()
+
+    override val context: CoroutineContext = this
+
+    private val handler = suspend handler@{
+        try {
+            yield()
+            if (!state.isActive) return@handler
```
cy6erGn0m: Is it ok to not close `_content` if cancelled immediately?

cy6erGn0m: Is it ok to keep hasData = true in this case?

ktor-client/ktor-client-apache/jvm/src/io/ktor/client/engine/apache/ContentReader.kt
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    init {
+        handler.startCoroutineUninterceptedOrReturn(this)
+    }
+
+    fun consume(decoder: ContentDecoder, ioControl: IOControl) {
+        hasData = true
+        if (state.isCompleted) {
+            ioControl.shutdown()
```
cy6erGn0m: You just told that one shouldn't shutdown an io control...

ktor-client/ktor-client-apache/jvm/src/io/ktor/client/engine/apache/ContentReader.kt
```diff
@@ -0,0 +1,119 @@
...
+            return
+        }
+
+        currentDecoder = decoder
+        control = ioControl
+
+        while (hasData) {
+            runNextEvent()
```
cy6erGn0m: Can't we get an infinite loop here if no tasks in the queue?

ktor-client/ktor-client-core/common/src/io/ktor/client/response/HttpResponse.kt
```diff
@@ -68,7 +68,7 @@ abstract class HttpResponse : HttpMessage, CoroutineScope, Closeable {
         if (!closed.compareAndSet(false, true)) return
 
         launch {
-            content.discard()
+            content.cancel()
```
cy6erGn0m: So we don't wait for content transfer completion but abort it, right? Is it safe?

ktor-client/ktor-client-core/js/src/io/ktor/client/engine/js/compatible/Utils.kt
```diff
@@ -26,19 +26,38 @@ internal fun readBody(
...
+    val body = response.body ?: error("Fail to get body")
+    val responseData = Channel<ByteArray>(Channel.UNLIMITED)
+
+    body.on("data") { chunk: ArrayBuffer ->
+        responseData.offer(Uint8Array(chunk).asByteArray())
+    }
+
+    body.on("end") {
+        responseData.close()
+    }
```
cy6erGn0m: Shouldn't we listen for [error events](https://nodejs.org/api/stream.html#stream_event_error_1) as well?

ktor-client/ktor-client-tests/build.gradle
```diff
@@ -19,8 +19,8 @@ class KtorTestServer extends DefaultTask {
             def main = mainClass.getMethod("startServer")
             server = main.invoke(null)
         } catch (cause) {
-            println("[TestServer] ERROR: $cause")
-            cause.printStackTrace()
+//            println("[TestServer] ERROR: $cause")
+//            cause.printStackTrace()
```
cy6erGn0m: ?

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/FullFormTest.kt
```diff
@@ -4,50 +4,27 @@
...
-    fun testGet() = clientTest(factory) {
+    fun testGet() = clientTests {
         test { client ->
             val text = client.call {
                 url {
                     protocol = URLProtocol.HTTP
                     host = "127.0.0.1"
-                    port = serverPort
-                    encodedPath = "/hello"
+                    port = 8080
```
cy6erGn0m: Why a constant port?

ktor-client/ktor-client-android/jvm/src/io/ktor/client/engine/android/AndroidClientEngine.kt
e5l: It was slow to edit code with dev plugin :(

ktor-client/ktor-client-apache/jvm/src/io/ktor/client/engine/apache/ApacheEngine.kt
e5l: We report exceptions with our client coroutines and channels. Here we prevent to log cancellation exceptions unhandled by Apache.

ktor-client/ktor-client-tests/build.gradle
e5l: It prevents logging of address already in use(when testing locally)

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/FullFormTest.kt
e5l: We use external test server on the fixed port here

ktor-client/ktor-client-core/common/src/io/ktor/client/response/HttpResponse.kt
e5l: Yes

ktor-client/ktor-client-apache/jvm/src/io/ktor/client/engine/apache/ContentReader.kt
e5l: We close `_content` and switch `hasData` in finally block here. It should be enough.

ktor-client/ktor-client-apache/jvm/src/io/ktor/client/engine/apache/ApacheHttpRequest.kt
e5l: Yes and we checked it early.

PR #1211, opened 2019-06-28
Remove private modifier from Routing.interceptor

eld0727:
Move all inner logic of Routing to separated class to allow user use routing feature in detached mode.
Also make interceptor function public to use it independently of `install(Routing)`

Main motivation of that pull request is to allow user use dynamic routing

### DIFF
--- a/ktor-server/ktor-server-core/jvm/src/io/ktor/routing/Routing.kt
+++ b/ktor-server/ktor-server-core/jvm/src/io/ktor/routing/Routing.kt
@@ -27,7 +27,7 @@ class Routing(val application: Application) :
         tracers.add(block)
     }
 
-    private suspend fun interceptor(context: PipelineContext<Unit, ApplicationCall>) {
+    suspend fun interceptor(context: PipelineContext<Unit, ApplicationCall>) {
         val resolveContext = RoutingResolveContext(this, context.call, tracers)
         val resolveResult = resolveContext.resolve()
         if (resolveResult is RoutingResolveResult.Success) {


cy6erGn0m:
Why making `interceptor` function public is not enough? What are advantages of extra inheritance and nullable application? Do you build routing with no running application? Why?

cy6erGn0m:
Locations couldn't be registered in a routing with no application

eld0727:
> Do you build routing with no running application? Why?

I want to do something like that:
```kotlin
val applicationRef = AtomicRef(MyApplication())
val server = embeddedServer(Netty, port = 8080) {
    intercept(ApplicationCallPipeline.Call) {
        applicationRef.get().routing.intercept(this)
    }
    routing {
        route("some static routes") {
            ...
        }
    }
}

//and in some moment reload all application components atomically
applicationRef.set(newVersionOfMyApplication)
```
> Why making interceptor function public is not enough? What are advantages of extra inheritance and nullable application?

So it allows us to create detached routing object instead of  collecting `(Routing) -> Unit`

> Locations couldn't be registered in a routing with no application

Yeah, as i can see Locations feature could not be used with detached routes at this moment.
Authentication and web sockets have the same issue. 

 @cy6erGn0m  is it a blocker for detached routes?

cy6erGn0m:
So you actually need application reloading, like hot reload but with a custom trigger, right?

eld0727:
Yes, i think that hot reload with no downtime could solve this issue

eld0727:
Ok, lets just make interceptor function public

PR #1218, opened 2019-07-02
Fix apache performance; Benchmarks; Auth; Minor fixes

e5l:


ktor-client/ktor-client-apache/jvm/src/io/ktor/client/engine/apache/ApacheResponseConsumer.kt
```diff
@@ -4,61 +4,159 @@
...
+                        written(count)
+                        flush()
+                    }
+
+                    if (decoder.isCompleted) {
+                        done = true
+                    }
+                }
+
+                done = done || completed.value
```
cy6erGn0m: check for cancellation/failure

Issue #1182, opened 2019-06-11
How can I configure https connector in Jetty?


sskki:
### Ktor Version
1.2.1
### Ktor Engine Used(client or server and name)
Jetty
### JVM Version, Operating System and Relevant Context
java version "1.8.0_201"
OS: Ubuntu 18.04
### Feedback
I'm trying to use server side client certificate in Jetty as the code below. But I failed because the Jetty server uses sslConnector in applicationEngineEnvironment(), ignoring httpsConnector defined in embeddedServer().

```kotlin

import io.ktor.application.*
import io.ktor.http.*
import io.ktor.response.*
import io.ktor.routing.*
import io.ktor.server.engine.*
import io.ktor.server.jetty.*
import java.io.*
import java.security.KeyStore
import org.eclipse.jetty.server.ServerConnector
import org.eclipse.jetty.util.ssl.SslContextFactory

fun main() {
    val keyStore = KeyStore.getInstance(KeyStore.getDefaultType())
    val environment = applicationEngineEnvironment {
        module { module() }
        connector {
            host = "127.0.0.1"
            port = 8080
        }

        keyStore.load(File("build/keystore.jks").inputStream(), "changeit".toCharArray())

        sslConnector(
            keyStore, "keystore", { "changeit".toCharArray() }, { "changeit".toCharArray() }
        ) {
            host = "127.0.0.1"
            port = 8443
        }
    }

    embeddedServer(Jetty, environment) {
        configureServer = {
            val sslContextFactory = SslContextFactory().apply {
                this.keyStore = keyStore
                setKeyStorePassword("changeit")
                setKeyManagerPassword("changeit")
                this.trustStore = keyStore
                setTrustStorePassword("changeit")
                wantClientAuth = true
            }
            val httpsConnector = ServerConnector(this, sslContextFactory).apply {
                port = 8443
            }
            this.addConnector(httpsConnector)
        }
    }.start(wait = true)
}

fun Application.module(testing: Boolean = false) {
    routing {
        get("/") {
            call.respondText("HELLO WORLD!", contentType = ContentType.Text.Plain)
        }
    }
}

```

Deleting connectors  in applicationEngineEnvironment() results in only an inaccessible server. It seems that org.eclipse.jetty.server.Server.connectors is overwitten by applicationEngineEnvironment() after the configuration by embeddedServer().

[...]

PR #1233, opened 2019-07-13
Enable Jetty server to use arbitrary customized HTTP connectors.

sskki:
**Subsystem**
Server

**Motivation**
Enable Jetty server to use arbitrary customized HTTP connectors. #1182 .

**Solution**
exchange the order of `configuration.configureServer(this)` and `initializeServer(environment)`.

### DIFF
--- a/ktor-server/ktor-server-jetty/jvm/src/io/ktor/server/jetty/JettyApplicationEngineBase.kt
+++ b/ktor-server/ktor-server-jetty/jvm/src/io/ktor/server/jetty/JettyApplicationEngineBase.kt
@@ -41,8 +41,8 @@ open class JettyApplicationEngineBase(
      * Jetty server instance being configuring and starting
      */
     protected val server: Server = Server().apply {
-        configuration.configureServer(this)
         initializeServer(environment)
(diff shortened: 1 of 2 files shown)


cy6erGn0m:
Unfortunately, it may break already existing initialization blocks. I would add another config option to `JettyApplicationEngineBase.Configuration`, something like `configureServerPost` or (`configureServerAfter`) with the corresponding kdoc. Not sure if we need to rename the old one.

sskki:
Thank you for reviewing and pointing the problem. I'd take another way shown in #1182.
Changing [ServerInitializer.kt](https://github.com/ktorio/ktor/blob/master/ktor-server/ktor-server-jetty/jvm/src/io/ktor/server/jetty/ServerInitializer.kt) will
1. not change the order of `initializeServer(environment)` and `configuration.configureServer(this)`,
2. not overwrite `Server.connectors` defined in `configuration.configureServer(this)`,
3. not need another api,
4. not break existing initialization blocks, unless they wrongly contain connectors which did not take effect before the change.

I wish this will be some help.

PR #1236, opened 2019-07-18
 Introduce support for server double receive

cy6erGn0m:
**Subsystem**
Server, core + content negotiation

**Motivation**
It is quite a usual mistake that one does `call.receive` several times. Currently, there is no double receive diagnostics at all. So sometimes double receive leads to an empty result or unexpected errors. This is definitely not user-friendly behaviour. The other problem is that sometimes it is very unlikely that there is no way to receive a post form content twice. For example, the form auth provider consumes the whole form content so one is unable to receive it again in spite of that it is the first receive invocation from the user's point of view.

**Solution**
The suggested solution is to provide a feature that supports double receive. Without the feature installed, any extra receive invocation should produce an exception.

This feature provides the following:
- receiving the same content twice produces the same result for simple types such as `String`, `ByteArray`, `Parameters` and deserialized values;
- receiving a non-repeatable content such as channels, streams, `Multipart` and so on causes `RequestAlreadyConsumedException` error;
- a second receive invocation after a failed request leads to a repeated exception thrown;
- bonus: receiving a text, stream or something after a `ByteArray` received successfully should work since the byte array content is already cached.

I believe it is non-breaking change since double receive did never work properly anyway.

### DIFF
--- a/ktor-server/ktor-server-core/jvm/src/io/ktor/features/ContentNegotiation.kt
+++ b/ktor-server/ktor-server-core/jvm/src/io/ktor/features/ContentNegotiation.kt
@@ -153,7 +153,7 @@ class ContentNegotiation(val registrations: List<ConverterRegistration>,
                 val converted = suitableConverter.converter.convertForReceive(this)
                     ?: throw UnsupportedMediaTypeException(requestContentType)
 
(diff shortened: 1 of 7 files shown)


cy6erGn0m:
Merged manually.

PR #1239, opened 2019-07-20
ktor-bom module for better dependency management

Prototik:
For a large project it's a pain to keep all dependency versions in sync. To solve that problem maven (and gradle since 5.0) uses BOM (Bill of Materials) - list of all provided artifacts with their versions.

I suggest to add BOM to ktor project, as it already does [Kotlin itself](https://youtrack.jetbrains.com/issue/KT-18398) and [kotlinx-coroutines](https://github.com/Kotlin/kotlinx.coroutines/issues/1110).

After that ktor could be used more easily:

Gradle:
```gradle
dependencies {
  implementation(platform("io.ktor:ktor-bom:$ktor_version"))

  implementation("io.ktor:ktor-server-core")
  implementation("io.ktor:ktor-server-jetty") // whatever else
}
```

Maven:
```xml
<dependencyManagement>
    <dependencies>
        <dependency>
            <groupId>io.ktor</groupId>
            <artifactId>ktor-bom</artifactId>
            <version>${ktor.version}</version>
            <type>pom</type>
            <scope>import</scope>
        </dependency>
    </dependencies>
</dependencyManagement>
<dependencies>
    <dependency>
        <groupId>io.ktor</groupId>
        <artifactId>ktor-server-core</artifactId>
    </dependency>
<dependencies>
```


ktor-bom/build.gradle
```diff
@@ -0,0 +1,29 @@
...
+                    dependency(group: it.groupId, name: it.artifactId, version: it.version)
+                }
+            }
+        }
+    }
+}
+
+publishing {
+    publications {
+        maven(MavenPublication) {
```
cy6erGn0m: What does it produce? What `packaging` it will produce? artifact id and so on? maven central have quite strict requirements for pom files

ktor-bom/build.gradle
```diff
@@ -0,0 +1,29 @@
+plugins {
```
cy6erGn0m: Can't we avoid using custom plugin? I believe it may lead to serious gradle version limitations. For example, when we will need to upgrade Gradle to support newer metadata, the plugin may be not ready yet for newer versions and  it's release cycles may be too slow.

Prototik: It produce standard artifact like all others: io.ktor:ktor-bom, packaging is pom, all pom metadata handled by `gradle/publish.gradle` just like all others artifacts in the repo

ktor-bom/build.gradle
Prototik: It produce standard artifact like all others: io.ktor:ktor-bom, packaging is pom, all pom metadata handled by `gradle/publish.gradle` just like all others artifacts in the repo.
All maven central restrictions are satisfied.

ktor-bom/build.gradle
Prototik: Well, this plugin managed by spring team, and already used for bom in kotlinx-coroutines. I assume it pretty stable and doesn't cause issues with new gradle releases. It could be rewrited to pure xml stuff, but I don't think it necessary. Do you insist on removing this plugin?

Prototik:
[BOM sample, generated by this PR](https://pastebin.com/NfgFbpVz)

build.gradle
```diff
@@ -68,7 +69,8 @@ apply from: 'gradle/verifier.gradle'
  * Don't create `posix` and `darwin` sourceSets in single project.
  */
 def platforms = ["common", "jvm", "js", "posix", "darwin"]
-rootProject.ext.skipPublish = ["binary-compatibility-validator", "ktor-server-benchmarks", "ktor-client-benchmarks"]
+ext.skipPublish = ["binary-compatibility-validator", "ktor-server-benchmarks", "ktor-client-benchmarks"]
+ext.skipDefault = ["ktor-bom"]
```
cy6erGn0m: Unlike `skipPublish`, `skipDefault` is not self-explaining name. So for me, it is not clear what the _default_ is it and how is it skipped.

Prototik: Good catch, renamed to `nonDefaultProjectStructure`

PR #1253, opened 2019-07-30
Feature/url parsing

### LINKED ISSUE #1204: Url(…) doesn't parse non-HTTP-like URL strings correctly
`Url` is [supposed to](https://kotlinlang.slack.com/archives/C0A974TJ9/p1561553541222200?thread_ts=1561546449.221000&cid=C0A974TJ9) not just support HTTP-like URLs but [any kind of URLs](https://tools.ietf.org/html/rfc1738).

However the string parser seems to properly support only HTTP-like URLs:

### Example
```kotlin
Url("mailto:marc@knaup.io").toString()
Url("file:///var/www").toString()
```

### Expected
```kotlin
"mailto:marc@knaup.io"
"file:///var/www"
```

### Actual
```kotlin
"mailto://localhost/marc@knaup.io"
"file://var/www"
```

Ktor 1.2.2

karthikbhat13:
ktor-http client - UrlBuilder & UrlParser

This PR fixes string parser of url to cover all protocols, not just HTTP-like protocols.
fixes #1204

There was no logic to handle cases for the protocols like 'file' or 'mailto' where there is no host or number of slashes differ from conventional http ones. So have added the code snippets which will handle when there is no slash or more than two when bulding Url object from the string. Added control flow to handle all cases in string parser for Url.


e5l:
Hi @karthikbhat13, could you rebase the changes on `master`?

e5l:
Rebased in #2131

PR #1272, opened 2019-08-08
Add Scheduler feature

inkyfox:
Adds a new ktor server feature module for the simple task scheduler which is easily set up by DSL.

```kotlin
install(Scheduler)

schedule {
    everySecond {
        // do something at every second
    }

    everySecond(repeat = 10) {
        // do something for 10 times at every second
    }

    everyMinute {
        // do something at every minute "*:*:00"
    }

    everyMinute(second = 30) {
        // do something at every minute "*:*:30"
    }


    everyHour {
        // do something at every hour "*:00:00"
    }

    everyHour(minute = 15, second = 45) {
        // do something at every hour "*:15:45"
    }

    everyDay(hourOfDay = 14) {
        // do something at every day "14:00:00"
    }

    everyWeek(dayOfWeek = DayOfWeek.MONDAY, hourOfDay = 9) {
        // do something at every Monday "9:00:00"
    }
}
```


razvn:
It would be nice to know why is not approuved or give any feedback... 
that doesn't encourage people to contribute when seeing that they recieve no feedback and PR are closed because no review/validation is done.  At least thank him for the time he spends writing.

e5l:
Hey @razvn, it looks like it's the separate feature and not related to `Ktor`. Maybe it should be a separate library?

razvn:
Wrong choice IMHO. SpringBoot, Micronaut, Quarkus they all support natively a way to Schedule tasks. It's quite a usefull feature to have insde the framework I think... I just need it and we may have to go with other framework because of that even if the other parts of the project are using Ktor,

e5l:
I have no objections about the feature. The only doubt I have is to include it inside Ktor. I think it's nice to have a separate library, so anyone can use it with Ktor or without.

hhariri:
@inkyfox Have an idea about this. Is there a way we could reach you directly?

inkyfox:
Actually, yes, I expected the scheduler feature in SpringBoot because the server, which work like a daemon, is the best environment for the scheduler. But I agree with @e5l , it would be better that this become a separate library.
But we need to listen to the @razvn's opinion. I agree with reject of this PR, but the reason should be explained before the issue is closed.

hhariri:
@inkyfox The PR was closed by a bot. What we should do is make sure we are monitoring and responding to PRs ASAP. And we're working towards this. In regard to this specific functionality, you could provide it as a separate library which can be then used via Ktor, and we could point to this in the documentation (with potentially recommendations. Alternatively if you prefer not to, and want to contribute this code to the Ktor org, we would make it a separate repo and module.

razvn:
I'm just a user who needed that feature and was unhappy to see the PR existed but close without explanation. As always Hadi's proposal is wise and I would be happy with it. Thanks Leonid and Hadi for fallowing up these comments and of course Yongha for implementing it.

PR #1274, opened 2019-08-09
Fix multipart form header entity separator

drampelt:
**Subsystem**
Client multipart forms

**Motivation**
The entities in headers for multipart form parts should be separated by a semicolon and a space (`form-data; name=...`) not just a semicolon (`form-data;name=...`). Certain webservers (Rails in my case) can't correctly parse the headers otherwise. See: 
* https://tools.ietf.org/html/rfc7578#section-4.2 
* https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition#As_a_header_for_a_multipart_body

**Solution**
Added a space after the semicolon when building form multipart headers

### DIFF
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/request/forms/FormDataContent.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/request/forms/FormDataContent.kt
@@ -48,7 +48,7 @@ class MultiPartFormDataContent(
     private val rawParts: List<PreparedPart> = parts.map { part ->
         val headersBuilder = BytePacketBuilder()
         for ((key, values) in part.headers.entries()) {
-            headersBuilder.writeStringUtf8("$key: ${values.joinToString(";")}")
+            headersBuilder.writeStringUtf8("$key: ${values.joinToString("; ")}")
             headersBuilder.writeFully(RN_BYTES)
         }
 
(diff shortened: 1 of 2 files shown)


e5l:
Hi @drampelt, thanks for the PR.

Could you provide the server sample to add it in integration tests?

drampelt:
Could you clarify where I should be making this test? Sorry I'm not too familiar with how the tests are structured. I see that there's a `ContentTest` which does form requests but I'm not really sure where I should be testing the format of these headers.

e5l:
You can provide repo example. I'll setup the test.

drampelt:
Got it thanks! I setup a test repo here: https://github.com/drampelt/ktor-form-test

It's a simple rails app that just responds with the parameters that are passed into it. Not sure if you're familiar with rails but once you have ruby installed you should be able to run `bundle install` and `rails s` to start the server on port 3000.

I also added a ktor client sample (`src/main/kotlin/Application.kt`) that you can just run with `./gradlew run` showcasing the problem.

Here is the expected output:
```bash
# curl -X POST -F 'some_key=some value' -F 'another_key=another value' http://localhost:3000/ktor_test
{"parameters":[{"key":"some_key","value":"some value"},{"key":"another_key","value":"another value"}]}
```

And the ktor output:
```bash
# ./gradlew run
{"parameters":[{"key":"text/plain","value":["some value","another value"]}]}
```

The ktor client sample also has a quick fix that is commented out to show that the missing space is the problem.

e5l:
Merged. Thanks for the PR :)

PR #1282, opened 2019-08-12
Introduce common date parser

e5l:


ktor-http/common/src/io/ktor/http/DateUtils.kt
```diff
@@ -6,36 +6,40 @@ package io.ktor.http
...
+    for (format in HTTP_DATE_FORMATS) {
+        val date = try {
+            val parser = GMTDateParser(format)
+            parser.parse(this@fromHttpToGmtDate)
+        } catch (cause: InvalidDateString) {
+            null
+        }
+
+        if (date != null) {
+            return date
```
cy6erGn0m: Can't we simply have return inside of the try block instead?

ktor-utils/common/src/io/ktor/util/date/GMTDateParser.kt
```diff
@@ -0,0 +1,133 @@
...
+
+    var dayOfMonth: Int? = null
+    lateinit var month: Month
+    var year: Int? = null
+
+    fun build(): GMTDate = GMTDate(seconds!!, minutes!!, hours!!, dayOfMonth!!, month, year!!)
+}
+
+@Suppress("KDocMissingDocumentation")
+class InvalidDateString(
```
cy6erGn0m: `InvalidDateStringException`

ktor-utils/common/src/io/ktor/util/date/GMTDateParser.kt
```diff
@@ -0,0 +1,133 @@
...
+ * | Year     | Y            | parse year                                           |
+ * | Any char | *            | Match any character                                  |
+ *
+ * All other chars are checked to match exact pattern. If you want to exact match any of pattern chars,
+ * you should escape it with `\`:
+ *
+ * @sample: pattern "\s ss" matches string "s 15"
+ */
+@KtorExperimentalAPI
+class GMTDateParser(private val pattern: String) {
```
cy6erGn0m: Not sure if it should be open since it is definitely not a general purpose date parser (at least because of missing variable length component support, am I right?)

ktor-utils/common/src/io/ktor/util/date/GMTDateParser.kt
```diff
@@ -0,0 +1,133 @@
...
+        MONTH -> {
+            month = Month.from(chunk)
+        }
+        YEAR -> {
+            year = chunk.toInt()
+        }
+        ZONE ->
+            check(chunk == "GMT")
+        ANY -> Unit
+        else -> {
```
cy6erGn0m: What about escape character `\` ?

ktor-utils/common/src/io/ktor/util/date/GMTDateParser.kt
```diff
@@ -0,0 +1,133 @@
...
+    var hours: Int? = null
+
+    var dayOfMonth: Int? = null
+    lateinit var month: Month
+    var year: Int? = null
+
+    fun build(): GMTDate = GMTDate(seconds!!, minutes!!, hours!!, dayOfMonth!!, month, year!!)
+}
+
+@Suppress("KDocMissingDocumentation")
```
cy6erGn0m: Why suppressed?

ktor-utils/common/src/io/ktor/util/date/GMTDateParser.kt
e5l: We don't have `\` in our masks

Issue #1285, opened 2019-08-15
Retry on HttpCode


imanushin:
**Subsystem**
Client

**Is your feature request related to a problem? Please describe.**
Sometimes external service returns 500/502 errors, however next call can be successful. In this case do connection retry after a small delay.

**Describe the solution you'd like**
I'd like to have separate Ktor Client feature to automate these retries. For example, it can get handler like ```needRetry: Response -> Bool``` on input and configurable retry pauses and retry count.

**Motivation to include to ktor**
Of course retries can be implemented above the call. However client can do them better, because:
* It can preserve low-level resources (like data buffers, resolved headers, etc.)
* It already does retry logic on 401 responses

e5l:
Hi @imanushin. Thanks for the report, it's good idea.
 The retry mechanism is implemented in `HttpSend` feature, it looks easy to introduce the feature to use it.

e5l:
Sure :)

imanushin:
Hi @augustorsouza , I did not look on it.

augustorsouza:
Hello guys, I think I have an alpha version and I would love feedbacks from you :)

```kotlin

package io.ktor.client.features

import io.ktor.client.HttpClient
import io.ktor.client.HttpClientConfig
import io.ktor.client.request.HttpRequestBuilder
import io.ktor.client.request.takeFrom
import io.ktor.client.response.HttpReceivePipeline
import io.ktor.client.response.HttpResponse
import io.ktor.util.AttributeKey

typealias NeedRetryHandler = suspend (response: HttpResponse) -> Boolean

class NeedRetry(
    private val retryHandlers: List<NeedRetryHandler>
) {
    class Config {
        internal val retryHandlers: MutableList<NeedRetryHandler> = mutableListOf()

        fun needRetryHandler(block: NeedRetryHandler) {
            retryHandlers += block
        }
    }

    companion object : HttpClientFeature<Config, NeedRetry> {
        override val key: AttributeKey<NeedRetry> = AttributeKey("NeedRetry")

        override fun prepare(block: Config.() -> Unit): NeedRetry {
            val config = Config().apply(block)

            config.retryHandlers.reversed()

            return NeedRetry(config.retryHandlers)
        }

        override fun install(feature: NeedRetry, scope: HttpClient) {
            scope.receivePipeline.intercept(HttpReceivePipeline.After) {
                try {
                    val isRetryNeeded = feature.retryHandlers.map { it(context.response) }.contains(true)

[...]

PR #1310, opened 2019-09-02
Adding need retry http client feature

augustorsouza:
**Subsystem**
client

**Motivation**
Fix issue #1285  

**Solution**
Added a feature based on Logging and HttpSend, that run a user defined block to check if retry is needed.


augustorsouza:
Marked as wip, because I still need to document things up, but I wanted @e5l and @imanushin to have access to what solution I am developing :)

ktor-client/ktor-client-core/common/src/io/ktor/client/features/NeedRetry.kt
```diff
@@ -0,0 +1,59 @@
...
+
+import io.ktor.client.HttpClient
+import io.ktor.client.HttpClientConfig
+import io.ktor.client.request.HttpRequestBuilder
+import io.ktor.client.request.takeFrom
+import io.ktor.client.response.HttpReceivePipeline
+import io.ktor.client.response.HttpResponse
+import io.ktor.util.AttributeKey
+
+typealias NeedRetryHandler = suspend (response: HttpResponse) -> Boolean
```
e5l: Would be nice to have `HttpRequestBuilder` for the future request here.

augustorsouza: great! I will change... Also, a friend of mine suggeste calling this type `RetryCondition` what do you think?

e5l: `RetryCondition` looks perfect

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/features/NeedRetryTest.kt
```diff
@@ -0,0 +1,37 @@
...
+            RetryCondition {
+                retryCondition { requestBuilder, response ->
+                    retryConditionBlockCalled = true
+                    false
+                }
+            }
+        }
+
+        test { client ->
+            client.get<HttpResponse>()
```
e5l: Consider closing the response here.

ktor-client/ktor-client-core/common/src/io/ktor/client/features/NeedRetry.kt
```diff
@@ -0,0 +1,83 @@
...
+                        .map { it(requestBuilder, context.response) }
+                        .contains(true)
+
+                    if (isRetryNeeded) {
+                        context.client.execute(requestBuilder)
+                    }
+
+                    proceedWith(it)
+                } catch (cause: Throwable) {
+                    throw cause
```
e5l: It looks like the `catch` block is redundant.

ktor-client/ktor-client-core/common/src/io/ktor/client/features/NeedRetry.kt
```diff
@@ -0,0 +1,83 @@
...
+
+            return NeedRetry(config.retryHandlers)
+        }
+
+        override fun install(feature: NeedRetry, scope: HttpClient) {
+            scope.receivePipeline.intercept(HttpReceivePipeline.After) {
+                try {
+                    val requestBuilder = HttpRequestBuilder().takeFrom(context.request)
+                    val isRetryNeeded = feature.retryHandlers
+                        .map { it(requestBuilder, context.response) }
```
e5l: `map { ... }.contains(true)` -> `any { ... }`

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/features/NeedRetryTest.kt
e5l: Btw. We need the test with different two responses for first and second requests to check that we're getting a proper response instance in the end.

augustorsouza:
@e5l what do you think about those last changes ?

augustorsouza:
> @e5l what do you think about those last changes ?

Can any one review this PR please? :)

ktor-client/ktor-client-core/common/src/io/ktor/client/features/NeedRetry.kt
```diff
@@ -0,0 +1,77 @@
...
+        }
+    }
+
+    companion object : HttpClientFeature<Config, NeedRetry> {
+        override val key: AttributeKey<NeedRetry> = AttributeKey("NeedRetry")
+
+        override fun prepare(block: Config.() -> Unit): NeedRetry {
+            val config = Config().apply(block)
+
+            config.retryHandlers.reversed()
```
Thomas-Vos: Just like in [HttpCallValidator](https://github.com/ktorio/ktor/blob/master/ktor-client/ktor-client-core/common/src/io/ktor/client/features/HttpCallValidator.kt#L72-L73) this is a useless call. The function `reversed` does NOT modify the `MutableList` but returns a new `List`. So either the return value needs to be used or replace it with `reverse`. See the following for more details: https://github.com/ktorio/ktor/issues/1345

augustorsouza: agreed :) thank you

Thomas-Vos: This isn't solved yet. You just removed the reversed call. What about your comment in line 39 which says:

> Last added handler executes first

Can you confirm this is the case? Maybe with a test?

mdivkovic-speech:
@augustorsouza, @Thomas-Vos, any updates on this? This would be quite useful.

ktor-client/ktor-client-core/common/src/io/ktor/client/features/NeedRetry.kt
Thomas-Vos: This still needs to be fixed.

wkornewald:
@e5l I was trying to implement something similar, but that would also deal with other kinds of connection errors (timeout, dns lookup failures / host not found, ...). Is that even possible? I'm very confused by the pipeline model with separate request and response pipelines. Wouldn't it be better to just have a repeatable pipeline of request-response lambda functions and then you could just retry by calling nextCallback() again?

Thomas-Vos:
> @e5l I was trying to implement something similar, but that would also deal with other kinds of connection errors (timeout, dns lookup failures / host not found, ...). Is that even possible?

@wkornewald I would like that as well but was not able to figure it out. What I would like to do is execute a HTTPS request, and in case that fails (throws `SSLHandshakeException`), retry the request with HTTP. Unfortunately I was not able to find a solution.

See discussion here: https://kotlinlang.slack.com/archives/C0A974TJ9/p1582058751273000

dimartiro-py:
Any updates on this PR? I was trying to configure retry on my API but i didn't find a way yet

ktor-client/ktor-client-core/common/src/io/ktor/client/features/NeedRetry.kt
```diff
@@ -0,0 +1,74 @@
...
+            retryHandlers += block
+        }
+    }
+
+    companion object : HttpClientFeature<Config, NeedRetry> {
+        override val key: AttributeKey<NeedRetry> = AttributeKey("NeedRetry")
+
+        override fun prepare(block: Config.() -> Unit): NeedRetry {
+            val config = Config().apply(block)
+            return NeedRetry(config.retryHandlers)
```
Thomas-Vos, line 51:
```suggestion
            return NeedRetry(config.retryHandlers.reversed())
```
Should reverse the list because line 39 mentions "Last added handler executes first.".

Thomas-Vos (changes requested):

e5l:
Regarding https://youtrack.jetbrains.com/issue/KTOR-1142, we're not going to introduce new features in the core. 
Please consider creating a separate repo for that feature

PR #1318, opened 2019-09-04
Typeof support in kotlinx.serializer

e5l:


ktor-client/ktor-client-features/ktor-client-json/ktor-client-serialization/common/src/io/ktor/client/features/json/serializer/KotlinxSerializer.kt
```diff
@@ -35,6 +27,7 @@ class KotlinxSerializer(
     /**
      * Set mapping from [type] to generated [KSerializer].
      */
+    @Deprecated("[setMapper] is obsolete with 1.3.50 `typeOf` feature", level = DeprecationLevel.WARNING)
     fun <T : Any> setMapper(type: KClass<T>, serializer: KSerializer<T>) {
         @Suppress("UNCHECKED_CAST")
         mappers[type as KClass<Any>] = serializer as KSerializer<Any>
```
cy6erGn0m: Do we actually need `mappers` and `listMappers`? We probably can keep these deprecated methods empty

PR #1319, opened 2019-09-04
Bug fixes

e5l:


ktor-client/ktor-client-ios/darwin/src/io/ktor/client/engine/ios/IosClientEngineConfig.kt
```diff
@@ -11,10 +11,18 @@ import platform.Foundation.*
  * Custom [IosClientEngine] config.
  */
 class IosClientEngineConfig : HttpClientEngineConfig() {
-    internal var requestConfig: NSMutableURLRequest.() -> Unit = {}
+    /**
+     * Request configuration.
+     */
+    var requestConfig: NSMutableURLRequest.() -> Unit = {}
```
cy6erGn0m: Why is it made open? Isn't it dangerous that one can accidentally replace configure function instead of chaining?

e5l: Let's notice the difference in docs

Issue #1302, opened 2019-08-27
HTTP GET request body is not preserved with x-www-form-urlencoded content type


sjyn:
**Ktor Version and Engine Used (client or server and name)**
Ktor version: 1.2.3
Engine: Netty server

**Describe the bug**
According to the HTTP/1.x spec, a GET request is allowed to have a body. However, the `Parameters` are always empty on the request object. Additionally, inspecting the request in any middleware (i.e. `intercept` or a `ContentNegotiator` for the content type) yields an empty list of parameters.

**To Reproduce**
1. Create any HTTP GET endpoint in a ktor application and install `DefaultHeaders`.
2. Run a CURL command with any `--data-urlencoded` values and a `GET` method type to the endpoint.
3. Inspect the result of `call.receiveParameters()` or `call.receive<Parameters>()`. It will be empty despite the `GET` request from CURL containing a body. 

**Expected behavior**
A user should be able to retrieve the body of an HTTP GET request as parameters on the request.

PR #1320, opened 2019-09-04
Cy/server get body

cy6erGn0m:
**Subsystem**
server CIO, server netty

**Motivation**
Currently, we do not support GET requests with a body (it works only with Jetty/Tomcat). This leads to issues like #1302 

**Solution**
Actually, GET requests with a body are allowed by the HTTP spec so we need to support it as well in spite of the fact that we ignored it for years.


cy6erGn0m:
Merged manually.

PR #1324, opened 2019-09-05
Migrate to gradle 5.6.1

e5l:
- update gradle
- remove bintray publish plugin


gradle/publish.gradle
```diff
@@ -10,9 +9,6 @@ task emptyJar(type: Jar) {
 
 def selectArtifactId(project, type, defaultName) {
     def name = project.name
-
-    if (rootProject.ext.nonDefaultProjectStructure.contains(name)) return defaultName
```
cy6erGn0m: So for `ktor-bom` we switch from `project.artifactId` to `project.name`, right?

PR #1325, opened 2019-09-05
Cy/server fixes

cy6erGn0m:
**Subsystem**
server, ktor-serialization, `Compression`

**Motivation**
1. Since 1.2.3 we have initial `typeOf` support with ktor private serializer detection (there was no kotlinx.serialization built-in function like that). We need to avoid it.
2. Responses with the "identity" content-encoding lack `Content-Length` header and always switched to chunked transfer encoding in spite of the fact that no actual compression applied.

**Solution**
1. We need to avoid our implementation and use kotlinx.serialization instead. 
2. An encoder may predict content-length or say that it is impossible to predict (the default behaviour). The "identity" encoder always predict the length since we do not change the content.


cy6erGn0m:
Merged manually.

PR #1326, opened 2019-09-05
Use lowercase header names as per HTTP/2.

marshallpierce:
**Subsystem**
Server

**Motivation**
HTTP/2 requires that header names be lowercase.

**Solution**
Use lowercase headers. Since HTTP/1 mandates case-insensitive header names, it should be backwards compatible.

N.B. ktor doesn't build for me on Linux so I can't run the tests; waiting to see what CI says. (`java.lang.UnsatisfiedLinkError: /home/mbp/.konan/dependencies/clang-llvm-6.0.1-linux-x86-64/lib/libclang.so.6.0: libtinfo.so.5: cannot open shared object file: No such file or directory`)


cy6erGn0m:
What is the reason to make them lower-case for HTTP/1.x?

marshallpierce:
Wouldn't `HttpHeaders.Accept` be used by app code regardless of what protocol was being served?

marshallpierce:
Sigh.

PR #1331, opened 2019-09-10
FIx minor issues

e5l:


ktor-client/ktor-client-core/jvm/src/io/ktor/client/engine/HttpClientJvmEngine.kt
```diff
@@ -54,9 +55,9 @@ abstract class HttpClientJvmEngine(engineName: String) : HttpClientEngine {
     }
 
     override fun close() {
-        clientContext.complete()
+        (clientContext[Job] as CompletableJob).complete()
 
-        clientContext.invokeOnCompletion {
+        clientContext[Job]!!.invokeOnCompletion {
```
cy6erGn0m: Can't we avoid double key search?

Something like 
```
(clientContext[Job] as CompletableJob).let { job ->
    job.complete()
    job.invokeOnCompletion { ... }
}
```

ktor-network/ktor-network-tls/jvm/src/io/ktor/network/tls/extensions/SignatureAlgorithm.kt
```diff
@@ -2,12 +2,14 @@
  * Copyright 2014-2019 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
  */
 
+@file:Suppress("INVISIBLE_REFERENCE", "INVISIBLE_MEMBER")
```
cy6erGn0m: Relaxing such serious errors for the whole file... well...

ktor-utils/common/src/io/ktor/util/CoroutinesUtils.kt
```diff
@@ -29,3 +29,10 @@ internal expect inline fun <R, A>
     arg: A,
     continuation: Continuation<Unit>
 ): Any?
+
+/**
+ * Supervisor with coroutine exception handler.
```
cy6erGn0m:
```suggestion
 * Supervisor with coroutine exception handler ignoring all exceptions.
```

PR #1332, opened 2019-09-10
Cy/exceptions

cy6erGn0m:
**Subsystem**
Server

**Motivation**
Currently, we are losing unhandled exceptions in servers so they are handled by the default exception handler that is crashing an application on Android and does not log errors properly. 

**Solution**
Introduce `DefaultUnhandledExceptionHandler` that logs exceptions using the application environment logger (configurable) and apply everywhere in server engines.

### DIFF
--- a/ktor-server/ktor-server-jetty/jvm/src/io/ktor/server/jetty/JettyKtorHandler.kt
+++ b/ktor-server/ktor-server-jetty/jvm/src/io/ktor/server/jetty/JettyKtorHandler.kt
@@ -13,7 +13,6 @@ import io.ktor.util.pipeline.*
 import kotlinx.coroutines.*
 import org.eclipse.jetty.server.*
 import org.eclipse.jetty.server.handler.*
-import java.io.*
 import java.util.concurrent.*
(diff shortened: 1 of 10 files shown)


ktor-server/ktor-server-jetty/jvm/src/io/ktor/server/jetty/JettyKtorHandler.kt
```diff
@@ -32,15 +33,24 @@ internal class JettyKtorHandler(
 ) : AbstractHandler(), CoroutineScope {
     private val environmentName = environment.connectors.joinToString("-") { it.port.toString() }
     private val queue: BlockingQueue<Runnable> = LinkedBlockingQueue()
-    private val executor = ThreadPoolExecutor(configuration.callGroupSize, configuration.callGroupSize * 8, 1L, TimeUnit.MINUTES, queue) { r ->
+    private val executor = ThreadPoolExecutor(
+        configuration.callGroupSize,
+        configuration.callGroupSize * 8,
+        1L,
```
e5l: magic

PR #1336, opened 2019-09-12
Add support for additional params to locations.href

LEDfan:
**Subsystem**
`ktor-locations`

**Motivation**
In the http API of the application I'm working on you can specify additional query parameters which I can't provide in the class annotated with `@Location`.
For example, you can add a `formatting` parameter to the query strings, but this can also be specified using the `Accept` header, hence I don't want to specify a (wrong) default in the location class.

**Solution**
I added an extra function `href` which gets a list of pairs with query parameters, which will be added to the url.
I wasn't able to do this by creating an extension method, hence this PR.

Is this something you want to add? If not, please let me know so I can look for another solution.
If you like this I will open a PR for the docs (https://github.com/ktorio/ktorio.github.io/blob/master/servers/features/locations.md#building-urls)

### DIFF
--- a/ktor-features/ktor-locations/jvm/src/io/ktor/locations/Locations.kt
+++ b/ktor-features/ktor-locations/jvm/src/io/ktor/locations/Locations.kt
@@ -226,6 +226,20 @@ open class Locations(private val application: Application, private val routeServ
             ""
     }
 
+    /**
+     * Constructs the url for [location] with additional query parameters.
+     *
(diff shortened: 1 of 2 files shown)


rsinukov:
Hi! Thanks for your contribution and sorry for not coming back to you earlier. 
Unfortunately, we will reject this PR. We want to keep locations in a way that you can transform URL to location object and back and get the original URL. This change breaks such behavior.

PR #1342, opened 2019-09-13
finishTasksAndInvalidate after executing a request (#1341).

### LINKED ISSUE #1341: iOS client engine creates a new NSURLSession for each request, which can result in failures on real devices
1.2.4, iOS

**Describe the bug**
There seems to be an internal limit on the total number of NSURLSessions allowed per app; it may differ depending on model or OS version, but the limit is definitely around 650 on an iPhone 8 running iOS 11. The following logs appear when the limit is exceeded, and the request fails:

```
nw_path_evaluator_request_nexus_and_stats NECP_CLIENT_ACTION_REQUEST_NEXUS_INSTANCE [28: No space left on device]

nw_endpoint_flow_setup_channel [520.1 2600:9000:2145:4400:7:307d:92c0:21.443 in_progress channel-flow (satisfied)] failed to request nexus instance: Error Domain=kNWErrorDomainPOSIX Code=28 "No space left on device" UserInfo={NSDescription=No space left on device}
```

We have also reproduced the bug on an iPhone X. It does _not_ reproduce in the simulator.

**To Reproduce**
1. Make 650+ requests on a real device
2. Observe the subsequent errors

**Expected behavior**
Requests should continue to succeed as usual.

6bangs:
**Subsystem**
iOS engine client

**Motivation**
Fixes #1341

**Solution**
Since a new session is used for each request, we can safely allow the session to be invalidated after the request has completed.

### DIFF
--- a/ktor-client/ktor-client-ios/darwin/src/io/ktor/client/engine/ios/IosClientEngine.kt
+++ b/ktor-client/ktor-client-ios/darwin/src/io/ktor/client/engine/ios/IosClientEngine.kt
@@ -123,6 +123,8 @@ internal class IosClientEngine(override val config: IosClientEngineConfig) : Htt
 
             config.requestConfig(nativeRequest)
             session.dataTaskWithRequest(nativeRequest).resume()
+        }.invokeOnCompletion {
+            session.finishTasksAndInvalidate()
         }
     }
 


ktor-client/ktor-client-ios/darwin/src/io/ktor/client/engine/ios/IosClientEngine.kt
```diff
@@ -123,6 +123,7 @@ internal class IosClientEngine(override val config: IosClientEngineConfig) : Htt
 
             config.requestConfig(nativeRequest)
             session.dataTaskWithRequest(nativeRequest).resume()
+            session.finishTasksAndInvalidate()
```
e5l: The session is in progress after `resume()` call. Maybe we should call this in `invokeOnCompletion` of call context?

6bangs: Good idea; updated.

e5l:
Nice, thanks for the PR. Merged in `e5l/develop` for testing

PR #1348, opened 2019-09-16
Cy/websocket close to cancellation migration (wip)

cy6erGn0m:
**Subsystem**
Server+client, ktor-websockets

**Motivation**
Currently, we are always closing an outgoing WebSocket channel at the receive side that causes `ClosedSendChannelException` at a send attempt that leads to a lot of logged errors with coroutines `1.3.0+`.

**Solution**
Instead, we should only close an outgoing channel when a WS handler does WebSocket close with `CloseReason`. Otherwise, we need to do a cancellation, just the same as `actor` coroutines do. Unfortunately, due to missing channel resources cleanup in kx.coroutines, we can't just cancel a channel. However, we still can close it with a proper cancellation exception to emulate `cancel()` since the only difference between `cancel()` and `close(CancellationException(...))` is that the first discards all messages, that is not desired.

### DIFF
--- a/ktor-client/ktor-client-core/js/src/io/ktor/client/features/websocket/JsWebSocketSession.kt
+++ b/ktor-client/ktor-client-core/js/src/io/ktor/client/features/websocket/JsWebSocketSession.kt
@@ -5,7 +5,6 @@
 package io.ktor.client.features.websocket
 
 import io.ktor.http.cio.websocket.*
-import io.ktor.util.*
 import kotlinx.coroutines.*
 import kotlinx.coroutines.channels.*
 import io.ktor.utils.io.core.*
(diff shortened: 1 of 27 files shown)


ktor-client/ktor-client-core/js/src/io/ktor/client/features/websocket/JsWebSocketSession.kt
```diff
@@ -86,28 +88,21 @@ internal class JsWebSocketSession(
...
-                websocket.close(CloseReason.Codes.UNEXPECTED_CONDITION.code, "Client failed")
+                websocket.close(CloseReason.Codes.INTERNAL_ERROR.code, "Client failed")
             }
         }
     }
 
     override suspend fun flush() {
     }
 
+    @Deprecated(
```
e5l: Consider adding level explicit

cy6erGn0m:
Merged

Issue #1351, opened 2019-09-18
Add an option to disable URL Encoding


uharaqo:
**Subsystem**
ktor-client, ktor-http/common

**Is your feature request related to a problem? Please describe.**
ktor-client cannot be used for URLs which are not supported by the URL Encoding format. 

I failed to replace an HTTP client with ktor-client just because we needed to call a URL like `/api?k1=v1;k1=v2` (non-encoded `;` instead of `%3B`). As far as I checked, there's no way to bypass URL Encoding in ktor-client.

In other libraries, typically we can set parameters just like a simple `Map<String, List<String>>`. I agree that it's better to always use a sanitized request object, but it would be too strict not to have an option to disable encoding.

**Describe the solution you'd like**
I opened [a PR to add an option to bypass the URL Encoding](https://github.com/ktorio/ktor/pull/1352) when building a URL.

**Motivation to include to ktor**
I believe that users of ktor-client shouldn't be blocked by the implementation of a server URL because they might not have control over it.

(Please let me know if there's an option to resolve the above problem.)

kevincianfarini:
I agree. In order for me to use ktor-client, I need URL encoding to have an option to be disabled. I would much rather use ktor over retrofit for my use case.

kevincianfarini:
This is my use case for needed this, as commented on the PR. 


> 
> > The only question here: why do we disable escaping at all? The format you mentioned is not URL parameters at all so the only you need is to specify raw query string without parameters. Disabling escaping for sure will lead to security issues and strange bugs.
> 
> My personal use case is that the [bittorrent protocal specifies that data encoded in the URL must be in binary format when making announce calls](https://wiki.theory.org/index.php/BitTorrentSpecification#Tracker_Request_Parameters). It's a binary protocall after all.
> 
> > Note that all binary data in the URL (particularly info_hash and peer_id) must be properly escaped. This means any byte not in the set 0-9, a-z, A-Z, '.', '-', '_' and '~', must be encoded using the "%nn" format, where nn is the hexadecimal value of the byte. (See [RFC1738](http://www.faqs.org/rfcs/rfc1738.html) for details.)
> 
> > For a 20-byte hash of \x12\x34\x56\x78\x9a\xbc\xde\xf1\x23\x45\x67\x89\xab\xcd\xef\x12\x34\x56\x78\x9a.
> 
> > The right encoded form is %124Vx%9A%BC%DE%F1%23Eg%89%AB%CD%EF%124Vx%9A
> 
[...]

PR #1352, opened 2019-09-18
Add an option to disable URL Encoding

uharaqo:
**Subsystem**
ktor-http/common

**Motivation**
ktor-client cannot be used for URLs which are not supported by the URL Encoding format. It's because ktor-client always uses `URLBuidler` and there's no way to bypass the query encoding.

This PR adds an option to disable URL Encoding via `ParametersBuilder` (which can be passed to `URLBuidler`) without breaking changes.

(I mentioned in [this issue](https://github.com/ktorio/ktor/issues/1351) that `/api?k1=v1;k1=v1` (non-encoded `;`) is preventing us to use ktor-client. We needed to switch to another library just because of one character in the URL.)

**Solution**
`formUrlEncodeTo()` method is doing actual encoding. Added an option to bypass the encoding to `ParametersBuilder` and passed it to the method.

- Created `UrlEncodingOption` enum which provides encoding options (`DEFAULT`, `KEY_ONLY`, `VALUE_ONLY`, `NO_ENCODING`)
- Added the option to `ParametersBuilder` constructor
- Added the option to `Parameters` interface as a method and to `ParametersImpl` as a private property
- Changed `formUrlEncodeTo()` method to do URL Encoding only when it's enabled by the option
- The option added to methods and interfaces are defaulted to `DEFAULT` (i.e. keys and values are encoded) so as not to break existing code

### DIFF
--- a/ktor-http/common/src/io/ktor/http/HttpUrlEncoded.kt
+++ b/ktor-http/common/src/io/ktor/http/HttpUrlEncoded.kt
@@ -12,7 +12,7 @@ import io.ktor.utils.io.charsets.*
  * Keys and values are encoded only when [encodeKey] and [encodeValue] are `true` respectively.
  */
 @KtorExperimentalAPI
-enum class UrlEncodingOption(internal val encodeKey: Boolean, internal val encodeValue: Boolean) {
+public enum class UrlEncodingOption(internal val encodeKey: Boolean, internal val encodeValue: Boolean) {
     DEFAULT(true, true),
     KEY_ONLY(true, false),
     VALUE_ONLY(false, true),
@@ -47,8 +47,9 @@ public fun String.parseUrlEncodedParameters(defaultEncoding: Charset = Charsets.
 public fun List<Pair<String, String?>>.formUrlEncode(): String = formUrlEncode(UrlEncodingOption.DEFAULT)
 
 /**
- * Encode form parameters from a list of pairs by using [option]
+ * Encode form parameters from a list of pairs
  */
+@KtorExperimentalAPI
 public fun List<Pair<String, String?>>.formUrlEncode(option: UrlEncodingOption = UrlEncodingOption.DEFAULT): String =
     buildString { formUrlEncodeTo(this, option) }
 
@@ -56,11 +57,13 @@ public fun List<Pair<String, String?>>.formUrlEncode(option: UrlEncodingOption =
  * Encode form parameters from a list of pairs to the specified [out] appendable
  */
 @Deprecated("Binary compatibility", level = DeprecationLevel.HIDDEN)
-public fun List<Pair<String, String?>>.formUrlEncodeTo(out: Appendable) = formUrlEncodeTo(out, UrlEncodingOption.DEFAULT)
+public fun List<Pair<String, String?>>.formUrlEncodeTo(out: Appendable): Unit =
+    formUrlEncodeTo(out, UrlEncodingOption.DEFAULT)
 
 /**
  * Encode form parameters from a list of pairs to the specified [out] appendable by using [option]
  */
+@KtorExperimentalAPI
 public fun List<Pair<String, String?>>.formUrlEncodeTo(
     out: Appendable,
     option: UrlEncodingOption = UrlEncodingOption.DEFAULT
@@ -82,7 +85,7 @@ public fun List<Pair<String, String?>>.formUrlEncodeTo(
  */
 public fun Parameters.formUrlEncode(): String = entries()
     .flatMap { e -> e.value.map { e.key to it } }
-    .formUrlEncode(urlEncodingOption())
+    .formUrlEncode(urlEncodingOption)
 
 /**
  * Encode form parameters to the specified [out] appendable
@@ -92,11 +95,19 @@ public fun Parameters.formUrlEncodeTo(out: Appendable) {
 }
 
 internal fun ParametersBuilder.formUrlEncodeTo(out: Appendable) {
-    entries().formUrlEncodeTo(out)
+    entries().formUrlEncodeTo(out, urlEncodingOption)
 }
 
-internal fun Set<Map.Entry<String, List<String>>>.formUrlEncodeTo(out: Appendable) {
+@Deprecated("Binary compatibility", level = DeprecationLevel.HIDDEN)
+internal fun Set<Map.Entry<String, List<String>>>.formUrlEncodeTo(out: Appendable) =
+    formUrlEncodeTo(out, UrlEncodingOption.DEFAULT)
+
+@KtorExperimentalAPI
+internal fun Set<Map.Entry<String, List<String>>>.formUrlEncodeTo(
+    out: Appendable,
+    option: UrlEncodingOption = UrlEncodingOption.DEFAULT
+) {
     flatMap { (key, value) ->
         if (value.isEmpty()) listOf(key to null) else value.map { key to it }
-    }.formUrlEncodeTo(out, urlEncodingOption)
+    }.formUrlEncodeTo(out, option)
 }
--- a/ktor-http/common/src/io/ktor/http/HttpUrlEncoded.kt
+++ b/ktor-http/common/src/io/ktor/http/HttpUrlEncoded.kt
@@ -43,11 +43,7 @@ public fun String.parseUrlEncodedParameters(defaultEncoding: Charset = Charsets.
 /**
  * Encode form parameters from a list of pairs
  */
-@Deprecated(
-    "Use formUrlEncode(option)",
-    ReplaceWith("formUrlEncode(UrlEncodingOption.DEFAULT)", "io.ktor.http.UrlEncodingOption"),
-    DeprecationLevel.HIDDEN
-)
+@Deprecated("Binary compatibility", level = DeprecationLevel.HIDDEN)
 public fun List<Pair<String, String?>>.formUrlEncode(): String = formUrlEncode(UrlEncodingOption.DEFAULT)
 
 /**
@@ -59,11 +55,7 @@ public fun List<Pair<String, String?>>.formUrlEncode(option: UrlEncodingOption =
 /**
  * Encode form parameters from a list of pairs to the specified [out] appendable
  */
-@Deprecated(
-    "Use formUrlEncodeTo(out, option)",
-    ReplaceWith("formUrlEncodeTo(out, UrlEncodingOption.DEFAULT)", "io.ktor.http.UrlEncodingOption"),
-    DeprecationLevel.HIDDEN
-)
+@Deprecated("Binary compatibility", level = DeprecationLevel.HIDDEN)
 public fun List<Pair<String, String?>>.formUrlEncodeTo(out: Appendable) = formUrlEncodeTo(out, UrlEncodingOption.DEFAULT)
 
 /**
--- a/ktor-http/common/src/io/ktor/http/HttpUrlEncoded.kt
+++ b/ktor-http/common/src/io/ktor/http/HttpUrlEncoded.kt
@@ -41,12 +41,30 @@ public fun String.parseUrlEncodedParameters(defaultEncoding: Charset = Charsets.
 }
 
 /**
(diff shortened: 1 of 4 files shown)


ktor-http/common/src/io/ktor/http/HttpUrlEncoded.kt
```diff
@@ -29,25 +41,29 @@ fun String.parseUrlEncodedParameters(defaultEncoding: Charset = Charsets.UTF_8,
 }
 
 /**
- * Encode form parameters from a list of pairs
+ * Encode form parameters from a list of pairs by using [option]
  */
-fun List<Pair<String, String?>>.formUrlEncode(): String = StringBuilder().apply {
```
Prototik: Breaking ABI change, please keep original function and annotate it with `@Deprecated` with hidden visibility to fix that.

ktor-http/common/src/io/ktor/http/HttpUrlEncoded.kt
```diff
@@ -29,25 +41,29 @@ fun String.parseUrlEncodedParameters(defaultEncoding: Charset = Charsets.UTF_8,
...
+fun List<Pair<String, String?>>.formUrlEncode(option: UrlEncodingOption = UrlEncodingOption.DEFAULT): String =
+    StringBuilder().apply {
+        formUrlEncodeTo(this, option)
+    }.toString()
 
 /**
- * Encode form parameters from a list of pairs to the specified [out] appendable
+ * Encode form parameters from a list of pairs to the specified [out] appendable by using [option]
  */
-fun List<Pair<String, String?>>.formUrlEncodeTo(out: Appendable) {
```
Prototik: Same breaking ABI change

ktor-http/common/src/io/ktor/http/Parameters.kt
```diff
@@ -10,6 +10,12 @@ import io.ktor.util.*
  * Represents HTTP parameters as a map from case-insensitive names to collection of [String] values
  */
 interface Parameters : StringValues {
+    /**
+     * Returns a [UrlEncodingOption] instance
+     */
+    @KtorExperimentalAPI
+    fun urlEncodingOption(): UrlEncodingOption = UrlEncodingOption.DEFAULT
```
Prototik: Shouldn't it be a property instead of function?

ktor-http/common/src/io/ktor/http/Parameters.kt
```diff
@@ -27,11 +33,14 @@ interface Parameters : StringValues {
 }
 
 @Suppress("KDocMissingDocumentation")
-class ParametersBuilder(size: Int = 8) : StringValuesBuilder(true, size) {
```
Prototik: Breaking ABI change, secondary constructor helps here

ktor-http/common/src/io/ktor/http/Parameters.kt
```diff
@@ -74,7 +83,11 @@ fun parametersOf(vararg pairs: Pair<String, List<String>>): Parameters = Paramet
 
 @Suppress("KDocMissingDocumentation")
 @InternalAPI
-class ParametersImpl(values: Map<String, List<String>> = emptyMap()) : Parameters, StringValuesImpl(true, values) {
```
Prototik: Breaking ABI change

ktor-http/common/src/io/ktor/http/HttpUrlEncoded.kt
```diff
@@ -29,25 +41,29 @@ fun String.parseUrlEncodedParameters(defaultEncoding: Charset = Charsets.UTF_8,
...
 
 /**
- * Encode form parameters from a list of pairs
+ * Encode form parameters from a list of pairs by using [option]
  */
-fun List<Pair<String, String?>>.formUrlEncode(): String = StringBuilder().apply {
-    formUrlEncodeTo(this)
-}.toString()
+fun List<Pair<String, String?>>.formUrlEncode(option: UrlEncodingOption = UrlEncodingOption.DEFAULT): String =
+    StringBuilder().apply {
```
Prototik: Notabug, but that could be replaced by `buildString { ... }` for cleaner code

ktor-http/common/src/io/ktor/http/Parameters.kt
uharaqo: I thought that adding a variable to this interface would break compatibility since all implementation classes should override the variable.

uharaqo:
> Please fix breaking ABI and run `./gradlew -Poverwrite.output=true :binary-compatibility-validator:jvmTest` to ensure that all original methods exists.

Thanks for reviewing this @Prototik . Could you review this again?
I've recovered the original code to keep binary compatibility, and updated the file in `binary-compatibility-validator/reference-public-api`.

ktor-http/common/src/io/ktor/http/Parameters.kt
Prototik: You can define default values for properties too:
```kotlin
val urlEncodingOption: UrlEncodingOption get() = UrlEncodingOption.DEFAULT
```

ktor-http/common/src/io/ktor/http/HttpUrlEncoded.kt
```diff
@@ -31,23 +43,45 @@ fun String.parseUrlEncodedParameters(defaultEncoding: Charset = Charsets.UTF_8,
...
  */
-fun List<Pair<String, String?>>.formUrlEncode(): String = StringBuilder().apply {
-    formUrlEncodeTo(this)
-}.toString()
+@Deprecated(
+    "Use formUrlEncode(option)",
+    ReplaceWith("formUrlEncode(UrlEncodingOption.DEFAULT)", "io.ktor.http.UrlEncodingOption"),
+    DeprecationLevel.HIDDEN
+)
+fun List<Pair<String, String?>>.formUrlEncode(): String = formUrlEncode(UrlEncodingOption.DEFAULT)
```
Prototik: Few notes about `Deprecated` here and other places:
1. Change message here to "Binary compatibility"
2. Drop ReplaceWith, it will never used

ktor-http/common/src/io/ktor/http/Parameters.kt
```diff
@@ -75,6 +88,14 @@ fun parametersOf(vararg pairs: Pair<String, List<String>>): Parameters = Paramet
 @Suppress("KDocMissingDocumentation")
 @InternalAPI
 class ParametersImpl(values: Map<String, List<String>> = emptyMap()) : Parameters, StringValuesImpl(true, values) {
+    private var option: UrlEncodingOption = super.urlEncodingOption
```
Prototik: Looks like a hack for me.

Move `urlEncodingOption` to primary constructor and provide a secondary constructor to keep abi (the same applies to `ParametersBuilder` too).

All other things LGFM

uharaqo: Now the code looks straightforward to me 🙂 Thanks a lot for many useful advice!

cy6erGn0m:
The only question here: why do we disable escaping at all? The format you mentioned is not URL parameters at all so the only you need is to specify raw query string without parameters. Disabling escaping for sure will lead to security issues and strange bugs.

uharaqo:
I totally agree with you. Raw query string would be much better. Actually I tried to add raw query support, but I couldn't come up with any solution. If you have any ideas, please let me know.
These are the restrictions which prevented to me to go forward:

- actual query string is built from the `Parameters` in the format like `encodedKey=encodedVal&...`
- `Parameters` wraps `Map<String, List<String>>` and built from `ParametersBuilder`
- `ParametersBuilder` is exposed as a property of `URLBuilder` which is used everywhere

By the way, it seems that `?k1=v1;k1=v2` is a valid query.
(I spent some time to check URL and query specifications. Good discussions in [here](https://stackoverflow.com/questions/3481664/semicolon-as-url-query-separator). The syntax for `query` is defined in [here](https://tools.ietf.org/html/rfc3986#section-3.4). `query` can be anything as long as they are composed of valid characters. It doesn't need to be key/value pairs. W3C defines the `k1=v1&k2=v2` syntax in [here](https://www.w3.org/TR/2014/REC-html5-20141028/forms.html#url-encoded-form-data), but that is specific to `application/x-www-form-urlencoded`. It seems that the well known 'query param' syntax is just a convention, which is surprising to me. For example, W3C used to recommend `;` as a separator instead of `&`.)

So, the answer to this question `why do we disable escaping at all?` would be:
- Since query param syntax is not defined in the URL spec, raw query string should be supported and query param support should be an additional feature
- But raw query support requires many changes. I don't think those edge case scenarios deserve the effort
- Letting users take care of the encoding looks the easiest and reasonable solution

I hope this would help other developers who need to call legacy APIs...

kevincianfarini:
> The only question here: why do we disable escaping at all? The format you mentioned is not URL parameters at all so the only you need is to specify raw query string without parameters. Disabling escaping for sure will lead to security issues and strange bugs.

My personal use case is that the [bittorrent protocal specifies that data encoded in the URL must be in binary format when making announce calls](https://wiki.theory.org/index.php/BitTorrentSpecification#Tracker_Request_Parameters). It's a binary protocall after all. 

So, when I make the following request
```kotlin
        val response = client.get<ByteArray>(announceUrl) {
            url {
                parameters.append("uploaded", uploaded.toString())
                parameters.append("downloaded", downloaded.toString())
                parameters.append("compact", if (acceptCompact) "1" else "0")
                parameters.
                parameters.append(
                        "info_hash",
                        infoHash.chunked(2).joinToString("") { "%$it" }
                )
                parameters.append("peer_id", peerId)
                parameters.append("port", port.toString())
                parameters.append("left", "${torrentSize - downloaded}")
            }
        }
```

I expect that the `%` characters I put in the URl have a way to avoid being encoded. I expect my call to look something like this

```
GET /announce?info_hash=%65%14%5e%d4%d7%45%cf%c9%3f%5f%fe%34%92%e9%cd%e5%4b%55%7d%9f&amp; peer_id=-CR0001-012345678901&amp; uploaded=0&amp; downloaded=0&amp; left= 2082816000&amp; port=6889&amp; compact=1 HTTP/1.1
Host: torrent.ubuntu.com
User-Agent: PostmanRuntime/7.19.0
Accept: */*
Cache-Control: no-cache
Postman-Token: 1c54b245-137c-470d-93fc-72b0edaf83e3,443df566-cf19-4c7c-a625-5c24997f6f5b
Host: torrent.ubuntu.com
Accept-Encoding: gzip, deflate
Connection: keep-alive
cache-control: no-cache
```

but instead I get this. 

```
https://torrent.ubuntu.com/announce?uploaded=0&downloaded=0&compact=1&info_hash=%2565%2514%255e%25d4%25d7%2545%25cf%25c9%253f%255f%25fe%2534%2592%25e9%25cd%25e5%254b%2555%257d%259f&peer_id=-CR0001-012345678901&port=0&left=2082816000
```

notice here that the `%` have been encoded. I see no way around this without having a way to disable URL encoding.

kevincianfarini:
Is there any more discussion on this that needs to happen?

I've since moved to retrofit which allows optional encoding via `@Query("info_hash", encoded = true) infoHash: String` where `encoded` specifies that the user has already URL encoded the value. 

Retrofit is heavy handed for my use case and for the time being limits multiplatform capabilities.

kevincianfarini:
@uharaqo could you rebase this branch on master? @e5l said he'd be able to review it :)

jschneidereit:
I'm affected by this as well in a pretty unpleasant way, and am wondering about getting this added 🤔

klausner17:
Some news about this issue?

uharaqo:
I've rebased this branch on the latest master. 

Please let me know if I need to change the code. I stopped using Ktor because of this issue, but updating this again because I received messages from some people who faced similar problems.

hhariri:
https://youtrack.jetbrains.com/issue/KTOR-1717

e5l (approved):
LGTM

e5l:
Thanks for the PR! Merged.

PR #1355, opened 2019-09-23
Problem: partial content without the passed range doesn't have content length

adelnizamutdinov:
**Subsystem**
Server Core

**Motivation**
Partial content without the passed range doesn't have content length. But it should

**Solution**
override ContentLength in PartialContent.Bypass


cy6erGn0m:
Nice catch, thanks!

Issue #1300, opened 2019-08-26
TestApplicationEngine.hookRequests loses old handler's state


LoneEngineer:
Ktor-1.2.3 (ktor-server-test-host-1.2.3.jar)

** Description **
Seems like a bug:
```
@KtorExperimentalAPI`
    fun hookRequests(
        processRequest: TestApplicationRequest.(setup: TestApplicationRequest.() -> Unit) -> Unit,
        processResponse: TestApplicationCall.() -> Unit,
        block: () -> Unit
    ) {
        val oldProcessRequest = processRequest    /// <<<----------------------- HERE
        val oldProcessResponse = processResponse  /// <<<----------------------- HERE
        this.processRequest = {
            oldProcessRequest {
                processRequest(it)
            }
        }
```
there is a clash in names between this.processRequest and function's argument processRequest and it resolved to function's argument. So instead preserving old request processor (and restore later) new processor is stored.
I found this behavior with using cookieSession {} extention:

```
  cookiesSession { some requests here }
  handleRequest(...).apply { <<<--- still contains cookies
  }

```

**Expected behavior**
I expected that 'hookRequest' will restore original set of processors (without cookie handling).

LoneEngineer:
Just curios, why issue is not marked as 'closed' ?

oleg-larshin:
Please check the <a href="https://youtrack.jetbrains.com/issue/KTOR-567">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

PR #1357, opened 2019-09-25
referencing private processors when storing old state

alxgarcia:
**Subsystem**
Server, ktor-server/ktor-server-test-host

**Motivation**
There is a clash in names between this.processRequest and function's argument processRequest and it resolved to function's argument (bug #1300)

**Solution**
`hookRequest` will restore the original set of processors

Kudos to @LoneEngineer


cy6erGn0m:
Thanks! Any chance to write a test for that?

alxgarcia:
> Thanks! Any chance to write a test for that?

Sure!

PR #1360, opened 2019-09-26
[SECURITY] Forbid the use of the sample key in production

JLLeitschuh:
If the Ktor documentation is going to provide an example with a default value, it's only responsible to forbid the use of that default value in the actual codebase.

**Subsystem**
Session

**Motivation**
Developers have a bad habit of copy and pasting sample code without understanding the security implications of their decisions. This will prevent this from happening with the default signing key used for signing sessions.

**Solution**
Black list the keys used in the sample documentation.


cy6erGn0m:
Good idea, but how do we run samles then? :D

JLLeitschuh:
You could put in an environment variable to disable this check?

JLLeitschuh:
> You could put in an environment variable to disable this check?

Would you like me to move forward with this idea?

PR #1364, opened 2019-09-27
Add support for compiling with curl 7.66.0

JLLeitschuh:
**Subsystem**
Curl client.

**Motivation**
I wanted to test the new curl support, but brew pulls down the latest version (7.66) so the build breaks. This adds support for it.

**Solution**
Add the new header location.


jameshilliard:
The current approach for setting curl include paths seems extremely fragile, one can use the command `brew --prefix curl` to programmatically detect the correct curl include path instead.

JLLeitschuh:
That assumes that brew is installed though unfortunately.

jameshilliard:
Yes, the command requires brew to be installed, but since `/usr/local/Cellar` is a homebrew specific prefix one can replace all the hardcoded `/usr/local/Cellar` prefixed includes with a dynamic path generated from `brew --prefix curl`, you'll probably need to just check that you actually have `brew` installed and that you get a valid path from `brew --prefix curl` before using it for generating the full include path.

e5l:
Hi, @JLLeitschuh. Thanks for the RP. Sorry, I didn't recognize it early, but the same fix is already in master.

PR #1370, opened 2019-09-30
Introduce TestHttpClient shortcut in a test application

cy6erGn0m:
**Subsystem**
server, client, tests

**Motivation**
Currently, creating a test client in a test application is not that easy and counterintuitive. Sometimes, it is important to do create a test client rather than use `handleRequest`. For example, if you need relatively complex features such as serialization that can't be installed into `handleRequest` function.

**Solution**
Introduce a function and a property capturing test application server engine and providing a connected test client engine.

### DIFF
--- a/ktor-features/ktor-auth/jvm/test/io/ktor/tests/auth/OAuth1a.kt
+++ b/ktor-features/ktor-auth/jvm/test/io/ktor/tests/auth/OAuth1a.kt
@@ -404,10 +404,11 @@ private fun createOAuthServer(server: TestingOAuthServer): HttpClient {
 
         }
     }
-    val engine = TestApplicationEngine(environment)
-    engine.start()
-    return HttpClient(TestHttpClientEngine.config { app = engine }) {
-        expectSuccess = false
+    with(TestApplicationEngine(environment)) {
+        start()
+        return client.config {
+            expectSuccess = false
+        }
     }
 }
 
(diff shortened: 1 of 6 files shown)


ktor-server/ktor-server-test-host/jvm/src/io/ktor/server/testing/client/TestHttpClient.kt
```diff
@@ -0,0 +1,25 @@
...
+
+package io.ktor.server.testing.client
+
+import io.ktor.client.engine.*
+import io.ktor.server.testing.*
+
+/**
+ * A [TestHttpClientConfig] connected to the application engine.
+ */
+val TestApplicationEngine.TestHttpClient: HttpClientEngineFactory<TestHttpClientConfig>
```
e5l: @InternalAPI

cy6erGn0m: No, this is not internal, this is a shortcut

cy6erGn0m: So one could write

```kotlin
HttpClient(TestHttpClient).use { client ->   ... }
```

cy6erGn0m:
@e5l  so how about this?

e5l:
lgtm

cy6erGn0m:
Merged.

Issue #1368, opened 2019-09-29
CallLogging MDC with sessions: Application feature Sessions is not installed 


AndreasVolkmann:
**Ktor Version and Engine Used (client or server and name)**
1.2.4, server, [cio, serialization, auth, locations, client-cio]

**Describe the bug**
Just tried out the CallLogging MDC feature, but as soon as I try to get a value from sessions, my server becomes unresponsive. The call just takes forever and doesn't return. No log shows up. 
Then I tried the `CallId` feature with the same logic and this time it throws: 
`io.ktor.application.MissingApplicationFeatureException: Application feature Sessions is not installed`. 

Once I switched to netty, it shows the same exception when using the `CallLogging` feature. 

**To Reproduce**
1. Write the following
```kotlin
class OAuthSession(val userId: Int)

fun main() {
    embeddedServer(CIO, 5000) {
        install(CallLogging) {
            mdc("test") {
                it.sessions.get<OAuthSession>()?.userId?.toString()
            }
        }
        install(Sessions) {
            cookie<OAuthSession>("oauthSampleSessionId") {
                val secretSignKey = hex("000102030405060708090a0b0c0d0e0f")
                transform(SessionTransportTransformerMessageAuthentication(secretSignKey))
            }
        }
        install(Routing) {
            get {
                call.respond("Ok")
            }
        }
    }.start(true)
}
```

2. Start the server and call it via `localhost:5000`.
3. In CIO, no error is shown. With netty, exception is thrown. 

**Expected behavior**
The session should be retrieved and used for mdc / call id.

cy6erGn0m:
The first issue is that it reports the missing feature that is not true. This is just wrong diagnostics. 
However, the root cause of your issue is that MDC captures too early so sessions are not yet configured for this call. 

And even worse, we can't fix that easily because it's not clear what should be first. Let me explain the problem: what should happen if a session extraction is failed? Should error message be logged with MDC? How do we get MDC with no sessions?

AndreasVolkmann:
@cy6erGn0m Thanks for the explanation. For me this functionality is not a must. Whatever the outcome, it should probably be reflected in the docs.

oleg-larshin:
Please check the <a href="https://youtrack.jetbrains.com/issue/KTOR-550">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

PR #1373, opened 2019-09-30
Improve missing session diagnostics

cy6erGn0m:
**Subsystem**
Server

**Motivation**
Currently, we are reporting missing feature error when a session is requested too early before the sessions feature get chance to act. For example, in #1368, due to the imprecise diagnostic, the error looks a bit mysterious and causes confusion. 

**Solution**
When a session is not found, check feature first and then report the right error.


cy6erGn0m:
Merged manually.

Issue #1369, opened 2019-09-29
java.lang.NoClassDefFoundError: Failed resolution of: Ljava/time/ZoneId;


JajaComp:
**Ktor Version and Engine Used (client or server and name)**
io.ktor:ktor-server-netty:1.2.5

**Describe the bug**
Ktor not work on Android API =< 25

**To Reproduce**
When i try to create server i get error:

E/AndroidRuntime: FATAL EXCEPTION: main
    Process: com.example.myapplication, PID: 5399
    java.lang.NoClassDefFoundError: Failed resolution of: Ljava/time/ZoneId;
        at io.ktor.features.DefaultHeaders.<clinit>(DefaultHeaders.kt:70)

ZoneId was added in android just in API 26

cy6erGn0m:
As a workaround, you could simply rewrite [DefaultHeaders](https://github.com/ktorio/ktor/blob/master/ktor-server/ktor-server-core/jvm/src/io/ktor/features/DefaultHeaders.kt) feature to avoid using java.time package.

cy6erGn0m:
However, I would expect it to not work anyway since Netty may require a lot of API that doesn't fit.

dtmilano:
I experienced a similar problem

```
uiAutomatorHelper(com.dtmilano.android.culebratester2.UiAutomatorHelper)
java.lang.NoClassDefFoundError: Failed resolution of: Ljava/time/Duration;
	at com.typesafe.config.impl.ConfigImpl.fromAnyRef(ConfigImpl.java:243)
	at com.typesafe.config.impl.ConfigImpl.fromPathMap(ConfigImpl.java:200)
	at com.typesafe.config.ConfigFactory.parseMap(ConfigFactory.java:1038)
	at io.ktor.server.engine.CommandLineKt.commandLineEnvironment(CommandLine.kt:32)
	at io.ktor.server.cio.EngineMain.main(EngineMain.kt:20)
...
[...]

PR #1375, opened 2019-09-30
Reduce required JDK version for DefaultHeaders feature

cy6erGn0m:
**Subsystem**
Server, [DefaultHeaders](https://ktor.io/servers/features/default-headers.html)

**Motivation**
Currently, [DefaultHeaders](https://ktor.io/servers/features/default-headers.html) feature requires JDK8 due to java.time dependency. However, the only we need is GMT time formatting.

#1369 java.lang.NoClassDefFoundError: Failed resolution of: Ljava/time/ZoneId;

**Solution**
Migrate to ktor's `GMTDate` with `Calendar` instantiation optimization.

### DIFF
--- a/ktor-server/ktor-server-core/jvm/src/io/ktor/features/DefaultHeaders.kt
+++ b/ktor-server/ktor-server-core/jvm/src/io/ktor/features/DefaultHeaders.kt
@@ -8,7 +8,9 @@ import io.ktor.application.*
 import io.ktor.http.*
 import io.ktor.response.*
 import io.ktor.util.*
-import java.time.*
+import io.ktor.util.date.*
+import kotlinx.atomicfu.*
+import java.util.*
 
 /**
  * Adds standard HTTP headers `Date` and `Server` and provides ability to specify other headers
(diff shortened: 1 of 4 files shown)


imanushin:
In regarding to [this coroutines issue](https://github.com/Kotlin/kotlinx.coroutines/issues/1589), probably it is better to avoid JDK with version lower than 1.8.

ktor-server/ktor-server-core/jvm/src/io/ktor/features/DefaultHeaders.kt
```diff
@@ -8,20 +8,20 @@ import io.ktor.application.*
...
  * that are included in responses.
  */
 class DefaultHeaders(config: Configuration) {
     private val headers = config.headers.build()
-
-    private val UTC = ZoneId.of("UTC")!! // it is very important to get it like that
-    private val zoneUTCRules = UTC.rules
+    private val clock = config.clock
 
     private var cachedDateTimeStamp: Long = 0L
```
e5l: atomic or volatile?

Issue #1365, opened 2019-09-28
GET request without body (EmptyContent) results in error on iOS 13


alekseychugaev:
**Ktor Version and Engine Used (client or server and name)**
ktor_version=1.3.0-beta-1
kotlin_version=1.3.50
implementation "io.ktor:ktor-client-ios:$ktor_version"
implementation "io.ktor:ktor-client-core-native:$ktor_version"
implementation "io.ktor:ktor-client-json-native:$ktor_version"
implementation "io.ktor:ktor-client-serialization-native:$ktor_version"

**Describe the bug**
When sending a GET request without any body, the resulting request still contains body as `{}`.
My request:
```
client.get<MyResponse>("$apiEndpoint/summary?v=1") {
    contentType(ContentType.Application.Json)
}
...
```
Logs:
```
REQUEST: http://0.0.0.0:8080/summary?v=1
METHOD: HttpMethod(value=GET)
COMMON HEADERS
-> User-Agent: timezone=Australia/Melbourne;
-> Authorization: d0460e3b-8ef7-4cd5-bf03-f02feb109ace
-> Accept: application/json
-> Accept-Charset: UTF-8
CONTENT HEADERS
BODY Content-Type: application/json
BODY START
{}
BODY END
```
Looks like iOS 13 does not allow body content for GET requests, so as a result the client fails with error: 
```
GET method must not have a body
finished with error [-1103] Error Domain=NSURLErrorDomain Code=-1103 "resource exceeds maximum size"
```
Another way to reproduce:
```
val unitJson = KotlinxSerializer().write(Unit)
println("$unitJson")
```

**Expected behavior**
When request body for a GET request is specified as `var body: Any = EmptyContent` the actual body sent by ktor should be empty.

From what I can see the body is created here:
https://github.com/ktorio/ktor/blob/master/ktor-client/ktor-client-features/ktor-client-json/common/src/io/ktor/client/features/json/JsonFeature.kt#L100
It tries to serialize a Unit object.

cy6erGn0m:
In terms of json, an empty content is `{}` so I see no error here.

alekseychugaev:
Removing content-type from request worked, thanks for your help.

cy6erGn0m:
Didn't you mean Accept instead of Content-Type?

alekseychugaev:
I think it was just a copy/paste of a post request I had. It worked fine with Xcode 10.x (so I didn't notice it), but started failing after upgrading to Xcode 11.

[...]

PR #1376, opened 2019-10-01
Send empty body on GET request

Monchi:
**Subsystem**
Client, JsonFeature

**Motivation**
As mentioned in https://github.com/ktorio/ktor/issues/1365, When sending a GET request without any body, the resulting request still contains body as {}. But iOS 13 does not allow body content for GET requests, so as a result the client fails with error.

**Solution**
Send `EmptyContent` on GET request.


cy6erGn0m:
Well, the reason why json feature is applied is that you are setting request content type. Why do you do so? This is not correct and this leads to the error.

Monchi:
@cy6erGn0m Thanks. Removing content-type worked well.

PR #1377, opened 2019-10-01
Problem: PartialContent crashes with empty If-Range header

adelnizamutdinov:
**Subsystem**
Server, ktor-server-core

**Motivation**
PartialContent crashes with empty If-Range header (trying to parse it)

**Solution**
Dont' crash, check if If-Range is blank

### DIFF
--- a/ktor-server/ktor-server-core/jvm/src/io/ktor/features/PartialContent.kt
+++ b/ktor-server/ktor-server-core/jvm/src/io/ktor/features/PartialContent.kt
@@ -109,7 +109,7 @@ class PartialContent(private val maxRangeCount: Int) {
         val versions = conditionalHeadersFeature?.versionsFor(content) ?: content.defaultVersions
         val ifRange = call.request.header(HttpHeaders.IfRange)
 
-        return ifRange == null || versions.all { version ->
+        return ifRange.isNullOrBlank() || versions.all { version ->
             when (version) {
                 is EntityTagVersion -> version.etag in ifRange.parseMatchTag()
                 is LastModifiedVersion -> version.lastModified <= ifRange.fromHttpToGmtDate()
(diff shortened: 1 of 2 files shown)


cy6erGn0m:
Is there any meaning for the empty header? Can we simply ignore it?

adelnizamutdinov:
I don’t really understand your question, but my problem is that my server responds 500 when Overcast iOS podcast player sends this empty header

cy6erGn0m:
The question is about what is right. Unfortunately, the RFC doesn't clearly specify what should be done when `If-Range` header is wrong. I would fallback to non-partial content in this case. Do you have any idea why the client send the broken header? Don't you send an empty E-Tag or something?

cy6erGn0m:
Since there is no strictly specified behaviour for an empty `If-Range`, I'd try to "guess" the intension. `If-Range` header seems to be similar to `If-Match`/`If-None-Match`/`If-Modified-Since`/`If-Unmodified-Since`. Obviously, an empty header couldn't be interpresed as a date so we concentrate on `If-Match` and `If-None-Match` headers. Both allows a list of entity tags. So an empty header means an empty list of entities that is not prohibited by the RFC. Unfortunately, `If-Range` doesn't allow a list but a single entity tag. However, it looks like we already interpret `If-Range` as a list of etags (that is not exactly correct) so we just shouldn't try to parse it as a date when it doesn't have a GMT suffix (that is always required in a case when `If-Range` is a date).

adelnizamutdinov:
thanks for taking a concern about this issue.

I was not sending an `ETag` to the client. I recently switched to serving partial content with nginx, it sends `ETag`s, and has no issues with empty `If-Range`. Or maybe Overcast is not sending `If-Range` to nginx. I can check that later

cy6erGn0m:
Will be fixed in the related PR. Thank you for pointing out us the problem and for your contribution.

cy6erGn0m:
PS: declined since we have more complex solution for this.

PR #1392, opened 2019-10-09
Introduce Local Port Routing Builder

hosamaly:
**Subsystem**
ktor-server

**Motivation**
I'm building an application that has some functions that should be exposed on a separate port (e.g. metrics). The existing (experimental) [`HostRouteSelector`](https://api.ktor.io/1.2.4/io.ktor.routing/-host-route-selector/index.html) (introduced in #835) is geared towards multi-tenant/VirtualHost-style configurations, allowing a client to send a `Host` header to get information that perhaps shouldn't be exposed to them.

**Solution**
I suggest a simple `RouteSelector` based on the `local: RequestConnectionPoint` port, whose documentation says (emphasis mine):
> Contains http request and connection details such as a host name used to connect, port, scheme and so on. ***No proxy headers could affect it.***

Sadly, some implementations don't abide by this rule. Particularly noteworthy is the [`TestApplicationRequest`](https://api.ktor.io/1.2.4/io.ktor.server.testing/-test-application-request/index.html), which extracts the port from headers. I think that it should be changed to enable setting the port in the request builder, but this could be left for another PR.

### DIFF
--- a/ktor-server/ktor-server-core/jvm/src/io/ktor/routing/LocalPortRoutingBuilder.kt
+++ b/ktor-server/ktor-server-core/jvm/src/io/ktor/routing/LocalPortRoutingBuilder.kt
@@ -1,0 +1,54 @@
+/*
+ * Copyright 2014-2019 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.routing
+
+import io.ktor.http.parametersOf
+import io.ktor.util.*
+
+/**
+ * Create a route to match the port on which the call was received.
+ *
+ * The selector checks the [io.ktor.request.ApplicationRequest.local] request port,
+ * _ignoring_ HTTP headers such as "Host" or "X-Forwarded-Host".
+ * This is useful for securing routes under separate ports.
+ *
+ * For multi-tenant applications, you may want to use [io.ktor.routing.port],
+ * which takes HTTP headers into consideration.
+ *
+ * @param port the port to match against
+ *
+ * @throws IllegalArgumentException if the port is outside the range of acceptable, non-ephemeral ports
+ */
+@KtorExperimentalAPI
+fun Route.localPort(port: Int, build: Route.() -> Unit): Route {
+    require(port in 1..49151) { "Port $port must be a positive number within the range of non-ephemeral ports" }
+
+    val selector = LocalPortRouteSelector(port)
+    return createChild(selector).apply(build)
+}
+
+/**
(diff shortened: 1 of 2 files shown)


spand:
I dont think its correct to check that the port is outside an ephemeral port range. I believe the range is configurable and different on many systems.

Also a separate pr maybe but I feel an ip version of the selector ought to be there also.

hosamaly:
Thanks for the feedback, @spand. I'm happy to remove the check for ephemeral ports; perhaps it was too specific to my design decisions. I'll change this now.

As for the IP version, I agree with you that it's better left to a separate PR.

hosamaly:
Hi, @spand. How does this PR look now from your perspective?

ktor-server/ktor-server-core/jvm/src/io/ktor/routing/LocalPortRoutingBuilder.kt
```diff
@@ -19,11 +19,11 @@ import io.ktor.util.*
  *
  * @param port the port to match against
  *
- * @throws IllegalArgumentException if the port is outside the range of acceptable, non-ephemeral ports
+ * @throws IllegalArgumentException if the port is outside the range of TCP/UDP ports
  */
 @KtorExperimentalAPI
 fun Route.localPort(port: Int, build: Route.() -> Unit): Route {
-    require(port in 1..49151) { "Port $port must be a positive number within the range of non-ephemeral ports" }
+    require(port in 1..65535) { "Port $port must be a positive number between 1 and 65,535" }
```
spand: Entering bike shedding territory here but afaik 0 is also valid. A quick look in `InetSocketAddress` has:
```
private static int checkPort(int port) {
        if (port < 0 || port > 0xFFFF)
            throw new IllegalArgumentException("port out of range:" + port);
        return port;
    }
```

spand:
I put in a comment but looks good to me.. though I am just a nobody in this project :)

ktor-server/ktor-server-core/jvm/src/io/ktor/routing/LocalPortRoutingBuilder.kt
hosamaly: According to [Wikipedia](https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers#Well-known_ports), `0` is reserved. It also adds that:
> in programming APIs (not in communication between hosts), [it] requests a system-allocated (dynamic) port

That wouldn't be useful as a filter.

PR #1393, opened 2019-10-09
HttpCookies feature cookie encoding retention

T-Fowl:
**Subsystem**
Client, HttpCookies feature.

**Motivation**
The HttpCookies feature attempted to encode all cookies using DQUOTES rather than respecting the cookies own encoding, which defaults to URI during parsing anyway.

**Solution**
HttpCookies feature now renders cookies using their own encoding.


cy6erGn0m:
This breaks tests in `CookiesTest`

PR #1394, opened 2019-10-10
Fix http-redirect conformance

T-Fowl:
**Subsystem**
Client, HttpRedirect feature

**Motivation**
While MDN **incorrectly** states that rewriting redirected POST requests to use GET is against the specification, it is actually **allowed** in RFC7231, specifically for [301 Moved Permanently and 302 Found](https://tools.ietf.org/html/rfc7231#page-56).
Most notably chrome, and probably all major browsers, do this (i.e. when submitting a login form which then receives a 302 Found back to the authenticated page).

**Solution**
Updated the HttpRedirect feature with a configuration class which allows the user to explicitly allow the rewrites of POST to GET (defaults to false) as well as a maximum number of redirects to follow (default is [RFC2068's 5](https://tools.ietf.org/html/rfc2068#section-10.3)).

This now allows users to do the following:
```kotlin
install(HttpRedirect) {
    maximumRedirects = 23 // Follow up to 23 redirects
    rewritePostAsGet = true // Enable rewrite to GET
}
```

Something that needs to be discussed, in the event that the client is redirected more than the configured number of times, I was unsure as to what to actually do:
```kotlin
// TODO: Just return last call, return original (closed) call, or throw an exception?
if (redirects > feature.maximumRedirects) return call
```

### DIFF
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/features/HttpRedirect.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/features/HttpRedirect.kt
@@ -13,9 +13,7 @@ import io.ktor.util.*
 /**
  * [HttpClient] feature that handles http redirect
  */
-class HttpRedirect(config: Config) {
-    val maximumRedirects: Int = config.maximumRedirects
-    val rewritePostAsGet: Boolean = config.rewritePostAsGet
+class HttpRedirect(val maximumRedirects: Int, val rewritePostAsGet: Boolean) {
 
     class Config {
         /** RFC2068 recommends a maximum of five redirection */
@@ -29,7 +27,10 @@ class HttpRedirect(config: Config) {
     companion object Feature : HttpClientFeature<Config, HttpRedirect> {
         override val key: AttributeKey<HttpRedirect> = AttributeKey("HttpRedirect")
 
-        override fun prepare(block: Config.() -> Unit): HttpRedirect = HttpRedirect(Config().apply(block))
+        override fun prepare(block: Config.() -> Unit): HttpRedirect {
+            val config = Config().apply(block)
+            return HttpRedirect(config.maximumRedirects, config.rewritePostAsGet)
+        }
 
         override fun install(feature: HttpRedirect, scope: HttpClient) {
             scope.feature(HttpSend)!!.intercept { origin ->
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/features/HttpRedirect.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/features/HttpRedirect.kt
@@ -15,10 +15,15 @@ import io.ktor.util.*
  */
 class HttpRedirect(config: Config) {
     val maximumRedirects: Int = config.maximumRedirects
+    val rewritePostAsGet: Boolean = config.rewritePostAsGet
 
     class Config {
-        /* RFC2068 recommends a maximum of five redirection */
+        /** RFC2068 recommends a maximum of five redirection */
         var maximumRedirects: Int = 5
+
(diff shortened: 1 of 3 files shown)


cy6erGn0m:
Could you please provide a test?

T-Fowl:
Added some tests but have spent all afternoon trying to work out why some of them aren't working. Specifically ones that involve HttpRedirect throwing an exception when too many redirects are received. HttpSend currently does this (at 20) but I cannot manage to replicate this in HttpRedirect.  


The following is in HttpRedirect:
```kotlin
if (redirects >= feature.maximumRedirects)
    throw RedirectCountExceedException("Max redirect count ${feature.maximumRedirects} exceeded")
```

Then on line 62-76 of HttpSend:
```kotlin
var currentCall = sender.execute(context) // SendCountExceedException gets thrown here
var callChanged: Boolean

do {
    callChanged = false

    passInterceptors@ for (interceptor in feature.interceptors) {
        val transformed = interceptor(sender, currentCall) // will be thrown here
        if (transformed === currentCall) continue@passInterceptors

        currentCall = transformed
        callChanged = true
        break@passInterceptors
    }
} while (callChanged)
```

The exception thrown in HttpRedirect should flow up the call stack into the interceptor loop in HttpSend and cause the same effect as the SendCountExceedException exception in HttpSend. However this exception seems to be lost somewhere.

Tests requiring this exception to be thrown simply timeout.

The redirect loop in HttpRedirect exits, but the call does not seem to complete at all (let alone throwing)

However this exception is not having an effect on the pipeline execution during testing and is causing tests to timeout.

ktor-client/ktor-client-core/common/src/io/ktor/client/features/HttpRedirect.kt
```diff
@@ -39,9 +70,18 @@ class HttpRedirect {
                     url.parameters.clear()
 
                     location?.let { url.takeFrom(it) }
+
+                    if (call.response.status.isRewritable()) {
+                        if (feature.rewritePostAsGet && call.request.method == HttpMethod.Post)
```
e5l: Could you merge it on one `if`?

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/HttpRedirectTest.kt
```diff
@@ -30,6 +31,33 @@ class HttpRedirectTest : ClientLoader() {
...
+    @Test
+    fun rewrittenRedirectTest(): Unit = clientTests {
+        config {
+            install(HttpRedirect) {
+                rewritePostAsGet = true
+            }
+        }
+
+        test { client ->
+            client.post<HttpResponse>("$TEST_URL_BASE/post-expecting-get-301").use {
```
e5l: Could you rebase tests on master?

ktor-client/ktor-client-core/common/src/io/ktor/client/features/HttpRedirect.kt
```diff
@@ -39,9 +70,18 @@ class HttpRedirect {
...
+                            method = HttpMethod.Get
+                    }
                 })
+                redirects++
 
                 if (!call.response.status.isRedirect()) return call
+
+                if (redirects >= feature.maximumRedirects)
+                    throw RedirectCountExceedException("Max redirect count ${feature.maximumRedirects} exceeded")
```
e5l: It's quite ambiguous with a limit in the `HttpSend` feature - we throw 2 kinds of exceptions for similar reasons.
Could you tell me if you have an idea of how we can simplify the handling of such cases?

e5l:
Hi @T-Fowl, thanks for the PR. It's almost good, except minor usability issues.

PR #1396, opened 2019-10-11
option to include/ exclude query parameters in the constructed url

selectAll:
**Subsystem**
Locations API
**Motivation**
To construct the url based on location, query parameters are also included by default. There is no way to exclude the query parameters.
**Solution**
Added ability to specify whether to include the query params. By providing the default as true, this is not a breaking change.


cy6erGn0m:
`href(x, false) == href(x).substrinfAfter('?')`

Issue #1277, opened 2019-08-11
support application.conf  with env type


jicui:
I found it is pretty useful to manage couple of different profile (env) ,like application-dev.conf /application-prod.conf/application-test.conf

so each profile runs by different start up command ,similiar like how sprintboot does 

but unfortunately I found env does not support well 
https://ktor.io/servers/configuration/environments.html

does ktor has plan to add this feature 

java -jar xxx.jar -profile=prod  ==> will look for application-prod.conf to load

jicui:
I'm able to enhance this by provide local EngineMain ,by override 
CommandLine.kt ==>commandLineEnvironment ,so once detect --profile ,will append as application-prod |application-dev to load

can this be consider merged into common

oleg-larshin:
Please check the <a href="https://youtrack.jetbrains.com/issue/KTOR-574">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

PR #1398, opened 2019-10-14
#1277 introduce new command line argument 'profile'

doyaaaaaken:
**Subsystem**
Server

**Motivation**

refs #1277
Introduce new command line argument 'profile', and we can switch configuration files easily according to the environment like using `java -jar xxx.jar -profile=prd`.


dragneelfps:
any update on merging this? I really need this in ktor for even basic env based configuration.

PR #1399, opened 2019-10-14
Fix #1397 - Headers::getAll(String) return null when key not found

### LINKED ISSUE #1397: Inconsistent behaviour of Headers::getAll(String)
ktor-server-core:1.2.5
ktor-server-netty:1.2.5
ktor-server-cio:1.2.5

**Describe the bug**
There is an inconsistency in `Headers::getAll(String)`(actually `StringValues::getAll(String)`). The KDoc says:

> Gets all values associated with the [name], **or null if the name is not present**

The problem is that for both CIO and Netty engines it is never null. Even if **name** was not present empty collection is returned.

For Netty this behaviour is defined in `io.netty.handler.codec.HeadersUtils::getAllAsString`.
For CIO in `io.ktor.http.cio.HttpHeadersMap::getAll`

**Expected behavior**
I would expect this function to actually work as its own KDoc says.

T-Fowl:
**Subsystem**
ktor-client-okhttp
ktor-server-netty
ktor-server-jetty
ktor-server-cio

**Motivation**
Fix for #1397 

**Solution**
Updated all the implementations of Headers::getAll(String) to return null instead of an empty list when the key does not exist.

In general this was done with `.takeIf { it.isNotEmpty() }` as the underlying native header containers tend to return an empty list when the key does not exist.


cy6erGn0m:
Merged manually, thanks

PR #1405, opened 2019-10-21
Fix hanging suspension in CIO endpoint

dmitrievanthony:
**Subsystem**
Client, CIOEngine

**Motivation**
In CIOEngine `endpoint.execute(data, callContext)` hangs if `callContext` is cancelled.

**Solution**
Add `invokeOnCompletion` to a dedicated request task that checks if task completed with an exception and passes this exception to response continuation.


dmitrievanthony:
It will be closed at https://github.com/ktorio/ktor/pull/1409 because it's impossible to reproduce the problem so far.

PR #1407, opened 2019-10-21
Fixes #1406

### LINKED ISSUE #1406: Http-Client Cache Feature Implements "must-revalidate" incorrectly
1.2.5 Ktor Client CIO version - HttpCache Feature

**Describe the bug**
https://github.com/ktorio/ktor/blob/master/ktor-client/ktor-client-core/common/src/io/ktor/client/features/cache/HttpCacheEntry.kt#L88

`must-revalidate` cache header only needs to revalidate once the expiry has passed.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control

**To Reproduce**
1. Hit: https://www.googleapis.com/robot/v1/metadata/x509/securetoken@system.gserviceaccount.com
2. Hit: https://www.googleapis.com/robot/v1/metadata/x509/securetoken@system.gserviceaccount.com again it should come from cache for about 12 hours 
3. It will revalidate everytime as cache-control has `must-validate` in it.

**Expected behavior**

Has cache-control header?
Yes: Has Expiry Header?
Yes: Has Expired?
Yes: `must-revalidate`?
- No? Try and fetch new response otherwise serve cached response
- Yes? Fetch new response or fail.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control

**Screenshots**

chrisjenx:
Fixed must-revalidate on the request side for now as per details in #1406.

Note: Response interceptor should check cache on response error to see if we can serve the cached response - in the case of `must-revalidate` it can serve the cache as long as it's not stale, once stale and must-revalidate then it cache should be purged and return the response error

**Subsystem**
Client/Core

**Motivation**
See #1406 for how must-revalidate details

**Solution**
For now we disable `must-revalidate` on the request side, as it doesn't really do anything different to a stale response.

When error response recover is introduced - must-revalidate + max-age will act together to see if they can return a cached response in the case of a response failure. 
Normal cacheing would try and refresh once expired but return if it fails to refresh. Once you add `must-revalidate` a stale response can not be returned when getting a fresh response fails.

### DIFF
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/features/cache/HttpCacheEntry.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/features/cache/HttpCacheEntry.kt
@@ -87,9 +87,13 @@ internal fun HttpResponse.cacheExpires(): GMTDate {
 @KtorExperimentalAPI
 internal fun HttpCacheEntry.shouldValidate(): Boolean {
     val cacheControl = responseHeaders[HttpHeaders.CacheControl]?.let { parseHeaderValue(it) } ?: emptyList()
-    var result = GMTDate() > expires
-    result = result || CacheControl.MUST_REVALIDATE in cacheControl
-    result = result || CacheControl.NO_CACHE in cacheControl
-
-    return result
+    val isStale = GMTDate() > expires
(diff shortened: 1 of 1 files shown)


cy6erGn0m:
@e5l what a shame we didn't fix the commit message, it's so ugly

PR #1408, opened 2019-10-22
HTTP headers validation

cy6erGn0m:
**Subsystem**
ktor utils, server, client

**Motivation**
Currently, header name and values requirements are not strict enough since we only prohibit setting so-called "unsafe" headers while we allow setting illegal headers. Setting illegal header names or values may cause HTTP protocol violations, software incompatibility, security issues and other negative consequences.

**Solution**
- restrict CIO HTTP parser so illegal names and values causing the request to be aborted
- restrict API to not allow setting illegal headers via `HeadersBuilder` and server's `ResponseHeaders`

### DIFF
--- a/ktor-http/ktor-http-cio/common/src/io/ktor/http/cio/internals/Tokenizer.kt
+++ b/ktor-http/ktor-http-cio/common/src/io/ktor/http/cio/internals/Tokenizer.kt
@@ -4,13 +4,25 @@
 
 package io.ktor.http.cio.internals
 
-import io.ktor.http.cio.*
-
 internal fun nextToken(text: CharSequence, range: MutableRange): CharSequence {
     val spaceOrEnd = findSpaceOrEnd(text, range)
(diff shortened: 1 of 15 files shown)


ktor-http/ktor-http-cio/common/src/io/ktor/http/cio/internals/Tokenizer.kt
```diff
@@ -4,15 +4,27 @@
...
-import io.ktor.http.cio.*
-
 internal fun nextToken(text: CharSequence, range: MutableRange): CharSequence {
     val spaceOrEnd = findSpaceOrEnd(text, range)
     val s = text.subSequence(range.start, spaceOrEnd)
     range.start = spaceOrEnd
     return s
 }
 
+internal fun skipSpHTab(
```
e5l: Consider avoiding shortcuts in the parser

cy6erGn0m:
Merged manually

PR #1429, opened 2019-11-08
Handle callContext correctly in Mock client engine.

dmitrievanthony:
**Subsystem**
Mock client

**Motivation**
Call context is not handled correctly in the Mock engine and as a result, we get hanging tests.

**Solution**
Pass call context to the `HttpResponseData`.

### DIFF
--- a/ktor-client/ktor-client-mock/common/src/io/ktor/client/engine/mock/MockEngineConfig.kt
+++ b/ktor-client/ktor-client-mock/common/src/io/ktor/client/engine/mock/MockEngineConfig.kt
@@ -8,11 +8,17 @@ import io.ktor.client.call.*
 import io.ktor.client.engine.*
 import io.ktor.client.request.*
 import io.ktor.client.statement.*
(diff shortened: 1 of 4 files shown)


ktor-client/ktor-client-mock/common/src/io/ktor/client/engine/mock/MockEngineConfig.kt
```diff
@@ -8,11 +8,17 @@ import io.ktor.client.call.*
...
 /**
  * Single [HttpClientCall] to [HttpResponse] mapper.
  */
-typealias MockRequestHandler = suspend (request: HttpRequestData) -> HttpResponseData
+typealias MockRequestHandler = suspend MockRequestHandleScope.(request: HttpRequestData) -> HttpResponseData
+
+/**
+ * Scope for [MockRequestHandler].
+ */
+class MockRequestHandleScope(internal val callContext: CoroutineContext)
```
e5l: Consider using old name

PR #1434, opened 2019-11-11
Bug fixes. Migrate to `HttpStatement`

e5l:


ktor-client/ktor-client-core/common/src/io/ktor/client/call/HttpClientCall.kt
```diff
@@ -142,8 +116,16 @@ data class HttpEngineCall(val request: HttpRequest, val response: HttpResponse)
  * Constructs a [HttpClientCall] from this [HttpClient] and with the specified [HttpRequestBuilder]
  * configured inside the [block].
  */
-suspend fun HttpClient.call(block: suspend HttpRequestBuilder.() -> Unit = {}): HttpClientCall =
-    execute(HttpRequestBuilder().apply { block() })
+@Deprecated(
+    "Unbound [HttpClientCall] is deprecated. Consider using [HttpResponse] in request method instead.",
```
dmitrievanthony: I guess the message should be about using `request()` instead of `call()`, not about `HttpResponse`. Please, check here and below.

ktor-client/ktor-client-core/common/src/io/ktor/client/features/DefaultResponseValidation.kt
```diff
@@ -24,21 +23,19 @@ fun HttpClientConfig<*>.addDefaultResponseValidation() {
...
+            when (statusCode) {
+                in 300..399 -> throw RedirectResponseException(exceptionResponse)
+                in 400..499 -> throw ClientRequestException(exceptionResponse)
+                in 500..599 -> throw ServerResponseException(exceptionResponse)
+            }
 
-                if (statusCode >= 600) {
-                    throw ResponseException(exceptionResponse)
-                }
+            if (statusCode >= 600) {
```
dmitrievanthony: Can we move it into `when(...) else ->` section?

ktor-client/ktor-client-core/common/src/io/ktor/client/statement/HttpResponse.kt
```diff
@@ -0,0 +1,76 @@
...
+import io.ktor.http.*
+import io.ktor.util.*
+import io.ktor.util.date.*
+import io.ktor.utils.io.*
+import kotlinx.coroutines.*
+
+/**
+ * A response for [HttpClient], second part of [HttpClientCall].
+ */
+abstract class HttpResponse : HttpMessage, CoroutineScope {
```
dmitrievanthony: Why not an interface?

ktor-client/ktor-client-core/common/src/io/ktor/client/statement/HttpStatement.kt
```diff
@@ -0,0 +1,139 @@
...
+     *
+     * Note if T is a streaming type, you should manage how to close it manually.
+     */
+    @UseExperimental(ExperimentalStdlibApi::class)
+    suspend inline fun <reified T> receive(): T = when (T::class) {
+        HttpStatement::class -> this as T
+        HttpResponse::class -> execute() as T
+        else -> {
+            val response = executeUnsafe()
+            val result =  try {
```
dmitrievanthony: Why not just

```
try {
    response.receive<T>()
}  finally {
    response.complete()
}
```

?

ktor-client/ktor-client-android/jvm/test/io/ktor/client/engine/android/AndroidHttpsTest.kt
```diff
@@ -103,8 +104,9 @@ class AndroidHttpsTest : TestWithKtor() {
     fun external(): Unit = runBlocking {
         val client = HttpClient(Android)
 
-        val response = client.get<HttpResponse>("https://kotlinlang.org")
-        assertEquals(HttpStatusCode.OK, response.status)
+        client.get<HttpStatement>("https://kotlinlang.org").execute { response ->
```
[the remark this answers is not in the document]
e5l: Please note that we need `execute { block }` only if the content is a stream or it doesn't fit in memory. 

Probably you'll use: 
```
client.get<User>()
client.get<Photo>()
```
in production.

Btw to get streaming body you can `get<ByteReadChannel>()`.

ktor-client/ktor-client-core/common/src/io/ktor/client/call/HttpClientCall.kt
```diff
@@ -142,8 +116,16 @@ data class HttpEngineCall(val request: HttpRequest, val response: HttpResponse)
...
+    level = DeprecationLevel.ERROR,
+    replaceWith = ReplaceWith(
+        "this.request<HttpResponse>(block)",
+        "io.ktor.client.request.request",
+        "io.ktor.client.statement.*"
+    )
+)
+suspend fun HttpClient.call(block: suspend HttpRequestBuilder.() -> Unit = {}): HttpClientCall = error("")
```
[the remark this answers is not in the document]
e5l: It will make no effect in public API

ktor-client/ktor-client-core/common/src/io/ktor/client/features/DefaultTransform.kt
```diff
@@ -48,27 +50,37 @@ fun HttpClient.defaultTransformers() {
 
         when (info.type) {
             Unit::class -> {
-                response.close()
+                body.cancel()
                 proceedWith(HttpResponseContainer(info, Unit))
             }
+            Int::class -> {
```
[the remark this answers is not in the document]
e5l: Could you tell me what kind of primitives do you want to receive with client?

ktor-client/ktor-client-core/common/src/io/ktor/client/statement/HttpStatement.kt
e5l: ?

ktor-client/ktor-client-core/common/src/io/ktor/client/statement/HttpResponse.kt
e5l: It could be. We must consider use cases.

ktor-client/ktor-client-android/jvm/test/io/ktor/client/engine/android/AndroidHttpsTest.kt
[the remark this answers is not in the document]
dmitrievanthony: Agree. But just as an idea we could also use syntax like this:

```
client.withCleanup {
    val usersRes = get<HttpStatement>("http://localhost/users")
    for (user in usersRes) {
        val userRes = get<HttpStatement>("http://localhost/details/${user}")
        val photoRes = get<HttpStatement>("http://localhost/photo/${userRes.photoId}")
        // etc...
    }
}
```

ktor-client/ktor-client-core/common/src/io/ktor/client/call/HttpClientCall.kt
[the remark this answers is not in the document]
dmitrievanthony: What about `HttpClient.call` for example? Will `@InternalAPI` remove this method from suggestions?

ktor-client/ktor-client-core/common/src/io/ktor/client/features/DefaultTransform.kt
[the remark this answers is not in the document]
dmitrievanthony: It just looks weird that we handle such a specific class like `Int`, but don't handle other primitives. What about `Boolean` of `Float`?

ktor-client/ktor-client-core/common/src/io/ktor/client/call/SavedCall.kt
```diff
@@ -35,13 +34,9 @@ internal class SavedHttpResponse(
 
     override val headers: Headers = origin.headers
 
-    override val coroutineContext: CoroutineContext = origin.coroutineContext + Job()
+    override val coroutineContext: CoroutineContext = origin.coroutineContext
```
cy6erGn0m: Well... who is the parent job coming from `origin`? Doesn't it break anything if one cancel a response?

ktor-client/ktor-client-core/common/src/io/ktor/client/call/utils.kt
```diff
@@ -27,38 +23,37 @@ class UnsupportedUpgradeProtocolException(
...
-internal fun HttpResponse.channelWithCloseHandling(): ByteReadChannel = writer {
-    try {
-        content.joinTo(channel, closeOnEnd = true)
-    } catch (cause: CancellationException) {
-        this@channelWithCloseHandling.cancel(cause)
-    } finally {
-        this@channelWithCloseHandling.close()
-    }
-}.channel
+): HttpClientCall = error("")
```
cy6erGn0m: One more emtpy

ktor-client/ktor-client-core/common/src/io/ktor/client/engine/HttpClientEngineConfig.kt
```diff
@@ -40,16 +27,17 @@ open class HttpClientEngineConfig {
...
      * See [ProxyBuilder] to create proxy.
      */
     @KtorExperimentalAPI
     var proxy: ProxyConfig? = null
+
+    @Deprecated(
+        "Response config is deprecated. See [HttpPlainText] feature for charset configuration",
+        level = DeprecationLevel.ERROR
+    )
+    val response: Unit get() = error("")
```
cy6erGn0m: `Unit` -> `Nothing`?

cy6erGn0m: + error message

ktor-client/ktor-client-core/common/src/io/ktor/client/features/DefaultTransform.kt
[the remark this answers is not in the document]
cy6erGn0m: Why do we support it? What use-cases for reading a single integer? One could simply do `get<String>(...).toInt()` instead

ktor-client/ktor-client-core/common/src/io/ktor/client/features/HttpRedirect.kt
```diff
@@ -32,15 +34,14 @@ class HttpRedirect {
...
+                val requestBuilder = HttpRequestBuilder().apply {
                     takeFrom(origin.request)
                     url.parameters.clear()
 
                     location?.let { url.takeFrom(it) }
-                })
+                }
 
+                call = execute(requestBuilder)
                 if (!call.response.status.isRedirect()) return call
```
cy6erGn0m: I'm just curious, what if there is a redirect response with body, what would happend and who does consume/discard it here?

ktor-client/ktor-client-core/common/src/io/ktor/client/features/HttpSend.kt
```diff
@@ -80,14 +80,25 @@ class HttpSend(
...
-            sentCount++
+            currentCall?.cancel()
+
+            if (sentCount >= maxSendCount) {
+                throw SendCountExceedException("Max send count $maxSendCount exceeded")
+            }
 
-            return client.sendPipeline.execute(requestBuilder, requestBuilder.body) as HttpClientCall
+            sentCount++
+            val call = client.sendPipeline.execute(requestBuilder, requestBuilder.body) as HttpClientCall
```
cy6erGn0m: Shouldn't we provide a more specific error message here instead of ClassCastException so that it will explain what the problem is?

ktor-client/ktor-client-core/common/src/io/ktor/client/features/websocket/builders.kt
```diff
@@ -39,14 +41,24 @@ suspend fun HttpClient.webSocketSession(
...
+    val session = request<HttpStatement> {
+        url {
+            protocol = URLProtocol.WS
+            port = protocol.defaultPort
+        }
+        request()
+    }
+
+    session.receive<DefaultClientWebSocketSession, Unit> {
+        // Workaround: https://youtrack.jetbrains.com/issue/KT-34041
```
cy6erGn0m: This should already work well with eap2 so we probably need to eliminate this workaround

ktor-client/ktor-client-core/common/src/io/ktor/client/statement/HttpResponse.kt
cy6erGn0m: We need to think about it since "abstract class <-> interface" conversion is not a binary compatible change. The other think we need to notice is that we can restrict users from inheriting an abstract class while we can't do the same for an interface.

The first question is why one would like to inherit it?

ktor-client/ktor-client-core/common/src/io/ktor/client/statement/HttpStatement.kt
```diff
@@ -0,0 +1,139 @@
...
+    private val builder: HttpRequestBuilder,
+    private val client: HttpClient
+) {
+    /**
+     * Executes this statement and call the [block] with the streaming [response].
+     *
+     * The [response] argument holds a network connection until the [block] isn't completed. You can read the body
+     * on-demand or at once with [receive<T>()] method.
+     *
+     * After [block] finishes, [response] will be completed body will be discarded or released depends on the engine configuration.
```
cy6erGn0m: so a `response` instance shouldn't be captrued outside `block`, we could notice it

ktor-client/ktor-client-core/common/src/io/ktor/client/statement/HttpStatement.kt
```diff
@@ -0,0 +1,139 @@
...
+        try {
+            return block(response)
+        } finally {
+            val job = response.coroutineContext[Job]!! as CompletableJob
+
+            job.apply {
+                complete()
+                try {
+                    response.content.cancel()
+                } catch (_: Throwable) {}
```
cy6erGn0m: So what kind of errors do we suppress here?

ktor-client/ktor-client-core/common/src/io/ktor/client/statement/HttpStatement.kt
```diff
@@ -0,0 +1,139 @@
...
+/**
+ * Read the [HttpResponse.content] as a String. You can pass an optional [charset]
+ * to specify a charset in the case no one is specified as part of the Content-Type response.
+ * If no charset specified either as parameter or as part of the response,
+ * [HttpResponseConfig.defaultCharset] will be used.
+ *
+ * Note that [charset] parameter will be ignored if the response already has a charset.
+ *      So it just acts as a fallback, honoring the server preference.
+ */
+suspend fun HttpResponse.readText(charset: Charset? = null): String {
```
cy6erGn0m: We probably can rename `charset` parameter to something like `fallbackCharset` or `defaultCharset`

ktor-client/ktor-client-core/common/src/io/ktor/client/statement/Readers.kt
```diff
@@ -2,22 +2,21 @@
...
 
 import io.ktor.utils.io.*
 import io.ktor.utils.io.core.*
 
 /**
  * Exactly reads [count] bytes of the [HttpResponse.content].
  */
-suspend fun HttpResponse.readBytes(count: Int): ByteArray =
-    ByteArray(count).also { content.readFully(it) }
+suspend fun HttpResponse.readBytes(count: Int): ByteArray = content.readRemaining(count.toLong()).readBytes()
```
cy6erGn0m: This is not identical since `readRemaining` reads at most count bytes

ktor-client/ktor-client-features/ktor-client-logging/common/src/io/ktor/client/features/logging/Logging.kt
```diff
@@ -50,19 +50,14 @@ class Logging(
...
-    private suspend fun logResponse(response: HttpResponse): Unit = response.use {
+    private suspend fun logResponse(response: HttpResponse) {
         if (level.info) {
             logger.log("RESPONSE: ${response.status}")
             logger.log("METHOD: ${response.call.request.method}")
             logger.log("FROM: ${response.call.request.url}")
         }
 
         if (level.headers) logHeaders(response.headers.entries())
-        if (level.body) {
```
cy6erGn0m: So who does `logResponseBody` instead of this?

ktor-client/ktor-client-jetty/jvm/src/io/ktor/client/engine/jetty/JettyResponseListener.kt
```diff
@@ -75,7 +80,9 @@ internal class JettyResponseListener(
             headersBuilder.append(field.name, field.value)
         }
 
-        if (frame.isEndStream) backendChannel.close()
+        if (frame.isEndStream || request.method == HttpMethod.Head) {
```
cy6erGn0m: Shouldn't head response always have `isEndStream` set? Since a head response has no body, a headers frame is alwasys the last with end stream mark set, isn't it?

cy6erGn0m:
@e5l  why do we change `io/ktor/client/response/X` and `io/ktor/client/request/X` to `io/ktor/client/statement/X`?

ktor-client/ktor-client-jetty/jvm/src/io/ktor/client/engine/jetty/JettyResponseListener.kt
e5l: Yes, it always the last, and no, there is no flag in it(at least in jetty). It was a bug discovered by the statement.

ktor-client/ktor-client-core/common/src/io/ktor/client/statement/HttpStatement.kt
e5l: Already canceled

ktor-client/ktor-client-core/common/src/io/ktor/client/features/DefaultTransform.kt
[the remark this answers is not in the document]
e5l: There are many API that responds with single number and it takes no efforts to support it.

ktor-client/ktor-client-core/common/src/io/ktor/client/statement/HttpResponse.kt
e5l: I would like to keep it as an abstract class here: 
  - there are no use cases to use it with other abstract classes 
  - we have more flexibility in future wo breaking compatibility

ktor-client/ktor-client-core/common/src/io/ktor/client/call/SavedCall.kt
e5l: The `Job()` in the `SavedCall` existed to have different lifecycles in `HttpResponse` and `SavedResponse`. Now there is no public response lifecycle and we have no need to split it.

ktor-client/ktor-client-core/common/src/io/ktor/client/features/HttpRedirect.kt
e5l: We handle it in the send feature

ktor-client/ktor-client-core/common/src/io/ktor/client/statement/HttpResponse.kt
cy6erGn0m: Shouldn't we restrict inheritance for now? (e.g. by making the constructor internal or something)

ktor-client/ktor-client-core/common/src/io/ktor/client/features/HttpSend.kt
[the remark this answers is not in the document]
e5l: I did it to avoid `!!` in return

ktor-client/ktor-client-core/common/src/io/ktor/client/statement/HttpStatement.kt
cy6erGn0m: Cancelling already cancelled should have no effect and should be no exception, isn't it?

ktor-client/ktor-client-features/ktor-client-logging/common/src/io/ktor/client/features/logging/Logging.kt
e5l: In the feature itsel :168

ktor-client/ktor-client-core/common/src/io/ktor/client/statement/HttpStatement.kt
```diff
@@ -0,0 +1,145 @@
...
+     *
+     * The [response] argument holds a network connection until the [block] isn't completed. You can read the body
+     * on-demand or at once with [receive<T>()] method.
+     *
+     * After [block] finishes, [response] will be completed body will be discarded or released depends on the engine configuration.
+     *
+     * Please note: the [response] instance will be canceled and shouldn't be passed outside of [block].
+     */
+    suspend fun <T> execute(block: suspend (response: HttpResponse) -> T): T {
+        println("Execute ${builder.url.buildString()}")
```
cy6erGn0m: println and printDebugTree() later

ktor-client/ktor-client-core/common/src/io/ktor/client/statement/HttpStatement.kt
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    /**
+     * Executes this statement and run [HttpClient.responsePipeline] with the response and expected type [T].
+     *
+     * Note if T is a streaming type, you should manage how to close it manually.
+     */
+    @UseExperimental(ExperimentalStdlibApi::class)
+    suspend inline fun <reified T> receive(): T = when (typeOf<T>()) {
+        typeOf<HttpStatement>() -> this as T
```
cy6erGn0m: Can we compare typeOf results? Is it guaranteed that `typeOf<T>() == typeOf<T>()` ?

PR #1443, opened 2019-11-15
Allow passing ObjectMapper to serializer

Koriit:
**Subsystem**
Client Jackson

**Motivation**
I have my own ObjectMapper instance which I would like to use as JsonSerializer.

**Solution**
Just allow passing it as argument with current implementation as default. This is also more in line with content converters for ktor server.

### DIFF
--- a/ktor-client/ktor-client-features/ktor-client-json/ktor-client-jackson/jvm/src/io/ktor/client/features/json/JacksonSerializer.kt
+++ b/ktor-client/ktor-client-features/ktor-client-json/ktor-client-jackson/jvm/src/io/ktor/client/features/json/JacksonSerializer.kt
@@ -11,9 +11,9 @@ import io.ktor.http.*
 import io.ktor.http.content.*
 import io.ktor.utils.io.core.*
 
-class JacksonSerializer(block: ObjectMapper.() -> Unit = {}) : JsonSerializer {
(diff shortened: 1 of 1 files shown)


e5l:
Hi @Koriit, thanks for the PR. Merged

cy6erGn0m:
@e5l this caused binary change

```diff
        public fun <init> ()V
-       public fun <init> (Lkotlin/jvm/functions/Function1;)V
-       public synthetic fun <init> (Lkotlin/jvm/functions/Function1;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
+       public fun <init> (Lcom/fasterxml/jackson/databind/ObjectMapper;Lkotlin/jvm/functions/Function1;)V
+       public synthetic fun <init> (Lcom/fasterxml/jackson/databind/ObjectMapper;Lkotlin/jvm/functions/Function1;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
```

Koriit:
```kotlin
val block : ObjectMapper.() -> Unit = {}
JacksonSerializer() // 1
JacksonSerializer {} // 2
JacksonSerializer({}) // 3
JacksonSerializer(block) //4
```
Indeed option 3 and 4 won't compile now. I am not proficient in creating multiplatform code in Kotlin yet, so I don't know if there are any other implications.

I believe option 3 can be ignored, unless ktor follows strict backward compatibility rule. However, I'm not so sure about option 4. It seems to be against the idea behind ktor's configuration of features - layers's argument - however, there are no widespread conventions against such use like in the case of 3rd option.

PR #1460, opened 2019-11-26
Mark exceptions with CopyableThrowable to help stacktrace recovery

e5l:


ktor-client/ktor-client-okhttp/jvm/src/io/ktor/client/engine/okhttp/OkHttpWebsocketSession.kt
```diff
@@ -111,4 +111,10 @@ internal class OkHttpWebsocketSession(
 }
 
 @Suppress("KDocMissingDocumentation")
-class UnsupportedFrameTypeException(frame: Frame) : IllegalArgumentException("Unsupported frame type: $frame")
+class UnsupportedFrameTypeException(
+    private val frame: Frame
```
cy6erGn0m: Not sure how it may work in future _mt_ coroutines

cy6erGn0m:
So, generally saying, having custom exception properties forces us to implement `CopyableThrowable`.

ktor-client/ktor-client-okhttp/jvm/src/io/ktor/client/engine/okhttp/OkHttpWebsocketSession.kt
```diff
@@ -111,4 +111,10 @@ internal class OkHttpWebsocketSession(
...
+) : IllegalArgumentException("Unsupported frame type: $frame"), CopyableThrowable<UnsupportedFrameTypeException> {
+    override fun createCopy(): UnsupportedFrameTypeException? = UnsupportedFrameTypeException(frame).also {
+        it.initCause(this)
```
dmitrievanthony: Just to understand, do we actually need to call `it.initCause(this)`, not `it.initCause(this.cause)`?

ktor-server/ktor-server-core/jvm/src/io/ktor/features/CallId.kt
```diff
@@ -29,7 +30,11 @@ typealias CallIdVerifier = (String) -> Boolean
  * An exception that could be thrown to reject a call due to illegal call id
  * @param illegalCallId that caused rejection
  */
-class RejectedCallIdException(val illegalCallId: String) : IllegalArgumentException()
+class RejectedCallIdException(
+    val illegalCallId: String
+) : IllegalArgumentException(), CopyableThrowable<RejectedCallIdException> {
+    override fun createCopy(): RejectedCallIdException? = RejectedCallIdException(illegalCallId)
```
dmitrievanthony: Did we forget to add `initCause` here?

ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/BaseApplicationResponse.kt
```diff
@@ -239,21 +242,39 @@ abstract class BaseApplicationResponse(override val call: ApplicationCall) : App
      * [OutgoingContent] is trying to set some header that is not allowed for this content type.
      * For example, only upgrade content can set `Upgrade` header.
      */
-    class InvalidHeaderForContent(name: String, content: String) : IllegalStateException("Header $name is not allowed for $content")
+    class InvalidHeaderForContent(
+        private val name: String, private val content: String
```
dmitrievanthony: Looks like this class is public, does this change break compatibility (I mean `public` -> `private` replacement)?

e5l: There were no extra fields in this class before, only constructor params

ktor-client/ktor-client-okhttp/jvm/src/io/ktor/client/engine/okhttp/OkHttpWebsocketSession.kt
e5l: Yep, we should do it as the reflective approach in `kotlinx.coroutines`.

PR #1464, opened 2019-11-27
Optimize JS module import time

ezhaka:
**Subsystem**
Client

**Motivation**
It was doing a lot of unnecessary work when JS module were imported:

Alt text

**Solution**
Just make those properties lazy-initialized.


e5l:
lgtm

PR #1468, opened 2019-12-02
Don't send Authorization after redirect

### LINKED ISSUE #1467: Ktor client resends auth data to redirect location
**Ktor Version and Engine Used**
Ktor HTTP client 1.2.5 with Apache engine.

**Describe the bug**

Enable `followRedirects` for http client and send request with auth data to the host `https://foo/redirect` which redirects to the host `https://bar` with auth data from host `foo`.

**To Reproduce**

```
HttpClient(Apache) {
        followRedirects = true
    }.use { httpClient ->
    httpClient.get<HttpResponse>("https://foo/redirect") {
        basicAuthorization("login", "***")
    }
}
```

Actual result: `Authorization` header data is leaked to the host `https://bar`.

**Expected behavior**
The `Authorization` header from original request to the host `foo` is cleared before establishing redirect request to the host `bar`. It's a good security practice.

**Comment**
It could be solved by cleaning the `Authorization` header in the `HttpRedirect` feature:
https://github.com/ktorio/ktor/blob/master/ktor-client/ktor-client-core/common/src/io/ktor/client/features/HttpRedirect.kt#L39

e5l:
Fix #1467

### DIFF
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/features/HttpRedirect.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/features/HttpRedirect.kt
@@ -7,15 +7,28 @@ package io.ktor.client.features
 import io.ktor.client.*
 import io.ktor.client.call.*
 import io.ktor.client.request.*
-import io.ktor.client.statement.*
 import io.ktor.http.*
 import io.ktor.util.*
-import kotlinx.coroutines.*
+import kotlin.jvm.*
+import kotlin.native.concurrent.*
+
+@ThreadLocal
+private val ALLOWED_FOR_REDIRECT: Set<HttpMethod> = setOf(HttpMethod.Get, HttpMethod.Head)
 
 /**
  * [HttpClient] feature that handles http redirect
  */
 class HttpRedirect {
+    /**
+     * Check if the HTTP method is allowed for redirect.
+     * Only [HttpMethod.Get] and [HttpMethod.Head] is allowed for implicit redirect.
+     *
+     * Please note: changing this flag could lead to security issues, consider changing the request URL instead.
+     */
+    @KtorExperimentalAPI
+    @Volatile
+    var checkHttpMethod: Boolean = true
+
     companion object Feature : HttpClientFeature<Unit, HttpRedirect> {
         override val key: AttributeKey<HttpRedirect> = AttributeKey("HttpRedirect")
 
@@ -23,6 +36,10 @@ class HttpRedirect {
 
         override fun install(feature: HttpRedirect, scope: HttpClient) {
             scope.feature(HttpSend)!!.intercept { origin ->
+                if (feature.checkHttpMethod && origin.request.method !in ALLOWED_FOR_REDIRECT) {
+                    return@intercept origin
+                }
+
                 handleCall(origin)
             }
         }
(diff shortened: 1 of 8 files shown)


ktor-http/common/src/io/ktor/http/URLBuilder.kt
```diff
@@ -160,3 +160,47 @@ data class Url(
...
+            append(it)
+        }
+
+        if (isNotEmpty()) {
+            append('@')
+        }
+
+        append(host)
+        append(':')
+        append(port)
```
cy6erGn0m: Well, we may add `0` here while we don't do the same for `Url` that is always resolved to a specific port

ktor-client/ktor-client-core/common/src/io/ktor/client/features/HttpRedirect.kt
```diff
@@ -31,13 +48,26 @@ class HttpRedirect {
...
                 val location = call.response.headers[HttpHeaders.Location]
 
                 val requestBuilder = HttpRequestBuilder().apply {
                     takeFrom(origin.request)
                     url.parameters.clear()
 
+                    /**
+                     * Disallow redirect with a security downgrade.
+                     */
+                    if (originProtocol.isSecure() && !url.protocol.isSecure()) {
```
cy6erGn0m: We need a switch to be able to allow downgrading

ktor-client/ktor-client-core/common/src/io/ktor/client/features/HttpRedirect.kt
```diff
@@ -7,22 +7,39 @@ package io.ktor.client.features
...
 class HttpRedirect {
+    /**
+     * Check if the HTTP method is allowed for redirect.
+     * Only [HttpMethod.Get] and [HttpMethod.Head] is allowed for implicit redirect.
+     *
+     * Please note: changing this flag could lead to security issues, consider changing the request URL instead.
+     */
+    @KtorExperimentalAPI
+    @Volatile
+    var checkHttpMethod: Boolean = true
```
cy6erGn0m: This may work well with REST API, however, in good old web app, it was a common pattern to do a redirect after posting form. So it is wise to keep this switch.

ktor-client/ktor-client-core/common/src/io/ktor/client/features/HttpRedirect.kt
```diff
@@ -31,13 +48,26 @@ class HttpRedirect {
...
                     url.parameters.clear()
 
+                    /**
+                     * Disallow redirect with a security downgrade.
+                     */
+                    if (originProtocol.isSecure() && !url.protocol.isSecure()) {
+                        return call
+                    }
+
+                    if (originAuthority != url.authority) {
```
cy6erGn0m: Should we only check for auth/host/port or we need to check path as well?

ktor-client/ktor-client-core/common/src/io/ktor/client/features/HttpRedirect.kt
cy6erGn0m: It looks like we need HSTS support here as well

ktor-client/ktor-client-core/common/src/io/ktor/client/features/HttpRedirect.kt
e5l: Path is ok. We should only check for `authority` part of the url: https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Generic_syntax

cy6erGn0m:
I'am going to fix it separately.

ktor-http/common/src/io/ktor/http/URLBuilder.kt
dtretyakov: @cy6erGn0m, you are right, authority should be composed like that:
```
authority = [userinfo@]host[:port]
```
so if the port if default we should not add it at all.
We could cover it by tests as well.

PR #1477, opened 2019-12-09
Fix Apache client engine freeze at request initiation

cy6erGn0m:
**Subsystem**
ktor-client-apache

**Problem and solution**
When sending a request, it is likely that consumeContent callback will
be triggered before inDecoderThread invocation, especially on Windows.
Usually, it is not a problem since we do requestInput. However, if the response
is short or empty or it is the last chunk, then there will be no more callback invocations so there will be no
change for rendezvous. So to avoid this unfortunate case we introduce a decoder
reference that is populated in the callback in the case when no suspension yet
and then we pick it up in decoding loop initiation later.


cy6erGn0m:
It was closed as proved as not fixing issues.

PR #1478, opened 2019-12-10
Bug fixes

e5l:


ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/WebSocketTest.kt
```diff
@@ -0,0 +1,30 @@
...
+class WebSocketTest : ClientLoader() {
+
+    @Test
+    fun testEcho() = clientTests {
+        config {
+            install(WebSockets)
+        }
+
+        test { client ->
+            client.webSocket("ws://127.0.0.1:8080/websockets/echo") {
```
cy6erGn0m: Shouldn't we use `$TEST_SERVER` instead?

e5l: `$TEST_SERVER` has a different protocol. I introduce a constant instead.

Issue #1420, opened 2019-10-31
iOS Memory Leak when KTOR Get Request is Made


Tony-T-R-A-N:
Hello, I am having an issue with iOS memory leaks when I make a ktor get request.
The dispatcher is being built on the iosX64/iosArm32/iosArm32 side through an class implementing CoroutineDispatcher() with the override fun dispatch doing
```
dispatch_async(dispatch_get_main_queue()) {
    block.run()
}
```
```
class DummyClass() : CoroutineScope {
​
	override val coroutineContext: CoroutineContext
        get() = dispatcher
	
	fun execute(suspendFunc: suspend () -> Unit) {
		launch { suspendFunc() }
	}    
​
	fun request(
		parameter: String,
		callback: (MyResponse?, Throwable?) -> Unit
	) {
		execute {
			try {
				val result = httpClient.get<MyResponse> {
					url("PUT A URL HERE")
				}
​
				callback(result, null)
			} catch (e: Throwable) {
				callback(null, e)
			} finally {
				executor.cancel()
			}
		}
	}
	
}
```

bherbst:
I spent some time over the last few days digging into this issue, and believe I have found the issue (and solved it).

### Background
We made a sample app that simply makes 300 API calls in quick succession in order to (roughly) test Ktor's performance and memory profile. 

As it stands today (with Ktor 1.3.0-beta-2), the memory profile looks something like this, with the peak of the graph being roughly 122MB:

![ktor_leak_graph](https://user-images.githubusercontent.com/3267925/70550277-21a98680-1b3b-11ea-9d30-cd83eb077df2.png)

Looking at what is leaking, it appears as though the biggest source is a large number of 96KB mallocs in `SSLBuffer`, which are tracked on the green graph above. Here's a breakdown of some of those allocations and a stack trace:

![ktor_leaked_objects](https://user-images.githubusercontent.com/3267925/70550379-44d43600-1b3b-11ea-8f7b-45b288093524.png)

![ktor_leak_trace](https://user-images.githubusercontent.com/3267925/70550432-5a496000-1b3b-11ea-8c97-caf4f9ba7777.png)

After looking through the call traces for these allocations, it appears as though the entire coroutine used for making the request was leaked, and these objects just happened to be the biggest single object connected to that leak.

### The fix
[...]

PR #1480, opened 2019-12-10
Fix NSUrlSession memory leak by finishing session

bherbst:
**Subsystem**
iOS client engine

**Motivation**
#1420 describes a memory leak in the iOS client. I have posted more details in that issue, but the gist of it is that when opening an `NSUrlSession` using `sessionWithConfiguration()`, the `delegate` is leaked until the app exits unless we call `invalidateAndCancel()` or `finishTasksAndInvalidate()` at some point. See the documentation for the `delegate` parameter [on developer.apple.com](https://developer.apple.com/documentation/foundation/nsurlsession/1411597-sessionwithconfiguration?language=objc) for more information.

In effect, this meant that the engine, the session, and the coroutine used for executing the request are all leaked until the app dies.

**Solution**
After adding the data task to `NSUrlSession`, we immediate call [`finishTasksAndInvalidate()`](https://developer.apple.com/documentation/foundation/nsurlsession/1407428-finishtasksandinvalidate?language=objc), which blocks creation of new tasks but allows the existing data task to complete. Once that data task completes, `NSUrlSession` will then break references to the delegate and callback objects, allowing those objects to become garbage collected.

### DIFF
--- a/ktor-client/ktor-client-ios/darwin/src/io/ktor/client/engine/ios/IosClientEngine.kt
+++ b/ktor-client/ktor-client-ios/darwin/src/io/ktor/client/engine/ios/IosClientEngine.kt
@@ -118,6 +118,7 @@ internal class IosClientEngine(override val config: IosClientEngineConfig) : Htt
 
                 config.requestConfig(nativeRequest)
                 session.dataTaskWithRequest(nativeRequest).resume()
+                session.finishTasksAndInvalidate()
             }
         }
     }
--- a/ktor-client/ktor-client-ios/darwin/src/io/ktor/client/engine/ios/IosClientEngine.kt
+++ b/ktor-client/ktor-client-ios/darwin/src/io/ktor/client/engine/ios/IosClientEngine.kt
@@ -118,6 +118,7 @@ internal class IosClientEngine(override val config: IosClientEngineConfig) : Htt
 
                 config.requestConfig(nativeRequest)
                 session.dataTaskWithRequest(nativeRequest).resume()
+            }.invokeOnCompletion {
                 session.finishTasksAndInvalidate()
             }
         }


cy6erGn0m:
Do you know, if `dataTaskWithRequest` produce additional helper tasks? Shouldn't we invoke the finish function somewhere in the delegates callback instead just before resuming the continuation?

dmitrievanthony:
HI @bherbst, thanks for the PR, seems like it fixes a significant issue. Speaking about the change, do I understand correctly that `NSURLSession` usage should be followed by `invalidateAndCancel()` or `finishTasksAndInvalidate()` in all cases? If it's right I think it makes sense to wrap this usage into `try-finally` block or, maybe even better, add `fun NSURLSession.use { ... }` extension with such a block. What do you think?

cy6erGn0m:
The other alternative would be setting `invalidateAndCancel` on the call context completion if we are unsure about the exact time when to terminate.

bherbst:
I'll start by saying I'm certainly no iOS networking expert, but here's my understanding:

> Do you know, if dataTaskWithRequest produce additional helper tasks? Shouldn't we invoke the finish function somewhere in the delegates callback instead just before resuming the continuation?

The docs for `dataTaskWithRequest` seem to indicate only one is spawned, but `finishTasksAndInvalidate` will wait for _all_ tasks to finish before invalidating, so that shouldn't be an issue. I put it directly after the `resume()` call because my assumption is that the task isn't yet running until we call `resume()` and thus if we tried to `finishTasksAndInvalidate` we would immediately invalidate the session since no tasks are running.

>  do I understand correctly that NSURLSession usage should be followed by invalidateAndCancel() or finishTasksAndInvalidate() in all cases? If it's right I think it makes sense to wrap this usage into try-finally block or, maybe even better, add fun NSURLSession.use { ... } extension with such a block. 

Not necessarily- after talking to some iOS developers my belief is that most iOS developers use NSURLSession a bit differently than KTOR does. Instead of creating a new NSURLSession for each network request, they typically have a singleton session that they reuse for all requests. Since it is being reused, the session can't be invalidated. A `use{ ... }` extension method could be useful though! Would it be more appropriate to `finishTasksAndInvalidate()` or `invalidateAndCancel`?

matarasso:
isn't this the same issue that was resolved here? https://github.com/ktorio/ktor/pull/1342

bherbst:
@matarasso Indeed it looks to be the same! Good find. Unfortunately that PR was only ever merged into ktorio:e5l/develop.

I've updated this PR to move the `finishTasksAndInvalidate()` call to `invokeOnCompletion()` as suggested in the other PR as well, so the two are now identical.

matarasso:
A shame that this couldn't make it to 1.3.0, I've been using my own fork for months because this is a show stopper for an app that downloads hundreds of files such as ours.

PR #1488, opened 2019-12-12
Update to experimental dispatcher; Fix benchmarks

e5l:


ktor-client/ktor-client-core/jvm/src/io/ktor/client/utils/CoroutineDispatcherUtils.kt
```diff
@@ -6,19 +6,14 @@ package io.ktor.client.utils
...
-    return Executors.newFixedThreadPool(threadCount) {
-        Thread(it).apply {
-            isDaemon = true
-            name = threadName.format(threadsNum.getAndIncrement())
-        }
-    }.asCoroutineDispatcher()
-}
+fun Dispatchers.clientDispatcher(
+    threadCount: Int,
+    dispatcherName: String = "thread-pool-%d"
```
dmitrievanthony: I'm not sure "thread-pool-%d" is a good default name for client dispatcher. I think we could avoid it because `ExperimentalCoroutineDispatcher` has it's own default name as far as I understand.

ktor-client/ktor-client-android/jvm/src/io/ktor/client/engine/android/AndroidClientEngine.kt
```diff
@@ -26,7 +26,7 @@ import kotlin.coroutines.*
 class AndroidClientEngine(override val config: AndroidEngineConfig) : HttpClientEngineBase("ktor-android") {
 
     override val dispatcher by lazy {
-        Dispatchers.fixedThreadPoolDispatcher(
+        Dispatchers.clientDispatcher(
             config.threadsCount,
             "ktor-android-thread-%d"
```
dmitrievanthony: Does `ExperimentalCoroutineDispatcher` resolve these `-%d` in names?

e5l:
Merged manually

Issue #1462, opened 2019-11-26
[CIO] java.io.EOFException when parsing response to multi-part form data


serebit:
**Ktor Version and Engine Used (client or server and name)**
1.2.6 and 1.3.0-beta-2, `ktor-client-cio`

**Describe the bug**
About 30% of the time, when sending a multi-part form data request to Discord's servers, the CIO client fails to parse the response sent by the server and ceases to function. The error thrown is `java.io.EOFException: Failed to parse HTTP response: unexpected EOF`, which causes a `java.io.IOException: Connection reset by peer`, which essentially softlocks the HttpClient. **This does not happen with `ktor-client-okhttp`, only CIO.** The trace is here: https://hastebin.com/cikecature.txt

**To Reproduce**
I can't provide an easy method of reproduction, as I'm not sure what part of the response is causing this behavior, and it could be specific to this scenario. If it helps, I can provide a packet capture from Wireshark.

**Expected behavior**
I expected the client to correctly parse the response.

e5l:
Hi @serebit, thanks for the report. Could you provide the request example?

serebit:
The request is a standard HTTP POST request with authorization headers, no URL parameters, and the following body construction:
```kotlin
MultiPartFormDataContent(formData {
    append("file", fileData, Headers.build {
        append(HttpHeaders.ContentDisposition,
            """form-data; name="file"; filename="$fileName"""")
    })
})
```
where `fileData` is a `ByteArray` read directly from an image file (a PNG with transparency).

serebit:
If need be, I can provide a simplified single-block example of the call, although it won't be runnable due to how the Discord HTTP API works.

e5l:
I failed to reproduce the issue. It looks like the issue is in the specific response parsing. It would be nice if you can provide some request/response captures for that case, but I continue investigation.

dmitrievanthony:
Hi folks,

I spent some time and prepared the code that reproduces the problem, you can find it [here](https://github.com/dmitrievanthony/KPlayground/blob/%231462/src/main/kotlin/app.kt). It works fine with `OkHttp` and `Android` clients but fails with `CIO` with the following stack trace:

```
Exception in thread "main" java.io.EOFException: Failed to parse HTTP response: unexpected EOF
	at io.ktor.client.engine.cio.UtilsKt.readResponse(utils.kt:81)
[...]

Issue #1456, opened 2019-11-22
CIO client half-closes the TCP connection after the request send


rocketraman:
**Subsystem**
Client, CIO

**Is your feature request related to a problem? Please describe.**
It is to some extent. I had an issue with an outbound proxy when using the CIO client. Upon investigation of the packet dump, I saw that the CIO client half-closes the TCP connection immediately after the request is sent. Upon further research, it appears there is debate in the HTTP community about whether this is "legal" or not for HTTP clients to do, but consensus that for best compatibility its probably not a good idea [1].

[1] See https://github.com/httpwg/http-core/issues/22.

**Describe the solution you'd like**
Do not half-close client connections once the request has been sent. Leave the connection open until after the response has been received.

**Motivation to include to ktor**
As per the linked discussion above in the HTTP WG, it is odd for HTTP clients to do this, and the WG recommends that clients do not do it.

dmitrievanthony:
Hi @rocketraman,

I've just merged fix that allows avoiding TCP half-closes (see https://github.com/ktorio/ktor/pull/1489), so it will be included in `1.3.0`.

PR #1489, opened 2019-12-12
Make CIO TCP half close behavior optional and disabled by default.

dmitrievanthony:
**Subsystem**
Client, CIO

**Motivation**
In the case of using `WriteChannelContent` CIO client closes the output channel and so triggers TCP connection half-close. Although this behavior is correct from the protocol perspective it's not encouraged in some cases and is not supported by all servers (Netty for example).

There are several issues caused by current implementation: #1456 and #1462.

**Solution**
This change makes TCP half-close optional and disabled by default.

### DIFF
--- a/ktor-client/ktor-client-cio/jvm/src/io/ktor/client/engine/cio/CIOEngineConfig.kt
+++ b/ktor-client/ktor-client-cio/jvm/src/io/ktor/client/engine/cio/CIOEngineConfig.kt
@@ -71,4 +71,11 @@ class EndpointConfig {
      * Maximum number of connection attempts.
      */
     var connectRetryAttempts: Int = 5
+
+    /**
+     * Allow socket to close output channel immediately on writing completion (TCP connection half close).
+     */
+    var allowHalfClose: Boolean = false
+        @Deprecated("Half closed TCP connection is not supported by all servers, use it at your own risk.")
+        set
 }
--- a/ktor-client/ktor-client-cio/jvm/src/io/ktor/client/engine/cio/utils.kt
+++ b/ktor-client/ktor-client-cio/jvm/src/io/ktor/client/engine/cio/utils.kt
@@ -13,69 +13,76 @@ import io.ktor.http.cio.*
 import io.ktor.http.cio.websocket.*
 import io.ktor.http.content.*
 import io.ktor.util.date.*
-import kotlinx.coroutines.*
 import io.ktor.utils.io.*
 import io.ktor.utils.io.errors.*
+import kotlinx.coroutines.*
 import kotlin.coroutines.*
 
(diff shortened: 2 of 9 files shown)


ktor-client/ktor-client-cio/jvm/src/io/ktor/client/engine/cio/CIOEngineConfig.kt
```diff
@@ -71,4 +71,11 @@ class EndpointConfig {
      * Maximum number of connection attempts.
      */
     var connectRetryAttempts: Int = 5
+
+    /**
+     * Allow socket to close output channel immediately on writing completion (TCP connection half close).
+     */
+    var allowHalfClose: Boolean = false
```
e5l: Let's turn it on by default

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/MultiPartFormDataTest.kt
```diff
@@ -0,0 +1,39 @@
...
+                        "file",
+                        ByteArray(1024 * 1024),
+                        Headers.build {
+                            append(HttpHeaders.ContentDisposition, """form-data; name="file"; filename="test.png"""")
+                        }
+                    )
+                })
+            }.execute()
+
+            check(result.status == HttpStatusCode.OK)
```
e5l: assertEquals

ktor-client/ktor-client-cio/jvm/src/io/ktor/client/engine/cio/utils.kt
```diff
@@ -125,3 +132,29 @@ internal suspend fun readResponse(
...
+    if (closeOnCoroutineCompletion) {
+        // Pure output represents a socket output channel that is closed when request fully processed or after
+        // request sent in case TCP half-close is allowed.
+        coroutineContext[Job]!!.invokeOnCompletion {
+            close()
+        }
+    }
+
+    return GlobalScope.reader(coroutineContext, autoFlush = true) {
+        this.channel.copyTo(this@withoutClosePropagation, Long.MAX_VALUE)
```
e5l: this

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/MultiPartFormDataTest.kt
dmitrievanthony: Done.

ktor-client/ktor-client-cio/jvm/src/io/ktor/client/engine/cio/utils.kt
dmitrievanthony: Done.

Issue #1498, opened 2019-12-14
The ApplicationStopPreparing event is fired several times in withTestApplication


BorzdeG:
**Ktor Version and Engine Used (client or server and name)**
version, client/server, additional ktor modules included
version: master branch (

**To Reproduce**
Demo test added to PR: `io.ktor.tests.server.application.ApplicationEventTest`

**Expected behavior**
ApplicationStopPreparing event is fired once

cy6erGn0m:
Fixed in 1.3.0-rc2

PR #1499, opened 2019-12-14
Added test for #1498

BorzdeG:
Added test for #1498 

> The ApplicationStopPreparing event is fired several times in withTestApplication


cy6erGn0m:
Merged as per #1514

PR #1500, opened 2019-12-15
Add @SharedImmutable over a readonly private var to avoid exceptions

GoatHunter:
Add @SharedImmutable over a readonly private var to avoid exceptions when used in multithreading

**Subsystem**
Client core.

**Motivation**
Encountered when making a request on iOS from another thread.

**Solution**
SharedImmutable should prevent the exception thrown by Kotlin when this var gets accessed on another thread.


e5l:
Thanks for the PR. LGTM

Issue #1505, opened 2019-12-17
ktor-client-mock engine is not thread safe


haneev:
**Ktor Version and Engine Used (client or server and name)**
Ktor Client 1.2.5 used for all modules:
- ktor-client-core-jvm
- ktor-client-apache
- ktor-client-json-jvm
- ktor-client-jackson

**Describe the bug**
I get a `IndexOutOfBoundsException` or a `Unhandled http://x...` error.

Line 41 of MockEngine.kt gives a IndexOutOfBoundsException, but also line 40 sometimes happens, then I get an `Unhandled http://x...` error. Both errors indicate that there is something wrong with the `invocationCount` and is mutated at different locations. And i can imagine that also the `requestHistory` is not thread safe.

In my test we execute multiple calls on the HttpClient at the same time using coroutines. I therefore suspect that the mock engine is not thread safe, because `invocationCount` is mutated at different locations (line 41/43). When I run my unit tests it will give me this error about 1 in 100. 

https://github.com/ktorio/ktor/blob/master/ktor-client/ktor-client-mock/common/src/io/ktor/client/engine/mock/MockEngine.kt#L41

Engine is created using the following snippet:

```
val httpClient = HttpClient(MockEngine) {
    engine {
        addHandler {
            respond("{}", HttpStatusCode.OK, headers = headersOf("Content-Type" to listOf(ContentType.Application.Json.toString())))
        }
    }
    install(JsonFeature) {
        serializer = JacksonSerializer()
    }
}
```

**To Reproduce**
The following test can show the behavior.

```
for (i in 0..100) {
    launch { httpClient.get<HttpResponse> { url(request.endpoint) } }
}
```

**Expected behavior**
No crash :P

haneev:
@cy6erGn0m If you agree with my fix in my branch, I can open a PR?

cy6erGn0m:
@haneev Yes, open a PR, please

PR #1510, opened 2019-12-18
Improve ktor-client-mock engine to be thread safe #1505

haneev:
**Subsystem**
Http Mock Engine

**Motivation**
#1505 InvocatioCounter was not thread safe for http client mock engine

**Solution**
Add a lock surrounding the important code paths. I did not make use of SynchronizedLists because i want only protect some parts of the code. For instance reading history is not required to be thread safe.


cy6erGn0m:
Thanks, merged!

PR #1511, opened 2019-12-18
Add Application/Problem Content type RFC 7808

devmachiine:
**Subsystem**
Relevant to both Client/Server

**Motivation**
RFC 7808 provides a way to add more problem Details for HTTP APIs than just an HTTP status code.

See example here:
https://tools.ietf.org/html/rfc7807#section-3

**Solution**
Added both IETF content types.


cy6erGn0m:
Merged, thanks!

PR #1513, opened 2019-12-20
Several minor fixes

cy6erGn0m:
**Subsystem**
Client and server

**Motivation**
There are a lot of warnings. 
Several places when we do upper/lowercase transformation using locale rules that may potentially break HTTP-related logic. 

**Solution**
- Cleanup warnings
- Avoid using toLowerCase/toUppwerCase, introduce internal utility function that preserves ASCII characters rules
- Avoid accidentally boxed precomputed lookup tables

### DIFF
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/call/HttpClientCall.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/call/HttpClientCall.kt
@@ -58,10 +58,10 @@ open class HttpClientCall internal constructor(
         internal set
 
     /**
-     * Tries to receive the payload of the [response] as an specific [expectedType].
-     * Returns [response] if [expectedType] is [HttpResponse].
+     * Tries to receive the payload of the [response] as a specific expected type provided in [info].
+     * Returns [response] if [info] corresponds to [HttpResponse].
      *
-     * @throws NoTransformationFoundException If no transformation is found for the [expectedType].
+     * @throws NoTransformationFoundException If no transformation is found for the type [info].
      * @throws DoubleReceiveException If already called [receive].
      */
     suspend fun receive(info: TypeInfo): Any {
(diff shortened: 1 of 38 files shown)


ktor-client/ktor-client-core/common/src/io/ktor/client/call/HttpClientCall.kt
```diff
@@ -158,7 +159,7 @@ class DoubleReceiveException(call: HttpClientCall) : IllegalStateException() {
  * Exception representing fail of the response pipeline
  * [cause] contains origin pipeline exception
  */
-@Suppress("KDocMissingDocumentation")
+@Suppress("KDocMissingDocumentation", "unused")
```
dmitrievanthony: Why `unused` is in uppercase somewhere and in lowercase here?

ktor-utils/common/src/io/ktor/util/Text.kt
```diff
@@ -39,6 +39,66 @@ inline fun String.chomp(separator: String, onMissingDelimiter: () -> Pair<String
...
+        append(original, 0, firstIndex)
+
+        for (index in firstIndex .. original.lastIndex) {
+            append(toUpperCasePreservingASCII(original[index]))
+        }
+    }
+}
+
+private fun toLowerCasePreservingASCII(ch: Char): Char = when (ch) {
+    in 'A' .. 'Z' -> ch + 32
```
dmitrievanthony: Do we actually need here two different branches for `A` .. `Z` and else?

cy6erGn0m: Yes, this is the point: we always do lowercase manually for A-Z. The else block does `toLowerCase` that for letters in range A-Z may produce characters other than a-z.

ktor-client/ktor-client-core/common/src/io/ktor/client/call/HttpClientCall.kt
cy6erGn0m: Well.. we have a lot of both in the codebase. Not sure exactly why

ktor-utils/common/src/io/ktor/util/Text.kt
cy6erGn0m: This is not obvious but it may happen due to locale-specific rules

cy6erGn0m:
Merged

PR #1514, opened 2019-12-20
Fix ApplicationStopPreparing fired multiple times with test server engine

cy6erGn0m:
**Subsystem**
Server, test engine

**Motivation**
Issue #1498 

**Solution**
Apply the same solution as we do for other server engines: we simply launch a coroutine that is always sleeping and resuming at cancellation and doing stop.


cy6erGn0m:
Merged

PR #1516, opened 2019-12-20
Fix dropLeadingTopDirs to not hang at some cases

cy6erGn0m:
**Subsystem**
Server

**Motivation**
Currently, dropLeadingTopDir that is used in several path resolution functions may hang at some inputs.

**Solution**
Fix bug, add tests

### DIFF
--- a/ktor-utils/jvm/src/io/ktor/util/Path.kt
+++ b/ktor-utils/jvm/src/io/ktor/util/Path.kt
@@ -44,29 +44,48 @@ private fun File.notRooted(): File {
     return File(path.drop(current.name.length).dropWhile { it == '\\' || it == '/' })
 }
 
-private fun File.dropLeadingTopDirs(): File {
+/**
+ * Discards all leading path separators, top dir (..) and current dir (.) references.
+ * The only exception is when the whole path is a current dir reference (.)
+ */
+internal fun dropLeadingTopDirs(path: String): Int {
     var startIndex = 0
-    val path = path!!
     val lastIndex = path.length - 1
 
-    while (startIndex < lastIndex) {
-        if (path[startIndex] == '.') {
-            val second = path[startIndex + 1]
-            if (second == '\\' || second == '/') {
-                startIndex += 2 // skip 2 characters: ./ or .\
-            } else if (second == '.') {
-                if (startIndex + 2 == path.length) {
-                    startIndex += 2 // skip the only 2 characters remaining: ..
-                } else if (path[startIndex + 2].let { it == '/' || it == '\\' }) {
-                    startIndex += 3 // skip 3 characters: ../ or ..\
-                }
-            } else {
+    while (startIndex <= lastIndex) {
+        val first = path[startIndex]
+        if (first.isPathSeparator()) {
+            startIndex++
+            continue
+        }
+        if (startIndex == lastIndex || first != '.') {
+            break
+        }
+
+        val second: Char = path[startIndex + 1]
+        if (second.isPathSeparator()) {
+            startIndex += 2 // skip 2 characters: ./ or .\
+        } else if (second == '.') {
+            if (startIndex + 2 == path.length) {
+                startIndex += 2 // skip the only 2 characters remaining: ..
+            } else if (path[startIndex + 2].isPathSeparator()) {
+                startIndex += 3 // skip 3 characters: ../ or ..\
+            } else { // we have a path component starting with two dots that shouldn't be discarded
                 break
             }
-        } else {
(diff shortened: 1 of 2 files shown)


ktor-utils/jvm/test/io/ktor/tests/utils/DropLeadingTopDirsTest.kt
```diff
@@ -0,0 +1,59 @@
...
+    fun testEmptyPath() {
+        assertEquals(0, dropLeadingTopDirs(""))
+        assertEquals(0, dropLeadingTopDirs(" "))
+        assertEquals(1, dropLeadingTopDirs("/"))
+        assertEquals(1, dropLeadingTopDirs("\\"))
+    }
+
+    @Test
+    fun testSingleDot() {
+        assertEquals(0, dropLeadingTopDirs("."))
```
dmitrievanthony: Dot is just a pointer to the current directory as well as `..` is a pointer to parent. It's not clear why the behavior is different, why we left `.`, but remove `..`.

ktor-utils/jvm/src/io/ktor/util/Path.kt
```diff
@@ -44,30 +44,49 @@ private fun File.notRooted(): File {
     return File(path.drop(current.name.length).dropWhile { it == '\\' || it == '/' })
 }
 
-private fun File.dropLeadingTopDirs(): File {
+/**
+ * Discards all leading path separators, top dir (..) and current dir (.) references.
+ * The only exception is when the whole path is a current dir reference (.)
+ */
+internal fun dropLeadingTopDirs(path: String): Int {
```
dmitrievanthony: Is it possible to make it using `Paths` API? It looks like this code should look much simpler if we use this API.

ktor-utils/jvm/test/io/ktor/tests/utils/DropLeadingTopDirsTest.kt
```diff
@@ -0,0 +1,59 @@
...
+        assertEquals(2, dropLeadingTopDirs(".."))
+        assertEquals(3, dropLeadingTopDirs("../a"))
+        assertEquals(3, dropLeadingTopDirs("..\\a"))
+    }
+
+    @Test
+    fun testDoubleTopDir() {
+        assertEquals(5, dropLeadingTopDirs("../.."))
+        assertEquals(5, dropLeadingTopDirs("..\\.."))
+        assertEquals(6, dropLeadingTopDirs("../../"))
```
dmitrievanthony: Do we handle `/` and `\` similarly for all platforms? For UNIX it's OK to have a file with `\` in name:

```
$ cat > ..\..
hello
^C
$ cat ..\..
hello
```

ktor-utils/jvm/src/io/ktor/util/Path.kt
cy6erGn0m: No, we should avoid using Paths API because of Android.

ktor-utils/jvm/test/io/ktor/tests/utils/DropLeadingTopDirsTest.kt
cy6erGn0m: Because `.` is safe and the only dot is not a problem at all. However, parent dir dots should be avoided to get directory "sandbox".

ktor-utils/jvm/test/io/ktor/tests/utils/DropLeadingTopDirsTest.kt
cy6erGn0m: We probably could but for simplicity and with lack of real use-cases, it is a common pattern to treat both as path separators.

ktor-utils/jvm/test/io/ktor/tests/utils/DropLeadingTopDirsTest.kt
cy6erGn0m: Also note that this function is always called after path normalization so we don't need to look deep in a path since `..` could be only at the beginning.

cy6erGn0m: The other reason to keep the single dot is that an empty path is questionable while the dot-path is for sure valid.

cy6erGn0m: Fixed to remove single dot as well. An empty path result is ok.

cy6erGn0m:
Merged manually including fix of concerns.

Issue #1519, opened 2019-12-23
KotlinxSerializer unnable to serialize JsonElement/Map<String, *>


varahash:
**Ktor Version and Engine Used (client or server and name)**
kotlin: 1.3.61
kotlinx-serialization-runtime: 0.14.0
ktor-client: 1.3.0-rc

**Describe the bug**
Setup HttpClient with `JsonFeature` and serializer `KotlinxSerializer`.
When running following request:
```kotlin
import kotlinx.serialization.json.json

httpClient.post<String>("http://example.com") {
    contentType(ContentType.Application.Json)
    body = json {
        "key1" to "abc"
        "map" to json {
            "key2" to "abc"
        }
    }
}
```
got following exception:
```
java.lang.ClassCastException: kotlinx.serialization.json.JsonObject cannot be cast to kotlinx.serialization.json.JsonLiteral

	at kotlinx.serialization.json.JsonLiteralSerializer.serialize(JsonElementSerializer.kt:103)
	at kotlinx.serialization.json.internal.StreamingJsonOutput.encodeSerializableValue(StreamingJsonOutput.kt:228)
	at kotlinx.serialization.ElementValueEncoder.encodeSerializableElement(ElementWise.kt:72)
	at kotlinx.serialization.internal.MapLikeSerializer.serialize(CollectionSerializers.kt:134)
	at kotlinx.serialization.json.internal.StreamingJsonOutput.encodeSerializableValue(StreamingJsonOutput.kt:228)
	at kotlinx.serialization.CoreKt.encode(Core.kt:74)
	at kotlinx.serialization.json.Json.stringify(Json.kt:95)
	at io.ktor.client.features.json.serializer.KotlinxSerializer.write(KotlinxSerializer.kt:63)
	at io.ktor.client.features.json.JsonFeature$Feature$install$1.invokeSuspend(JsonFeature.kt:101)
	at io.ktor.client.features.json.JsonFeature$Feature$install$1.invoke(JsonFeature.kt)
	at io.ktor.util.pipeline.SuspendFunctionGun.loop(PipelineContext.kt:273)
	at io.ktor.util.pipeline.SuspendFunctionGun.proceed(PipelineContext.kt:141)
	at io.ktor.util.pipeline.SuspendFunctionGun.proceedWith(PipelineContext.kt:151)
	at io.ktor.client.features.HttpCallValidator$Companion$install$1.invokeSuspend(HttpCallValidator.kt:87)
[...]

PR #1521, opened 2019-12-24
Introduce support for json structures and improve diagnostics

cy6erGn0m:
**Subsystem**
client and server

**Motivation**
Currently we do not support `json {}` structures from kotlinx.serialization. See #1519 

**Solution**
- Use `JsonElementSerializer` when applicable
- Improve diagnostics to produce an error when non-uniform collections are serialized

### DIFF
--- a/ktor-client/ktor-client-features/ktor-client-json/ktor-client-serialization/common/src/io/ktor/client/features/json/serializer/KotlinxSerializer.kt
+++ b/ktor-client/ktor-client-features/ktor-client-json/ktor-client-serialization/common/src/io/ktor/client/features/json/serializer/KotlinxSerializer.kt
@@ -10,6 +10,7 @@ import io.ktor.http.*
 import io.ktor.http.content.*
 import io.ktor.utils.io.core.*
 import kotlinx.serialization.*
+import kotlinx.serialization.internal.*
 import kotlinx.serialization.json.*
 import kotlin.reflect.*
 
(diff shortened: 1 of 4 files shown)


ktor-client/ktor-client-features/ktor-client-json/ktor-client-serialization/common/src/io/ktor/client/features/json/serializer/KotlinxSerializer.kt
```diff
@@ -92,5 +94,25 @@ private fun buildSerializer(value: Any): KSerializer<*> = when (value) {
 }
 
 @UseExperimental(ImplicitReflectionSerializer::class)
-private fun Collection<*>.elementSerializer(): KSerializer<*> =
-    firstOrNull()?.let { buildSerializer(it) } ?: String.serializer()
+private fun Collection<*>.elementSerializer(): KSerializer<*> {
+    val serializers = filterNotNull().map { buildSerializer(it) }.distinctBy { it.descriptor.name }
+
+    if (serializers.size > 1) {
+        error("Different serializers selected: ${serializers.map { it.descriptor.name }}")
```
qwwdfsad: Not sure this is the most approachable error message.
Something like "Collection has heterogeneous elements that cannot be serialized by a single serializer"
Also, it is not clear how to fix this problem on the user's side

cy6erGn0m:
Merged

Issue #1515, opened 2019-12-20
Throw an error ktor-auth-jwt validate function if null or default value is returned


brendanw:
**Subsystem**
ktor-auth-jwt

**Is your feature request related to a problem? Please describe.**
It is non-obvious that the function passed to validate should return an object when the user is successfully authenticated. 

**Describe the solution you'd like**
The validate block should throw a NotImplementedError if a null or default type is returned.

**Motivation to include to ktor**
I lost quite a bit of time walking through the source figuring out how this middleware works and is to be used.

PR #1523, opened 2019-12-24
Make default auth validate functions fail to force users to implement them

cy6erGn0m:
Make default auth validate functions fail to force users to implement them

**Subsystem**
Server, ktor-auth, ktor-auth-jwt

**Motivation**
Currently we simply always forbid access if the validation function is not specified. We should throw an exception in all auth providers by default to notice users not implementing the function that is always required. Also see #1515 

**Solution**
Simply replace the default function `{ null }` with `{ error(...descriptive message...) }`


cy6erGn0m:
Merged.

PR #1527, opened 2019-12-25
Use a property from Gradle plugin to force optimized test binaries

homuroll:


cy6erGn0m:
@homuroll Thanks for the hint. What plugin version is required? Does it work with 1.3.61 or 1.3.70-eap*?

homuroll:
@cy6erGn0m I think, should work with both.

PR #1529, opened 2019-12-26
Fix Apache client bug

cy6erGn0m:
**Subsystem**
ktor-client-apache

**Motivation**
Apache consumers decoding loop was waiting for -1 returned from decoder that means that it is completed. Return value 0 was treated as incomplete so the consumer loop goes to suspend. However, when the last chunk of size 0 is decoded, a decoder returns 0 and mark as completed. So suspension will be never resumed.

**Solution**
To avoid this, we should check for completion as well so if we got 0 because of completion, we exit loop immediately.


cy6erGn0m:
Merged

PR #1531, opened 2019-12-26
Add CIOApplicationRequest.remoteHost implementation.

programingjd:
**Subsystem**

**Motivation**

**Solution**
Describe your solution.


ktor-http/ktor-http-cio/jvm/src/io/ktor/http/cio/RemoteHost.kt
```diff
@@ -0,0 +1,27 @@
+package io.ktor.http.cio
+
+import io.ktor.network.sockets.*
+import java.net.*
+import java.nio.channels.*
+
+fun remoteHost(socket: AConnectedSocket) = remoteHost(socket.remoteAddress)
```
cy6erGn0m: These functions look like they construct a remote host. More Kotlinish would be an extension property returning String

ktor-http/ktor-http-cio/jvm/src/io/ktor/http/cio/RemoteHost.kt
```diff
@@ -0,0 +1,27 @@
...
+
+fun remoteHost(socketChannel: SocketChannel) = remoteHost(socketChannel.remoteAddress)
```
cy6erGn0m: I believe this one and the functions beyond shouldn't be public

ktor-http/ktor-http-cio/jvm/src/io/ktor/http/cio/RemoteHost.kt
cy6erGn0m: Also, since it is a public API, it should be documented properly describing what does it return, when it is hostname or a stringified IP address.

ktor-http/ktor-http-cio/jvm/src/io/ktor/http/cio/RemoteHost.kt
```diff
@@ -0,0 +1,27 @@
...
+fun remoteHost(address: InetAddress) = when (address) {
+    is Inet4Address -> address.address.joinToString(".") {
+        (it.toInt() and 0xFF).toString()
+    }
+    is Inet6Address -> address.address.toList().chunked(2).joinToString(":") { pair ->
+        val first = pair.first().toInt() and 0xFF
+        val last = pair.last().toInt() and 0xFF
+        (first * 256 + last).toString(16)
+    }
+    else -> ""
```
cy6erGn0m: Well, not sure about it. An empty host is definitely not a good result. We could fail with some description or return `null`

ktor-http/ktor-http-cio/jvm/src/io/ktor/http/cio/RemoteHost.kt
```diff
@@ -0,0 +1,27 @@
...
+
+fun remoteHost(socketAddress: SocketAddress): CharSequence {
+    return (socketAddress as? InetSocketAddress)?.address?.let { remoteHost(it) } ?: ""
+}
+
+fun remoteHost(address: InetAddress) = when (address) {
+    is Inet4Address -> address.address.joinToString(".") {
+        (it.toInt() and 0xFF).toString()
+    }
+    is Inet6Address -> address.address.toList().chunked(2).joinToString(":") { pair ->
```
cy6erGn0m: Creating so many instances for every request may significantly slowdown server at high rates because of high memory pressure. Unfortunately, this needs to be rewritten in a more optimal manner. Btw, doesn't `Inet6Address` already have a proper formatting function? What about `Inet6Address.getHostAddress`?

ktor-server/ktor-server-cio/jvm/src/io/ktor/server/cio/HttpServer.kt
```diff
@@ -14,6 +14,7 @@ import io.ktor.server.engine.*
 import io.ktor.util.*
 import kotlinx.coroutines.*
 import org.slf4j.*
+import java.net.*
```
cy6erGn0m: Why do we need this import?

ktor-http/ktor-http-cio/common/src/io/ktor/http/cio/RequestResponse.kt
```diff
@@ -42,6 +42,7 @@ class Request internal constructor(
     val method: HttpMethod,
     val uri: CharSequence,
     val version: CharSequence,
+    val remoteHost: CharSequence,
```
cy6erGn0m: Not sure about it: can't we pass it somehow else? Perhaps we can pass some "client info" in `HttpRequestHandler` instead. I would prefer to group input, output and connection info in a "client info"

programingjd:
I have changed quite a few things after reading your comments.
The socket address is no longer converted to a CharSequence early on, it's only converted to a String in the remoteHost accessor of CIOApplicationRequest.RequestConnectionPoint
The string conversion is now consistant with what was already done in the netty implementation.
The implementation was missing from the http2 version for netty, so I have added the implementation there too.
Finally, I have added a test in the EngineTestSuite.

cy6erGn0m:
Compilation failed: 
> e: /mnt/agent/work/a85294440dc5c6e/ktor-http/ktor-http-cio/jvm/src/io/ktor/http/cio/RemoteHost.kt: (9, 73): Unresolved reference: remoteAddress

Try with older java

programingjd:
I reverted all my changes and made new commits for cleaner commits.
I also compiled with java 8 this time.

cy6erGn0m:
oops, @programingjd , could you please run the following and commit a change

> ./gradlew clean binary-compatibility-validator:jvmTest -Poverwrite.output=true

programingjd:
> oops, @programingjd , could you please run the following and commit a change
> 
> > ./gradlew clean binary-compatibility-validator:jvmTest -Poverwrite.output=true

Done

cy6erGn0m:
@programingjd My bad, sorry: the recommendation to use `InetSocketAddress.getHostString` was wrong since it doesn't work with older Androids (requires Android API 19 that is too high). Looks like we need to revert your handmade implementation for this.

programingjd:
I looked at the implementation of hostString and hostname ?: address.hostAddress should do the same thing, so I replaced all 3 calls to hostString to that.

cy6erGn0m:
This will be merged as per #1593

PR #1535, opened 2019-12-28
Implementing generateNonce() for iOS

SebOh:
While checking, if ktor already supports websockets on iOS I realized this method is not implemented yet.

I implemented this method based on the another commit here https://gist.github.com/tidwall/475bd2dd0a37d8195d7830ed0ef65e6a
(Row 1039-1043)

### DIFF
--- a/ktor-utils/posix/src/io/ktor/util/CryptoNative.kt
+++ b/ktor-utils/posix/src/io/ktor/util/CryptoNative.kt
@@ -5,7 +5,7 @@
 package io.ktor.util
 
 import kotlinx.cinterop.*
-import platform.Security.*
+import platform.posix.*
 
 private const val NONCE_SIZE_IN_BYTES = 16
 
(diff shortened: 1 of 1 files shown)


ktor-utils/posix/src/io/ktor/util/CryptoNative.kt
```diff
@@ -4,12 +4,31 @@
...
 @InternalAPI
-actual fun generateNonce(): String = error("[generateNonce] is not supported on iOS")
+actual fun generateNonce(): String {
+    return hex(generateRandomByteArray())
+}
+
+private fun generateRandomByteArray(): ByteArray {
+    val result = ByteArray(NONCE_SIZE_IN_BYTES)
+    for (i in 0..3) {
+        val random = arc4random().toLong()
```
cy6erGn0m: Is `arc4random()` still cryptographically nowadays?

SebOh: So far I can only find the following online: https://stackoverflow.com/questions/44751523/is-there-a-cryptographically-secure-alternative-to-arc4random-on-ios 

An alternative could be https://developer.apple.com/documentation/security/1399291-secrandomcopybytes

SebOh:
Rewrote arc4random() to SecRandomCopyBytes() which is recommended for security reasons.

ktor-utils/posix/src/io/ktor/util/CryptoNative.kt
```diff
@@ -4,12 +4,37 @@
...
-actual fun generateNonce(): String = error("[generateNonce] is not supported on iOS")
+actual fun generateNonce(): String {
+    return hex(generateNonceByteArray())
+}
+
+private fun generateNonceByteArray(): ByteArray {
+    memScoped {
+        val array = allocArray<ByteVarOf<Byte>>(NONCE_SIZE_IN_BYTES)
+        val status = SecRandomCopyBytes(
+            kSecRandomDefault,
```
cy6erGn0m: What is it's threading model? Can we use it outside of the main event queue? If yes, can we do it concurrently?

ktor-utils/posix/src/io/ktor/util/CryptoNative.kt
```diff
@@ -4,12 +4,37 @@
...
+            array
+        )
+        if (status == errSecSuccess) {
+            val result = ByteArray(NONCE_SIZE_IN_BYTES)
+            for (i in 0 until NONCE_SIZE_IN_BYTES) {
+                result[i] = array[i]
+            }
+            return result
+        }
+        error("Nonce could not be generated")
```
cy6erGn0m: Can we have some more debug info here? It is slightly strange that the documentation doesn't state what kind of errors may happen and what do they mean

cy6erGn0m:
This is placed in the POSIX source set that is wrong: should be in the Darwin source set

cy6erGn0m:
At least a sanity test is required just to ensure that it works and doesn't crash

ktor-utils/posix/src/io/ktor/util/CryptoNative.kt
SebOh: I checked the web a bit and found the following: 
[https://stackoverflow.com/questions/27011825/random-256bit-key-using-secrandomcopybytes-in-ios](https://stackoverflow.com/questions/27011825/random-256bit-key-using-secrandomcopybytes-in-ios
)

In general I don't see this failing. But to have more feedback I added the result code to the error message for easier debugging.

ktor-utils/posix/src/io/ktor/util/CryptoNative.kt
SebOh: It can be called from outside the main event queue and also concurrently. Should I provide a sample for this?

SebOh:
> This is placed in the POSIX source set that is wrong: should be in the Darwin source set

I totally agree. But in the main build.gradle it says, that 

> Don't create `posix` and `darwin` sourceSets in single project.

I'm not sure how to setup the darwin source set now. Could you help me on this? @cy6erGn0m

SebOh:
I rewrote the randomByteArray generator in a way, that we use the /dev/urandom on the system. So it could be used for linux and macos without any trouble.

urandom is also non blocking and can be called from background threads.

https://en.wikipedia.org/wiki//dev/random

PR #1539, opened 2019-12-31
Fix potential TLS uninitialized property access

cy6erGn0m:
**Subsystem**
CIO TLS

**Motivation**
The cio-tls-parser coroutine may crash in the case when it receives ServerHelloDone, ChangeCipherSpec and Finished too quickly because we construct data required for `cipher` during ServerDone processing while the parser coroutine may continue parsing and may switch `useCipher` too early.

**Solution**
Defer cipher and add wait when needed to ensure that we have constructed cipher.


cy6erGn0m:
I see, according to the RFC this should never happen

Issue #84, opened 2017-04-07
Request TwigFeature 


jmfayard:
It would be great to be able to write `twig` templates (not a fan of freemarker :)

https://github.com/jtwig/jtwig

orangy:
Can you may be contribute support for it? It shouldn't be too hard.

jmfayard:
I will give it a try

orangy:
Did you get it working?

jmfayard:
I made a quick local test and looked how free marker support was implemented. That looks straightforward enough. I think I can work on that before the end of the week

orangy:
Great!

BoD:
@jmfayard I'm looking at Jtwig right now and was wondering if you got the chance to make a PR in the end?

PR #1545, opened 2020-01-03
Add support for Jtwig template engine

pmendelski:
**Subsystem**
It's a server feature related with templates.

**Motivation**
Adds support for [Jtwig](http://jtwig.org/), another template engine. Related issue: https://github.com/ktorio/ktor/issues/84

**Solution**
Render templates using [jtwig-core](https://github.com/jtwig/jtwig-core) library. 
Template prefixes and suffixes are not supported out of the box in jtwig-core so I implemented it in the feature. Documentation: https://github.com/ktorio/ktorio.github.io/pull/247


cy6erGn0m:
Should't we use Pebble instead?

pmendelski:
True, Pebble seems to be in a [better shape](https://java.libhunt.com/compare-pebble-vs-jtwig-jtwig). Closing.

PR #1547, opened 2020-01-05
Fix CIO headers and chunked parsers

cy6erGn0m:
**Subsystem**
ktor-server-cio, ktor-client-cio

**Motivation**
1. If a header is too long, then CIO parser simply cut it into two that is not correct.
2. If both content length and chunked encoding specified, then we discard content length. This is correct according to RFC. However, this may lead to undesired consequences. 
Both may cause potential security issues, especially with a proxy server.

**Solution**
1. Terminate request processing when a header body is too long
2. Terminate request processing AND connection when the specified content length and decoded chunked content length do not match. Note that this violates RFC however it looks like we have no choice.


e5l:
Commit names

cy6erGn0m:
Merged

Issue #1542, opened 2020-01-01
Executing a HttpStatement in a suspending main results in NPE


BartArys:
**Ktor Version and Engine Used (client or server and name)**
io.ktor:ktor-client-cio:1.3.0-rc2 (JVM)

**Describe the bug**

Executing a `HttpStatement` in a `suspend` main function results in a `KotlinNullPointerException` being thrown. 

**To Reproduce**
1. Write the following
```kotlin
suspend fun main()  {
    val client = HttpClient(CIO)
    val statement = client.get<HttpStatement>("https://kotlinlang.org/")
    statement.execute()
}
```
2. Run the program 
3. See error

```
Exception in thread "main" kotlin.KotlinNullPointerException
	at io.ktor.client.engine.HttpClientEngine$DefaultImpls.createCallContext(HttpClientEngine.kt:137)
	at io.ktor.client.engine.HttpClientEngine$DefaultImpls.executeWithinCallContext(HttpClientEngine.kt:73)
	at io.ktor.client.engine.HttpClientEngine$install$1.invokeSuspend(HttpClientEngine.kt:56)
	at io.ktor.client.engine.HttpClientEngine$install$1.invoke(HttpClientEngine.kt)
	at io.ktor.util.pipeline.SuspendFunctionGun.loop(PipelineContext.kt:273)
	at io.ktor.util.pipeline.SuspendFunctionGun.proceed(PipelineContext.kt:141)
	at io.ktor.util.pipeline.SuspendFunctionGun.execute(PipelineContext.kt:161)
	at io.ktor.util.pipeline.Pipeline.execute(Pipeline.kt:27)
	at io.ktor.client.features.HttpSend$DefaultSender.execute(HttpSend.kt:98)
	at io.ktor.client.features.HttpSend$Feature$install$1.invokeSuspend(HttpSend.kt:62)
	at io.ktor.client.features.HttpSend$Feature$install$1.invoke(HttpSend.kt)
	at io.ktor.util.pipeline.SuspendFunctionGun.loop(PipelineContext.kt:273)
	at io.ktor.util.pipeline.SuspendFunctionGun.proceed(PipelineContext.kt:141)
	at io.ktor.util.pipeline.SuspendFunctionGun.proceedWith(PipelineContext.kt:151)
	at io.ktor.client.features.HttpCallValidator$Companion$install$1.invokeSuspend(HttpCallValidator.kt:87)
[...]

PR #1549, opened 2020-01-06
Fix NPE when executing a call within a coroutine with no job

cy6erGn0m:
**Subsystem**
ktor-client-core

**Motivation**
When running a client call within a coroutine with no job (suspend main fun or manually started), a client crashes with NPE. See #1542 

**Solution**
Simply ignore missing `Job` instance when attaching to a parent job


cy6erGn0m:
Merged

PR #1553, opened 2020-01-09
Improve logging, add explicit port in proxy

e5l:


ktor-client/ktor-client-core/posix/src/io/ktor/client/engine/ProxyConfigNative.kt
```diff
@@ -10,7 +10,22 @@ import io.ktor.http.*
  * @param url: proxy url address.
  */
 actual class ProxyConfig(val url: Url) {
-    override fun toString(): String = url.toString()
+    override fun toString(): String = buildString {
+        url.apply {
+            append(protocol.name)
```
cy6erGn0m: Don't we already have it somewhere?

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/ProxyTest.kt
```diff
@@ -25,4 +29,22 @@ class ProxyTest : ClientLoader() {
...
+        config {
+            engine {
+                proxy = ProxyBuilder.http(HTTP_PROXY_SERVER)
+            }
+
+            install(JsonFeature)
+        }
+
+        test { client ->
+            val response = client.get<ProxyResponse>("http://google.com/json")
```
cy6erGn0m: Isn't it better to not specify google but some non-existing domain since we expect it to not requested anyway?

e5l: To get wrong response in case of failure

Issue #1051, opened 2019-04-01
Basic Authentication fails with counter-intuitive exception


suushiemaniac:
### Ktor Version
1.1.3

### Ktor Engine Used(client or server and name)
Apache
client only

### JVM Version, Operating System and Relevant Context
JVM from OpenJDK 1.8.0_191
Ubuntu 18.04 operating system

operating with `io.ktor:ktor-client-core` and `io.ktor:ktor-client-apache` only.
Connecting to an external 3rd party service, so the server is not on ktor and is to be considered a black box.

`io.ktor:ktor-client-auth` plugin for Basic Authentication.

### Feedback
When (accidentally) using the wrong username/password for `BasicAuthProvider` in the `io.ktor:ktor-client-auth` plugin, the client ultimately throws a `SendCountExceedException` because the engine just tries sending again and again until the treshold is reached/surpassed.

I would expect and wish for a more meaningful exception to be thrown when authorization fails. The server does correctly yield response code `401`.

fdc-maku:
I have the same problem. A 3rd party service I was trying to connect to already blocked the user because I accidentally used the wrong password and the ktor client retries to connect to the service with these wrong credentials 20 times.

e5l:
Hi, @suushiemaniac, thanks for the report.
It's a bug and obviously should be fixed.

suushiemaniac:
Is there any progress on this? Just a kind ping, no stress :)

e5l:
Not yet. Btw PR's are always welcome :)

Brunomachadob:
I could try to help with this one, it seems to be a fairly straightforward issue to start with.
I managed to reproduce the problem here.

But just to have a better context, why do we have `ktor-client-auth` and `ktor-client-auth-basic`?

As far as I understood `ktor-client-auth-basic` just intercept the request and append the header, while the other use this complex retry mechanism which I'm still digging into.
Shouldn't we have just one of them with the possibility to retry or not?

Any considerations before I start working on this?

e5l:
Hi @Brunomachadob, awesome to hear :) 
`ktor-client-auth` was JVM-only and `ktor-client-auth-basic` multiplatform, now `ktor-client-auth` is multiplatform and `ktor-client-auth-basic` deprecated.

Probably we can add `simple` mode without retry as we had before. 

BTW. If you have more questions feel free to ask me here or in our slack: `kotlinlang.slack.com`  #ktor channel.

[...]

PR #1555, opened 2020-01-10
Prevent infinite request-response loop for client authentication (#1051)

thumannw:
**Subsystem**
ktor-client:ktor-client-features:ktor-client-auth

**Motivation**
Solves Issue #1051 

**Solution**
The while loop on the returned status code (Unauthorized) was not the only problem for the potential infinite request-response loop: The interceptor is triggering itself infinitely. Solution was to introduce a circuit breaker in the form of an attribute on the HTTP request. My hypothesis here was that these attributes are kind of a meta-data on the client side only and are not sent to the server as a payload. If this is not correct, please let me know.

The first commit enables JVM tests for the unit tests already present (they did not run before). Also, I added a test which provokes the SendCountExceedException as described in the issue. The second commit is the actual fix. The added test and all other tests are successful on my machine with this fix.

### DIFF
--- a/ktor-client/ktor-client-features/ktor-client-auth/common/src/io/ktor/client/features/auth/Auth.kt
+++ b/ktor-client/ktor-client-features/ktor-client-auth/common/src/io/ktor/client/features/auth/Auth.kt
@@ -35,7 +35,7 @@ class Auth(
                 }
             }
 
-            val circuitBreaker = AttributeKey<Int>("auth-request")
+            val circuitBreaker = AttributeKey<Unit>("auth-request")
             scope.feature(HttpSend)!!.intercept { origin, context ->
                 if (origin.response.status != HttpStatusCode.Unauthorized) return@intercept origin
                 if (origin.request.attributes.contains(circuitBreaker)) return@intercept origin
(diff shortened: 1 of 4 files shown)


ktor-client/ktor-client-features/ktor-client-auth/common/src/io/ktor/client/features/auth/Auth.kt
```diff
@@ -35,24 +35,26 @@ class Auth(
...
-                    val provider = feature.providers.find { it.isApplicable(authHeader) } ?: return@intercept call
-
-                    if (provider in usedProviders || provider in feature.alwaysSend) return@intercept call
+                    val provider = candidateProviders.find { it.isApplicable(authHeader) } ?: return@intercept call
+                    candidateProviders.remove(provider)
 
                     val request = HttpRequestBuilder()
                     request.takeFrom(call.request)
                     provider.addRequestHeaders(request)
+                    request.attributes.put(circuitBreaker, 0)
```
e5l: Could you use `Unit` instead of `0` here?

thumannw: Good idea, thanks!

thumannw:
Fixed comment, rebased to current master and fixed conflict.
Thanks for the review.

thumannw:
@e5l Concerning failing TeamCity build: Is this something which has to be looked at, or are these flaky tests?

e5l:
Yep. Thanks for the PR. Merged

Issue #1340, opened 2019-09-13
OkHttp client can't be closed and leaks a lot of threads


kimble:
**Ktor Version and Engine Used (client or server and name)**
Client 1.3.0-beta-1 with okhttp

**Describe the bug**
Closing the client does not terminate threads started by okthttp.

**To Reproduce**

    import io.ktor.client.HttpClient
    import io.ktor.client.engine.okhttp.OkHttp
    import io.ktor.client.request.get
    import io.ktor.client.response.HttpResponse
    import io.ktor.client.response.readBytes
    import kotlinx.coroutines.runBlocking
    import org.junit.jupiter.api.Test
    import java.lang.management.ManagementFactory
    
    class Bug {
    
        @Test
        fun never_closed() {
            var counter = 0
    
            while (!Thread.interrupted() && counter < 20) {
                println("Creating client ${++counter}")
                HttpClient(OkHttp).use { client ->
                    runBlocking {

                        // Remove this request and the threads seems to be removed.
                        client.get<HttpResponse>("https://www.vg.no").use { response ->
                            println("Got " + response.readBytes().size + " bytes")
                        }
                    }
                }
    
                println(" - Number of threads: " + ManagementFactory.getThreadMXBean().threadCount)
                Thread.sleep(1000)
            }
        }
    
    }

**Expected behavior**
I would expect threads to be stopped, but running the test above and you will see the number of threads increase with every iteration. 

I've debugged the code and it seems like the code launched in OkHttpEngine:58 is never executed. 
https://github.com/ktorio/ktor/blob/master/ktor-client/ktor-client-okhttp/jvm/src/io/ktor/client/engine/okhttp/OkHttpEngine.kt#L58

dmitrievanthony:
Hi @kimble, I reproduced the problem on `1.3.0-beta-1` as well as on `master`. 

The problem could also be easily reproduced without `Ktor`. It's enough to create an `OkHttp` client, make a request and close client. Making these actions repeatedly we'll get an increasing number of threads in the system.

[...]

PR #1562, opened 2020-01-15
Fix OkHttp leaking threads.

dmitrievanthony:
**Subsystem**
`HttpClient`, `OkHttp` client infrastructure.

**Motivation**
The problem mentioned in https://github.com/ktorio/ktor/issues/1340.

**Solution**
The problem could be fixed by using a shared `OkHttp` client (aka prototype). Such an approach leads to the inability to clear resources fully closing all internal objects of the prototype. To allow it there is also separated `HttpClient` and `HttpClientEngine` closing.

### DIFF
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/HttpClient.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/HttpClient.kt
@@ -37,8 +37,15 @@ fun <T : HttpClientEngineConfig> HttpClient(
 ): HttpClient {
     val config: HttpClientConfig<T> = HttpClientConfig<T>().apply(block)
     val engine = engineFactory.create(config.engineConfig)
+    val client = HttpClient(engine, config)
 
-    return HttpClient(engine, config)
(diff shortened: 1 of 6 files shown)


ktor-client/ktor-client-core/common/src/io/ktor/client/HttpClient.kt
```diff
@@ -37,8 +37,15 @@ fun <T : HttpClientEngineConfig> HttpClient(
 ): HttpClient {
     val config: HttpClientConfig<T> = HttpClientConfig<T>().apply(block)
     val engine = engineFactory.create(config.engineConfig)
+    val client = HttpClient(engine, config)
 
-    return HttpClient(engine, config)
+    // If the engine was created using factory Ktor is responsible for its lifecycle management. Otherwise user has to
+    // close engine by himself.
+    client.coroutineContext[Job]!!.invokeOnCompletion {
```
e5l: consider adding a general test `client->engine` lifecycle

dmitrievanthony: Added, see `CommonHttpClientTest.kt`.

Issue #1323, opened 2019-09-05
ktor-client-cio missing first 4 characters of headers in Android


czp3009:
**Ktor Version and Engine Used (client or server and name)**
ktor-client-cio 1.2.4

**Describe the bug**
When i use ktor-client-cio in Android as HttpClient engine, i found that first 4 character of headers always missing.

![image](https://user-images.githubusercontent.com/17246126/64315874-de90c200-cfe5-11e9-89b3-a0618c0e23e4.png)

For example, i send request to github, the key of first header is "Date", but cio receive a empty key because 4 character lost. Sometimes "Server" will be the first, and cio receive "er". If first header is "Transfer-Encoding", the incorrect header "sfer-Encoding" cause response body always empty.

![image](https://user-images.githubusercontent.com/17246126/64316012-29123e80-cfe6-11e9-990d-bc9db1bcb034.png)

See the log below

![image](https://user-images.githubusercontent.com/17246126/64316260-e3a24100-cfe6-11e9-89ac-1086c699275a.png)

```
I/i.k.c.HttpClient: 14:05:41.321 [DefaultDispatcher-worker-1] INFO  io.ktor.client.HttpClient - REQUEST: https://api.github.com/orgs/JetBrains/repos
I/i.k.c.HttpClient: 14:05:41.326 [DefaultDispatcher-worker-1] INFO  io.ktor.client.HttpClient - METHOD: HttpMethod(value=GET)
I/i.k.c.HttpClient: 14:05:41.340 [DefaultDispatcher-worker-1] INFO  io.ktor.client.HttpClient - COMMON HEADERS
I/i.k.c.HttpClient: 14:05:41.342 [DefaultDispatcher-worker-1] INFO  io.ktor.client.HttpClient - -> Accept-Charset: UTF-8
I/i.k.c.HttpClient: 14:05:41.344 [DefaultDispatcher-worker-1] INFO  io.ktor.client.HttpClient - -> Accept: */*
I/i.k.c.HttpClient: 14:05:41.348 [DefaultDispatcher-worker-1] INFO  io.ktor.client.HttpClient - CONTENT HEADERS
I/i.k.c.HttpClient: 14:05:41.359 [DefaultDispatcher-worker-1] INFO  io.ktor.client.HttpClient - BODY Content-Type: null
I/i.k.c.HttpClient: 14:05:41.361 [DefaultDispatcher-worker-1] INFO  io.ktor.client.HttpClient - BODY START
[...]

PR #1568, opened 2020-01-16
Fix UTF#decodeUTF8Line() wrong behavior in case ByteBuffer has an offset.

dmitrievanthony:
**Subsystem**
`ktor-io`

**Motivation**
The problem was mentioned in #1323. In case when `ByteBuffer` has an offset (particularly on Android when we call `ByteBuffer.allocateDirect(SIZE)`) function `decodeUTF8Line()` works incorrectly.

**Solution**
Fix position movements in `decodeUTF8Line()` function.

### DIFF
--- a/ktor-io/jvm/src/io/ktor/utils/io/charsets/UTF.kt
+++ b/ktor-io/jvm/src/io/ktor/utils/io/charsets/UTF.kt
@@ -154,7 +154,7 @@ private fun ByteBuffer.decodeUTF8_array(out: CharArray, offset: Int, length: Int
                 // 2 bytes, always valid
 
                 if (srcPos >= srcEnd) {
-                    position(srcPos - 1)
+                    position(srcPos - 1 - arrayOffset())
                     return decodeUtf8Result(outPos - offset, 2)
                 }
 
@@ -164,7 +164,7 @@ private fun ByteBuffer.decodeUTF8_array(out: CharArray, offset: Int, length: Int
             vi and 0xf0 == 0xe0 -> {
                 // 3 bytes
                 if (srcEnd - srcPos < 2) {
-                    position(srcPos - 1)
+                    position(srcPos - 1 - arrayOffset())
                     return decodeUtf8Result(outPos - offset, 3)
                 }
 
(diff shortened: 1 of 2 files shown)


cy6erGn0m:
The explanation is not correct (direct buffer has no array and no offset). However the fix seems to be correct and should be merged in any case.

dmitrievanthony:
> The explanation is not correct (direct buffer has no array and no offset). However the fix seems to be correct and should be merged in any case.

@cy6erGn0m , I'm afraid I don't understand your idea. `DirectByteBuffer` does have an offset (see this [code](https://android.googlesource.com/platform/libcore/+/f33eae7e84eb6d3b0f4e86b59605bb3de73009f3/luni/src/main/java/java/nio/DirectByteBuffer.java#55)), and it means "offset from the base address at which this buffer logically starts". Let me also mention that `hasArray()` method returns `true` on Android, so it emphasizes that buffer does have an array, do you agree?

noboto-niklas:
Sorry for commenting on a closed issue, but after upgrading to ktor 1.3.1 our session storage stopped working. I have suspicions that it is because of this fix. I've written a minimal test which causes the same error as we're seeing.

```
java.lang.IllegalArgumentException: newPosition > limit: (3 > 2)
	at java.base/java.nio.Buffer.createPositionException(Buffer.java:318)
	at java.base/java.nio.Buffer.position(Buffer.java:293)
	at java.base/java.nio.ByteBuffer.position(ByteBuffer.java:1086)
	at java.base/java.nio.ByteBuffer.position(ByteBuffer.java:262)
	at io.ktor.utils.io.charsets.UTFKt.decodeUTF8Line_array(UTF.kt:661)
	at io.ktor.utils.io.charsets.UTFKt.decodeUTF8Line(UTF.kt:36)
	at io.ktor.utils.io.charsets.UTFKt.decodeUTF8Line$default(UTF.kt:34)
	at UTFTest.testDecodeUTF8Line(UTFTest.kt:11)
```

```
@Test
fun testDecodeUTF8Line() {
    val array = "{}".toByteArray()
    val buffer = ByteBuffer.wrap(array)

    buffer.decodeUTF8Line(CharArray(array.size))
}
```

dmitrievanthony:
Hi @noboto-niklas, yes, we faced this issue before and fixed it at https://github.com/ktorio/ktor/pull/1721. I checked your test and it works fine in the current master, so the next release will include the fix.

noboto-niklas:
Hi! I completely missed that issue. Ok, that is great to hear! I look
forward to the next release then. Thanks for fast response!

Den ons 6 maj 2020 09:05Anton Dmitriev <notifications@github.com> skrev:

Issue #1262, opened 2019-08-03
OkHttp Websocket's incoming channel never terminates when server gracefully closes connection


matejdro:
**To Reproduce**
1. Get [example project](https://github.com/ktorio/ktor/files/3463541/sockets.zip)
1. Install web socket library with  `pip install aiohttp==3.5.4`
2. Run `python server.py` from the attached project folder. This starts up basic websocket server that waits for the `close` text message. When it receives, it just closes the socket gracefully and then shuts down itself. 
3. Run the main() function of the attached kotlin program

**Expected behavior**
`incoming` channel on the socket should immediately terminate since the socket is not open anymore and even worse, server is not reachable at all 

**Actual behavior**
WebSocket seems to get stuck. incoming is never finished.

Might be related to #1226

dmitrievanthony:
Hi @matejdro, thanks for reporting. I reproduced this issue and found out that the root cause is the absence of `onClosing()` handler in our web socket listener.

I prepared a fix, it currently on the review (see https://github.com/ktorio/ktor/pull/1571).

dmitrievanthony:
Hi @matejdro, I've just merged the fixes so the current master and next release will contain a fixed version.

PR #1571, opened 2020-01-17
Fix onClosing() web socket handler in OKHttp client (#1262).

dmitrievanthony:
**Subsystem**
OkHttp client.

**Motivation**
The initial problem is mentioned in https://github.com/ktorio/ktor/issues/1262.

**Solution**
Add `onClosing()` handler to `OkHttpWebsocketSession.kt`.

### DIFF
--- a/ktor-client/ktor-client-okhttp/jvm/src/io/ktor/client/engine/okhttp/OkHttpWebsocketSession.kt
+++ b/ktor-client/ktor-client-okhttp/jvm/src/io/ktor/client/engine/okhttp/OkHttpWebsocketSession.kt
@@ -97,7 +97,9 @@ internal class OkHttpWebsocketSession(
         super.onClosing(webSocket, code, reason)
 
         _closeReason.complete(CloseReason(code.toShort(), reason))
-        outgoing.sendBlocking(Frame.Close(CloseReason(code.toShort(), reason)))
+        if (!outgoing.isClosedForSend) {
+            outgoing.sendBlocking(Frame.Close(CloseReason(code.toShort(), reason)))
+        }
         _incoming.close()
     }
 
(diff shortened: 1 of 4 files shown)


ktor-client/ktor-client-okhttp/jvm/src/io/ktor/client/engine/okhttp/OkHttpWebsocketSession.kt
```diff
@@ -93,6 +93,14 @@ internal class OkHttpWebsocketSession(
             "${CloseReason.Codes.byCode(code.toShort())?.toString() ?: code}."))
     }
 
+    override fun onClosing(webSocket: WebSocket, code: Int, reason: String) {
+        super.onClosing(webSocket, code, reason)
+
+        _closeReason.complete(CloseReason(code.toShort(), reason))
+        outgoing.sendBlocking(Frame.Close(CloseReason(code.toShort(), reason)))
```
cy6erGn0m: Sending to a closed channel leads to an exception. Not sure what could happen when throwing it to okhttp

cy6erGn0m:
Looks like the test doesn't pass on CIO :)

dmitrievanthony:
> Looks like the test doesn't pass on CIO :)

Yes, looks like CIO has the same problem. I'm figuring out.

ktor-client/ktor-client-okhttp/jvm/src/io/ktor/client/engine/okhttp/OkHttpWebsocketSession.kt
dmitrievanthony: I added an additional check here, but still not sure about what we shell do in such a case.

PR #1582, opened 2020-01-23
bugfixes for 1.3.1

e5l:
- revert LocalFileContent
- User-Agent fix
- add coverage
- support logging of `Input` request


build.gradle
```diff
@@ -180,8 +180,11 @@ allprojects {
             }
         }
     }
+
```
dmitrievanthony: It doesn't look like a necessary change here.

ktor-client/ktor-client-core/common/src/io/ktor/client/engine/Utils.kt
```diff
@@ -80,3 +81,5 @@ internal suspend inline fun attachToUserJob(callJob: Job) {
         cleanupHandler.dispose()
     }
 }
+
+private fun needUserAgent(): Boolean = !PlatformUtils.IS_BROWSER
```
dmitrievanthony: Does it make sense to make it `inline`?

ktor-utils/common/src/io/ktor/util/ByteChannels.kt
```diff
@@ -42,6 +42,33 @@ fun ByteReadChannel.split(coroutineScope: CoroutineScope): Pair<ByteReadChannel,
...
+ */
+@InternalAPI
+fun ByteReadChannel.copyToBoth(first: ByteWriteChannel, second: ByteWriteChannel) {
+    GlobalScope.launch(Dispatchers.Unconfined) {
+        try {
+            while (!isClosedForRead && (!first.isClosedForWrite || !second.isClosedForWrite)) {
+                readRemaining(CHUNK_BUFFER_SIZE).use {
+                    try {
+                        first.writePacket(it.copy())
+                    } catch (_: Throwable) {
```
dmitrievanthony: Do we intentionally miss all exceptions here?

build.gradle
```diff
@@ -256,3 +259,70 @@ afterEvaluate {
...
+        sourceDirectories.from files(coverageSourceDirs)
+        additionalSourceDirs.from files(coverageSourceDirs)
+        executionData.from files("${buildDir}/jacoco/jvmTest.exec")
+
+        reports {
+            xml.enabled true
+            csv.enabled false
+            html.enabled true
+
+            html.destination file("${buildDir}/jacoco-reports/html")
```
cy6erGn0m: Doesn't Gradle have reports directory for such purposes?

ktor-utils/common/src/io/ktor/util/PlatformUtils.kt
```diff
@@ -0,0 +1,13 @@
+/*
+ * Copyright 2014-2020 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.util
+
+@InternalAPI
+expect object PlatformUtils {
```
cy6erGn0m: This is definitely a thing that everyone would abuse :skull_and_crossbones:

ktor-utils/common/src/io/ktor/util/ByteChannels.kt
e5l: Yep, it's only for logging purposes

ktor-utils/common/src/io/ktor/util/PlatformUtils.kt
e5l: Probably, but we also can't live without it :)

build.gradle
e5l: I faced an issue when Gradle overwrite this directory in parrallel

build.gradle
e5l: tnx

PR #1585, opened 2020-01-24
Cy/server default exceptions

cy6erGn0m:
**Subsystem**
ktor-server-core

**Motivation**
When an http request contains an illegal Content-Type or Accept header value, ktor is producing a `BadContentTypeFormatException` that is reported as 500 and logged as regular exceptions. So log may contain a lot of such errors that is too noisy.

**Solution**
The better behaviour is to produce Bad Request (400) response and log error with debug level. This is much more convenient.

### DIFF
--- a/ktor-server/ktor-server-core/jvm/src/io/ktor/routing/RouteSelector.kt
+++ b/ktor-server/ktor-server-core/jvm/src/io/ktor/routing/RouteSelector.kt
@@ -4,6 +4,7 @@
 
 package io.ktor.routing
 
+import io.ktor.features.*
 import io.ktor.http.*
 import io.ktor.request.*
 import io.ktor.util.*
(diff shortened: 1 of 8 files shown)


ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/DefaultEnginePipeline.kt
```diff
@@ -101,6 +101,9 @@ private fun ApplicationEnvironment.logFailure(call: ApplicationCall, cause: Thro
             is CancellationException -> log.info("$status: $logString, cancelled")
             is ClosedChannelException -> log.info("$status: $logString, channel closed")
             is ChannelIOException -> log.info("$status: $logString, channel failed")
+            is BadRequestException -> log.debug("$status: $logString", cause)
+            is NotFoundException -> log.debug("$status: $logString", cause)
+            is UnsupportedMediaTypeException -> log.debug("$status: $logString, cause)")
```
e5l: Could you check the `"` correctness?

ktor-server/ktor-server-core/jvm/src/io/ktor/routing/RouteSelector.kt
```diff
@@ -380,14 +381,19 @@ data class HttpHeaderRouteSelector(val name: String, val value: String) : RouteS
...
+            val parsedHeaders = parseAndSortContentTypeHeader(acceptHeaderContent)
+
+            if (parsedHeaders.isEmpty())
+                return RouteSelectorEvaluation.Missing
+            val header = parsedHeaders.firstOrNull { contentType.match(it.value) }
+            if (header != null)
+                return RouteSelectorEvaluation(true, header.quality)
+            return RouteSelectorEvaluation.Failed
+        } catch (failedToParse: BadContentTypeFormatException) {
+            throw BadRequestException("Illegal Accept header format: $acceptHeaderContent", failedToParse)
```
e5l: Should we set the cause here?

ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/DefaultEnginePipeline.kt
cy6erGn0m: fixed, thanks

cy6erGn0m:
Merged

Issue #1589, opened 2020-01-27
File descriptor leaks when connection interrupted on routes that provides static content


robober:
**Ktor Version and Engine Used (client or server and name)**
0.9.0-1.3.0, server, ktor-server-jetty

**Describe the bug**
File descriptor leaks when connection interrupted on routes that provides static content

**To Reproduce**
Stable reproduce:
1. Create file in /some/where/filesystem/SOME_FILE
2. Create server (listen port 8080) with static route
`static("/path/files") {
    files("/some/where/filesystem")
}`
3. wget http://localhost:8080/path/files/SOME_FILE
4. Interrupt downloading
5. `sudo lsof -p PID_OF_SERVER | grep SOME_FILE`

**Expected behavior**
The file descriptor must be closed when the download is interrupted.

PR #1590, opened 2020-01-27
Fix server's outgoing content channel cancellation on any failure

cy6erGn0m:
**Subsystem**
ktor-server-host-common

**Motivation**
As described in #1589, when a response transferring is cancelled, we don't cancel content's channel so the file reading coroutine is not cancelled and get stuck. This leads to a file descriptor leakage.

**Solution**
Always cancel channel at the end of a content processing.

**Priority**
This is critical and the fix should be introduced in 1.3.1

### DIFF
--- a/ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/BaseApplicationResponse.kt
+++ b/ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/BaseApplicationResponse.kt
@@ -118,13 +118,17 @@ abstract class BaseApplicationResponse(override val call: ApplicationCall) : App
                 return respondWriteChannelContent(content)
             }
 
-            // Pipe is least efficient
+            // Pipe is the least efficient
             is OutgoingContent.ReadChannelContent -> {
                 // First call user code to acquire read channel, because it could fail
                 val readChannel = content.readFrom()
-                // If channel is fine, commit headers and pipe data
-                commitHeaders(content)
-                return respondFromChannel(readChannel)
+                try {
+                    // If channel is fine, commit headers and pipe data
+                    commitHeaders(content)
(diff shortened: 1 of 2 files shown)


ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/BaseApplicationResponse.kt
```diff
@@ -118,13 +118,17 @@ abstract class BaseApplicationResponse(override val call: ApplicationCall) : App
...
 
-            // Pipe is least efficient
+            // Pipe is the least efficient
             is OutgoingContent.ReadChannelContent -> {
                 // First call user code to acquire read channel, because it could fail
                 val readChannel = content.readFrom()
-                // If channel is fine, commit headers and pipe data
-                commitHeaders(content)
-                return respondFromChannel(readChannel)
+                try {
```
robober: Why **io.ktor.utils.io.ByteReadChannel** does not extends **io.ktor.utils.io.core.Closeable**?
Then we could use try-with-resource statement (**use** for kotlin).

Otherwise cancel() execute close() with kotlinx.coroutines.CancellationException but the reasons for close may be different.

P.S. Its just a question Im interested of patterns used in ktor project.

cy6erGn0m: ReadChannel is similar to ReceiveChannel that have cancel, not close. This doesn't prevent us from having a specialized use function that does cancel. We probably will do in the new kotlinx.io.

robober: Ok. Thanks for reply!)

cy6erGn0m:
Merged

PR #1593, opened 2020-01-29
Introduce remoteAddress support

cy6erGn0m:
**Subsystem**
ktor-server-cio, ktor-http-cio

**Motivation**
Some engines did not support remote address properly. A user opened PR #1531 addressing the issue. However, binary compatibility wasn't met so we were unable to apply it. 

**Solution**
Pick the author's commit and rework a bit CIO part:
- simplify RAW CIO server API so there is no need to write all request handler parameters
- introduce fallback functions so old code will compile (unless `HttpRequestHandler` type alias is used)
- only API changes, no behaviour changes

```kotlin
httpServer(...) { request -> 
  // input, output, remoteAddress, upgraded are available here on the receiver
}
```

instead of 
```kotlin
httpServer(...) { remoteAddress, request, input, output, upgraded ->
}
```

So changing parameters list will be easier in the future.

### DIFF
--- a/ktor-http/ktor-http-cio/jvm/src/io/ktor/http/cio/Pipeline.kt
+++ b/ktor-http/ktor-http-cio/jvm/src/io/ktor/http/cio/Pipeline.kt
@@ -4,16 +4,10 @@
 
 package io.ktor.http.cio
 
-import io.ktor.http.*
 import io.ktor.http.cio.internals.*
-import io.ktor.util.*
-import kotlinx.coroutines.*
-import kotlinx.coroutines.CancellationException
-import kotlinx.coroutines.channels.*
+import io.ktor.server.cio.backend.*
 import io.ktor.utils.io.*
-import java.io.*
-import java.net.*
-import kotlin.coroutines.*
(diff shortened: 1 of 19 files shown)


ktor-http/ktor-http-cio/jvm/src/io/ktor/http/cio/Pipeline.kt
```diff
@@ -22,13 +24,47 @@ fun lastHttpRequest(http11: Boolean, connectionOptions: ConnectionOptions?): Boo
...
+    val output: ByteWriteChannel,
+    val remoteAddress: SocketAddress?,
+    val upgraded: CompletableDeferred<Boolean>?
+) : CoroutineScope {
+    /**
+     * Creates another request scope with same parameters except coroutine context
+     */
+    @KtorExperimentalAPI
+    fun withContext(coroutineContext: CoroutineContext): ServerRequestScope =
+        ServerRequestScope(this.coroutineContext + coroutineContext,
```
e5l: codestyle

cy6erGn0m:
Merged

Issue #1591, opened 2020-01-27
Exception after form and session auth 


Implex1v:
**Ktor Version and Engine Used (client or server and name)**
Tested on 1.3.0 and 1.2.6

**Describe the bug**
Redirect after form and session login does not work. I am getting an exception.
```
17:11:16.611 [nioEventLoopGroup-4-1] ERROR ktor.application - Unhandled: POST - /login
java.util.NoSuchElementException: No session data found for id c1b0142e44d5e67c
	at io.ktor.sessions.DirectoryStorage.read(DirectoryStorage.kt:49)
	at io.ktor.sessions.CacheStorage$referenceCache$1.invokeSuspend(CacheStorage.kt:15)
	at io.ktor.sessions.CacheStorage$referenceCache$1.invoke(CacheStorage.kt)
	at io.ktor.sessions.ReferenceCache$container$1.invokeSuspend(Cache.kt:50)
	at io.ktor.sessions.ReferenceCache$container$1.invoke(Cache.kt)
	at io.ktor.sessions.BaseCache$getOrCompute$2$1.invokeSuspend(Cache.kt:36)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
	at kotlinx.coroutines.DispatchedKt.resumeCancellable(Dispatched.kt:457)
	at kotlinx.coroutines.intrinsics.CancellableKt.startCoroutineCancellable(Cancellable.kt:26)
	at kotlinx.coroutines.CoroutineStart.invoke(CoroutineStart.kt:109)
	at kotlinx.coroutines.AbstractCoroutine.start(AbstractCoroutine.kt:154)
	at kotlinx.coroutines.BuildersKt__Builders_commonKt.async(Builders.common.kt:89)
	at kotlinx.coroutines.BuildersKt.async(Unknown Source)
	at kotlinx.coroutines.BuildersKt__Builders_commonKt.async$default(Builders.common.kt:82)
	at kotlinx.coroutines.BuildersKt.async$default(Unknown Source)
	at io.ktor.sessions.BaseCache$getOrCompute$2.apply(Cache.kt:36)
	at io.ktor.sessions.BaseCache$getOrCompute$2.apply(Cache.kt:31)
	at java.base/java.util.concurrent.ConcurrentHashMap.computeIfAbsent(ConcurrentHashMap.java:1705)
	at io.ktor.sessions.BaseCache.getOrCompute(Cache.kt:36)
[...]

Issue #1586, opened 2020-01-24
Having more than one authentication configurations for a request, only evaluate the first


gdesantos:
**Ktor Version and Engine Used (client or server and name)**
ktor server version 1.3.0

**Describe the bug**
Having a simple application, with 3 JWT authentication providers installed, if i write a request authenthicated with the 3 providers, only the first is evaluated.

**To Reproduce**
Here is a simple full code to reproduce it

```
package com.example

import com.auth0.jwt.JWT
import com.auth0.jwt.algorithms.Algorithm
import io.ktor.application.*
import io.ktor.response.*
import io.ktor.request.*
import io.ktor.routing.*
import io.ktor.http.*
import io.ktor.auth.*
import com.fasterxml.jackson.databind.*
import io.ktor.auth.jwt.jwt
import io.ktor.jackson.*
import io.ktor.features.*

fun main(args: Array<String>): Unit = io.ktor.server.netty.EngineMain.main(args)

@Suppress("unused") // Referenced in application.conf
@kotlin.jvm.JvmOverloads
fun Application.module(testing: Boolean = false) {

    install(ContentNegotiation) {
        jackson {

        }
    }

    install(Authentication) {
        val jwtVerifier = JWT
            .require(Algorithm.HMAC512("secret"))
            .withIssuer("issuer")
            .build()

        jwt("auth1") {
            verifier(jwtVerifier)
            validate { credential ->
                print("Validating 1")
                null
            }
        }

        jwt("auth2") {
            verifier(jwtVerifier)
            validate { credential ->
                print("Validating 2")
                null
            }
        }

        jwt("auth3") {
            verifier(jwtVerifier)
            validate { credential ->
                print("Validating 3")
                null
            }
        }
    }

    routing {
        post("/auth") {
            val token = JWT.create()
                .withIssuer("issuer")
                .sign(Algorithm.HMAC512("secret"))
            call.respond(hashMapOf("token" to token))
        }

        authenticate("auth1", "auth2", "auth3") {
            get("test") {
                call.respond(HttpStatusCode.OK)
            }
        }
    }
}
```

To reproduce the bug, first call the /auth POST to obtain a token, and then call /test GET with that token trying to check the token against the 3 providers.
 
**Expected behavior**
The log should write:
Validating 1
Validating 2
Validating 3

But the log only write:
Validating 1

This means that only the first JWT authentication provider is being evaluated, but not the other two. The first JWT authentication provider is returning null, so ktor engine should try with the next

[...]

PR #1594, opened 2020-01-29
Fix server auth feature and session cache issues

cy6erGn0m:
**Subsystem**
Server ktor-auth, ktor-auth-jwt

**Motivation**
There are the following issues:
- Sometimes an exception from session loader is accidentally re-thrown and cancelled a request handling job in spite of the fact that it is caught (#1591)
- If a session loader crashes, the session cache watchdog may crash (without possible restart).
- When there are multiple auth providers registered, only the first has a chance to handle request while extra providers will be never invoked if the first is failed with invalid credentials (#1586)
- There is minor confusion with auth failure property name

**Solution**
- Fix exceptions handling caused by async sub-coroutine that accidentally killed all parents. Such behaviour is not intended and an exception should be caught and logger in debug level
- Fix cache invalidate functions to not rethrow exceptions. The invalidate function caused watchdog crashes.
- Upgrade JWT/JWKS implementation libraries.
- Simplify server auth procedure and always process all challenges in the end (after all providers actually started).
- Enable binary compatibility validator for two more server modules because 3rd part libraries depend on them: ktor-auth and ktor-websockets


cy6erGn0m:
Merged

PR #1597, opened 2020-01-30
Prepare socket exceptions

e5l:


binary-compatibility-validator/reference-public-api/ktor-client-core.txt
```diff
@@ -1169,3 +1154,21 @@ public final class io/ktor/client/utils/HeadersKt {
 	public static synthetic fun buildHeaders$default (Lkotlin/jvm/functions/Function1;ILjava/lang/Object;)Lio/ktor/http/Headers;
 }
 
+public final class io/ktor/network/sockets/ConnectTimeoutException : java/net/ConnectException {
+	public fun <init> (Lio/ktor/client/request/HttpRequestData;)V
```
dmitrievanthony: If we are going to commonize these exceptions whether it makes sense to pass `HttpRequestData` as a parameter?

ktor-client/ktor-client-core/common/src/io/ktor/network/sockets/TimeoutExceptions.kt
```diff
@@ -0,0 +1,38 @@
...
+ * This exception is thrown in case socket timeout (read or write) exceeded.
+ */
+expect class SocketTimeoutException(request: HttpRequestData) : IOException
+
+/**
+ * Convert long timeout in milliseconds to int value. To do that we need to consider [HttpTimeout.INFINITE_TIMEOUT_MS]
+ * as zero and convert timeout value to [Int].
+ */
+@InternalAPI
+fun convertLongTimeoutToIntWithInfiniteAsZero(timeout: Long): Int = when {
```
dmitrievanthony: Not sure it's the best place for these util functions.

ktor-client/ktor-client-ios/darwin/src/io/ktor/client/engine/ios/IosClientEngine.kt
```diff
@@ -45,7 +46,9 @@ internal class IosClientEngine(override val config: IosClientEngineConfig) : Htt
 
                     if (didCompleteWithError != null) {
                         val mappedException = when (didCompleteWithError.code) {
-                            NSURLErrorTimedOut -> HttpSocketTimeoutException(data)
+                            NSURLErrorTimedOut -> SocketTimeoutException(
```
cy6erGn0m: strange codestyle

ktor-client/ktor-client-okhttp/jvm/src/io/ktor/client/engine/okhttp/OkUtils.kt
```diff
@@ -25,10 +24,10 @@ internal suspend fun OkHttpClient.execute(request: Request, requestData: HttpReq
                 }
 
                 val mappedException = when (cause) {
-                    is SocketTimeoutException -> if (cause.message?.contains("connect") == true) {
-                        HttpConnectTimeoutException(requestData)
+                    is java.net.SocketTimeoutException -> if (cause.message?.contains("connect") == true) {
```
cy6erGn0m: Don't we want to pass the original exception as cause? This could be helpful

ktor-client/ktor-client-ios/darwin/src/io/ktor/client/engine/ios/IosClientEngine.kt
e5l: tnx!

Issue #985, opened 2019-02-27
When returning a String, content negotiation is ignored


mjburghoffer:
### Ktor Version
1.1.3

### Ktor Engine Used(client or server and name)
Server - Jetty

### JVM Version, Operating System and Relevant Context
Azul 10, OSX / Debian Linux

### Feedback
Currently, when you respond with a string, even if you have a header of `Accept=application/json`, due to `fun ApplicationSendPipeline.installDefaultTransformations()`, `TextContent` is always returned.  This means you can't return a bare json string, ie `"foo"`.

The special case of `String` is handled here:

```kotlin
@KtorExperimentalAPI
fun PipelineContext<Any, ApplicationCall>.transformDefaultContent(value: Any): OutgoingContent? = when (value) {
    is OutgoingContent -> value
    is String -> {
        val contentType = call.defaultTextContentType(null)
        TextContent(value, contentType, null)
    }
    is ByteArray -> {
        ByteArrayContent(value)
    }
    is HttpStatusCode -> {
        HttpStatusCodeContent(value)
    }
    is URIFileContent -> {
        if (value.uri.scheme == "file")
            LocalFileContent(File(value.uri))
        else
            null
    }
    else -> null
}
```

oleg-larshin:
Please check the <a href="https://youtrack.jetbrains.com/issue/KTOR-662">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

Stexxe:
You can use explicitly set `Content-Type` header to `application/json` and respond with a JSON:
```kt
val server = embeddedServer(Jetty, port = 7070) {
    routing {
        get("/") {
            call.respondText("\"foo\"", contentType = ContentType.Application.Json)
        }
    }
}
```

PR #1599, opened 2020-02-01
Reduce complexity of Pipeline implementation

thumannw:
**Subsystem**
`ktor-utils`

**Motivation**
In order to develop a clean solution for #985, I wanted to extend the `Pipeline` class with a new feature. However, the complexity of the `Pipeline` class blocked me from doing this.

**Solution**
I propose a re-implementation of the class. Goals of the re-implementation are:
* reduce size
* reduce complexity due to shared variable accesses
* make the code easier to understand and reason about, hence making it better to maintain
* keep the public interface as is
* the new implementation should not imply a major performance regression

**Caveats**
Some tests in `PipelineContractsTest` had to be removed because of the different nature of internal interceptors handling. If these tests were meant as an absolute requirement, the PR has to be rejected. On the other hand, I added further tests to increase code coverage.

**Side note**
The diff for `Pipeline.kt` is probably not so helpful for reviewing, since it is a complete rework of the code. However, note the removal of the `Volatile` annotation. I removed it since it is JVM specific but the class is part of the common source code. Should it be re-added to the `cachedInterceptors` variable?

### DIFF
--- a/ktor-utils/common/src/io/ktor/util/pipeline/Pipeline.kt
+++ b/ktor-utils/common/src/io/ktor/util/pipeline/Pipeline.kt
@@ -5,447 +5,195 @@
 package io.ktor.util.pipeline
 
 import io.ktor.util.*
-import kotlin.jvm.*
 
 /**
  * Represents an execution pipeline for asynchronous extensible computations
  */
-open class Pipeline<TSubject : Any, TContext : Any>(vararg phases: PipelinePhase) {
+open class Pipeline<TS : Any, TC : Any>(vararg phases: PipelinePhase) {
+
     /**
-     * Provides common place to store pipeline attributes
+     * Provides a common place to store pipeline attributes
      */
     val attributes = Attributes()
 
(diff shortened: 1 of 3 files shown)


cy6erGn0m:
Since it is a very core part, so a lot of complexity of its implementation is caused by performance and memory pressure optimizations. So changing the implementation may have a significant impact. So benchmarks should be rerun: complete HTTP server hight load benchmark, client benchmarks, pipelines microbenchmark. Both CPU load and memory pressure shouldn't significantly increase.

thumannw:
Are these benchmarks contained in
`ktor-client:ktor-client-benchmarks`
`ktor-server:ktor-server-benchmarks`?

Are the server benchmarks run like this?
`gradlew :k-s:k-s-b:runBenchmark -PbenchmarkName=<name>`
If yes, I get a lot of Gradle errors of type `Could not resolve project ...`. Do you get the same errors?

thumannw:
@cy6erGn0m Is there any interest in this improvement? If not, we can just close this PR.

PR #1602, opened 2020-02-03
Remove org.eclipse.jetty.alpn:alpn-api from ktor-server-netty

alexburlton-sonocent:
**Subsystem**
ktor-server-netty

**Motivation**
See https://github.com/ktorio/ktor/issues/1307 - this dependency appears to be unused, and is creating noise for us when running the gradle dependency check (flags up https://nvd.nist.gov/vuln/detail/CVE-2007-5615).

**Solution**
I've removed the dependency and run gradle build - all seems fine. I have 56 tests passing locally with no failures. 

Let me know if there's anything else I should do to verify that I haven't broken anything :)

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -27,7 +27,6 @@ atomicfu_version=0.14.1
 netty_version=4.1.44.Final
 netty_tcnative_version=2.0.27.Final
 jetty_version=9.4.24.v20191120
-jetty_alpn_api_version=1.1.3.v20160715
 
 # utility
 logback_version=1.2.3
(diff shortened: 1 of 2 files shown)


cy6erGn0m:
Well, CVE flag is false, isn't it? The versions mentioned in the CVE are quite old and not referenced from the dependency?

cy6erGn0m:
Looks like we need it for proper HTTP/2 support. It's a shame that we don't support running tests for HTTP/2 with Netty (we are planning to fix it soon).

alexburlton-sonocent:
You're right that the CVE is old, but the dependency is even older (2016) and hasn't been maintained since. I think it *may* be a false positive, but not sure how to prove that one way or the other.

As someone else had already raised an issue saying this was unused, though, I would've thought that from a code hygiene perspective it'd still be a good course of action to get rid of it?

PR #1607, opened 2020-02-04
Add a list of exempted path prefixes for https redirection.

clydebarrow:
**Subsystem**
Server, HttpsRedirect feature.

**Motivation**
When using  the `HttpsRedirect` feature in some contexts it is necessary to exempt certain path prefixes from redirection. A specific example is Google App Engine cron jobs which use an http:// url to trigger scheduled tasks. Currently these fail with a 301 status if `HttpsRedirect` is used.

**Solution**
Add a configuration option to `HttpsRedirect` that specifies an list of uri prefixes to be exempted from redirection.


cy6erGn0m:
I would prefer a lambda predicates instead of exact prefixes.

Something like this:
```kotlin
install(HttpsRedirect) {
    exclude { call ->
        call.request.parameters["x"] == "1" || call.request.path.startsWith("/some/path")
    }
}
```

clydebarrow:
That would certainly be more flexible, but seems more complex than required for the use cases I've identified, and since I would have to spend time working out how to implement it, I'll defer that to someone with more Kotlin knowledge.

cy6erGn0m:
Merged.

PR #1610, opened 2020-02-04
Fix netty transports native dependencies

daniilguit:
**Subsystem**
Server

**Motivation**
Currently netty transports native dependencies are used without platform classifiers thus lacking actual compiled JNI libraries so epoll/kqueue are not really loaded on linux/osx

**Solution**
Added platform classifier to netty transports native dependencies

### DIFF
--- a/ktor-server/ktor-server-netty/build.gradle
+++ b/ktor-server/ktor-server-netty/build.gradle
@@ -7,8 +7,8 @@ kotlin.sourceSets {
         api group: 'io.netty', name: 'netty-codec-http2', version: netty_version
         api group: 'org.eclipse.jetty.alpn', name: 'alpn-api', version: jetty_alpn_api_version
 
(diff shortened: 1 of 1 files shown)


cy6erGn0m:
Well, this is the point: one need to opt-in by adding a proper dependency to get it working. Otherwise, a safe JVM implementation is applied by default. 

The other solution would be to provide a configuration parameter instead. Not sure if it is safe to enable native hacks by default.

daniilguit:
@cy6erGn0m : valid point. My issue was that I saw -epoll library in dependencies and did not notice that it does not have JNIs and was wondering for some time why epoll did not work. After  that I had to exclude from dependencies JNI-less jars to avoid class files duplication. Not sure what is the best solution to avoid both of this issues.

e5l:
I think we should keep them as it is and clarify this point in docs: @andreyaksenov

PR #1611, opened 2020-02-04
Fix for stress tests

cy6erGn0m:
**Subsystem**
ktor-server-cio, ktor-server-jetty, ktor-server-netty

**Motivation**
Currently, stress tests for upgrade requests are failing on both Linux and Windows. 

On Netty we were logging exceptions so that caused test failures. 
On Jetty, coroutines were hanging forever without proper cancellation. 
Also, test filtering in stress tests was not working properly.

**Solution**
Netty was broken at the point when the closed send exception stopped being cancellation. So now it is caught and wrapped to CancellationException.

On Jetty, there were no proper cancellation on upgrade job completion. So the reading job was hanging if the channel is full and nobody read from it anymore. 

Now we assume that if an upgrade job is completed, then nobody will use channels anymore so we close output and cancel input.


cy6erGn0m:
Merged

PR #1616, opened 2020-02-05
Fix IOOBE in coroutine debug probes

cy6erGn0m:
**Subsystem**
ktor-utils (Pipelines)

**Motivation**
Sometimes, debug probes do walk through stack frames when the analyzed coroutine is running concurrently. This causes IOOBE exceptions in tests from time to time and makes tests flaky.

**Solution**
Catch all crashes and fallback to `null` on any accident.

### DIFF
--- a/ktor-utils/common/src/io/ktor/util/pipeline/PipelineContext.kt
+++ b/ktor-utils/common/src/io/ktor/util/pipeline/PipelineContext.kt
@@ -92,7 +92,7 @@ private class SuspendFunctionGun<TSubject : Any, TContext : Any>(
                     return rootContinuation
                 }
                 is ArrayList<*> -> {
-                    if (rootContinuation.isEmpty()) return null
+                    if (rootContinuation.isEmpty()) return StackWalkingFailedFrame
                     @Suppress("UNCHECKED_CAST")
                     return peekContinuationFromList(rootContinuation as List<Continuation<*>>)
                 }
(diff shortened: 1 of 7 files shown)


ktor-utils/common/src/io/ktor/util/pipeline/PipelineContext.kt
```diff
@@ -84,28 +84,44 @@ private class SuspendFunctionGun<TSubject : Any, TContext : Any>(
...
+            when (val rootContinuation = rootContinuation) {
                 null -> return null
                 is Continuation<*> -> {
                     --lastPeekedIndex
                     return rootContinuation
                 }
                 is ArrayList<*> -> {
-                    if (rootContinuation.isEmpty()) return null
-                    return rootContinuation[lastPeekedIndex--] as Continuation<*>
+                    if (rootContinuation.isEmpty()) return StackWalkingFailedFrame
```
qwwdfsad: Isn't empty list a valid situation?

cy6erGn0m: Right, it is.

cy6erGn0m: Well, the answer is both yes and no. If the list is empty, then `lastPeekedIndex` should be `-1` and `null` should be already returned at the beginning of the function. So if we are here and see that the list is empty but lastPeekedIndex wasn't `-1`, then it means that element(s) has disappeared concurrently. 

I think we have two options here: we may return `null` and say that it's ok in spite of the accident, or we may report it as a problem as it is done in this PR.

cy6erGn0m:
Merged

Issue #1189, opened 2019-06-14
httpclient gzip decoding failed


rockyou:
### Ktor Version
1.2.1
### Ktor Engine Used(client or server and name)
Http Client with Apache engine
### JVM Version, Operating System and Relevant Context
JDK 11,MAC OS
### Feedback
If ContentEncoding feature turned on 
```
    install(ContentEncoding) {
                gzip()
                identity()
            }
```
and get content from any website which supports compression

An exception thrown:
```
java.lang.IllegalStateException: Gzip padding invalid.
```

e5l:
Hi @rockyou, can you provide the request example? The request log also would be nice.

rockyou:
Hi! Thanks so much for quick response.

Test code:
    runBlocking {
        val client = HttpClient(Apache) {

            install(ContentEncoding) {
//                gzip()
                identity()
            }
        }
        client.call("http://mockbin.com/request") {
            method = HttpMethod.Post
        }.response.use { println(it.readText()) }

    }

when gzip-encoding is disabled,the request result is correct.
when gzip-encoding is enabled,exception is thrown:
Exception in thread "pool-1-thread-3" java.lang.IllegalStateException: Gzip padding invalid.
	at io.ktor.util.EncodersJvmKt$inflate$1.invokeSuspend(EncodersJvm.kt:55)
	at io.ktor.util.EncodersJvmKt$inflate$1.invoke(EncodersJvm.kt)
	at kotlinx.coroutines.io.CoroutinesKt$launchChannel$job$1.invokeSuspend(Coroutines.kt:123)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
	at kotlinx.coroutines.DispatchedTask.run(Dispatched.kt:238)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)

luishiga:
In ktor/ktor-utils/jvm/src/io/ktor/utilEncoders.jvm the check padding should be only size?
line 55:
`check(padding.contentEquals(GZIP_HEADER_PADDING)) { "Gzip padding invalid." }
`

change the condition to something like
`check(padding.size == GZIP_HEADER_PADDING.size) { "Gzip padding invalid." }`

[...]

PR #1621, opened 2020-02-06
Fix httpclient gzip decoding failed #1189 - gzip header check fix

pecanw:
**Subsystem**
ktor-utils

**Motivation**
Fix httpclient gzip decoding failed #1189

**Solution**
Changing the gzip header check to conform [specification](http://www.zlib.org/rfc-gzip.html)

### DIFF
--- a/ktor-utils/jvm/src/io/ktor/util/EncodersJvm.kt
+++ b/ktor-utils/jvm/src/io/ktor/util/EncodersJvm.kt
@@ -48,11 +48,32 @@ private fun CoroutineScope.inflate(
         val header = source.readPacket(GZIP_HEADER_SIZE)
         val magic = header.readShortLittleEndian()
         val format = header.readByte()
-        val padding = header.readBytes()
+        val flags = header.readByte().toInt()
+        // val time = header.readInt()
+        // val extraFlags = header.readByte()
+        // val osType = header.readByte()
+
+        /* flags bits:
+            bit 0   FTEXT
+            bit 1   FHCRC
+            bit 2   FEXTRA
+            bit 3   FNAME
+            bit 4   FCOMMENT
+            bit 5   reserved
+            bit 6   reserved
+            bit 7   reserved
+        */
+        if (flags and 0b0000_0100 != 0) { //FLG.FEXTRA
+            val extraLen = source.readShort().toInt()
+            val extra = source.readPacket(extraLen)
+        }
 
         check(magic == GZIP_MAGIC) { "GZIP magic invalid: $magic" }
         check(format.toInt() == Deflater.DEFLATED) { "Deflater method unsupported: $format." }
-        check(padding.contentEquals(GZIP_HEADER_PADDING)) { "Gzip padding invalid." }
+        //FLG.FNAME
+        check(flags and 0b0000_1000 == 0) { "Gzip file name not suported" }
+        //FLG.FCOMMENT
+        check(flags and 0b0001_0000 == 0) { "Gzip file comment not suported" }
     }
 
     try {
--- a/ktor-utils/jvm/src/io/ktor/util/EncodersJvm.kt
+++ b/ktor-utils/jvm/src/io/ktor/util/EncodersJvm.kt
@@ -14,7 +14,7 @@ import java.util.zip.*
 private const val GZIP_HEADER_SIZE: Int = 10
 
 // GZIP header flags bits
-private object FLG {
+private object GzipHeaderFlags {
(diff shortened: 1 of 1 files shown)


ktor-utils/jvm/src/io/ktor/util/EncodersJvm.kt
```diff
@@ -48,11 +48,32 @@ private fun CoroutineScope.inflate(
         val header = source.readPacket(GZIP_HEADER_SIZE)
         val magic = header.readShortLittleEndian()
         val format = header.readByte()
-        val padding = header.readBytes()
+        val flags = header.readByte().toInt()
+        // val time = header.readInt()
+        // val extraFlags = header.readByte()
+        // val osType = header.readByte()
```
Nohus: Why are the commented out variables necessary? May be worth either deleting them or using them.

ktor-utils/jvm/src/io/ktor/util/EncodersJvm.kt
```diff
@@ -48,11 +48,32 @@ private fun CoroutineScope.inflate(
...
+            bit 2   FEXTRA
+            bit 3   FNAME
+            bit 4   FCOMMENT
+            bit 5   reserved
+            bit 6   reserved
+            bit 7   reserved
+        */
+        if (flags and 0b0000_0100 != 0) { //FLG.FEXTRA
+            val extraLen = source.readShort().toInt()
+            val extra = source.readPacket(extraLen)
```
Nohus: I suspect this is here only to skip past the extra packet, a comment would help make this visible.

ktor-utils/jvm/src/io/ktor/util/EncodersJvm.kt
pecanw: I added an extra comment, but I would keep them for reference or future usage.

ktor-utils/jvm/src/io/ktor/util/EncodersJvm.kt
pecanw: I did some refactoring to make the code even more readable and added a comment as proposed.

ktor-utils/jvm/src/io/ktor/util/EncodersJvm.kt
```diff
@@ -13,6 +13,17 @@ import java.util.zip.*
 
 private const val GZIP_HEADER_SIZE: Int = 10
 
+// GZIP header flags bits
+private object FLG {
```
cy6erGn0m: GzipHeaderFlags ?

ktor-utils/jvm/src/io/ktor/util/EncodersJvm.kt
```diff
@@ -48,11 +59,31 @@ private fun CoroutineScope.inflate(
         val header = source.readPacket(GZIP_HEADER_SIZE)
         val magic = header.readShortLittleEndian()
         val format = header.readByte()
-        val padding = header.readBytes()
+        val flags = header.readByte().toInt()
```
cy6erGn0m: Memory leak here: header should be consumed entirely or discarded

ktor-utils/jvm/src/io/ktor/util/EncodersJvm.kt
pecanw: The `FLG` name was taken from the RFC spec, but I agree that `GzipHeaderFlags` is better. Renamed in commit e209b0b90101794b3505731012c5c0cb1d5d26b1.

ktor-utils/jvm/src/io/ktor/util/EncodersJvm.kt
pecanw: Fixed in 6bdf82850036b0c532307429bc460c988263dddd.

pecanw:
> LGTM, but we should fix memory leak(by consuming) all header bytes.

All issues from comments have been fixed, so I believe you can merge the code. @cy6erGn0m & @Nohus - can you please close your comments?

pecanw:
However some unit test (not related to the changed code) have failed with a timeout. I haven't figured out how to re-run the build. Can someone check it please?

pecanw:
When are you going to release a new version (1.3.2?) including this fix?

e5l:
Yep, but no dates for now. Probably on the next week.

pecanw:
> Yep, but no dates for now. Probably on the next week.

Any news? Can you release a new version please?

PR #1629, opened 2020-02-09
Move the DefaultByteBufferPool to ktor-network module

MariusVolkhart:
The `DefaultByteBufferPool` lived in the _ktor-http-cio_ module, but was used by the _ktor-network-tls_ module. This meant the the _ktor-network-tls_ module had both the _ktor-http-cio_ and _ktor-http_ modules as dependencies. This is undesirable as the HTTP code is not required for TLS or plain-socket connections, and it increases the size of applications, as well as adding unnecessary items to the classpath.

The _ktor-network_ already had a file, `Pools.kt`, for declaring an `ObjectPool` for UDP datagrams, so the general purpose pool from the _ktor-http-cio_ module has been moved to that file. A further benefit is that there is now only 1 implementation of `DirectByteBufferPool` in the ktor project.

After moving the `ObjectPool` to the _ktor-network_ module, the _ktor-network_ module no longer uses anything from _ktor-http-cio_ and the dependency was replaced with the _ktor-utils_ module.


e5l:
One little thing remaining: could you update binary compatibility according to the build log?

MariusVolkhart:
@e5l I ran the binary compatibility check with `-Poverwrite.output=true` to update the text files. I hope this is correct, since the API changes is marked with `@InternalAPI`

PR #1631, opened 2020-02-10
Fix maxFrameSize for fragmented packets

e5l:


ktor-http/ktor-http-cio/jvm/src/io/ktor/http/cio/websocket/DefaultWebSocketSessionImpl.kt
```diff
@@ -195,6 +202,13 @@ class DefaultWebSocketSessionImpl(
         }
     }
 
+    private suspend fun checkMaxFrameSize(size: Int) {
+        if (size > maxFrameSize) {
+            close(CloseReason(CloseReason.Codes.TOO_BIG, "Frame is too big: $size. Max size is $maxFrameSize"))
```
cy6erGn0m: Is it safe? Doest the following exception cancel this close frame? Hooe it doesn't

Issue #1614, opened 2020-02-05
URLParser doesn't handle all scheme/protocols correctly


vdesmet93:
**Ktor Version and Engine Used (client or server and name)**
Tested with Ktor Client Common 1.3.70-eap-42, but bug is present in current stable version as well.

**Describe the bug**
The URLParser (used by the URLBuilder) does not support schemes with numbers, plus
   ("+"), period ("."), or hyphen ("-") characters. 
This is not according to the RFC 3986 specification: https://tools.ietf.org/html/rfc3986#page-17

The `findScheme` function only handles schems with letters and when no scheme can be parsed it will default to `http` as scheme and `localhost` as host.

**To Reproduce**
Execute
```
val urlBuilder = URLBuilder(urlString = "test-scheme123://ktor.io/index.html")
println("Test URL: ${urlBuilder.buildString()}")
```
This will print
> Test URL: http://localhost/test-scheme123://ktor.io/index.html

**Expected behavior**
The URLParser should correctly handle these types of URL schemes, and in this case it should print the initial URL that was passed to the URLBuilder.

dmitrievanthony:
Hi @vdesmet93,

Thanks for reporting. I fixed this bug in #1635, so the fix will be included in the next release `1.3.2`.

PR #1635, opened 2020-02-11
Fix URL scheme parser (#1614).

dmitrievanthony:
**Subsystem**
Module `ktor-http`. 

**Motivation**
The bug described in #1614. URL scheme parser does not handle URLs that are correct from the RFC perspective.

### DIFF
--- a/ktor-http/common/src/io/ktor/http/URLParser.kt
+++ b/ktor-http/common/src/io/ktor/http/URLParser.kt
@@ -6,7 +6,6 @@ package io.ktor.http
 
 import io.ktor.util.*
 
-
 /**
  * Take url parts from [urlString]
  * throws [URLParserException]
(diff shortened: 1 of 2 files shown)


ktor-http/common/src/io/ktor/http/AsciiCharSet.kt
```diff
@@ -0,0 +1,85 @@
...
+    /**
+     * Adds all given [chars] to the set.
+     * Throws [IllegalArgumentException] is some characters are not ASCII characters.
+     */
+    fun add(chars: Iterable<Char>) {
+        for (char in chars) {
+            val index = char.toInt()
+
+            // Only ASCII characters are handled.
+            if (index !in 0..0x7F) {
```
e5l: `index < 0`

ktor-http/common/src/io/ktor/http/URLParser.kt
```diff
@@ -132,15 +131,53 @@ private fun URLBuilder.fillHost(urlString: String, startIndex: Int, endIndex: In
...
+    if (firstChar !in 'a'..'z' && firstChar !in 'A'..'Z') {
+        incorrectSchemePosition = current
+    }
+
     while (current < endIndex) {
         val char = urlString[current]
-        if (char == ':') return current
-        if (!char.isLetter()) return -1
+
+        // Character ':' means the end of the scheme and at thi point the length of the scheme should be returned or the
```
e5l: typo

dmitrievanthony: Fixed.

Issue #1618, opened 2020-02-06
Ktor-client-logging : Add an option to filter logged calls


eMZet-zz:
**Subsystem**
Kotlin multiplatform library: ktor-client-logging:1.3.0

**Is your feature request related to a problem? Please describe.**
Problem: not able to filter calls to be logged. All calls are being logged by default.

**Describe the solution you'd like**
Add a filter option in Logging configuration block (this solution is available on Android in CallLogging feature, but it is absent in kotlin multiplatform)
https://ktor.io/servers/features/call-logging.html

```
install(Logging) {
    filter { call -> call.request.path().startsWith("/section1") }
}
```

**Motivation to include to ktor**
Logging all queries is helpful in debug mode. However, for a production version in which, for example, logs are saved to a file, saving only failed requests would be sufficient. It would be convenient to use the function that is already in the library.

oleg-larshin:
Please check the <a href="https://youtrack.jetbrains.com/issue/KTOR-481">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

AndresAusecha:
Hey I would like to try with this one

mervyn-mccreight:
Is this really still open?
Looking at `ktor-client/ktor-client-plugins/ktor-client-logging/common/src/io/ktor/client/plugins/logging/Logging.kt` it looks like this is there already:

```kotlin
    /**
     * [Logging] plugin configuration
     */
    @KtorDsl
    public class Config {
        /**
         * filters
         */
        internal var filters = mutableListOf<(HttpRequestBuilder) -> Boolean>()

        /**
         * [Logger] instance to use
         */
        public var logger: Logger = Logger.DEFAULT

        /**
         * log [LogLevel]
         */
        public var level: LogLevel = LogLevel.HEADERS

        /**
         * Log messages for calls matching a [predicate].
         */
        public fun filter(predicate: (HttpRequestBuilder) -> Boolean) {
            filters.add(predicate)
        }
    }
```

Stexxe:
Fixed by https://github.com/ktorio/ktor/commit/da7cb98d829596fe66342e47e5a5f270e63c261c

PR #1638, opened 2020-02-11
Add an option to filter logged calls for ktor-client-logging.

dragneelfps:
**Subsystem**
Kotlin multiplatform library: ktor-client-logging:1.3.0

**Motivation**
Problem: not able to filter calls to be logged. All calls are being logged by default.
This is inspired by issue https://github.com/ktorio/ktor/issues/1618.

**Solution**
Added filters of type `(HttpRequestBuilder) -> Boolean` to `Logging` Config.

### DIFF
--- a/binary-compatibility-validator/reference-public-api/ktor-client-logging.txt
+++ b/binary-compatibility-validator/reference-public-api/ktor-client-logging.txt
@@ -32,8 +32,11 @@ public final class io/ktor/client/features/logging/LoggerKt {
 public final class io/ktor/client/features/logging/Logging {
 	public static final field Companion Lio/ktor/client/features/logging/Logging$Companion;
 	public fun <init> (Lio/ktor/client/features/logging/Logger;Lio/ktor/client/features/logging/LogLevel;)V
+	public fun <init> (Lio/ktor/client/features/logging/Logger;Lio/ktor/client/features/logging/LogLevel;Ljava/util/List;)V
+	public final fun getFilters ()Ljava/util/List;
 	public final fun getLevel ()Lio/ktor/client/features/logging/LogLevel;
 	public final fun getLogger ()Lio/ktor/client/features/logging/Logger;
+	public final fun setFilters (Ljava/util/List;)V
 	public final fun setLevel (Lio/ktor/client/features/logging/LogLevel;)V
 }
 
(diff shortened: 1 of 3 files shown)


cy6erGn0m:
Unfortunately, the primary constructor is public so adding constructor parameter breaks binary and source compatibility. A backward compatibility constructor should be added without the new parameter. After that, please run 

```
./gradlew clean binary-compatibility-validator:jvmTest -Poverwrite.output=true
```

Then review signature changes and commit the changed file.

dragneelfps:
@cy6erGn0m I have added the ctor and done the stuff you asked for.

Webunity:
@dragneelfps i don't see any unit-tests which are created. Can you add some?

dragneelfps:
Will do.

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

dragneelfps:
I have added test for request log filtering. 

One question comes to my mind:
With this PR we can have multiple filters. So, should we log requests for requests that pass ALL filters or pass ANY filter or support both.
My solution will be to use `&&` for ALL use-cases and separate `filter {}` for ANY use case.

dragneelfps:
Any update when will this be merged?

e5l:
LGTM. I can't merge it with GitHub right now. Are you ok if I manually cherry-pick your commit?

dragneelfps:
@e5l sure. Go ahead.

e5l:
Merged manually. Thanks for the PR

eMZet-zz:
The filtering solution for common client is different than for jvm one:
`var filters: List<(HttpRequestBuilder) -> Boolean> //common`
`val filters: List<(ApplicationCall) -> Boolean>  //jvm` 

therefore there is no way to access call's response and as a result log only unsuccessful calls and responses.
Would it be possible to change the signature so that we have more flexibility while logging?

dragneelfps:
@eMZet-zz Ill look into it.

dragneelfps:
@eMZet-zz I have created a draft PR https://github.com/ktorio/ktor/pull/1728. Can you take a look and see if the proposed changes are fine and how can they be improved.

geoffreywiseman:
This isn't released yet, correct? I'd be happy to use it, but doesn't look like it's in 1.3.1.

Issue #1472, opened 2019-12-07
The content type application/json+hal causes a NoTransformationFoundException


rekire:
**Ktor client 1.2.6 and with OkHttp Engine and GsonSerializer**
Used dependencies:
```
compile "io.ktor:ktor-client-okhttp:1.2.6"
compile "io.ktor:ktor-client-json:1.2.6"
compile "io.ktor:ktor-client-gson:1.2.6"
```

**Describe the bug**
The content type application/json+hal causes a NoTransformationFoundException

**To Reproduce**
Create a REST request which returns an data class when the response has the content type `application/json+hal`.  
More details and a bit too [complex example](https://stackoverflow.com/q/59221697/995926) at Stack Overflow

**Expected behavior**
I would expect that the ktor behaves the same with the content type `application/json` and just returns my data class.

**Screenshots**
```
Exception in thread "main" io.ktor.client.call.NoTransformationFoundException: No transformation found: class kotlinx.coroutines.io.ByteBufferChannel -> class my.package.name.Skills
        at io.ktor.client.call.HttpClientCall.receive(HttpClientCall.kt:88)
        at my.package.name.AmazonApi$fetchSkills$1.invokeSuspend(AmazonApi.kt:200)
        at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
        at io.ktor.util.pipeline.SuspendFunctionGun.resumeRootWith(PipelineContext.kt:215)
        at io.ktor.util.pipeline.SuspendFunctionGun.loop(PipelineContext.kt:172)
        at io.ktor.util.pipeline.SuspendFunctionGun.access$loop(PipelineContext.kt:67)
        at io.ktor.util.pipeline.SuspendFunctionGun$continuation$1.resumeWith(PipelineContext.kt:122)
        at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:46)
        at io.ktor.util.pipeline.SuspendFunctionGun.resumeRootWith(PipelineContext.kt:215)
        at io.ktor.util.pipeline.SuspendFunctionGun.loop(PipelineContext.kt:172)
        at io.ktor.util.pipeline.SuspendFunctionGun.access$loop(PipelineContext.kt:67)
        at io.ktor.util.pipeline.SuspendFunctionGun$continuation$1.resumeWith(PipelineContext.kt:122)
[...]

PR #1640, opened 2020-02-12
Improve JsonFeature configuration DSL (#1472).

dmitrievanthony:
**Subsystem**
Ktor Client, `JsonFeature`.

**Motivation**
Lack of a good DSL described in #1472.

### DIFF
--- a/ktor-client/ktor-client-features/ktor-client-json/common/src/io/ktor/client/features/json/JsonFeature.kt
+++ b/ktor-client/ktor-client-features/ktor-client-json/common/src/io/ktor/client/features/json/JsonFeature.kt
@@ -60,16 +60,32 @@ class JsonFeature internal constructor(
         var serializer: JsonSerializer? = null
 
         /**
+         * Backing field with mutable list of content types that are handled by this feature.
+         */
+        private val _acceptContentTypes: MutableList<ContentType> = mutableListOf(ContentType.Application.Json)
+
+        /**
          * List of content types that are handled by this feature.
          * It also affects `Accept` request header value.
          * Please note that wildcard content types are supported but no quality specification provided.
          */
         @KtorExperimentalAPI
-        var acceptContentTypes: List<ContentType> = listOf(ContentType.Application.Json)
-            set(newList) {
-                require(newList.isNotEmpty()) { "At least one content type should be provided to acceptContentTypes" }
-                field = newList
+        var acceptContentTypes: List<ContentType>
+            set(value) {
+                require(value.isNotEmpty()) { "At least one content type should be provided to acceptContentTypes" }
+                _acceptContentTypes.clear()
+                _acceptContentTypes.addAll(value)
             }
+            get() = _acceptContentTypes
+
+        /**
+         * Adds accepted content types. Be aware that [ContentType.Application.Json] accepted by default is removed from
+         * the list if you use this function to provide accepted content types.
+         * It also affects `Accept` request header value.
+         */
+        fun accept(vararg contentTypes: ContentType) {
+            _acceptContentTypes += contentTypes
+        }
     }
 
     /**
--- a/ktor-http/common/src/io/ktor/http/ContentTypes.kt
+++ b/ktor-http/common/src/io/ktor/http/ContentTypes.kt
@@ -121,6 +121,7 @@ class ContentType private constructor(val contentType: String, val contentSubtyp
         val Any = ContentType("application", "*")
         val Atom = ContentType("application", "atom+xml")
         val Json = ContentType("application", "json")
+        val HalJson = ContentType("application", "hal+json")
         val JavaScript = ContentType("application", "javascript")
         val OctetStream = ContentType("application", "octet-stream")
         val FontWoff = ContentType("application", "font-woff")
--- a/binary-compatibility-validator/reference-public-api/ktor-client-json.txt
+++ b/binary-compatibility-validator/reference-public-api/ktor-client-json.txt
@@ -11,6 +11,7 @@ public final class io/ktor/client/features/json/JsonFeature {
 
 public final class io/ktor/client/features/json/JsonFeature$Config {
 	public fun <init> ()V
+	public final fun accept ([Lio/ktor/http/ContentType;)V
 	public final fun getAcceptContentTypes ()Ljava/util/List;
 	public final fun getSerializer ()Lio/ktor/client/features/json/JsonSerializer;
 	public final fun setAcceptContentTypes (Ljava/util/List;)V
--- a/binary-compatibility-validator/reference-public-api/ktor-http.txt
+++ b/binary-compatibility-validator/reference-public-api/ktor-http.txt
@@ -149,6 +149,7 @@ public final class io/ktor/http/ContentType$Application {
 	public final fun getFontWoff ()Lio/ktor/http/ContentType;
 	public final fun getFormUrlEncoded ()Lio/ktor/http/ContentType;
 	public final fun getGZip ()Lio/ktor/http/ContentType;
+	public final fun getHalJson ()Lio/ktor/http/ContentType;
 	public final fun getJavaScript ()Lio/ktor/http/ContentType;
 	public final fun getJson ()Lio/ktor/http/ContentType;
 	public final fun getOctetStream ()Lio/ktor/http/ContentType;
--- a/ktor-client/ktor-client-features/ktor-client-json/common/test/JsonFeatureTest.kt
+++ b/ktor-client/ktor-client-features/ktor-client-json/common/test/JsonFeatureTest.kt
@@ -1,0 +1,39 @@
+/*
+ * Copyright 2014-2020 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.client.features.json
+
+import io.ktor.http.*
+import kotlin.test.*
+
+class JsonFeatureTest {
+    @Test
+    fun testDefaultContentTypes() {
+        val config = JsonFeature.Config()
+
+        assertEquals(1, config.acceptContentTypes.size)
+        assertTrue { config.acceptContentTypes.contains(ContentType.Application.Json) }
+    }
+
+    @Test
+    fun testAcceptCall() {
+        val config = JsonFeature.Config()
+        config.accept(ContentType.Application.Xml)
+
+        assertEquals(2, config.acceptContentTypes.size)
+        assertTrue { config.acceptContentTypes.contains(ContentType.Application.Json) }
+        assertTrue { config.acceptContentTypes.contains(ContentType.Application.Xml) }
+    }
+
+    @Test
+    fun testContentTypesListAssignment() {
+        val config = JsonFeature.Config()
+        config.acceptContentTypes = listOf(ContentType.Application.Pdf, ContentType.Application.Xml)
+
+        assertEquals(2, config.acceptContentTypes.size)
+        assertFalse { config.acceptContentTypes.contains(ContentType.Application.Json) }
+        assertTrue { config.acceptContentTypes.contains(ContentType.Application.Xml) }
+        assertTrue { config.acceptContentTypes.contains(ContentType.Application.Pdf) }
+    }
+}


ktor-client/ktor-client-features/ktor-client-json/common/src/io/ktor/client/features/json/JsonFeature.kt
```diff
@@ -66,10 +66,20 @@ class JsonFeature internal constructor(
...
                 field = newList
             }
+
+        /**
+         * Adds accepted content types.
+         * It also affects `Accept` request header value.
+         */
+        @KtorExperimentalAPI
+        fun accept(vararg contentTypes: ContentType) {
+            acceptContentTypes = acceptContentTypes + contentTypes
```
e5l: append

ktor-client/ktor-client-features/ktor-client-json/common/src/io/ktor/client/features/json/JsonFeature.kt
```diff
@@ -66,10 +66,20 @@ class JsonFeature internal constructor(
          */
         @KtorExperimentalAPI
         var acceptContentTypes: List<ContentType> = listOf(ContentType.Application.Json)
+            @Deprecated("Use accept() function instead.")
```
e5l: ReplaceWith

ktor-client/ktor-client-features/ktor-client-json/common/src/io/ktor/client/features/json/JsonFeature.kt
cy6erGn0m: 1. `+=`
2. Deprecation warning should be suppressed here

ktor-client/ktor-client-features/ktor-client-json/ktor-client-json-tests/jvm/src/io/ktor/client/features/json/tests/JsonTest.kt
```diff
@@ -151,7 +151,7 @@ abstract class JsonTest : TestWithKtor() {
     @Test
     fun testCustomContentTypesMultiple() = testWithEngine(CIO) {
         configCustomContentTypeClient {
-            acceptContentTypes = listOf(ContentType.Application.Json, customContentType)
+            accept(ContentType.Application.Json, customContentType)
```
cy6erGn0m: So we accept `application/json` twice?

ktor-client/ktor-client-features/ktor-client-json/ktor-client-json-tests/jvm/src/io/ktor/client/features/json/tests/JsonTest.kt
```diff
@@ -113,7 +113,7 @@ abstract class JsonTest : TestWithKtor() {
     @Test
     fun testCustomContentTypes() = testWithEngine(CIO) {
         configCustomContentTypeClient {
-            acceptContentTypes = listOf(customContentType)
+            accept(customContentType)
```
cy6erGn0m: Only the custom content type was accepted before the change, now we accept both custom and json

ktor-client/ktor-client-features/ktor-client-json/ktor-client-json-tests/jvm/src/io/ktor/client/features/json/tests/JsonTest.kt
dmitrievanthony: Well, I have no idea, I didn't change the logic, just updated the syntax in tests.

ktor-client/ktor-client-features/ktor-client-json/common/src/io/ktor/client/features/json/JsonFeature.kt
cy6erGn0m: How can we provide it? This is not possible, I believe.

ktor-client/ktor-client-features/ktor-client-json/common/src/io/ktor/client/features/json/JsonFeature.kt
dmitrievanthony: I made `acceptContentTypes` mutable and added the suppression.

dmitrievanthony: I made `acceptContentTypes `, so please have a look.

ktor-client/ktor-client-features/ktor-client-json/common/src/io/ktor/client/features/json/JsonFeature.kt
dmitrievanthony: I added `replaceWith = ReplaceWith("accept(newList, onlyThese = true)")`. Not sure if it works, but hope so.

ktor-client/ktor-client-features/ktor-client-json/ktor-client-json-tests/jvm/src/io/ktor/client/features/json/tests/JsonTest.kt
dmitrievanthony: Fixed.

ktor-client/ktor-client-features/ktor-client-json/ktor-client-json-tests/jvm/src/io/ktor/client/features/json/tests/JsonTest.kt
dmitrievanthony: Fixed.

ktor-client/ktor-client-features/ktor-client-json/common/src/io/ktor/client/features/json/JsonFeature.kt
cy6erGn0m: No, it doesn't (just tested). And it couldn't. Because setter is used by all different operators like `=`, `+=`, `-=` and so on.

cy6erGn0m: Note that deprecation itself works well for all the operators, but not replacements.

dmitrievanthony: Thanks for testing, but I don't see any reason why it could not work in theory. The Idea knows how to replace `+=` and `-=` to `list = list + ...` and at the same time it's clear how to make the replacement of `= list` to `accept(list)`. Hope it will be implemented eventually :)

ktor-client/ktor-client-features/ktor-client-json/common/src/io/ktor/client/features/json/JsonFeature.kt
```diff
@@ -65,11 +65,30 @@ class JsonFeature internal constructor(
...
                 require(newList.isNotEmpty()) { "At least one content type should be provided to acceptContentTypes" }
                 field = newList
             }
+
+        /**
+         * Adds accepted content types. Be aware that [ContentType.Application.Json] is accepted by default. Depending
+         * on the given [onlyThese] property previously added content types will be removed or will remain.
+         * It also affects `Accept` request header value.
+         */
+        @KtorExperimentalAPI
```
cy6erGn0m: Well, still there are issues with this. The first one is that one may accept the same content type multiple times. Shouldn't we use a set instead of the list or a check before adding? Also, the flag makes this function order-sensitive. Consider the following example:

```kotlin
install(JsonFeature) {
    accept(ContentType.Application.OctetStream, onlyThese = true) 
    accept(OctetStream.Application.Zip)
    accept(OctetStream.Application.FormUrlEncoded)
    accept(OctetStream.Application.ProblemXml)
}
```

So what if I want to simply reorder lines (to make it look "more beautiful", sort by length or something)? What is a change of that I forget about `onlyThese`? This `onlyThese` could be only at the beginning and makes no sense in the middle or at the last accept invocation, right?

dmitrievanthony: I added `onlyThese` to cover the case when we need to remove the default content type. Well, let's think about how we can avoid it.

cy6erGn0m: We may always apply the flag at the first accept invocation.

```kotlin
install(JsonFeature) { // only application/octet-stream
   accept(ContentType.Application.OctetStream)
}
```

```kotlin
install(JsonFeature) // the default application/json
```

```kotlin
install(JsonFeature) {
   accept(ContentType.Application.Json)
   accept(ContentType.Application.OctetStream)
}
```

dmitrievanthony: Good point, done.

ktor-client/ktor-client-features/ktor-client-json/common/src/io/ktor/client/features/json/JsonFeature.kt
```diff
@@ -66,10 +66,34 @@ class JsonFeature internal constructor(
...
+
+        /**
+         * Adds accepted content types. Be aware that [ContentType.Application.Json] accepted by default is removed from
+         * the list if you use this function to provide accepted content types.
+         * It also affects `Accept` request header value.
+         */
+        @KtorExperimentalAPI
+        fun accept(vararg contentTypes: ContentType) {
+            @Suppress("DEPRECATION")
+            if (!acceptAlreadyCalled) {
```
cy6erGn0m:
```
acceptAlreadyCalled = true
```

cy6erGn0m: Looks like a test is required for those 3 cases :)

ktor-client/ktor-client-features/ktor-client-json/common/src/io/ktor/client/features/json/JsonFeature.kt
cy6erGn0m: I see, IDEA could do it, right. The only issue is that `accept` is not `=`, it's `+=`. So there is no valid replacement for `-=` :)

ktor-client/ktor-client-features/ktor-client-json/common/src/io/ktor/client/features/json/JsonFeature.kt
dmitrievanthony: Your truth, I added the tests :)

ktor-client/ktor-client-features/ktor-client-json/common/src/io/ktor/client/features/json/JsonFeature.kt
dmitrievanthony: Well, that's the truth. So, what's the decision? Do you think it makes sense to remove the replacement?

ktor-http/common/src/io/ktor/http/ContentTypes.kt
```diff
@@ -121,6 +121,7 @@ class ContentType private constructor(val contentType: String, val contentSubtyp
         val Any = ContentType("application", "*")
         val Atom = ContentType("application", "atom+xml")
         val Json = ContentType("application", "json")
+        val JsonHal = ContentType("application", "json+hal")
```
thumannw: According to spec, the type is `application/hal+json`, see e.g. http://stateless.co/hal_specification.html.

dmitrievanthony: You are right, thank you, I haven't checked it. Anyway, I fixed it.

ktor-client/ktor-client-features/ktor-client-json/common/src/io/ktor/client/features/json/JsonFeature.kt
```diff
@@ -59,17 +59,33 @@ class JsonFeature internal constructor(
          */
         var serializer: JsonSerializer? = null
 
+        /**
+         * Backing field with mutable list of content types that are handled by this feature.
+         */
+        private val acceptContentTypesMutable: MutableList<ContentType> = mutableListOf(ContentType.Application.Json)
```
e5l: We have the convention of naming mutable backing fields as: `_ acceptContentTypes`

PR #1642, opened 2020-02-12
Upgrade versions

cy6erGn0m:
**Subsystem**
ktor server features

**Motivation**
Some libraries are a bit outdated.

**Solution**
Upgrade libraries where possible. Webjars requires an additional fix since it started to crash with NPE.

### DIFF
--- a/ktor-features/ktor-metrics/build.gradle
+++ b/ktor-features/ktor-metrics/build.gradle
@@ -3,8 +3,8 @@ kotlin {
     sourceSets {
         jvmMain {
             dependencies {
-                api group: 'io.dropwizard.metrics', name: 'metrics-core', version: '4.1.0'
-                api group: 'io.dropwizard.metrics', name: 'metrics-jvm', version: '4.1.0'
+                api group: 'io.dropwizard.metrics', name: 'metrics-core', version: '4.1.2'
+                api group: 'io.dropwizard.metrics', name: 'metrics-jvm', version: '4.1.2'
             }
         }
     }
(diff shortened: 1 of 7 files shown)


ktor-features/ktor-metrics/build.gradle
```diff
@@ -3,8 +3,8 @@ kotlin {
     sourceSets {
         jvmMain {
             dependencies {
-                api group: 'io.dropwizard.metrics', name: 'metrics-core', version: '4.1.0'
-                api group: 'io.dropwizard.metrics', name: 'metrics-jvm', version: '4.1.0'
+                api group: 'io.dropwizard.metrics', name: 'metrics-core', version: '4.1.2'
```
e5l: could we extract all versions in `gradle.properties`?

ktor-features/ktor-webjars/jvm/src/io/ktor/webjars/Webjars.kt
```diff
@@ -22,6 +22,9 @@ import java.time.*
 @Suppress("KDocMissingDocumentation")
 @KtorExperimentalAPI
 class Webjars(private val configuration: Configuration) {
+    private val locator = WebJarAssetLocator()
+    private val knownWebJars = locator.webJars?.keys?.toSet() ?: emptySet()
+    private val lastModified = ZonedDateTime.now(configuration.zone)
```
e5l: Is it ok to use `jdk > 6` here?

cy6erGn0m: Nobody complained, it actually never worked before. There is also `ZoneId` usage. We could eliminate it separately.

Issue #1482, opened 2019-12-11
Prevent wrapping exception in `CancellationException`


e5l:
**Ktor Version and Engine Used (client or server and name)**

`1.3.0-beta-2`

**Describe the bug**
In async fail client wraps an exception in `CancellationExeption`
After introducing `CompletableJob` it'll be right to fail it instead of canceling `coroutineContext`. 

Sample can be found here: https://github.com/ktorio/ktor/blob/3a9bc56d42281bb4433aef500dc4113c2d02a6da/ktor-client/ktor-client-apache/jvm/src/io/ktor/client/engine/apache/ApacheResponseConsumer.kt#L118

dtretyakov:
@dmitrievanthony will be this fix merged into 1.3.2 release?

dmitrievanthony:
@dtretyakov, I think so. Meanwhile, I'm waiting for review so far, some concerns should be clarified.

oleg-larshin:
Please check the <a href="https://youtrack.jetbrains.com/issue/KTOR-516">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

e5l:
Fixed in 1.3.2

PR #1643, opened 2020-02-13
Unwrap cancellation exception in client code (#1482).

dmitrievanthony:
**Subsystem**
Client, `ktor-io`.

**Motivation**
The client throws `CancellationException` with the actual exception put into the cause. It is not convenient in many cases.

### DIFF
--- a/ktor-client/ktor-client-tests/js/src/io/ktor/client/tests/utils/Assertions.kt
+++ b/ktor-client/ktor-client-tests/js/src/io/ktor/client/tests/utils/Assertions.kt
@@ -11,6 +11,14 @@ import io.ktor.util.*
  */
 @InternalAPI
 actual inline fun <reified T : Throwable> assertFailsAndContainsCause(block: () -> Unit) {
+    assertFailsWith<T>(block)
+}
+
+/**
+ * Asserts that a [block] fails with a specific exception of type [T] being thrown.
+ */
+@InternalAPI
+actual inline fun <reified T : Throwable> assertFailsWith(block: () -> Unit) {
     try {
         block()
         error("Expected ${T::class} exception, but it wasn't thrown")
--- a/ktor-io/jvm/src/io/ktor/utils/io/ByteBufferChannel.kt
+++ b/ktor-io/jvm/src/io/ktor/utils/io/ByteBufferChannel.kt
@@ -70,7 +70,7 @@ internal open class ByteBufferChannel(
         job.invokeOnCompletion(onCancelling = true) { cause ->
             attachedJob = null
             if (cause != null) {
-                cancel(CancellationException("Channel closed due to job failure").apply { initCause(cause) })
+                cancel(cause)
             }
         }
     }
--- a/ktor-io/jvm/test/io/ktor/utils/io/ByteBufferChannelTest.kt
+++ b/ktor-io/jvm/test/io/ktor/utils/io/ByteBufferChannelTest.kt
@@ -1,0 +1,19 @@
+/*
+ * Copyright 2014-2020 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.utils.io
+
+import kotlinx.coroutines.*
+import java.io.*
+import kotlin.test.*
+
+class ByteBufferChannelTest {
+    @Test
+    fun testCompleteExceptionallyJob() {
+        val channel = ByteBufferChannel(false)
+        Job().also { channel.attachJob(it) }.completeExceptionally(IOException("Text exception"))
+
+        assertFailsWith<IOException> { runBlocking { channel.readByte() } }
+    }
+}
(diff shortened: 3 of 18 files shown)


ktor-io/jvm/src/io/ktor/utils/io/ByteBufferChannel.kt
```diff
@@ -70,7 +70,7 @@ internal open class ByteBufferChannel(
         job.invokeOnCompletion(onCancelling = true) { cause ->
             attachedJob = null
             if (cause != null) {
-                cancel(CancellationException("Channel closed due to job failure").apply { initCause(cause) })
+                cancel(cause)
```
cy6erGn0m: This is potentially dangerous: all write operations everywhere will produce logged errors. Most likely it will undo fixes made before to avoid I/O errors logging. Need to be changed with care

dmitrievanthony: Could you please clarify the I/O errors logging issue? I'm not sure I'm aware of it.

cy6erGn0m: The problem is that people getting undesired IO exceptions logged at server when working with channels. In real server connection issues are very common including those that caused by browser termination and user cancellation (button stop). At server side users don't want to see such exceptions or want to have them logged at the debug level.

I am not exactly sure, if it leads to such issues. May be not. But for sure this changes behaviour in many places that we can't easily trace in mind.

dmitrievanthony: Well, I do understand the problem, but I struggle to understand how the wrapping of all exceptions into `CancellationException` can help to distinguish exceptions that user wants to see in logs from exceptions he doesn't want to see.

ktor-client/ktor-client-tests/js/src/io/ktor/client/tests/utils/Assertions.kt
```diff
@@ -11,6 +11,14 @@ import io.ktor.util.*
...
 @InternalAPI
 actual inline fun <reified T : Throwable> assertFailsAndContainsCause(block: () -> Unit) {
+    assertFailsWithJsSupported<T>(block)
+}
+
+/**
+ * Asserts that a [block] fails with a specific exception of type [T] being thrown.
+ */
+@InternalAPI
+actual inline fun <reified T : Throwable> assertFailsWithJsSupported(block: () -> Unit) {
```
e5l: Consider using a different name here: `assertFailsWith` should be enough

ktor-io/jvm/test/io/ktor/utils/io/ByteBufferChannelTest.kt
```diff
@@ -0,0 +1,19 @@
...
+ * Copyright 2014-2020 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.utils.io
+
+import kotlinx.coroutines.*
+import java.io.*
+import kotlin.test.*
+
+class ByteBufferChannelTest {
```
e5l: Consider adding WebSocket tests with a cancellation

ktor-client/ktor-client-core/common/src/io/ktor/client/utils/exceptions.kt
```diff
@@ -0,0 +1,14 @@
+/*
```
e5l: filename

ktor-client/ktor-client-core/jvm/src/io/ktor/client/utils/exceptions.kt
```diff
@@ -0,0 +1,21 @@
...
+import io.ktor.util.*
+import io.ktor.utils.io.*
+
+/**
+ * If the exception contains cause that differs from [CancellationException] returns it otherwise returns itself.
+ */
+@InternalAPI
+actual fun Throwable.unwrapCancellationException(): Throwable {
+    var exception: Throwable? = this
+    while (exception is CancellationException) {
```
e5l: Could we have an infinite loop here?

ktor-client/ktor-client-core/common/src/io/ktor/client/utils/exceptions.kt
dmitrievanthony: I'm not sure I understand your point, could you clarify it?

ktor-client/ktor-client-tests/js/src/io/ktor/client/tests/utils/Assertions.kt
dmitrievanthony: Yes, I fixed.

ktor-io/jvm/test/io/ktor/utils/io/ByteBufferChannelTest.kt
dmitrievanthony: Not sure how it relates to the unwrapping of cancellation exception, but I added `WebSocketTest.testCancel`, please have a look.

ktor-client/ktor-client-core/jvm/src/io/ktor/client/utils/exceptions.kt
dmitrievanthony: Yep, we can. Fixed.

Issue #1333, opened 2019-09-10
Apache HTTP Client does not send Content-Length header if body is empty content


SIY1121:
**Ktor Version and Engine Used (client or server and name)**
io.ktor:ktor-client-core:1.2.4
io.ktor:ktor-client-core-jvm:1.2.4
io.ktor:ktor-client-apache:1.2.4

**Describe the bug**
**Apache** HTTP Client does not send Content-Length header if body is empty content.
CIO send the header correctly.

**To Reproduce**
1. Write the following
```kotlin
val apacheClient = HttpClient(Apache)
apacheClient.put<HttpResponseData>("http://example.com/") 
apacheClient.post<HttpResponseData>("http://example.com/")
apacheClient.patch<HttpResponseData>("http://example.com/")
```
2. Run on jvm
3. The sent header does not contain Content-Length
```
{
  "accept-charset": "UTF-8",
  "accept": "*/*",
  "user-agent": "Ktor client",
  "host": "example.com",
  "connection": "Keep-Alive" 
}
```

Some endpoints that do not require body return **411 Length Required**.

**Expected behavior**
`Content-Length: 0` Should be sent.

Apache seems not to send `Content-Length` if entity is null.
Following code sent the header correctly.
```kotlin
apacheClient.put<HttpResponseData>("http://example.com/"){
    body = ""
}
```

The relevant part of the code in ktor:
https://github.com/ktorio/ktor/blob/f3858b2bd95dcb35aa7755451403e7cab1dd678c/ktor-client/ktor-client-apache/jvm/src/io/ktor/client/engine/apache/ApacheRequestProducer.kt#L122-L155

and org.apache.http.client.methods.RequestBuilder 's build() method.

SIY1121:
No. It's a client issue, not a server issue.
I tried to send a empty put request to a third-party api server.

```kotlin
val apacheClient = HttpClient(Apache)
apacheClient.put<HttpResponseData>("http://example.com/") 
```

However, the server returns **411 Length Required**.
The cause is Ktor client with Apache engine didn't send content-length header.

content-length or transfer-encoding headers should be present. 

[RFC 7230, section 3.3.2: Content-Length](https://tools.ietf.org/html/rfc7230#section-3.3.2)
>  For example, a Content-Length header
   field is normally sent in a POST request even when the value is 0
   (indicating an empty payload body).

christophpickl:
[...]

PR #1644, opened 2020-02-13
#1333 Set content-length for empty request payload for Apache Http Client

christophpickl:
**Subsystem**
ktor-client

**Motivation**
When sending a POST/PUT without a payload, then some servers might reject the request with a `411 Length Required`. Other HTTP client implementations (CIO) are properly sending this header (`content-length = 0`), only the Apache client doesn't send the right header.

Refers to: #1333 (Apache HTTP Client does not send Content-Length header if body is empty content)

**Solution**
Simply check whether A) there is no request body and B) we are dealing with POST/PUT (which are the only HTTP methods which support a request body, so we don't want to do it for GET requests for example), and in that case hardcoded set the `content-length` header to `0`.

### DIFF
--- a/ktor-client/ktor-client-android/jvm/src/io/ktor/client/engine/android/AndroidClientEngine.kt
+++ b/ktor-client/ktor-client-android/jvm/src/io/ktor/client/engine/android/AndroidClientEngine.kt
@@ -64,8 +64,12 @@ class AndroidClientEngine(override val config: AndroidEngineConfig) : HttpClient
 
             config.requestConfig(this)
 
-            if (outgoingContent !is OutgoingContent.NoContent) {
-                if (data.method in listOf(HttpMethod.Get, HttpMethod.Head)) error(
+
+            if (outgoingContent is OutgoingContent.NoContent && data.method.supportsBody) {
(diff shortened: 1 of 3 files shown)


cy6erGn0m:
Remember that all methods may have request payload (including even HEAD, [RFC7231 sec 4.3.2](https://tools.ietf.org/html/rfc7231#section-4.3.2)). So, using `isSupportingPayload` function could be only useful if Apache itself doesn't support payloads for some reason.

christophpickl:
right, i was thinking and doing "some" research about that, but seems i was wrong about my assumption. thanks for the input 👍 
so that basically means that the content-length should ALWAYS be set for empty payloads, nevertheless which HTTP method being used, right?

christophpickl:
FYI: there was a failing test for the android client when i've tried to set the content-length (via `setFixedLengthStreamingMode()` which required a `doOutput=true`) also for GET/HEAD requests. solution was to get back to my initial attempt (do NOT support content-length for those HTTP methods), and weaken the http client, not asserting on the `content-length` header for GET requests.
sounds good enough for you?

PR #1645, opened 2020-02-13
Cleanup webjars feature API, reduce JDK7+ dependencies

cy6erGn0m:
**Subsystem**
server: ktor-webjars

**Motivation**
The module depends on JDK7+ API with no actual reason. 

**Solution**
Reduce unnecessary dependencies, deprecate useless API.


cy6erGn0m:
Merged

PR #1649, opened 2020-02-13
Add option to disable URL encoding

kevincianfarini:
These code changes are made by another contributor, @uharaqo who wasn't responding. The original pull request can be found from #1352 

We've got someone who is willing to review, @e5l, who requested that the feature be rebased on master. I've gone ahead and squashed and rebased this PR on master. 

I hope that is all the context needed to get this reviewed :)

### DIFF
--- a/ktor-http/common/src/io/ktor/http/HttpUrlEncoded.kt
+++ b/ktor-http/common/src/io/ktor/http/HttpUrlEncoded.kt
@@ -39,9 +39,7 @@ fun List<Pair<String, String?>>.formUrlEncode(): String = StringBuilder().apply 
  * Encode form parameters from a list of pairs to the specified [out] appendable
  */
 fun List<Pair<String, String?>>.formUrlEncodeTo(out: Appendable) {
-    joinTo(
-        out, "&"
-    ) {
+    joinTo(out, separator = "&") {
         val key = it.first.encodeURLParameter(spaceToPlus = true)
         if (it.second == null) {
             key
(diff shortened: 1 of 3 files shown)


cy6erGn0m:
So much work here. Shouldn't we simply provide a way to specify `var URLBuilder.queryString: String` instead? So people may put any characters by any convention they need? The only problem is that we'll have both `queryString` and `parameters` here. I believe, we already discussed it, didn't we @e5l ?

kevincianfarini:
@cy6erGn0m what was the result of that discussion? 

As an API consumer of ktor, having both `queryString` exposed as well as `parameters` would be confusing. It would not be clear to be that the raw `queryString` should be used to bypassing URL encoding. 

It would be nice if there were one, and exactly one, way to deal with parameters.

cy6erGn0m:
The other option would be providing a lambda to customize encoding: `var queryStringEncoder: (Parameters) -> String`

cy6erGn0m:
What if somebody says at some point that they need to encode parameters by protobuf and then encode to query string using BASE64? Should we provide an option too?

kevincianfarini:
I like the lambda option. Seems flexible and robust. 

> What if somebody says at some point that they need to encode parameters by protobuf and then encode to query string using BASE64? Should we provide an option too?

This is a valid point. I don't want to discount it saying it's never going to happen, but I haven't seen any other http library which allows something like this. However, it is fairly commonplace that disabling url encoding is an option.

kevincianfarini:
> The other option would be providing a lambda to customize encoding: var queryStringEncoder: (Parameters) -> String

Would the default of this option be to do "normal" URL encoding. If set, then circumvent any default encoding options?

Issue #1626, opened 2020-02-07
What to use in place of deprecated takeFrom(HttpRequestBuilder)?


brewin:
With HttpRequestBuilder.takeFrom(HttpRequestBuilder) deprecated in 1.3.1, what's the recommended way to reuse common request data?

dmitrievanthony:
Hi @brewin,

We initially assumed `HttpRequestBuilder.takeFrom` as an internal method. Starting from `1.3.1` started to copy `executionContext` of the initial request builder and so that became a bit more "dangerous" so we marked it as deprecated.

Could you describe your use case for me to understand how we can bypass the problme?

brewin:
I may have been using it incorrectly then. My use case is like this:

```kotlin
class SomeDataSource(private val httpClient: HttpClient) {

    private val requestBuilder = HttpRequestBuilder().apply {
        url {
            protocol = URLProtocol.HTTPS
            host = "example.com"
        }
        parameter("param_shared_by_all_requests", 42)
    }
    
    suspend fun getSomething(): Something = httpClient.get {
        takeFrom(requestBuilder)
        parameter("foo", "bar")
    }

    suspend fun getSomethingElse(): SomethingElse = httpClient.get {
        takeFrom(requestBuilder)
        parameter("baz", true)
    }
}
```

brewin:
Added back in 1.3.2. Thanks.

PR #1655, opened 2020-02-18
Recover HttpRequestBuilder.takeFrom function (#1626).

dmitrievanthony:
**Subsystem**
Client, `HttpRequestBuilder`

**Motivation**
When we deprecated `takeFrom` we lost the support of important use case described in https://github.com/ktorio/ktor/issues/1626.

**Solution**
Make a separate `takeFromWithExecutionContext` method for internal usage and leave `takeFrom` to be used in client code.

### DIFF
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/request/HttpRequest.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/request/HttpRequest.kt
@@ -117,9 +117,15 @@ class HttpRequestBuilder : HttpMessageBuilder {
      * Mutates [this] copying all the data from another [builder] using it as base.
      */
     @InternalAPI
-    @Deprecated("For internal usage only")
-    fun takeFrom(builder: HttpRequestBuilder): HttpRequestBuilder {
+    fun takeFromWithExecutionContext(builder: HttpRequestBuilder): HttpRequestBuilder {
         executionContext = builder.executionContext
+        return takeFrom(builder)
+    }
+
+    /**
+     * Mutates [this] copying all the data but execution context from another [builder] using it as base.
+     */
+    fun takeFrom(builder: HttpRequestBuilder): HttpRequestBuilder {
         method = builder.method
         body = builder.body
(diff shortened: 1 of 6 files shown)


ktor-client/ktor-client-core/common/src/io/ktor/client/request/HttpRequest.kt
```diff
@@ -117,9 +117,15 @@ class HttpRequestBuilder : HttpMessageBuilder {
...
-    fun takeFrom(builder: HttpRequestBuilder): HttpRequestBuilder {
+    fun takeFromWithExecutionContext(builder: HttpRequestBuilder): HttpRequestBuilder {
         executionContext = builder.executionContext
+        return takeFrom(builder)
+    }
+
+    /**
+     * Mutates [this] copying all the data but execution context from another [builder] using it as base.
+     */
+    fun takeFrom(builder: HttpRequestBuilder): HttpRequestBuilder {
```
cy6erGn0m: @e5l The reason for the problem is a mistake we did. I believe it is a broken design to keep a job in the builder. It looks even more strange if we notice that the builder could be built multiple times. Can't we eliminate the context (job) at all in the builder? Can't we keep it somewhere else?

dmitrievanthony: As far as I see it we use `HttpRequestBuilder` as a context of the request and send pipelines. We pass through all steps and uses them in many places and we use `executionContext` it contains also.

cy6erGn0m: I believe that request building and request execution are separate things and only the second is tracked and executed under a job while the first could be prepared at any time or, in theory, even serialized/deserialized.

cy6erGn0m: What I am saying is that just a mental model. Currently, we have them mixed up.

Issue #1657, opened 2020-02-19
Expose LocationInfo in ktor locations


bizarre:
**Subsystem**
Server

**Is your feature request related to a problem? Please describe.**
We wanted our typed locations to be able to define a permission node / access level. We did this via an additional annotation but then we had to write an interceptor that would get the class annotated with `@Location` **from the request path** and check for ours. The problem is that there doesn't seem to be any public access to the mapping done in Locations. Right now we're using a pretty-narly looking reflection solution that's less than ideal.

**Describe the solution you'd like**
Exposing `io.ktor.locations.Locations$LocationInfo` and an immutable copy of `io.ktor.locations.Locations.info` should be enough.

**Motivation to include to ktor**
I'm sure this could be a beneficial feature to a lot of users, even if it's not implemented in the way described above. It'll allow users to add additional properties to their routes and implement them in interceptors (and elsewhere) without having to do any reflection or map what is already mapped within _ktor.locations_.

oleg-larshin:
Please check the <a href="https://youtrack.jetbrains.com/issue/KTOR-468">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

e5l:
Fixed in 1.5.3

PR #1658, opened 2020-02-19
Expose LocationInfo (#1657)

bizarre:
**Subsystem**
Server

**Motivation**
This PR exposes `LocationInfo` in `io.ktor.locations.Locations`. This'll allow, for example, the ability to get a defined class annotated with `@Location` from a URL path. See #1657 .

**Solution**
I exposed `io.ktor.locations.Locations.LocationInfo` and added a public function that returns an immutable list of the currently mapped `LocationInfo`'s.

```kotlin
fun getLocationInfos() = info.values.toList()
```


cy6erGn0m:
Fixed in a separate PR

PR #1675, opened 2020-02-24
Switch conditional branch PlatformUtils.IS_NODE and PlatformUtils.IS_BROWSER

### LINKED ISSUE #1674: "TypeError: this.local$body.on is not a function" on ktor-client-js with node-fetch environment
Version: 1.3.1
Engine: Js-Client

**Describe the bug**
The module specifying `js { nodejs() }` and building called from JavaScript, throws "TypeError".

For example, the code blow.

```kotlin
private const val HOSTNAME = "sparql.crssnky.xyz"
private const val ENDPOINT = "/spql/imas/query"

val client = HttpClient(Js) {
    defaultRequest {
        accept(ContentType("application", "sparql-results+json"))
    }
    install(JsonFeature) {
        serializer = KotlinxSerializer(Json.nonstrict)
        acceptContentTypes += ContentType("application", "sparql-results+json")
    }
}

// -> Throws "TypeError: this.local$body.on is not a function"! 
val response = httpClient.get<Response>(buildQuery()) {
    url {
        protocol = URLProtocol.HTTPS
        host = HOSTNAME
    }
}

private fun buildQuery(): String = buildString {
    append(ENDPOINT)
    append("?output=json")
    append("&query=")

    val query = """
        some queries...
    """.trimIndent()

    append(URLEncoder.encode(query))
}

@Serializable
data class Response(...)
```

Repository URL: https://github.com/subroh0508/ktor-client-mpp-sample/tree/by-node-fetch

**Expected behavior**
Does not throw any errors on node-fetch environment.

**Screenshots**
MacScreenShot 2020-02-24 23 13 39

subroh0508:
**Subsystem**
Js-Client

**Motivation**
Fixes #1674 .

**Solution**
`Response#body` returns `ReadableStream`. But, it does not have `on` method.
So I replaced `Response#arrayBuffer` method instead of `ReadableStream` methods on `CoroutineScope.readBodyNode`.

And I use `jsRequire`, always throws `"Error loading module 'node-fetch': Error: Cannot find module 'node-fetch'"`. So, I deleted the method.

The operation has been confirmed in my sample repository.

### DIFF
--- a/ktor-client/ktor-client-core/js/src/io/ktor/client/engine/js/compatibility/Utils.kt
+++ b/ktor-client/ktor-client-core/js/src/io/ktor/client/engine/js/compatibility/Utils.kt
@@ -22,10 +22,10 @@ internal suspend fun commonFetch(
         controller.abort()
     }
 
-    val promise: Promise<Response> = if (PlatformUtils.IS_NODE) {
-        jsRequire("node-fetch")(input, init)
-    } else {
+    val promise: Promise<Response> = if (PlatformUtils.IS_BROWSER) {
         fetch(input, init)
+    } else {
+        jsRequire("node-fetch")(input, init)
     }
 
     promise.then(
@@ -39,20 +39,20 @@ internal suspend fun commonFetch(
 }
 
 internal fun AbortController(): AbortController {
-    return if (PlatformUtils.IS_NODE) {
+    return if (PlatformUtils.IS_BROWSER) {
+        js("new AbortController()")
+    } else {
         val controller = js("require('abort-controller')")
         js("new controller()")
-    } else {
-        js("new AbortController()")
     }
 }
 
 internal fun CoroutineScope.readBody(
     response: Response
-): ByteReadChannel = if (PlatformUtils.IS_NODE) {
-    readBodyNode(response)
-} else {
+): ByteReadChannel = if (PlatformUtils.IS_BROWSER) {
     readBodyBrowser(response)
+} else {
+    readBodyNode(response)
 }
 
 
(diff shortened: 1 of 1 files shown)


e5l:
Hi @subroh0508, thanks for the PR. 
It looks like the issue problem is different. Btw it looks like `arrayBuffer` doesn't work well with streaming body or long polling, could you add some tests to check that?

subroh0508:
@e5l I see. The issue problem is using `node-fetch`, despite being able to access `window.fetch`, isn't it?

Fixed my code, and I'm glad if you can review.

ktor-client/ktor-client-core/js/src/io/ktor/client/engine/js/compatibility/Utils.kt
```diff
@@ -22,10 +22,10 @@ internal suspend fun commonFetch(
         controller.abort()
     }
 
-    val promise: Promise<Response> = if (PlatformUtils.IS_NODE) {
-        jsRequire("node-fetch")(input, init)
-    } else {
+    val promise: Promise<Response> = if (PlatformUtils.IS_BROWSER) {
```
e5l: We saw that sometimes it can work in the opposite. Let's dig in some details: could you tell me the values of the internal predicates in `PlatformUtils.IS_BROWSER` and `IS_NODE` on your platform? Maybe we can force them to work well on each platform.

subroh0508: On my platform at the line

```
console.log(window)
// -> global {parent: global, postMessage: f, blur: f, ...}
console.log(window.document)
// -> document { URL: "http://localhost:8088/", activeElement: body, ... }

console.log(process)
// -> process {version: "v12.8.1", versions: {...}, arch: "x64", platform: "darwin", ...}
console.log(process.versions)
// -> {node: "12.8.1", v8: "7.8.279.23-electron.0". uv: "1.30.1", ...}
console.log(process.versions.node)
// -> "12.8.1"
```

e5l:
Thanks a lot. Will be merged after the remote run!

Issue #1275, opened 2019-08-10
Websocket close on client side (JavaFX webkit) breaks future websocket connections


altavir:
**Ktor Version and Engine Used (client or server and name)**
CIO server on Windows, version 1.2.3 + websockets.

**Describe the bug**
The page opens a websocket connection on connect and server starts to send data with small intervals. When page is reloaded or switched, the socket should be closed and server should receive close event. It works as expected in chrome, but in JavaFX browser, the following error is thrown:
```
java.io.IOException: An established connection was aborted by the software in your host machine
	at java.base/sun.nio.ch.SocketDispatcher.write0(Native Method)
	at java.base/sun.nio.ch.SocketDispatcher.write(SocketDispatcher.java:51)
	at java.base/sun.nio.ch.IOUtil.writeFromNativeBuffer(IOUtil.java:113)
	at java.base/sun.nio.ch.IOUtil.write(IOUtil.java:58)
	at java.base/sun.nio.ch.IOUtil.write(IOUtil.java:50)
	at java.base/sun.nio.ch.SocketChannelImpl.write(SocketChannelImpl.java:466)
	at io.ktor.network.sockets.CIOWriterKt$attachForWritingDirectImpl$1$1.invokeSuspend(CIOWriter.kt:75)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
	at kotlinx.coroutines.DispatchedContinuation.resumeWith(Dispatched.kt:108)
	at kotlinx.coroutines.io.internal.CancellableReusableContinuation.resumeWith(CancellableReusableContinuation.kt:93)
	at kotlinx.coroutines.io.ByteBufferChannel.resumeReadOp(ByteBufferChannel.kt:2211)
	at kotlinx.coroutines.io.ByteBufferChannel.flushImpl(ByteBufferChannel.kt:156)
	at kotlinx.coroutines.io.ByteBufferChannel.flush(ByteBufferChannel.kt:162)
	at kotlinx.coroutines.io.ByteBufferChannel.flushImpl(ByteBufferChannel.kt:140)
	at kotlinx.coroutines.io.ByteBufferChannel.flush(ByteBufferChannel.kt:162)
	at io.ktor.http.cio.websocket.WebSocketWriter.drainQueueAndSerialize(WebSocketWriter.kt:120)
	at io.ktor.http.cio.websocket.WebSocketWriter.writeLoop(WebSocketWriter.kt:47)
[...]

PR #1677, opened 2020-02-25
Fix DefaultWebSocketSession that does not complete closeReason on disconnection.

dmitrievanthony:
**Subsystem**
`ktor-http`, `ktor-http-cio`, websockets.

**Motivation**
https://github.com/ktorio/ktor/issues/1275#issuecomment-589670631

**Solution**
Cancel `closeReason` in case incoming channel is closed.

### DIFF
--- a/ktor-http/ktor-http-cio/jvm/src/io/ktor/http/cio/websocket/DefaultWebSocketSessionImpl.kt
+++ b/ktor-http/ktor-http-cio/jvm/src/io/ktor/http/cio/websocket/DefaultWebSocketSessionImpl.kt
@@ -11,6 +11,7 @@ import io.ktor.utils.io.pool.*
 import kotlinx.atomicfu.*
 import kotlinx.coroutines.*
 import kotlinx.coroutines.channels.*
+import java.lang.IllegalStateException
 import java.nio.*
 import kotlin.coroutines.*
 
@@ -92,11 +93,13 @@ class DefaultWebSocketSessionImpl(
         IncomingProcessorCoroutineName + Dispatchers.Unconfined
     ) {
         var last: BytePacketBuilder? = null
+        var closeFramePresented = false
         try {
             raw.incoming.consumeEach { frame ->
                 when (frame) {
                     is Frame.Close -> {
                         outgoing.send(Frame.Close(frame.readReason() ?: NORMAL_CLOSE))
+                        closeFramePresented = true
                         return@launch
                     }
                     is Frame.Pong -> pinger.value?.send(frame)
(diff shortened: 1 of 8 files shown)


ktor-http/ktor-http-cio/jvm/test/io/ktor/tests/http/cio/websocket/DefaultWebSocketSessionImplTest.kt
```diff
@@ -0,0 +1,50 @@
...
+        val closeReason = runBlocking {
+            withTimeout(1000) {
+                try {
+                    defaultSession.closeReason.await()
+                } catch (_: CancellationException) { /* Do nothing. */
+                }
+            }
+        }
+
+        assertNotNull(closeReason)
```
cy6erGn0m: What a non-null value do we have here? `Unit`?

ktor-http/ktor-http-cio/jvm/src/io/ktor/http/cio/websocket/DefaultWebSocketSessionImpl.kt
```diff
@@ -131,6 +131,9 @@ class DefaultWebSocketSessionImpl(
             ponger.close()
             last?.release()
             filtered.close()
+
+            // Job will be cancelled if it's not completed yet.
+            closeReasonRef.cancel()
```
cy6erGn0m: Because of this change we will never have a real close reason received from a peer, always cancellation. It will only have a value if we close the session first but not when a peer initiates the termination.

cy6erGn0m: Otherwise, we need a test to ensure that.

ktor-http/ktor-http-cio/jvm/src/io/ktor/http/cio/websocket/DefaultWebSocketSessionImpl.kt
```diff
@@ -92,11 +92,13 @@ class DefaultWebSocketSessionImpl(
         IncomingProcessorCoroutineName + Dispatchers.Unconfined
     ) {
         var last: BytePacketBuilder? = null
+        var closeFrameSent = false
         try {
             raw.incoming.consumeEach { frame ->
```
e5l: `outgoing` or `filtered` cancellation should cancel `raw.incoming`

e5l: consider adding a test to verify the correctness of `cancel()` instead of `terminate()`

e5l:
Also please consider adding `raw` session `close()` to verify that outgoing is closed.

PR #1702, opened 2020-03-05
Cy/fixes2

cy6erGn0m:
**Subsystem**
Server

**Motivation**
Several bugfixes including fix for `call.request.origin` that doesn't work properly in 1.3.1 and report IP address instead of hostname.

### DIFF
--- a/ktor-io/common/src/io/ktor/utils/io/ByteChannelSequential.kt
+++ b/ktor-io/common/src/io/ktor/utils/io/ByteChannelSequential.kt
@@ -547,7 +547,12 @@ abstract class ByteChannelSequentialBase(
 
     override fun discard(n: Int): Int {
         closedCause?.let { throw it }
-        return readable.discard(n).also { afterRead() }
+        if (n == 0) return 0
+
+        return readable.discard(n).also {
+            afterRead()
+            requestNextView(1)
+        }
     }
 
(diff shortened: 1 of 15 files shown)


ktor-utils/common/src/io/ktor/util/pipeline/Pipeline.kt
```diff
@@ -14,7 +14,7 @@ open class Pipeline<TSubject : Any, TContext : Any>(vararg phases: PipelinePhase
     /**
      * Provides common place to store pipeline attributes
      */
-    val attributes = Attributes()
+    val attributes = Attributes(true)
```
e5l: consider providing name here

ktor-io/common/src/io/ktor/utils/io/ByteChannelSequential.kt
```diff
@@ -547,14 +547,23 @@ abstract class ByteChannelSequentialBase(
 
     override fun discard(n: Int): Int {
         closedCause?.let { throw it }
-        return readable.discard(n).also { afterRead() }
+        if (n == 0) return 0
```
e5l: please consider adding test for all byte channel fixes

cy6erGn0m:
Merged.

PR #1707, opened 2020-03-06
Mime types

raharrison:
**Subsystem**
Server/http

**Motivation**
Some mime type mapping were incorrect/had typo's (I bet there are more in the file). The mime type for .mp4 files was incorrectly set to application/mp4 instead of video/mp4 due to ordering within the file.

**Solution**
Corrected typo's and ordering


cy6erGn0m:
Thanks!

PR #1712, opened 2020-03-10
Workaround crash in iOS(by introducing memory leak), add task parameter to challenge

e5l:


ktor-client/ktor-client-ios/darwin/src/io/ktor/client/engine/ios/IosClientEngineConfig.kt
```diff
@@ -12,6 +12,7 @@ import platform.Foundation.*
  */
 typealias ChallengeHandler = (
```
cy6erGn0m: This is the disadvantage of properties with functional types: we can't simply add a parameter without source breaking changes :( Looks like we should always use functions with labda instead of plain properties

E.g. 
```kotlin
class MyConfig {
    var handler: () -> Unit = {}
}
```

should be always

```kotlin
class MyConfig {
    private var handler: () -> Unit = {}

    fun handler(block: () -> Unit) {
        handler = block
    }
}
```

e5l: This param was never been released in the past

e5l: Btw I’ll replace it with the method

Issue #1727, opened 2020-03-16
Improve ktor-client request/response logging


dragneelfps:
**Subsystem**
Client, ktor-client-logging

**Is your feature request related to a problem? Please describe.**
Related to https://github.com/ktorio/ktor/pull/1638#issuecomment-599395334

**Describe the solution you'd like**
Change to `filters: List<(HttpClientCall) -> Boolean>`

**Motivation to include to ktor**
Related to https://github.com/ktorio/ktor/pull/1638#issuecomment-599395334

oleg-larshin:
Please check the <a href="https://youtrack.jetbrains.com/issue/KTOR-443">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

PR #1728, opened 2020-03-16
Improve ktor-client request/response logging

dragneelfps:
Related to https://github.com/ktorio/ktor/issues/1727

### DIFF
--- a/ktor-client/ktor-client-features/ktor-client-logging/common/src/io/ktor/client/features/logging/Logging.kt
+++ b/ktor-client/ktor-client-features/ktor-client-logging/common/src/io/ktor/client/features/logging/Logging.kt
@@ -24,12 +24,13 @@ import kotlinx.coroutines.*
 class Logging(
     val logger: Logger,
     var level: LogLevel,
-    var filters: List<(HttpRequestBuilder) -> Boolean>
+    val filters: List<(HttpClientCall) -> Boolean>,
+    val logContentType: LogContentType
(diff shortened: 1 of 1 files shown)


eMZet-zz:
@dragneelfps 
The changes seem fine.
`val logContentType: LogContentType` is a nice addition :)

eMZet-zz:
Just wondering if 
`io.ktor.client.HttpClientConfig` 's

```
/**
* Terminate [HttpClient.responsePipeline] if status code is not success(>=300).
*/
var expectSuccess: Boolean = true
```
will affect logging error responses if not changed to `false`.

ktor-client/ktor-client-features/ktor-client-logging/common/src/io/ktor/client/features/logging/Logging.kt
```diff
@@ -24,12 +24,13 @@ import kotlinx.coroutines.*
 class Logging(
     val logger: Logger,
     var level: LogLevel,
-    var filters: List<(HttpRequestBuilder) -> Boolean>
+    val filters: List<(HttpClientCall) -> Boolean>,
```
e5l: I'm not sure if it's ok to replace the request builder with the call here. The call will be created only after the network request, but we should print the request log before any network activity. 
Could you tell me if you have an idea of fixing this?

rsinukov:
Closing this PR as stale

PR #1732, opened 2020-03-17
Bug fixes

e5l:
- fix race in testMaxSize
- cleanup proxy configuration
- update okhttp version
- cleanup gradle config
- fix engine close


gradle.properties
```diff
@@ -47,7 +47,7 @@ typesafe_config_version=1.3.1
 apache_version=4.1.4
 apache_core_version=4.4.13
 gson_version=2.8.6
-okhttp_version=3.14.2
```
cy6erGn0m: Do we know, what minimal Android API level change does it introduce?

ktor-client/ktor-client-cio/jvm/src/io/ktor/client/engine/cio/CIOEngine.kt
```diff
@@ -24,11 +26,29 @@ internal class CIOEngine(override val config: CIOEngineConfig) : HttpClientEngin
...
+    private val requestsJob: CoroutineContext
+
+    override val coroutineContext: CoroutineContext
+
+    init {
+        val parent = super.coroutineContext[Job]!!
+        requestsJob = SilentSupervisor(parent)
+        coroutineContext = super.coroutineContext + requestsJob
+
+        GlobalScope.launch(super.coroutineContext) {
```
cy6erGn0m: ❗️ This will cause unclosed selector manager when the context is already cancelled. Use atomic launch mode to prevent it

ktor-client/ktor-client-cio/jvm/src/io/ktor/client/engine/cio/CIOEngine.kt
```diff
@@ -59,10 +79,7 @@ internal class CIOEngine(override val config: CIOEngineConfig) : HttpClientEngin
     override fun close() {
         super.close()
         endpoints.forEach { (_, endpoint) -> endpoint.close() }
-
-        coroutineContext[Job]!!.invokeOnCompletion {
-            selectorManager.close()
-        }
+        (requestsJob[Job] as CompletableJob).complete()
```
cy6erGn0m: ❓ Shouldn't we do it in a finally block to ensure completion? Do we know that close will never crash?

ktor-client/ktor-client-core/common/src/io/ktor/client/HttpClient.kt
```diff
@@ -109,17 +122,21 @@ class HttpClient(
...
+     * Client engine config.
      */
     val engineConfig: HttpClientEngineConfig = engine.config
 
     internal val config = HttpClientConfig<HttpClientEngineConfig>()
 
     init {
+        val engineJob = engine.coroutineContext[Job]!!
+        @Suppress("DEPRECATION_ERROR")
+        clientJob.attachChild(engineJob as ChildJob)
```
cy6erGn0m: ❓  Can't we avoid it and implement it manually?

ktor-client/ktor-client-okhttp/jvm/src/io/ktor/client/engine/okhttp/OkHttpEngine.kt
```diff
@@ -27,19 +27,39 @@ import kotlin.coroutines.*
...
     /**
      * Cache that keeps least recently used [OkHttpClient] instances.
      */
     private val clientCache = createLRUCache(::createOkHttpClient, {}, config.clientCacheSize)
+    init {
+        val parent = super.coroutineContext[Job]!!
+        requestsJob = SilentSupervisor(parent)
+        coroutineContext = super.coroutineContext + requestsJob
+
+        GlobalScope.launch(super.coroutineContext) {
```
cy6erGn0m: ❗️ same her: atomic launch

ktor-client/ktor-client-cio/jvm/src/io/ktor/client/engine/cio/CIOEngine.kt
e5l: I'll add a test with manual cancellation

gradle.properties
e5l:
```
The OkHttp 3.12.x branch supports Android 2.3+ (API level 9+) and Java 7+. 
OkHttp works on Android 5.0+ (API level 21+) and on Java 8+.
```
https://square.github.io/okhttp/
so it stays the same

Issue #1595, opened 2020-01-30
Add OAuth2/Space client authentication 


vladimirdolzhenko:
**Subsystem**
Client

**Is your feature request related to a problem? Please describe.**
There is no oauth2 client authentication, and especially JetBrains Space auth 

**Describe the solution you'd like**
I'd like to have smth like 
```
val spaceClient = client.config {
    Auth {
       Space(clientId, clientSecret) 
    }
}
```

actually to authenticate in Space I have to do smth like
```
private data class AccessToken(val token_type: String, val expires_in: Int, val access_token: String)

private suspend fun issueAccessToken(endpoint: String, clientId: String, clientSecret: String): AccessToken {
        return client.submitForm(
            url = "${endpoint}/oauth/token",
            formParameters = Parameters.build {
                append("grant_type", "client_credentials")
                append("scope", "**")
            }
        ) {
            header(
                HttpHeaders.Authorization,
                HttpAuthHeader.Single(
                    AuthScheme.Basic,
                    String(Base64.getEncoder().encode("$clientId:$clientSecret".toByteArray()))
                ).render()
            )
        }
    }
```

and later on for each query use 

```
private fun HttpRequestBuilder.authHeader() {
        header(
            HttpHeaders.Authorization,
            HttpAuthHeader.Single(issueAccessToken.token_type, issueAccessToken.access_token).render()
        )
    }
```

**Motivation to include to ktor**
OAuth2 is quite common auth way

hardysim:
I tried to write my own `OAuthAuthProvider` based on the existing providers in `io.ktor.client.features.auth.providers` which seems to work but with the restriction that I can only retry a request once.

But I might need to retry the request multiple times when using a refresh token:
1. try request
1. request fails with 401
1. the provider is asked to add an authentication-header
   - it does by adding the current access token
1. request is retried with the added header by the auth-feature
1. request fails again because the token is invalid
1. now the provider should
   - invalidate the current access token
   - get a new access token (via refresh token or username/password)
1. retry the request one more time with the new access token
1. request is ok or fails again (start over or stop)

So my question is: How can I retry the request multiple times in an `io.ktor.client.features.auth.AuthProvider`?

PS: When my provider is working, I'm happy to contribute it here.

e5l:
Hi, @hardysim. You can try using ‘HttpSend’ feature. You can add interceptor there and retry request multiple times(See HttpRedirect feature as the example)

hardysim:
[...]

PR #1733, opened 2020-03-17
set AuthScheme.OAuth to `Bearer` as defined in RFC6750

hardysim:
In RFC6750 the auth-scheme is defined as `Bearer` (not `OAuth`). See https://tools.ietf.org/html/rfc6750#section-3:

> All challenges defined by this specification MUST use the auth-scheme value "Bearer".

This is also commonly used by servers which respond with `Bearer` as well so this must match to be used by an (not yet implemented) `OAuthAuthProvider` (see https://github.com/ktorio/ktor/issues/1595).

### DIFF
--- a/ktor-http/common/src/io/ktor/http/auth/AuthScheme.kt
+++ b/ktor-http/common/src/io/ktor/http/auth/AuthScheme.kt
@@ -40,11 +40,11 @@ object AuthScheme {
     const val Negotiate = "Negotiate"
 
     /**
-     * OAuth Authentication described in the RFC-6749:
+     * OAuth Authentication described in the RFC-6750:
      *
-     * see https://tools.ietf.org/html/rfc6749
+     * see https://tools.ietf.org/html/rfc6750#section-3
      */
-    const val OAuth = "OAuth"
+    const val OAuth = "Bearer"
 
     @Suppress("KDocMissingDocumentation", "unused")
     @Deprecated("Compatibility", level = DeprecationLevel.HIDDEN)


hardysim:
Any progress on this? It's not that much to review ;)

cy6erGn0m:
Sorry for the huge delay. But I see the problem: this value is passed in OAuth1.a implementation when obtaining a token. So it couldn't be simply changed. Most likely the RFC number is wrong that makes confusion. I believe we simply need to add Bearer constant instead of replacing the existing one

hardysim:
@cy6erGn0m which RFC do you mean? I changed RFC-6749 to RFC-6750 but both are about OAuth 2.0. I want to use this on OAuth 2.0 (only - I don't think anyone is still using OAuth 1).

Do you mean this constant is used in multiple places (and one of it is the OAuth 1 flow)? Then a second constant would be better, correct.

---

I see that #1740 has been merged already. In there, I added an extension-property `AuthScheme.Bearer` so there we have our additional constant. Just not in a "nice" place.

But I think it has been merged into `e5l/develop` only so it might not be in the actual `master`.

@e5l Any updates on this? [You mentioned, you wanted to check binary compatibility and add some tests](https://github.com/ktorio/ktor/pull/1740#issuecomment-604839760).

Should we add `AuthScheme.Bearer` directly to the `AuthScheme.kt`-file? If so, this (#1733) MR would nicely complement my code from #1740 and we could merge both to `master` if you're happy with it.

LouisCAD:
Why has this issue been closed, @e5l?

e5l:
I reopened the related YT issue, will fix it in the next minor.

hardysim:
> I reopened the related YT issue, will fix it in the next minor.

Would you please link the YT issue so we can keep track of the progress? Thanks!

LouisCAD:
Yep, please, always link the related issue.

PR #1740, opened 2020-03-20
add OAuthProvider

hardysim:
**Subsystem**
client, auth-feature

**Motivation**
There is no `AuthProvider` for OAuth. This is needed in many cases.

refs #1595 (and hopefully closes it as well)

**Solution**
I've modified `Auth` to let the `AuthProvider` decide, if the request should be retried or not. This is indicated by `AuthProvider.addRequestHeaders()` now *returning* a request or not. This is just like OkHttp handles this and many people might be used to. This is why, this was also renamed to `authenticate()`.

To let the `AuthProvider` get original request as a `HttpRequestBuilder`, I needed to use `HttpRequestBuilder().takeFrom(call.request)` because `takeFromWithExecutionContext()` does not add a header which was just added by the `AuthProvider` (not sure why).

I'm not sure how to add this provider to the test cases. Plus, all tests in `AuthTest` fail on my machine (using jvm as the target).

The documentation at https://github.com/ktorio/ktorio.github.io/blob/master/clients/http-client/features/auth.md should be updated as well when this new provider is added.

---

### Usage

To use it, you need to pass a `TokenProvider` to `Auth.oauth()`. This is prepared to be used in a multiplatform project where it can be implemented by platform-dependent versions (e.g. for android you can use the systems `AccountManager`:

**common**
Use expect/actual to use different approaches for each platform.
```kotlin
expect class MultiplatformTokenProvider : TokenProvider {

    override suspend fun getToken(): String?

    override suspend fun invalidateToken(token: String)
}
```

**android**
On android, we can still use suspended functions so we can simply delegate the calls to the `AccountManager` which can then be queried in a suspended function as well.
```kotlin
actual class MultiplatformTokenProvider(
    val getTokenFromAccountManager: suspend () -> String?,
    val invalidateTokenInAccountManager: suspend (String) -> Unit
) : TokenProvider {

    actual override suspend fun getToken(): String? = getTokenFromAccountManager()

    actual override suspend fun invalidateToken(token: String) = invalidateTokenInAccountManager(token)
}
```

**ios**
On iOS we cannot use suspended functions and need to use a `Continuation` which will be passed to copies of the original interface-functions (I did not tried this out yet but should be possible).
```kotlin
actual open class MultiplatformTokenProvider : TokenProvider {

    actual override suspend fun getToken(): String? {
        return suspendCancellableCoroutine { continuation ->
            getToken(continuation)
        }
    }

    actual override suspend fun invalidateToken(token: String) {
        return suspendCancellableCoroutine { continuation ->
            invalidateToken(token, continuation)
        }
    }

    open fun getToken(continuation: CancellableContinuation<String?>) {
        TODO("Not yet implemented")
    }

    open fun invalidateToken(token: String, continuation: CancellableContinuation<Unit>) {
        TODO("Not yet implemented")
    }

}
```

### DIFF
--- a/ktor-client/ktor-client-features/ktor-client-auth/common/src/io/ktor/client/features/auth/providers/oauth/OAuthProvider.kt
+++ b/ktor-client/ktor-client-features/ktor-client-auth/common/src/io/ktor/client/features/auth/providers/oauth/OAuthProvider.kt
@@ -72,7 +72,7 @@ class OAuthProvider(
         return true
     }
 
-    override suspend fun addRequestHeaders(request: HttpRequestBuilder): HttpRequestBuilder? {
+    override suspend fun authenticate(request: HttpRequestBuilder): HttpRequestBuilder? {
         /**
          * Use [REQUEST_TRIES_HEADER] to store [tryCount].
          * Cannot use [AttributeKey] because those are not copied by [HttpRequestBuilder.takeFrom].
@@ -95,7 +95,7 @@ class OAuthProvider(
              * If so, the token did not work and should be invalidated.
              *
              * But still return the request (and not null) in this case
-             * so [addRequestHeaders] is called again and we can try a new token
+             * so [authenticate] is called again and we can try a new token
              * (which should be requested by the [TokenProvider] after the old token has been invalidated).
              */
             tokenProvider.invalidateToken(token)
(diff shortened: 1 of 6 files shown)


ktor-client/ktor-client-features/ktor-client-auth/common/src/io/ktor/client/features/auth/providers/oauth/OAuthProvider.kt
```diff
@@ -0,0 +1,113 @@
...
+import io.ktor.http.*
+import io.ktor.http.auth.*
+
+const val DEFAULT_MAX_REQUEST_TRIES = 3 // original request, refreshToken, username/password
+const val REQUEST_TRIES_HEADER = "NumRequestTries"
+
+// `AuthScheme.OAuth` is currently wrong, so we use our own (see https://github.com/ktorio/ktor/pull/1733)
+val AuthScheme.Bearer: String
+    get() = "Bearer"
+
```
hardysim: This could be removed when #1733 gets merged before.

e5l:
Merged in `e5l/develop` for checking binary compatibility and adding tests.

e5l:
Thanks for the PR! :)

Issue #1735, opened 2020-03-18
Clarify behavior of get<Unit> with appropriate error/warning message


Ribesg:
**Ktor Version and Engine Used (client or server and name)**
1.3.2, client, `ResponseObserver`

**Describe the bug**
When using `ResponseObserver` and a request like `get<Unit>(…)` to an endpoint which actually returns a body (which we would want to ignore in this case), the `response` passed to the observer has a cancelled `content: ByteReadChannel`, preventing logging of the body which was returned.

**Expected behavior**
It would be nice if there was a way to indicate why the body isn't available in this case to the developer

dmitrievanthony:
Hi @Ribesg, 

Could you clarify the exact reproducer of the issue and the desired behaviour?

I used the following code to reproduce:

```
suspend fun main() {
    HttpClient(CIO) {
        install(Logging) {
            logger = Logger.DEFAULT
            level = LogLevel.ALL
        }
    }.use { client ->
        val res = client.get<Unit>("http://localhost:4567/test")
        println("Result: $res")
    }
}
```

And got the following output:

```
[main] INFO io.ktor.client.HttpClient - REQUEST: http://localhost:4567/test
[main] INFO io.ktor.client.HttpClient - METHOD: HttpMethod(value=GET)
[main] INFO io.ktor.client.HttpClient - COMMON HEADERS
[main] INFO io.ktor.client.HttpClient - -> Accept-Charset: UTF-8
[main] INFO io.ktor.client.HttpClient - -> Accept: */*
[main] INFO io.ktor.client.HttpClient - CONTENT HEADERS
[main] INFO io.ktor.client.HttpClient - BODY Content-Type: null
[main] INFO io.ktor.client.HttpClient - BODY START
[main] INFO io.ktor.client.HttpClient - 
[main] INFO io.ktor.client.HttpClient - BODY END
[DefaultDispatcher-worker-1] INFO io.ktor.client.HttpClient - BODY Content-Type: text/plain; charset=UTF-8
[DefaultDispatcher-worker-1] INFO io.ktor.client.HttpClient - BODY START
[DefaultDispatcher-worker-1] INFO io.ktor.client.HttpClient - Hello, world!
[DefaultDispatcher-worker-1] INFO io.ktor.client.HttpClient - BODY END
[ktor-cio-dispatcher-worker-1] INFO io.ktor.client.HttpClient - RESPONSE: 200 OK
[...]

PR #1746, opened 2020-03-23
Fix client logging with cancelled content (#1735).

dmitrievanthony:
**Subsystem**
Client, Logging

**Motivation**
https://github.com/ktorio/ktor/issues/1735

### DIFF
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/features/observer/ResponseObserver.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/features/observer/ResponseObserver.kt
@@ -42,7 +42,7 @@ class ResponseObserver(
         override fun install(feature: ResponseObserver, scope: HttpClient) {
 
             scope.receivePipeline.intercept(HttpReceivePipeline.Before) { response ->
-                val (loggingContent, responseContent) = response.content.split(response)
+                val (loggingContent, responseContent) = response.content.splitOnMainAndDependant(response)
 
                 val newClientCall = context.wrapWithContent(responseContent)
                 val sideCall = newClientCall.wrapWithContent(loggingContent)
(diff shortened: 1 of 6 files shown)


ktor-client/ktor-client-core/common/src/io/ktor/client/features/observer/ResponseObserver.kt
```diff
@@ -42,7 +42,7 @@ class ResponseObserver(
         override fun install(feature: ResponseObserver, scope: HttpClient) {
 
             scope.receivePipeline.intercept(HttpReceivePipeline.Before) { response ->
-                val (loggingContent, responseContent) = response.content.split(response)
+                val (loggingContent, responseContent) = response.content.splitOnMainAndDependant(response)
```
e5l: Could you tell me why we need new method here? It looks like the split should propagate cancellation to all channels

ktor-utils/common/src/io/ktor/util/ByteChannels.kt
```diff
@@ -10,6 +10,37 @@ import kotlinx.coroutines.*
...
+fun ByteReadChannel.splitOnMainAndDependant(coroutineScope: CoroutineScope): Pair<ByteReadChannel, ByteReadChannel> {
+    val first = ByteChannel(autoFlush = true)
+    val second = ByteChannel(autoFlush = true)
+
+    coroutineScope.launch {
+        try {
+            while (!isClosedForRead) {
+                readRemaining(CHUNK_BUFFER_SIZE).use { chunk ->
+                    second.writePacket(chunk.copy())
+                    while (second.availableForRead != 0) { yield() }
```
e5l: codestyle

ktor-client/ktor-client-core/common/src/io/ktor/client/features/observer/ResponseObserver.kt
dmitrievanthony: Method `split` does split the channel and returns two fully asynchronous channels. As a result, logging can be fully done and completed even before the "main" channel will be canceled.

From my perspective, it's wrong and the logging channel should propagate only the data the main channel already read (it should wait for the main channel). To implement this idea we need a separate `splitOnMainAndDependant` method.

ktor-utils/common/src/io/ktor/util/ByteChannels.kt
dmitrievanthony: Fixed.

e5l:
Fixed in `1.5.0`

Issue #1431, opened 2019-11-09
SSL pinning for iOS


Rjmaurya13:
How to make SSL pinning for iOS. As I am getting response but for other thing its not working.

Getting crash near this code  
`val remoteCertificateData : NSData = SecCertificateCopyData(certificate) as NSData`

This is the error.

> Uncaught Kotlin exception: kotlin.TypeCastException

Here is my code.

`override fun URLSession(
                session: NSURLSession,
                didReceiveChallenge: NSURLAuthenticationChallenge,
                completionHandler: (NSURLSessionAuthChallengeDisposition, NSURLCredential?) -> Unit
            ) {
                val serverTrust = didReceiveChallenge.protectionSpace.serverTrust
                val certificate = SecTrustGetCertificateAtIndex(serverTrust,0)

                var result: SecTrustResultType = 0u

                memScoped{
                    val nativeResult = alloc<SecTrustResultTypeVar>()
                    nativeResult.value = result

                    SecTrustEvaluate(serverTrust!!, nativeResult.ptr)
                }

                val remoteCertificateData : NSData = SecCertificateCopyData(certificate) as NSData

                val bundle = NSBundle.bundleForClass(objc_getRequiredClass("IosClientEngine"))
                Logger.debug("pathToCert","$bundle")

                val pathToCert = bundle.pathForResource("MyCertificate","cer")

                val localCertificate : NSData = NSData.dataWithContentsOfFile(pathToCert!!)!!

                if (localCertificate == remoteCertificateData) {
                    completionHandler(NSURLSessionAuthChallengeUseCredential,NSURLCredential.create(serverTrust))
            
                } else {
                    completionHandler(NSURLSessionAuthChallengeUseCredential, null)
            
                }
            }`

danailalexiev:
Any suggestions for a workaround for self-signed certificate handling on iOS?

alistairsykes:
I recently had to tackle implementing iOS certificate pinning. Here is a blog post I just published detailing my solution. Feel free to use it to help you work out your solution. Hope it helps someone.

https://medium.com/@alistairsykes/kotlin-multiplatform-ios-certificate-pinning-fd1abba5ca8f

hardysim:
When I get it right, I need to add your `CertificatePinner` and use it like so:

```kotlin

HttpClient(Ios) {
 engine {
        challengeHandler = CertificatePinner.Builder()
[...]

PR #1750, opened 2020-03-25
iOS Certificate Pinning

alistairsykes:
**Subsystem**
Client, iOS

**Motivation**
Pinning certificates defends against attacks on certificate authorities. It also prevents connections through man-in-the-middle certificate authorities either known or unknown to the application's user.
#1431

**Solution**
Added class `CertificatePinner` which confirms to `ChallengeHandler`. In doing so it can override:
```
fun invoke(
    session: NSURLSession,
    task: NSURLSessionTask,
    challenge: NSURLAuthenticationChallenge,
    completionHandler: (NSURLSessionAuthChallengeDisposition, NSURLCredential?) -> Unit
)
```
and therefore ensure that the certificates in the serverTrust match those that are pinned.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -14,7 +14,7 @@ org.gradle.daemon=true
 org.gradle.jvmargs=-Xmx2g -XX:MaxPermSize=2048m -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8
 
 # kotlin
-kotlin_version=1.3.70
+kotlin_version=1.3.71
 kotlin.native.ignoreDisabledTargets=true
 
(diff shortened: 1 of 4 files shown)


alistairsykes:
As discussed with @e5l I have been unable to build and validate this PR due to an issue with gradle configuration. 
The tests with this class are pretty scarce. I did not want to introduce any mocking frameworks or the like, so I added tests for what I could.
Happy to work on this further, if you could point me in the right direction.

e5l:
Merged in `e5l/develop` for fixing gradle configuration. Thanks for the PR

PR #1752, opened 2020-03-26
Add Configurability to Client WebSockets

halilozercan:
**Subsystem**
Client

**Motivation**
Currently there is a small problem with WebSockets feature in ktor-client that originates from non-configurability. Here is a short explanation of how I encountered this issue:

I've been using ktor in both server and client nodes in a side project. I discovered that my client agents are disconnecting due to an internet connectivity issue and not realizing they’ve lost the connection, meanwhile the server is well aware of disconnection.

I’ve been using DefaultWebSocket which means ping-pong should have caught this issue. After some digging, it turns out that ping-pong mechanism actually is not enabled by default on client sockets. The reason is DefaultWebSocket expects a positive value for pingInterval while `WebSockets` is initialized with non-configurable `-1L`. 

Although client WebSockets feature takes pingInterval and maxFrameSize as constructor parameters, install(WebSockets) does not have a configuration option. So, I’m now stuck, not able to turn on a feature that already exists.

Furthermore, maxFrameSize is passed to DefaultWebSocketSession  as
```return DefaultWebSocketSession(this, pingInterval, maxFrameSize)```
it turns out, third parameter is actually timeoutMillis not related to frameSize at all. 

**Solution**
Simply changing the name of maxFrameSize to timeoutMillis and providing a configuration lambda in `prepare` call of the feature gives support for users to enable ping-pong check in their clients. Details of implementation can be found in the commit message.

### DIFF
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/features/websocket/WebSockets.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/features/websocket/WebSockets.kt
@@ -15,20 +15,22 @@ import io.ktor.util.*
 /**
  * Client WebSocket feature.
  *
- * @property pingInterval - interval between [FrameType.PING] messages.
- * @property maxFrameSize - max size of single websocket frame.
+ * @property pingIntervalMillis - interval between [FrameType.PING] messages.
+ * @property timeoutMillis - timeout to receive a [FrameType.PONG] message.
  */
 @KtorExperimentalAPI
 @OptIn(WebSocketInternalAPI::class)
 class WebSockets(
-    val pingInterval: Long = -1L,
-    val maxFrameSize: Long = Int.MAX_VALUE.toLong()
+    var pingIntervalMillis: Long = -1L,
+    var timeoutMillis: Long = 15000L
 ) {
     @Suppress("KDocMissingDocumentation")
-    companion object Feature : HttpClientFeature<Unit, WebSockets> {
+    companion object Feature : HttpClientFeature<WebSockets, WebSockets> {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by halilozercan** [comment]
Related documentation PR 

https://github.com/ktorio/ktorio.github.io/pull/263

File: ktor-client/ktor-client-core/common/src/io/ktor/client/features/websocket/WebSockets.kt [hunk 1]
```diff
@@ -15,20 +15,22 @@ import io.ktor.util.*
...
- * @property maxFrameSize - max size of single websocket frame.
+ * @property pingIntervalMillis - interval between [FrameType.PING] messages.
+ * @property timeoutMillis - timeout to receive a [FrameType.PONG] message.
  */
 @KtorExperimentalAPI
 @OptIn(WebSocketInternalAPI::class)
 class WebSockets(
-    val pingInterval: Long = -1L,
-    val maxFrameSize: Long = Int.MAX_VALUE.toLong()
+    var pingIntervalMillis: Long = -1L,
```
e5l: Consider keeping old names for now, we can't change them until ktor `1.4.0`

halilozercan: As I understand, we should keep the old names and default values but `maxFrameSize` would still be passed as timeoutMillis?

halilozercan:
I think changing variable names back to what they were would defeat the purpose of the fix. I'm going to reopen this pull request with `1.4.0` as destination.

PR #1755, opened 2020-03-27
#1751 uri support

### LINKED ISSUE #1751: Better URI support
Ktor 1.3.1

1. `URLBuilder.takeFrom(URI)` returns `Unit` but should return `URLBuilder` for chaining.
2. There's `Url.toURI()` but no `Url(URI)`.
    One has to use `URLBuilder().apply { takeFrom(URI) }.build()`.
    `Url(URI)` should be added for symmertry.

Excuta:
**Subsystem**
JVM/Http

**Motivation**
closes #1751 

**Solution**
 change URLBuilder.takeFrom(uri: URI) and made return URLBuilder to be able to chain calls
Added a static extension to concisely create a Url object from a java.net.URI object


e5l:
Hi @Excuta, thanks for the PR :). LGTM, Merged

PR #1769, opened 2020-04-02
Fix max-age directive on the Cache-Control header to use '=' instead of ':'.

bernatpl:
**Subsystem**
Client

**Motivation**
See https://github.com/ktorio/ktor/issues/1944
Using the ktor-client on Android with the OkHttp engine I noticed that using the `maxAge()` function to set the `Cache-Control` header had no effect on the caching strategy.
It was caused because using that function the `max-age` directive key and value are separated with ':' and the OkHttp client expected '=' as a separator.

According to the HTTP standard, the `max-age` directive of the `Cache-Control` header must use the '=' symbol instead of ':' to separate the directive key with the value.

**Solution**
The ':' separator has been replaced by '=' on the `max-age` directive key-value pair


e5l:
Hi @bernatpl, thanks for the PR. Merged.

Issue #1767, opened 2020-04-01
Get client certificate information from request


jorispz:
**Subsystem**
Server

**Is your feature request related to a problem? Please describe.**
We want to use client-side certificates to authenticate HTTPS requests from connected devices. In our code, we need to get the client certificate information to determine the DN in the certificate.

In a regular servlet, client certificate information can be accessed using `request.getAttribute("javax.servlet.request.X509Certificate")`, which returns the certificate chain.

Our strategy to get this to work in ktor was to run our application as a WAR in a Tomcat instance configured to do client authentication. Unfortunately, request attributes do not seem to be available. (Note that a Spring Boot application deployed under the same Tomcat can access these attributes, so we know Tomcat is configured correctly for TLS.)

**Describe the solution you'd like**
We'd like some way to access servlet request attributes from ktor application code.

**Motivation to include to ktor**
Client certificates are an important part of IoT deployments, where ktor would shine because typically a server needs to manage a lot of concurrent connections. Ktor itself doesn't support client certificate authentication - our solution would be a workaround if only we could access request attributes.

oleg-larshin:
Please check the <a href="https://youtrack.jetbrains.com/issue/KTOR-424">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

jorispz:
Closed see https://youtrack.jetbrains.com/issue/KTOR-424

ydanneg:
is it possible to get a client certificate using other engines, like Netty?

PR #1771, opened 2020-04-03
Expose servlet request attributes as call attributes

jorispz:
**Subsystem**
ktor-server-servlet

**Motivation**
[KTOR-424](https://youtrack.jetbrains.com/issue/KTOR-424) Get client certificate information from request 
This implements feature request #1767 by exposing servlet request attributes as call attributes

**Solution**
On instantiating a ServletApplicationRequest from a HttpServletRequest, the attributes in the servlet request are extracted and stored in the call attributes using a key that is exposed in a companion object to ServletApplicationRequest.

Note that the map with servlet attributes is also stored as a property on ServletApplicationRequest. This currently serves no purpose but has been done to be symmetrical to the headers. But one could certainly argue this is not useful

### DIFF
--- a/ktor-server/ktor-server-servlet/jvm/src/io/ktor/server/servlet/ServletApplicationRequest.kt
+++ b/ktor-server/ktor-server-servlet/jvm/src/io/ktor/server/servlet/ServletApplicationRequest.kt
@@ -8,6 +8,7 @@ import io.ktor.application.*
 import io.ktor.http.*
 import io.ktor.request.*
 import io.ktor.server.engine.*
+import io.ktor.util.*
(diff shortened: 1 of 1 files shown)


jorispz:
The opinion of stale[bot] notwithstanding, I'd still like to see this merged....

StefanOltmann:
Damn, I just migrated with my private project to Ktor because it's the new stuff. Seeing that basic features like this are missing and many month later pull requests don't get reviewed I think I need to rollback and stay away from it until it gets mature. 👎

jorispz:
Merged as part of https://github.com/ktorio/ktor/pull/2234

PR #1798, opened 2020-04-16
Locations serialization support + common

cy6erGn0m:
**Subsystem**
ktor-locations

**Description**
- Introduce support for kotlinx.serialization in locations
- Introduce common locations
- Keep backward compatible behaviour:
    - backward compatible implementation is available using key
    - the new implementation does support old behaviour on JVM only

note: commits/messages will be revisited later just before merge


e5l:
Could we close this PR?

Issue #1729, opened 2020-03-16
ktor client does not log request body


AlexCzar:
**Ktor Version and Engine Used (client or server and name)**
ver: 1.3.1, 1.3.2
engine: Apache
app: client
dependencies:
```kotlin
dependencies {
    implementation(kotlin("stdlib-jdk8"))
    implementation(platform("io.ktor:ktor-bom:1.3.2"))
    implementation("io.ktor:ktor-client-core")
    implementation("io.ktor:ktor-client-logging-jvm")
    implementation("io.ktor:ktor-client-apache")
    implementation("io.ktor:ktor-client-json-jvm")
    implementation("org.slf4j:slf4j-simple:1.7.26")
    implementation("org.apache.httpcomponents:httpasyncclient:4.1.4")
}
```

**Describe the bug**
Request body is not logged even with LogLevel.ALL

**To Reproduce**
```kotlin
package com.example

import io.ktor.client.HttpClient
import io.ktor.client.engine.apache.Apache
import io.ktor.client.features.logging.DEFAULT
import io.ktor.client.features.logging.LogLevel
import io.ktor.client.features.logging.Logger
import io.ktor.client.features.logging.Logging
import io.ktor.client.request.post
import io.ktor.util.KtorExperimentalAPI
import kotlinx.coroutines.runBlocking

@KtorExperimentalAPI
fun main() {
    runBlocking {
        val client = HttpClient(Apache) {
            install(Logging) {
                logger = Logger.DEFAULT
                level = LogLevel.ALL
            }
        }
        client.post<String>("http://localhost:8080") {
            body = """{ "test": "testVal" }"""
        }
    }
}
```

**Expected behavior**
Request body is in the log

**Actual result**
```text
[main] INFO io.ktor.client.HttpClient - REQUEST: http://localhost:8080/
[main] INFO io.ktor.client.HttpClient - METHOD: HttpMethod(value=POST)
[main] INFO io.ktor.client.HttpClient - COMMON HEADERS
[main] INFO io.ktor.client.HttpClient - -> Accept-Charset: UTF-8
[main] INFO io.ktor.client.HttpClient - -> Accept: */*
[main] INFO io.ktor.client.HttpClient - CONTENT HEADERS
[main] INFO io.ktor.client.HttpClient - BODY Content-Type: text/plain; charset=UTF-8
[main] INFO io.ktor.client.HttpClient - RESPONSE: 200 
---   remaining response stuff skipped for brevity  ---
```

[...]

PR #1808, opened 2020-04-19
fix log channel is not closed for ByteArrayContent

giylmi:
fix log channel is not closed for ByteArrayContent and
add tests for OutgoingContent types handled in ObservingUtils.kt
replace deprecated call to `writeStringUtf8` with `writeText` in `ByteWriteChannel.kt`

**Subsystem**
Client

**Motivation**
Solving the problem stated in https://github.com/ktorio/ktor/issues/1729

**Solution**
The log channel was not closed for ByteArrayContent. Added closing and added tests for all the types of OutgoingContent that are used in ObservingUtils.kt


eliekarouz:
Hello, any idea when this fix will be released?

PR #1810, opened 2020-04-20
Logging

cy6erGn0m:
**Subsystem**
All

**Motivation**
We need common multi-platform logging replacement for slf

### DIFF
--- a/.idea/dictionaries/cy.xml
+++ b/.idea/dictionaries/cy.xml
@@ -5,6 +5,7 @@
       <w>agentlib</w>
       <w>algo</w>
       <w>api</w>
+      <w>appenders</w>
       <w>attrs</w>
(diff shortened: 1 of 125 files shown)


ColinHebert:
It seems wrong to me to build a complete logging framework within ktor. Shouldn't that be a side library?
On a related note, MicroUtils/kotlin-logging has experimental support of multi-platform ( see https://github.com/MicroUtils/kotlin-logging/wiki/Multiplatform-support ) it might be worth putting effort into that library rather than creating another one?

NahroTo:
https://github.com/AAkira/Napier is also a nice library. Was built with multiplatform in mind from the start.

e5l:
Could we close this PR?

Issue #1748, opened 2020-03-24
When CIO channel closed,   IO exception not catch


xfront:
**Ktor Version and Engine Used (client or server and name)**
ktor 1.3.0 CIO tcp socket on android

**Describe the bug**
When  channel closed,  IO exception not catch

**To Reproduce**
1. android client using cio, connect to server
2. close wifi
3. See error

**Expected behavior**
IO exception.

**Screenshots**

![图片](https://user-images.githubusercontent.com/7844106/77413133-d8cfc200-6df9-11ea-852d-2bf70275aa14.png)

PR #1828, opened 2020-04-27
bugfix for CIO socket

xfront:
**Subsystem**
Clien ,CIO Raw socket

**Motivation**
https://github.com/ktorio/ktor/issues/1748

**Solution**
Let out side class control the socket run scope


ktor-network/jvm/src/io/ktor/network/sockets/NIOSocket.kt
```diff
@@ -28,9 +28,6 @@ internal abstract class NIOSocketImpl<out S>(
 
     override val socketContext: CompletableJob = Job()
```
cy6erGn0m, line 29: Not sure if the job should be outside the jobs hierarchy here. Shouldn't we tie it to selector's job?

Issue #1812, opened 2020-04-21
Ktor with Android client throws IllegalStateException - responseMessage must not be null


Hakky54:
**Ktor Version and Engine Used (client or server and name)**
Version 1.3.1 of Ktor client with Android module

**Describe the bug**
When I do a get request I am getting `Exception in thread "main" java.lang.IllegalStateException: connection.responseMessage must not be null`

**To Reproduce**
* Running with Java 11, with maven, kotlin stdlib and org.jetbrains.kotlin maven plugin.
```
import io.ktor.client.HttpClient
import io.ktor.client.engine.android.Android
import io.ktor.client.request.get
import io.ktor.client.statement.HttpResponse
import io.ktor.utils.io.readUTF8Line
import kotlinx.coroutines.runBlocking

fun main() {

    runBlocking {
        val response: HttpResponse = HttpClient(Android).get("http://api.chucknorris.io/jokes/random")

        print(response.status.value)
        print(response.content.readUTF8Line())
    }

}
```

**Expected behavior**
I would expect it should print the status code and response message, but it doesn't reach at that point.

**Screenshots**
<img width="813" alt="Screenshot 2020-04-21 at 12 53 59" src="https://user-images.githubusercontent.com/16032204/79858163-435f2800-83cf-11ea-8ede-f670a999ad43.png">

Hakky54:
I discovered that the some servers don't always send the reasonPrase. I am using Tomcat and they mentioned it is disabled by default, but can be enabled. However this option is deprecated and the switch will be removed for the next release. I am not quite sure if other servers always do send the reasonPhrase.

AndroidClientEngine requires the reasonPhrase at line number 85 `val statusCode = HttpStatusCode(connection.responseCode, connection.responseMessage)` Would it be possible to make the reasonPhrease optional within the AndroidClientEngine?

Hakky54:
This issue is fixed with the following pull request: https://github.com/ktorio/ktor/pull/1852

rolfwessels:
I'm on Version **1.3.2** of Ktor client and I am still having this issues. The API calls work when I am running in app but under MPP unit testing I get the same exception. 

```groovy
sourceSets["androidTest"].dependencies {
        implementation("io.ktor:ktor-client-android:1.3.2")
       ...
}
```

```kotlin
    @Test
    @Tag("IntegrationTest")
    fun `Failing`() = runBlocking  {

[...]

PR #1852, opened 2020-05-07
bug-fix: enhanced handling of statusCode for AndroidEngine

Hakky54:
**Subsystem**
Client/Android Engine

**Motivation**
This pull request should solve the following issue: https://github.com/ktorio/ktor/issues/1812

**Solution**
Describe your solution.
The underlying http client of AndroidEngine can't extract the reasonPhrase when the server is not sending it. The reason phrase is actually optional, so it could be that some servers are sending and some dont. The http specification has also mentioned that it is optional:
> The reason-phrase element exists for the sole purpose of providing a
textual description associated with the numeric status code, mostly
out of deference to earlier Internet application protocols that were
more frequently used with interactive text clients. A client SHOULD
ignore the reason-phrase content.

See here for more information: https://tools.ietf.org/html/rfc7230#section-3.1.2

By switching over to HttpStatusCode#fromValue method AndroidEngine can create HttpStatusCode from the status code. It won't rely on the underlying http client for getting the reason phrase anymore but it will use the internal contants of HttpStatusCode to find a match based on the status code.

### DIFF
--- a/ktor-client/ktor-client-android/jvm/src/io/ktor/client/engine/android/AndroidClientEngine.kt
+++ b/ktor-client/ktor-client-android/jvm/src/io/ktor/client/engine/android/AndroidClientEngine.kt
@@ -82,7 +82,9 @@ class AndroidClientEngine(override val config: AndroidEngineConfig) : HttpClient
 
         connection.timeoutAwareConnect(data)
 
-        val statusCode = HttpStatusCode.fromValue(connection.responseCode)
+        val responseCode = connection.responseCode
+        val responseMessage = connection.responseMessage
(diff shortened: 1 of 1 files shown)


Hakky54:
@e5l have you had a chance to look at this, anything you want done differently?

e5l:
Hi @Hakky54, thanks for the PR.

I'm not sure if we can always replace the reason-phrase. There are several cases where the server is using it for a purpose.

Could we check for the exact values of the reason phrase before replacing it?

Hakky54:
Glad to hear you liked it!

I understand your concerns. You would only like to replace it when it is actually absent and not when it is present. I refactored it a bit. It will now behave the same as before, but when it is absent it will still create the statusCode from the responseCode. What do you think of the refactored solution?

e5l:
Thanks, that's awesome! Merged

PR #1860, opened 2020-05-13
Fix executor service termination in okhttp

e5l:


ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/WebSocketTest.kt
```diff
@@ -51,19 +51,36 @@ class WebSocketTest : ClientLoader() {
...
+            install(WebSockets)
+        }
+
+        test { client ->
+            client.webSocket("wss://echo.websocket.org") {
+                outgoing.send(Frame.Text("PING"))
+                val frame = incoming.receive()
+                if (frame is Frame.Text) {
+                    assertEquals("PING", frame.readText())
+                }
```
Ololoshechkin, line 82: Is it possible for some reason for frame not to be Frame.Text because of some network problem?
Maybe `assert(frame is Frame.Text && frame.readText() == "PING")`

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/WebSocketTest.kt
```diff
@@ -51,19 +51,36 @@ class WebSocketTest : ClientLoader() {
     fun testCancel() = clientTests(listOf("Apache", "Android", "Js", "iOS")) {
```
Ololoshechkin, line 51: Nit: it feels a bit confusing to me that `clientTests(engines) { ... }` means `clientTests(skipEngines = engines ) { ... }`. I would add explicit parameter name, but feel free to ignore :)

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/WebSocketTest.kt
```diff
@@ -51,19 +51,36 @@ class WebSocketTest : ClientLoader() {
...
+                            delay(250)
                         }
                     }
                 }
             }
         }
     }
+
+    @Test
+    fun testEchoWSS() = clientTests(listOf("Apache", "Android", "Js", "iOS")) {
```
Ololoshechkin, line 71: Please, add comments here describing why this test was failing before the change and is not failing now. How is it connected to the change?

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/WebSocketTest.kt
e5l: It's not possible and we can perform the strict check here

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/WebSocketTest.kt
e5l: It wouldn't help I think. Maybe migration to `vararg` can help, but I'm not sure

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/WebSocketTest.kt
e5l: I'm going to make a separate test for OkHttp. Let's keep this test like `wss` only.

e5l:
Added test for termination. Regarding the test with `wss`, it's copied from the related report.

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/WebSocketTest.kt
Ololoshechkin: I don't know how varargs can help, not sure as well

Issue #1869, opened 2020-05-15
Scheme instead of version in OriginConnectionPoint


mariotrucco:
**Ktor Version and Engine Used (client or server and name)**
1.3.2, server

**Describe the bug**
`ApplicationCall.request.origin.remoteHost` holds the scheme instead of the version (like 'http' instead of `HTTP/1.1`).
The root cause is this [line](https://github.com/ktorio/ktor/blob/master/ktor-server/ktor-server-core/jvm/src/io/ktor/features/OriginConnectionPoint.kt#L59).

I provided a fix here: https://github.com/ktorio/ktor/pull/1871

**To Reproduce**
1. [Configure request logging](https://ktor.io/servers/features/call-logging.html) for a ktor server
2. Install 'XForwardedHeaderSupport`
3. print `call.request.origin.remoteHost`
4. Run the server behind a proxy
5. Send a request to the proxy
6. Observe that `http` is printed

**Expected behavior**
`HTTP/1.1` should be printed

oleg-larshin:
Please check the <a href="https://youtrack.jetbrains.com/issue/KTOR-377">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

PR #1871, opened 2020-05-15
Fix protocol version in OriginConnectionPoint

mariotrucco:
**Subsystem**
Server

**Motivation**
Fixes [1869](https://github.com/ktorio/ktor/issues/1869),
When using a proxy and wanting, for example, access log, we should not have the protocol scheme instead of the version

**Solution**
Replaced faulty return `scheme` with `version`


cy6erGn0m (approved):

cy6erGn0m:
Merged in 319426282aa5a446594a654ea16a89d6fca3c77d

Issue #1874, opened 2020-05-16
Content Negotiation: Gson: Should be able to return 400 for badly formatted request.


lgwillmore:
**Ktor Version and Engine Used 1.3.2**
Additionally using `ktor-gson` plugin.

**Describe the bug**
gson converter returns a 500 status request if the gson is badly formatted. I think it should be a 400.

**To Reproduce**
1. Add content negotiation plugin. Use the gson converter
2. Call your endpoint with badly formatted json
3. Get a 500 status

**Expected behavior**
Should be a 400 status.

gson converter should catch the `JsonSyntaxException` and throw a `BadRequestException`

lgwillmore:
I have used status pages to handle this - so maybe the intent is for status pages to be the way that gson parsing errors are handled?

oleg-larshin:
Please check the <a href="https://youtrack.jetbrains.com/issue/KTOR-373">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

PR #1875, opened 2020-05-16
Add json syntax exception handling to gson content negotiation converter

lgwillmore:
**Subsystem**
Server - Features - ktor-gson

**Motivation**
Make gson converter return a 400 status code for bad json rather than a 500. More detail here: https://github.com/ktorio/ktor/issues/1874

**Solution**
Catch the "JsonSyntaxException" and throw a "BadRequestException"

### DIFF
--- a/ktor-features/ktor-gson/jvm/src/io/ktor/gson/GsonSupport.kt
+++ b/ktor-features/ktor-gson/jvm/src/io/ktor/gson/GsonSupport.kt
@@ -5,7 +5,6 @@
 package io.ktor.gson
 
 import com.google.gson.*
-import com.google.gson.stream.*
 import io.ktor.application.*
 import io.ktor.features.*
 import io.ktor.features.ContentTransformationException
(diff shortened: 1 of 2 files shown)


ktor-features/ktor-gson/jvm/src/io/ktor/gson/GsonSupport.kt
```diff
@@ -40,7 +41,11 @@ class GsonConverter(private val gson: Gson = Gson()) : ContentConverter {
             throw ExcludedTypeGsonException(type)
         }
 
-        return gson.fromJson(reader, type.javaObjectType) ?: throw UnsupportedNullValuesException()
+        return try{
```
mav3n, line 44: Code formatting doesn't seem to be proper, missing spaces around try block.

lgwillmore: fixed

lgwillmore:
Looks like the test and fix was copypastad into master. Closing.

PR #1881, opened 2020-05-20
Kotlin 1.4-M2 & HMPP fixes

h0tk3y:


gradle.properties
```diff
@@ -65,3 +65,16 @@ ws_version=6.2.1
...
+systemProp.org.gradle.internal.publish.checksums.insecure=true
+
+kotlin.mpp.enableGranularSourceSetsMetadata=true
+kotlin.mpp.enableCompatibilityMetadataVariant=true
+
+kotlin.native.disableCompilerDaemon=true
+
+kotlin.native.distribution.type=prebuilt
+
+org.gradle.parallel=false
```
e5l, line 80: It takes lots of time to build and publish without parallel. 
Could we fix this?

ktor-client/ktor-client-features/ktor-client-tracing/ktor-client-tracing-stetho/gradle.properties
```diff
@@ -0,0 +1,6 @@
+#
+# Copyright 2014-2020 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+#
+
+# Workaround for KT-38872
```
e5l, line 5: Please consider mentioning issues consistent: a link or a ticket id.

ktor-http/ktor-http-cio/posix/src/io/ktor/http/cio/websocket/FrameNative.kt
```diff
@@ -26,7 +26,7 @@ actual sealed class Frame actual constructor(
      * (separated into several text frames so they have [fin] = false except the last one).
      * Note that usually there is no need to handle fragments unless you have a RAW web socket session.
      */
-    actual class Binary actual constructor(fin: Boolean, data: ByteArray) : Frame(fin, FrameType.BINARY, data) {
+    actual class Binary actual constructor(fin: Boolean, data: ByteArray) : Frame(fin, FrameType.BINARY, data, NonDisposableHandle) {
```
e5l, line 29: Could you tell me the reason for the explicit argument here?

gradle.properties
h0tk3y: Fixed it, thanks. That was a local change that I should not have committed.

ktor-http/ktor-http-cio/posix/src/io/ktor/http/cio/websocket/FrameNative.kt
h0tk3y: Otherwise, it's an error: no value passed for parameter `disposableHandle`. It seems like some change happened in Kotlin with expect constructor default parameters, so that in the actual class no default arguments are available.

dsavvinov:
Superseded by https://github.com/ktorio/ktor/pull/1905

PR #1888, opened 2020-05-22
Fix doubling host

e5l:
Fix doubling `Host` header


e5l:
Hi @proton72, could you take a look?

proton72:
Looks good =) Many thanks for quick fix!

PR #1896, opened 2020-05-26
Change IosHttpRequestException parent to improve usability

e5l:


e5l:
Hi @Alex009, could you take a look?

Alex009:
hi!
yes, i think it will be fix problem

Issue #1127, opened 2019-05-14
Content-Type in a request headers


mister11:
### Ktor Version
1.1.4

### Ktor Engine Used(client or server and name)
Apache

### JVM Version, Operating System and Relevant Context
1.8, macOS Mojave and Linux Mint, IDEA 2019.1.2 CE

### Feedback

I'm accessing and API that requires me to set `Content-Type` for all requests, but I'm getting `io.ktor.http.UnsafeHeaderException: Header Content-Type is controlled by the engine and cannot be set explicitly`

Current implementation:

```kotlin
val client = HttpClient(Apache) {
        install(JsonFeature) {
            serializer = GsonSerializer()
        }
        install(Logging) {
            level = LogLevel.HEADERS
        }
        defaultRequest {
            header("Content-Type", "application/vnd.api+json")
        }
    }
```

```kotlin
get("/test") {
            client.get<String>("url...") {
                headers {
                    // other headers
                }
            }
        }
```

mister11:
As an additional information, I can solve this issue using `OkHttp` with a network interceptor, but I'm curious how to do the same thing using Apache client

e5l:
Hi @mister11, thanks for the report.
We introduced `acceptContentTypes` in `JsonFeature` since `1.2.0`.  It provides you possibility to set custom `Accept` and `Content-Type` headers automatically.

Could you check and report if it solves your problem?

mister11:
Not sure if I'm doing something wrong, but this is not working:

```kotlin
val client = HttpClient(Apache) {
        install(JsonFeature) {
            serializer = GsonSerializer()
            acceptContentTypes = acceptContentTypes + listOf(ContentType.parse("application/vnd.api+json; ext=bulk"))
        }
    }
```

This just sets `Accept` header and I need `Content-Type` header for requests.

To make it more clear, here is a working version of OkHttp interceptor implementation:

```kotlin
fun provideHeadersInterceptor() = Interceptor { chain ->
    val requestBuilder = chain.request().newBuilder()
        .addHeader("Content-Type", "application/vnd.api+json")

    chain.proceed(requestBuilder.build())
}
```

```kotlin
fun provideOkHttpClient(): HttpClient = HttpClient(OkHttp) {
    engine {
        addNetworkInterceptor(provideHeadersInterceptor())
    }
}
```

sh1nen:
any solutions? I am in the same situation, but for me even interceptior does not seem to work as I would expected and I'm getting
[...]

PR #1897, opened 2020-05-26
Verify sending Content-Type and custom object body via POST

### LINKED ISSUE #997: ClassCastException when POSTing an object without explicitly setting a Content-Type
### Ktor Version

1.1.3

OkHttp with GsonSerializer

### JVM Version, Operating System and Relevant Context

Android 9.0

### Feedback

When I try to do a POST call with a body object, it gives me a ClassCastException:
```
java.lang.ClassCastException: be.gekrabbel.ktortest.MyCustomObject cannot be cast to io.ktor.client.call.HttpClientCall
        at io.ktor.client.HttpClient.execute(HttpClient.kt:151)
        at io.ktor.client.call.HttpClientCallKt.call(HttpClientCall.kt:80)
        at be.gekrabbel.ktortest.MainActivity.callDoesntWork(MainActivity.kt:63)
        at be.gekrabbel.ktortest.MainActivity$onCreate$1.invokeSuspend(MainActivity.kt:41)
```

It works if the body is a String instead of an object, and it also works after explicitly adding `contentType(ContentType.Application.Json)`.

Full sample app: https://github.com/Arne517/ktor-test/blob/master/app/src/main/java/be/gekrabbel/ktortest/MainActivity.kt

e5l:
Close #997, #1127


e5l:
Hi @arnestockmans, @Ribesg, @lamba92, @mister11. The issue was fixed some time ago, could you take a look and confirm that all cases were covered?

e5l:
@1hsyn1, could you also take a look?

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/SerializationTest.kt
```diff
@@ -0,0 +1,53 @@
...
+class SerializationTest : ClientLoader() {
+
+    @Test
+    fun testSendCustomObject() = clientTests {
+        config {
+            install(JsonFeature)
+        }
+
+        test { client ->
+            assertFailsWith<ClientRequestException> {
```
Ololoshechkin, line 27: should the exception code (`response.status`) describe the reason of failure (bad content type in header)? If yes -- maybe add the check for it.

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/SerializationTest.kt
```diff
@@ -0,0 +1,53 @@
...
+                client.post {
+                    url("https://google.com")
+                    header(HttpHeaders.ContentType, ContentType.Application.Json)
+                    body = MyCustomObject(message = "Hello World")
+                }
+            }
+        }
+    }
+
```
Ololoshechkin, line 36: Am i correct that 2/2 tests are checking the bad situation (`assertFailsWith`). Shouldn't we add one more test with no failure?

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/SerializationTest.kt
e5l: Nope. The message should be default in status
The custom message is usually used for custom status codes

Ololoshechkin:
Also it is not clear to me how this PR solves #997. I think it is required to add checks for content type that was never specified but I couldn't find a test for the situation where no Content-Type header was specified at all.

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/SerializationTest.kt
e5l: Will add. Thanks

e5l:
That's true. I'll check that case too.

e5l:
Hey, I fixed the exception message for #997(the behavior is expected, but the message looked confusing) and realized that there are lots of general serialization tests. Could you take a look?

Issue #882, opened 2019-01-16
Can't set FormUrlEncoded header for ktor http client request


wilyarti:
### Ktor Version 
1.1.1

### Ktor Engine Used(client or server and name) 
Apache

### JVM Version, Operating System and Relevant Context 
Ubuntu 18.04
Information:Kotlin: kotlinc-jvm 1.3.11 (JRE 1.8.0_191-8u191-b12-0ubuntu0.18.04.1-b12)

### Feedback
I am trying to POST a request to an API that uses Content-Type: application/x-www-form-urlencoded.

```
val message = client.post<Unit> {
    url(URL("https://tapi.telstra.com/v2/oauth/token"))
    contentType(ContentType.Application.FormUrlEncoded)
    body = getAuth(client_id = "overriddenkey", client_secret = "overridenpass")
}
```
I get the error: io.ktor.http.UnsafeHeaderException: Header Content-Type is controlled by the engine and cannot be set explicitly

I can't figure out how to set the headers in the client constructor. The documentation [https://ktor.io/clients/http-client/calls/requests.html](url) specifies that custom headers can be set using HttpRequestBuilder.

How do I specify the contentType and set the header data? I could not figure it out from your documentation.

tasmalinakis:
Hi, we bumped into this problem a while ago as well. Going through the code, I believe that the current way of doing this is by using `FormDataContent` like so (without specifying a contentType):

` body = FormDataContent(parametersOf(
       "client_id" to "overriddenkey", 
       "client_secret" to "overridenpass") )
`

Last time I checked this was not stated somewhere in the documentation though.

wilyarti:
Thanks that seems to work. Although I still need to also be able to set the body of the message with JSON to communicate with the API.

But at least I am 50% there now.

FrancoSabadini:
I'm having this same issue, I need to add the header and the body, but seems like it's not possible to add both, which makes the API call fail. Has anyone found a workaround for this?

semyannikov:
@e5l I will start changing documentation in few days, any comments related to PR are welcome.

oleg-larshin:
Please check the <a href="https://youtrack.jetbrains.com/issue/KTOR-695">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

woniupapa:
Anyone fixed this problem? help, the same problem.

wilyarti:
[...]

PR #1902, opened 2020-05-27
Issue #882 Can't set FormUrlEncoded header for ktor http client request

semyannikov:
**Subsystem**
Ktor client
Ktor server (more convenient method to set unsafe header).

**Motivation**
issue #882

**Solution**
1. Added processing content-related headers in default transformers feature of `HttpClient`.
2. Added flag `HttpClientEngineConfig. allowUnsafeHeaders` that allows clients forwarding unsafe headers to an engine, by default that ability is disabled. If there is a conflict between a header and `OutgoingContent` object, data from the content object will be preferred. 
This is a controversial decision and, actually, I am not sure it is a right way, but that flag may be useful in situations when user can not affect `OutgoingContent` that does not contain `Content-Type` or when it's hard to do. Another option might be to make some decorator that overrides `Content-Type` from `OutgoingContent` or something.

PS: I disabled engine `CIO` in my tests because it hangs for some reasons even if I remove all my changes. At the moment I am not familiar with that engine and would be grateful if you could help me whit this.

### DIFF
--- a/ktor-client/ktor-client-apache/jvm/src/io/ktor/client/engine/apache/ApacheRequestProducer.kt
+++ b/ktor-client/ktor-client-apache/jvm/src/io/ktor/client/engine/apache/ApacheRequestProducer.kt
@@ -125,19 +125,19 @@ internal class ApacheRequestProducer(
         builder.uri = url.toURI()
 
         val content = requestData.body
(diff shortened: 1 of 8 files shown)


e5l:
Hey, @semyannikov. Thanks for the PR. I'm not sure that we're going to introduce this change in the API: it looks mostly like an ad-hoc solution and requires good design first. Please consider making an issue instead.

PR #1908, opened 2020-05-30
JsonFeature: Fixed header behavior

wkornewald:
**Edit:** The preferred solution is in PR #1927.

**Subsystem**
ktor-client-json

**Motivation**
This fixes several related issues:

*First problem:*
Without this PR it's impossible to define a precise accept header on a per-request basis, but still be able to handle all supported JSON content types in the response.

Some backends/endpoints require exactly one accept header value to select the correct response content type. For instance, this happens when each endpoint is versioned using the accept header (e.g. application/x-vnd.foo.v2+json) or even versioned **and** typed (json v3 and xml v2) and you want to use one single client for all endpoints. While it's possible to configure all accepted content types for the response via `accept()` when installing `JsonFeature`, the caveat is that all of those types would also be added to each request.

In our case, this makes the endpoint respond with a 400 status because it doesn't know which version to choose. You might argue that the endpoint should pick the first best match, but that's out of our control and it makes ktor-client-json impractical to use for us.

In other words, there must be a solution for choosing the accept header per individual request, but still be able to deal with all response types. 

*Second problem:*
ktor-client-json adds accept headers, no matter if the exact same header value already exists. Some backend software doesn't like redundant headers and fails with e.g. a 400 response code. Again, the software is out of our control.

**Solution**
This PR allows to set a new enum config `addAcceptHeader` with values `Always` (the current behavior), `IfAcceptHeaderMissing` (acceptable behavior for us), and `Never` (our most desired behavior).

Also, this PR adds a check, so we never add redundant accept header values.

**Alternative**
This is implemented in the alternative PR #1927. It also adds more flexible Accept/Content-Type handling.

We could go even further and simply never add accept headers to the request. The accept header can already be defined via `defaultRequest`, so adding extra headers in ktor-client-json is redundant, unexpected and causes strange problems like in our case. This solution would be the most efficient, simple, clear, and precise. I just thought you might have reasons for the current behavior, so I wanted to start from a configurable solution that contains all options for you to test.

### DIFF
--- a/ktor-client/ktor-client-features/ktor-client-json/common/src/io/ktor/client/features/json/JsonFeature.kt
+++ b/ktor-client/ktor-client-features/ktor-client-json/common/src/io/ktor/client/features/json/JsonFeature.kt
@@ -25,6 +25,25 @@ import io.ktor.utils.io.*
 expect fun defaultSerializer(): JsonSerializer
 
 /**
+ * Configuration values for [JsonFeature.Config.addAcceptHeader], defining when accept header is
+ * added to requests.
+ */
+enum class AddAcceptHeader {
+    /** Always adds accept header(s) even if the request already has such a header. */
+    Always,
+
+    /**
+     * Only adds accept header(s) if none exists in the request.
+     *
+     * This can be used to make sure that only one accept header is ever sent to the server.
+     */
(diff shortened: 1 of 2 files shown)


wkornewald:
Closing in favor of #1927

PR #1916, opened 2020-06-01
Fix cinterop publishing

dsavvinov:
https://youtrack.jetbrains.com/issue/KT-39302

### DIFF
--- a/gradle/posix.gradle
+++ b/gradle/posix.gradle
@@ -57,7 +57,7 @@ kotlin {
                 project.ext.nativeTargets = [iosX64, iosArm64, iosArm32, macosX64, linuxX64, mingwX64, tvosArm64, tvosX64, watchosArm32, watchosArm64, watchosX86]
             }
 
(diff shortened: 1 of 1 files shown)


gradle/posix.gradle
```diff
@@ -104,3 +104,21 @@ kotlin {
...
 project.ext.set("hasNative", true)
+
+// hack for 1.4-M2, remove after
+// see https://youtrack.jetbrains.com/issue/KT-39302
+afterEvaluate {
+    def metadataTarget = kotlin.targets.getByName("metadata")
+
+    def currentHostKonanTarget = hostManager.Companion.getHost()
+
+    if (metadataTarget != null) {
```
h0tk3y, line 115: Unlike `findByName`, `getByName` never returns null, it throws an exception if the element is absent. However, in Kotlin MPP modules that have `kotlin.targets`, there's always a target under the name `metadata`. So it seems that the null check is redundant.

PR #1919, opened 2020-06-02
Apache should use existing approach when merging headers

semyannikov:
**Subsystem**
Ktor client - Apache engine

**Motivation**
We should use existing approach when merging content-related headers and common ones, it will make engines safely replaceable.

**Solution**
Content-Type and Content-Length of OutgoingContent will be preferred now.


e5l:
LGTM

PR #1926, opened 2020-06-04
Fixes #1345

### LINKED ISSUE #1345: Useless Iterable.reversed calls in HttpCallValidator, the result is not used
1.2.4, 1.3.0-beta-1, and probably older versions.

**Describe the bug**
The HttpCallValidator contains two calls to reversed but the result is ignored.
https://github.com/ktorio/ktor/blob/4cc95c26d8f11cece38beb676f7ccab037c4941e/ktor-client/ktor-client-core/common/src/io/ktor/client/features/HttpCallValidator.kt#L72-L73
This makes these calls completely useless.

**Expected behavior**
The result should probably be used when calling the HttpCallValidator constructor in line 75-78. As an alternative if a list copy is not correct behaviour, `MutableList.reverse()` could be used.

alllex:
**Subsystem**
Client

**Motivation**
Fixes #1345

**Solution**
Reverse the order of validators and exception handlers as the documentation to this feature claims.

### DIFF
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/features/HttpCallValidator.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/features/HttpCallValidator.kt
@@ -71,12 +71,9 @@ class HttpCallValidator(
         override fun prepare(block: Config.() -> Unit): HttpCallValidator {
             val config = Config().apply(block)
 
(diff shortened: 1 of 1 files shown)


e5l:
Hi @alllex, thanks for the PR. LGTM, Merged.

PR #1927, opened 2020-06-04
JsonFeature: Fixed header behavior and made it more flexible

wkornewald:
**Subsystem**
ktor-client-json

**Motivation**
This fixes several related issues:

*First problem:*
Without this PR it's impossible to define a precise accept header on a per-request basis, but still be able to handle all supported JSON content types in the response.

Some backends/endpoints require exactly one accept header value to select the correct response content type. For instance, this happens when each endpoint is versioned using the accept header (e.g. application/x-vnd.foo.v2+json) or even versioned **and** typed (json v3 and xml v2) and you want to use one single client for all endpoints. While it's possible to configure all accepted content types for the response via `accept()` when installing `JsonFeature`, the caveat is that all of those types would also be added to each request.

In our case, this makes the endpoint respond with a 400 status because it doesn't know which version to choose. You might argue that the endpoint should pick the first best match, but that's out of our control and it makes ktor-client-json impractical to use for us.

In other words, there must be a solution for choosing the accept header per individual request, but still be able to deal with all response types. 

*Second problem:*
ktor-client-json adds accept headers, no matter if the exact same header value already exists. Some backend software doesn't like redundant headers and fails with e.g. a 400 response code. Again, the software is out of our control.

*Third problem*:
ktor-client-json detects whether to serialize/deserialize JSON for requests and responses based on the Content-Type header, but you can only define `ContentType` objects with rather limited matching rules. We need more complex content type patterns than just `application/*`. We need e.g. `application/*+json` because we have to deal with many different Content-Type / Accept variants.

**Solution**
Simply never add accept headers to the request. The accept header can already be defined via `defaultRequest`, so adding extra headers in ktor-client-json is redundant, unexpected and causes strange problems like in our case. This solution is efficient, simple, clear, and precise.

Also, this allows defining an arbitrary matching function for content types (with a sensible default) that simplifies triggering the JsonFeature serialization.

However, this might only be the first step. What if the backend doesn't set the Content-Type header because it's broken? We must be able to enforce JSON handling per individual request and/or response. I haven't solved that aspect, yet.

This is the improved and simplified solution based on PR #1908.


wkornewald:
@e5l Could you please have a look? This was a real show-stopper for us and after having tried out all options I think explicit is better than implicit.

ktor-client/ktor-client-features/ktor-client-json/common/src/io/ktor/client/features/json/JsonFeature.kt
```diff
@@ -79,13 +81,21 @@ class JsonFeature internal constructor(
...
         fun accept(vararg contentTypes: ContentType) {
             _acceptContentTypes += contentTypes
         }
+
+        /**
+         * Defines when to handle a given [ContentType] (used in addition to [accept]).
+         *
+         * By default accepts any `application/<...>+json` pattern.
+         */
+        var shouldHandle: (ContentType) -> Boolean = {
```
e5l, line 95: It's easy to break source compatibility if we have `public` functional property.
Let's make it simple: instead of introducing the public field, let's make a config function:
```
fun filterContentType(block: (ContentType) -> Boolean {
 ...
}
```
And make the field at internal or private. I think the `filter` is the proper word here because it's naturally what we're doing.

ktor-client/ktor-client-features/ktor-client-json/common/src/io/ktor/client/features/json/JsonFeature.kt
```diff
@@ -79,13 +81,21 @@ class JsonFeature internal constructor(
...
+            val value = it.toString()
+            value.startsWith("application/") && value.endsWith("+json")
```
e5l, line 97: Consider using the `ContentType` class and `match` method. I also would keep the old default, but if you have any reasons we can discuss it

ktor-client/ktor-client-features/ktor-client-json/common/src/io/ktor/client/features/json/JsonFeature.kt
e5l: And we need at least one test using that API

ktor-client/ktor-client-features/ktor-client-json/common/src/io/ktor/client/features/json/JsonFeature.kt
```diff
@@ -57,7 +57,10 @@ class JsonFeature internal constructor(
          *
          * Default value for [serializer] is [defaultSerializer].
          */
-        var serializer: JsonSerializer? = null
+        var serializer: JsonSerializer
```
e5l, line 60: Please consider using `lazy` here

e5l: Or we can move the default in the constructor parameters

e5l:
Hi @wkornewald.
Thanks for the PR, and sorry for the delay :(. That's a good idea, and I will merge the PR after small fixes.

ktor-client/ktor-client-features/ktor-client-json/common/src/io/ktor/client/features/json/JsonFeature.kt
wkornewald: The `ContentType` class doesn't support matching on arbitrary patterns like `application/*+json`. The HTTP spec doesn't seem to allow that, either, so I think `ContentType` shouldn't be extended for this use-case.

The only alternatives for matching rules are regexes and lambda functions. I think functions are the most flexible and can still use a regex under the hood very easily.

ktor-client/ktor-client-features/ktor-client-json/common/src/io/ktor/client/features/json/JsonFeature.kt
wkornewald: Yeah, that solution wasn't ideal. I'd like to keep the lazy behavior, but `lazy` doesn't support `var`, so I'll add a `lazyVar` to ktor-utils.

ktor-client/ktor-client-features/ktor-client-json/common/src/io/ktor/client/features/json/JsonFeature.kt
wkornewald: The advantage of a field is that it's composable. You can do:

```kotlin
httpClient.config {
    install(JsonFeature) {
        val prevFilter = filterContentType
        filterContentType = { prevFilter(it) || myOtherFilter(it) }
    }
}
```

With a function you lose that property unless you introduce a complicated API.

I do see your argument that a function can be extended with additional arguments in the future, but I believe that composability is more important in this case and also that a simple field is no different from inheritance/polymorphism where you override a single method. With inheritance you have the sample problem that changes to the function signature are breaking changes.

wkornewald:
Hi @e5l now I've rebased on master and updated the implementation and unit tests. I hope you agree with the reasons why I've designed the API like that.

What I don't quite understand is that in the `responsePipeline` the `body` is of type `String` during the `testStringWithJsonFeature` unit test. In all other cases it's correct, again. I've just added a check for `String`, but maybe you can verify if that's the correct solution?

ktor-client/ktor-client-features/ktor-client-json/common/src/io/ktor/client/features/json/JsonFeature.kt
wkornewald: By the way, this enables both "application/json" and "application/...+json" by default. Both content types are very common. That's why I think this default would make sense.

wkornewald:
Maybe I should add that I was trying to not fundamentally break the existing API. Otherwise, we could, for example, introduce a `ContenTypeMatcher` interface providing a `match(ContentType)` method. `ContentType` could implement that interface. Then, instead of `Config.acceptContentTypes` being a `List<ContentType>` it would become a `List<ContentTypeMatcher>`. This would allow implementing arbitrary matching rules like `application/*+json`.

ktor-client/ktor-client-features/ktor-client-json/common/src/io/ktor/client/features/json/JsonFeature.kt
e5l: That's a good use-case! So let's fit it in our usual ktor model: we can make the field public with `private` set, and introduce method as I mentioned before.

So we will get a nice builder-like API with the ability to reuse previously configured method!

ktor-client/ktor-client-features/ktor-client-json/common/src/io/ktor/client/features/json/JsonFeature.kt
e5l: Could we fit it in the `ContentType.match` method?

e5l:
Hi @wkornewald, the most considerations are right, and I agree with you. I think that idea of introducing `ContentTypeMatcher` is beautiful for ktor `1.4`, so if you're interested, you can proceed further with it :)

e5l:
Btw. Feel free to use the `Re-request review` button here if you need the review. 

I think all remaining problems are minor, but we definitely should fix the `public` setter before a merge

ktor-client/ktor-client-features/ktor-client-json/common/src/io/ktor/client/features/json/JsonFeature.kt
wkornewald: This is resolved with the ContentTypeMatcher solution.

ktor-client/ktor-client-features/ktor-client-json/common/src/io/ktor/client/features/json/JsonFeature.kt
wkornewald: The ContentTypeMatcher interface allows this now.

wkornewald:
I've added the ContentTypeMatcher interface, but this has the consequence that `ContentType.match()` is inverted now. Previously, the parameter was a pattern. Now, `this` is the pattern (like a Regex) and `match()` receives a non-pattern argument (like `Regex.matches()`).

e5l:
Hey, some tests are failed on the CI. Could you check it?(You can run `cleanJvmTest`, `jvmTest`, `jvmTestClasses`, `jvmMainClasses` to avoid a full build)

wkornewald:
I've indeed missed a few places and I hadn't checked the whole test suite. Unfortunately, I get lots of socket timeouts when running `jvmTest` (even on master), so it's difficult to check locally. Can you please run the test suite on the CI again? Also, can you please send me the report in case of any remaining errors?

e5l:
It almost passes.
The last thing should be done here: `clean apiDump` task. And commit the API changes in `Update public API` commit, so we will review the API and provide migration if it's possible.

wkornewald:
@e5l Finally found the time to finish this. Can you please check?

e5l:
Sure. Will merge after CI run

wkornewald:
Ok, I'm at a loss since I can't run the unit test suite without lots of timeouts. :( Maybe I need to run it in Docker to reproduce your CI environment more closely. Or can you just tell me which tests are broken?

e5l:
Merged :)  Great work! Thanks for the PR

PR #1928, opened 2020-06-05
Apache decoder fix

e5l:


ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/DownloadTest.kt
```diff
@@ -34,8 +33,20 @@ class DownloadTest : ClientLoader() {
...
             assertTrue(size > 0)
         }
     }
+
+    @Test
+    fun testDownload8175() = clientTests {
+        test { client ->
+            repeat(100) {
+                val uri =
+                    "https://raw.githubusercontent.com/ssp/ktor/4f1986df68e3594714ea12949c8af8274be99d01/ktor-client/ktor-client-apache/jvm/test/io/ktor/client/engine/apache/8175.txt"
```
dtretyakov, line 47: We should store this test file in official ktor/ktor repository.

ktor-io/jvm/src/io/ktor/utils/io/ByteBufferChannel.kt
```diff
@@ -1552,9 +1558,11 @@ internal open class ByteBufferChannel(
         }
     }
 
-    override suspend fun write(min: Int, block: (ByteBuffer) -> Unit) {
+    override fun writeAvailable(min: Int, block: (ByteBuffer) -> Unit): Int {
         require(min > 0) { "min should be positive" }
+        require(min <= 4088) { "min should be positive" }
```
dtretyakov, line 1563: The error message should reflect the number `4088`.

dtretyakov: Maybe we need to introduce a constant for this number?

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/DownloadTest.kt
e5l: Sure. I will replace it with the generated data of specific length.

ktor-io/jvm/src/io/ktor/utils/io/ByteBufferChannel.kt
e5l: Sure, will do

PR #1937, opened 2020-06-09
Make JVM MetricSet registration optional

marshallpierce:
**Subsystem**
ktor-metrics

**Motivation**
I was unpleasantly surprised to see that the `DropwizardMetrics` feature registered various MetricSets from `metrics-jvm`. Having them added for me in the Ktor feature means I cannot control the naming or configuration of those metrics, nor can I opt out entirely. It is easy to add those MetricSets outside of Ktor, so it's not like it's saving the user some huge burden by bundling that here. Ultimately, I don't think it's appropriate for Ktor to register metrics that aren't at all related to the core problem of measuring _Ktor specific_ things. 

**Solution**
Add a boolean to the feature's config that makes it possible to opt out. The default setting preserves the old behavior.


e5l (approved):

e5l:
Hey @marshallpierce, lgtm. Thanks for the PR. Could you rebase it on `1.6.0-eap` and run `./gradlew apiDump`?

marshallpierce:
@e5l done

e5l:
Thanks for the PR, merged!

PR #1938, opened 2020-06-10
Update ktor to kotlin 1.4.0

e5l:


gradle/interop-as-source-set-klib.gradle
```diff
@@ -0,0 +1,60 @@
...
+ */
+
+project.ext.registerInteropAsSourceSetOutput = { interopName, sourceSet ->
+    afterEvaluate {
+        def currentHostKotlinTarget = kotlin.targets.matching { it.platformType.name == "native" }.find {
+            it.konanTarget == hostManager.Companion.getHost()
+        }
+
+        def interop = currentHostKotlinTarget.compilations["main"].cinterops[interopName]
+        def cinteropTask = tasks.named(interop.interopProcessingTaskName)
```
Ololoshechkin, line 12: What if `currentHostKotlinTarget` was not found (equals `null`) and hence `interop` also equals `null`.
What would be the result of `tasks.named(null)` in this case?

gradle/interop-as-source-set-klib.gradle
```diff
@@ -0,0 +1,60 @@
...
+
+        def prepareKlibTaskProvider = tasks.register("prepare${sourceSet.name.capitalize()}InteropKlib", Sync) {
+            from(files(zipTree(cinteropKlib).matching {
+                exclude("targets/**", "default/targets/**")
+            }).builtBy(cinteropTask))
+
+            into(tempDir)
+
+            doLast {
+                def manifest140 = file("$tempDir/default/manifest")
```
Ololoshechkin, line 27: Why such numbers? (140, 1371)

Ololoshechkin: PS: oh, I see it must stand for 1.4.0, 1.3.71, right?
Maybe `1_4_0` and `1_3_71` then?

gradle/interop-as-source-set-klib.gradle
```diff
@@ -0,0 +1,60 @@
...
+                def manifest1371 = file("$tempDir/manifest")
+                def manifest = manifest140.exists() ? manifest140 : manifest1371
```
Ololoshechkin, line 29: What if neither of these two files exist? Maybe add descriptive log output here, like "manifest not found"

gradle/interop-as-source-set-klib.gradle
```diff
@@ -0,0 +1,60 @@
...
+
+                def lines = manifest.readLines()
+                def modifiedLines = lines.collect { line ->
+                    line.startsWith("depends=") ? "depends=stdlib ${manifest == manifest140 ? 'org.jetbrains.kotlin.native.platform.posix' : 'posix'}" :
+                        line.startsWith("native_targets=") ? "native_targets=" :
+                            line
+                }
+                manifest.text = modifiedLines.join("\n")
```
Ololoshechkin, line 37: Maybe
`manifest
`.readLines()`
`.collect { ... }`
`.join("\n")`

gradle/interop-as-source-set-klib.gradle
```diff
@@ -0,0 +1,60 @@
...
+            archiveFileName.set("${project.name}_${fakeCinteropCompilation.name}.klib")
+            dependsOn cinteropTask
+        }
+
+        fakeCinteropCompilation.output.classesDirs.from(files().builtBy(copyCinteropTaskProvider))
+
+        kotlin.sourceSets.matching {
+            def visited = new HashSet()
+            def visit
+            visit = { s -> if (visited.add(s)) s.dependsOn.each { visit(it) } }
```
Ololoshechkin, line 53: Is there a chance for `visited.add(s)` to be `false`? I think if it's `false` then it means we have cyclic dependencies between source sets. Is it a valid situation? If no -- maybe report it via some meaningful message?

gradle/js.gradle
```diff
@@ -45,10 +45,6 @@ kotlin {
     sourceSets {
         jsMain.dependencies {
             api "org.jetbrains.kotlin:kotlin-stdlib-js:$kotlin_version"
-
-            api group: 'org.jetbrains.kotlinx', name: 'kotlinx-coroutines-core-js', version: coroutines_version
```
Ololoshechkin, line 49: Don't we need to change it to `org.jetbrains.kotlinx:kotlinx-coroutines-core:$coroutines_version` or is it configured somewhere in a common module?

gradle/jvm.gradle
```diff
@@ -63,8 +62,6 @@ kotlin {
             api group: 'org.jetbrains.kotlin', name: 'kotlin-test-junit', version: kotlin_version
             api group: 'junit', name: 'junit', version: junit_version
 
-            api "org.jetbrains.kotlinx:atomicfu:$atomicfu_version"
```
Ololoshechkin, line 66: ditto

gradle/posix.gradle
```diff
@@ -72,7 +72,7 @@ kotlin {
             }
             disableCompilation(crossCompileTargets)
             configure([mingwX64Test]) {
-                environment "PATH": "C:\\msys64\\mingw64\\bin;C:\\Tools\\msys64\\mingw64\\bin"
+                environment "PATH": "C:\\msys64\\mingw64\\bin;C:\\Tools\\msys64\\mingw64\\bin;C:\\Tools\\msys2\\mingw64\\bin"
```
Ololoshechkin, line 75: Is there a way to find its location without hardcoding directories?

gradle/posix.gradle
```diff
@@ -104,3 +104,21 @@ kotlin {
 }
 
 project.ext.set("hasNative", true)
+
+// hack for 1.4-M2, remove after
+// see https://youtrack.jetbrains.com/issue/KT-39302
```
Ololoshechkin, line 109: Please, add a TODO here to remove this after M3 since it's fixed there according to the issue link

gradle/publish.gradle
```diff
@@ -100,20 +99,39 @@ publishing {
...
+}
 
-    // Disable gradle metadata in root jvm modules
-    def hasNative = project.ext.has("hasNative") && project.ext.hasNative
-    if (!hasNative) {
-        def rootPublication = publishing.publications.findByName('kotlinMultiplatform')
-        rootPublication.moduleDescriptorGenerator = null
+publish.dependsOn publishToMavenLocal
+
+class PublishWithRetryTask extends DefaultTask {
```
Ololoshechkin, line 111: Why do we need retries now? :)

ktor-client/ktor-client-cio/jvm/test/io/ktor/client/engine/cio/CIOHttpsTest.kt
```diff
@@ -140,6 +141,7 @@ class CIOHttpsTest : TestWithKtor() {
     }
 
     @Test
+    @Ignore
```
Ololoshechkin, line 144: Why the test is ignored? Maybe leave a comment?

ktor-client/ktor-client-curl/build.gradle.kts
```diff
@@ -49,7 +49,7 @@ kotlin {
                     afterEvaluate {
                         if (platform.name == "mingwX64") {
                             val winTests = tasks.getByName("mingwX64Test") as KotlinNativeTest
-                            winTests.environment("PATH", "c:\\msys64\\mingw64\\bin;c:\\tools\\msys64\\mingw64\\bin")
+                            winTests.environment("PATH", "c:\\msys64\\mingw64\\bin;c:\\tools\\msys64\\mingw64\\bin;C:\\Tools\\msys2\\mingw64\\bin")
                         }
```
Ololoshechkin, line 53: ditto

ktor-client/ktor-client-curl/build.gradle.kts
```diff
@@ -70,8 +70,18 @@ kotlin {
             dependencies {
                 api(project(":ktor-client:ktor-client-features:ktor-client-logging"))
                 api(project(":ktor-client:ktor-client-features:ktor-client-json"))
-                api("org.jetbrains.kotlinx:kotlinx-serialization-runtime-native:$serialization_version")
             }
         }
+
+        // Hack: register the Native interop klibs as outputs of Kotlin source sets:
```
Ololoshechkin, line 76: Please, add more description here with the reason of this hack

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/HttpRedirectTest.kt
```diff
@@ -60,6 +60,7 @@ class HttpRedirectTest : ClientLoader() {
     }
 
     @Test
+    @Ignore
```
Ololoshechkin, line 63: didto

settings.gradle
```diff
@@ -54,7 +54,7 @@ include ':ktor-client:ktor-client-features:ktor-client-logging'
 include ':ktor-client:ktor-client-features:ktor-client-encoding'
 include ':ktor-client:ktor-client-features:ktor-client-websockets'
 //include ':ktor-client:ktor-client-features:ktor-client-tracing'
-//include ':ktor-client:ktor-client-featurlass'
+//include ':ktor-client:ktor-client-features:ktor-client-tracing:ktor-client-tracing-stetho'
```
Ololoshechkin, line 57: still commented out?

gradle/js.gradle
e5l: It's in common

gradle/posix.gradle
e5l: Nope. It's from our experience :(

gradle/posix.gradle
e5l: The issue is fixed, I try to remove.

gradle/publish.gradle
e5l: Because of the Bintray timeouts

ktor-client/ktor-client-cio/jvm/test/io/ktor/client/engine/cio/CIOHttpsTest.kt
e5l: Because of the availability of remote resources. I'm going to replace that test.

gradle/interop-as-source-set-klib.gradle
e5l: It's a workaround from `MPP` team and I bet that it was copy-pasted from the other projects. I will not change this file to avoid merge conflicts in future.

Ololoshechkin: Ok. Maybe at least add some report message in case of something is null or not found?

gradle/posix.gradle
Ololoshechkin: Ok. 
Then I have 2 concerns here:
1. Shouldn't we add a property or env variable that users can specify in case they have non-default path in their system. Something like: `System.findProperty(MINGW64_PATH) ?: DEFAULT_MINGW64_PATH`
2. Nit: There are at least 2 places with that string. Maybe move to some const?

nikolaykasyanov:
👋 great to see the progress!

I believe this can be uncommented now: https://github.com/ktorio/ktor/blob/0b86cd8f6b933facd473698e873c3d86e9afde2d/ktor-client/ktor-client-ios/darwin/src/io/ktor/client/engine/ios/IosClientEngine.kt#L65-L67

e5l:
Thanks for poiting out! Done

PR #1939, opened 2020-06-10
Fix saved call early completion

### LINKED ISSUE #1780: Ktor client throws JobCancellationException: Parent job is Completed when using compression and logging features
io.ktor:ktor-client-core-jvm:1.3.2
io.ktor:ktor-client-json-jvm:1.3.2
io.ktor:ktor-client-apache:1.3.1"
io.ktor:ktor-client-encoding-jvm:1.3.2
io.ktor:ktor-client-features:1.3.2
io.ktor:ktor-client-serialization-jvm:1.3.2
io.ktor:ktor-client-auth-jvm:1.3.2
io.ktor:ktor-client-logging-jvm:1.3.2

**Describe the bug**
When using the gzip compression feature together 
with the logging feature (with `LogLevel.ALL`) the client fails with the 
`kotlinx.coroutines.JobCancellationException: Parent job is Completed` exception.

**To Reproduce**
Get [ktorGzipTest](https://github.com/pecanw/ktorGzipTest) repository and run `log_all_with_gzip` test. See also readme in the repo for more details.

**Expected behavior**
The httpClient call ends successfully even when using gzip together with logging all details.

e5l:
Close #1780

### DIFF
--- a/ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/HttpStatementTest.kt
+++ b/ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/HttpStatementTest.kt
@@ -5,10 +5,12 @@
 package io.ktor.client.tests
 
 import io.ktor.client.call.*
+import io.ktor.client.features.compression.*
 import io.ktor.client.request.*
 import io.ktor.client.statement.*
 import io.ktor.client.tests.utils.*
 import io.ktor.utils.io.core.*
+import kotlinx.coroutines.*
 import kotlin.test.*
 
 class HttpStatementTest : ClientLoader() {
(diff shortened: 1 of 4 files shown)


ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/HttpStatementTest.kt
```diff
@@ -32,4 +34,22 @@ class HttpStatementTest : ClientLoader() {
...
+    @Test
+    fun testGZipFromSavedResponse() = clientTests {
+        config {
+            ContentEncoding {
+                gzip()
+            }
+        }
+
+        test { client ->
+            val response = client.get<HttpResponse>("$TEST_SERVER/compression/gzip")
```
Ololoshechkin, line 47: Why is it guaranteed here that the result of `get` will be always a `SavedHttpResponse` but not just `HttpResponse`?

e5l: Because we can get the simple `HttpResponse` only when executing `HttpStatement` with `block`. Otherwise, it will be `SavedHttpResponse`

PR #1949, opened 2020-06-16
Fix empty multipart body for client

e5l:


e5l:
Hey @Agcom, could you take a look?

Agcom:
@e5l 
I mentioned (in the issue) that the problem is not related to any particular client.
You can also use curl to reproduce the same behavior.
```
curl -X POST localhost:8080 --header "content-type:multipart/formdata; boundary=-----"
```
The changes are just pointing to the Ktor client core. I don't think this solves the issue.

e5l:
It solves the multipart body in client at least. Let's fix it step by step.

ktor-client/ktor-client-tests/jvm/src/io/ktor/client/tests/utils/TestServer.kt
```diff
@@ -22,7 +21,7 @@ internal fun startServer(): Closeable {
 
     val proxyServer = TestTcpServer(HTTP_PROXY_PORT, ::proxyHandler)
 
-    val server = embeddedServer(Netty, DEFAULT_PORT) {
+    val server = embeddedServer(CIO, DEFAULT_PORT) {
```
Ololoshechkin, line 24: Why do you change to CIO everywhere in tests? How is it related to the fix of empty multipart body?

e5l: This test strikes only with `CIO`. `Netty` ignores wrong prepared request

ktor-server/ktor-server-test-host/build.gradle
```diff
@@ -26,8 +26,6 @@ kotlin.sourceSets.jvmMain.dependencies {
     implementation "org.jetbrains.kotlinx:kotlinx-coroutines-debug:$coroutines_version"
 }
 
-if (!project.ext.ideaActive) {
```
Ololoshechkin, line 29: Why do we need ideaActive at all?

Ololoshechkin: or why did we need*

e5l: It was a bug in old kotlin plugin, the didn't support custom Gradle configurations

PR #1951, opened 2020-06-17
Add locale to ThymeleafContent

AnonymousWalker:
#### Subsystem
ktor-thymeleaf
#### Motivation
In order to support i18n in templates, Thymeleaf provides accessors to resource bundles based on a set locale. When this feature's process method is called a new Context object is created with an empty constructor ```(Thymeleaf.kt:84)```, which then initializes a locale property to the default locale of the current jvm instance. Ktor provides ```call.request.acceptedLanguageItems()``` to access the ```accepted-languages``` field of the http header, and it would be helpful to use a locale based on these requested languages for template rendering.
#### Solution
As the ThymeleafContent object is the public way to provide content from Ktor to Thymeleaf using this feature, an optional locale parameter is added to the ThymeleafContent constructor. If a locale is provided, the Thymeleaf context object is constructed with it. If a locale is not provided, the old behavior is preserved by using a default value of ```Locale.getDefault()```. Previously omitting a locale in the context constructor would initialize a locale with ```Locale.getDefault()``` in the context's super class ```(org.thymeleaf.context.AbstractContext:62)```.
#### Routing
```
get("/demo") {
 val locale = Locale("es") // default locale is English
 call.respond(
 ThymeleafContent(
 template = "demo",
 model = mapOf(),
 locale = locale
 )
 )
}
```
#### demo.html
```
<!DOCTYPE html>
<html xmlns:th="http://www.thymeleaf.org">
<head><meta charset="UTF-8"></head>
 <body>
 <h2 th:text="#{greeting}"></h2>
 </body>
</html>
```
#### demo_en.properties
```
greeting=Hello, World!
```
#### demo_es.properties
```
greeting=¡Hola, Mundo!
```
#### Result
```
<!DOCTYPE html>
<html>
<head><meta charset="UTF-8"></head>
 <body>
 <h2>¡Hola, Mundo!</h2>
 </body>
</html>
```

**Subsystem**

**Motivation**

**Solution**
Describe your solution.

### DIFF
--- a/ktor-features/ktor-thymeleaf/jvm/src/io/ktor/thymeleaf/Thymeleaf.kt
+++ b/ktor-features/ktor-thymeleaf/jvm/src/io/ktor/thymeleaf/Thymeleaf.kt
@@ -18,6 +18,7 @@ import io.ktor.util.cio.bufferedWriter
 import io.ktor.utils.io.ByteWriteChannel
 import org.thymeleaf.TemplateEngine
 import org.thymeleaf.context.Context
+import java.util.Locale
 
 /**
  * Represents a content handled by [Thymeleaf] feature.
@@ -26,12 +27,14 @@ import org.thymeleaf.context.Context
  * @param model to be passed during template rendering
  * @param etag value for `E-Tag` header (optional)
  * @param contentType of response (optional, `text/html` with UTF-8 character encoding by default)
+ * @param locale to be used with template rendering (optional, Locale.getDefault() by default)
  */
 class ThymeleafContent(
     val template: String,
     val model: Map<String, Any>,
     val etag: String? = null,
-    val contentType: ContentType = ContentType.Text.Html.withCharset(Charsets.UTF_8)
+    val contentType: ContentType = ContentType.Text.Html.withCharset(Charsets.UTF_8),
+    val locale: Locale = Locale.getDefault()
 )
 
 /**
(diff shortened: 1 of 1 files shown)


jsarabia:
@e5l Thanks for your quick response!

I've added a test on behalf of @AnonymousWalker. It tests that the template rendering matches the locale requested, as well as a case where no locale is provided and the default is used.

If there are any other tests you'd like here, or any other edge cases you can think of, I am happy to add them!

Thank you!

AnonymousWalker:
@e5l Can you check on this and let me know if it needs anything else? Thanks!

twiese99:
Hello, I would be very happy if a localisation of Thymeleaf templates in Ktor would be possible soon. Is there a way to make this possible in a timely manner?

Regarding the changes: If the ThymeleafContent class is already adapted, shouldn't the respondTemplate extension also be adapted accordingly?

AnonymousWalker:
We've forked this PR and used for a while in our org. Working fine
https://github.com/Bible-Translation-Tools/Fetcher/blob/audio.bibleineverylanguage.org/fetcher-web/src/main/kotlin/org/bibletranslationtools/fetcher/web/controllers/HomeController.kt#L23

e5l (approved):
Hey @AnonymousWalker, thanks for the PR. LGTM
Can be merged after rebase

PR #1952, opened 2020-06-17
Fixed serialization of empty body

wkornewald:
**Subsystem**
Client: JsonFeature/JsonSerializer

**Motivation**
When using `JsonFeature` the empty body (e.g. a `get()` request without a body) gets serialized to `{}` instead of an empty string. This causes problems with some backends which expect the body to be empty.

For more context, let's assume we define an HttpClient with JsonFeature and a default `Content-Type` header like this:

```
val client = HttpClient(...) {
    defaultRequest {
        contentType(ContentType.Application.Json)
    }
    install(JsonFeature) {
        // ...
    }
}
```

This gives us some more convenience because we don't need to add the Content-Type header to each `post()`/`put()` call. However, now every request where we don't assign a body (e.g. a simple `get("https://...")`) gets turned into a request containing `{}` in the body and having a `Content-Type: application/json` header.

This causes some backends to return an error because they expect an empty body for `get()` requests. Often, the backend isn't under our control, so we have to comply with the usual expection: No body in `get()` should mean no body in the resulting HTTP request.

**Solution**
With this PR, `JsonFeature` simply passes through `EmptyContent`. I think this really is the most sensible behavior, mapping the intent 1:1. It also removes the `Content-Type` header because there really is no content.

If someone wants to send an empty `{}` then that's still possible by explicitly setting e.g. `body = Unit` (or any serializable object without attributes). At least, now it's explicit.

If someone wants to send an empty body, but with `Content-Type` set (for obscure backend requirements), that's also still possible by customizing the serializer and using a special object as a marker.

### DIFF
--- a/ktor-client/ktor-client-features/ktor-client-json/common/src/io/ktor/client/features/json/JsonFeature.kt
+++ b/ktor-client/ktor-client-features/ktor-client-json/common/src/io/ktor/client/features/json/JsonFeature.kt
@@ -10,6 +10,7 @@ import io.ktor.client.request.*
 import io.ktor.client.statement.*
 import io.ktor.client.utils.*
 import io.ktor.http.*
+import io.ktor.http.content.*
 import io.ktor.util.*
 import io.ktor.utils.io.*
 import io.ktor.utils.io.core.*
@@ -109,7 +110,7 @@ class JsonFeature internal constructor(val config: Config) {
                 context.headers.remove(HttpHeaders.ContentType)
 
                 val serializedContent = when (payload) {
-                    is EmptyContent -> config.serializer.write(Unit, contentType)
+                    is EmptyContent -> EmptyContent
                     else -> config.serializer.write(payload, contentType)
                 }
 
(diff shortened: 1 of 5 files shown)


wkornewald:
@e5l Since I can't add reviewers, could you please have a look at this fix? :)

wkornewald:
I've updated the solution to directly deal with this in `JsonFeature`. At first I thought it would be better to leave this decision to the serializer (because who knows how some serialization formats represent "no content"), but in the end that's just too error-prone (I almost forgot about `KotlinxSerializer` because it's not covered by `JsonTest`).

ktor-client/ktor-client-features/ktor-client-json/common/src/io/ktor/client/features/json/JsonFeature.kt
```diff
@@ -109,7 +110,7 @@ class JsonFeature internal constructor(val config: Config) {
                 context.headers.remove(HttpHeaders.ContentType)
 
                 val serializedContent = when (payload) {
-                    is EmptyContent -> config.serializer.write(Unit, contentType)
+                    is EmptyContent -> EmptyContent
```
e5l, line 113: Could you also make the same behavior for `Unit` type?

wkornewald:
@e5l You mean, like in the new commit I just pushed?

e5l:
LGTM, merged

Issue #377, opened 2018-04-02
receiveOrNull susceptible to type erasure


mjburghoffer:
https://github.com/ktorio/ktor/blob/84f213a320f38e11ffaafe7b5e7f321a6b7adaab/ktor-server/ktor-server-core/src/io/ktor/request/ApplicationReceiveFunctions.kt#L70

The highlighted code is susceptible to type erasure.  If I want to `receiveOrNull<Map<String, List<String>>>()`, for instance, at runtime I will only have access to `Map<*, *>::class`.  This is a problem for a lot of serializations frameworks, like jackson, who use `TypeToken`s or the like.

nikolajakshic:
This [sample](https://gist.github.com/nikolajakshic/0a7ea29a53fab557e40deb14d7d201be) throws [this exception](https://gist.github.com/nikolajakshic/0cba31c828c8b970da8d4e399396e0ad). I just want to be sure that this is related, right? If so, any updates on this?

mjburghoffer:
Yes this is 100% related. As far as I know there have been no updates to
this issue.

mkobit:
From the client side, there are https://github.com/ktorio/ktor/issues/346 and https://github.com/ktorio/ktor/issues/365 which are similar

dmakarenko-connyun:
Guys, any progress on this?

nikolajakshic:
@cy6erGn0m commenting just as a reminder, similar issues on the client side are solved, somehow this got forgotten.

tylerbobella:
Commenting on this because I started to create a small web app with ktor + jackson and this was a full-stop for me since i could not receive a `List<MyType>`. Is there a workaround for this? Does the gson plugin work with something like `List<MyType>` or @mjburghoffer example?

nikolajakshic:
@tylerbobella Same thing happens with Gson as well. And for workaround, well you can still take the whole string and convert it with jackson/gson manually.

With Gson, instead of:
```kotlin
val items = call.receive<List<MyType>>()
```
use
```kotlin
val items = Gson().fromJson<List<MyType>>(call.receiveText(), object : TypeToken<List<MyType>>() {}.type)
```

You can write extension function and make it look prettier.

tylerbobella:
Thanks for the reply @nikolajakshic. I will try something similar with Jackson tonight and will report back.

tylerbobella:
For those using jackson I created this extension function that allows us to serialize a list of objects

```
[...]

PR #1956, opened 2020-06-19
Fix type erasure problem when using receive and jackson, e.g. receive<List<Entity>>()

vansickle:
**Subsystem**

Jackson

**Motivation**

The problem is described in #377. 
Basically I expect that I can receive generic objects via `receive<List<MyEntity>>()`, but currently when trying to do so I get an error like:

```
java.util.LinkedHashMap cannot be cast to io.ktor.tests.jackson.MyEntity
java.lang.ClassCastException: java.util.LinkedHashMap cannot be cast to io.ktor.tests.jackson.MyEntity
```
Which is result of type erasure or more specifically that type information available in KType is not properly provided to Jackson

Test for this is available in PR.

**Workarounds**

If I get it right the #377 is closed with the solution to use kotlinx.serialization but I believe it's wrong way to fix the problem - kotlinx.serialization is totally different beast than Jackson and if Jackson is mentioned in docs as possible feature for Content-Negotiation then it should be fully supported

Also I can use other workarounds like instantiate my own objectmapper or make custom jacksonconverter, but I do believe it should be fixed in ktor upstream.

**Solution**

Make an adapter from `KType` to Jackson's `TypeReference`. 

Actually need a little help with the solution - wanna to use `ktor-server-core` `ReflectionUtils` classes but don't know hot to properly move/reuse them, so just copy/pasted for now

**Follow-Up**

I guess the same approach can be used for Gson and it's TypeToken implementation

### DIFF
--- a/ktor-features/ktor-jackson/jvm/src/io/ktor/jackson/JacksonConverter.kt
+++ b/ktor-features/ktor-jackson/jvm/src/io/ktor/jackson/JacksonConverter.kt
@@ -4,6 +4,7 @@
 
 package io.ktor.jackson
 
+import com.fasterxml.jackson.core.type.*
 import com.fasterxml.jackson.core.util.*
 import com.fasterxml.jackson.databind.*
 import com.fasterxml.jackson.module.kotlin.*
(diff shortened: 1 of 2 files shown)


rsinukov:
Thanks for you contribution and sory for such a huge delay. It seems like this problem is already fixed in previous releases. Closing this PR.
Sorry again for the delay.

Issue #604, opened 2018-09-24
Cannot log status code with MDC


windoze:
Code snippet:
```
fun Application.module() {
    install(CallLogging) {
        mdc("status_code") {
            it.response.status()?.toString()
        }
    }
    routing {
        get("/") {
            call.respondText("OK", contentType = ContentType.Text.Plain)
        }
    }
}
```
The MDC value of "status_code" is always null.

The reason is CallLogging feature evaluates all MDC values **before** the call gets processed.
In CallLogging.kt:
```
            if (feature.mdcEntries.isNotEmpty()) {
                pipeline.intercept(loggingPhase) {
                    val mdc = feature.setupMdc(call)
                    withContext(MDCSurvivalElement(mdc)) {
                        try {
                            proceed()
                            feature.logSuccess(call)
                        } finally {
                            feature.cleanupMdc()
                        }
                    }
                }
            }
```
Simply calling `feature.setupMdc(call)` after `proceed()` again got the issue fixed, but I'm not sure it's the correct way.

cy6erGn0m:
Unfortunately it is not going to work properly as we should fill all MDC values before processing to pass them through all coroutines. The second reason is that we don't know how many times and when we should evaluate lambdas. Adding it after `proceed()` is too ad-hoc.

cy6erGn0m:
The other question is what for? Why do you need to add status code to MDC? `CallLogging` feature logs all status codes at trace log level

windoze:
Hi, @cy6erGn0m,
I'was trying to make Ktor work with Logstash, which supports log in JSON format, so user can add additional key/values along with normal log messages.
I checked out the CallLogging code but didn't find any appropriate places to add K/V except MDC, hence the question.
If this is going nowhere, maybe a new LogstashCallLogging feature from scratch is needed, looks like some minor modification though.

cy6erGn0m:
Why it should be MDC? Can't logstash parse regular log messages?

windoze:
It's not about MDC actually, what I wanted is to log in JSON format and add some customized properties, with Fluentd/Elasticsearch I can dig into logs easily.
Logback has a Logstash encoder (net.logstash.logback:logstash-logback-encoder), which can output logs in JSON format and add properties via `log.info("message", kv(someKey, someValue))`, but I cannot find anywhere to customize format and content with CallLogging.
The encoder also supports MDC, which looks like an alternative but turned out it didn't work as I expected.

[...]

PR #1957, opened 2020-06-19
Add the ability to defer MDC entry computation

CarrotCodes:
**Subsystem**
ktor-server-jvm

**Motivation**
As discussed in https://github.com/ktorio/ktor/issues/604, it's not currently possible to add MDC entries with values that contain response information.

Services like Datadog are set up to pull response codes out of predefined MDC entries, so it would be helpful to support this out of the box.

**Solution**
I've added a parameter to `MDCEntry` called `defer`, with a default value of `false` so as to keep existing behaviour the same.

Deferred MDC entries have the same lifecycle / management as regular entries, but their values are set to `null` to start with. After a request has been handled, but before the result is logged, we compute the value of any `MDCEntry` marked as being deferred.

I've also added a couple of tests to make sure existing behaviour, without `defer`, is maintained, and that the deferred entries survive context switches (using the existing tests as examples to follow).

This is my first PR - I'm more than happy to change this as desired, and for you to make changes, just let me know.

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-call-logging/jvm/src/io/ktor/server/plugins/CallLogging.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-call-logging/jvm/src/io/ktor/server/plugins/CallLogging.kt
@@ -29,7 +29,7 @@ public class CallLogging private constructor(
     private val formatCall: (ApplicationCall) -> String
 ) : MDCProvider {
 
-    internal class MDCEntry(val name: String, val provider: (ApplicationCall) -> String?)
+    internal class MDCEntry(val name: String, val defer: Boolean = false, val provider: (ApplicationCall) -> String?)
 
     /**
      * Configuration for [CallLogging] plugin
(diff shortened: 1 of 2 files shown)


e5l (approved):
Hey @CarrotCodes, thanks for the PR! LGTM, could you rebase it on `main`?
We can merge it in `2.0.0`

CarrotCodes:
> Hey @CarrotCodes, thanks for the PR! LGTM, could you rebase it on `main`? We can merge it in `2.0.0`

Thanks for taking a look @e5l - I've rebased and tweaked the tests a little to match the code style around them. I'm having trouble running the entire `jvmTest` suite but everything in `CallLoggingTest` is happy locally at least.

CarrotCodes:
I rebased to fix the conflicts and I believe this PR might not be necessary any more after https://github.com/ktorio/ktor/pull/2716 - tests seem to show status codes being logged just fine 🤞 

Let me know if it's still needed and I'll reopen, otherwise going to close now to reduce the number of PRs you have to look at.

e5l:
@CarrotCodes, thank you for the contribution!

Issue #720, opened 2018-11-12
Add support for ports in withTestApplication


dave08:
```kotlin
fun TestApplicationEngine.handleRequest(
    method: HttpMethod,
    port: Int = 80,
    uri: String,
    setup: TestApplicationRequest.() -> Unit = {}
): TestApplicationCall = handleRequest {
    this.uri = uri
    this.method = method
    this.addHeader(HttpHeaders.Host, "localhost:$port")
    setup()
}
```

brushmate:
This functionality can be simulated by setting the `Host` header:

    handleRequest(Get, "/") {
        addHeader(HttpHeaders.Host, "localhost:1234")
    }.apply {
        assertEquals(1234, request.port())
    }

oleg-larshin:
Please check the <a href="https://youtrack.jetbrains.com/issue/KTOR-725">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

PR #1961, opened 2020-06-21
Add optional port parameter in withTestApplication (Issue #720)

Bartosz-D3V:
**Subsystem**
ktor-server-test-host

**Motivation**
This PR solves issue #720. Developers can now pass additional port parameter to every handleRequest method that will make a call to specific port.

**Solution**
handleRequest method will now accept additional port parameter. The parameter is marked as optional with default null value for backward compatibility purpose. If parameter is not provided then engine will look for header HOST if header is not present either then default one (80) is used.

Example use:
```
handleRequest(HttpMethod.Get, "/endpoint") { port = 8080 }
```


e5l (changes requested):
Hey, @Bartosz-D3V. Thanks for the PR, LGTM. We can merge it in the next minor branch `1.6.0-eap`. Could you rebase on it and run  `./gradlew apiDump` to verify that API change is compatible?

Bartosz-D3V:
Hi @e5l 
I just pushed couple of changes - I rebased my changes on top of 1.6.0-eap, and run `./gradlew apiDump` successfully.

PR #1963, opened 2020-06-22
Fix parsing of quoted header parameter value

e5l:


ktor-http/common/src/io/ktor/http/HttpHeaderValueParser.kt
```diff
@@ -155,36 +155,43 @@ private fun parseHeaderValueParameter(text: String, start: Int, parameters: Lazy
...
+    return position to value.subtrim(start, position)
 }
 
 private fun parseHeaderValueParameterValueQuoted(value: String, start: Int): Pair<Int, String> {
     var pos = start
     val sb = StringBuilder()
     while (pos <= value.lastIndex) {
-        val c = value[pos]
-        when (c) {
+        val current = value[pos]
```
Ololoshechkin, line 180: Nit: maybe `currentSymbol` or any shorter version of it would be a better name here?

ktor-http/common/src/io/ktor/http/HttpHeaderValueParser.kt
```diff
@@ -155,36 +155,43 @@ private fun parseHeaderValueParameter(text: String, start: Int, parameters: Lazy
...
-            '"' -> return parseHeaderValueParameterValueQuoted(value, pos + 1)
-            ';', ',' -> return pos to value.subtrim(start, pos)
-            else -> pos++
+    if (value.length == start) {
+        return start to ""
+    }
+
+    var position = start
+    if (value[start] == '"') {
+        return parseHeaderValueParameterValueQuoted(value, position + 1)
```
Ololoshechkin, line 164: What would happen if `value == "\""`?

Ololoshechkin: I hope that case is being handled on the call side?

ktor-http/common/src/io/ktor/http/HttpHeaderValueParser.kt
e5l: Thanks, will change

ktor-http/common/src/io/ktor/http/HttpHeaderValueParser.kt
e5l: I can't get the difference here. Could you clarify?

Ololoshechkin: According to the way `parseHeaderValueParameterValueQuoted` works, the result would be the same for `"\""` and `"\"\""` (in both cases result would be a mapping from size to an empty string, and the last parsed symbol is `\"` in both cases). 
I guess, `\"\"` must be a quote of empty string and `\"` must be either an error or a text that just contains single  quotation mark.
So I think result must be empty for `"\"\""` and `\"` for `\"`

e5l: Thanks. I will add the test and fix the problem.

ktor-http/common/src/io/ktor/http/HttpHeaderValueParser.kt
```diff
@@ -174,27 +178,43 @@ private fun parseHeaderValueParameterValue(value: String, start: Int): Pair<Int,
...
+}
+
+private fun String.nextIsSemicolonOrEnd(start: Int): Boolean {
+    var position = start + 1
+    loop@ while (position < length) {
+        when (get(position)) {
+            ';' -> return true
+            ' ' -> {
+                position += 1
+                continue@loop
```
Ololoshechkin, line 212: Nit: 
`while (position < length && get(position) == ' ') position++`
`return get(position) == ';'`

probably would be easier to read
But feel free to leave it as it is.
LGTM

e5l: Yep, I actually like your way. Except for missing `{}` in while loop

e5l: `parseHeaderValueParameterValueQuoted` performs the check and append `"` at the start if failed to find a closing one

Ololoshechkin: Yep, sounds good

ktor-http/common/test/io/ktor/tests/http/HeadersTest.kt
```diff
@@ -153,7 +153,7 @@ class HeadersTest {
             parseHeaderValue("justValue;x=\"\"")
         )
         assertEquals(
-            listOf(HeaderValue("justValue", listOf(HeaderValueParam("x", "abc\\")))),
+            listOf(HeaderValue("justValue", listOf(HeaderValueParam("x", "\"abc\\")))),
```
Ololoshechkin, line 156: Why?

PR #1971, opened 2020-06-25
Fix followRedirects setting in JS client implementation

### LINKED ISSUE #1968: [JS client] Cannot change redirect policy by followRedirects=false
Ktor HTTP client 1.3.2 on Kotlin/JS; dependencies: 
```
                implementation("io.ktor:ktor-client-json:$ktorVersion")
                implementation("io.ktor:ktor-client-js:$ktorVersion")
                implementation("io.ktor:ktor-client-json-js:$ktorVersion")
                implementation("io.ktor:ktor-client-serialization-js:$ktorVersion")
```

**Describe the bug**
While debugging the situation with redirects location, I found that if you call something like `println(resp)` where resp is a `HttpResponse` result of a request that returns `HTTP/1.1 302 Found`, string will be `HttpResponse[http://localhost:3000/api/v1/auth/logout, 200 OK]`, where I excected something like `HttpResponse[http://localhost:3000/api/v1/auth/logout, 302 Found]`.

In network tab of browser developer tools response shown as `302 Found` with correct `Location` header, but ktor client shows different results...

**To Reproduce**
1. Take [this repo](https://github.com/kam1sh/kvision-sandbox/tree/ktor-http-issue) for example, change url in `App.kt` to something suitable in your environment (httpbin, for example)
2. Open in browser localhost:3000
3. See string representation of response with 200 OK

**Expected behavior**
302 Found with `Location` header

**Screenshots**
image

If what, I can try to debug this and fix by myself, but later...

P.S. Thank you for ktor framework and client. They're awesome.

kam1sh:
**Subsystem**
Client, `ktor-client-core` (common and js) and `ktor-utils` (common)

**Motivation**
I want to resolve #1968 -- catch `HTTP 302 Found` and similiar redirect responses from backend and handle them manually in my frontend application.

**Solution**
I solved it by adding new attribute (`redirect`) that gets value from `config.followRedirects` and by using `attributes` container value passes to `HttpRequestData.toRaw`, that uses JS Fetch API.

Also, I changed AttributeKey signature from class to data class. I decided to do this because of need to have equals and hashCode methods for proper using AttributeKey as a mapping key. For example, now `AttributeKey("test") == AttributeKey("test")`, and so on.

Hope it gets merged soon, because I don't know other such powerful HTTP toolkits as ktor-client =)

### DIFF
--- a/ktor-client/ktor-client-core/js/src/io/ktor/client/engine/js/JsClientEngine.kt
+++ b/ktor-client/ktor-client-core/js/src/io/ktor/client/engine/js/JsClientEngine.kt
@@ -4,6 +4,7 @@
 
 package io.ktor.client.engine.js
 
+import io.ktor.client.*
 import io.ktor.client.engine.*
 import io.ktor.client.features.*
 import io.ktor.client.engine.js.compatibility.*
--- a/ktor-client/ktor-client-core/js/src/io/ktor/client/engine/js/JsUtils.kt
+++ b/ktor-client/ktor-client-core/js/src/io/ktor/client/engine/js/JsUtils.kt
@@ -8,6 +8,7 @@ import io.ktor.client.engine.*
 import io.ktor.client.fetch.RequestInit
 import io.ktor.client.request.*
 import io.ktor.http.content.*
+import io.ktor.util.*
 import io.ktor.utils.io.*
 import io.ktor.utils.io.core.*
 import kotlinx.coroutines.*
@@ -35,8 +36,7 @@ internal suspend fun HttpRequestData.toRaw(callContext: CoroutineContext): Reque
     return buildObject {
         method = this@toRaw.method.value
         headers = jsHeaders
-        redirect = RequestRedirect.FOLLOW
-
+        redirect = attributes.getOrNull(AttributeKey<String>("redirect")) ?: RequestRedirect.FOLLOW
         bodyBytes?.let { body = Uint8Array(it.toTypedArray()) }
     }
 }
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/engine/HttpClientEngine.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/engine/HttpClientEngine.kt
@@ -57,6 +57,8 @@ interface HttpClientEngine : CoroutineScope, Closeable {
             val requestData = HttpRequestBuilder().apply {
                 takeFromWithExecutionContext(context)
                 body = content
+                if (!client.config.followRedirects)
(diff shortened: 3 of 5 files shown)


kam1sh:
> that is quite a specific extension point

I agree with that =)

> Could we introduce the extension property redirect on HttpRequestBuilder for that

You mean add in `JsUtils.kt` this extension property?
```kotlin
var HttpRequestBuilder.redirect: String
    get() = attributes.getOrNull(AttributeKey<String>("redirect")) ?: RequestRedirect.FOLLOW.toString()
    set(value) = attributes.put(AttributeKey("redirect"), value)
```

But how then could I access this property from `HttpRequestData.toRaw()`? And how would it be set in `config {...}`? I don't understand.

> It also looks like that HttpRedirect feature should be aware of that flag

I saw that value `RequestRedirect.FOLLOW` was hardcoded in `JsUtils`... but anyway, I'll look at `HttpRedirect` feature.

kam1sh:
About API cleanup: do you have gradle task for linting/formatting code, or I can just use IDEA's feature `Folder -> Reformat code`? I've tried that, but it formatted too many files, though I enabled both `Use import with *` in formatting settings...

e5l:
I mean the method introducing by that :)

ktor-client/ktor-client-core/js/src/io/ktor/client/engine/js/JsClientEngine.kt
```diff
@@ -4,6 +4,7 @@
 
 package io.ktor.client.engine.js
 
+import io.ktor.client.*
```
shoresea, line 7: Looks like redundant import and can be removed

ktor-client/ktor-client-core/js/src/io/ktor/client/engine/js/JsUtils.kt
```diff
@@ -35,8 +36,7 @@ internal suspend fun HttpRequestData.toRaw(callContext: CoroutineContext): Reque
     return buildObject {
         method = this@toRaw.method.value
         headers = jsHeaders
-        redirect = RequestRedirect.FOLLOW
-
+        redirect = attributes.getOrNull(AttributeKey<String>("redirect")) ?: RequestRedirect.FOLLOW
```
shoresea, line 39: How about adding a `getOrDefault` function to `Attributes`?
`fun <T : Any> getOrDefault(key: AttributeKey<T>, default: T): T = getOrNull(key) ?: default`

e5l:
I'm not sure that `manual` redirect is working: it behaves the same as `no-cors` mode, so the response has nothing inside: status code is 0, no headers, no body, and etc. I'm going to investigate further.

PR #1973, opened 2020-06-26
Split engine test suite to improve IDE performance

e5l:


ktor-server/ktor-server-jetty/jvm/test/io/ktor/tests/server/jetty/JettyServletApplicationEngine.kt
```diff
@@ -7,42 +7,24 @@ package io.ktor.tests.server.jetty
...
-    }
-}
-
-// the factory and engine are only suitable for testing
-// you shouldn't use it for production code
-
-private class Servlet(private val async: Boolean) :
+/**
+ * The factory and engine are only suitable for testing. You shouldn't use it for production code.
+ */
```
Ololoshechkin, line 15: Is there a way to use it if it's internal in `*.test.*` package?

Ololoshechkin: Upd: Well, I get the point. I think we should make some annotation for that, like `@TestOnly` and check that it's being used only in test module

ktor-server/ktor-server-test-host/jvm/src/io/ktor/server/testing/suites/CompressionTestSuite.kt
```diff
@@ -0,0 +1,133 @@
...
+import java.io.*
+import java.util.zip.*
+import kotlin.test.*
+
+abstract class CompressionTestSuite<TEngine : ApplicationEngine, TConfiguration : ApplicationEngine.Configuration>(
+    hostFactory: ApplicationEngineFactory<TEngine, TConfiguration>
+) : EngineTestBase<TEngine, TConfiguration>(hostFactory) {
+
+    @Test
+    fun testLocalFileContentWithCompression() {
```
Ololoshechkin, line 29: Nit: I was thinking of how to make this similar-looking tests table-driven or at least extract some common code, but maybe it's not worth it as there are quite a few parameters to introduce.
Maybe let's leave it as it is.

ktor-server/ktor-server-netty/jvm/test/io/ktor/tests/server/netty/NettyEngineTest.kt
```diff
@@ -5,12 +5,45 @@
...
+}
+
+class NettyContentTest : ContentTestSuite<NettyApplicationEngine, NettyApplicationEngine.Configuration>(Netty) {
+    init {
+        enableSsl = true
+    }
+
+    override fun configure(configuration: NettyApplicationEngine.Configuration) {
+        configuration.shareWorkGroup = true
+    }
```
Ololoshechkin, line 27: Nit: Maybe extract a a parent class with enableSsl = true, configuration.shareWorkGroup = true ?

ktor-server/ktor-server-test-host/jvm/src/io/ktor/server/testing/suites/CompressionTestSuite.kt
e5l: I'm experimenting with that on the client-side, and haven't found any good solutions yet.

Maybe we should look in different test frameworks, something like `spec`. But I'm not sure if they work with multiplatform.

Issue #1974, opened 2020-06-26
qop missing from auth header when using digest auth


RyanGrif:
When using the Ktor http client (ktor 1.3.1) with digest auth, calls to a downstream service fail with a 401 status code. This is because the "qop" field isn't included in the authentication header.

The problem is caused by a bug in DigestAuthProvider.addRequestHeaders. This line:

`actualQop?.let { this["qop"] }`

should be

`actualQop?.let { this["qop"] = it }`

Ololoshechkin:
Thanks for the bug report!
Would you like to make a pull request?

RyanGrif:
Yes please. Does someone need to give me permission to contribute?

e5l:
Thanks for the fix, merged and closed

PR #1980, opened 2020-06-30
Fix missing qop in DigestAuthProvider (Issue #1974)

RyanGrif:
**Subsystem**
Client

**Motivation**
The problem is described in issue #1974. When using digest auth, the qop field is never included in the authentication header. This breaks digest authentication.

**Solution**
Add missing `= it` so that the qop field is assigned to the correct value.


e5l:
Hi @RyanGrif, thanks for the PR. 
LGTM, Merged

PR #1982, opened 2020-06-30
Fix exception logged twice in CIO launch channel

e5l:

### DIFF
--- a/ktor-io/common/src/io/ktor/utils/io/Coroutines.kt
+++ b/ktor-io/common/src/io/ktor/utils/io/Coroutines.kt
@@ -115,13 +115,24 @@ private fun <S : CoroutineScope> CoroutineScope.launchChannel(
     attachJob: Boolean,
     block: suspend S.() -> Unit
 ): ChannelJob {
-
+    val originJob = coroutineContext[Job]
     val job = launch(context) {
         if (attachJob) {
             channel.attachJob(coroutineContext[Job]!!)
         }
+
         @Suppress("UNCHECKED_CAST")
-        block(ChannelScope(this, channel) as S)
+        val scope = ChannelScope(this, channel) as S
+
+        try {
+            block(scope)
+        } catch (cause: Throwable) {
+            if (originJob != null) {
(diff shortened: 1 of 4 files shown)


e5l:
Sorry, I tested and found the critical here. 
This avoids logging, but doesn't cancel launcher job - decomposition stopped working.

I'm looking for the fix.

ktor-io/common/src/io/ktor/utils/io/Coroutines.kt
```diff
@@ -43,32 +43,6 @@ fun CoroutineScope.reader(
...
-    coroutineContext: CoroutineContext,
-    autoFlush: Boolean = false, parent: Job? = null,
-    block: suspend ReaderScope.() -> Unit
-): ReaderJob {
-    val channel = ByteChannel(autoFlush)
-    return reader(coroutineContext, channel, parent, block).also {
-        channel.attachJob(it)
-    }
-}
-
```
Ololoshechkin, line 71: Why did we need this Reader/Writer jobs before and why they are not used anymore?

e5l: That's deprecated API without `CoroutineScope` that we don't want to support

ktor-io/common/src/io/ktor/utils/io/Coroutines.kt
```diff
@@ -115,15 +64,23 @@ private fun <S : CoroutineScope> CoroutineScope.launchChannel(
     attachJob: Boolean,
     block: suspend S.() -> Unit
 ): ChannelJob {
+    val originJob = coroutineContext[Job]
     val job = launch(context) {
         if (attachJob) {
             channel.attachJob(coroutineContext[Job]!!)
```
Ololoshechkin, line 70: is it `channel.attachJob(originJob!!)` ?

e5l: Thanks. Will fix

PR #1984, opened 2020-06-30
fix #1970 - make MultiPartFormDataContent open and exposed boundary

### LINKED ISSUE #1970: Objections to changing boundary to internal on MultiPartFormDataContent?
https://github.com/ktorio/ktor/blob/3ce3906c2bfa3c86238d40e9b67006b2b020bdaf/ktor-client/ktor-client-core/common/src/io/ktor/client/request/forms/FormDataContent.kt#L43

I'd like to change this to internal so I can consume the boundary like so:

```kotlin
class MultiPartMixedDataContent(
        parts: List<PartData>
) : MultiPartFormDataContent {
    override val contentType: ContentType = ContentType.MultiPart.Mixed.withParameter("boundary", boundary)
}
```

jschneidereit:
Fixes #1970 

**Subsystem**
Client

**Motivation**
Before this change `MultiPartFormDataContent` was not compatible with servers that accepted `multipart/mixed` content. By exposing `boundary` clients of ktor client can implement their own classes to override the content type boundary behavior like so:

```kotlin
class MultiPartMixedDataContent(
        parts: List<PartData>
) : MultiPartFormDataContent(parts) {
    override val contentType: ContentType = ContentType.MultiPart.Mixed.withParameter("boundary", boundary)
}
```

**Solution**
I made `MultiPartFormDataContent` an `open` class and exposed `boundary` to subclasses to allow for the above code snippet

### DIFF
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/request/forms/FormDataContent.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/request/forms/FormDataContent.kt
@@ -37,10 +37,10 @@ class FormDataContent(
  *
  * @param parts: form part data
  */
-class MultiPartFormDataContent(
+open class MultiPartFormDataContent(
     parts: List<PartData>
 ) : OutgoingContent.WriteChannelContent() {
-    private val boundary: String = generateBoundary()
+    protected val boundary: String = generateBoundary()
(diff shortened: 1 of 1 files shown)


jschneidereit:
Hey there @e5l, any updates on this?

Issue #57, opened 2016-12-10
Documentation to use with Kotlin 1.1 EAP?


Jire:
I tried to get it working with both the Netty and Jetty embedded server. I am using Ktor 0.2.4 and Kotlin 1.1-M03.

```kotlin
import org.jetbrains.ktor.application.call
import org.jetbrains.ktor.application.install
import org.jetbrains.ktor.features.DefaultHeaders
import org.jetbrains.ktor.jetty.embeddedJettyServer
import org.jetbrains.ktor.logging.CallLogging
import org.jetbrains.ktor.response.respondText
import org.jetbrains.ktor.routing.Routing
import org.jetbrains.ktor.routing.get

fun main(args: Array<String>) = embeddedJettyServer {
	install(DefaultHeaders)
	install(CallLogging)
	install(Routing) {
		get("/") {
			call.respondText("Hello, World!")
		}
	}
}.start()
```

```
2016-12-09 20:23:11.000 [main] INFO  org.eclipse.jetty.util.log - Logging initialized @426ms
2016-12-09 20:23:11.064 [main] TRACE embedded - Starting server...
2016-12-09 20:23:11.065 [main] INFO  org.eclipse.jetty.server.Server - jetty-9.3.13.v20161014
2016-12-09 20:23:11.126 [main] INFO  o.e.jetty.server.AbstractConnector - Started ServerConnector@69a10787{HTTP/1.1,[http/1.1, h2c, h2c-17, h2c-16, h2c-15, h2c-14]}{0.0.0.0:80}
2016-12-09 20:23:11.126 [main] INFO  org.eclipse.jetty.server.Server - Started @555ms
2016-12-09 20:23:11.126 [main] TRACE embedded - Server running.
2016-12-09 20:23:13.826 [ktor-pool-1-thread-1] ERROR embedded - null: GET - /
java.lang.IncompatibleClassChangeError: class kotlin.reflect.jvm.internal.KFunctionImpl has interface kotlin.jvm.internal.FunctionImpl as super class
	at java.lang.ClassLoader.defineClass1(Native Method) ~[na:1.8.0_112]
	at java.lang.ClassLoader.defineClass(ClassLoader.java:763) ~[na:1.8.0_112]
	at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142) ~[na:1.8.0_112]
	at java.net.URLClassLoader.defineClass(URLClassLoader.java:467) ~[na:1.8.0_112]
[...]

PR #1985, opened 2020-06-30
fix webpack warning about ktor-client-core critical dependency

toth-istvan-zoltan:
**Subsystem**
ktor-client-core

**Motivation**
Solve webpack warning about a critical depencency (using require with an expression as a parameter).

```
WARNING in /Users/tiz/src/sandbox/ktor-client-test/build/js/packages_imported/ktor-ktor-client-core/1.3.2/ktor-ktor-client-core.js 16801:13-32
Critical dependency: the request of a dependency is an expression
 @ ./kotlin/ktor-client-test.js
 @ multi ./kotlin/ktor-client-test.js
```

**Solution**
Changed the require call to use the fixed 'node-fetch' string instead of a passed parameter.

This will cause builds to fail because of the missing 'node-fetch' dependency.

However, builds already fail because of missing dependencies. Users have to declare them in gradle build or add them to Webpack's externals so the build. See  #961 and #57

As of now I think this should be handled by documentation. There are some proposed solutions in  #961 and #57

In the future a separation of node and browser sources (to the necessary level) might be the best solution but that requires much more time to implement.

I've found the best way is to handle this by adding an externals.js into webpack.config.d with the following content:

```javascript
config.externals = [
    'text-encoding',
    'utf-8-validate',
    'abort-controller',
    'bufferutil',
    'fs',
    'node-fetch'
]
```

### DIFF
--- a/ktor-client/ktor-client-core/js/src/io/ktor/client/engine/js/compatibility/Utils.kt
+++ b/ktor-client/ktor-client-core/js/src/io/ktor/client/engine/js/compatibility/Utils.kt
@@ -25,7 +25,7 @@ internal suspend fun commonFetch(
     val promise: Promise<Response> = if (PlatformUtils.IS_BROWSER) {
         fetch(input, init)
     } else {
-        jsRequire("node-fetch")(input, init)
+        jsRequireNodeFetch()(input, init)
     }
 
     promise.then(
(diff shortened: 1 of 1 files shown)


e5l:
Hi @toth-istvan-zoltan, thanks for the PR.
LGTM, Merged

Issue #1967, opened 2020-06-24
HTTP-client auth with Bearer token


Savrov:
**Subsystem**
Client/Server, particular relevant ktor module(s) if applicable.
ktor HTTP-client, auth

**Is your feature request related to a problem? Please describe.**
Currently, Ktor HTTP-client supports only 2 types of providers: Basic and Digest. But in a real-life project is a standard to use Bearer token pair - access and refresh. So it will be great to add support of Bearer tokens to multiplatform Ktor's HTTP-client.
So the problem itself can be split into 4 sub-tasks:
1. Persist token pair in a safe place
2. Add them to auth headers
3. In case when the request with access token returns 401, 
3.1. pause all present requests
3.2 refresh it with refresh token.. and so on, regular flow.
3.3 restart interrupted requests from 3.1

**Describe the solution you'd like**
Create BearerAuthConfig, BearerAuthProvider like its Basic and Digest versions. Update Auth class to be able to save, load and refresh tokens. Something like this:
```
fun provide(): HttpClient = HttpClient {
        install(Auth) {
            bearer {
                refresh = TODO("lambda function that get new access token and return a string"),
                revoke = TODO("lambda function that removes tokens")
            }
        }
    }
```

**Motivation to include to ktor**
Add support of Bearer tokens for clients.

Ololoshechkin:
Thanks for the feature request!
What you suggest is actually 2 features:
1. Bearer (jwt) support in HttpClient
2. Automatic token refresh.

The first one should be added as a part of Auth feature and we'll be working on that.
And the second one (token renewal) is currently a bit out of our scope but we'll keep our eye on it in future.

Savrov:
Thank you for a quick response,
I personally think that Bearer support without refresh functionality is pretty useless and un-safe (to make access token work long enough, access token's lifetime should be extended to a week at least. According to my knowledge access token should be valid for 5-10 min). So, will be much better, if these two features will be delivered together.
If there is a way to help you in solving these tasks, please let me know.

Ololoshechkin:
Thank you for offering your help!
We will appreciate if you make a PR with your changes. Please tell if you need any help or guidance from us.

Savrov:
@Ololoshechkin, I don't see a way to add support of token refresh w/o complete rewrite of Auth.kt
As I understand, the whole job must happen inside the `install` function of `Auth` class:
we check if response status is Unauthorized and a current provider is BearerAuthProvider (that must be created)
[...]

PR #1991, opened 2020-07-05
Add a bearer auth feature for http-client

Savrov:
**Subsystem**
Http-Client

**Motivation**
Add support of Bearer tokens to the client-side. (Issue #1967 )

**Solution**
Added a BearerAuthProvider with an ability to load and refresh tokens.

### DIFF
--- a/ktor-client/ktor-client-features/ktor-client-auth/common/src/io/ktor/client/features/auth/providers/BearerAuthProvider.kt
+++ b/ktor-client/ktor-client-features/ktor-client-auth/common/src/io/ktor/client/features/auth/providers/BearerAuthProvider.kt
@@ -14,11 +14,11 @@ import io.ktor.http.auth.*
  */
 fun Auth.bearer(block: BearerAuthConfig.() -> Unit) {
     with(BearerAuthConfig().apply(block)) {
-        providers.add(BearerAuthProvider(refreshTokenFun, loadBearerTokenFun, true, realm))
+        providers.add(BearerAuthProvider(refreshTokensFun, loadTokensFun, true, realm))
     }
 }
 
-data class BearerToken(
+data class BearerTokens(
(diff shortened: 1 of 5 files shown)


ktor-client/ktor-client-features/ktor-client-auth/common/src/io/ktor/client/features/auth/providers/BearerAuthProvider.kt
```diff
@@ -0,0 +1,84 @@
...
+fun Auth.bearer(block: BearerAuthConfig.() -> Unit) {
+    with(BearerAuthConfig().apply(block)) {
+        providers.add(BearerAuthProvider(refreshTokenFun, loadTokensFun, true, realm))
+    }
+}
+
+data class Tokens(
+    val accessToken: String,
+    val refreshToken: String,
+    val type: String
```
Ololoshechkin, line 24: Why do we need type here?
If it's always "Bearer" maybe it worth renaming the class to `BearerTokens` and removing the `type` param?

Savrov: Agree. Will fix it

Savrov:
@Ololoshechkin , will be it merged? Or another fix needed?

Ololoshechkin:
@Savrov Thanks! I will merge it to a separate branch for now and then we need to think about the API and before that we need to consider user stories. But from the backend perspective I am happy with your change, just we needs to design the API before adding to ktor.

moallemi:
Any update about when you add it to ktor?

ovitrif:
What else is needed to get this into the lib?

grrigore:
Any updates?

Issue #1990, opened 2020-07-02
Digest auth fails for urls with query params


RyanGrif:
When using the Ktor http client (ktor 1.3.1) with digest auth, calls to a downstream service fail with a 401 status code when query parameters are included. This is because the uri field does not include the query params.

Changing all places in `DigestAuthProvider.addRequestHeaders` which currently use `url.encodedPath` to use `url.fullPath` instead seems to fix this problem (at least for the API I'm calling).

I can raise PR for this if someone agrees agrees this is the correct fix?

e5l:
Hey @RyanGrif, thanks for the report. 
It looks like the right direction. Could you make the PR? :)

e5l:
Merged. Closed. Thanks for the PR

PR #1992, opened 2020-07-07
Fix uri field in digest auth header to include query params

RyanGrif:
Subsystem
Client

Motivation
The problem is described in issue #1990 . When using digest auth, the uri field does not include the request parameters included in the original request. This breaks digest authentication.

Solution
Changing all places in DigestAuthProvider.addRequestHeaders which currently use url.encodedPath to use url.fullPath so that the uri field and token include the request parameters.

### DIFF
--- a/ktor-client/ktor-client-features/ktor-client-auth/common/src/io/ktor/client/features/auth/providers/DigestAuthProvider.kt
+++ b/ktor-client/ktor-client-features/ktor-client-auth/common/src/io/ktor/client/features/auth/providers/DigestAuthProvider.kt
@@ -82,7 +82,7 @@ class DigestAuthProvider(
         val credential = makeDigest("$username:$realm:$password")
 
         val start = hex(credential)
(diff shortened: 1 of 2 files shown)


e5l:
Hey @RyanGrif, Thanks for the request. 
LGTM.
Could you add the test to verify the behavior?

RyanGrif:
> Hey @RyanGrif, Thanks for the request.
> LGTM.
> Could you add the test to verify the behavior?

Hey @e5l, I've added tests for the DispatchAuthProvider as requested. They pass for jvm, js browser and node, but fail for posix - I get the error `[generateNonce] is not supported on iOS` which suggests that Digest auth just isn't designed to work with iOS. If that's right, is there a way I can skip these tests for posix? Apologies if this is confusing - I'm quite new to Kotlin and not at all familiar with this multiplatform build stuff!

e5l:
Merged

PR #1998, opened 2020-07-17
add direct access to payload in JWTCredential & JWTPrincipal

thiakil:
**Subsystem**
JWT Auth

**Motivation**
When interacting with `JWTCredential` & `JWTPrincipal`, to access the decoded JWT token, it's a bit clunky to need to do `credential.payload.<foo>` (or its variable based workarounds) - the classes are just wrappers to apply the Ktor interfaces

**Solution**
This PR simply adds an implementation of auth0-jwt's `Payload` interface by delegation.
It doesn't provide _much_ benefit, but the more fields accessed the more evident it becomes.

Before (snippet from the docs)
```
validate { credential ->
    if (credential.payload.audience.contains(jwtAudience)) JWTPrincipal(credential.payload) else null
}
```

After
```
validate { credential ->
    if (credential.audience.contains(jwtAudience)) JWTPrincipal(credential.payload) else null
}
```

### DIFF
--- a/ktor-features/ktor-auth-jwt/jvm/src/io/ktor/auth/jwt/JWTAuth.kt
+++ b/ktor-features/ktor-auth-jwt/jvm/src/io/ktor/auth/jwt/JWTAuth.kt
@@ -27,18 +27,115 @@ private val JWTAuthKey: Any = "JWTAuth"
 private val JWTLogger: Logger = LoggerFactory.getLogger("io.ktor.auth.jwt")
 
 /**
+ * Shortcut functions for standard registered [JWT Claims](https://tools.ietf.org/html/rfc7519#section-4.1)
+ */
+public interface JWTPayloadHolder {
+    /**
+     * The JWT payload
+     */
+    public val payload: Payload
+
+    /**
+     * Get the value of the "iss" claim, or null if it's not available.
+     *
+     * The "iss" (issuer) claim identifies the principal that issued the
+     * JWT.  The processing of this claim is generally application specific.
+     * The "iss" value is a case-sensitive string containing a StringOrURI
+     * value.  Use of this claim is OPTIONAL.
+     */
+    public val issuer: String? get() = payload.issuer
+
+    /**
+     * Get the value of the "sub" claim, or null if it's not available.
+     *
+     * The "sub" (subject) claim identifies the principal that is the
+     * subject of the JWT.  The claims in a JWT are normally statements
+     * about the subject.  The subject value MUST either be scoped to be
+     * locally unique in the context of the issuer or be globally unique.
+     * The processing of this claim is generally application specific.  The
+     * "sub" value is a case-sensitive string containing a StringOrURI
+     * value.  Use of this claim is OPTIONAL.
+     */
+    public val subject: String? get() = payload.subject
+
+    /**
+     * Get the value of the "aud" claim, or an empty list if it's not available.
+     *
+     * The "aud" (audience) claim identifies the recipients that the JWT is
+     * intended for.  Each principal intended to process the JWT MUST
+     * identify itself with a value in the audience claim.  If the principal
+     * processing the claim does not identify itself with a value in the
+     * "aud" claim when this claim is present, then the JWT MUST be
+     * rejected.  In the general case, the "aud" value is an array of case-
+     * sensitive strings, each containing a StringOrURI value.  In the
+     * special case when the JWT has one audience, the "aud" value MAY be a
+     * single case-sensitive string containing a StringOrURI value.  The
+     * interpretation of audience values is generally application specific.
+     * Use of this claim is OPTIONAL.
+     */
+    public val audience: List<String> get() = payload.audience ?: emptyList()
+
+    /**
+     * Get the value of the "exp" claim, or null if it's not available.
+     *
+     * The "exp" (expiration time) claim identifies the expiration time on
+     * or after which the JWT MUST NOT be accepted for processing.  The
+     * processing of the "exp" claim requires that the current date/time
+     * MUST be before the expiration date/time listed in the "exp" claim.
+     * Implementers MAY provide for some small leeway, usually no more than
+     * a few minutes, to account for clock skew. Use of this claim is OPTIONAL.
+     */
+    public val expiresAt: Date? get() = payload.expiresAt
+
+    /**
+     * Get the value of the "nbf" claim, or null if it's not available.
+     *
+     * The "nbf" (not before) claim identifies the time before which the JWT
+     * MUST NOT be accepted for processing.  The processing of the "nbf"
+     * claim requires that the current date/time MUST be after or equal to
+     * the not-before date/time listed in the "nbf" claim.  Implementers MAY
+     * provide for some small leeway, usually no more than a few minutes, to
+     * account for clock skew. Use of this claim is OPTIONAL.
+     */
+    public val notBefore: Date? get() = payload.notBefore
+
+    /**
+     * Get the value of the "iat" claim, or null if it's not available.
+     *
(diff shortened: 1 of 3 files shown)


cy6erGn0m:
The disadvantage of that decision is that it will tie us to the particular library forever. Even worse it will make it forever JVM that may hit us in the future when and if we will migrate to MPP.

cy6erGn0m:
Perhaps we just need to add shortcuts for easier components access

thiakil:
Good point. Should we add shortcut(s) for the non standard claims? I.e the methods which return a `Claim` object

thiakil:
I've made an attempt at doing this - though it's probably a breaking one.
Separated `Credential` and `Principal` don't make much sense for JWT anyway

wyaeld:
One observation, I'm just finishing up an integration with Microsoft Identity Platform on an internal Ktor app, and their library pulls in Nimbus for those objects, rather than Auth0 (which I've also used).

Links
https://connect2id.com/products/nimbus-oauth-openid-connect-sdk

https://connect2id.com/products/nimbus-jose-jwt

Usages
https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/dev/pom.xml#L38

https://github.com/spring-projects/spring-security/blob/master/oauth2/oauth2-jose/spring-security-oauth2-jose.gradle#L7

I'm in favour of convenience helpers, but expect the Nimbus libraries, which are used by Spring and Microsoft, are more widespread than Auth0

cy6erGn0m:
@thiakil So what is the current status? Do you think we need to improve something particular?

thiakil:
I made some changes to make things more generic back in July and heard nothing.

If you want them I can look at resolving the merge conflicts that have popped up since.

ktor-features/ktor-auth-jwt/jvm/src/io/ktor/auth/jwt/JWTAuth.kt
```diff
@@ -25,20 +25,6 @@ private val JWTAuthKey: Any = "JWTAuth"
 
 private val JWTLogger: Logger = LoggerFactory.getLogger("io.ktor.auth.jwt")
 
-/**
```
cy6erGn0m, line 28: Those couldn't be simply removed. A smooth migration is required. The new type `JWToken` should still have `payload` property so old validation code should keep working.

```
validate { jwt ->
    if (jwt.payload .....) jwt else ... // payload usage should lead to deprecation warning
}
```

The other issue is binary compatibilty. Most like we'll need few steps to migrate

ktor-features/ktor-auth-jwt/jvm/src/io/ktor/auth/jwt/JWTPrincipal.kt
```diff
@@ -0,0 +1,111 @@
...
+
+package io.ktor.auth.jwt
+
+import com.auth0.jwt.exceptions.*
+import com.auth0.jwt.interfaces.*
+import io.ktor.auth.*
+import java.util.*
+import kotlin.reflect.*
+
+interface JWToken: Credential, Principal {
```
cy6erGn0m, line 13: kdoc for interface

ktor-features/ktor-auth-jwt/jvm/src/io/ktor/auth/jwt/JWTPrincipal.kt
```diff
@@ -0,0 +1,111 @@
...
+    /**
+     * Get the value of the "iss" claim, or null if it's not available.
```
cy6erGn0m, line 15: Instead of referring to some field or something called claim, it's better to explain the meaning to the reader so one shouldn't look for the field "iss" somewhere in sourcecodes or google for it. We better to not refer to the particular implementation library.

ktor-features/ktor-auth-jwt/jvm/src/io/ktor/auth/jwt/JWTPrincipal.kt
```diff
@@ -0,0 +1,111 @@
...
+     */
+    fun <T: Any> getClaim(name: String, clazz: KClass<T>): T?
+
+    /**
+     * Attempt to locate a claim with the specified name, and deserialize it as List<[clazz]>
+     */
+    fun <T: Any> getListClaim(name: String, clazz: KClass<T>): List<T>?
+}
+
+inline fun <reified T: Any> JWToken.getClaim(name: String): T? {
```
cy6erGn0m, line 65: kdoc

ktor-features/ktor-auth-jwt/jvm/src/io/ktor/auth/jwt/JWTPrincipal.kt
```diff
@@ -0,0 +1,111 @@
...
+}
+
+inline fun <reified T: Any> JWToken.getListClaim(name: String): List<T>? {
+    return getListClaim(name, T::class)
+}
+
+/**
+ * JWT payload from auth0-jwt
+ */
+open class Auth0Jwt(private val payload: Payload) : JWToken {
```
cy6erGn0m, line 76: What is the reason to have it public and open? I would have it internal instead. What usecase for inheriting it?

cy6erGn0m (changes requested):
The most painful here is migration. Also, because of serious breaking changes, it couldn't be integrated before 2.0.0 that is not even planned yet. We definitely need a migration plan for this and, most likely, we'll need to make few iterations in 1.5.x or even 1.6.x

ktor-features/ktor-auth-jwt/jvm/src/io/ktor/auth/jwt/JWTPrincipal.kt
thiakil: `iss` and claim are part of the JWT specification, you want to explain that specification in kdoc?

ktor-features/ktor-auth-jwt/jvm/src/io/ktor/auth/jwt/JWTPrincipal.kt
thiakil: I suppose it's not strictly necessary in a kotlin world, but one could do so in order to cache non-standard claims from the token as fields instead of querying them every time

ktor-features/ktor-auth-jwt/jvm/src/io/ktor/auth/jwt/JWTAuth.kt
thiakil: Indeed the breakage is not great, and perhaps went too far from the initial goal of QoL changes

thiakil:
@cy6erGn0m I've re-done this in a non-breaking manner, and added a link to the RFC, and copied the RFC's description for the Claims.

e5l (approved):
LGTM, @cy6erGn0m WDYT? :)

e5l:
One little notice, @thiakil could you rebase it on the `1.6.0-eap`. We can merge new API in the next minor branch

cy6erGn0m (changes requested):
Generally it's fine. However, there's one thing to be improved. The problem is that we have interfaces here that are hard to migrate in case because we even can't add a property without major version. To improve it, we may think of abstract class or even sealed...

e5l:
@cy6erGn0m, totally agree. Abstract class with internal constructor would be great.

thiakil:
Ah, I had assumed Kotlin properties with implementations would translate to Java `default` interface methods

thiakil:
I'm not a fan of the internal constructor idea - it ruins the usability if someone wants their own `Principal` implementation with the helpers

e5l:
Good job! Merged

PR #2001, opened 2020-07-21
Revert breaking changes for 1.4 release

e5l:


ktor-client/ktor-client-features/ktor-client-json/jvm/src/io/ktor/client/features/json/DefaultJvm.kt
```diff
@@ -17,5 +17,5 @@ actual fun defaultSerializer(): JsonSerializer {
             " - ktor-client-serialization"
     )
 
-    return serializers.maxBy { it::javaClass.name }!!
+    return serializers.maxByOrNull { it::javaClass.name }!!
```
cy6erGn0m, line 20: This makes it kotlin 1.4.0 only, no compatibility with previously versions. Are we fine with it?

cy6erGn0m: Ah, or may be it's ok since it's inline

ktor-utils/jvm/src/io/ktor/util/cio/Semaphore.kt
```diff
@@ -0,0 +1,42 @@
+/*
+ * Copyright 2014-2020 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.util.cio
+
+import kotlinx.coroutines.sync.Semaphore
+
+@Deprecated(
+    "Ktor Semaphore is deprecated and will be removed in ktor 2.0.0. Consider using kotlinx.coroutines Semaphore instead.",
```
cy6erGn0m, line 10: Are we sure about 2.0.0? Isn't it better to no state exact version? I know we want to become more predictable. Are we going to do the same for other deprecations in the future?

ktor-utils/jvm/src/io/ktor/util/cio/Semaphore.kt
```diff
@@ -0,0 +1,42 @@
...
+    level = DeprecationLevel.WARNING,
+    replaceWith = ReplaceWith("Semaphore", "kotlinx.coroutines.sync.Semaphore")
+)
+class Semaphore(val limit: Int) {
+    private val delegate = Semaphore(limit)
+
+    @Deprecated(
+        "Ktor Semaphore is deprecated and will be removed in ktor 2.0.0. Consider using kotlinx.coroutines Semaphore instead.",
+        level = DeprecationLevel.WARNING,
+        replaceWith = ReplaceWith("acquire()", "kotlinx.coroutines.sync.Semaphore")
```
cy6erGn0m, line 20: What it will produce on existing code? Doesn't IDEA generate mess?

ktor-utils/jvm/src/io/ktor/util/cio/Semaphore.kt
e5l: Yes, we are. We're going to remove deprecations only in major updates. I see no reason not to say it publicly.

ktor-utils/jvm/src/io/ktor/util/cio/Semaphore.kt
e5l: Just checked. Everything works fine. Could you record the screencast if it doesn't work for you?

PR #2003, opened 2020-07-22
fix #1970 - update MultiPartFormDataContent to allow contentType override using optional builder

### LINKED ISSUE #1970: Objections to changing boundary to internal on MultiPartFormDataContent?

jschneidereit:
Fixes #1970 

**Subsystem**
Client

**Motivation**
Before this change `MultiPartFormDataContent` was not compatible with servers that accepted `multipart/mixed` content. 

**Solution**
Now a user can pass their own lambda for building content type like so:

```kotlin
val mixed = MultiPartFormDataContent(
    parts = listOf(),
    buildContentType = { b -> ContentType.MultiPart.Mixed.withParameter("boundary", b) }
)
```


ktor-client/ktor-client-core/common/src/io/ktor/client/request/forms/FormDataContent.kt
```diff
@@ -31,14 +31,18 @@ class FormDataContent(
...
 
 /**
  * [OutgoingContent] for multipart/form-data formatted request.
  *
  * @param parts: form part data
  */
 class MultiPartFormDataContent(
-    parts: List<PartData>
+    parts: List<PartData>,
+    buildContentType: (boundary: String) -> ContentType = buildFormDataContentType
```
e5l, line 45: Could we move `boundary` and `contentType` fields in the constructor?

e5l:
Hey @jschneidereit, sorry for the delay. Could you check one thing before the merge?

ktor-client/ktor-client-core/common/src/io/ktor/client/request/forms/FormDataContent.kt
jschneidereit: @e5l sure! pushed the fix!

jschneidereit:
> Hey @jschneidereit, sorry for the delay. Could you check one thing before the merge?

@e5l did I miss something?

ktor-client/ktor-client-core/common/src/io/ktor/client/request/forms/FormDataContent.kt
```diff
@@ -31,16 +31,16 @@ public class FormDataContent(
...
-
 /**
  * [OutgoingContent] for multipart/form-data formatted request.
  *
  * @param parts: form part data
  */
 public class MultiPartFormDataContent(
-    parts: List<PartData>
+    parts: List<PartData>,
+    val boundary: String = generateBoundary(),
```
e5l, line 41: Could you tell me why do we need `boundary` in the constructor?

e5l (changes requested):

ktor-client/ktor-client-core/common/src/io/ktor/client/request/forms/FormDataContent.kt
jschneidereit: Hi @e5l I had originally made it override-able but @cy6erGn0m requested an optional constructor argument in the discussion in #1970

jschneidereit: The other option to only having `contentType` in the constructor and not `boundary` would be a lambda that accepts a string like `contentTypeFn: (String) -> ContentType` but I think that's a little overkill.

e5l (approved):

PR #2004, opened 2020-07-22
Moshi content negotiation support for servers

veeti:
**Subsystem**
Server, new feature

**Motivation**
[Moshi](https://github.com/square/moshi) is a JSON serialization library for Java and Kotlin. Ktor core support for Moshi is desired because:

* Moshi can informally be considered to be ["Gson 3.0"](https://www.reddit.com/r/androiddev/comments/684flw/why_use_moshi_over_gson/dgx3gpm/) with various new features and design fixes.

* Unlike Gson or Jackson, Moshi has compile-time and reflection adapters for Kotlin including support for nullable types and default values.

* It's much lighter than Jackson having a size in kilobytes instead of megabytes.

I rebased this from old pull request #551.


cy6erGn0m:
Could you please claify benefits over kotlinx.serialization?

yogurtearl:
@cy6erGn0m moshi supports streaming so that it can start parsing in parallel while the request is still coming in.

Issue #875, opened 2019-01-15
JsonSerializer.write should also receive a type


ansman:
### Ktor Version
1.1.1

### Feedback
Currently `JsonSerializer.read` accepts a `TypeInfo` object which is great since it handles parameterized types. The `write` method does not which is unfortunate since it can lead to problematic cases.

If you consider the case where you pass `emptyList<String>()` the actual type is `EmptyList` which on some JSON libraries (for example Moshi) fails to return an adapter.

It can also lead to problems when using inheritance:
```kotlin
interface Named {
  val name: String
}

data class Person(val email: String, override val name: String) : Named

fun getName(): Named {
  return Person("info@example.com", "Nicklas")
}

client.post<SomeResponseType>("https://example.com") {
  // Since the return type of getName is Named we expect the JSON to be written to be
  // {"name": "Nicklas"} but it will instead be serialized as a Person
  body = getName()
}
```

A proposed solution could be to introduce a `bodyType` property of `HttpRequestBuilder` and add a `body` function that is reified and will get the type of the object as far as the call site is concerned.

e5l:
Hi @ansman, thanks for the report. 
Do you have the idea of how to pass body type info?

jbenoit2011:
This feature would be great since currently serialization of parameterized type is not possible since type parameter is erased at compile-time and inferred data type does not contain type parameter.

oleg-larshin:
Please check the <a href="https://youtrack.jetbrains.com/issue/KTOR-698">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

Issue #831, opened 2018-12-25
ContentConverter for list, map or other generic types


syhily:
### Ktor Version

1.0.1

### Ktor Engine Used(client or server and name)

Jetty Server

### JVM Version, Operating System and Relevant Context

1.8.0 Update 191, macOS 10.14.2

### Feedback

![image](https://user-images.githubusercontent.com/1761698/50422822-e6aa3680-0888-11e9-9d22-3da36afad60e.png)

I use `post<List<ConfigCorepage>>("/api/pageName/update")` to accept the json content from user request, but ktor couldn't recognize the type `List<ConfigCorepage>`, it use `List` instead.

I debug the code, and find that the [code](https://github.com/ktorio/ktor/blob/master/ktor-server/ktor-server-core/jvm/src/io/ktor/request/ApplicationReceiveFunctions.kt#L57) here shouldn't use `T::class`.

Jackson use a `TypeReference` class for preserving class info, ktor should use the same tricks.

syhily:
Could we move the https://github.com/ktorio/ktor/blob/master/ktor-client/ktor-client-core/common/src/io/ktor/client/call/TypeInfo.kt to upper level to let the server & client share the same typeinfo definition.

ansman:
Just as an FYI, this relates to both serializing and deserializing. If the user uses the type `List<String>` but passes `EmptyList` libraries like `Moshi` won't find an adapter for it so it would be great if the type could be explicitly specified rather than being inferred from the object that is sent.

respinha:
Will this be fixed? It can really stop you from using Jackson as a serializer.

oleg-larshin:
Please check the <a href="https://youtrack.jetbrains.com/issue/KTOR-707">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

e5l:
Fixed since Ktor 1.3.2

PR #2005, opened 2020-07-23
Add Moshi support

ZacSweers:
**Subsystem**
Json serialization

**Motivation**
This is a proposal PR with Moshi support. Moshi is a spiritual successor to Gson, faster, actively maintained, and has true first first-party support for Kotlin (both via reflect artifact or code gen). Two of its original authors are also two of the three maintainers of Gson as well.

**Solution**
Its implementation and use is mostly similar to the existing Gson implementation, but one significant caveat is that unlike gson it will not try to auto-magically fall back to Map/Set/List adapters if it receives concrete implementations.

Normally moshi handles this by requesting the explicit adapter type, but Ktor does not currently offer a way to indicate the encoding type for serialization (only offers the object, with encoders expected to introspect it via reflection). This is described in detail in #875 and [here](https://github.com/ktorio/ktor/issues/831#issuecomment-454179156). I could add custom logic for this like the `KotlinxSerializer` does, but that would then short-circuit any handling the `Moshi` instance itself has.


ktor-client/ktor-client-features/ktor-client-json/ktor-client-moshi/jvm/src/io/ktor/client/features/json/MoshiSerializer.kt
```diff
@@ -0,0 +1,30 @@
...
+import io.ktor.util.*
+import io.ktor.utils.io.core.*
+import okio.*
+import kotlin.reflect.*
+
+/**
+ * [JsonSerializer] using [Moshi] as backend.
+ */
+@ExperimentalStdlibApi
+class MoshiSerializer(block: Moshi.Builder.() -> Unit = {}) : JsonSerializer {
```
bradynpoulsen, line 20: Any reason to be inconsistent here by having a builder block passed to the client instead of a configured `Moshi` instance like the server implementation?

bradynpoulsen: Maybe a secondary constructor that accepts the builder block could allow it to be consistent and work with the block

```kotlin
constructor(block: Moshi.Builder.() -> Unit) : this(Moshi.Builder().apply(block).build())
```

ZacSweers: I just matched the Gson serializer as convention. I personally don't like this and would prefer it just accept a Moshi instance and let the user build it separately.

ktor-features/ktor-moshi/jvm/src/io/ktor/moshi/MoshiSupport.kt
```diff
@@ -0,0 +1,65 @@
...
+        return moshi.adapter<Any>(type).fromJson(source) ?: throw UnsupportedNullValuesException()
+    }
+}
+
+/**
+ * Register Moshi to [ContentNegotiation] feature
+ */
+fun ContentNegotiation.Configuration.moshi(
+    contentType: ContentType = ContentType.Application.Json,
+    block: Moshi.Builder.() -> Unit = {}
```
bradynpoulsen, line 47: This might be a good `inline` function with this `block` specified as `crossinline`

ktor-client/ktor-client-features/ktor-client-json/ktor-client-moshi/jvm/src/io/ktor/client/features/json/MoshiSerializer.kt
plannigan: A secondary constructor is what I did with [my implementation](https://github.com/plannigan/ktor-moshi/blob/dd3c55063ec598b71581d7db0d046c000dac4cfa/client/src/main/kotlin/com/hypercubetools/ktor/moshi/MoshiSerializer.kt#L13). It provides a bit more flexibility.

ZacSweers: Let's see what the ktor maintainers think

cy6erGn0m:
The main question here is: why do we want Mosi in ktor? What are benefits of Moshi over kotlinx.serialization and other backends? Could you please explain?

ZacSweers:
The best comparison is against Gson. Moshi is a spiritual successor to Gson, a sort of 3.0 that Gson would never have had on its own. Two of the three last Gson maintainers started Moshi. Gson today is largely unmaintained.

Compared to Gson, Moshi has:
* A smaller, better API
* Better defaults
* Better performance in most conventional conditions
* Actual first-class kotlin support in both reflection and code gen. Gson doesn't natively support kotlin in any capacity beyond standard java reflection.
* Actual, ongoing development
* Moshi 2.0 (starting now) will be written in Kotlin.
* Built on top of Okio, another kotlin-first (and now multiplatform) project

I would claim that if Gson has space as a first party supported option, so should Moshi on the basis of its similarities, Kotlin story, and ongoing development.

hhariri:
@ZacSweers We're moving Ktor into the direction of having most features [no longer part of core](https://youtrack.jetbrains.com/issue/KTOR-1142). In addition, we will be providing a market place for 3rd party features. While Moshi really does feel like a successor to Gson, it would probably contradict the direction we're taking somewhat. Makes sense?

ZacSweers:
Makes sense 👍

PR #2008, opened 2020-07-25
Prevent Empty Cookie addition

gincos:
**Subsystem**
Client.

**Motivation**
Some systems may behave differently if an empty cookie is sent, rather than omitting the Cookie header.

**Solution**
The header is only added if the Cookies list is not empty.


e5l:
Hey @gincos, thanks for the PR. LGTM! Merged :)

Issue #1891, opened 2020-05-22
TestHttpClientEngine KNPE


agcom:
**Ktor Version and Engine Used (client or server and name)**
1.3.2, test client/server, ktor-server-test-host

**Describe the bug**
Take this example.
```kotlin
fun main() {
    
    withTestApplication(Application::helloModule) {
        
        runBlocking {
    
            val res = client.get<HttpResponse>("NonExistenceRoute")
            
            println(res)
    
        }
        
    }
    
}

fun Application.helloModule() {
    
    routing {
        
        get {
            
            call.respond(HttpStatusCode.OK, "Hello")
            
        }
        
    }
    
}
```
Throws KNPE because route doesn't exists. If you follow stack trace, it's obvious where KNPE was thrown.
It's pretty nice to have this kind of tests. Is `TestHttpClientEngine` in experimental state? I also had another issue where the tests would stuck using multiparts.

**Expected behavior**
A response with 404 not found status.

**Screenshots**
Stack trace of the thrown exception,
```kotlin
Exception in thread "main" kotlin.KotlinNullPointerException
	at io.ktor.server.testing.client.TestHttpClientEngine.execute(TestHttpClientEngine.kt:37)
	at io.ktor.client.engine.HttpClientEngine$executeWithinCallContext$2.invokeSuspend(HttpClientEngine.kt:83)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:56)
	at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:571)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:738)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:678)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:665)
```

oleg-larshin:
Please check the <a href="https://youtrack.jetbrains.com/issue/KTOR-365">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

PR #2011, opened 2020-07-29
Ktor test client returns 404 if request not handled. Fixing NPE #1891

PokimonZerg:
**Subsystem**
Test Ktor Client
ktor-server-test-host

**Motivation**
Fixing NPE #1891 
We should never throw NPE

**Solution**
We should emulate real client behavor with test client engine. So, if request not handled we should return 404 like real ktor.

### DIFF
--- a/ktor-server/ktor-server-test-host/jvm/src/io/ktor/server/testing/client/TestHttpClientEngine.kt
+++ b/ktor-server/ktor-server-test-host/jvm/src/io/ktor/server/testing/client/TestHttpClientEngine.kt
@@ -31,15 +31,29 @@ class TestHttpClientEngine(override val config: TestHttpClientConfig) : HttpClie
     override val coroutineContext: CoroutineContext = dispatcher + clientJob
 
     override suspend fun execute(data: HttpRequestData): HttpResponseData {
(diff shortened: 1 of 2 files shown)


ktor-server/ktor-server-test-host/jvm/src/io/ktor/server/testing/client/TestHttpClientEngine.kt
```diff
@@ -31,15 +31,30 @@ class TestHttpClientEngine(override val config: TestHttpClientConfig) : HttpClie
...
+            with(testServerCall.response) {
+                HttpResponseData(
+                    status()!!, GMTDate(),
+                    headers.allValues(),
+                    HttpProtocolVersion.HTTP_1_1,
+                    ByteReadChannel(byteContent ?: byteArrayOf()),
+                    callContext()
+                )
+            }
+        }
```
cy6erGn0m, line 46: codestyle

PokimonZerg: fixed

e5l (approved):

PR #2014, opened 2020-07-31
Support RFC 7692 Compression Extensions for WebSocket

volkert-fastned:
**Subsystem**

**Motivation**

**Solution**
Describe your solution.

### DIFF
--- a/ktor-features/ktor-websockets/jvm/test/io/ktor/tests/websocket/ParserTest.kt
+++ b/ktor-features/ktor-websockets/jvm/test/io/ktor/tests/websocket/ParserTest.kt
@@ -20,6 +20,23 @@ class ParserTest {
 
         assertEquals(FrameType.TEXT, parser.frameType)
         assertTrue { parser.fin }
+        assertFalse { parser.rsv1 }
+        assertFalse { parser.mask }
+        assertEquals(5, parser.length)
+        assertTrue { parser.bodyReady }
+
+        assertEquals("Hello", buffer.decodeString())
+    }
+
+    @Test
+    fun testParserSimpleFrameWithRsv1Set() {
+        val buffer = bufferOf("0xC1 0x05 0x48 0x65 0x6c 0x6c 0x6f")
+        val parser = FrameParser()
+        parser.frame(buffer)
+
+        assertEquals(FrameType.TEXT, parser.frameType)
+        assertTrue { parser.fin }
+        assertTrue { parser.rsv1 }
         assertFalse { parser.mask }
         assertEquals(5, parser.length)
         assertTrue { parser.bodyReady }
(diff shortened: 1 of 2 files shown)


edvorg:
@volkert-fastned I have been looking at this problem recently and have found that there is already DEFLATE support existing in netty: https://netty.io/4.1/api/io/netty/handler/codec/http/websocketx/extensions/compression/WebSocketServerCompressionHandler.html and https://netty.io/4.0/api/io/netty/handler/codec/compression/package-summary.html but haven't found a way to use it with ktor.

Since ktor is based on netty, do you think we could somehow get it to work?

volkert-fastned:
> @volkert-fastned I have been looking at this problem recently and have found that there is already DEFLATE support existing in netty: https://netty.io/4.1/api/io/netty/handler/codec/http/websocketx/extensions/compression/WebSocketServerCompressionHandler.html and https://netty.io/4.0/api/io/netty/handler/codec/compression/package-summary.html but haven't found a way to use it with ktor.
> 
> Since ktor is based on netty, do you think we could somehow get it to work?



> @volkert-fastned I have been looking at this problem recently and have found that there is already DEFLATE support existing in netty: https://netty.io/4.1/api/io/netty/handler/codec/http/websocketx/extensions/compression/WebSocketServerCompressionHandler.html and https://netty.io/4.0/api/io/netty/handler/codec/compression/package-summary.html but haven't found a way to use it with ktor.
> 
> Since ktor is based on netty, do you think we could somehow get it to work?

Netty is only one of Ktor's supported transports. And wasn't Ktor supposed to also work with Kotlin/JS and Kotlin/Native? Let's try getting it to work with the `FrameParser` class first, since that's actual Ktor code.

Right now, I'm trying to figure out how best to process the protocol extension header in the HTTP request preceding the WebSocket protocol upgrade.

cy6erGn0m:
First of all, the ability to handle frame "extensions" should be introduced in frame parser. Unfortunately, we simply skip extension parts. Then a frame interceptor should be sintroduced and it should properly work with frame defragmentation coroutine.

volkert-fastned:
Thanks for the preliminary review and feedback, @cy6erGn0m. :slightly_smiling_face: I hope to have some time to work on this next week. By the way, weren't all GitHub issues slated to be closed and migrated to YouTrack? Or has that decision been reversed? No offense to JetBrains, but I'm kind of hoping for the latter.

e5l:
This feature was finished and released in `1.5.0`

volkert-fastned:
@e5l Thanks, and good work! :slightly_smiling_face:

Any idea why it wasn't mentioned in the release notes? It's certainly a feature that's relevant enough to mention there.

e5l:
It was here: https://ktor.io/changelog/1.5 and in the 1.5.0 blogpost

volkert-fastned:
@e5l Ah, I see, thanks. By the way, the YouTrack ticket mentioned that the fix version would be `1.6.0`: https://youtrack.jetbrains.com/issue/KTOR-688

Good to know you got it in earlier. :ok_hand:

PR #2016, opened 2020-08-07
Fix ByteBufferPool recycle

e5l:


ktor-io/jvm/src/io/ktor/utils/io/pool/ByteBufferPools.kt
```diff
@@ -0,0 +1,52 @@
...
+
+import io.ktor.utils.io.core.*
+import java.nio.*
+import java.nio.ByteOrder
+
+private const val DEFAULT_POOL_CAPACITY: Int = 2000
+
+private const val DEFAULT_BUFFER_SIZE: Int = 4096
+
+@ExperimentalIoApi
```
cy6erGn0m, line 15: Don't we want to make them internal (with annotation)?

ktor-utils/posix/src/io/ktor/util/CryptoNative.kt
```diff
@@ -4,18 +4,27 @@
...
+import platform.posix.*
+
 
 /**
  * Generates a nonce string. Could block if the system's entropy source is empty
  */
 @InternalAPI
-actual fun generateNonce(): String = error("[generateNonce] is not supported on iOS")
+actual fun generateNonce(): String {
+    val data = CharArray(16) { random().toChar() }
```
cy6erGn0m, line 16: Well, isn't it too dangerous to use `random()` function here?

ktor-utils/posix/src/io/ktor/util/internal/CRCDigest.kt
```diff
@@ -0,0 +1,28 @@
...
+import io.ktor.util.*
+import kotlinx.cinterop.*
+import platform.zlib.*
+
+internal class CRCDigest: Digest {
+    private var current = crc32(0, null, 0)
+
+    override fun plusAssign(bytes: ByteArray) {
+        bytes.asUByteArray().usePinned {
+            val buf  = it.addressOf(0)
```
cy6erGn0m, line 16: This will crash at an empty ByteArray

e5l: Yep, let's remove unrelated changes

ktor-io/jvm/src/io/ktor/utils/io/pool/ByteBufferPools.kt
e5l: We don't have such for IO, and we can introduce that. I think it's ok to leave it as Experimental, because someone may want to use it in the utility methods.

PR #2024, opened 2020-08-24
Fix client WebSocket maxFrameSize and timeout configuration

e5l:
Fix #KTOR-923


Stexxe:
The test, that will fail before the change and will pass after, is missing.

e5l:
Done, could you take a look?

ktor-client/ktor-client-core/jvm/test/WebSocketTest.kt
```diff
@@ -0,0 +1,42 @@
+import io.ktor.client.features.websocket.*
+import io.ktor.http.cio.websocket.*
+import kotlinx.coroutines.channels.*
+import kotlin.coroutines.*
+import kotlin.test.*
+
+/*
+ * Copyright 2014-2020 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
```
Stexxe, line 8: Should the license info be on top of the file?

ktor-client/ktor-client-core/jvm/test/WebSocketTest.kt
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+class WebSocketTest {
+
+    @Test
+    fun testAsDefault() {
+        val feature = WebSockets(42, 16)
+        val session = object : WebSocketSession {
```
Stexxe, line 16: What prevents the direct creation of `WebSocketSession`?

Stexxe (changes requested):

ktor-client/ktor-client-core/jvm/test/WebSocketTest.kt
e5l: It's an interface so we need connected channels to create any `real` instance. So I think it's enough to use mock for the test.

ktor-client/ktor-client-core/jvm/test/WebSocketTest.kt
e5l: Sorry, IDEA autoimport. Fixed

Stexxe (approved):

PR #2025, opened 2020-08-24
[KTOR-941] Add option to use only specific alias of KeyStore

hfhbd:
**Subsystem**
CIO - JVM

Fixing [KTOR-941](https://youtrack.jetbrains.com/issue/KTOR-941)

**Motivation**
This fix adds the option to use only a specific alias instead all alias for a given JVM Keystore.
This change is addictive only, no breaking change.
 
**Solution**
Added null-able default null alias which will be use if present. Otherwise, all aliases from a Keystore will be used as known behavior.


Stexxe (changes requested):
Please add a test

hfhbd:
Sorry, I rebased my pull request again -.- How can I prevent/fix it?

hfhbd:
@e5l @Stexxe Please review #2230
This PR has too many breaking rebased commits, so I simple create a new PR

PR #2026, opened 2020-08-24
[KTOR-940] CIO TLSConfigBuilder JVM allow null as password

hfhbd:
**Subsystem**
CIO JVM TLSConfigBuilder

Fixing [KTOR-940](https://youtrack.jetbrains.com/issue/KTOR-940)
**Motivation**
It is legal (and possible with Apache HttpClientEngine) to use a keystore without a password. One valid and secure use-case is using a pki/smartcard middleware which is loaded as a keystore, and enforces itself the password/the smartcard pin.

**Solution**
Made password nullable

### DIFF
--- a/ktor-network/ktor-network-tls/api/ktor-network-tls.api
+++ b/ktor-network/ktor-network-tls/api/ktor-network-tls.api
@@ -210,6 +210,8 @@ public final class io/ktor/network/tls/TLSConfigBuilderKt {
 	public static final synthetic fun addKeyStore (Lio/ktor/network/tls/TLSConfigBuilder;Ljava/security/KeyStore;[C)V
 	public static final fun addKeyStore (Lio/ktor/network/tls/TLSConfigBuilder;Ljava/security/KeyStore;[CLjava/lang/String;)V
 	public static synthetic fun addKeyStore$default (Lio/ktor/network/tls/TLSConfigBuilder;Ljava/security/KeyStore;[CLjava/lang/String;ILjava/lang/Object;)V
+	public static final fun addKeyStoreNullablePassword (Lio/ktor/network/tls/TLSConfigBuilder;Ljava/security/KeyStore;[CLjava/lang/String;)V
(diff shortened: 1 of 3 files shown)


Stexxe (changes requested):
Please add a test

e5l (changes requested):
LGTM, but we really need a test for that

hfhbd:
@e5l Unfortunately the JavaKeyStore requires a non-null password. Some other KeyStoreSpis, like the KeyStoreSpi from IT Solution GmbH/trustWare CSP+ v1.1.0.7 needs `null` as passwords, because they act as a middleware and en-/decrypt the keys with other passwords, like a PIN.
I did implement a fake KeyStoreSPI. Do you really want this?

hfhbd:
@cy6erGn0m Currently, TLSConfigBuilder is used only by CIO Client, so Server is the wrong label :)

hfhbd:
Switched base to master

e5l (approved):
Hey @hfhbd, thanks for the PR. Will be merged after CI run

hfhbd:
@e5l After reading Binary validator, this would be a breaking change, won't it? I will create a second overload method!

e5l:
@hfhbd that's right. Thanks for catch

hfhbd:
@e5l added overload. Do you want a deprecation warning too?

e5l:
Would be really great

hfhbd:
@e5l here you go

e5l:
It looks like there is a compilation problem:
```
e: Z:\buildAgent\work\8d547b974a7be21f\ktor-network\ktor-network-tls\jvm\test\io\ktor\network\tls\tests\TLSConfigBuilderTest.kt: (127, 63): Type mismatch: inferred type is String but Double was expected
```

Could you check?

hfhbd:
@e5l I was wondering, why this build only fails on Java 8, but not on Java 11. I am using the new API, introduced in Java 9, which deprecated the constructor in Java 8. In Java 15, it is still available.

hfhbd:
@e5l Could you rerun the tests?

e5l:
Yep, sure

e5l:
Good job! Merged

PR #2028, opened 2020-08-25
WIP: Adding client certificate requests

hfhbd:
**Subsystem**
CIO JVM TLSHandshake Clientrequest

Fixing: [KTOR-937](https://youtrack.jetbrains.com/issue/KTOR-937)

**Motivation**
Currently, CIO crashes when a server requests a client certificate and restricts the authority to some specific. CIO reads the size bytes for all the authorities, but did not add the size of each authorities to the position counter. As a result, CIO tries to read more bytes than available and crashes.

**Solution**
Add missing 2 Bytes to position after reading the size of authority


ktor-network/ktor-network-tls/jvm/src/io/ktor/network/tls/TLSClientHandshake.kt
```diff
@@ -242,7 +242,7 @@ internal class TLSClientHandshake(
                     var position = 0
                     while (position < authoritiesSize) {
                         val size = packet.readShort().toInt() and 0xFFFF
-                        position += size
+                        position += size + 2 // short = 2 bytes for reading size
```
Stexxe, line 245: Please express the intention for `+ 2` in the code instead of the comment

Stexxe (changes requested):
Add a test that fails before the change and passes after it.

ktor-network/ktor-network-tls/jvm/src/io/ktor/network/tls/TLSClientHandshake.kt
hfhbd: Changed :)

hfhbd:
> Add a test that fails before the change and passes after it.

@Stexxe I created a test. Where should it located?

Stexxe:
> > Add a test that fails before the change and passes after it.
> 
> @Stexxe I created a test. Where should it located?

I meant that you should write one test for your bug fix which fails without your change and passes with it. Please have a look at this [example](https://github.com/ktorio/ktor/blob/7a7144c9bab15533ffeef5f3337ed716610d35cf/ktor-network/ktor-network-tls/jvm/test/io/ktor/network/tls/tests/ConnectionTests.kt#L39). You can use the same location as of `certificateTest` for your test.

hfhbd:
@Stexxe Thanks for your help. But how do you create a webserver with requiring a client certificate in your example path, without adding dependencies to this module? At your example package, there is only the possibility to connect to a public server, which has to require a client request with limited authorities. This usage is very rare in public, but often used in internal networks, like my corporate network. So I think, it would be better to write the test inside a module, which has access to `embeddedServer` to create such a server in the test.

andrey-vasilyev:
Nice job! Was about to make a PR with the same fix.

ElbowBaggins:
@andrey-vasilyev you might want to anyway if it has tests in it

andrey-vasilyev:
> @andrey-vasilyev you might want to anyway if it has tests in it

This PR already has a test, isn't it? I actually have the test. But for it to work I had to refactor all parsing code into internal function and then test the function alone. That way (internal functions) it actually looks more like JDK's sun.security.ssl.CertificateRequest.T12CertificateRequestMessage (which I believe is a prototype of the ktor code). Basically I am not sure ktor team will like it that way. If the answer is yes, I'll be happy to contribute

hfhbd:
> > @andrey-vasilyev you might want to anyway if it has tests in it
> 
> This PR already has a test, isn't it? I actually have the test. But for it to work I had to refactor all parsing code into internal function and then test the function alone. That way (internal function) it actually looks more like JDK's sun.security.ssl.CertificateRequest.T12CertificateRequestMessage (which I believe is a prototype of the ktor code). Basically I am not sure ktor team will like it that way. If the answer is yes, I'll be happy to contribute

Hey Andrey,
I have the problem with the test. The tests @Stexxe mentions use public web servers. I dislike using external source in normal unit tests. These kind of tests are integrations tests, which should move into a separate test module.
So, to write my test, I also extend the code base to use the ktor web server. Fortunately the Jetty engine supports HTTPS Client Authentication Requests.

Unfortunately, IDEA does not find my test...

ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/EngineConnectorConfig.kt
```diff
@@ -75,7 +76,15 @@ interface EngineSSLConnectorConfig : EngineConnectorConfig {
...
-    val privateKeyPassword: () -> CharArray
+    public val privateKeyPassword: () -> CharArray
+
+    /**
+     * If not empty, the list of allowed CA certificates which will be used for client certificate authentication.
+     * If the list is empty, no client certification request will be requested.
+     *
+     * Currently only supported by Jetty Engine
+     */
+    val requiredClientCACertificates: List<X509Certificate>
```
andrey-vasilyev, line 87: Why not trustStore(key store) with the certificates? With key store parameter, one can use same key store for both server and client auth, which is most common I believe as client certificates are self signed by server owners

Actually I have already made an implementation for Netty engine but chose to provide keyStore instead of list of certificates. I can contribute the implementation to this PR

hfhbd: Good idea :)
Yes please. Help/contribution is always welcome.

andrey-vasilyev:
> Before I/we merge the fix with tests, we should discuss, what other changes we want in ktor. Then I would like to create final pull request with "better, smaller" commits

Hi. I would refactored certificate request parsing into separate internal function and write a test for that - little and easy change. Also quite easy to change it back - inline this new function or make it private. Like if we (or someone else) implements client certificates auth for all server engines and this code can be tested as part of some component/integration tests. I just would like to know what ktor team thinks about it

hfhbd:
@andrey-vasilyev 
I want to split this pull request:
1. Fixing KTOR-937, by adding 2 bytes to the size, in a new pull request
2. Adding client authentication to the server engines

andrey-vasilyev:
totally agree

hfhbd:
> totally agree

Do you want to review the new one ? :)

andrey-vasilyev:
sure

hfhbd:
Merged netty implementation and test by @andrey-vasilyev 
@e5l Should I recreate this pull request with less WIP commits?

hfhbd:
Created smaller pr in #2112

PR #2029, opened 2020-08-26
Fix nullable response breaking change

e5l:
Fix [KTOR-916](https://youtrack.jetbrains.com/issue/KTOR-916)

### DIFF
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/features/DefaultResponseValidation.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/features/DefaultResponseValidation.kt
@@ -47,7 +47,8 @@ fun HttpClientConfig<*>.addDefaultResponseValidation() {
 open class ResponseException(
     response: HttpResponse
 ) : IllegalStateException("Bad response: $response") {
-    public val response: HttpResponse? by threadLocal(response)
+    private val _response: HttpResponse? by threadLocal(response)
+    public val response: HttpResponse get() = _response ?: error("Failed to access response from a different native thread")
 }
 
 /**
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/features/DefaultResponseValidation.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/features/DefaultResponseValidation.kt
@@ -18,7 +18,7 @@ private val ValidateMark = AttributeKey<Unit>("ValidateMark")
  * Default response validation.
  * Check the response status code in range (0..299).
  */
-fun HttpClientConfig<*>.addDefaultResponseValidation() {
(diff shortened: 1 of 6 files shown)


ktor-client/ktor-client-core/common/src/io/ktor/client/features/DefaultResponseValidation.kt
```diff
@@ -47,7 +47,8 @@ fun HttpClientConfig<*>.addDefaultResponseValidation() {
 open class ResponseException(
     response: HttpResponse
 ) : IllegalStateException("Bad response: $response") {
-    public val response: HttpResponse? by threadLocal(response)
+    private val _response: HttpResponse? by threadLocal(response)
+    public val response: HttpResponse get() = _response ?: error("Failed to access response from a different native thread")
```
Stexxe, line 51: How do you know that `null` in this case means that was a failed attempt of access from a different native thread?

ktor-client/ktor-client-core/common/src/io/ktor/client/features/DefaultResponseValidation.kt
```diff
@@ -47,7 +47,8 @@ fun HttpClientConfig<*>.addDefaultResponseValidation() {
 open class ResponseException(
     response: HttpResponse
 ) : IllegalStateException("Bad response: $response") {
-    public val response: HttpResponse? by threadLocal(response)
+    private val _response: HttpResponse? by threadLocal(response)
```
Stexxe, line 50: Why do we need to have this additional property?

Stexxe (changes requested):
We probably should have a test checking that accessing `response` multiple times leads to the same reference.

ktor-client/ktor-client-core/common/src/io/ktor/client/features/DefaultResponseValidation.kt
e5l: It has thread-local behavior only in the case of kotlin-native

ktor-client/ktor-client-core/common/src/io/ktor/client/features/DefaultResponseValidation.kt
e5l: Could you tell me how we can replace it? We can't write `?:` inside `threadLocal` delegate to get `non-null` type.

ktor-client/ktor-client-core/common/src/io/ktor/client/features/DefaultResponseValidation.kt
Stexxe: Then it would be great to document this assumption

Stexxe (approved):

PR #2030, opened 2020-08-26
socketTimeout in EndpointConfig was read only

kimble:
**Subsystem**
Client 

**Motivation**
I assume this is a typo. 

**Solution**
Make configuration option writable.


e5l:
Hey, @kimble. Thanks for the notice. LGTM

PR #2031, opened 2020-08-26
Serialize/deserialize outside of the main thread

wkornewald:
**Subsystem**
Client / JsonFeature

**Motivation**
Depending on the request size and the underlying serialization (which could e.g. also do encryption/decryption) we might be blocking the main thread for too long.

**Solution**
In order to stay with the abstraction that Ktor takes care of thread switching internally, this PR moves serialization/deserialization out of the main thread.


wkornewald:
@e5l I hope that's the correct way to solve it. Is there any other dispatcher I can use?

PR #2033, opened 2020-08-28
Fix parsing of url with space at start

e5l:
Fix #KTOR-886

### DIFF
--- a/ktor-http/common/src/io/ktor/http/URLParser.kt
+++ b/ktor-http/common/src/io/ktor/http/URLParser.kt
@@ -157,7 +157,7 @@ private fun findScheme(urlString: String, startIndex: Int, endIndex: Int): Int {
                 throw IllegalArgumentException("Illegal character in scheme at position $incorrectSchemePosition")
             }
 
-            return current
+            return current - startIndex
         }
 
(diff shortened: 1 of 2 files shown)


ktor-http/common/src/io/ktor/http/URLParser.kt
```diff
@@ -157,7 +157,7 @@ private fun findScheme(urlString: String, startIndex: Int, endIndex: Int): Int {
                 throw IllegalArgumentException("Illegal character in scheme at position $incorrectSchemePosition")
             }
 
-            return current
+            return current - startIndex
```
Ololoshechkin, line 160: I cannot comment on the old code so I'll leave it here.
I would personally do the following on the top of this function:
a) Comment that `startIndex` and `endIndex` are indexes of the first and last spaces in urlString. Because otherwise it's not quite obvious that it fixes the test case (for sure, it does, because I've checked the usage of this function. But I'll leave an explicit comment).
b) Or maybe just rename `startIndex` and `endIndex` to something more explanative.

But still feel free to ignore as soon as it's not connected to the current commit

Ololoshechkin (approved):

PR #2034, opened 2020-08-29
KTOR-1263 Allow any header in CORS configuration

jozefdransfield:
**Subsystem**
Server

**Motivation**
I would like to support custom headers in a rest api called from a web browser, that I cannot know at application load.
I would therefore like to be able to mirror the allowed headers as requested in the preflight call, I believe this functionality is implemented in existing frameworks: for example expressjs
https://expressjs.com/en/resources/middleware/cors.html
"allowedHeaders: Configures the Access-Control-Allow-Headers CORS header. Expects a comma-delimited string (ex: ‘Content-Type,Authorization’) or an array (ex: ['Content-Type', 'Authorization']). If not specified, defaults to reflecting the headers specified in the request’s Access-Control-Request-Headers header."

**Solution**
Attempted to replicate the way anyHost works to extend this to anyHeader.

[KTOR-1263](https://youtrack.jetbrains.com/issue/KTOR-1263)


ktor-server/ktor-server-tests/jvm/test/io/ktor/tests/server/features/CORSTest.kt
```diff
@@ -670,4 +670,29 @@ class CORSTest {
...
+                anyHeader()
+            }
+
+            application.routing {
+                get("/") {
+                    call.respond("OK")
+                }
+            }
+
+            handleRequest(HttpMethod.Options, "/") {
```
cy6erGn0m, line 688: Two cases need to be tested including missing `AccessControlRequestHeaders` in the request.

cy6erGn0m:
The main question here is security. Allowing any headers may lead to serious vulnerabilities. What are observations of expressjs community? Any related reports?

cy6erGn0m:
Please run `./gradlew apiDump` and commit the change in server-core module

cy6erGn0m:
The feature only looks similar to any host. While the purpose is quite different. Any host is required for public API services (e.g. GitHub API) that want to provide the ability to make requests from any hosts. If API is secured properly, it is relatively safe to allow any hosts.
On the other hand, allowing any headers looks like some kind debug/testing feature and opens potentially unplanned capabilities with unknown effects. Such a "too open" strategy seems to be too dangerous because one can't plan all possible incoming headers because there are too many different headers with different semantics. This is why in CORS the list of headers is designed as a white list while all other headers are forbidden.

What a shame that CORS doesn't support partial wildcards in header names.

jozefdransfield:
My use case is I am using custom headers to exchange meta data about a resource. (e.g PU/ GET /POST) 

All my headers are prefixed so I know which are specific to the data.

Perhaps a solution could be to emulate a wildcard matcher, where you say you allow headers that match prefix-* then you could take the Access-Control-Allow-Headers and mirror the ones that match the pattern, plus any headers that where explicitly configured?

jozefdransfield:
@cy6erGn0m Should I investigate the solution I proposed?

cy6erGn0m:
@jozefdransfield Hi, Jozef. Sorry for the delay. I think yes, we may try the wildcard solution.

jozefdransfield:
@cy6erGn0m Made those changes now.

cy6erGn0m (changes requested):
Thanks! This is way better. But what if we need to match two or more different prefixes? Not sure, perhaps we need a list of prefixes or even a list of functional predicates...  or maybe both

```kotlin
allowHeadersPrefixed("X-My-")
allowHeaders { name -> name.startsWith("X-My-") }
```

The disadvantage of lambda is that it encurages `allowHeaders { true }` or something like that

jozefdransfield:
@cy6erGn0m I have added an allowHeaders to take a predicate for matching the header names and it now accepts multiple conditions so you can add more than one prefix etc.

ktor-server/ktor-server-core/jvm/src/io/ktor/features/CORS.kt
```diff
@@ -128,9 +138,12 @@ public class CORS(configuration: Configuration) {
         if (methodsListHeaderValue.isNotEmpty()) {
             response.header(HttpHeaders.AccessControlAllowMethods, methodsListHeaderValue)
         }
-        if (headersListHeaderValue.isNotEmpty()) {
-            response.header(HttpHeaders.AccessControlAllowHeaders, headersListHeaderValue)
-        }
+
+        val requestHeadersMatchingPrefix = requestHeaders.filter { header -> headerMatchesAPredicate(header) }
+
+        val headersListHeaderValue = listOf(headersList, requestHeadersMatchingPrefix).flatten().sorted().joinToString(", ")
```
cy6erGn0m, line 144: Can't we simply do `+` instead of flattening?

ktor-server/ktor-server-core/jvm/src/io/ktor/features/CORS.kt
```diff
@@ -42,6 +43,11 @@ public class CORS(configuration: Configuration) {
             if (configuration.allowNonSimpleContentTypes) headers else headers.minus(HttpHeaders.ContentType)
         }
 
+    /**
+     * Prefix for permitted headers
+     */
+    public val headerPredicates : List<Predicate<String>> = configuration.headerPredicates
```
cy6erGn0m, line 49: Depending on Java-only `Predicate` is unnecessary. Can't we replace it with Kotlin function `(String) -> Boolean`? It looks a bit more verbose but it will let us migrate out of JVM in case..

cy6erGn0m (changes requested):
Thanks! I think it is the last iteration

jozefdransfield:
@cy6erGn0m No problem, made those changes

cy6erGn0m (approved):

cy6erGn0m:
Merged manually. Thanks for your contribution!

PR #2036, opened 2020-08-29
Jackson: Do not convert to String before converting to JSON

antonha:
**Subsystem**
ktor-jackson

**Motivation**
The existing implementation of the JacksonConverter used a TextContent, which meant that the ObjectMapper had to serialize the entire output value to a String before responding.

For large output objects, this causes significant allocation pressure. It can avoid by instead streaming the output. This change does that.

**Solution**

The change is fairly small, simply using different existing streaming methods on the ObjectMapper in combination with a OutputStreamContent.

There is an added test to verify that it got UTF-16 conversion right.


cy6erGn0m (approved):

cy6erGn0m:
Created ticket [KTOR-1186](https://youtrack.jetbrains.com/issue/KTOR-1186). Waiting for build.

PR #2037, opened 2020-08-31
Skip invalid cookies instead of failing

e5l:
fixes #KTOR-860

### DIFF
--- a/ktor-client/ktor-client-apache/jvm/src/io/ktor/client/engine/apache/ApacheHttpRequest.kt
+++ b/ktor-client/ktor-client-apache/jvm/src/io/ktor/client/engine/apache/ApacheHttpRequest.kt
@@ -36,13 +36,18 @@ internal suspend fun CloseableHttpAsyncClient.sendRequest(
         val statusLine = rawResponse.statusLine
 
         val status = HttpStatusCode(statusLine.statusCode, statusLine.reasonPhrase)
-        val version = with(rawResponse.protocolVersion) { HttpProtocolVersion.fromValue(protocol, major, minor) }
-        val headers = Headers.build {
-            rawResponse.allHeaders.forEach { headerLine ->
-                append(headerLine.name, headerLine.value)
-            }
+        val version = with(rawResponse.protocolVersion) {
+            HttpProtocolVersion.fromValue(protocol, major, minor)
         }
 
+        val rawHeaders = rawResponse.allHeaders.filter {
+            it.name != null || !it.name.isBlank()
+        }.groupBy(
+            { it.name },
+            { it.value ?: "" }
+        )
+
+        val headers = HeadersImpl(rawHeaders)
         return HttpResponseData(status, requestTime, headers, version, consumer.responseChannel, callContext)
     } catch (cause: Exception) {
         future.cancel(true)
--- a/ktor-client/ktor-client-cio/common/src/io/ktor/client/engine/cio/utils.kt
+++ b/ktor-client/ktor-client-cio/common/src/io/ktor/client/engine/cio/utils.kt
@@ -13,7 +13,9 @@ import io.ktor.http.cio.*
 import io.ktor.http.content.*
 import io.ktor.util.date.*
 import io.ktor.utils.io.*
+import io.ktor.utils.io.core.*
 import io.ktor.utils.io.errors.*
+import io.ktor.utils.io.errors.EOFException
 import kotlinx.coroutines.*
 import kotlin.coroutines.*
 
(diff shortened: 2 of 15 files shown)


ktor-http/ktor-http-cio/common/src/io/ktor/http/cio/HttpParser.kt
```diff
@@ -258,7 +264,7 @@ internal fun parseHeaderValue(text: CharArrayBuilder, range: MutableRange) {
         when {
             ch == HTAB || ch == ' ' -> {
             }
-            ch < ' ' -> characterIsNotAllowed(text, ch)
+            ch < ' ' -> skip = true
```
cy6erGn0m, line 267: This could potentially lead to security vulnerabilities since we simply skip all "strange" characters instead of rejecting and stopping processing the HTTP pipeline. Perhaps, we need to handle particular character codes mode precisely

ktor-http/ktor-http-cio/common/src/io/ktor/http/cio/HttpParser.kt
```diff
@@ -116,14 +116,16 @@ internal suspend fun parseHeaders(
...
 
             val valueStart = range.start
             val valueEnd = range.end
             val valueHash = builder.hashCodeLowerCase(valueStart, valueEnd)
             range.start = headerEnd
 
-            headers.put(nameHash, valueHash, nameStart, nameEnd, valueStart, valueEnd)
+            if (!shouldSkipHeader) {
+                headers.put(nameHash, valueHash, nameStart, nameEnd, valueStart, valueEnd)
+            }
```
cy6erGn0m, line 128: We need some kind diagnostics here, like logging on trace level. Not sure what logger we could use here....

cy6erGn0m (changes requested):
I believe we can't simply skip all undesired characters as it is too dangeruous

ktor-http/ktor-http-cio/common/src/io/ktor/http/cio/HttpParser.kt
e5l: Could you explain what exact cases do you mean here? Cause we're excluding all parts with forbidden symbols from output

cy6erGn0m: What I fear is some kind of split header issue. A surprising `\n` character or something that may break HTTP pipeline processing and provide a way of spoofing

e5l: But if we have `\n` we're proceeding to check the next header

cy6erGn0m: Ah I see, the line is already delimited so we can't face \r\n here. But what about the RFC? It clearly states that the field content is a VCHAR (0x21..0x7e) or a obs-text (0x80+). Why do we want to allow such a violation?

cy6erGn0m:
The other question: according to the ticket, any engine is affected but we only fix CIO. Don't other clients also forbid such bad bytes?

ktor-http/ktor-http-cio/common/src/io/ktor/http/cio/HttpParser.kt
e5l: Because there were multiple RFC about encoding header value and there are servers that don't encode such symbols(for cookies for example). CURL and other HTTP clients are handling that in the same way

e5l:
Sure, let's add an integration test as well

e5l:
The general idea of the fix
 - we keep any characters as is during parsing
 - expose parsing result in client as is

We still check if you're allowed to use such value in header during request building, so it's not symmetric.

ktor-client/ktor-client-apache/jvm/src/io/ktor/client/engine/apache/ApacheHttpRequest.kt
```diff
@@ -36,13 +36,16 @@ internal suspend fun CloseableHttpAsyncClient.sendRequest(
...
-        val headers = Headers.build {
-            rawResponse.allHeaders.forEach { headerLine ->
-                append(headerLine.name, headerLine.value)
-            }
+        val version = with(rawResponse.protocolVersion) {
+            HttpProtocolVersion.fromValue(protocol, major, minor)
         }
 
+        val rawHeaders = rawResponse.allHeaders.groupBy(
+            { it.name ?: "" },
```
cy6erGn0m, line 44: ❓ I wonder what is an empty header name and what semantics does it have? Some clients use `null` or `""` as a special value for the status line.

ktor-client/ktor-client-tests/jvm/src/io/ktor/client/tests/utils/tests/Proxy.kt
```diff
@@ -54,11 +59,23 @@ private fun buildResponse(
     status: HttpStatusCode,
     headers: Headers = Headers.Empty,
     body: String = "proxy"
+): String {
+    val headersList: List<String> = headers.toMap().entries.toList().map {  (key, values) ->
```
cy6erGn0m, line 63: format (space)

ktor-http/ktor-http-cio/jvm/src/io/ktor/server/cio/backend/ServerPipeline.kt
```diff
@@ -87,7 +87,7 @@ public fun CoroutineScope.startServerConnectionPipeline(
 
             val response = ByteChannel()
 
-            val transferEncoding = request.headers["Transfer-Encoding"]
+            val transferEncoding = request.headers["Transfer-Encoding"]?.trim()
```
cy6erGn0m, line 90: `trim()` is relatively slow (because of the additional allocation and unicode checks). Can't we avoid it? Like by tuning `expectHttpBody` to do case-insensitive indexOf instead of equals?

ktor-client/ktor-client-cio/common/src/io/ktor/client/engine/cio/utils.kt
```diff
@@ -102,38 +104,52 @@ internal suspend fun readResponse(
...
-    }
+        val body = when {
+            request.method == HttpMethod.Head ||
+                status in listOf(HttpStatusCode.NotModified, HttpStatusCode.NoContent) ||
+                status.isInformational() -> {
+                ByteReadChannel.Empty
+            }
+            else -> {
+                val httpBodyParser = GlobalScope.writer(Dispatchers.Unconfined, autoFlush = true) {
+                    parseHttpBody(contentLength, transferEncoding, connectionType, input, channel)
```
cy6erGn0m, line 129: A little bit "slippery" here since we pass pooled values here to a launched coroutine and the only hope is that it is unconfined. How was release done in before this change? Was the response object released at the context completion? Wasn't it the issue before this change too since the coroutine is running outside of the hierarchy?
❓  Also, we have the same issue with potential spaces in the header so should we fix it everywhere or we give up and don't care.

ktor-client/ktor-client-cio/common/src/io/ktor/client/engine/cio/utils.kt
```diff
@@ -102,38 +104,52 @@ internal suspend fun readResponse(
...
-                parseHttpBody(contentLength, transferEncoding, connectionType, input, channel)
+                httpBodyParser.channel
             }
+        }
+
+        return HttpResponseData(status, requestTime, headers, version, body, callContext)
+    }
+}
+
+internal fun HttpHeadersMap.toMap(): Map<String, List<String>> {
```
cy6erGn0m, line 140: ❓ so we are doing like this to avoid header name/value checks, right?

ktor-http/ktor-http-cio/common/test/io/ktor/tests/http/cio/HttpParserTest.kt
```diff
@@ -33,6 +35,23 @@ class HttpParserTest {
...
+        val rawHeaders = """
+            Set-Cookie: ___utmvazauvysSB=kDuxSkE; path=/; Max-Age=900
+        """.trimIndent() + "\r\n\r\n"
+
+        val channel = ByteReadChannel(rawHeaders)
+        val headers = parseHeaders(channel)
+
+        try {
+            val actual = headers[HttpHeaders.SetCookie].toString()
+            assertEquals("___utmvazauvysSB=kDu\u0001xSkE; path=/; Max-Age=900", actual)
```
cy6erGn0m, line 49: Does it pass? I see no `\u0001` in the raw source text

cy6erGn0m (changes requested):
So, by having it asymmetric, we allow receiving suspicious values but forbid sending them. So the question is: if we receive a cookie with a bad character, then how do we send it to the server in a further request?

ktor-http/ktor-http-cio/common/src/io/ktor/http/cio/internals/Tokenizer.kt
```diff
@@ -19,7 +19,7 @@ internal fun skipSpacesAndHorizontalTabs(
     var index = start
     while (index < end) {
         val ch = text[index]
-        if (ch != ' ' && ch != HTAB) break
+        if (!ch.isWhitespace() && ch != HTAB) break
```
cy6erGn0m, line 22: Isn't it too slow to use (it is much slower that just cmpeq because of double table lookup and indirection)? Why should we ignore all unicode whitespaces here?

cy6erGn0m (changes requested):

cy6erGn0m (approved):

PR #2038, opened 2020-08-31
Fix netty resource leak

e5l:
[`#KTOR-939`](https://youtrack.jetbrains.com/issue/KTOR-939)

### DIFF
--- a/ktor-server/ktor-server-netty/jvm/test/io/ktor/tests/server/netty/NettySpecificTest.kt
+++ b/ktor-server/ktor-server-netty/jvm/test/io/ktor/tests/server/netty/NettySpecificTest.kt
@@ -1,0 +1,19 @@
+/*
+ * Copyright 2014-2020 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.tests.server.netty
+
+import io.ktor.server.engine.*
+import io.ktor.server.netty.*
+import org.junit.*
+
+class NettySpecificTest {
+
+    @Test
+    fun testNoLeakWithoutStartAndStop() {
+        repeat(100000) {
+            embeddedServer(Netty, applicationEngineEnvironment {  })
+        }
+    }
+}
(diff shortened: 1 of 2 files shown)


gradle.properties
```diff
@@ -6,12 +6,14 @@ version=1.4.0
...
 kotlin.incremental.multiplatform=true
 kotlin.parallel.tasks.in.project=true
-org.gradle.parallel=true
 kotlin.native.ignoreIncorrectDependencies=true
 #kotlin.native.enableParallelExecutionCheck=false
 
 # gradle
+org.gradle.parallel=true
 org.gradle.daemon=true
+#org.gradle.configureondemand=true
```
cy6erGn0m, line 15: irrelevant change

cy6erGn0m:
Am I right that it's about unsarted servers? Would be nice to write a brief description here in PR for the future history

e5l:
Sure. I linked the issue in the description and commit message

cy6erGn0m:
The only side effect I can imagine is that concurrent start and stop could lead to undefined behaviour due to concurrent creating and disposing separate components.

e5l:
I'm not sure if we should support concurrent `start` and `stop` call. Short lookup shows that also we don't support it in other engines

ktor-server/ktor-server-netty/jvm/test/io/ktor/tests/server/netty/NettySpecificTest.kt
```diff
@@ -0,0 +1,19 @@
+/*
+ * Copyright 2014-2020 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.tests.server.netty
+
+import io.ktor.server.engine.*
+import io.ktor.server.netty.*
+import org.junit.*
```
cy6erGn0m, line 9: kotlin.test?

ktor-server/ktor-server-netty/jvm/test/io/ktor/tests/server/netty/NettySpecificTest.kt
```diff
@@ -0,0 +1,19 @@
...
+
+class NettySpecificTest {
```
cy6erGn0m, line 11: Class name?

e5l: That's a good name for the test, but not for the suite I guess

ktor-server/ktor-server-netty/jvm/test/io/ktor/tests/server/netty/NettySpecificTest.kt
e5l: +

cy6erGn0m (approved):

PR #2039, opened 2020-08-31
Fix ktor client cookie encoding

e5l:
Fix [`#KTOR-917`](https://youtrack.jetbrains.com/issue/KTOR-917)

### DIFF
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/features/cookies/AcceptAllCookiesStorage.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/features/cookies/AcceptAllCookiesStorage.kt
@@ -7,16 +7,18 @@ package io.ktor.client.features.cookies
 import io.ktor.http.*
 import io.ktor.util.*
 import io.ktor.util.date.*
+import io.ktor.util.withLock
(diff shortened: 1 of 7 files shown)


ktor-client/ktor-client-core/common/src/io/ktor/client/features/cookies/AcceptAllCookiesStorage.kt
```diff
@@ -39,9 +41,7 @@ class AcceptAllCookiesStorage : CookiesStorage {
         }
     }
 
-    override fun close() {
-        mutex.close()
-    }
+    override fun close() {}
```
cy6erGn0m, line 44: So coroutines' lock does pooling for mutex_t so no need to terminate it explicitly, right? 

codestyle?

cy6erGn0m (approved):

ktor-client/ktor-client-core/common/src/io/ktor/client/features/cookies/AcceptAllCookiesStorage.kt
e5l: Yep. Thanks will fix

PR #2041, opened 2020-09-02
KTOR-792 - Fix priority of routing with query parameter selector

rsinukov:
**Subsystem**
Server

**Motivation**
[Routing: get matcher has higher priority than param matcher on the same level
](https://youtrack.jetbrains.com/issue/KTOR-792)

**Solution**
Original problem happens because `ParameterRouteSelector` uses the same priority (quality) as `PathSegmentParameterRouteSelector`, but they should not be the same

Example:

1. path: `/users/admin`
selectors:
* `/users/admin` - `qualityConstant` (1.0)
* `/users/{username}` - `qualityParameter` (0.8)
in this case chosen selector is `/users/admin` because it's more specific - everything works as expected

2. path: `/users?username=admin`
selectors:
* `users?[username]` - `qualityParameter` (0.8)
* `users:get` - `qualityConstant` (1.0)
Here selector `users:get` wins, but is should be the opposite. Route with parameter is more specific and should have higher priority.

To make fix without api changes I introduces `internal` constant `qualityQueryParameter = 1.0` and use it for routing with query parameter. Downside of it - now we have `qualityQueryParameter` and `qualityParameter` and this  can be confusing.
 
In the follow up PR I will make `qualityQueryParameter` public,  create new constant `qualityPathParameter` and deprecate `qualityParameter` in favour of the new one. This will introduce public API change and can't be done in patch release.

### DIFF
--- a/ktor-server/ktor-server-core/jvm/src/io/ktor/routing/RouteSelector.kt
+++ b/ktor-server/ktor-server-core/jvm/src/io/ktor/routing/RouteSelector.kt
@@ -29,6 +29,11 @@ data class RouteSelectorEvaluation(val succeeded: Boolean,
         const val qualityConstant = 1.0
 
         /**
+         * Quality of [RouteSelectorEvaluation] when a query parameter has matched
+         */
+        internal const val qualityQueryParameter = 1.0
+
+        /**
          * Quality of [RouteSelectorEvaluation] when a parameter has matched
          */
         const val qualityParameter = 0.8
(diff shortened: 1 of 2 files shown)


ktor-server/ktor-server-tests/jvm/test/io/ktor/tests/server/routing/RoutingProcessingTest.kt
```diff
@@ -89,6 +89,77 @@ class RoutingProcessingTest {
 
     }
 
+    enum class SelectedRoute { Get, Param, None }
```
e5l, line 92: Can be private, consider moving it outside of the class

ktor-server/ktor-server-core/jvm/src/io/ktor/routing/RouteSelector.kt
```diff
@@ -28,6 +28,11 @@ data class RouteSelectorEvaluation(val succeeded: Boolean,
          */
         const val qualityConstant = 1.0
 
+        /**
+         * Quality of [RouteSelectorEvaluation] when a query parameter has matched
+         */
+        internal const val qualityQueryParameter = 1.0
```
e5l, line 34: Should we order selection between `qualityConstant` and `qualityQueryParameter`?

Could you write the test case with both on the same level?

e5l (changes requested):
LGTM in general, but have some things to check

e5l (approved):

e5l:
LGTM, feel free to squash and merge after CI build

Issue #482, opened 2018-07-17
MultiPartData.readAllParts() throws "java.io.IOException: Broken delimiter occurred" with empty multipart list


SebastianAigner:
**Problem description**: When receiving an empty multipart-form request on a POST route, the request handler throws an IOException, preventing graceful handling or the display of an error message. Instead, a 500 error message is returned.

Minimum working example:
```kotlin
import io.ktor.application.*
import io.ktor.content.readAllParts
import io.ktor.response.*
import io.ktor.request.*
import io.ktor.routing.post
import io.ktor.routing.routing

fun main(args: Array<String>): Unit = io.ktor.server.netty.DevelopmentEngine.main(args)

fun Application.module() {
    routing {
        post("/add") {
            if(call.request.isMultipart()) {
                val multipart = call.receiveMultipart()
                val formItems = multipart.readAllParts() // <-- this breaks when sent an empty request
                call.respondText("recvd: $multipart")
            }
            else {
                call.respond("no multipart, fix request!")
            }
        }
    }
}
```

Send a request with an empty multipart body to see the issue:
```
curl -X POST \
  http://localhost:8080/add \
  -H 'Cache-Control: no-cache' \
  -H 'content-type: multipart/form-data; boundary=----WebKitFormBoundary7MA4YWxkTrZu0gW'
```

Error Message:
```
22:57:11.866 [nettyCallPool-4-4] ERROR Application - Unhandled: POST - /add
java.io.IOException: Broken delimiter occurred
	at kotlinx.coroutines.experimental.io.DelimitedKt$skipDelimiterSuspend$2.doResume(Delimited.kt:57)
	at kotlinx.coroutines.experimental.io.DelimitedKt$skipDelimiterSuspend$2.invoke(Delimited.kt)
	at kotlinx.coroutines.experimental.io.DelimitedKt$skipDelimiterSuspend$2.invoke(Delimited.kt)
	at kotlinx.coroutines.experimental.io.ByteBufferChannel.lookAheadSuspend(ByteBufferChannel.kt:1746)
	at kotlinx.coroutines.experimental.io.DelimitedKt.skipDelimiterSuspend(Delimited.kt:55)
	at kotlinx.coroutines.experimental.io.DelimitedKt.skipDelimiter(Delimited.kt:50)
	at io.ktor.http.cio.MultipartKt.boundary(Multipart.kt:89)
	at io.ktor.http.cio.MultipartKt$parseMultipart$1.doResume(Multipart.kt:158)
[...]

PR #2043, opened 2020-09-03
Fix empty multipart payload processing

cy6erGn0m:
**Subsystem**
Both client and server

**Motivation**
[#KTOR-767](https://youtrack.jetbrains.com/issue/KTOR-767) (same as #482 )

I personally don't see if an empty payload is exactly conforming the specification but for sure we need to handle it safely since browsers generate such requests.

**Solution**
Don't demand a delimiter but try to skip and succeed if EOF detected. 

**Note**
For review purpose, the reformat/"explicit public" commit will be added later.

### DIFF
--- a/ktor-http/ktor-http-cio/jvm/src/io/ktor/http/cio/Multipart.kt
+++ b/ktor-http/ktor-http-cio/jvm/src/io/ktor/http/cio/Multipart.kt
@@ -199,9 +199,12 @@ public suspend fun boundary(boundaryPrefixed: ByteBuffer, input: ByteReadChannel
 
 /**
  * Skip multipart boundary
+ * @return `true` if end channel encountered
  */
 private suspend fun skipBoundary(boundaryPrefixed: ByteBuffer, input: ByteReadChannel): Boolean {
-    input.skipDelimiter(boundaryPrefixed)
+    if (!input.skipDelimiterOrEof(boundaryPrefixed)) {
+        return true
+    }
 
     var result = false
     @Suppress("DEPRECATION")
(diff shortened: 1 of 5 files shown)


ktor-io/jvm/src/io/ktor/utils/io/Delimited.kt
```diff
@@ -52,11 +53,39 @@ suspend fun ByteReadChannel.skipDelimiter(delimiter: ByteBuffer) {
     }
 }
 
-private suspend fun ByteReadChannel.skipDelimiterSuspend(delimiter: ByteBuffer) {
+/**
+ * Tries to skip the specified [delimiter] or fails if encounters bytes differs from the required.
+ * @return `true` if the delimiter was found and skipped or `false` when EOF.
+ */
+@ExperimentalIoApi
+suspend fun ByteReadChannel.trySkipDelimiter(delimiter: ByteBuffer): Boolean {
```
e5l, line 61: no tests

e5l (changes requested):

ktor-io/api/ktor-io.api
```diff
@@ -295,6 +295,7 @@ public final class io/ktor/utils/io/CoroutinesKt {
 public final class io/ktor/utils/io/DelimitedKt {
 	public static final fun readUntilDelimiter (Lio/ktor/utils/io/ByteReadChannel;Ljava/nio/ByteBuffer;Ljava/nio/ByteBuffer;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
 	public static final fun skipDelimiter (Lio/ktor/utils/io/ByteReadChannel;Ljava/nio/ByteBuffer;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
+	public static final fun trySkipDelimiter (Lio/ktor/utils/io/ByteReadChannel;Ljava/nio/ByteBuffer;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
```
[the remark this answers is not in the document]
cy6erGn0m: This will require to copy a lot of code from io to http-cio. Unfortunately, we can't use `@Internal` annotation (because it is in utils)

ktor-http/ktor-http-cio/jvm/src/io/ktor/http/cio/Multipart.kt
```diff
@@ -196,9 +196,12 @@ suspend fun boundary(boundaryPrefixed: ByteBuffer, input: ByteReadChannel): Bool
 
 /**
  * Skip multipart boundary
+ * @return `true` if end channel encountered
  */
 private suspend fun skipBoundary(boundaryPrefixed: ByteBuffer, input: ByteReadChannel): Boolean {
-    input.skipDelimiter(boundaryPrefixed)
+    if (!input.trySkipDelimiter(boundaryPrefixed)) {
```
e5l, line 202: changed behavior without test

ktor-io/api/ktor-io.api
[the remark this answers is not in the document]
e5l: Nope! we made a strict commitment and have excuses to do that

ktor-http/ktor-http-cio/jvm/src/io/ktor/http/cio/Multipart.kt
cy6erGn0m: Actually two tests were added

e5l: and no one is testing this method directly

ktor-io/jvm/src/io/ktor/utils/io/Delimited.kt
```diff
@@ -52,11 +53,18 @@ suspend fun ByteReadChannel.skipDelimiter(delimiter: ByteBuffer) {
     }
 }
 
-private suspend fun ByteReadChannel.skipDelimiterSuspend(delimiter: ByteBuffer) {
+private suspend fun ByteReadChannel.skipDelimiterSuspend(delimiter: ByteBuffer): Boolean {
```
e5l, line 56: this change also requires test

cy6erGn0m: The behaviour is not changed, only a result flag is provided. Also, this is not a separate function but a companion function for tail-suspend `skipDelimiter`. We actually don't need this code anymore, since we copied it.

e5l: Do we have a test checking that method?

ktor-io/jvm/src/io/ktor/utils/io/Delimited.kt
```diff
@@ -52,11 +53,18 @@ suspend fun ByteReadChannel.skipDelimiter(delimiter: ByteBuffer) {
...
+    var result = true
+
     lookAheadSuspend {
-        awaitAtLeast(delimiter.remaining())
+        if (!awaitAtLeast(delimiter.remaining()) && !awaitAtLeast(1)) {
```
e5l, line 60: this change is not obvious and requires test

ktor-io/jvm/src/io/ktor/utils/io/Delimited.kt
cy6erGn0m: I'll remove this change but apply a test to the copied version with change

ktor-io/jvm/src/io/ktor/utils/io/Delimited.kt
cy6erGn0m: Reverted and added in the copied part (coverted by test)

ktor-http/ktor-http-cio/jvm/src/io/ktor/http/cio/Multipart.kt
cy6erGn0m: Added tests

ktor-io/jvm/src/io/ktor/utils/io/Delimited.kt
cy6erGn0m: Fixed

e5l (approved):
LGTM, consider removing the commit message to match others in master

cy6erGn0m:
Merged

PR #2044, opened 2020-09-03
KTOR-372 Fix routing with trailing slash in path

rsinukov:
**Subsystem**
Server

**Motivation**
[Ktor should resolve trailing slash as a different URL
](https://youtrack.jetbrains.com/issue/KTOR-372)

**Solution**
Save flag if route ends with trailing slash and check it in selectors

### DIFF
--- a/ktor-server/ktor-server-core/jvm/src/io/ktor/routing/RouteSelector.kt
+++ b/ktor-server/ktor-server-core/jvm/src/io/ktor/routing/RouteSelector.kt
@@ -107,6 +107,7 @@ class RootRouteSelector(rootPath: String = "") : RouteSelector(RouteSelectorEval
         true, RouteSelectorEvaluation.qualityConstant,
         segmentIncrement = parts.size
     )
+    private val hasTrailingSlash = rootPath.endsWith('/')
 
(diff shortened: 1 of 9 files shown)


e5l (changes requested):
The `apiCheck` task is failing on TeamCity. Please consider rerun it with `clean` before

ktor-server/ktor-server-core/jvm/src/io/ktor/routing/RouteSelector.kt
```diff
@@ -181,8 +185,17 @@ data class OptionalParameterRouteSelector(val name: String) : RouteSelector(Rout
...
-data class PathSegmentConstantRouteSelector(val value: String) : RouteSelector(RouteSelectorEvaluation.qualityConstant) {
+data class PathSegmentConstantRouteSelector internal constructor(
+    val value: String,
+    private val hasTrailingSlash: Boolean
+) : RouteSelector(RouteSelectorEvaluation.qualityConstant) {
+
+    public constructor(value: String) : this(value, false)
+
     override fun evaluate(context: RoutingResolveContext, segmentIndex: Int): RouteSelectorEvaluation {
+        if (segmentIndex == context.segments.lastIndex && hasTrailingSlash != context.hasTrailingSlash) {
```
e5l, line 196: could be done with `when`

Issue #1007, opened 2019-03-11
multiple authenticate fail to redirect


PinkieSwirl:
### Ktor Version

1.1.3

### Ktor Engine Used(client or server and name)

netty

### JVM Version, Operating System and Relevant Context

jdk 11, windows 10

### Feedback

From my understanding the following code should redirect to the `/login` route, however nothing happens:
```
fun main(args: Array<String>): Unit = io.ktor.server.netty.EngineMain.main(args)

@Suppress("unused") // Referenced in application.conf
@kotlin.jvm.JvmOverloads
fun Application.module(testing: Boolean = false) {
    install(Authentication) {
        form {
            challenge = FormAuthChallenge.Redirect { "/login" }
        }
    }

    routing {
        get("/login") {
            call.respondText("LOGIN")
        }
        authenticate(optional = true) {
            static("/") {
                resources("static")
            }
        }

        authenticate {
            get("/") {
                call.respondText("HELLO WORLD!", contentType = ContentType.Text.Plain)
            }
        }
    }
}
```
The problem seems to be that static (`static("/")`) has the same root path as get (` get("/")`) in conjunction with `authenticate(optional = true) `.

The redirect works again, if the paths are different or/and `optional = false`.

If I activate call logging I get: `INFO  Application - Unhandled: GET - /`

oleg-larshin:
Please check the <a href="https://youtrack.jetbrains.com/issue/KTOR-654">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

Stexxe:
Cannot reproduce it using Ktor 1.6.0 with the following modification:
```kt
install(Authentication) {
    form {
        challenge {
            call.respondRedirect("/login")
        }
    }
}
```

PR #2046, opened 2020-09-07
Server session ID handling

cy6erGn0m:
**Subsystem**
ktor-server-core

**Motivation**
This PR corrects the wrong behaviour in the case when a client sends a wrong session ID through cookies. This could happen if a session is expired or after the server restart with in-memory session storage.

Logging related issue: [YT #776](https://youtrack.jetbrains.com/issue/KTOR-776)
Cookie handling: [YT #1007](https://youtrack.jetbrains.com/issue/KTOR-1007)

**Solution**
- Remove the exception name that caused confusion
- Remove wrong session ids by sending cookie expiration so a client stops sending wrong IDs

### DIFF
--- a/ktor-server/ktor-server-core/jvm/src/io/ktor/sessions/SessionTrackerById.kt
+++ b/ktor-server/ktor-server-core/jvm/src/io/ktor/sessions/SessionTrackerById.kt
@@ -36,7 +36,7 @@ class SessionTrackerById<S : Any>(
                 serializer.deserialize(text)
             }
         } catch (notFound: NoSuchElementException) {
-            call.application.log.debug("Failed to lookup session: $notFound")
+            call.application.log.debug("Failed to lookup session: ${notFound.message ?: notFound.toString()}")
         }
 
         // we remove the wrong session identifier if no related session found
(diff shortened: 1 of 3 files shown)


ktor-server/ktor-server-core/jvm/src/io/ktor/sessions/SessionTrackerById.kt
```diff
@@ -36,7 +36,7 @@ class SessionTrackerById<S : Any>(
                 serializer.deserialize(text)
             }
         } catch (notFound: NoSuchElementException) {
-            call.application.log.debug("Failed to lookup session: $notFound")
+            call.application.log.debug("Failed to lookup session: ${notFound.message ?: notFound.toString()}")
```
hhariri, line 39: Should be "look up", as opposed to "lookup".

cy6erGn0m: notFound.message usually contains something like session ID [number] is not found. I believe it should be enough. So I simply removed exception class name "NoSuchElementException" that was too loud.

cy6erGn0m: Note that in production you can't see these message because of the debug log level.

cy6erGn0m: Perhaps, we could add something like "The session id is wrong or outdated"

hhariri: But would still need to be the verb "look up".

cy6erGn0m:
Closing to split into two PRs due to the new policy

cy6erGn0m:
New PRs:
#2078 removing staled/bad session id
#2079 improve logging message

PR #2049, opened 2020-09-07
KTOR-1240 Fix socket timeout

e5l:
Fixes [KTOR-1240](https://youtrack.jetbrains.com/issue/KTOR-1240)


cy6erGn0m:
mm.. the total diff adds API with no actual code change? Are you sure?

e5l:
Fixed, please take a look

Stexxe (changes requested):
Where exactly was a typo?

e5l:
`val`->`var`

Stexxe:
> `val`->`var`

Looks like it was done on purpose. Could you please describe in the https://youtrack.jetbrains.com/issue/KTOR-1240 why  `socketTimeout` needs to be mutable?

e5l:
Nope, it wasn't. It's a property in config.

e5l:
explained in the issue

Stexxe (approved):

PR #2050, opened 2020-09-07
KTOR-613 increase priority of route with prefix and suffix

rsinukov:
**Subsystem**
Server, routing

**Motivation**
First PR to fix [KTOR-613](https://youtrack.jetbrains.com/issue/KTOR-613). Route with parameter that has prefix or suffix should have higher priority to the route that has only parameter

**Solution**
Add quality that is higher then `qualityParameter` but lower then `qualityConstant`


ktor-server/ktor-server-core/jvm/src/io/ktor/routing/RouteSelector.kt
```diff
@@ -218,7 +223,13 @@ data class PathSegmentParameterRouteSelector(val name: String, val prefix: Strin
                     return RouteSelectorEvaluation.Failed
 
             val values = parametersOf(name, suffixChecked)
-            return RouteSelectorEvaluation(true, RouteSelectorEvaluation.qualityParameter, values, segmentIncrement = 1)
+            return RouteSelectorEvaluation(
+                true,
```
e5l, line 227: consider adding parameter names

e5l (changes requested):
LGTM, except codestyle

e5l (approved):

e5l:
One more thing to do: could you move `Subsystem` and `Motivation` to YT issue?

ktor-server/ktor-server-core/jvm/src/io/ktor/routing/RouteSelector.kt
e5l: consider extracting common code here, but I don't insist

ktor-server/ktor-server-core/jvm/src/io/ktor/routing/RouteSelector.kt
```diff
@@ -233,27 +244,34 @@ data class PathSegmentParameterRouteSelector(val name: String, val prefix: Strin
...
-                prefixChecked
-            else
-                if (prefixChecked.endsWith(suffix))
-                    prefixChecked.dropLast(suffix.length)
-                else
-                    return RouteSelectorEvaluation.Missing
+        return evaluate(context.segments, segmentIndex)
+    }
+
+    internal fun evaluate(segments: List<String>, segmentIndex: Int): RouteSelectorEvaluation {
```
e5l, line 252: Is this code duplicate? Can extract the top-level method instead?

e5l (changes requested):

ktor-server/ktor-server-core/jvm/src/io/ktor/routing/RouteSelector.kt
rsinukov: Yes and no. In case of failure one method returns `RouteSelectorEvaluation.Failed` and another `RouteSelectorEvaluation.Missing`. They can be extracted to one with extra parameter, I will do it.

ktor-server/ktor-server-core/jvm/src/io/ktor/routing/RouteSelector.kt
```diff
@@ -403,3 +381,41 @@ data class HttpAcceptRouteSelector(val contentType: ContentType) : RouteSelector
...
+    name: String,
+    prefix: String?,
+    suffix: String?,
+    isOptional: Boolean
+): RouteSelectorEvaluation {
+    fun failedEvaluation(): RouteSelectorEvaluation {
+        return if (isOptional) RouteSelectorEvaluation.Missing else RouteSelectorEvaluation.Failed
+    }
+
+    if (segmentIndex < segments.size) {
```
e5l, line 397: Consider invert expression and early `return` to avoid lots of nested scopes

e5l (changes requested):

e5l (approved):

PR #2052, opened 2020-09-09
#KTOR-936 Migrate to explicit api mode(part 1)

e5l:

### DIFF
--- a/ktor-io/common/src/io/ktor/utils/io/core/AbstractInput.kt
+++ b/ktor-io/common/src/io/ktor/utils/io/core/AbstractInput.kt
@@ -12,15 +12,15 @@ import io.ktor.utils.io.pool.*
  * The default abstract base class implementing [Input] interface.
  * @see [AbstractInput.fill] and [AbstractInput.closeSource].
  */
-abstract class AbstractInput(
+public abstract class AbstractInput(
     head: ChunkBuffer = ChunkBuffer.Empty,
     remaining: Long = head.remainingAll(),
-    val pool: ObjectPool<ChunkBuffer> = ChunkBuffer.Pool
+    public val pool: ObjectPool<ChunkBuffer> = ChunkBuffer.Pool
 ) : Input {
 
     @Suppress("DEPRECATION")
     @Deprecated("Binary compatibility.", level = DeprecationLevel.HIDDEN)
-    constructor(head: IoBuffer = IoBuffer.Empty,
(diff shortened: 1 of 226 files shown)


ktor-io/jvm/src/io/ktor/utils/io/ByteBufferChannel.kt
```diff
@@ -27,7 +29,7 @@ internal open class ByteBufferChannel(
 ) : ByteChannel, ByteReadChannel, ByteWriteChannel, LookAheadSuspendSession, HasReadSession, HasWriteSession {
 
     // internal constructor for reading of byte buffers
```
cy6erGn0m, line 31: Looks like it is not true anymore. We could remove the comment.

cy6erGn0m (approved):

gradle.properties
```diff
@@ -2,7 +2,7 @@
 kotlin.code.style=official
 
 # config
-version=1.4.0
+version=1.4.1-SNAPSHOT
```
Stexxe, line 5: Most likely commit "KTOR-1027 Update versions" shouldn't be here

ktor-io/common/src/io/ktor/utils/io/core/AbstractInput.kt
```diff
@@ -314,13 +314,13 @@ abstract class AbstractInput(
     }
 
     @Deprecated("Binary compatibility.", level = DeprecationLevel.HIDDEN)
-    final override fun readShort() = readShort()
+    final override fun readShort(): Short = readShort()
```
Stexxe, line 317: Should `public` modifier be here?

Stexxe (changes requested):

ktor-io/common/src/io/ktor/utils/io/core/AbstractInput.kt
e5l: Nope. We should set visibility modifiers for `overrides` only when it changes visibility

gradle.properties
e5l: rebased

ktor-io/jvm/src/io/ktor/utils/io/ByteBufferChannel.kt
e5l: thanks, done

Stexxe (approved):

PR #2054, opened 2020-09-09
#KTOR-936 Migrate to explicit api mode(part 3) 

e5l:


ktor-network/common/src/io/ktor/network/sockets/TypeOfService.kt
```diff
@@ -9,23 +9,23 @@ package io.ktor.network.sockets
  * @property value an unsigned byte IP_TOS value
  */
 @OptIn(ExperimentalUnsignedTypes::class)
```
cy6erGn0m, line 11: FYI. Btw, it is a potential compatibility bomb. It wasn't an issue when it was initially introduced because we had no such strict commitments and the module was more auxiliary. Hope Kotlin will not change anything in unsigned types generation.

cy6erGn0m (approved):

Stexxe (approved):

PR #2057, opened 2020-09-09
#KTOR-936 Migrate to explicit api mode(part 6) 

e5l:


ktor-server/ktor-server-benchmarks/src/jmh/kotlin/io/ktor/server/benchmarks/AsyncIntegrationBenchmark.kt
```diff
@@ -23,7 +23,7 @@ import java.io.*
 import java.util.concurrent.*
 
 @State(Scope.Group)
-abstract class AsyncIntegrationBenchmark<TEngine : ApplicationEngine> {
+public abstract class AsyncIntegrationBenchmark<TEngine : ApplicationEngine> {
```
cy6erGn0m, line 26: Can't we simply add benchmarks module to the exclusion list? This module is not used outside and looks more like a testing module? Or if we want to keep our codebase uniform, shouldn't we avoid exclusions instead?

cy6erGn0m (changes requested):

ktor-server/ktor-server-benchmarks/src/jmh/kotlin/io/ktor/server/benchmarks/AsyncIntegrationBenchmark.kt
e5l: Let's exclude benchmarks. It was accidentally touched by regex.

e5l: Fixed in the fixup in the last PR

Stexxe (approved):

cy6erGn0m (changes requested):
Please add the apidump change we discussed yesterday in private. Would be nice to add a brief comment regarding why do we believe that it's safe to ignore it (just for history). The question is where should we do that: in a commit message or right here in PR.

cy6erGn0m (approved):

PR #2058, opened 2020-09-09
#KTOR-936 Migrate to explicit api mode(part 7) 

e5l:

### DIFF
--- a/ktor-features/ktor-locations/jvm/src/io/ktor/locations/LocationsInfo.kt
+++ b/ktor-features/ktor-locations/jvm/src/io/ktor/locations/LocationsInfo.kt
@@ -16,7 +16,7 @@ import kotlin.reflect.*
  * @property queryParameters is a list of properties stored in query parameters
  */
 @KtorExperimentalLocationsAPI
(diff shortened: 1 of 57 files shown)


ktor-features/ktor-locations/jvm/src/io/ktor/locations/LocationsInfo.kt
```diff
@@ -16,7 +16,7 @@ import kotlin.reflect.*
  * @property queryParameters is a list of properties stored in query parameters
  */
 @KtorExperimentalLocationsAPI
-data class LocationInfo internal constructor(
+public data class LocationInfo internal constructor(
     val klass: KClass<*>,
```
cy6erGn0m, line 20: So it's not required here to provide visibility modifiers? Why? Is it relaxed for internals or for data classes?

e5l: for data classes

cy6erGn0m (approved):

Stexxe (approved):

PR #2059, opened 2020-09-09
#KTOR-936 Migrate to explicit api mode

e5l:


build.gradle
```diff
@@ -160,11 +167,13 @@ allprojects {
         }
     }
 
-
-
     configurations { testOutput }
 
     kotlin {
+        if (!ignoreExplicitApi.contains(project.name)){
+            explicitApi()
```
cy6erGn0m, line 174: So it becomes error, right?

e5l: Yep

build.gradle
```diff
@@ -99,6 +99,13 @@ def check(Object version, String libVersion, String libName) {
     }
 }
 
+def ignoreExplicitApi = [
+    "ktor-client-tests",
+    "ktor-client-json-tests",
+    "ktor-server-test-host",
```
cy6erGn0m, line 105: Why is the test host excluded?

e5l: To exclude suites

build.gradle
```diff
@@ -99,6 +99,14 @@ def check(Object version, String libVersion, String libName) {
     }
 }
 
+def ignoreExplicitApi = [
```
Stexxe, line 102: This name is better suited for a method, not for a list of things. I suggest  `DisabledExplicitApiModeProjects` or something shorter.

build.gradle
```diff
@@ -247,7 +253,7 @@ if (build_snapshot_train) {
                     include 'META-INF/MANIFEST.MF'
                 }.getFiles().first()
 
-                manifest.readLines().each {
+                manifes.readLines().each {
```
Stexxe, line 256: Typo?

Stexxe (changes requested):

build.gradle
e5l: yep. thanks for catching

Stexxe (approved):

PR #2060, opened 2020-09-09
KTOR-949 fix reusing HttpRequestBuilder for different network clients

rsinukov:
**Subsystem**
Client

**Motivation**
[OkHttp Can't reuse same HttpRequestBuilder for different network clients](https://youtrack.jetbrains.com/issue/KTOR-949)

**Solution**
When first client respond with timeout, it also cancels `executionContext` of a builder. Second client doesn't receive request because job is cancelled. `SupervisorJob` fix this issue.

### DIFF
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/request/HttpRequest.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/request/HttpRequest.kt
@@ -85,7 +85,7 @@ public class HttpRequestBuilder : HttpMessageBuilder {
      * A deferred used to control the execution of this request.
      */
     @KtorExperimentalAPI
-    public var executionContext: Job = Job()
+    public var executionContext: Job = SupervisorJob()
         .also { it.makeShared() }
         internal set(value) {
             value.makeShared()
--- a/ktor-client/ktor-client-okhttp/jvm/test/io/ktor/client/engine/okhttp/RequestTests.kt
+++ b/ktor-client/ktor-client-okhttp/jvm/test/io/ktor/client/engine/okhttp/RequestTests.kt
@@ -4,12 +4,32 @@
 
 package io.ktor.client.engine.okhttp
 
+import io.ktor.application.*
+import io.ktor.client.*
 import io.ktor.client.request.*
 import io.ktor.client.tests.utils.*
+import io.ktor.http.*
+import io.ktor.network.sockets.*
+import io.ktor.response.*
+import io.ktor.routing.*
+import io.ktor.server.cio.*
+import io.ktor.server.engine.*
+import kotlinx.coroutines.*
 import okhttp3.*
+import java.time.*
 import kotlin.test.*
 
-class RequestTests {
+class RequestTests : TestWithKtor() {
+
+    override val server = embeddedServer(CIO, serverPort) {
+        routing {
+            get("/delay") {
+                val delay = call.parameters["delay"]!!.toLong()
+                delay(delay)
+                call.respondText("OK")
+            }
+        }
+    }
 
     class LoggingInterceptor : Interceptor {
         override fun intercept(chain: Interceptor.Chain): Response {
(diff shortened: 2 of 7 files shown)


ktor-client/ktor-client-okhttp/jvm/test/io/ktor/client/engine/okhttp/RequestTests.kt
```diff
@@ -32,4 +52,32 @@ class RequestTests {
             client.get<String>("https://google.com")
         }
     }
+
+    @Test
+    fun testReusingRequestBuilderOnMultipleClients() {
+        val rb = HttpRequestBuilder()
+        rb.url.takeFrom("http://0.0.0.0:$serverPort/delay?delay=500")
```
e5l, line 59: Consider using `$TEST_URL` constant

ktor-client/ktor-client-okhttp/jvm/test/io/ktor/client/engine/okhttp/RequestTests.kt
```diff
@@ -32,4 +52,32 @@ class RequestTests {
             client.get<String>("https://google.com")
         }
     }
+
+    @Test
+    fun testReusingRequestBuilderOnMultipleClients() {
+        val rb = HttpRequestBuilder()
```
e5l, line 58: Please avoid using such shortcuts for names

ktor-client/ktor-client-core/common/src/io/ktor/client/request/HttpRequest.kt
```diff
@@ -85,7 +85,7 @@ class HttpRequestBuilder : HttpMessageBuilder {
      * A deferred used to control the execution of this request.
      */
     @KtorExperimentalAPI
-    public var executionContext: Job = Job()
+    public var executionContext: Job = SupervisorJob()
```
e5l, line 88: I would like to see more tests with `request.await()`: the case where you start request and awaiting completion on context. Especially cases with redirect, timeout, and cancel.

e5l (changes requested):

ktor-client/ktor-client-okhttp/jvm/test/io/ktor/client/engine/okhttp/RequestTests.kt
```diff
@@ -4,12 +4,32 @@
...
+import io.ktor.server.engine.*
+import kotlinx.coroutines.*
 import okhttp3.*
+import java.time.*
 import kotlin.test.*
 
-class RequestTests {
+class RequestTests : TestWithKtor() {
+
+    override val server = embeddedServer(CIO, serverPort) {
```
e5l, line 24: Please rewrite this suite using `ClientLoader` and move the test in common

rsinukov: This issue is related to OkHttp only and is not reproduced with common configuration

ktor-client/ktor-client-core/common/src/io/ktor/client/request/HttpRequest.kt
cy6erGn0m: A lot of different coroutines are started under the job. Are we sure that it's safe to avoid auto crash handling?

ktor-client/ktor-client-okhttp/jvm/test/io/ktor/client/engine/okhttp/RequestTests.kt
rsinukov: can't be constant because it uses `$serverPort`, but extracted to read-only property in the base class

ktor-client/ktor-client-core/common/src/io/ktor/client/request/HttpRequest.kt
e5l: That thing we can test actually. For instance, add a test that cancels body or launch and cancel child coroutine

rsinukov: @cy6erGn0m afaik, `SupervisorJob` doesn't avoid auto crash handling but keep parent job active for reuse. All exceptions are still propagating to the caller. Do I miss anything?

Added test for failure in child coroutine to verify that caller will receive an exception and the parent job is still active.

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/ExceptionsTest.kt
```diff
@@ -35,4 +39,26 @@ class ExceptionsTest {
...
+            val requestBuilder = HttpRequestBuilder()
+            requestBuilder.url.takeFrom("$TEST_SERVER/download/infinite")
+
+            assertFailsWith<IllegalStateException> {
+                client.get<HttpStatement>(requestBuilder).execute { response ->
+                    try {
+                        CoroutineScope(response.coroutineContext)
+                            .launch { throw IllegalStateException("failed on receive") }
+                            .join()
+                    } catch (e: Exception) {
```
e5l, line 55: cause

e5l (approved):
lgtm

PR #2062, opened 2020-09-11
Fix HttpStatement.toString

e5l:
fixes #KTOR-1005

### DIFF
--- a/ktor-client/ktor-client-core/common/test/HttpStatementTest.kt
+++ b/ktor-client/ktor-client-core/common/test/HttpStatementTest.kt
@@ -1,0 +1,20 @@
+import io.ktor.client.*
+import io.ktor.client.request.*
+import io.ktor.client.statement.*
+import kotlin.test.*
+
+/*
+ * Copyright 2014-2020 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+class HttpStatementTest {
+
+    @Test
+    fun testStatementToString() {
+        val statement = HttpStatement(HttpRequestBuilder(), HttpClient(TestEngine))
+
+        assertEquals("HttpStatement[http://localhost/]", statement.toString())
+        assertEquals("HttpStatement[http://localhost/]", statement.toString())
+    }
+}
+
--- a/ktor-http/common/src/io/ktor/http/HttpUrlEncoded.kt
+++ b/ktor-http/common/src/io/ktor/http/HttpUrlEncoded.kt
@@ -4,7 +4,6 @@
 
 package io.ktor.http
 
-import io.ktor.util.*
 import io.ktor.utils.io.charsets.*
 
 /**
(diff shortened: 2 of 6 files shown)


ktor-http/common/src/io/ktor/http/HttpUrlEncoded.kt
```diff
@@ -67,3 +67,9 @@ fun Parameters.formUrlEncodeTo(out: Appendable) {
         .flatMap { e -> if (e.value.isEmpty()) listOf(e.key to null) else e.value.map { e.key to it } }
         .formUrlEncodeTo(out)
 }
+
+fun ParametersBuilder.formUrlEncodeTo(out: Appendable) {
```
cy6erGn0m, line 71: Visibility modifier.
Shouldn't it be internal to avoid ABI change?

ktor-client/ktor-client-core/common/test/HttpStatementTest.kt
```diff
@@ -0,0 +1,20 @@
...
+import kotlin.test.*
+
+/*
+ * Copyright 2014-2020 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+class HttpStatementTest {
+
+    @Test
+    fun testStatementToString() {
```
cy6erGn0m, line 13: The copied encode function is not covered by tests so could be potentially broken in the future. Shouldn't we copy tests as well?

cy6erGn0m (changes requested):

ktor-client/ktor-client-core/common/test/HttpStatementTest.kt
e5l: Sorry it's still draft. Forgot to switch the mode

ktor-http/common/src/io/ktor/http/URLUtils.kt
```diff
@@ -98,3 +98,21 @@ internal fun Appendable.appendUrlFullPath(
 
     queryParameters.formUrlEncodeTo(this)
 }
+
+internal fun Appendable.appendUrlFullPath(
+    encodedPath: String,
+    queryParameters: ParametersBuilder,
+    trailingQuery: Boolean
+) {
+    if (encodedPath.isNotBlank() && !encodedPath.startsWith("/")) {
```
cy6erGn0m, line 107: `&& encodedPath[0] != '/'` ?

ktor-http/common/src/io/ktor/http/URLUtils.kt
```diff
@@ -98,3 +98,21 @@ internal fun Appendable.appendUrlFullPath(
...
+        append('/')
+    }
+
+    append(encodedPath)
+
+    if (!queryParameters.isEmpty() || trailingQuery) {
```
cy6erGn0m, line 113: isNotEmpty?

cy6erGn0m (approved):
Not sure, since the function is just copied, most likey we don't need to fix the things I've noticed

PR #2065, opened 2020-09-14
KTOR-307 update documentation of HttpResponse::content property

rsinukov:
Added note that this property is not touched by any interceptors from HttpResponsePipeline

**Subsystem**
Server

**Motivation**
Property `content` of `HttpResponse` is a raw response from the server, that is not modified by `HttpResponsePipeline`. It created confusions and questions, like [KTOR-307](https://youtrack.jetbrains.com/issue/KTOR-307)

**Solution**
Added note that this property is not touched by any interceptors from HttpResponsePipeline

### DIFF
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/statement/HttpResponse.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/statement/HttpResponse.kt
@@ -45,7 +45,10 @@ public abstract class HttpResponse : HttpMessage, CoroutineScope {
     public abstract val responseTime: GMTDate
 
     /**
-     * [ByteReadChannel] with the payload of the response.
+     * Unmodified [ByteReadChannel] with the raw payload of the response.
+     *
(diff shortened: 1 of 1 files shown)


ktor-client/ktor-client-core/common/src/io/ktor/client/statement/HttpResponse.kt
```diff
@@ -45,7 +45,10 @@ public abstract class HttpResponse : HttpMessage, CoroutineScope {
     public abstract val responseTime: GMTDate
 
     /**
-     * [ByteReadChannel] with the payload of the response.
+     * Unmodified [ByteReadChannel] with the raw payload of the response.
+     *
+     * **Note:** this content doesn't go through any interceptors from [HttpResponsePipeline].
+     * If you need modified content, use [HttpResponse::receive<T>] function.
```
e5l, line 51: Could you highlight `receive<ByteReadChannel>`  here?

e5l (approved):

PR #2066, opened 2020-09-15
KTOR-1036 Ignore content-length if transfer-encoding is chunked

e5l:
Ignore content length during parsing.

### DIFF
--- a/ktor-http/ktor-http-cio/jvm/test/io/ktor/tests/http/cio/ChunkedTest.kt
+++ b/ktor-http/ktor-http-cio/jvm/test/io/ktor/tests/http/cio/ChunkedTest.kt
@@ -23,6 +23,31 @@ class ChunkedTest {
         decodeChunked(ch, parsed)
     }
 
+    @Test
+    fun testChunkedWithContentLength() = runBlocking {
+        val chunkedContent = listOf(
+            "3\r\n",
+            "a=1\r\n",
+            "0\r\n",
+            "\r\n",
+        )
+
+        val input = writer {
+            chunkedContent.forEach {
+                channel.writeStringUtf8(it)
+            }
+        }.channel
+
+        val output = ByteChannel()
+        launch {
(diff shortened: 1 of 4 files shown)


ktor-http/ktor-http-cio/jvm/test/io/ktor/tests/http/cio/ChunkedTest.kt
```diff
@@ -23,6 +23,32 @@ class ChunkedTest {
...
+
+        val input = writer {
+            chunkedContent.forEach {
+                channel.writeStringUtf8(it)
+            }
+        }.channel
+
+        val output = ByteChannel()
+        launch {
+            @Suppress("DEPRECATION")
```
Stexxe, line 43: Why suppressing?

ktor-http/ktor-http-cio/jvm/src/io/ktor/server/cio/backend/ServerPipeline.kt
```diff
@@ -87,7 +87,7 @@ public fun CoroutineScope.startServerConnectionPipeline(
 
             val response = ByteChannel()
 
-            val transferEncoding = request.headers["Transfer-Encoding"]
+            val transferEncoding = request.headers["Transfer-Encoding"]?.trim()
```
Stexxe, line 90: Is this covered by `testChunkedWithContentLength`?

Stexxe (changes requested):

ktor-http/ktor-http-cio/jvm/src/io/ktor/server/cio/backend/ServerPipeline.kt
e5l: It should be removed, accidentally cherry picked. Thanks for the notice

Stexxe (approved):

PR #2068, opened 2020-09-15
KTOR-1034 Fix ConcurrentList clear

e5l:
Finish KTOR-1034 with fix of `clear` method

### DIFF
--- a/ktor-utils/common/test/io/ktor/util/ConcurrentListTest.kt
+++ b/ktor-utils/common/test/io/ktor/util/ConcurrentListTest.kt
@@ -37,4 +37,9 @@ class ConcurrentListTest {
             assertEquals(it, list[it])
         }
     }
+
+    @Test
+    fun testClearHasNoExceptions() {
+        ConcurrentList<Unit>().clear()
+    }
 }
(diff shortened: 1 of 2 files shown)


ktor-utils/common/test/io/ktor/util/ConcurrentListTest.kt
```diff
@@ -37,4 +37,9 @@ class ConcurrentListTest {
             assertEquals(it, list[it])
         }
     }
+
+    @Test
+    fun testClear() {
+        ConcurrentList<Unit>().clear()
```
Stexxe, line 43: This test doesn't have any assertions. Do you expect that some exception won't be thrown?

Stexxe (changes requested):

ktor-utils/common/test/io/ktor/util/ConcurrentListTest.kt
e5l: Yep. There was exception in this method

Stexxe: Can we illustrate that no exception is expected?

Stexxe (approved):

PR #2071, opened 2020-09-15
Closes #KTOR-366

andrey-vasilyev:
Subsystem
Client, ktor-network-tls

Motivation
Fix for [KTOR-366](https://youtrack.jetbrains.com/issue/KTOR-366)

Solution
Corrected the search of a Principal in the collection of principals. I.e. comparing X500Principal against other implementation of Principal interface always returns false. And since io.ktor.network.tls.CertificateInfo is an internal class I believe it is ok to make sure authorities are always X500Principals and then convert any other principal to X500Principal before comparison. That way the comparison will always be correct.

### DIFF
--- a/ktor-network/ktor-network-tls/build.gradle.kts
+++ b/ktor-network/ktor-network-tls/build.gradle.kts
@@ -1,3 +1,5 @@
+val netty_version: String by project.extra
+
 kotlin.sourceSets {
     val commonMain by getting {
         dependencies {
@@ -8,6 +10,7 @@ kotlin.sourceSets {
     val jvmTest by getting {
         dependencies {
             api(project(":ktor-network:ktor-network-tls:ktor-network-tls-certificates"))
+            api("io.netty:netty-handler:$netty_version")
         }
     }
 }
--- a/ktor-network/ktor-network-tls/jvm/test/io/ktor/network/tls/tests/ConnectionTests.kt
+++ b/ktor-network/ktor-network-tls/jvm/test/io/ktor/network/tls/tests/ConnectionTests.kt
@@ -8,12 +8,21 @@ import io.ktor.network.selector.*
 import io.ktor.network.sockets.*
 import io.ktor.network.tls.*
 import io.ktor.network.tls.certificates.*
+import io.ktor.utils.io.*
+import io.netty.bootstrap.*
+import io.netty.channel.*
+import io.netty.channel.nio.*
+import io.netty.channel.socket.*
+import io.netty.channel.socket.nio.*
+import io.netty.handler.ssl.*
 import kotlinx.coroutines.*
 import kotlinx.coroutines.debug.junit4.*
-import io.ktor.utils.io.*
 import org.junit.*
 import java.io.*
 import java.net.*
+import java.net.ServerSocket
+import java.security.*
+import java.security.cert.*
 import javax.net.ssl.*
 
 class ConnectionTests {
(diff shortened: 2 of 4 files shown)


ktor-network/ktor-network-tls/jvm/test/io/ktor/network/tls/tests/ConnectionTests.kt
```diff
@@ -66,4 +75,62 @@ class ConnectionTests {
         output.close()
         socket.close()
     }
+
+    @Test
+    fun test() {
```
rsinukov, line 80: please give a proper name to the test

andrey-vasilyev: Fixed it, thank you!

ktor-network/ktor-network-tls/build.gradle.kts
```diff
@@ -8,6 +10,7 @@ kotlin.sourceSets {
     val jvmTest by getting {
         dependencies {
             api(project(":ktor-network:ktor-network-tls:ktor-network-tls-certificates"))
+            api("io.netty:netty-handler:$netty_version")
```
e5l, line 13: Could we implement the comparison instead of bringing additional dependency here? We want to avoid using additional dependencies here.

e5l (changes requested):

e5l (approved):

rsinukov (approved):

andrey-vasilyev:
Broken test seems unrelated. Is it possible proceed with the merge?

e5l:
Could you rebase it on `master` to run CI build?

andrey-vasilyev:
done

andrey-vasilyev:
Decided to update the PR with new closely related findings.

According to [RFC 5246](https://tools.ietf.org/html/rfc5246#section-7.4.8): 

> 7.4.8.  Certificate Verify
>
>   When this message will be sent:
>
>     This message is used to provide explicit verification of a client
>     certificate.  This message is only sent following a client
>     certificate that has signing capability (i.e., all certificates
>     except those containing fixed Diffie-Hellman parameters).

I believe when client's `Certificate` message is empty (absolutely valid case) then it has **no** signing capability and Ktor must not send `CertificateVerify`. So I've added a check before sending `CertificateVerify`. Before the check was added Ktor would throw NPE because `chain` variable was actually `null`. 

I also had to guard `output.send` because it started to fail with `kotlinx.coroutines.channels.ClosedSendChannelException: Channel was closed` exception sometimes due to channel after `TLSException`.

ktor-network/ktor-network-tls/jvm/test/io/ktor/network/tls/tests/ConnectionTests.kt
```diff
@@ -66,4 +79,70 @@ class ConnectionTests {
...
+        val keyStoreFile = File("build/temp.jks")
+        val keyStore = generateCertificate(keyStoreFile, algorithm = "SHA256withRSA", keySizeInBits = 4096)
+        val chain1 = keyStore.getCertificateChain("mykey").toList() as List<X509Certificate>
+        val certs = chain1.toList().toTypedArray()
+        val password = "changeit".toCharArray()
+        val pk = keyStore.getKey("mykey", password) as PrivateKey
+        val trustManagerFactory = TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm()) .also { it.init(keyStore) }
+        val workerGroup: EventLoopGroup = NioEventLoopGroup()
+        try {
+            val b = ServerBootstrap()
```
e5l, line 94: One last little note. Could you avoid using shortcuts and single letters in variable names?

e5l (changes requested):
LGTM, except code style

ktor-network/ktor-network-tls/jvm/test/io/ktor/network/tls/tests/ConnectionTests.kt
andrey-vasilyev: Thanks. I am sorry, not sure what `shortcuts` you are referring to, could you please elaborate? Variable names fixed.

e5l: that's what I mean. thanks, lgtm

andrey-vasilyev:
Just want to check if there is anything I need to do to merge this?

ElbowBaggins:
@e5l I also have this issue. Is @andrey-vasilyev's change ready? The sooner this gets merged the sooner I also get unblocked.

andrey-vasilyev:
Dear ktor team, could you please help us understand what blocks the merge of this PR?

e5l (approved):

e5l:
Sorry for the delay, will be released with `1.4.2`

rsinukov (approved):

andrey-vasilyev:
Great! Thanks you!

hfhbd:
I was about to create a new pull request, thanks

cy6erGn0m:
@e5l @rsinukov I wonder why had you passed these "guards"? They definitely will lead to random behaviour due to race. Shouldn't we let it crash instead? Why suppress negotiation errors here? What a strange idea to silently create a TLS session with closed channels. And couldn't it crash anyway somewhere else before or after?

cy6erGn0m:
Would you mind if I elimianate this strange behaviour?

andrey-vasilyev:
@cy6erGn0m I believe you are correct on all accounts. My bad to add these stupid guards. Thanks for pointing this out. 

I wonder would it be appropriate to wrap exception rethrowing in your change into `TLSException`?

cy6erGn0m:
@andrey-vasilyev yes, I did it in the next commit, will fixup later after review

andrey-vasilyev:
Cool. I was asking partly because I believe `ConnectionTests.clientCertificatesAuthTest` should not flicker after your change and `TLSException` wrapping. Might be worth checking.

PR #2072, opened 2020-09-16
Fix logging of body in native-mt

rsinukov:
**Subsystem**
Client/Server

**Motivation**
[Empty body in response using macosx64 target](https://youtrack.jetbrains.com/issue/KTOR-479)

**Solution**
Fixed in coroutines_version=1.3.9-native-mt-2. Added tests to verify


ktor-utils/common/src/io/ktor/util/ByteChannels.kt
```diff
@@ -15,18 +15,20 @@ private const val CHUNK_BUFFER_SIZE = 4096L
...
-    coroutineScope.launch {
+    GlobalScope.launch {
         try {
             while (!isClosedForRead) {
-                this@split.readRemaining(CHUNK_BUFFER_SIZE).use { chunk ->
-                    listOf(
-                        async { first.writePacket(chunk.copy()) },
-                        async { second.writePacket(chunk.copy()) }
-                    ).awaitAll()
+                supervisorScope {
```
e5l, line 25: Does it mean that we changed the `split` behaviour? The fail of single channel led to fail the whole split

e5l (changes requested):
Please respond the comments above

ktor-utils/common/src/io/ktor/util/ByteChannels.kt
rsinukov: reverted changes because the original issue is fixed in coroutines

e5l:
lgtm

e5l (approved):

Issue #1909, opened 2020-05-30
configureBootstrap hook overwritten by Ktor settings


carbotaniuman:
**Subsystem**
ktor-server-netty

**Is your feature request related to a problem? Please describe.**
I'm trying to add some GlobalTrafficShapingHandler to ktor, but I cannot do anything because Ktor overwrites what I am trying to do in the `configureBootstrap` lambda. I cannot set the childHandler or the channelFactory , or even influence what they do in any way. I cannot touch childHandler nor channelFactory (it's set via the `channel()` call).

```
ServerBootstrap().apply {
	configuration.configureBootstrap(this)
	group(connectionEventGroup, workerEventGroup)
	channel(connectionEventGroup.channel.java)
	childHandler(
		NettyChannelInitializer(
			pipeline, environment,
			callEventGroup, engineDispatcherWithShutdown, dispatcherWithShutdown,
			connector,
			configuration.requestQueueLimit,
			configuration.runningLimit,
			configuration.responseWriteTimeoutSeconds,
			configuration.requestReadTimeoutSeconds,
			configuration.httpServerCodec
		)
	)
}
```

**Describe the solution you'd like**
I'd want a way to expose the underlying NettyPipeline, or at least have some way to meaninfully use the `configureBootstrap` hook. I would also take a hook called in `override fun initChannel(ch: SocketChannel) {` as that would let me do the configuration I want.

**Motivation to include to ktor**
Ktor already takes effort to expose this hook, but I've found the things I really want to do with this hook aren't allowed.

carbotaniuman:
I was trying to extend the classes and modify stuff that way, but the Ktor classes are closed and the implementations use internal variables and methods that I cant reach.

oleg-larshin:
Please check the <a href="https://youtrack.jetbrains.com/issue/KTOR-356">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

e5l:
Merged

PR #2073, opened 2020-09-17
Fix KTOR-356. allow custom channelHandler to be added in netty pipeline

thiyagu06:
**Subsystem**
Server, ktor-server, ktor-netty

**Motivation**
I'm working on project where we need to add few organization cross cutting channel handler. Current implementation doesn't expose any ways to add our channel handler.
https://github.com/ktorio/ktor/issues/1909

Fix for [KTOR-356](https://youtrack.jetbrains.com/issue/KTOR-356).

**Solution**

provide a way to configure netty pipeline with user provided custom `channelHandler`.  The user provided channel handler will be added in the channel pipeline along with the default channelHandler added [here].(https://github.com/ktorio/ktor/blob/master/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyChannelInitializer.kt#L94).

```
embeddedServer(Netty, port = configuration.server.port, configure = {
        channelPipelineConfig = {
            addAfter("codec","trafficMonitoring", TrafficMonitoringHandler())
        }
    }) {
        module()
    }.start(true)
```

### DIFF
--- a/ktor-features/ktor-auth/build.gradle
+++ b/ktor-features/ktor-auth/build.gradle
@@ -1,3 +1,7 @@
+/*
+ * Copyright 2014-2020 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
 description = ''
 
 kotlin {
@@ -8,6 +12,7 @@ kotlin {
         }
         jvmTest.dependencies {
             api project(':ktor-client:ktor-client-cio')
+            api project(':ktor-client:ktor-client-mock')
             api project(':ktor-server:ktor-server-test-host')
         }
     }


thiyagu06:
@e5l @cy6erGn0m can you please help me to review this? We really need this kind of flexibility when configuring netty pipeline in our project to unblock us from using ktor

e5l (changes requested):
Could you change the base branch to `1.5.0` and add the test?

e5l (approved):

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyChannelInitializer.kt
```diff
@@ -36,7 +36,8 @@ public class NettyChannelInitializer(
     private val runningLimit: Int,
     private val responseWriteTimeout: Int,
     private val requestReadTimeout: Int,
-    private val httpServerCodec: () -> HttpServerCodec
+    private val httpServerCodec: () -> HttpServerCodec,
+    private val channelPipelineConfig: ChannelPipeline.() -> Unit
```
cy6erGn0m, line 40: I would prefer to add a fallback constructor for those who call it directly for some reason.

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationEngine.kt
```diff
@@ -67,6 +67,11 @@ public class NettyApplicationEngine(
          * User-provided function to configure Netty's [HttpServerCodec]
          */
         public var httpServerCodec: () -> HttpServerCodec = ::HttpServerCodec
+
+        /*
```
cy6erGn0m, line 71: kdoc format: `/**` at the first line and `*/` at the last

ktor-server/ktor-server-netty/jvm/test/io/ktor/tests/server/netty/NettyEngineTest.kt
```diff
@@ -48,3 +49,21 @@ class NettySustainabilityTest :
...
+class NettyCustomChannelPipelineConfigurationTest :
+    HttpServerTestSuite<NettyApplicationEngine, NettyApplicationEngine.Configuration>(Netty) {
+    init {
+        enableSsl = true
+    }
+
+    override fun configure(configuration: NettyApplicationEngine.Configuration) {
+        configuration.shareWorkGroup = true
+        configuration.channelPipelineConfig = {
+            addLast("passThroughHandler", object : ChannelDuplexHandler() {
```
cy6erGn0m, line 62: Can't we check somehow that the config actually makes any difference? Like checking if the handler was invoked or something?  Should we actually run all integration tests against the modified netty server?

cy6erGn0m (changes requested):

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyChannelInitializer.kt
thiyagugk: Instead of having another constructor, I feel adding a default value to the parameter is more than sufficient.  Any comment?
`private val channelPipelineConfig: ChannelPipeline.() -> Unit = {}`

ktor-server/ktor-server-netty/jvm/test/io/ktor/tests/server/netty/NettyEngineTest.kt
thiyagu06: yes.. good point. No need run the entire test suite. added a new test to verify the configured netty handler invocation.

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyChannelInitializer.kt
thiyagu06: added a default parameter to the constructor.

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationEngine.kt
thiyagu06: Fixed

thiyagu06:
@cy6erGn0m can you help me with this?

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyChannelInitializer.kt
cy6erGn0m: Unfortunately, it doesn't help because the binary compatibility is broken here. We need a secondary constructor with the old signature. This is because of our compatibility guarantees. After doing this run `./gradlew apiDump` and ensure that we only add new signatures and don't remove/change anything.

cy6erGn0m (changes requested):
@thiyagu06 Sorry for the long response. Could you please fix the constructor issue?

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyChannelInitializer.kt
thiyagu06: @cy6erGn0m , sorry for the delayed response. I will fix it during this week

thiyagu06:
@cy6erGn0m  somehow, I messed with my commits of this PR and I see due for the release of 1.5.0 is sooner. I'm planning to create a new PR based out of master branch. Is that okay for you?

PR #2074, opened 2020-09-18
KTOR-973 Allow freezing of HttpClient

e5l:
Fixes [KTOR-973](https://youtrack.jetbrains.com/issue/KTOR-973)


ktor-client/ktor-client-core/common/src/io/ktor/client/features/HttpClientFeature.kt
```diff
@@ -43,6 +43,8 @@ public fun <B : Any, F : Any> HttpClient.feature(feature: HttpClientFeature<B, F
  * @throws [IllegalStateException] if [feature] is not installed.
  */
 public operator fun <B : Any, F: Any> HttpClient.get(feature: HttpClientFeature<B, F>): F {
+    feature(feature)?.let {  return it }
```
Stexxe, line 46: I don't know is it idiomatic in Kotlin or not but this `.let {  return it }` hides the intent of early return.

ktor-client/ktor-client-core/common/src/io/ktor/client/features/HttpRedirect.kt
```diff
@@ -19,30 +20,39 @@ private val ALLOWED_FOR_REDIRECT: Set<HttpMethod> = setOf(HttpMethod.Get, HttpMe
  * [HttpClient] feature that handles http redirect
  */
 public class HttpRedirect {
+    private val _checkHttpMethod = atomic(true)
```
Stexxe, line 23: Is it in our code style to prefix private members with an underscore?

ktor-client/ktor-client-core/common/src/io/ktor/client/utils/SharedCollections.kt
```diff
@@ -0,0 +1,24 @@
...
+
+import io.ktor.util.*
+import io.ktor.util.collections.*
+
+internal fun <K : Any, V : Any> sharedMap(): MutableMap<K, V> {
+    if (PlatformUtils.IS_NATIVE) {
+        return ConcurrentMap()
+    }
+
+    return mutableMapOf()
```
Stexxe, line 15: Is a mutable map thread-safe and can be shared safely?

ktor-client/ktor-client-tests/posix/test/io/ktor/client/tests/HttpClientFreezingTest.kt
```diff
@@ -0,0 +1,23 @@
...
+import io.ktor.client.tests.utils.*
+import io.ktor.utils.io.*
+import kotlin.test.*
+
+class HttpClientFreezingTest : ClientLoader() {
+
+    @Test
+    fun testRequestWithFrozenClient() = clientTests {
+        test { client ->
+            client.makeShared()
```
Stexxe, line 18: Will the clients that use `HttpClient` also call `makeShared` or is it just for testing?

ktor-utils/api/ktor-utils.api
```diff
@@ -491,8 +491,10 @@ public final class io/ktor/util/collections/ConcurrentList : java/util/List, kot
 	public fun clear ()V
 	public fun contains (Ljava/lang/Object;)Z
 	public fun containsAll (Ljava/util/Collection;)Z
+	public fun equals (Ljava/lang/Object;)Z
```
Stexxe, line 494: Isn't it a public API change?

ktor-utils/common/test/io/ktor/util/PipelineContractsTest.kt
```diff
@@ -166,6 +170,11 @@ class PipelineContractsTest {
 
     @Test
     fun executePipelineWithSuspension() {
+        /**
+         * Continuation doesn't support freeze in native.
+         */
+        if (PlatformUtils.IS_NATIVE) return
```
Stexxe, line 176: Maybe we should explicitly skip a test if the platform is native

Stexxe (changes requested):
I see only scarce `PipelineContractsTest.kt` and `ConcurrentListTest`. Where are unit tests for the rest changes, e.g. for changes in `ktor-utils/common/src/io/ktor/util/pipeline/Pipeline.kt`?

ktor-client/ktor-client-core/common/src/io/ktor/client/features/HttpRedirect.kt
e5l: Yep

ktor-client/ktor-client-core/common/src/io/ktor/client/utils/SharedCollections.kt
e5l: `Shared` doesn't mean `Concurrent` here.

e5l: This configuration shouldn't be used concurrently but can be accessed from the different threads, so it should be shared on native and can be usual on the JVM

ktor-client/ktor-client-tests/posix/test/io/ktor/client/tests/HttpClientFreezingTest.kt
e5l: It's just for testing cases when user share client in the different coroutine dispatcher or create a client on the top-level scope

ktor-utils/api/ktor-utils.api
e5l: Nope, the `equals` method is present by default. The difference is that we `override` it

ktor-utils/common/test/io/ktor/util/PipelineContractsTest.kt
e5l: How should we do it? Kotlin test(and multiplatform) doesn't support the test exclusion for the platform.

ktor-client/ktor-client-core/common/src/io/ktor/client/features/HttpClientFeature.kt
e5l: Agree, will change

ktor-client/ktor-client-core/common/src/io/ktor/client/utils/SharedCollections.kt
Stexxe: For the native, it's called`ConcurrentMap`. Maybe we should think about a more appropriate name?

ktor-utils/common/test/io/ktor/util/PipelineContractsTest.kt
Stexxe: is there in Kotlin test something similar to https://stackoverflow.com/questions/1689242/conditionally-ignoring-tests-in-junit-4?

ktor-client/ktor-client-core/common/src/io/ktor/client/utils/SharedCollections.kt
e5l: The native map is `Concurrent`(synchronized) and  `ConcurrentMap` is shared. We have no separate `shared` map in native, so it's ok to use concurrent here.

ktor-utils/common/test/io/ktor/util/PipelineContractsTest.kt
e5l: We have no JUnit in common, only the Kotlin Test, and as I remember it doesn't support test ignore.

e5l: Here is all we have in common: https://kotlinlang.org/api/latest/kotlin.test/kotlin.test/ 
So annotations only, andI'm not sure if they are working in kotlin-native

cy6erGn0m (changes requested):

ktor-utils/common/src/io/ktor/util/collections/ConcurrentMap.kt
```diff
@@ -22,20 +22,16 @@ public class ConcurrentMap<Key : Any, Value : Any>(
     private val lock: Lock = Lock(),
     initialCapacity: Int = INITIAL_CAPACITY
 ) : MutableMap<Key, Value> {
+    private var table by shared(SharedList<SharedForwardList<MapNode<Key, Value>>>(initialCapacity))
+    private var insertionOrder by shared(SharedForwardList<MapNode<Key, Value>>())
+
+    private val _size = atomic(0)
+    private val loadFactor get() = _size.value / table.size
```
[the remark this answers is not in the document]
e5l: Thanks, will fix

ktor-utils/common/src/io/ktor/util/collections/internal/SharedForwardList.kt
```diff
@@ -6,64 +6,41 @@ package io.ktor.util.collections.internal
...
     override fun hasNext(): Boolean = current?.item != null
 
     override fun next(): T {
-        last = current
-        current = current?.next
-        skipRemoved()
-
-        return last?.item!!
+        previous = current
+        return previous?.item!!
```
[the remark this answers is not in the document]
e5l: +1

ktor-utils/common/src/io/ktor/util/collections/internal/SharedForwardList.kt
```diff
@@ -6,64 +6,41 @@ package io.ktor.util.collections.internal
...
-    var current by shared(value)
-
-    init {
-        skipRemoved()
-    }
+private class ForwardListIterator<T>(head: ForwardListNode<T>) : MutableIterator<T> {
+    var previous by shared<ForwardListNode<T>?>(head)
+    val current: ForwardListNode<T>? get() = previous?.next
 
     override fun hasNext(): Boolean = current?.item != null
```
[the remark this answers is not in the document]
e5l: Fixed in a later commit

ktor-utils/common/src/io/ktor/util/pipeline/PhaseContent.kt
```diff
@@ -0,0 +1,86 @@
...
+
+    @Suppress("UNCHECKED_CAST")
+    constructor(
+        phase: PipelinePhase,
+        relation: PipelinePhaseRelation
+    ) : this(phase, relation, SharedArrayList as MutableList<PipelineInterceptor<TSubject, Call>>) {
+        check(SharedArrayList.isEmpty()) { "The shared empty array list has been modified" }
+    }
+
+    var shared: Boolean by shared(true)
```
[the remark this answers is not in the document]
e5l: Yep, that's what we have

ktor-utils/common/src/io/ktor/util/pipeline/Pipeline.kt
```diff
@@ -5,203 +5,85 @@
...
-//    }
+    @InternalAPI
+    public val isEmpty: Boolean
+        get() = interceptorsQuantity == 0
 
-    // array of phases or phase contents
-    private val phasesRaw: ArrayList<Any> = phases.mapTo(ArrayList(phases.size + 1)) {
-        it
-    }
+    private val _interceptors: AtomicRef<List<suspend PipelineContext<TSubject, TContext>.(TSubject) -> Unit>?> =
```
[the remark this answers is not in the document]
e5l: `interceptors` was `Volatile` before and we don't use any atomic-specific operations, it should be the same

cy6erGn0m (approved):

Stexxe (approved):

rsinukov (approved):

PR #2078, opened 2020-09-22
Fix wrong session id clean

cy6erGn0m:
We are sending expired cookies in the case when a session id is wrong

**Subsystem**
 ktor-server-core

**Motivation**
[KTOR-1007](https://youtrack.jetbrains.com/issue/KTOR-1007)
This PR corrects the wrong behaviour in the case when a client sends a wrong session ID through cookies. This could happen if a session is expired or after the server restart with in-memory session storage.

**Solution**
Remove wrong session ids by sending cookie expiration so a client stops sending wrong IDs


e5l (approved):

e5l:
LGTM, it would be nice if you also format the commit-name and put issue link in the PR description

PR #2079, opened 2020-09-22
Improve missing server session log message (KTOR-776)

cy6erGn0m:
NoSuchElementException prefix printed by Exception.toString() is too confusing
and makes impression that something is broken while it should simply explain why
there is no session in spite of that there is a session id.

**Subsystem**
ktor-server-core


e5l (approved):

cy6erGn0m:
Merged

PR #2085, opened 2020-09-22
KTOR-1067 Fix kotlin-native intialization order

e5l:
Fixes [KTOR-1067](https://youtrack.jetbrains.com/issue/KTOR-1067)


ktor-utils/common/src/io/ktor/util/collections/ConcurrentList.kt
```diff
@@ -139,7 +139,7 @@ public class ConcurrentList<T> : MutableList<T> {
         override fun next(): T = this@ConcurrentList[current++]
 
         override fun remove() {
-            removeAt(current)
+            removeAt(current - 1)
```
cy6erGn0m, line 142: Shouldn't we fix `curent` after removal? Consider the folloing scenario (add a test like this):

```
    // given l is ConcurrentList of 1, 2, 3
    val it = l.listIterator()

    println(it.next()) // 1
    println(it.next()) // 2
    it.remove()
    println(it.next()) // crash: index out of bounds, should be 3
```

cy6erGn0m (changes requested):

Stexxe (approved):

ktor-utils/common/src/io/ktor/util/collections/ConcurrentList.kt
e5l: nice catch, thanks

e5l:
Checked coverage and fixed other one more problem

cy6erGn0m (approved):

PR #2088, opened 2020-09-23
KTOR-835 Respond 500 in engines on not caught errors

rsinukov:
**Subsystem**
Server

**Motivation**
[Implement proper unhandled exception handling strategy](https://youtrack.jetbrains.com/issue/KTOR-835)

**Solution**
Every engine now handles exception in the pipeline with a simple `try/catch` block. In case of failure, they log the error and respond with 500 status code.

### DIFF
--- a/ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/DefaultEnginePipeline.kt
+++ b/ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/DefaultEnginePipeline.kt
@@ -53,13 +53,13 @@ public fun defaultEnginePipeline(environment: ApplicationEnvironment): EnginePip
     return pipeline
 }
 
-@InternalAPI
+@EngineAPI
 public suspend fun handleFailure(call: ApplicationCall, error: Throwable) {
     logError(call, error)
     tryRespondError(call, defaultExceptionStatusCode(error) ?: HttpStatusCode.InternalServerError)
 }
 
-@InternalAPI
+@EngineAPI
 public suspend fun logError(call: ApplicationCall, error: Throwable) {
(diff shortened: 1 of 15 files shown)


cy6erGn0m:
@rsinukov oops, could you please rebase to fix "trailing slash fix" leaked into this PR :)

rsinukov:
@cy6erGn0m done, thanks!

ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/DefaultEnginePipeline.kt
```diff
@@ -61,6 +56,21 @@ public fun defaultEnginePipeline(environment: ApplicationEnvironment): EnginePip
     return pipeline
 }
 
+@InternalAPI
```
cy6erGn0m, line 59: So these are more like `@EngineAPI` rather than internal, so one who is implementing an engine, should use this function to handler errors, right?

cy6erGn0m (approved):

ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/DefaultEnginePipeline.kt
rsinukov: Hm, I reused the same visibility as before, but it makes sense to have them visible for other engines. Will update.

cy6erGn0m:
And don't forget to update apidump as well

e5l:
Could you re-run tests, please? I can't see test results.

rsinukov:
@e5l done!

e5l (approved):

PR #2090, opened 2020-09-24
Avoid security constants in tests 

cy6erGn0m:
**Subsystem**
ktor-auth (Server)

**Motivation**
[KTOR-1078](https://youtrack.jetbrains.com/issue/KTOR-1078)  Avoid security constants in tests 
In tests, we use security constant and unsafe/vulnurable features. 
This is fine for tests, but the problem is that users look into tests for usage samples and may accidentally copy-paste such code lines. But they are not suitable for production and using them lead to vulnurabilities. 

**Solution**
- Avoid constants when possible
- Comment code lines to warn users from blind copy-pasting


rsinukov (approved):

cy6erGn0m:
@Stexxe I'll merge it, let me know if there are any objections: I'll create a separate PR if necessary.

Issue #2006, opened 2020-07-23
Ktor show startup duration


halimpuckjava:
Hi

Like spring boot why not show how mush ktor take time to startup

PR #2093, opened 2020-09-25
feat: log startup time

pavelgordon:
**Subsystem**
Server, ktor-server-host-common

**Motivation**
There is an issue https://github.com/ktorio/ktor/issues/2006. After consideration, I decided that it might be useful to log startup time.

Also, I wanted to contribute to open source, and this issue looks like a good first issue to start with.

**Solution**
`init` block in `BaseApplicationEngine` looks like a perfect place to log the information about startup time along with host and port(which were present already). 

On receiving the `ApplicationStarted` event -  the information about startup time is logged.

I know that there are other possible solutions - e.g. using StopWatch, but was not sure if bringing an additional library worth the drill. What do you think?

### DIFF
--- a/ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/BaseApplicationEngine.kt
+++ b/ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/BaseApplicationEngine.kt
@@ -5,6 +5,8 @@
 package io.ktor.server.engine
 
 import io.ktor.application.*
+import java.time.*
+import java.time.Duration.*
 
 /**
  * Base class for implementing [ApplicationEngine]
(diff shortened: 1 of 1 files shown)


cy6erGn0m:
What is the minimal android sdk requited for java.time.Duration?

pavelgordon:
Yep, I had the same idea so I decided to change implementation to use System.nanoTime because java.time.Duration is available in JDK 8, and according to [this link](https://developer.android.com/reference/java/time/Duration) - from API level 26.

I removed the imports.

ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/BaseApplicationEngine.kt
```diff
@@ -35,8 +36,10 @@ public abstract class BaseApplicationEngine(
             it.sendPipeline.installDefaultTransformations()
         }
         environment.monitor.subscribe(ApplicationStarted) {
+            val initializeFinishedAt = System.nanoTime()
             environment.connectors.forEach {
                 environment.log.info("Responding at ${it.type.name.toLowerCase()}://${it.host}:${it.port}")
+                environment.log.info("Startup time: ${(initializeFinishedAt - initializeStartedAt) / 1_000_000} ms")
```
cy6erGn0m, line 42: Sorry, didn't noticed that but why do we print the same message for evey connector?

cy6erGn0m (changes requested):

cy6erGn0m:
@pavelgordon could you please also rebase it?

PR #2094, opened 2020-09-28
HttpBenchmarkClient: Fix deprecated response.body() - moved to val

SebastianAigner:
**Subsystem**
HTTP Benchmark Client

**Motivation**
When checking out the current master branch and running `gradlew build`, the project fails with a compilation error: "Using 'body(): ResponseBody?' is an error. moved to val".

**Solution**
This changes the function invocation to a property access, which fixes the deprecation compilation error.


e5l (approved):
lgtm

PR #2096, opened 2020-09-30
KTOR-1092 Improve curl error handling

cy6erGn0m:
**Subsystem**
ktor-client-curl

**Motivation**
[KTOR-1092](https://youtrack.jetbrains.com/issue/KTOR-1092)
When curl fails to do a request, no addition information provided, there is no even any meaninful message describing the issue

**Solution**
Use libcurl's [curl_easy_strerror](https://curl.haxx.se/libcurl/c/curl_easy_strerror.html) function to get an error message.

**Problems**
- Not sure how do I test it since error messages may vary
- Not clear, if the function provides localized messages or english only

### DIFF
--- a/ktor-client/ktor-client-curl/posix/src/io/ktor/client/engine/curl/internal/CurlMultiApiHandler.kt
+++ b/ktor-client/ktor-client-curl/posix/src/io/ktor/client/engine/curl/internal/CurlMultiApiHandler.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2019 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ * Copyright 2014-2020 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
  */
 
 package io.ktor.client.engine.curl.internal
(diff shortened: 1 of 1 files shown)


ktor-client/ktor-client-curl/posix/src/io/ktor/client/engine/curl/internal/CurlMultiApiHandler.kt
```diff
@@ -217,10 +217,12 @@ internal class CurlMultiApiHandler : Closeable {
                         )
                     }
 
+                    val errorMessage = curl_easy_strerror(result)?.toKStringFromUtf8()
```
rsinukov, line 220: will timeout exception produce an error here? if yes, can you add such a test?

cy6erGn0m: No, it will be handled before this line (line 214 actually) by checking for the particular timeout error code. I believe we already have timeout tests so no need to introduce new

rsinukov: Are there any other errors that we can emulate to test this change?

cy6erGn0m: Yes, we could try to connect to a free port to emulate connection refused. On my machine I get the message `"Couldn't connect to server"` but I can't be sure that it won't change with difference systems/locales. I only can test that there are some characters at the corresponding position.

e5l (approved):

rsinukov (approved):

cy6erGn0m:
Merged

PR #2097, opened 2020-09-30
KTOR-349 Fix ClassCastException during server shutdown

cy6erGn0m:
**Subsystem**
ktor-server-cio

**Motivation**
[KTOR-349](https://youtrack.jetbrains.com/issue/KTOR-349)
Server coroutine machinery crashes at server shutdown if there are active connections (like open websockets).


e5l (approved):
Sorry for the delay, LGTM.
Actually, I don't have ideas how to rewrite the test

rsinukov (approved):
LGTM

PR #2098, opened 2020-09-30
KTOR-715 Provide access token request interceptor

cy6erGn0m:
**Subsystem**
ktor-auth (Server)

**Motivation**
Most of the OAuth providers do require `state` parameter to be passed when requesting for an access token. However, for some reason, Dropbox prohibits sending it so people need to write hacks to avoid this parameter that we are always sending.

**Solution**
[KTOR-715](https://youtrack.jetbrains.com/issue/KTOR-715)
This will provide the ability to intercept requests and remove the state parameter if needed. 
The reason why I don't want a switch for this particular case is that this is potentially insecure to not send the state so it shouldn't be cheap to do that. 

**Problems**
The disadvantage is that intercepting post body is not that easy (see the corresponding test). 
Adding any options is a change that requires 1.5.x

### DIFF
--- a/ktor-features/ktor-auth/jvm/src/io/ktor/auth/OAuth.kt
+++ b/ktor-features/ktor-auth/jvm/src/io/ktor/auth/OAuth.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2019 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ * Copyright 2014-2020 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
  */
 
 package io.ktor.auth
@@ -8,9 +8,9 @@ import io.ktor.application.*
 import io.ktor.client.*
 import io.ktor.client.request.*
 import io.ktor.http.*
-import io.ktor.util.pipeline.*
 import io.ktor.response.*
 import io.ktor.util.*
+import io.ktor.util.pipeline.*
 import kotlinx.coroutines.*
 import org.slf4j.*
 import java.io.*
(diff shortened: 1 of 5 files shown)


ktor-features/ktor-auth/jvm/src/io/ktor/auth/OAuth.kt
```diff
@@ -79,7 +79,9 @@ public sealed class OAuthServerSettings(public val name: String, public val vers
         public val accessTokenUrl: String,
 
         public val consumerKey: String,
-        public val consumerSecret: String
+        public val consumerSecret: String,
+
+        public val accessTokenInterceptor: HttpRequestBuilder.() -> Unit = {}
```
e5l, line 84: The purpose of this callback is not clear to the end-users: when it will be called and why I need to pass it.

e5l: btw It changes public API and should be for `1.5.0`

e5l (changes requested):

ktor-features/ktor-auth/jvm/src/io/ktor/auth/OAuth.kt
cy6erGn0m: Yes, this is for tuning request and people usually shouldn't use it unless something unusual happens. I would introduce a switch instead, but how many switches do we need? And also disabling status parameter is potentially dangerous in terms of security so not sure if it should be cheap in spite of that it is Dropbox that violates policy.

e5l (approved):

rsinukov (approved):

cy6erGn0m:
Merged into 1.5.0

PR #2099, opened 2020-10-01
KTOR-1091 Fix netty authority parser to process missing port properly

cy6erGn0m:
**Subsystem**
ktor-server-netty

**Motivation**
[KTOR-1091](https://youtrack.jetbrains.com/issue/KTOR-1091) Netty: NumberFormatException when make a request via HTTP/2 without explicit port 

**Solution**
Handle this case and write unit tests

### DIFF
--- a/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/http2/Http2LocalConnectionPoint.kt
+++ b/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/http2/Http2LocalConnectionPoint.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2019 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ * Copyright 2014-2020 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
  */
 
 package io.ktor.server.netty.http2
@@ -9,9 +9,9 @@ import io.netty.handler.codec.http2.*
 import java.net.*
 
 internal class Http2LocalConnectionPoint(
-                    private val nettyHeaders: Http2Headers,
-                    private val localAddress: InetSocketAddress?,
-                    private val remoteAddress: InetSocketAddress?
+    private val nettyHeaders: Http2Headers,
+    private val localAddress: InetSocketAddress?,
+    private val remoteAddress: InetSocketAddress?
 ) : RequestConnectionPoint {
     override val method: HttpMethod = nettyHeaders.method()?.let { HttpMethod.parse(it.toString()) } ?: HttpMethod.Get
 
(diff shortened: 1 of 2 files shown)


Stexxe (approved):

ktor-server/ktor-server-netty/jvm/test/io/ktor/tests/server/netty/NettyHttp2LocalConnectionPointTest.kt
```diff
@@ -0,0 +1,152 @@
...
+import kotlin.test.*
+
+class NettyHttp2LocalConnectionPointTest {
+    @Test
+    fun testMethod() {
+        val point = point {
+            method("PUT")
+        }
+
+        assertEquals(HttpMethod.Put, point.method)
```
Stexxe, line 20: This and some of the following tests don't relate to changed code

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/http2/Http2LocalConnectionPoint.kt
```diff
@@ -25,12 +25,13 @@ internal class Http2LocalConnectionPoint(
...
     override val host: String
-        get() = nettyHeaders.authority()?.toString() ?: "localhost"
+        get() = nettyHeaders.authority()?.toString()?.substringBefore(":") ?: "localhost"
 
     override val port: Int
-        get() = nettyHeaders.authority()?.toString()?.substringAfter(":")?.toInt()
-                ?: localAddress?.port
-                ?: 80
+        get() = nettyHeaders.authority()?.toString()
+            ?.substringAfter(":", "-1")?.toInt()?.takeIf { it > 0 }
```
rsinukov, line 32: I would prefer not to use `-1` as a fallback, like `?.substringAfter(":", "")?.takeIf { it.isNotEmpty() }?.toInt()`, but it's up to you

rsinukov (approved):
LGTM, but one optional suggestion

ktor-server/ktor-server-netty/jvm/test/io/ktor/tests/server/netty/NettyHttp2LocalConnectionPointTest.kt
cy6erGn0m: @rsinukov since we can't agree with @Stexxe  about extra tests I've added, could you please state your thoughts: is it acceptable to commit all the tests suite or only the relevant test? The reason why I did this is that the overhead for splitting this, creating branch, ticket and PR just for few related tests.

rsinukov: I can be wrong, but I think we agreed on covering with unit tests parts of the project that you touched. IMO creating extra ticket just for tests is redundant, because they can't break any functionality.

PR #2101, opened 2020-10-01
KTOR-404 Introduce support for X-Http-Method-Override

cy6erGn0m:
**Subsystem**
ktor-server-core

**Motivation**
[KTOR-404](https://youtrack.jetbrains.com/issue/KTOR-404)
At some point, we didn't provide a feature to support `X-Http-Method-Override` header overriding HTTP method. Unfortunately, we documented that it should work while it actually doesn't.

**Solution**
Provide a feature to support it.


rsinukov (approved):
LGTM

e5l:
Let's discuss first if we need more features inside the Ktor core

e5l (approved):

e5l (changes requested):
Let's add the feature in the separate module

Issue #1892, opened 2020-05-23
Ktor Client HTTP Cache for invalid or blank Expires header 


roemchine:
**Ktor Version and Engine Used (client or server and name)**
1.3.2, client, io.ktor.client.features.cache.HttpCache

**Describe the bug**
If a server returns a blank or an invalid Expires HTTP header, the ktor client crashes if the feature HttpCache is enabled `install(HttpCache)`

**To Reproduce**
1. Write the following:
```
val articleUrl = Url("https://www.nachrichten.at/panorama/weltspiegel/30-jahre-mauerfall-sinnbild-der-freiheit;art17,3184503")
val newsClient = HttpClient(Apache) {
    install(HttpCache)

    install(UserAgent) {
        agent = "Mozilla/5.0"
    }
}
try {
    val call: HttpStatement = newsClient.get(articleSimilarityRequest.articleUrl)
    val response: HttpResponse = call.receive()
    val requestUrl = response.request.url.toString()
    val rawHtml = response.readText()
    println(rawHtml)
} catch (cause: Throwable) {
    println("Unable to download article")
} finally {
    newsClient.close()
}
```
2. If you open the URL `https://www.nachrichten.at/panorama/weltspiegel/30-jahre-mauerfall-sinnbild-der-freiheit;art17,3184503` in the browser, you can see that the Expires response header is blank
3. An exception is thrown if you run the code:
```
2020-05-23 14:41:24.124 [DefaultDispatcher-worker-2 @request#48] ERROR a.s.api.util.service.DownloadService - Unable to download article
java.lang.IllegalStateException: Failed to parse date: 
	at io.ktor.http.DateUtilsKt.fromHttpToGmtDate(DateUtils.kt:40)
	at io.ktor.client.features.cache.HttpCacheEntryKt.cacheExpires(HttpCacheEntry.kt:84)
	at io.ktor.client.features.cache.HttpCacheEntryKt.HttpCacheEntry(HttpCacheEntry.kt:18)
	at io.ktor.client.features.cache.HttpCacheEntryKt$HttpCacheEntry$1.invokeSuspend(HttpCacheEntry.kt)
	(Coroutine boundary)
	at io.ktor.client.HttpClient$1.invokeSuspend(HttpClient.kt:123)
	at io.ktor.client.engine.HttpClientEngine$install$1.invokeSuspend(HttpClientEngine.kt:68)
[...]

PR #2102, opened 2020-10-01
KTOR-364 Update HttpCacheEntry.kt

gincos:
Fix crash on empty/invalid expires header

**Subsystem**
Client

**Motivation**
 #1892 .

**Solution**
Handle the case where the Expires header is empty or invalid, and return a GMTDate() as it would if such header was not present.

### DIFF
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/features/cache/HttpCacheEntry.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/features/cache/HttpCacheEntry.kt
@@ -83,9 +83,16 @@ internal fun HttpResponse.cacheExpires(): GMTDate {
     }
 
     val expires = headers[HttpHeaders.Expires]
-    if(expires != null && expires.isNotBlank()) {
-        return expires.fromHttpToGmtDate()
-    } else return GMTDate()
+    return expires?.let {
+        // Handle "0" case faster
+        if (it == "0") return GMTDate()
+
+        return try {
+            it.fromHttpToGmtDate()
+        } catch (e: Throwable) {
+            GMTDate()
+        }
+    } ?: GMTDate()
 }
 
 internal fun HttpCacheEntry.shouldValidate(): Boolean {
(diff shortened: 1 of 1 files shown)


ktor-client/ktor-client-core/common/src/io/ktor/client/features/cache/HttpCacheEntry.kt
```diff
@@ -82,8 +82,12 @@ internal fun HttpResponse.cacheExpires(): GMTDate {
         return call.response.requestTime + maxAge * 1000L
     }
 
-    headers[HttpHeaders.Expires]?.fromHttpToGmtDate()?.let { return it }
-    return GMTDate()
+    try {
+        val expires = headers[HttpHeaders.Expires]!!.fromHttpToGmtDate() 
+        return expires
+    } catch(e: Throwable) {
```
cy6erGn0m, line 88: We definitely shouldn't simply ignore the error. I would skip missing or "blank" content but crash at an invalid content.

cy6erGn0m (changes requested):
A test is required for different cases including missing header, empty header, blank header and some random characters

gincos:
@cy6erGn0m I believe the changes I included now are more representative of what you said. If it is invalid, it will throw an exception. If it is null or blank, then is handled like not present.

cy6erGn0m:
> @cy6erGn0m I believe the changes I included now are more representative of what you said. If it is invalid, it will throw an exception. If it is null or blank, then it is handled like the header is not present.

Ok, that's it. Now we need to add tests and clean the codestyle a bit. You can simply reformat the block.

cy6erGn0m:
@gincos Sorry for mess but I've just read the RFC and found the following: 

> A cache recipient MUST interpret invalid date formats, especially the
   value "0", as representing a time in the past (i.e., "already
   expired").

So we need to ignore parsing errors as you did at first, but just without KNPE (because it is expensive).

gincos:
I believe the current version correctly handles it? 🙂

cy6erGn0m:
@gincos yes, exactly. Now we need tests

cy6erGn0m:
@gincos we are running out of time for 1.4.2 so we'll move this patch to 1.4.3 since we still have no test

cy6erGn0m:
Ok, I've made a separate PR with a test

cy6erGn0m:
Thanks for your help

gincos:
Sorry I could not do this in due time! And thanks for creating the tests yourself 😁

PR #2104, opened 2020-10-05
Provide method to add part to existing url path

rsinukov:
**Subsystem**
Common

**Motivation**
[URLBuilder to append paths to currently encoded path](https://youtrack.jetbrains.com/issue/KTOR-403)

### DIFF
--- a/ktor-http/common/src/io/ktor/http/URLBuilder.kt
+++ b/ktor-http/common/src/io/ktor/http/URLBuilder.kt
@@ -189,3 +189,30 @@ public val URLBuilder.authority: String
             append(port.toString())
         }
     }
+
+
+/**
+ * Adds [parts] to current [encodedPath]
+ */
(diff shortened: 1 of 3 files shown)


ktor-http/common/src/io/ktor/http/URLBuilder.kt
```diff
@@ -189,3 +189,30 @@ public val URLBuilder.authority: String
             append(port.toString())
         }
     }
+
+
+/**
+ * Adds [parts] to current [encodedPath]
+ */
+public fun URLBuilder.pathPart(parts: List<String>): URLBuilder {
```
cy6erGn0m, line 197: Path component? Path segment? Not sure about "part" 

Also consider reflecting muptiple components in the function name

cy6erGn0m (changes requested):

cy6erGn0m (approved):

PR #2107, opened 2020-10-06
KTOR-613 - Add transparent quality for route evaluation result

rsinukov:
**Subsystem**
server routing

**Motivation**
https://youtrack.jetbrains.com/issue/KTOR-613

**Solution**
Introduce `RouteSelectorEvaluation.qualityTransparent` constant for route items that don't have their own quality and should delegate evaluation to their children

### DIFF
--- a/ktor-server/ktor-server-core/jvm/src/io/ktor/routing/RouteSelector.kt
+++ b/ktor-server/ktor-server-core/jvm/src/io/ktor/routing/RouteSelector.kt
@@ -71,6 +71,11 @@ public data class RouteSelectorEvaluation(
         public const val qualityTailcard: Double = 0.1
 
         /**
+         * Quality of [RouteSelectorEvaluation] that doesn't have it's own priority but should delegate evaluation to it's children
+         */
+        public const val qualityTransparent: Double = -1.0
+
+        /**
          * Route evaluation failed to succeed, route doesn't match a context
          */
         public val Failed: RouteSelectorEvaluation = RouteSelectorEvaluation(false, 0.0)
--- a/ktor-features/ktor-auth/jvm/src/io/ktor/auth/Authentication.kt
+++ b/ktor-features/ktor-auth/jvm/src/io/ktor/auth/Authentication.kt
@@ -322,9 +322,9 @@ public fun Route.authenticate(
  * unless you are writing an extension
  * @param names of authentication providers to be applied to this route
  */
-public class AuthenticationRouteSelector(public val names: List<String?>) : RouteSelector(RouteSelectorEvaluation.qualityConstant) {
+public class AuthenticationRouteSelector(public val names: List<String?>) : RouteSelector(RouteSelectorEvaluation.qualityTransparent) {
     override fun evaluate(context: RoutingResolveContext, segmentIndex: Int): RouteSelectorEvaluation {
-        return RouteSelectorEvaluation.Constant
+        return RouteSelectorEvaluation(true, RouteSelectorEvaluation.qualityTransparent)
     }
 
     override fun toString(): String = "(authenticate ${names.joinToString { it ?: "\"default\"" }})"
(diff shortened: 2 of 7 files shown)


ktor-server/ktor-server-core/jvm/src/io/ktor/routing/RouteSelector.kt
```diff
@@ -32,19 +32,29 @@ public data class RouteSelectorEvaluation(
...
-         */
-        internal const val qualityQueryParameter = 1.0
+        */
+        public const val qualityQueryParameter: Double = 1.0
 
         /**
          * Quality of [RouteSelectorEvaluation] when a parameter with prefix or suffix has matched
          */
-        internal const val qualityParameterWithPrefixOrSuffix = 0.9
+        public const val qualityParameterWithPrefixOrSuffix: Double = 0.9
```
cy6erGn0m, line 41: The reason why ot was internal is that i was unsure about the exact value. Since it's a const val, we can't change the value in spite of that it's not a part of abi because a const val get inlined into user's code forever. Making it internal, my intention was to let us change the value if needed at any time. 

So do we have any reason/demand to access the constant outside of the module?

ktor-server/ktor-server-core/jvm/src/io/ktor/routing/RouteSelector.kt
```diff
@@ -201,15 +217,43 @@ public data class OptionalParameterRouteSelector(val name: String) :
...
-    RouteSelector(RouteSelectorEvaluation.qualityConstant) {
-    override fun evaluate(context: RoutingResolveContext, segmentIndex: Int): RouteSelectorEvaluation {
-        if (segmentIndex < context.segments.size && context.segments[segmentIndex] == value)
-            return RouteSelectorEvaluation.ConstantPath
-        return RouteSelectorEvaluation.Failed
+public data class PathSegmentConstantRouteSelector(val value: String) : RouteSelector(RouteSelectorEvaluation.qualityConstant) {
+
+    private var hasTrailingSlash: Boolean = false
+
+    public constructor(value: String, hasTrailingSlash: Boolean): this(value) {
```
cy6erGn0m, line 224: So what if i create a segment selector for something in the middle, not last component?
 And why we don't expose the property? Shouldn't equals consider it and why?

ktor-server/ktor-server-core/jvm/src/io/ktor/routing/RouteSelector.kt
```diff
@@ -201,15 +217,43 @@ public data class OptionalParameterRouteSelector(val name: String) :
...
+        segmentIndex == context.segments.lastIndex && hasTrailingSlash != context.hasTrailingSlash ->
+            RouteSelectorEvaluation.Failed
+        segmentIndex < context.segments.size && context.segments[segmentIndex] == value ->
+            RouteSelectorEvaluation.ConstantPath
+        else -> RouteSelectorEvaluation.Failed
     }
 
     override fun toString(): String = value
+
+    // autogenerated
```
cy6erGn0m, line 238: If we think that the property should be considered, the why don't we simply include it in the primary constructor so we'll be able to omit explicit hashCode/equals implementations?

cy6erGn0m (changes requested):
Consider making hasTrailingSlash a data class property declared in primary constructor

ktor-server/ktor-server-core/jvm/src/io/ktor/routing/RouteSelector.kt
rsinukov: Sorry, this branch was outdated and needed a rebase. This change is obsolete now.

ktor-server/ktor-server-core/jvm/src/io/ktor/routing/RouteSelector.kt
rsinukov: Sorry, this branch was outdated and needed a rebase. This change is obsolete now.

ktor-server/ktor-server-core/jvm/src/io/ktor/routing/RouteSelector.kt
rsinukov: Sorry, this branch was outdated and needed a rebase. This change is obsolete now.

cy6erGn0m (approved):

e5l (approved):

PR #2108, opened 2020-10-07
KTOR-1059 handle surrogate symbols in url path

rsinukov:
**Subsystem**
common URL parser

**Motivation**
[Emoji in URL leads to an infinite loop](https://youtrack.jetbrains.com/issue/KTOR-1059)

**Solution**
Read surrogate symbols in pairs when encoding path


e5l (approved):

e5l:
lgtm

PR #2110, opened 2020-10-08
KTOR-455 quote separator fields in header values

rsinukov:
**Subsystem**
Client/Server

**Motivation**
[Content-Disposition additional parameters should be inside quotes](https://youtrack.jetbrains.com/issue/KTOR-455)

**Solution**
Handle all separator symbols listed in [RFC](https://tools.ietf.org/html/rfc2616#section-2.2)

### DIFF
--- a/ktor-http/common/src/io/ktor/http/HeaderValueWithParameters.kt
+++ b/ktor-http/common/src/io/ktor/http/HeaderValueWithParameters.kt
@@ -6,6 +6,10 @@ package io.ktor.http
 
 import io.ktor.util.*
 
+/** Separator symbols listed in RFC https://tools.ietf.org/html/rfc2616#section-2.2 */
+private val HeaderFieldValueSeparators =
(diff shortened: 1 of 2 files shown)


ktor-http/common/src/io/ktor/http/HeaderValueWithParameters.kt
```diff
@@ -80,15 +80,27 @@ private fun String.checkNeedEscape(): Boolean {
 
     for (index in 0 until length) {
         when (this[index]) {
+            '(',
```
e5l, line 83: Could we extract the constant and write a description(small notes about sources of these symbols)?

e5l (changes requested):

ktor-http/common/src/io/ktor/http/HeaderValueWithParameters.kt
```diff
@@ -6,6 +6,10 @@ package io.ktor.http
 
 import io.ktor.util.*
 
+/** separator symbols listed in RFC https://tools.ietf.org/html/rfc2616#section-2.2 */
```
e5l, line 9: `Separator`

e5l (approved):

PR #2111, opened 2020-10-09
KTOR-1103 handle 304 response without Vary header

rsinukov:
**Subsystem**
Client cache

**Motivation**
[Client: InvalidCacheStateException when Vary header differs for 200 and 304 responses](https://youtrack.jetbrains.com/issue/KTOR-1103)

**Solution**
According to [Calculating Secondary Keys with Vary](https://tools.ietf.org/html/rfc7234#section-4.1), if 304 response doesn't have `Vary` headers, the client should compare headers listed in Vary from cached responses with new request to find a match.

### DIFF
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/features/cache/HttpCache.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/features/cache/HttpCache.kt
@@ -14,7 +14,6 @@ import io.ktor.http.*
 import io.ktor.http.content.*
 import io.ktor.util.*
 import io.ktor.util.pipeline.*
-import kotlinx.coroutines.*
 
 internal object CacheControl {
     internal val NO_STORE = HeaderValue("no-store")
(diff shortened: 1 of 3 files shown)


ktor-client/ktor-client-core/common/src/io/ktor/client/features/cache/HttpCache.kt
```diff
@@ -123,14 +122,23 @@ public class HttpCache(
...
         val url = response.call.request.url
         val cacheControl = response.cacheControl()
 
         val storage = if (CacheControl.PRIVATE in cacheControl) privateStorage else publicStorage
-        val cache = storage.find(url, response.varyKeys()) ?: return null
 
-        storage.store(url, HttpCacheEntry(response.cacheExpires(), response.varyKeys(), cache.response, cache.body))
+        val varyKeysFrom304 = response.varyKeys()
+        val cache = when {
+            varyKeysFrom304.isNullOrEmpty() -> storage.findByUrl(url)
```
e5l, line 133: Consider simplifying the expression or extracting method here

e5l (approved):
lgtm, except minor

cy6erGn0m (approved):

PR #2112, opened 2020-10-10
KTOR-742, KTOR-1122 - Add Jetty, Netty and Tomcat Client Certificate Authentication

hfhbd:
**Subsystem**
Netty, Jetty and Tomcat Server

**Motivation**
Some TLS connections are secured by client certificate authentication. This feature is implemented in Netty, Jetty and Tomcat, but it is currently not possible to use it in Ktor.

**Fixing**
[KTOR-742](https://youtrack.jetbrains.com/issue/KTOR-742) (Netty)
[KTOR-1122](https://youtrack.jetbrains.com/issue/KTOR-1122) (Jetty)

**Solution**
@andrey-vasilyev and I implemented the client certificate authentication for the server engines Netty, Jetty and Tomcat.

Note: Smaller PR of #2028

- [x] apiDump

### DIFF
--- a/ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/EmbeddedServer.kt
+++ b/ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/EmbeddedServer.kt
@@ -54,16 +54,45 @@ CoroutineScope.embeddedServer(
     configure: TConfiguration.() -> Unit = {},
     module: Application.() -> Unit
 ): TEngine {
+    val connectors: Array<EngineConnectorConfig> = arrayOf(
+        EngineConnectorBuilder().apply {
+            this.port = port
+            this.host = host
+        }
+    )
+    return embeddedServer(
+        factory = factory,
+        connectors = connectors,
+        watchPaths = watchPaths,
+        parentCoroutineContext = parentCoroutineContext,
+        configure = configure,
+        module = module
+    )
+}
+
+/**
+ * Creates an embedded server with the given [factory], listening on given [connectors]
+ * @param connectors default listening on 0.0.0.0:80
+ * @param watchPaths specifies path substrings that will be watched for automatic reloading
+ * @param parentCoroutineContext specifies a coroutine context to be used for server jobs
+ * @param configure configuration script for the engine
+ * @param module application module function
+ */
+public fun <TEngine : ApplicationEngine, TConfiguration : ApplicationEngine.Configuration>
+CoroutineScope.embeddedServer(
+    factory: ApplicationEngineFactory<TEngine, TConfiguration>,
+    vararg connectors: EngineConnectorConfig = arrayOf(EngineConnectorBuilder()),
+    watchPaths: List<String> = listOf(WORKING_DIRECTORY_PATH),
+    parentCoroutineContext: CoroutineContext = EmptyCoroutineContext,
+    configure: TConfiguration.() -> Unit = {},
+    module: Application.() -> Unit
+): TEngine {
     val environment = applicationEngineEnvironment {
         this.parentCoroutineContext = coroutineContext + parentCoroutineContext
         this.log = LoggerFactory.getLogger("ktor.application")
         this.watchPaths = watchPaths
         this.module(module)
-
-        connector {
-            this.port = port
-            this.host = host
-        }
+        this.connectors.addAll(connectors)
     }
 
     return embeddedServer(factory, environment, configure)
--- a/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyChannelInitializer.kt
+++ b/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyChannelInitializer.kt
@@ -98,8 +98,8 @@ public class NettyChannelInitializer(
                             ApplicationProtocolNames.HTTP_1_1
                         )
                     )
-                    connector.trustManagerFactory()?.let { this.trustManager(it) }
                 }
+                connector.trustManagerFactory()?.let { this.trustManager(it) }
             }
                 .build()
         }
@@ -176,7 +176,7 @@ public class NettyChannelInitializer(
     private fun EngineSSLConnectorConfig.trustManagerFactory(): TrustManagerFactory? {
         val trustStore = trustStore ?: trustStorePath?.let { file ->
             FileInputStream(file).use { fis ->
-                KeyStore.getInstance(KeyStore.getDefaultType()).also { it.load(fis, null) }
+                KeyStore.getInstance("JKS").also { it.load(fis, null) }
             }
         }
         return trustStore?.let { store ->
--- a/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyChannelInitializer.kt
+++ b/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyChannelInitializer.kt
@@ -16,9 +16,11 @@ import io.netty.handler.ssl.*
 import io.netty.handler.timeout.*
 import io.netty.util.concurrent.*
 import kotlinx.coroutines.*
+import java.io.*
 import java.nio.channels.*
 import java.security.*
 import java.security.cert.*
+import javax.net.ssl.*
 import kotlin.coroutines.*
 
 /**
@@ -96,6 +98,7 @@ public class NettyChannelInitializer(
                             ApplicationProtocolNames.HTTP_1_1
                         )
                     )
+                    connector.trustManagerFactory()?.let { this.trustManager(it) }
                 }
             }
                 .build()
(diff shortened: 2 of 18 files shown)


cy6erGn0m (changes requested):
I generally like it. However, there are things to be sorted out first.

First of all, we need to make other engines (CIO and Tomcat) produce error if the options you added were set because they don't support them. 

The other issue is that we have two options for the same thing. We actually already have similar issue. So I would simply notice in kdocs that only one of the options will be applied while the second will be ignored if both configured.

hfhbd:
> The other issue is that we have two options for the same thing. We actually already have similar issue. So I would simply notice in kdocs that only one of the options will be applied while the second will be ignored if both configured.

Do you mean `public val trustStore: KeyStore?` and `public val trustStorePath: File?`? This implementation comes from the current design, see `public val keyStore: KeyStore` and `public val keyStorePath: File?`. I dislike this design either, and would use instead a sealed class (one variable based, one file based), but this would be a breaking change.

hfhbd:
> First of all, we need to make other engines (CIO and Tomcat) produce error if the options you added were set because they don't support them.
@cy6erGn0m CIO does not support HTTPS at all and throws already an exception. Instead throwing an exception for Jetty, the certification authentication is now implemented :)

ktor-server/ktor-server-jetty/build.gradle
```diff
@@ -15,6 +15,7 @@ kotlin {
             api project(':ktor-server:ktor-server-core')
             api project(':ktor-server:ktor-server-test-host')
             api project(':ktor-server:ktor-server-test-suites')
+            api project(':ktor-server:ktor-server-tomcat')
```
cy6erGn0m, line 18: Ooops, why do we depend on tomcat in jetty?

ktor-server/ktor-server-tomcat/jvm/src/io/ktor/server/tomcat/TomcatApplicationEngine.kt
```diff
@@ -72,12 +72,17 @@ public class TomcatApplicationEngine(environment: ApplicationEngineEnvironment,
                         if (ktorConnector.keyStorePath == null) {
                             throw IllegalArgumentException("Tomcat requires keyStorePath")
                         }
+                        if (ktorConnector.trustStorePath != null) {
```
cy6erGn0m, line 75: We probably should fail if only `trustStore` is specified with no `trustStorePath`, right?

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyChannelInitializer.kt
```diff
@@ -132,6 +141,19 @@ public class NettyChannelInitializer(
         }
     }
 
+    private fun EngineSSLConnectorConfig.hasTrustStore() = trustStore != null || trustStorePath != null
+
+    private fun EngineSSLConnectorConfig.trustManagerFactory(): TrustManagerFactory? {
+        val trustStore = trustStore ?: trustStorePath?.let { file ->
+            FileInputStream(file).use { fis ->
+                KeyStore.getInstance(KeyStore.getDefaultType()).also { it.load(fis, null) }
```
cy6erGn0m, line 149: In other places, we simply use `"JKS"` constant while here we are querying JCA for this value, is it intended?

cy6erGn0m (changes requested):
Great, we are almost done!

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyChannelInitializer.kt
hfhbd: No, and fixed

ktor-server/ktor-server-tomcat/jvm/src/io/ktor/server/tomcat/TomcatApplicationEngine.kt
hfhbd: For Tomcat only:
Usecase 1: Truststore null, truststorePath null => client cert Auth not wanted
Usecase 2: Truststore set, trustStorePath null => client cert wanted, but no correct config. Currently there is NO error shown to the user. Do you want an error here?
Usecase 3: truststore null, truststorePath set => client cert wanted and correct config (for tomcat only)
Usecase 4: truststore set, truststorePath set => client cert wanted and correct config

ktor-server/ktor-server-jetty/build.gradle
hfhbd: Simply because I did not know, what's the right module for the created multi engine test :)

ktor-server/ktor-server-tomcat/jvm/src/io/ktor/server/tomcat/TomcatApplicationEngine.kt
cy6erGn0m: I think yes, we shouldn't silently ignore config, especially auth

hfhbd:
Currently, Netty fails with TLS Handshake Failure: Bad Certificate. Currently, I don't know why.

ktor-server/ktor-server-tomcat/jvm/src/io/ktor/server/tomcat/TomcatApplicationEngine.kt
hfhbd: Exception added :)

hfhbd:
I finally fixed the blocking bug with Netty: the custom trustStore was not properly set, so the custom CA was not sent as an authority.

cy6erGn0m:
@hfhbd Sorry, we broke it: could you please rebase?

hfhbd:
@cy6erGn0m I did, but I don't get it. Why are in this pull request others commits too?

hfhbd:
update base from 1.5.0 to master

ktor-server/ktor-server-jetty/build.gradle
hfhbd: Fixed. I created a abstract class in server-testing-suites and in each engine testSrc a subclass

cy6erGn0m:
> @cy6erGn0m update base from the now outdated 1.5.0 to master again

I was already about to merge it but found a bug just before. Sorry for delay

hfhbd:
@cy6erGn0m What bug? Is it still in the commits?

Its okay, but you should really make this process more transparent and provide more feedback to the contributors. For example, we did look over every PR and their changes each friday, which took less than an hour. This improved our contribution process and the contributed quality :)

ktor-server/ktor-server-tomcat/jvm/src/io/ktor/server/tomcat/TomcatApplicationEngine.kt
```diff
@@ -76,11 +76,40 @@ public class TomcatApplicationEngine(environment: ApplicationEngineEnvironment,
                             throw IllegalArgumentException("Tomcat requires keyStorePath")
                         }
 
+                        if (ktorConnector.trustStore != null && ktorConnector.trustStorePath == null) {
```
hfhbd, line 79: duplicate

cy6erGn0m (approved):

cy6erGn0m:
I was wrong about bug, sorry. Due to API changes this should be merged into 1.6.0. Could you please change the base branch and rebase it? After that I'll rerun tests and merge it

hfhbd:
@cy6erGn0m rebased. 
The code style linter is very "interesting" regarding closure parameters and results not into the same code style from the auto-formatter using IntelliJ and Kotlin official code style.

e5l:
Hey, @hfhbd. thanks for the PR. LGTM! Could you rebase it on the main branch?

hfhbd:
@e5l rebased. The code style linter is very "interesting" regarding closure parameters and results not into the same code style from the auto-formatter using IntelliJ and Kotlin official code style.

e5l:
Unfortunately, auto-format can't fix all Kotlin style guide rules, so we use ktlint to ensure it

ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/EmbeddedServer.kt
```diff
@@ -28,13 +28,13 @@ public interface ApplicationEngineFactory<out TEngine : ApplicationEngine,
...
 public fun <TEngine : ApplicationEngine, TConfiguration : ApplicationEngine.Configuration>
 embeddedServer(
-    factory: ApplicationEngineFactory<TEngine, TConfiguration>,
-    port: Int = 80,
-    host: String = "0.0.0.0",
-    watchPaths: List<String> = listOf(WORKING_DIRECTORY_PATH),
-    configure: TConfiguration.() -> Unit = {},
-    module: Application.() -> Unit
-): TEngine = GlobalScope.embeddedServer(factory, port, host, watchPaths, EmptyCoroutineContext, configure, module)
+        factory: ApplicationEngineFactory<TEngine, TConfiguration>,
```
hfhbd, line 31: @e5l https://ktor.teamcity.com/buildConfiguration/Ktor_KtorCodeStyleVerifyKtLint/34757?showLog=34757_3754_426.3754&logFilter=debug
EmbeddedServer.kt:31:5: Lint error > [parameter-list-wrapping] Unexpected indentation (expected 8, actual 4)

After fix
https://ktor.teamcity.com/buildConfiguration/Ktor_KtorCodeStyleVerifyKtLint/34758?showLog=34758_3449_106.3449&logFilter=debug
t:31:9: Lint error > [parameter-list-wrapping] Unexpected indentation (expected 4, actual 8)

Which one is correct? :D

hfhbd:
@e5l I finally fixed the code style 🎉

e5l:
Hey @hfhbd, could you rebase the branch on top of `2.0.0-eap`?

hfhbd:
@e5l rebased

Issue #1720, opened 2020-03-12
Ktor CIO client, connect timeout handling


rrva:
1. Using the following code (full example at https://github.com/rrva/ktor-client-connection-leak)
2. Run the main method repeatedly. Sometimes, all futures created will not complete . Why? Expected behavior is that all created futures will complete eventually, so `Done with all` will be printed.

```kotlin
 class Client(val fooServiceUrl: String) {

    val client = HttpClient(CIO) {
        engine {
            requestTimeout = 500
            endpoint.connectTimeout = 100
            endpoint.connectRetryAttempts = 2
        }
        install(JsonFeature) {
            serializer = KotlinxSerializer(
                Json(
                    JsonConfiguration(
                        isLenient = true,
                        ignoreUnknownKeys = true,
                        serializeSpecialFloatingPointValues = true,
                        useArrayPolymorphism = true
                    )
                )
            )
        }
    }

    suspend fun fetchFoo(ids: List<String>): List<Foo> {
            return client.request {
                header("User-Agent", "Foo")
                parameter("input", ids.joinToString(","))
                url("$fooServiceUrl/foo")
            }
    }

}

@Serializable
data class Foo(val id: String)
val client = Client("http://localhost:9090")

fun fetchFooAsync(ids:List<String>): CompletableFuture<List<Foo>> {
    return CoroutineScope(Job() + Dispatchers.IO + MDCContext()).async {
        try {
            client.fetchFoo(ids)
        } catch (e: Throwable) {
            println(e.message)
            listOf<Foo>(Foo("timeout"))
        }
    }.asCompletableFuture()
}


fun main() {
    HttpServer.create(InetSocketAddress(9090), 1000).apply {

        createContext("/foo") { http ->
            http.responseHeaders.add("Content-type", "application/json")
            http.sendResponseHeaders(200, 0)
            PrintWriter(http.responseBody).use { out ->
                out.println("""[{"id":"1"}]""")
            }
        }

        start()
    }
    println("Start")

    val requests = (1..100).map {
        fetchFooAsync(listOf(it.toString()))
    }
    for (request in requests) {
        println("Waiting for coroutine...")
        println(request.get())
        println("Done waiting")
    }

    println("Done with all")

}
```

Reproduced on 1.3.1

e5l:
Hi @rrva, thanks for the report. Could you check it with 1.3.1?

cy6erGn0m:
*1.3.2

rrva:
[...]

PR #2113, opened 2020-10-12
Fix CIO hanging due to Endpoint timeout

e5l:
Fixes [KTOR-446](https://youtrack.jetbrains.com/issue/KTOR-446), #1720

### DIFF
--- a/ktor-client/ktor-client-cio/common/src/io/ktor/client/engine/cio/Endpoint.kt
+++ b/ktor-client/ktor-client-cio/common/src/io/ktor/client/engine/cio/Endpoint.kt
@@ -16,6 +16,7 @@ import io.ktor.utils.io.core.*
 import kotlinx.atomicfu.*
 import kotlinx.coroutines.*
 import kotlinx.coroutines.channels.*
+import kotlinx.coroutines.sync.*
 import kotlin.coroutines.*
 
 internal class Endpoint(
@@ -29,46 +30,50 @@ internal class Endpoint(
     private val onDone: () -> Unit
 ) : CoroutineScope, Closeable {
     private val address = NetworkAddress(host, port)
-
+    private val lastActivity = atomic(GMTDate())
     private val connections: AtomicInt = atomic(0)
-    private val tasks: Channel<RequestTask> = Channel(Channel.UNLIMITED)
     private val deliveryPoint: Channel<RequestTask> = Channel()
     private val maxEndpointIdleTime: Long = 2 * config.endpoint.connectTimeout
 
-    private val postman = launch {
+    private val timeout = launch(coroutineContext + CoroutineName("Endpoint timeout($host:$port)")) {
         try {
             while (true) {
-                val task = withTimeout(maxEndpointIdleTime) {
-                    tasks.receive()
+                val remaining = (lastActivity.value + maxEndpointIdleTime).timestamp - GMTDate().timestamp
+                if (remaining <= 0) {
+                    break
                 }
 
-                try {
-                    if (!config.pipelining || task.request.requiresDedicatedConnection()) {
-                        makeDedicatedRequest(task)
-                    } else {
-                        makePipelineRequest(task)
-                    }
-                } catch (cause: Throwable) {
-                    task.response.completeExceptionally(cause)
-                    throw cause
-                }
+                delay(remaining)
             }
         } catch (cause: Throwable) {
         } finally {
             deliveryPoint.close()
-            tasks.close()
             onDone()
         }
     }
 
-    public suspend fun execute(
+    private suspend fun processTask(task: RequestTask) {
+        try {
+            if (!config.pipelining || task.request.requiresDedicatedConnection()) {
+                makeDedicatedRequest(task)
+            } else {
+                makePipelineRequest(task)
+            }
+        } catch (cause: Throwable) {
+            task.response.completeExceptionally(cause)
+            throw cause
+        }
+    }
+
+    suspend fun execute(
         request: HttpRequestData,
         callContext: CoroutineContext
     ): HttpResponseData {
-        val result = CompletableDeferred<HttpResponseData>()
-        val task = RequestTask(request, result, callContext)
-        tasks.offer(task)
-        return result.await()
+        lastActivity.value = GMTDate()
+        val response = CompletableDeferred<HttpResponseData>()
+        val task = RequestTask(request, response, callContext)
+        processTask(task)
+        return response.await()
     }
 
     private suspend fun makePipelineRequest(task: RequestTask) {
@@ -229,7 +234,7 @@ internal class Endpoint(
     }
 
     override fun close() {
-        tasks.close()
+        timeout.cancel()
     }
 }
 
--- a/ktor-client/ktor-client-cio/jvm/test/io/ktor/client/engine/cio/CIORequestTest.kt
+++ b/ktor-client/ktor-client-cio/jvm/test/io/ktor/client/engine/cio/CIORequestTest.kt
@@ -14,10 +14,17 @@ import io.ktor.response.*
 import io.ktor.routing.*
 import io.ktor.server.engine.*
 import io.ktor.server.netty.*
+import kotlinx.coroutines.*
+import kotlinx.coroutines.debug.junit4.*
+import org.junit.*
 import kotlin.test.*
+import kotlin.test.Test
 
 class CIORequestTest : TestWithKtor() {
     private val testSize = 2 * 1024
+
+    @get:Rule
+    override val timeout = CoroutinesTimeout.seconds(10)
 
     override val server: ApplicationEngine = embeddedServer(Netty, serverPort) {
         routing {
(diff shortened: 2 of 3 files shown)


ktor-client/ktor-client-cio/common/src/io/ktor/client/engine/cio/EngineTasks.kt
```diff
@@ -15,7 +15,9 @@ internal data class RequestTask(
     val request: HttpRequestData,
     val response: CompletableDeferred<HttpResponseData>,
     val context: CoroutineContext
-)
+) {
+    override fun toString(): String = hashCode().toString()
```
rsinukov, line 19: Is it leftover from debugging or by design? I think it's useful to see at least a request url.

rsinukov (approved):
LGTM but minor comment

ktor-client/ktor-client-cio/common/src/io/ktor/client/engine/cio/EngineTasks.kt
e5l: Fixed! Thanks!

ktor-client/ktor-client-cio/common/src/io/ktor/client/engine/cio/Endpoint.kt
```diff
@@ -30,45 +31,52 @@ internal class Endpoint(
...
+    private val mutex = Mutex()
     private val connections: AtomicInt = atomic(0)
-    private val tasks: Channel<RequestTask> = Channel(Channel.UNLIMITED)
     private val deliveryPoint: Channel<RequestTask> = Channel()
     private val maxEndpointIdleTime: Long = 2 * config.endpoint.connectTimeout
 
-    private val postman = launch {
+    val lastActivity = atomic(GMTDate())
+
+    private val timeout = launch(coroutineContext + CoroutineName("Endpoint timeout($host:$port")) {
```
cy6erGn0m, line 41: Missing close bracket in the coroutine name pattern after the port substitution

cy6erGn0m: Btw, is it safe to put a hostname into coroutine name and to thread name that sometimes goes to some native stuff and may have some effects we can't trace

ktor-network/jvm/src/io/ktor/network/selector/ActorSelectorManager.kt
```diff
@@ -32,34 +32,33 @@ public class ActorSelectorManager(context: CoroutineContext) : SelectorManagerSu
...
         launch {
-            val selector = provider.openSelector() ?: error("openSelector() = null")
+            val selector = provider.openSelector() ?: error("Failed to open selector.")
             selectorRef = selector
-            try {
+
+            selector.use {
                 try {
-                    process(mb, selector)
+                    process(messages, it)
```
cy6erGn0m, line 46: I would prefer to see `selector` instead of `it` here since the `use` block is quite big and it's not obvious, what `it` represents

ktor-client/ktor-client-cio/jvm/test/io/ktor/client/engine/cio/CIORequestTest.kt
```diff
@@ -45,4 +53,25 @@ class CIORequestTest : TestWithKtor() {
...
+                endpoint {
+                    connectTimeout = 1
+                }
+            }
+        }
+
+        test { client ->
+            for (i in 0..1000) {
+                try {
+                    client.get<String>("http://something.wrong")
```
cy6erGn0m, line 70: Is it good to make so much resolve requests, especially on TC? Don't we risk to be rate-limited here?

ktor-client/ktor-client-cio/common/src/io/ktor/client/engine/cio/Endpoint.kt
```diff
@@ -30,45 +31,52 @@ internal class Endpoint(
...
+    suspend fun execute(
         request: HttpRequestData,
         callContext: CoroutineContext
     ): HttpResponseData {
-        val result = CompletableDeferred<HttpResponseData>()
-        val task = RequestTask(request, result, callContext)
-        tasks.offer(task)
-        return result.await()
+        val response = mutex.withLock {
+            lastActivity.value = GMTDate()
```
cy6erGn0m, line 72: Should we do that inside of the block or outside (before entering `withLock`) ?

ktor-client/ktor-client-cio/common/src/io/ktor/client/engine/cio/Endpoint.kt
```diff
@@ -30,45 +31,52 @@ internal class Endpoint(
...
-                    task.response.completeExceptionally(cause)
-                    throw cause
-                }
+                delay(maxEndpointIdleTime)
+                if (lastActivity.value + maxEndpointIdleTime < GMTDate()) break
             }
         } catch (cause: Throwable) {
         } finally {
             deliveryPoint.close()
-            tasks.close()
```
cy6erGn0m, line 59: Since we have no queue anymore, we can't stop users from calling execute on a dead endpoint, right? Do we think that it's safe to allow executing requests after a timeout? I guess it should be safe but not exactly sure

cy6erGn0m (changes requested):

ktor-client/ktor-client-cio/common/src/io/ktor/client/engine/cio/Endpoint.kt
```diff
@@ -30,45 +31,52 @@ internal class Endpoint(
...
-                        makeDedicatedRequest(task)
-                    } else {
-                        makePipelineRequest(task)
-                    }
-                } catch (cause: Throwable) {
-                    task.response.completeExceptionally(cause)
-                    throw cause
-                }
+                delay(maxEndpointIdleTime)
+                if (lastActivity.value + maxEndpointIdleTime < GMTDate()) break
```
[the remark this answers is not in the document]
e5l: Ok, we can delay for `maxEndpointIddleTime - (GMTDate() - lastActivity)`

ktor-client/ktor-client-cio/common/src/io/ktor/client/engine/cio/Endpoint.kt
```diff
@@ -30,45 +31,52 @@ internal class Endpoint(
...
+
+    suspend fun execute(
         request: HttpRequestData,
         callContext: CoroutineContext
     ): HttpResponseData {
-        val result = CompletableDeferred<HttpResponseData>()
-        val task = RequestTask(request, result, callContext)
-        tasks.offer(task)
-        return result.await()
+        val response = mutex.withLock {
```
rsinukov, line 71: Is mutex fair in coroutines? If not we lose ordering, but should not be a problem, I think.

ktor-network/jvm/src/io/ktor/network/selector/ActorSelectorManager.kt
e5l: Sure

ktor-client/ktor-client-cio/jvm/test/io/ktor/client/engine/cio/CIORequestTest.kt
e5l: It's the only way to reproduce the initial issue. I think it's ok

ktor-client/ktor-client-cio/common/src/io/ktor/client/engine/cio/Endpoint.kt
e5l: Yep, it doesn't matter.

e5l: It can be reordered in many places before

ktor-client/ktor-client-cio/common/src/io/ktor/client/engine/cio/Endpoint.kt
e5l: We can drop mutex here

e5l: I replaced `channel` with mutex cause it was `id` change, but since the separate timeout coroutine, we can drop it

ktor-client/ktor-client-cio/common/src/io/ktor/client/engine/cio/Endpoint.kt
e5l: Yep, it's safe. Cause the `deliveryPoint` subscribers will gracefully shutdown.

cy6erGn0m (approved):

cy6erGn0m:
❗ @e5l  Notice the compilation error!

cy6erGn0m (changes requested):
compilation failed

cy6erGn0m (approved):

PR #2114, opened 2020-10-12
KTOR-1125 - Rename misleading CIO config property

rsinukov:
**Subsystem**

**Motivation**
[CIO: client throws ConnectTimeoutException when endpoint.connectRetryAttempts = 0](https://youtrack.jetbrains.com/issue/KTOR-1125)

**Solution**
Deprecate misleading property. Add new property and delegate the old one to the new one.

### DIFF
--- a/ktor-client/ktor-client-cio/common/src/io/ktor/client/engine/cio/CIOEngineConfig.kt
+++ b/ktor-client/ktor-client-cio/common/src/io/ktor/client/engine/cio/CIOEngineConfig.kt
@@ -77,7 +77,7 @@ public class EndpointConfig {
     /**
      * Maximum number of connection attempts.
      */
-    @Deprecated("Misleading name.", replaceWith = ReplaceWith("connectAttempts"))
+    @Deprecated("This is deprecated due to the misleading name. Use connectAttempts instead.", replaceWith = ReplaceWith("connectAttempts"))
     public var connectRetryAttempts: Int
         get() = connectAttempts
         set(value) {
@@ -86,6 +86,7 @@ public class EndpointConfig {
 
     /**
      * Maximum number of connection attempts.
+     * Note: this property affects only connection retries, but not request retries
      */
     public var connectAttempts: Int = 1
 
--- a/ktor-client/ktor-client-cio/common/src/io/ktor/client/engine/cio/CIOEngineConfig.kt
+++ b/ktor-client/ktor-client-cio/common/src/io/ktor/client/engine/cio/CIOEngineConfig.kt
@@ -16,6 +16,7 @@ public class CIOEngineConfig : HttpClientEngineConfig() {
      * [Endpoint] settings.
      */
     public val endpoint: EndpointConfig = EndpointConfig()
+
     /**
      * [https] settings.
      */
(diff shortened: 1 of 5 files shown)


ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/HeadersTest.kt
```diff
@@ -15,7 +15,7 @@ class HeadersTest : ClientLoader() {
     @Test
     fun testHeadersReturnNullWhenMissing() = clientTests {
         test { client ->
-            client.get<HttpResponse>("$TEST_SERVER/headers/").let {
```
cy6erGn0m, line 18: How do these changes regarding suffices and paths relate to the original purpose of the PR?

cy6erGn0m (changes requested):

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/HeadersTest.kt
rsinukov: Sorry, it was rebase gone wrong. Pushed proper changes.

ktor-client/ktor-client-cio/common/src/io/ktor/client/engine/cio/CIOEngineConfig.kt
```diff
@@ -76,7 +77,17 @@ public class EndpointConfig {
     /**
      * Maximum number of connection attempts.
      */
-    public var connectRetryAttempts: Int = 1
+    @Deprecated("Misleading name.", replaceWith = ReplaceWith("connectAttempts"))
```
cy6erGn0m, line 80: A deprecation message should state what is deprecated and/or why, and what should be done to migrate. For example,
"This is deprecated due to the misleading name. Use connectAttempts instead." or "This is deprecated in favour of connectAttempts". In spite of that there is a "replace-with" block, a deprecation message is shown to user in the build log and the IDE popup and should explain what is going on and why. So that one sees that the deprecation makes sense.

ktor-client/ktor-client-cio/common/src/io/ktor/client/engine/cio/CIOEngineConfig.kt
```diff
@@ -76,7 +77,17 @@ public class EndpointConfig {
...
+    public var connectRetryAttempts: Int
+        get() = connectAttempts
+        set(value) {
+            connectAttempts = value
+        }
+
+    /**
```
cy6erGn0m, line 87: I know, you just copied the message as is. But it would be nice to clarifiy a bit here. It's important to notice, that this is about connections, not requests. So no request retries caused by this option.

cy6erGn0m (changes requested):

ktor-client/ktor-client-cio/common/src/io/ktor/client/engine/cio/CIOEngineConfig.kt
rsinukov: Thanks, updated

ktor-client/ktor-client-cio/common/src/io/ktor/client/engine/cio/CIOEngineConfig.kt
rsinukov: Updated. It definitely required making a note, because I also didn't understand this before you mentioned =)

cy6erGn0m (approved):
Please remember to create a PR to update the documentation. `./clients/http-client/1engines.md:170` as far as I can see requires to be amended

PR #2115, opened 2020-10-13
KTOR-842 Fix 404 StatusPages in tests

rsinukov:
**Subsystem**
Server, Status pages, tests

**Motivation**
[StatusPages: TestEngine doesn't handle requests when no route selector matches](https://youtrack.jetbrains.com/issue/KTOR-842)

**Solution**
Move replying with 404 status on unhandled response from `defaultEnginePipeline` to `BaseApplicationEngine`, because `TestApplicationEngine` do not use `defaultEnginePipeline` 
This unifies behaviour in tests and real life

### DIFF
--- a/ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/BaseApplicationEngine.kt
+++ b/ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/BaseApplicationEngine.kt
@@ -5,6 +5,8 @@
 package io.ktor.server.engine
 
 import io.ktor.application.*
+import io.ktor.http.*
+import io.ktor.response.*
 
 /**
  * Base class for implementing [ApplicationEngine]
@@ -33,10 +35,19 @@ public abstract class BaseApplicationEngine(
             it.sendPipeline.merge(pipeline.sendPipeline)
             it.receivePipeline.installDefaultTransformations()
             it.sendPipeline.installDefaultTransformations()
+            it.installDefaultInterceptors()
         }
         environment.monitor.subscribe(ApplicationStarted) {
             environment.connectors.forEach {
                 environment.log.info("Responding at ${it.type.name.toLowerCase()}://${it.host}:${it.port}")
+            }
+        }
+    }
+
+    private fun Application.installDefaultInterceptors() {
+        intercept(ApplicationCallPipeline.Fallback) {
+            if (call.response.status() == null) {
+                call.respond(HttpStatusCode.NotFound)
             }
         }
     }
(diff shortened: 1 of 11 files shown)


ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/BaseApplicationEngine.kt
```diff
@@ -33,11 +35,20 @@ public abstract class BaseApplicationEngine(
             it.sendPipeline.merge(pipeline.sendPipeline)
             it.receivePipeline.installDefaultTransformations()
             it.sendPipeline.installDefaultTransformations()
+            it.installDefaultInterceptors()
```
cy6erGn0m, line 38: What about the engines? Don't they setup the same and how does it interfer with this?

ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/BaseApplicationEngine.kt
```diff
@@ -33,11 +35,20 @@ public abstract class BaseApplicationEngine(
...
         }
         environment.monitor.subscribe(ApplicationStarted) {
             environment.connectors.forEach {
                 environment.log.info("Responding at ${it.type.name.toLowerCase()}://${it.host}:${it.port}")
             }
         }
     }
+
+    private fun Application.installDefaultInterceptors() {
```
cy6erGn0m, line 47: We had a lot of tests relying on isHandled (or isUnhandled). I wonder, why didn't they stop working after this change?

cy6erGn0m (changes requested):

ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/BaseApplicationEngine.kt
rsinukov: Engines used to set up it by using `defaultEnginePipeline()`, but the test engine does not use it. I removed 404 handling in real engines here https://github.com/ktorio/ktor/pull/2115/files#diff-304df4306f4d507c42cc92b972a3d7415cc80ae943455cea7411d65dbf2ee0ffL37

ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/BaseApplicationEngine.kt
rsinukov: I updated `isHandled` so it is false for 404 responses here https://github.com/ktorio/ktor/pull/2115/files#diff-a3f83b71ecd565b0a1089db3230bccfabdaf774b35a2101deba655f840245bebR75

rsinukov: Also, some tests were updated to compare status with 404

e5l (approved):

cy6erGn0m (approved):

PR #2116, opened 2020-10-14
KTOR-717 Fix Apache client freeze

e5l:
Fixes [KTOR-717](https://youtrack.jetbrains.com/issue/KTOR-717)

### DIFF
--- a/ktor-client/ktor-client-apache/jvm/src/io/ktor/client/engine/apache/ApacheResponseConsumer.kt
+++ b/ktor-client/ktor-client-apache/jvm/src/io/ktor/client/engine/apache/ApacheResponseConsumer.kt
@@ -45,8 +45,9 @@ internal class ApacheResponseConsumer(
     override fun consumeContent(decoder: ContentDecoder, ioctrl: IOControl) {
         check(!waiting.value)
 
-        var result = 0
+        var result: Int = 0
         do {
+            result = 0
             channel.writeAvailable {
                 result = decoder.read(it)
             }
@@ -65,7 +66,6 @@ internal class ApacheResponseConsumer(
                 check(waiting.getAndSet(false))
                 interestController.resumeInputIfPossible()
             }
-
         }
     }
 
--- a/ktor-client/ktor-client-apache/jvm/test/io/ktor/client/engine/apache/NoOpControl.kt
+++ b/ktor-client/ktor-client-apache/jvm/test/io/ktor/client/engine/apache/NoOpControl.kt
@@ -1,0 +1,24 @@
+/*
+ * Copyright 2014-2020 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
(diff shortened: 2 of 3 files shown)


e5l:
The failing is flaky happens randomly but stable after thousands of requests. Could you tell me if you have the idea of testing the fix?

ktor-client/ktor-client-apache/jvm/src/io/ktor/client/engine/apache/ApacheResponseConsumer.kt
```diff
@@ -45,8 +45,9 @@ internal class ApacheResponseConsumer(
     override fun consumeContent(decoder: ContentDecoder, ioctrl: IOControl) {
         check(!waiting.value)
 
-        var result = 0
+        var result: Int
```
Stexxe, line 48: Could you at least add a comment about how this change fixes the problem or add some information in the commit message?

Stexxe (changes requested):

ktor-client/ktor-client-apache/jvm/src/io/ktor/client/engine/apache/ApacheResponseConsumer.kt
e5l: I manage to add a test. Is this enough?

ktor-client/ktor-client-apache/jvm/test/io/ktor/client/engine/apache/FakeControl.kt
```diff
@@ -0,0 +1,24 @@
+/*
+ * Copyright 2014-2020 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.client.engine.apache
+
+import org.apache.http.nio.*
+
+internal class FakeControl : IOControl {
```
cy6erGn0m, line 9: Please consider more descriptive name like DoNothingControl or NoOpControl

cy6erGn0m: Also, don't we already such implementation? Shouldn't we fail instead of doing nothing in the test?

cy6erGn0m (changes requested):

ktor-client/ktor-client-apache/jvm/test/io/ktor/client/engine/apache/FakeControl.kt
e5l: Will rename. Should we try using Mockk instead?

cy6erGn0m: I don't think so, not just for a single class.

rsinukov (approved):

cy6erGn0m (approved):

Stexxe (approved):

PR #2119, opened 2020-10-14
KTOR-1139 Fix CertificatePinner buildErrorMessage outputs CertificatesInfo reference

clhols:
**Subsystem**
Client iOS

**Motivation**
This PR fixes a wrong error message output when the iOS certificate pinning check fails.
https://youtrack.jetbrains.com/issue/KTOR-1139

**Solution**
Instead of outputting a reference value to the CertificatesInfo object, it now outputs "sha256/" as expected.


cy6erGn0m (approved):

e5l (approved):
Thanks for the PR! LGTM

PR #2120, opened 2020-10-15
KTOR-1101 Prevent freezing when parking on backend event loop

cy6erGn0m:
**Subsystem**
ktor-io, ktor-http, ktor-server-core

**Motivation**
[KTOR-1101](https://youtrack.jetbrains.com/issue/KTOR-1101) Unexpected freezes 

Currently, people may start doing blocking operations on a thread that doesn't allow blocking/parking. A typical example is a coroutine event loop of `runBlocking`. This particular example is already covered and supported by playing the coroutine event loop under the hood instead of just parking. 

The other example is a server engine's event loop and/or epoll/kqueue handling thread. Blocking/parking that thread leads to stopping the underlying I/O machinery and prevents server from being operational.

In many cases we prevent users from doing such unsafe stuff by dispatching user's code on a safe dispatcher. However, there are cases when it is still possible to do. 

**Solution**
- Make `OutputStreamContent` and `WriterContent` dispatch to `Dispatchers.IO` if we running on a server's unsafe thread
- Prohibit taking/creating blocking adapters for channels on such threads

**Problems**
Since we are prohibiting creating blocking adapters, it is a behaviour change so a freezing code becomes crashing instead. So not sure if the target version is 1.4.x or 1.5.x.


ktor-http/jvm/src/io/ktor/http/content/OutputStreamContent.kt
```diff
@@ -20,6 +20,13 @@ public class OutputStreamContent(
 ) : OutgoingContent.WriteChannelContent() {
 
     override suspend fun writeTo(channel: ByteWriteChannel) {
-        channel.toOutputStream().use { it.body() }
+        withBlocking {
+            val stream = channel.toOutputStream()
+            // use bock should be inside because closing OutputStream is blocking as well
```
rsinukov, line 25: bock -> block

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationEngine.kt
```diff
@@ -204,16 +206,46 @@ public class EventLoopGroupProxy(public val channel: KClass<out ServerSocketChan
...
-            )
-            Epoll.isAvailable() -> EventLoopGroupProxy(
-                EpollServerSocketChannel::class,
-                EpollEventLoopGroup(parallelism)
-            )
-            else -> EventLoopGroupProxy(NioServerSocketChannel::class, NioEventLoopGroup(parallelism))
+        public fun create(parallelism: Int): EventLoopGroupProxy {
+            val defaultFactory = DefaultThreadFactory(EventLoopGroupProxy::class.java)
+
+            val factory: ThreadFactory = ThreadFactory { r ->
```
rsinukov, line 212: r -> runnable

ktor-http/jvm/src/io/ktor/http/content/BlockingBridge.kt
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+package io.ktor.http.content
+
+import kotlinx.coroutines.*
+import java.lang.reflect.*
+
+private val isParkingAllowedFunction: Method? by lazy {
+    try {
+        Class.forName("io.ktor.utils.io.jvm.javaio.PollersKt")
```
rsinukov, line 12: why using reflection here?

ktor-http/jvm/src/io/ktor/http/content/OutputStreamContent.kt
```diff
@@ -20,6 +20,13 @@ public class OutputStreamContent(
 ) : OutgoingContent.WriteChannelContent() {
 
     override suspend fun writeTo(channel: ByteWriteChannel) {
-        channel.toOutputStream().use { it.body() }
+        withBlocking {
+            val stream = channel.toOutputStream()
```
rsinukov, line 24: minor: stream val can be inlined

ktor-io/jvm/src/io/ktor/utils/io/internal/EventLoopExperimental.kt
```diff
@@ -1,8 +1,13 @@
+/*
+ * Copyright 2014-2020 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
 package io.ktor.utils.io.internal
 
 import java.lang.reflect.*
 
-
+@Deprecated("Invoke processNextEventInCurrentThread() directly instead.", level = DeprecationLevel.ERROR)
```
rsinukov, line 9: why deprecating internal class and removing it?

rsinukov (changes requested):

ktor-http/jvm/src/io/ktor/http/content/BlockingBridge.kt
cy6erGn0m: Because we can't invoke the function directly: it is made internal to avoid public API change and we are calling it via reflection. If the function is missing (like we have some kind jar hell), then no invocation happens and we fall back to the old behaviour. In 1.5.x we'll be able to make functions public and remove reflection magic.

cy6erGn0m: So it's a trick

ktor-io/jvm/src/io/ktor/utils/io/internal/EventLoopExperimental.kt
cy6erGn0m: We don't need it anymore. It was introduced to be compatible with different kotlinx.coroutines and fallback to some old behaviour. Nowadays, we don't need to support old coroutines so there is no need to do that magic anymore.

cy6erGn0m:
> Can we force such content types to always work on the IO dispatcher instead?

@e5l well, redispatching to `Dispatchers.IO` is expensive so we only do it when it is required, like when we are running on Netty's dispatcher (especially when the share work groups option is enabled).

ktor-io/jvm/src/io/ktor/utils/io/internal/EventLoopExperimental.kt
rsinukov: but can we just delete this code then?

rsinukov: sorry, my original comment had a typo, it should be
> why deprecating internal class and **not** removing it?

cy6erGn0m: Ah yes, you are right. Deleted.

rsinukov (approved):

e5l (approved):
LGTM

cy6erGn0m:
Merged

PR #2121, opened 2020-10-15
KTOR-895 Fix overwriting the default Jetty cipher exclusion list

cy6erGn0m:
**Subsystem**
ktor-server-jetty

**Motivation**
[KTOR-895](https://youtrack.jetbrains.com/issue/KTOR-895) 

**Solution**
Don't overwrite but amend the default Jetty cipher exlusion list.


rsinukov (approved):
LGTM but not sure if the test is required for this change

cy6erGn0m:
> LGTM but not sure if the test is required for this change

Unfortunately, I don't know how do I write a test except for writing a TLS client and validate all the suites that is definitely overkill

e5l (approved):

cy6erGn0m:
Merged

PR #2123, opened 2020-10-15
KTOR-750 Improve multiple cache-control directives behaviour (Stage 1)

cy6erGn0m:
**Subsystem**
ktor-server-core (Server), [CachingHeaders](https://ktor.io/docs/caching-headers.html) feature

**Motivation**
[KTOR-750](https://youtrack.jetbrains.com/issue/KTOR-750) Setting multiple cache control directives is impossible with current API 

Currently, specifying multiple cache-control directives causes server to produce multiple headers. In spite of it is not prohibited by the RFC, some clients may "don't like" this and stop caching at all. Also, repeating the same directive multiple time doesn't look smart. 

The other issue is that not all directives and options covered by the current API. For example, one can't specify `must-revalidate` without `max-age` that looks weird.

**Solution**
- Introduce merging caching directives

**Notes**
Adding extra-capabilities that not covered by the API is a change we can't afford in 1.4.x. This "Stage 1" is not intended to provide such functionality. We only merge directives into a single header (actually two).

### DIFF
--- a/ktor-server/ktor-server-core/jvm/src/io/ktor/util/CacheControlMerge.kt
+++ b/ktor-server/ktor-server-core/jvm/src/io/ktor/util/CacheControlMerge.kt
@@ -1,0 +1,70 @@
+/*
+ * Copyright 2014-2020 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.util
+
+import io.ktor.http.*
+
+/**
+ * Merge a list of cache control directives.
+ *
+ * Currently, visibility is pinned to the expiration directive,
+ * then to the first no-cache,
+ * then to the no-store directive.
+ * The RFC doesn't state, where a visibility modifier should be so we can place it at any place
+ * so there is nothing behind the rule beyond, therefore could be changed.
+ *
+ * Only one visibility specifier is kept.
+ *
+ * If there are different visibility modifiers specified, then the private wins.
+ * All max ages directives are reduced to a single with all minimal max age values.
+ *
+ * A no-cache directive is always placed first.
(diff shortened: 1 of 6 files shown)


ktor-server/ktor-server-core/jvm/src/io/ktor/util/CacheControlMerge.kt
```diff
@@ -0,0 +1,70 @@
...
+        any { it.visibility == CacheControl.Visibility.Public } -> CacheControl.Visibility.Public
+        else -> null
+    }
+
+    val noCacheDirective = firstOrNull { it is CacheControl.NoCache } as CacheControl.NoCache?
+    val noStoreDirective = firstOrNull { it is CacheControl.NoStore } as CacheControl.NoStore?
+
+    val maxAgeDirectives = filterIsInstance<CacheControl.MaxAge>()
+    val minMaxAge = maxAgeDirectives.minByOrNull { it.maxAgeSeconds }?.maxAgeSeconds
+    val minProxyMaxAge = maxAgeDirectives.mapNotNull { it.proxyMaxAgeSeconds }.minByOrNull { it }
```
rsinukov, line 45: `minByOrNull { it }` -> `.minOrNull()`

rsinukov: or even `maxAgeDirectives.minByOrNull { it.proxyMaxAgeSeconds }?.proxyMaxAgeSeconds`

rsinukov (approved):

e5l (approved):

cy6erGn0m:
Merged

PR #2124, opened 2020-10-16
KTOR-1120 Fix caching of jetty client

rsinukov:
**Subsystem**
Client, Jetty

**Motivation**
[KTOR-1120](https://youtrack.jetbrains.com/issue/KTOR-1120) Jetty client engine with high concurrency randomly freezes, throws NPEs, ISEs or EofExceptions

**Solution**
Timeout config was used as a key for HTTP clients cache, but it didn't have an implementation for `equals`/`hashcode`. That caused LRU cache overflow and closing of clients that were in use


e5l (approved):
lgtm

PR #2125, opened 2020-10-16
Fixed incorrect placement of setMaxConnPerRoute and setMaxConnTotal ApacheEngine

fluidsonic:
`setMaxConnPerRoute` and `setMaxConnPerRoute` configure the client, not the IO reactor.
Moving the two lines to the outer scope doesn't change what they do but make their intent clearer.

**Subsystem**
Apache Client Engine

**Motivation**
Code was confusing to read because it looked like that the two lines configure the IO reactor instead of the client.

**Solution**
Move calls to correct scope.

### DIFF
--- a/ktor-client/ktor-client-apache/jvm/src/io/ktor/client/engine/apache/ApacheEngine.kt
+++ b/ktor-client/ktor-client-apache/jvm/src/io/ktor/client/engine/apache/ApacheEngine.kt
@@ -58,11 +58,11 @@ internal class ApacheEngine(override val config: ApacheEngineConfig) : HttpClien
             disableAuthCaching()
             disableConnectionState()
             disableCookieManagement()
-            setDefaultIOReactorConfig(IOReactorConfig.custom().apply {
-                setMaxConnPerRoute(MAX_CONNECTIONS_COUNT)
-                setMaxConnTotal(MAX_CONNECTIONS_COUNT)
-                setIoThreadCount(IO_THREAD_COUNT_DEFAULT)
-            }.build())
+            setMaxConnPerRoute(MAX_CONNECTIONS_COUNT)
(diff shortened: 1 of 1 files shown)


cy6erGn0m (approved):

ktor-client/ktor-client-apache/jvm/src/io/ktor/client/engine/apache/ApacheEngine.kt
```diff
@@ -58,11 +58,11 @@ internal class ApacheEngine(override val config: ApacheEngineConfig) : HttpClien
             disableAuthCaching()
             disableConnectionState()
             disableCookieManagement()
-            setDefaultIOReactorConfig(IOReactorConfig.custom().apply {
-                setMaxConnPerRoute(MAX_CONNECTIONS_COUNT)
-                setMaxConnTotal(MAX_CONNECTIONS_COUNT)
-                setIoThreadCount(IO_THREAD_COUNT_DEFAULT)
-            }.build())
+            setMaxConnPerRoute(MAX_CONNECTIONS_COUNT)
+            setMaxConnPerRoute(MAX_CONNECTIONS_COUNT)
```
cy6erGn0m, line 62: You have it twice, should be `setMaxConnTotal`

cy6erGn0m (changes requested):
Sorry, didn't notice it first

ktor-client/ktor-client-apache/jvm/src/io/ktor/client/engine/apache/ApacheEngine.kt
fluidsonic: No idea how I've messed that one up. Thanks for the catch.

cy6erGn0m (approved):

e5l (approved):

PR #2126, opened 2020-10-16
KT-754 Fix applying compression option defaults

cy6erGn0m:
**Subsystem**
ktor-server-core ([Compression](https://ktor.io/docs/compression.html))

**Motivation**
[KTOR-754](https://youtrack.jetbrains.com/issue/KTOR-754) Sensible defaults for compression feature 

At the moment, the default configuration is only applied when no configuration for the feature provided. Consider the following example:

```kotlin
install(Compression) {
    gzip() // this disables all defaults
}
```

So doing this one need to configure everything manually to avoid "steel compression".

**Solution**
Apply default config in more cases so the example above will have the defaults with content types and size.
Note that any attempt to change the config still removes the default config so improved kdocs. Unfortunately, we can't fix that behaviour without breaking changes.

```kotlin
install(Compression) {
    gzip() // after fix will have defaults
}

install(Compression) {
    gzip()
    minimalSize(100) // this still reset defaults
}
```


e5l (approved):
LGTM.
It's a good first step. It also would be nice to be able to combine conditions(or write complex condition)

rsinukov (approved):

cy6erGn0m:
Merged

PR #2130, opened 2020-10-20
Reduce number of experimental/obsolete API usages

cy6erGn0m:
**Subsystem**
all

**Motivation**
At the moment, we are using too much experimental/obsolete API, especially because we have it whitelisted in the buildscript for all modules.

**Solution**
- Remove the annotation from the whitelist
- Reduce the number of usages
- make required opt-ins where necessary, as a small block as possible


rsinukov (approved):

cy6erGn0m:
Merged

PR #2131, opened 2020-10-21
KTOR-604 Add file and mailto protocols support

### LINKED ISSUE #1204: Url(…) doesn't parse non-HTTP-like URL strings correctly

e5l:
Fixes KTOR-604
Fixes #1204 

Rebased #1253

### DIFF
--- a/ktor-http/common/src/io/ktor/http/URLBuilder.kt
+++ b/ktor-http/common/src/io/ktor/http/URLBuilder.kt
@@ -61,6 +61,18 @@ public class URLBuilder(
 
     private fun <A : Appendable> appendTo(out: A): A {
         out.append(protocol.name)
+
+        when (protocol.name) {
+            "file" -> {
+                out.appendFile(host, encodedPath)
+                return out
+            }
+            "mailto" -> {
+                out.appendMailto(userAndPassword, encodedPath)
+                return out
+            }
+        }
+
         out.append("://")
         out.append(authority)
 
@@ -136,6 +148,19 @@ public data class Url(
 
     override fun toString(): String = buildString {
         append(protocol.name)
+
+        when (protocol.name) {
+            "file" -> {
+                appendFile(host, encodedPath)
+                return@buildString
+            }
+            "mailto" -> {
+                val userValue = user ?: error("User can't be empty.")
+                appendMailto(userValue, host)
+                return@buildString
+            }
+        }
+
         append("://")
         append(authority)
         append(fullPath)
@@ -149,19 +174,48 @@ public data class Url(
     public companion object
 }
 
+private fun Appendable.appendMailto(user: String, host: String) {
+    append(":")
+    append(user.encodeURLParameter())
+    append('@')
+    append(host)
+}
+
+private fun Appendable.appendFile(host: String, encodedPath: String) {
+    append("://")
+    append(host)
+    append(encodedPath)
+}
+
+internal val Url.userAndPassword: String
+    get() = buildString {
+        appendUserAndPassword(user, password)
+    }
+
+internal val URLBuilder.userAndPassword: String
+    get() = buildString {
+        appendUserAndPassword(user, password)
+    }
+
+private fun StringBuilder.appendUserAndPassword(user: String?, password: String?) {
+    user ?: return
+    append(user.encodeURLParameter())
+
+    if (password != null) {
+        append(':')
+        append(password.encodeURLParameter())
+    }
+
+    append("@")
+}
+
 /**
  * [Url] authority.
  */
 public val Url.authority: String
     get() = buildString {
-        if (user != null) {
-            append(user.encodeURLParameter())
-            if (password != null) {
-                append(':')
-                append(password.encodeURLParameter())
-            }
-            append('@')
-        }
+        append(userAndPassword)
+
         if (specifiedPort == DEFAULT_PORT) {
             append(host)
         } else {
@@ -174,14 +228,7 @@ public val Url.authority: String
  */
 public val URLBuilder.authority: String
     get() = buildString {
-        user?.let { user ->
-            append(user.encodeURLParameter())
-            password?.let { password ->
-                append(":")
-                append(password.encodeURLParameter())
-            }
-            append("@")
-        }
+        append(userAndPassword)
         append(host)
 
         if (port != DEFAULT_PORT && port != protocol.defaultPort) {
--- a/ktor-http/common/src/io/ktor/http/URLParser.kt
+++ b/ktor-http/common/src/io/ktor/http/URLParser.kt
@@ -41,6 +41,17 @@ internal fun URLBuilder.takeFromUnsafe(urlString: String): URLBuilder {
     val slashCount = count(urlString, startIndex, endIndex, '/')
     startIndex += slashCount
 
+    if (protocol.name == "file") {
+        parseFile(urlString, startIndex, endIndex, slashCount)
+        return this
+    }
+
+    if (protocol.name == "mailto") {
+        require(slashCount == 0)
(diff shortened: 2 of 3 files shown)


ktor-http/common/src/io/ktor/http/URLParser.kt
```diff
@@ -41,6 +41,30 @@ internal fun URLBuilder.takeFromUnsafe(urlString: String): URLBuilder {
     val slashCount = count(urlString, startIndex, endIndex, '/')
     startIndex += slashCount
 
+    if (protocol.name == "file") {
+        require(slashCount == 3)
```
cy6erGn0m, line 45: A message is required: we don't support host in file URIs so we need to clearly state it

ktor-http/common/src/io/ktor/http/URLParser.kt
```diff
@@ -41,6 +41,30 @@ internal fun URLBuilder.takeFromUnsafe(urlString: String): URLBuilder {
...
+        require(slashCount == 0)
+
+        val delimiter = urlString.indexOf("@", startIndex);
+        val passwordIndex = urlString.indexOfColonInHostPort(startIndex, delimiter)
+        if (delimiter < endIndex) {
+            if (passwordIndex != -1) {
+                user = urlString.substring(startIndex, passwordIndex)
+                password = urlString.substring(passwordIndex + 1, delimiter)
+            } else {
+                user = urlString.substring(startIndex, delimiter)
```
cy6erGn0m, line 60: So if we have no `@` character, then we get an empty username and a host without a first letter, right? I am not sure if it is allowed to have no `user@` in email, do you?

ktor-http/common/src/io/ktor/http/URLBuilder.kt
```diff
@@ -149,19 +179,42 @@ public data class Url(
...
+            append(':')
+            append(password.encodeURLParameter())
+        }
+
+        append("@")
+    }
+
+internal val URLBuilder.userAndPassword: String
+    get() = buildString {
+        val userValue = user ?: return@buildString
```
cy6erGn0m, line 199: duplication

ktor-http/common/src/io/ktor/http/URLParser.kt
```diff
@@ -41,6 +41,30 @@ internal fun URLBuilder.takeFromUnsafe(urlString: String): URLBuilder {
     val slashCount = count(urlString, startIndex, endIndex, '/')
     startIndex += slashCount
 
+    if (protocol.name == "file") {
+        require(slashCount == 3)
+        host = urlString.substring(startIndex, endIndex)
```
cy6erGn0m, line 46: Why do we say that the path is a host while the host is actually missing (empty)?

ktor-http/common/test/io/ktor/tests/http/UrlTest.kt
```diff
@@ -186,4 +189,23 @@ class UrlTest {
             assertEquals(null, password)
         }
     }
+
+    @Test
+    fun testForFileProtocol() {
+        val expectedUrl = "file:///var/www"
+        val result = Url(expectedUrl);
+        assertEquals(result.toString(), expectedUrl);
+        assertEquals(result.protocol.name, "file");
```
cy6erGn0m, line 198: assert "var/www" == result.path?

cy6erGn0m (changes requested):

ktor-http/common/src/io/ktor/http/URLParser.kt
e5l: Thanks for the notice, I will add support and tests

ktor-http/common/src/io/ktor/http/URLParser.kt
```diff
@@ -89,34 +99,72 @@ internal fun URLBuilder.takeFromUnsafe(urlString: String): URLBuilder {
...
-        rawParameters.forEach { key, values ->
-            parameters.appendAll(key, values)
+            host = urlString.substring(startIndex, nextSlash)
+            encodedPath = urlString.substring(nextSlash, endIndex)
         }
+        3 -> {
+            host = ""
+            encodedPath = "/" + urlString.substring(startIndex, endIndex)
+        }
+    }
```
cy6erGn0m, line 135: else -> error(....)

ktor-http/common/src/io/ktor/http/URLParser.kt
```diff
@@ -89,34 +99,72 @@ internal fun URLBuilder.takeFromUnsafe(urlString: String): URLBuilder {
...
 
-        startIndex = fragmentStart
+private fun URLBuilder.parseMailto(urlString: String, startIndex: Int, endIndex: Int) {
+    val delimiter = urlString.indexOf("@", startIndex);
+    if (delimiter == -1) {
+        throw IllegalArgumentException("Invalid mailto url: $urlString, it should contains '@'.")
     }
 
-    // Fragment
+    user = urlString.substring(startIndex, delimiter)
```
cy6erGn0m, line 144: Shouldn't we decode/de-escape user and password here?

cy6erGn0m (changes requested):

ktor-http/common/src/io/ktor/http/URLParser.kt
e5l: As I get it, we store everything in escaped format(the same as for regular URLs)

ktor-http/common/src/io/ktor/http/URLParser.kt
e5l: Good catch, thanks

ktor-http/common/src/io/ktor/http/URLParser.kt
cy6erGn0m: Then we shouldn't escape them when formatting otherwise it breaks parse-then-encode.

e5l: fixed as discussed, please take a look

ktor-http/common/src/io/ktor/http/URLParser.kt
```diff
@@ -89,34 +99,73 @@ internal fun URLBuilder.takeFromUnsafe(urlString: String): URLBuilder {
...
+        }
+        else -> throw IllegalArgumentException("Invalid file url: $urlString")
+    }
+}
 
-        startIndex = fragmentStart
+private fun URLBuilder.parseMailto(urlString: String, startIndex: Int, endIndex: Int) {
+    val delimiter = urlString.indexOf("@", startIndex);
+    if (delimiter == -1) {
+        throw IllegalArgumentException("Invalid mailto url: $urlString, it should contains '@'.")
```
cy6erGn0m, line 142: it should contain_

cy6erGn0m (approved):

ktor-http/common/src/io/ktor/http/URLParser.kt
e5l: Fixed, thanks

PR #2137, opened 2020-10-22
KTOR-456 Fix logging of request headers

### LINKED ISSUE #1690: Content-type header is missing in Request when using LogLevel.BODY
version: `1.3.1`
lib: `ktor-client-cio`
additional modules: `ktor-client-logging-jvm`

**Describe the bug**
Content-type header is missing when `LogLevel.BODY` is set. _(also I am using multi-part body, not sure if relevant)_. Header exists if removing logging or setting `LogLevel.HEADER`.

**To Reproduce**
1. Set logging 
```kotlin
HttpClient {
  install(Logging) {
  logger = Logger.ANDROID
  level = LogLevel.BODY
  }
}
```
2. Do multi-form request
```kotlin
val response = httpClient.post<String> {
  url("http://someurl")
  body = MultiPartFormDataContent(formData {
    append(
          "file",
          "file.txt",
          ContentType.parse("text/plain"),
          contentBytes.size.toLong()
    ) {
      writeFully(contentBytes)
    }
})
```
3. Observe missing `Content-Type: multipart/form-data; <...>` header

**Expected behavior**
`Content-Type: multipart/form-data; <...>` in request exists

e5l:
Fixes [KTOR-456](https://youtrack.jetbrains.com/issue/KTOR-456)
Fixes #1690


ktor-client/ktor-client-features/ktor-client-logging/common/src/io/ktor/client/features/logging/Logging.kt
```diff
@@ -80,6 +80,8 @@ public class Logging(
             logHeaders(request.headers.entries())
 
             logger.log("CONTENT HEADERS")
+            content.contentLength?.let { logger.logHeader(HttpHeaders.ContentLength, it.toString()) }
+            content.contentType?.let { logger.logHeader(HttpHeaders.ContentLength, it.toString()) }
```
rsinukov, line 84: should not it be `HttpHeaders.ContentType`?

rsinukov (changes requested):

ktor-client/ktor-client-features/ktor-client-logging/common/src/io/ktor/client/features/logging/Logging.kt
e5l: Thanks for the notice, fixed

rsinukov (approved):

PR #2139, opened 2020-10-22
Correct encoding of path parts in the locations feature.

spand:
**Subsystem**
Locations feature

**Motivation**
Currently locations not escape path values that contains chars such as / or # leading to urls like:
/a/da#/da
for a class
```
@Location("/{foo}/da")
class Foo(name: String)
Foo("a/da#")
```
This is obviously not possible to route back to this location.

**Solution**
Introduce a new encode method only used for url parts between the slashes and exhaustive tests.

The reason I do not exhaustively test characters after codepoint 0xD800 is the encoding utils will loop forever on 0xD800 as can be seen by expanding the range.

### DIFF
--- a/ktor-features/ktor-locations/jvm/src/io/ktor/locations/BackwardCompatibleImpl.kt
+++ b/ktor-features/ktor-locations/jvm/src/io/ktor/locations/BackwardCompatibleImpl.kt
@@ -95,7 +95,7 @@ internal class BackwardCompatibleImpl(
 
         val relativePath = substituteParts
             .filterNot { it.isEmpty() }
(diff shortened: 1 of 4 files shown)


cy6erGn0m:
Don't we already have `encodeURLPart` somewhere?

spand:
Maybe you think of this
```
/**
 * Encode url part as specified in
 * https://tools.ietf.org/html/rfc3986#section-2
 */
public fun String.encodeURLQueryComponent(
```

e5l:
@rsinukov, could you check if the PR is valid for the new Resources feature?

PR #2141, opened 2020-10-22
KTOR-879 Fix multipart form data client header

e5l:
Fixes [KTOR-879](https://youtrack.jetbrains.com/issue/KTOR-879)

### DIFF
--- a/ktor-client/ktor-client-core/common/test/MultiPartFormDataContentTest.kt
+++ b/ktor-client/ktor-client-core/common/test/MultiPartFormDataContentTest.kt
@@ -1,0 +1,30 @@
+import io.ktor.client.request.forms.*
+import io.ktor.test.dispatcher.*
+import io.ktor.utils.io.*
+import io.ktor.utils.io.core.*
+import kotlin.test.*
+
+/*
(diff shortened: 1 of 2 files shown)


ktor-client/ktor-client-core/common/test/MultiPartFormDataContentTest.kt
```diff
@@ -0,0 +1,30 @@
...
+            append("Hello", "World")
+        })
+
+        val channel = ByteChannel()
+        formData.writeTo(channel)
+        channel.close()
+
+        val actual = channel.readRemaining().readBytes()
+
+        assertNotEquals('\r', actual[0])
```
Stexxe, line 25: Could be more readable if you compare a slice for the first 4 bytes instead of comparing individual bytes.

Stexxe (approved):

ktor-client/ktor-client-core/common/test/MultiPartFormDataContentTest.kt
e5l: Nope, cause it will be a comparison of byte arrays instead of chars

PR #2142, opened 2020-10-22
KTOR-1163 Allow responding with polymorphic collections

rsinukov:
**Subsystem**
Server, `kotlinx.serialization`

**Motivation**
(Serializing collections of different element types)[https://youtrack.jetbrains.com/issue/KTOR-1163]

**Solution**
Save type of response to use it in the serializer


ktor-server/ktor-server-core/jvm/src/io/ktor/response/ResponseTypeAttributeKey.kt
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright 2014-2020 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.response
+
+import io.ktor.util.*
+import kotlin.reflect.*
+
+public val ResponseTypeAttributeKey: AttributeKey<KType> = AttributeKey("ResponseTypeAttributeKey")
```
cy6erGn0m, line 10: Do we need it public?

ktor-server/ktor-server-core/jvm/src/io/ktor/response/ApplicationResponseFunctions.kt
```diff
@@ -12,22 +12,29 @@ import io.ktor.http.content.*
...
+    response.pipeline.execute(this, message as Any)
 }
 
 /**
  * Sets [status] and sends a [message] as a response
  */
-@Suppress("NOTHING_TO_INLINE")
-public suspend inline fun ApplicationCall.respond(status: HttpStatusCode, message: Any) {
+@OptIn(ExperimentalStdlibApi::class)
+public suspend inline fun <reified T> ApplicationCall.respond(status: HttpStatusCode, message: T) {
```
cy6erGn0m, line 32: Intercepting a call pipeline is a relatively expensive operation that introduce a lot of allocations during pipeine merge. Also, it looks like we are adding this type every time even if we don't need it so we pay even with no serialization enabled

cy6erGn0m, line 32: Since the function is inline, we will inline the block every time, producing a lot of anonymous classes at call site. In such cases we usually make a pair of functions, like this:

```
inline fun <reified T> doSomething() 
fun <T> doSometing(KType)
```

or even

```
inline fun <reified T> doSomething() 
@PublishedApi
internal fun <T> doSometing(KType)
```

ktor-server/ktor-server-core/api/ktor-server-core.api
```diff
@@ -985,8 +985,6 @@ public abstract interface class io/ktor/response/ApplicationResponse {
 
 public final class io/ktor/response/ApplicationResponseFunctionsKt {
 	public static final fun defaultTextContentType (Lio/ktor/application/ApplicationCall;Lio/ktor/http/ContentType;)Lio/ktor/http/ContentType;
-	public static final fun respond (Lio/ktor/application/ApplicationCall;Lio/ktor/http/HttpStatusCode;Ljava/lang/Object;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
```
cy6erGn0m, line 988: I would keep old functions a well, at least for binary compatibility

ktor-server/ktor-server-core/jvm/src/io/ktor/response/ApplicationResponseFunctions.kt
```diff
@@ -12,22 +12,29 @@ import io.ktor.http.content.*
...
 /**
  * Sends a [message] as a response
  */
-@Suppress("NOTHING_TO_INLINE")
-public suspend inline fun ApplicationCall.respond(message: Any) {
-    response.pipeline.execute(this, message)
+@OptIn(ExperimentalStdlibApi::class)
+public suspend inline fun <reified T> ApplicationCall.respond(message: T) {
+    response.pipeline.intercept(ApplicationSendPipeline.Before) {
+        call.attributes.put(ResponseTypeAttributeKey, typeOf<T>())
```
cy6erGn0m, line 23: The problem with such an approach is that if a response object gets replaced, the paired type keeps remain that may lead to strange consequences. However, this is a good starting point. Why can't we put this type immediately instead of interception?

ktor-server/ktor-server-core/api/ktor-server-core.api
rsinukov: sure, it's still a draft PR that needs more fine-tuning. I published it to just run tests on CI

ktor-server/ktor-server-core/jvm/src/io/ktor/response/ApplicationResponseFunctions.kt
rsinukov: sure, I don't know why I made the interceptor here. but again, it's still a draft PR that will be updated

ktor-server/ktor-server-core/jvm/src/io/ktor/response/ResponseTypeAttributeKey.kt
rsinukov: Yes, because we write with this key in `ktor-server-core` but read in `ktor-serialization`

ktor-server/ktor-server-core/jvm/src/io/ktor/response/ApplicationResponseFunctions.kt
rsinukov: I removed interceptor and lambda, should not be an issue now

rsinukov: removed interception. unfortunately, we can't remove the evaluation of type here, because it's the only place where we still can access it

ktor-server/ktor-server-core/api/ktor-server-core.api
rsinukov: updated

ktor-server/ktor-server-core/jvm/src/io/ktor/response/ApplicationResponseFunctions.kt
```diff
@@ -12,22 +12,53 @@ import io.ktor.http.content.*
...
 import java.io.*
+import kotlin.reflect.*
 
 /**
  * Sends a [message] as a response
  */
-@Suppress("NOTHING_TO_INLINE")
+@OptIn(ExperimentalStdlibApi::class)
+@JvmName("respondWithType")
+public suspend inline fun <reified T> ApplicationCall.respond(message: T) {
```
cy6erGn0m, line 22: Unlike the old respond function, this one provides a way to respond with `null`, including with `T = Nothing?`. Not sure if we need to allow it. If yes, we need to add a test for this

ktor-features/ktor-serialization/jvm/test/io/ktor/tests/serialization/SerializationTest.kt
```diff
@@ -416,6 +417,50 @@ class SerializationTest {
...
+        handleRequest(HttpMethod.Get, "/") {
+            addHeader("Accept", "application/json")
+        }.let { call ->
+            assertEquals(HttpStatusCode.OK, call.response.status())
+            assertEquals(
+                """[{"x":777},{"x":888}]""",
+                call.response.content
+            )
+        }
+    }
```
cy6erGn0m, line 463: I would also test `null` (both typed `null` and `Nothing?`), and a case with `respond(when { ...})`

ktor-server/ktor-server-core/jvm/src/io/ktor/response/ResponseTypeAttributeKey.kt
cy6erGn0m: We could make it `@InternalAPI`

ktor-features/ktor-serialization/jvm/src/io/ktor/serialization/SerializerLookup.kt
```diff
@@ -36,33 +38,49 @@ private fun arraySerializer(type: KType): KSerializer<*> {
...
+internal fun serializerForSending(
+    context: PipelineContext<Any, ApplicationCall>,
+    value: Any,
+    module: SerializersModule
+): KSerializer<*> {
+    val responseType = context.call.attributes.getOrNull(ResponseTypeAttributeKey)
+    if (responseType != null) try {
+        return module.serializer(responseType)
+    } catch (_: SerializationException) {
+        // if serializer for type was not found, fallback to manual search
```
cy6erGn0m, line 50: When I don't like here is that all the default content types will always go through the exception. This is good and bad at the same time.
- Pros: one could configure serialization for types like `HttpStatusCode`
- Cons: looking for a serializer, creating an exception instance are slow

ktor-server/ktor-server-core/jvm/src/io/ktor/response/ApplicationResponseFunctions.kt
```diff
@@ -12,22 +12,53 @@ import io.ktor.http.content.*
...
+    try {
+        // We need to wrap getting type in try catch because of https://youtrack.jetbrains.com/issue/KT-42913
+        attributes.put(ResponseTypeAttributeKey, typeOf<T>())
+    } catch (_: Throwable) {
+    }
+    response.pipeline.execute(this, message as Any)
```
cy6erGn0m, line 28: On the other hand, a `null` value will crash here with KNPE. So we need to fix and allow this or we need to add an upper bound for type `T`.

cy6erGn0m (changes requested):
Still questionable, how the `typeOf` evaluation for every request affects the performance. A typical example is hello world example under high load, like one we have in framework benchmarks

ktor-server/ktor-server-core/jvm/src/io/ktor/response/ApplicationResponseFunctions.kt
rsinukov: Good point. Limited `T` to `T : Any`

ktor-features/ktor-serialization/jvm/test/io/ktor/tests/serialization/SerializationTest.kt
rsinukov: added test for response `when`

ktor-server/ktor-server-core/jvm/src/io/ktor/response/ApplicationResponseFunctions.kt
rsinukov: limited type to `T : Any`

ktor-features/ktor-serialization/jvm/src/io/ktor/serialization/SerializerLookup.kt
rsinukov: Unfortunately, there is no better way to check if a serializer is available. I created the issue in `kotlinx.serializer` project https://github.com/Kotlin/kotlinx.serialization/issues/1163

e5l (changes requested):
The code is LGTM for sure. One little thing I want to discuss: we locking the type in attributes and it looks like a hack. If I'm going to write an external serialization feature, I should discover that somehow. Could you tell me if Is it possible to add type info in the request object in a clear observable way? Maybe we need design here, @cy6erGn0m  what do you think?

rsinukov:
@e5l 
I see three possible options to increase the visibility of type in response:
1. We can create a wrapper class that will be a subject for `ApplicationSendPipeline`, similar to `ApplicationReceiveRequest` for `ApplicationReceivePipeline`. This will require changes in existing features and for sure will be a huge breaking change for our users, but it can be more future proof.
2. Keep type as it is in attributes, but provide extension property for easy access
3. Create a real property in `ApplicationResponse` that will store type (my least favorite one)

I would choose solution 2 here, what do you think?

e5l:
Yep, let's stay with the second option for now.

e5l (approved):

ktor-features/ktor-serialization/jvm/src/io/ktor/serialization/SerializerLookup.kt
rsinukov: I updated my code to create `KType` only when the user responds with a custom type

cy6erGn0m (approved):

PR #2147, opened 2020-10-23
KTOR-1169 Set lowest priority to jcenter

e5l:
Fixes [KTOR-1169](https://youtrack.jetbrains.com/issue/KTOR-1169)

It should help according to this guide: https://docs.gradle.org/current/userguide/declaring_repositories.html

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -32,7 +32,7 @@ buildscript {
 
     repositories {
         mavenLocal()
-        jcenter()
+        mavenCentral()
         google()
         gradlePluginPortal()
         maven { url "https://dl.bintray.com/jetbrains/kotlin-native-dependencies" }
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -32,7 +32,6 @@ buildscript {
 
     repositories {
         mavenLocal()
-        jcenter()
```
cy6erGn0m, line 35: ❓  Shouldn't we have mavenCentral here just like we have for plugins?

e5l: yep, let's add it

cy6erGn0m: Still no mavenCentral here :)

cy6erGn0m (changes requested):

cy6erGn0m (approved):

PR #2150, opened 2020-10-23
Add HTTP2 trailer support for Netty engine.

xiaodongw:
**Subsystem**
Server netty

**Motivation**
I'm working on a prototype to support GRPC on top of Ktor, one thing missing is the HTTP2 trailer support. With this change, the prototype is working for a GRPC unary request/respone.

**Solution**
Add OutgoingContent.trailers(), if it's HTTP2 it will be sent out as HTTP2 trailer frame. On HTTP1 the trailers will be ignored.
Right now it's only for Netty. I'll look into other engines later.


ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/cio/NettyResponsePipeline.kt
```diff
@@ -7,6 +7,8 @@ package io.ktor.server.netty.cio
 import io.ktor.util.cio.*
 import io.ktor.http.*
 import io.ktor.server.netty.*
+import io.ktor.server.netty.http2.*
+import io.ktor.server.netty.http2.NettyHttp2ApplicationResponse
```
cy6erGn0m, line 11: Why do we need ` import io.ktor.server.netty.http2.NettyHttp2ApplicationResponse`?

ktor-http/common/src/io/ktor/http/content/OutgoingContent.kt
```diff
@@ -56,6 +56,11 @@ public sealed class OutgoingContent {
         }
     }
 
+    /**
+     * Trailers to set when sending this content, will be ignored if request is not in HTTP2 mode
```
cy6erGn0m, line 60: Btw, trailing headers are possible in HTTP/1.1 but usually not processed and almost nobody cares. The question is: should we try to support it in HTTP/1 or we state that it is HTTP/2 only.

The other issue is that it's not easy to use since we have many content implementations and helper functions most of them don't provide a way to specify even regular headers. So how one could provide these headers?

cy6erGn0m (changes requested):
Thanks for your investigation. Generally, it would be fine except for question about user API. Btw, why does gRPC need it? How does it work over HTTP/1?

ktor-http/common/src/io/ktor/http/content/OutgoingContent.kt
xiaodongw: For HTTP/1.1, I think we should leave it right now since there is no use case for it.

GRPC only works on HTTP2. It utilizes the flow control for back pressure (I guess here it's already handled by Ktor so no need to worry about it), it also uses trailer to send its own status code, so the status can be sent after the stream.

This is a rarely used feature (seems except for GRPC no one uses it), it should be available on the most flexible API (OutgoingContent I assume), but not others for simplicity.

cy6erGn0m:
Ok, let try it. To proceed, we need to do the following:
- rebase on top of `1.5.0` branch and make change PR's base branch
- run `./gradlew apiDump` and commit changes in `*.api` files

e5l (approved):

e5l (approved):

e5l:
Hey, @xiaodongw. Thanks for the PR, lgtm

PR #2152, opened 2020-10-26
Fix using async dispatcher for blocking operations

cy6erGn0m:
**Subsystem**
ktor-client-apache, ktor-client-core, ktor-client-okhttp, ktor-client-cio, ... 

**Motivation**
[KTOR-1140](https://youtrack.jetbrains.com/issue/KTOR-1140)  Blocking calls (ReentrantLock.lock) on Apache engine dispatcher threads 

**Solution**
Stop using `ExperimentalCoroutinesDispatcher` and use it's blocking view instead

**Problem**
After doing this we will run CIO on this dispatcher. However, this could be not so bad since we do DNS resolution in CIO on this dispatcher as well.


e5l (approved):
lgtm

cy6erGn0m:
Merged

PR #2161, opened 2020-10-27
KTOR-1260 Introduce PublicAPICandidate annotation

cy6erGn0m:
**Subsystem**
Core

**Motivation**
[KTOR-1260](https://youtrack.jetbrains.com/issue/KTOR-1260)  Staging new public API 

**Solution**
Introduce `@PublicAPICandidate` annotation as described in the ticket.

### DIFF
--- a/ktor-utils/api/ktor-utils.api
+++ b/ktor-utils/api/ktor-utils.api
@@ -252,6 +252,10 @@ public final class io/ktor/util/PlatformUtils {
 	public final fun getIS_NODE ()Z
 }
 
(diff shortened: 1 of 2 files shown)


e5l:
Can we introduce it `internal` in `master` branch and mark with PublicAPI candidate? :D

cy6erGn0m:
> Can we introduce it `internal` in `master` branch and mark with PublicAPI candidate? :D

I had the same idea but it's not that beneficial since it's situated in ktor-utils so we can only do that with API in utils 🤣

e5l (approved):
lgtm

cy6erGn0m:
Merged

PR #2164, opened 2020-10-28
KTOR-1166 Fix UDP socket send exception

e5l:


ktor-network/jvm/src/io/ktor/network/sockets/DatagramSendChannel.kt
```diff
@@ -0,0 +1,93 @@
...
+        lock.withLock {
+            val buffer = element.prepareMessage()
+
+            val rc = channel.send(buffer, element.address)
+            if (rc != 0) {
+                socket.interestOp(SelectInterest.WRITE, false)
+                return
+            }
+
+            sendSuspend(buffer, element.address)
```
cy6erGn0m, line 62: I hope the state machine elimination didn't fail here because of the lambda block `withLock`

ktor-network/jvm/src/io/ktor/network/sockets/DatagramSendChannel.kt
```diff
@@ -0,0 +1,93 @@
...
+        get() = TODO("[DatagramSendChannel] doesn't support [onSend] select clause")
+
+    @ExperimentalCoroutinesApi
+    override fun invokeOnClose(handler: (cause: Throwable?) -> Unit) {
+        TODO("[DatagramSendChannel] doesn't support [invokeOnClose] operation.")
+    }
+}
+
+private fun Datagram.prepareMessage(): ByteBuffer {
+    val buffer = ByteBuffer.allocateDirect(packet.remaining.toInt())
```
cy6erGn0m, line 88: We could try direct buffer pool (`DefaultDatagramByteBufferPool`?) here but as it complicates it so I am in doubt since UDP is not that popular

ktor-network/jvm/src/io/ktor/network/sockets/DatagramSendChannel.kt
```diff
@@ -0,0 +1,93 @@
...
+    val channel: DatagramChannel,
+    val socket: DatagramSocketImpl
+) : SendChannel<Datagram> {
+    @ExperimentalCoroutinesApi
+    override val isClosedForSend: Boolean
+        get() = socket.isClosed
+
+    @ExperimentalCoroutinesApi
+    override val isFull: Boolean
+        get() = lock.isLocked
```
cy6erGn0m, line 27: > This function returns `false`  if the channel [is closed for `send`][isClosedForSend].

ktor-network/jvm/test/io/ktor/network/sockets/tests/UDPSocketTest.kt
```diff
@@ -0,0 +1,52 @@
...
+
+    @AfterTest
+    fun tearDown() {
+        testJob.cancel()
+        selector.close()
+    }
+
+    @Test
+    fun testBroadcastFails(): Unit = runBlocking {
+        assertFails {
```
cy6erGn0m, line 37: I would make the assertion more precise since we also may fail due to UDP bind failure (not enough fds and so on)

cy6erGn0m (changes requested):

ktor-network/jvm/src/io/ktor/network/sockets/DatagramSendChannel.kt
e5l: Thanks, fixed

ktor-network/jvm/test/io/ktor/network/sockets/tests/UDPSocketTest.kt
e5l: Thanks for the notice, will fix

ktor-network/jvm/src/io/ktor/network/sockets/DatagramSendChannel.kt
e5l: I'm not sure that it's critical here, but `withLock` is inline, so I hope for the best.

ktor-network/jvm/src/io/ktor/network/sockets/DatagramSendChannel.kt
e5l: done

e5l:
Thanks for the good review, fixed! Please take a look!

ktor-network/jvm/src/io/ktor/network/sockets/DatagramSendChannel.kt
```diff
@@ -0,0 +1,97 @@
...
+    private val lock = Mutex()
+
+    override fun close(cause: Throwable?): Boolean {
+        if (socket.isClosed) {
+            return false
+        }
+
+        socket.close()
+        return true
+    }
```
Thomas-Vos, line 40: This does not work as expected. Try the following test (just add it to `UDPSocketTest`:
```kt
@Test
fun testClose(): Unit = runBlocking {
    val socket = aSocket(selector)
        .udp()
        .bind()
    socket.close()
}
```
It will fail with a `StackOverflowError`:
```
java.lang.StackOverflowError
	at java.lang.Class.getEnclosingMethodInfo(Class.java:1072)
	at java.lang.Class.getEnclosingClass(Class.java:1272)
	at java.lang.Class.getSimpleBinaryName(Class.java:1443)
	at java.lang.Class.getSimpleName(Class.java:1309)
	at kotlinx.coroutines.DebugStringsKt.getClassSimpleName(DebugStrings.kt:20)
	at kotlinx.coroutines.AbstractCoroutine.cancellationExceptionMessage(AbstractCoroutine.kt:97)
	at kotlinx.coroutines.JobSupport.access$cancellationExceptionMessage(JobSupport.kt:29)
	at kotlinx.coroutines.channels.ChannelCoroutine.cancel(ChannelCoroutine.kt:49)
	at kotlinx.coroutines.channels.ReceiveChannel$DefaultImpls.cancel$default(Channel.kt:311)
	at io.ktor.network.sockets.DatagramSocketImpl.close(DatagramSocketImpl.kt:48)
	at io.ktor.network.sockets.DatagramSendChannel.close(DatagramSendChannel.kt:38)
	at kotlinx.coroutines.channels.SendChannel$DefaultImpls.close$default(Channel.kt:102)
	at io.ktor.network.sockets.DatagramSocketImpl.close(DatagramSocketImpl.kt:49)
	at io.ktor.network.sockets.DatagramSendChannel.close(DatagramSendChannel.kt:38)
	at kotlinx.coroutines.channels.SendChannel$DefaultImpls.close$default(Channel.kt:102)
	at io.ktor.network.sockets.DatagramSocketImpl.close(DatagramSocketImpl.kt:49)
	at io.ktor.network.sockets.DatagramSendChannel.close(DatagramSendChannel.kt:38)
	at kotlinx.coroutines.channels.SendChannel$DefaultImpls.close$default(Channel.kt:102)
	at io.ktor.network.sockets.DatagramSocketImpl.close(DatagramSocketImpl.kt:49)
	at io.ktor.network.sockets.DatagramSendChannel.close(DatagramSendChannel.kt:38)
	at kotlinx.coroutines.channels.SendChannel$DefaultImpls.close$default(Channel.kt:102)
	at io.ktor.network.sockets.DatagramSocketImpl.close(DatagramSocketImpl.kt:49)
	at io.ktor.network.sockets.DatagramSendChannel.close(DatagramSendChannel.kt:38)
... removed duplicate lines
```

Thomas-Vos (changes requested):
The `close` function throws a `StackOverflowError`, see review for details.

ktor-network/jvm/test/io/ktor/network/sockets/tests/UDPSocketTest.kt
Thomas-Vos: I saw that the latest commits changed this check to the following:
```
assertFailsWith<SocketException>("Permission denied") { ... }
```
Just so you know, the "Permission denied" string here is only for logging. So it does not actually check the exception message but only the exception type. Is that expected as @cy6erGn0m wanted the assertion to be more precise?

e5l: The message can be different on each platform

ktor-network/jvm/src/io/ktor/network/sockets/DatagramSendChannel.kt
e5l: Good catch! Thanks(and sorry for the delay). I fixed the problem in recent commits, could you take a look?

ktor-network/jvm/src/io/ktor/network/sockets/DatagramSendChannel.kt
```diff
@@ -0,0 +1,97 @@
...
+
+
+    override fun offer(element: Datagram): Boolean {
+        if (!lock.tryLock()) return false
+
+        try {
+            DefaultDatagramByteBufferPool.useInstance { buffer ->
+                element.writeMessageTo(buffer)
```
cy6erGn0m, line 48: Is Datagram reusable? How many times we can call writeMessageTo?

ktor-network/jvm/src/io/ktor/network/sockets/DatagramSendChannel.kt
```diff
@@ -0,0 +1,97 @@
...
+            }
+        }
+    }
+
+    override val onSend: SelectClause2<Datagram, SendChannel<Datagram>>
+        get() = TODO("[DatagramSendChannel] doesn't support [onSend] select clause")
+
+    @ExperimentalCoroutinesApi
+    override fun invokeOnClose(handler: (cause: Throwable?) -> Unit) {
+        TODO("[DatagramSendChannel] doesn't support [invokeOnClose] operation.")
```
cy6erGn0m, line 90: Unlike onSend, invokeOnClose seems to be useful. Some kxcoroutine parts depend on it. Can't we support it? Or, at least, a single handler registration?

cy6erGn0m (changes requested):

ktor-network/jvm/src/io/ktor/network/sockets/DatagramSendChannel.kt
e5l: fixed as dicussed

ktor-network/jvm/src/io/ktor/network/sockets/DatagramSendChannel.kt
e5l: Thanks for the catch! Implemented

ktor-network/jvm/src/io/ktor/network/sockets/DatagramSendChannel.kt
Thomas-Vos: Works great now!

Thomas-Vos (changes requested):
I have some suggestions for the `invokeOnClose` function. I added some more tests and fixed those tests. Please see my changes here: https://github.com/Thomas-Vos/ktor/commit/184a61848d0244cb6773685c93e3e05133c3920f
Let me know if you have any questions.

ktor-network/jvm/src/io/ktor/network/sockets/DatagramSendChannel.kt
Thomas-Vos: Looks great so far, here is what I would change: https://github.com/Thomas-Vos/ktor/commit/184a61848d0244cb6773685c93e3e05133c3920f

e5l: Hey @Thomas-Vos, thanks for the suggestion. The tests are great. One minor thing I found: there is a race with `cause` value in the suggestion(it's possible to find `null` cause(in `invokeOnClose`) before the cause was actually set(in `close`). I will take your suggestions with minor fixes. Thanks again.

e5l:
Added test and fixed concurrent close. @cy6erGn0m, @Thomas-Vos please take a look

cy6erGn0m (approved):

ktor-network/jvm/src/io/ktor/network/sockets/DatagramSocketImpl.kt
```diff
@@ -10,30 +10,26 @@ import io.ktor.util.network.*
...
+        get() = socket.localSocketAddress
             ?: throw IllegalStateException("Channel is not yet bound")
 
     override val remoteAddress: NetworkAddress
-        get() = socket.remoteSocketAddress as? NetworkAddress
+        get() = socket.remoteSocketAddress
             ?: throw IllegalStateException("Channel is not yet connected")
 
-    @OptIn(ObsoleteCoroutinesApi::class, ExperimentalCoroutinesApi::class)
-    private val sender = actor<Datagram>(Dispatchers.IO) {
```
Thomas-Vos, line 32: It appears that because of this pull request, datagram messages are no longer sent on the IO dispatcher but on the dispatcher that it is called from by the user. So for example on Android, if you call `send` on the datagram socket, it will crash with `android.os.NetworkOnMainThreadException`. The solution for the user is to call send from `Dispatchers.IO`, but that is a change in behaviour. Previously, as you can see in the code above, the actor would already use `Dispatchers.IO`. That is now no longer the case. Are you aware of this change in behaviour?

This could be fixed by using `withContext(Dispatchers.IO)` in `DatagramSendChannel.send`. However, the `offer` function would still fail because it cannot suspend. Any idea how to fix this? Should this be fixed? Maybe the `offer` function should not be available? What do you think?

e5l: `offer` should throw `NetworkOnMainThreadException` to force the user to switch dispatcher manually. Other case we can fix using dispatcher from the socket context

e5l: Thanks for the investigation, I will add fixes

e5l: Done, will merge after CI build

Thomas-Vos: Great, looks good!

Thomas-Vos (approved):

PR #2165, opened 2020-10-28
KTOR-1163 Fix polymorphic serialisation client

rsinukov:
**Subsystem**
Client, serialization

**Motivation**
[Serializing collections of different element types](https://youtrack.jetbrains.com/issue/KTOR-1163)

**Solution**
Save type of request body and use it in serialization.


ktor-client/ktor-client-features/ktor-client-json/common/src/io/ktor/client/features/json/JsonSerializer.kt
```diff
@@ -21,7 +29,7 @@ public interface JsonSerializer {
     /**
      * Convert data object to [OutgoingContent].
      */
-    public fun write(data: Any): OutgoingContent = write(data, ContentType.Application.Json)
+    public fun write(data: Any): OutgoingContent = write(data, ContentType.Application.Json, null)
```
hfhbd, line 32: Maybe this old function should be deprecated

PR #2171, opened 2020-10-30
KTOR-1227 Enable build cache and cache redirector

e5l:

### DIFF
--- a/buildSrc/src/main/kotlin/CacheRedirector.kt
+++ b/buildSrc/src/main/kotlin/CacheRedirector.kt
@@ -1,0 +1,152 @@
+/*
+ * Copyright 2014-2020 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+import org.gradle.api.*
+import org.gradle.api.artifacts.dsl.*
+import org.gradle.api.artifacts.repositories.*
+import org.gradle.api.initialization.dsl.*
+import java.net.*
+
+/**
+ * Enabled via environment variable, so that it can be reliably accessed from any piece of the build script,
+ * including buildSrc within TeamCity CI.
+ */
+private val cacheRedirectorEnabled = System.getenv("CACHE_REDIRECTOR_DISABLED")?.toBoolean() != true
+
+/**
+ *  The list of repositories supported by cache redirector should be synced with the list at https://cache-redirector.jetbrains.com/redirects_generated.html
+ *  To add a repository to the list create an issue in ADM project (example issue https://youtrack.jetbrains.com/issue/IJI-149)
+ */
+private val mirroredUrls = listOf(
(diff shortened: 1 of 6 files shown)


buildSrc/src/main/kotlin/CacheRedirector.kt
```diff
@@ -0,0 +1,152 @@
...
+)
+
+private val aliases = mapOf(
+    "https://repo.maven.apache.org/maven2" to "https://repo1.maven.org/maven2", // Maven Central
+    "https://kotlin.bintray.com/kotlin-dev" to "https://dl.bintray.com/kotlin/kotlin-dev",
+    "https://kotlin.bintray.com/kotlin-eap" to "https://dl.bintray.com/kotlin/kotlin-eap",
+    "https://kotlin.bintray.com/kotlinx" to "https://dl.bintray.com/kotlin/kotlinx"
+)
+
+private fun URI.toCacheRedirectorUri() = URI("https://cache-redirector.jetbrains.com/$host/$path")
```
cy6erGn0m, line 94: `path` already starts with the leading slash so we get it twice here

buildSrc/src/main/kotlin/CacheRedirector.kt
```diff
@@ -0,0 +1,152 @@
...
+    "https://www.myget.org/F/intellij-go-snapshots/maven",
+    "https://www.myget.org/F/rd-model-snapshots/maven",
+    "https://www.myget.org/F/rd-snapshots/maven",
+    "https://www.python.org/ftp",
+    "https://www.jetbrains.com/intellij-repository/nightly",
+    "https://www.jetbrains.com/intellij-repository/releases",
+    "https://www.jetbrains.com/intellij-repository/snapshots"
+)
+
+private val aliases = mapOf(
```
cy6erGn0m, line 87: It's important to have no trailing slashes in the map so I would add a disclaimer to avoid errors when adding new entries here. Otherwise, a map transformation could be applied to remove trailing slashes.

cy6erGn0m (changes requested):

hhariri (approved):

buildSrc/src/main/kotlin/CacheRedirector.kt
e5l: I will add runtime check for that

buildSrc/src/main/kotlin/CacheRedirector.kt
e5l: Strange, but it's working for kotlinx.coroutines and kotlin, will check.

e5l: checked, both URLs(with // and with /) are working the same. I will fix for consistency

cy6erGn0m (approved):

PR #2172, opened 2020-10-30
Use JvmName for deprecated-hidden conflicting overloads

udalov:
Instead of suppressing errors, which doesn't work with JVM IR.

### DIFF
--- a/ktor-features/ktor-serialization/jvm/src/io/ktor/serialization/JsonSupport.kt
+++ b/ktor-features/ktor-serialization/jvm/src/io/ktor/serialization/JsonSupport.kt
@@ -115,8 +115,7 @@ public fun ContentNegotiation.Configuration.serialization(
  * Register kotlinx.serialization converter into [ContentNegotiation] feature
  */
 @Deprecated("Use json function instead.", ReplaceWith("json(json, contentType)"))
-@JvmName("serialization")
-public fun ContentNegotiation.Configuration.serialization0(
+public fun ContentNegotiation.Configuration.serialization(
     contentType: ContentType,
     json: Json
 ) {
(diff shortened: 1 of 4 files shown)


ktor-features/ktor-serialization/jvm/src/io/ktor/serialization/JsonSupport.kt
```diff
@@ -114,9 +114,9 @@ public fun ContentNegotiation.Configuration.serialization(
 /**
  * Register kotlinx.serialization converter into [ContentNegotiation] feature
  */
-@Suppress("CONFLICTING_OVERLOADS") // conflict with hidden declaration
 @Deprecated("Use json function instead.", ReplaceWith("json(json, contentType)"))
-public fun ContentNegotiation.Configuration.serialization(
+@JvmName("serialization")
```
cy6erGn0m, line 118: This one doesn't work since it's deprecated with WARNING level

ktor-server/ktor-server-core/jvm/src/io/ktor/http/HttpDate.kt
```diff
@@ -19,21 +19,23 @@ public fun Long.toHttpDateString(): String = GMTDate(this).toHttpDate()
...
 /**
  * Default HTTP date format
  */
-@Suppress("CONFLICTING_OVERLOADS", "REDECLARATION", "unused")
+@Suppress("unused")
 @Deprecated("Binary compatibility.", level = DeprecationLevel.HIDDEN)
-public val httpDateFormat: DateTimeFormatter
-    get() = httpDateFormat
+public val httpDateFormat0: DateTimeFormatter
+    @JvmName("httpDateFormat") get() = httpDateFormat
```
cy6erGn0m, line 41: Shouldn't it be `getHttpDateFormat`?

cy6erGn0m (changes requested):

ktor-server/ktor-server-core/jvm/src/io/ktor/http/HttpDate.kt
udalov: Facepalm. You're right, thanks! Fixed.

cy6erGn0m (approved):

PR #2174, opened 2020-10-30
KTOR-1129 Delegate to original OutgoingContent when logging body

rsinukov:
**Subsystem**
Client, Logging

**Motivation**
[Ktor 1.4.1 fails to upload file (submit form) when logging level = LogLevel.BODY](https://youtrack.jetbrains.com/issue/KTOR-1129)

**Solution**
Logging creates a new instance for the content of the request and drops its properties. This PR keeps them by delegating from new content to the old one.

### DIFF
--- a/ktor-client/ktor-client-features/ktor-client-logging/common/src/io/ktor/client/features/logging/ObservingUtils.kt
+++ b/ktor-client/ktor-client-features/ktor-client-logging/common/src/io/ktor/client/features/logging/ObservingUtils.kt
@@ -4,7 +4,6 @@
 
 package io.ktor.client.features.logging
 
-import io.ktor.http.*
 import io.ktor.http.content.*
 import io.ktor.util.*
 import io.ktor.utils.io.*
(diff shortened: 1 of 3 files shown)


ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/LoggingTests.kt
```diff
@@ -510,6 +513,42 @@ class LoggingTest : ClientLoader() {
...
+                    append("file", "123")
+                })
+            }
+
+            assertNotNull(response.receive<String>())
+            val request = response.request
+            val contentLength = request.content.contentLength!!
+            val contentType = request.content.contentType!!
+
+            assertTrue(contentType.contentType == "multipart")
```
e5l, line 544: I'm not sure if it's ok to use such checks for `logfile`. I would prefer to list the whole here, as in other logging tests.

ktor-client/ktor-client-features/ktor-client-logging/common/src/io/ktor/client/features/logging/ObservingUtils.kt
```diff
@@ -20,21 +21,35 @@ internal suspend fun OutgoingContent.observe(log: ByteWriteChannel): OutgoingCon
...
+        LoggingContent(this, responseChannel)
     }
     else -> {
         log.close()
         this
     }
 }
 
-internal class LoggingContent(private val channel: ByteReadChannel) : OutgoingContent.ReadChannelContent() {
+internal class LoggingContent(
```
e5l, line 38: Consider extracting to the separate file

e5l (approved):
lgtm

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/LoggingTests.kt
rsinukov: Unfortunately, it's not possible here. Headers are in the same line with `boundary`. This line is flaky, so we can't rely on `TestLogger`

PR #2175, opened 2020-11-02
Add JVM IR target in build_snapshot_train mode

ALikhachev:


cy6erGn0m (changes requested):
What is the reason of such huge binary changes? We definitely can't merge it to the master because it will break build... And binary incompatibility is a problem, so its fine that it's failing: that's the point

ALikhachev:
JVM IR compiler generates bytecode whose binary API does not match binary API of old compiler at the moment. Thus API for JVM IR target is placed in separate folder.

udalov:
@ALikhachev I've just realized that since we (Kotlin/JVM team) will always be running JVM IR tests with `build_snapshot_train` against Kotlin master, the API dumps of the latest release might not be that useful for us. We'll observe the difference in old vs IR backend by running tests, not by comparing text files which might be outdated. So maybe it would be better to ignore API compatibility tests for JVM IR? I suppose we can enable them back as soon as we fix all the inconsistencies. Probably this relates to all projects in the library train.

cy6erGn0m:
The other solution would be to disable it and add a separate gradle step/confiugration to run checker and see the difference between the stable compiler and the new one

ALikhachev:
During private discussion with @qwwdfsad we've decided to try another way. I'll open new PR.

PR #2181, opened 2020-11-05
KTOR-1184 Fix Autoreload feature problems

e5l:
- Fix default reload path
- Fix lambda reload

### DIFF
--- a/ktor-server/ktor-server-host-common/api/ktor-server-host-common.api
+++ b/ktor-server/ktor-server-host-common/api/ktor-server-host-common.api
@@ -36,6 +36,7 @@ public final class io/ktor/server/engine/ApplicationEngineEnvironmentBuilder {
 	public final fun getClassLoader ()Ljava/lang/ClassLoader;
 	public final fun getConfig ()Lio/ktor/config/ApplicationConfig;
 	public final fun getConnectors ()Ljava/util/List;
+	public final fun getDevelopmentMode ()Z
 	public final fun getLog ()Lorg/slf4j/Logger;
 	public final fun getModules ()Ljava/util/List;
 	public final fun getParentCoroutineContext ()Lkotlin/coroutines/CoroutineContext;
@@ -44,6 +45,7 @@ public final class io/ktor/server/engine/ApplicationEngineEnvironmentBuilder {
 	public final fun module (Lkotlin/jvm/functions/Function1;)V
 	public final fun setClassLoader (Ljava/lang/ClassLoader;)V
 	public final fun setConfig (Lio/ktor/config/ApplicationConfig;)V
+	public final fun setDevelopmentMode (Z)V
 	public final fun setLog (Lorg/slf4j/Logger;)V
 	public final fun setParentCoroutineContext (Lkotlin/coroutines/CoroutineContext;)V
 	public final fun setRootPath (Ljava/lang/String;)V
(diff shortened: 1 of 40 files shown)


ktor-utils/common/src/io/ktor/util/pipeline/Pipeline.kt
```diff
@@ -21,6 +22,11 @@ public open class Pipeline<TSubject : Any, TContext : Any>(
      */
     public val attributes: Attributes = Attributes(concurrent = true)
 
+    /**
+     * Indicated if debug mode is enabled. In debug mode users will get more details in the stacktrace.
+     */
+    public open val developmentMode: Boolean = false
```
cy6erGn0m, line 28: Isn't it better to mark it as `@Internal`? Otherwise, what is the purpose of this flag for end-users? While it looks ok in config and environment, in a pipeline it doesn't look logical and seems to be some kind internal stuff. 

Also, it will not work if someone creates his own pipelines as he doesn't override it. Hopefully, nobody does it.

And the last question: should we check for `IS_DEVELOPMENT_MODE` instead of just `false`?

ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/ServerHostUtils.kt
```diff
@@ -0,0 +1,19 @@
...
+package io.ktor.server.engine
+
+import kotlin.reflect.*
+import kotlin.reflect.jvm.*
+
+/**
+ * Obtain function FQName.
+ */
+internal fun Function<*>.methodName(): String {
+    val method = (this as? KFunction<*>)?.javaMethod ?: return "${javaClass.name}.invoke"
```
cy6erGn0m, line 14: What if a lambda is suspend? It should be invokeSuspend? Shoull we prohibit it or allow?

cy6erGn0m (approved):
LGTM, but please check what about suspend lambdas

ktor-utils/common/src/io/ktor/util/pipeline/Pipeline.kt
e5l: Yep, let's mark it with internal with error level.

e5l: If we check `IS_DEVELOPMENT_MODE`, it can use the slow version of the pipeline context accidentally. It can be the cause of errors if someone is implementing the pipeline outside of the Ktor, but I don't insist. What do you think?

ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/ServerHostUtils.kt
e5l: It's a tricky question. We don't support it right now and it will not be breaking change if we support it in the future. I would keep it as it is and get more use cases before.

rsinukov (approved):

ktor-server/ktor-server-host-common/api/ktor-server-host-common.api
```diff
@@ -56,12 +58,13 @@ public final class io/ktor/server/engine/ApplicationEngineEnvironmentKt {
 
 public final class io/ktor/server/engine/ApplicationEngineEnvironmentReloading : io/ktor/server/engine/ApplicationEngineEnvironment {
 	public static final field Companion Lio/ktor/server/engine/ApplicationEngineEnvironmentReloading$Companion;
-	public fun <init> (Ljava/lang/ClassLoader;Lorg/slf4j/Logger;Lio/ktor/config/ApplicationConfig;Ljava/util/List;Ljava/util/List;Ljava/util/List;Lkotlin/coroutines/CoroutineContext;Ljava/lang/String;)V
-	public synthetic fun <init> (Ljava/lang/ClassLoader;Lorg/slf4j/Logger;Lio/ktor/config/ApplicationConfig;Ljava/util/List;Ljava/util/List;Ljava/util/List;Lkotlin/coroutines/CoroutineContext;Ljava/lang/String;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
+	public fun <init> (Ljava/lang/ClassLoader;Lorg/slf4j/Logger;Lio/ktor/config/ApplicationConfig;Ljava/util/List;Ljava/util/List;Ljava/util/List;Lkotlin/coroutines/CoroutineContext;Ljava/lang/String;Z)V
```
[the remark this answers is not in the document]
e5l: fixed, thanks

jamesward:
By default, when I start an app I get:
> Autoreload is disabled because the development mode is off.

But with no details on how to enable development mode.  And the [auto-reload docs](https://ktor.io/docs/auto-reload.html) also does not include details on how to do that.

jamesward:
I was able to turn on development mode:
```
    val env = applicationEngineEnvironment {
        developmentMode = true
        log = LoggerFactory.getLogger("ktor.application")
        watchPaths = listOf("ktor-server/build") // todo: gross
        module(Application::module)
        connector {
            port = System.getenv("PORT")?.toInt() ?: 8080
        }
    }

    val server = embeddedServer(CIO, env)
    server.start(true)
```

And now when the classes are rebuilt, I see:
```
2021-01-13 20:19:18.950 [DefaultDispatcher-worker-1] INFO  ktor.application - Changes in application detected.
2021-01-13 20:19:19.160 [DefaultDispatcher-worker-1] INFO  ktor.application - Responding at http://0.0.0.0:8080
```

Yet, making a request to an updated route still returns the original content.  Maybe I'm doing something wrong or maybe auto-reload is just currently broken.

e5l:
Yep, to reload change it should be recompiled first. Could you tell me how you recompiled it?

jamesward:
'gradle -t classes'

PR #2182, opened 2020-11-05
KTOR-1116 Fix chunked transfer encoding handling

e5l:

### DIFF
--- a/ktor-http/ktor-http-cio/jvm/src/io/ktor/server/cio/backend/ServerPipeline.kt
+++ b/ktor-http/ktor-http-cio/jvm/src/io/ktor/server/cio/backend/ServerPipeline.kt
@@ -109,10 +109,11 @@ public fun CoroutineScope.startServerConnectionPipeline(
                 throw cause
             }
 
-            val requestBody = if (expectedHttpBody || expectedHttpUpgrade)
+            val requestBody = if (expectedHttpBody || expectedHttpUpgrade) {
                 ByteChannel(true)
-            else
+            } else {
                 ByteReadChannel.Empty
+            }
 
             val upgraded = if (expectedHttpUpgrade) CompletableDeferred<Boolean>() else null
 
@@ -155,7 +156,9 @@ public fun CoroutineScope.startServerConnectionPipeline(
                     )
                 } catch (cause: Throwable) {
                     requestBody.close(cause)
-                    throw cause
+                    response.writePacket(BadRequestPacket.copy())
+                    response.close()
+                    break
                 } finally {
                     requestBody.close()
                 }
--- a/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationCallHandler.kt
+++ b/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationCallHandler.kt
@@ -11,9 +11,12 @@ import io.ktor.server.netty.http1.*
 import io.ktor.util.pipeline.*
 import io.ktor.utils.io.*
 import io.netty.channel.*
+import io.netty.handler.codec.http.*
 import kotlinx.coroutines.*
 import org.slf4j.*
 import kotlin.coroutines.*
+
+private const val CHUNKED_VALUE  = "chunked"
 
 internal class NettyApplicationCallHandler(
     userCoroutineContext: CoroutineContext,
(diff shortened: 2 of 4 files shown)


ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationCallHandler.kt
```diff
@@ -47,7 +50,30 @@ internal class NettyApplicationCallHandler(
...
+
+        if (index + 1 != encodings.size) return false
+
+        val chunkedIsNotLast = chunkedStart + CHUNKED_VALUE.length < header.length
+        if (chunkedIsNotLast) {
+            return false
+        }
+    }
+
+    return true
```
cy6erGn0m, line 78: So looks like if we have some unsupported encodings, we accept it. Not sure about it, probably it could be fine so it will be possible to handle custom encodings in the app level, or we should fail right here with no chance. I am in doubt

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationCallHandler.kt
```diff
@@ -47,7 +50,30 @@ internal class NettyApplicationCallHandler(
...
+    val encodings = headers().getAll(io.ktor.http.HttpHeaders.TransferEncoding) ?: return true
+
+    encodings.forEachIndexed { index, header ->
+        val chunkedStart = header.indexOf(CHUNKED_VALUE)
+
+        if (chunkedStart == -1) return@forEachIndexed
+
+        if (index + 1 != encodings.size) return false
+
+        val chunkedIsNotLast = chunkedStart + CHUNKED_VALUE.length < header.length
```
cy6erGn0m, line 72: What if we have `Zchunked`: we are going to fail this but we shouldn't, right? I would also check the beginning index to see 0 index or comma, or something.. that will complicate the function for sure 😔

cy6erGn0m (changes requested):

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationCallHandler.kt
e5l: The problem is that a single entry possibly contains multiple encodings, so yep. I will add a couple of unit tests to check this method

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationCallHandler.kt
e5l: I think it's a different problem, let's discuss it separately

ktor-http/ktor-http-cio/jvm/src/io/ktor/server/cio/backend/ServerPipeline.kt
```diff
@@ -155,7 +156,9 @@ public fun CoroutineScope.startServerConnectionPipeline(
                     )
                 } catch (cause: Throwable) {
                     requestBody.close(cause)
-                    throw cause
+                    response.writePacket(BadRequestPacket.copy())
+                    response.close()
+                    break
```
cy6erGn0m, line 161: What will happen with the requests that already running? Since we don't crash here anymore, they don't get cancelled, right?

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationCallHandler.kt
```diff
@@ -47,7 +50,50 @@ internal class NettyApplicationCallHandler(
...
+        if (chunkedStart > 0 && !header[chunkedStart - 1].isSeparator()) {
+            return@forEachIndexed
+        }
+
+        val afterChunked: Int = chunkedStart + CHUNKED_VALUE.length
+        if (afterChunked < header.length && !header[afterChunked].isSeparator()) {
+                return@forEachIndexed
+        }
+
+        if (headerIndex + 1 != size) {
```
cy6erGn0m, line 85: headerIndex != lastIndex?

cy6erGn0m (changes requested):

cy6erGn0m (approved):

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationCallHandler.kt
e5l: Thanks, fixed

rsinukov (approved):

PR #2185, opened 2020-11-06
KTOR-882 - allow access to SessionIdKey  attribute

lfmunoz:
**Subsystem**

Server Core, Sessions

**Motivation**

KTOR-882 Unable to get access to sessionId

https://youtrack.jetbrains.com/issue/KTOR-882

When using SessionId the client sends the id automatically, it is used to coordinate the state on server and client. It makes no sense to make this unavailable on the server side.   Every other framework lets you see the sessionId the client is sending you.  This was made too restrictive. 

**Solution**

The solution is to make  AttributeKey<String>("SessionId") public, so that it can be used in the code. For example

```

            // this is what you currently have to do to get the sessionId
            val  attrs = call.attributes
            val key: AttributeKey<String> = attrs.allKeys.filter {
                it.name == "SessionId"
            }.map {
                it as AttributeKey<String>
            }.first()
            val sessionId = call.attributes.get(key)

           // This won't work because attributes.get needs the exact object not one that looks the same
            val sessionId = call.attributes.get(AttributeKey<String>("SessionId"))

           // The code really should just be this: this is what the pull request will allow. 
          call.attributes.get( io.ktor.sessions.SessionIdKey)

```


ktor-server/ktor-server-core/jvm/src/io/ktor/sessions/Sessions.kt
```diff
@@ -224,7 +224,7 @@ private suspend fun <S : Any> SessionProviderData<S>.sendSessionData(call: Appli
 
 private data class SessionProviderData<S : Any>(var value: S?, val incoming: Boolean, val provider: SessionProvider<S>)
 
-private val SessionKey = AttributeKey<SessionData>("SessionKey")
```
cy6erGn0m, line 227: `SessionData` class is private so there is no reason to make it's key public. Also, it breaks compilation with the following error:

> 'public' property exposes its 'private' type argument SessionData

ktor-server/ktor-server-core/jvm/src/io/ktor/sessions/SessionTrackerById.kt
```diff
@@ -9,6 +9,8 @@ import io.ktor.util.*
 import io.ktor.utils.io.*
 import kotlin.reflect.*
 
+val SessionIdKey = AttributeKey<String>("SessionId")
```
cy6erGn0m, line 12: In ktor, we use so-called "explicit mode" so all visibility modifiers are required. So "public" is required, otherwise, a compiler error is reported

```suggestion
public val SessionIdKey = AttributeKey<String>("SessionId")
```

cy6erGn0m (changes requested):
Finally, a test is required for any new feature. At least we need to test that there is a value associated with the key when there is a session, and there is no key when it shouldn't exist.

ktor-server/ktor-server-core/jvm/src/io/ktor/sessions/Sessions.kt
lfmunoz: Yes SessionData is private, making the key public won't help unless SessionData is also public and I am not sure if that is the correct thing to do.  I reverted and forgetting about this change unless you have have a better idea.

ktor-server/ktor-server-core/jvm/src/io/ktor/sessions/SessionTrackerById.kt
lfmunoz: Also explicitly specifying the type was required.

cy6erGn0m:
Will be merged as per #2263

PR #2187, opened 2020-11-08
KTOR-348 Support Java HTTP Client

dtretyakov:
The [Java HTTP Client](https://openjdk.java.net/groups/net/httpclient/intro.html) provides HTTP 1/2 support and allows using WebSockets.

**Subsystem**
New client which uses Java HTTP Client as an engine.

**Motivation**
Resolves [Support Java HTTP Client](https://youtrack.jetbrains.com/issue/KTOR-348).


e5l:
Hey, @dtretyakov. Could you rebase your branch on `1.5.0` and change the PR base?

ktor-client/ktor-client-java/jvm/src/io/ktor/client/engine/java/JavaHttpEngine.kt
```diff
@@ -0,0 +1,120 @@
...
+    override val coroutineContext: CoroutineContext
+
+    private val requestsJob = SilentSupervisor(super.coroutineContext[Job])
+
+    private var httpClient: HttpClient? = null
+
+    /**
+     * Exposed for tests only.
+     */
+    internal val executor = Executors.newCachedThreadPool {
```
e5l, line 42: Should we use this `executor` as `dispatcher` as well?

ktor-client/ktor-client-java/jvm/src/io/ktor/client/engine/java/JavaHttpEngine.kt
```diff
@@ -0,0 +1,120 @@
...
+        Thread(it, "ktor-client-java").apply {
```
e5l, line 43: It would be nice to also have some `id` number here

ktor-client/ktor-client-java/jvm/src/io/ktor/client/engine/java/JavaHttpEngine.kt
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    override fun close() {
+        super.close()
+
+        (requestsJob[Job] as CompletableJob).complete()
+    }
+
+    private fun getJavaHttpClient(data: HttpRequestData): HttpClient {
+        return httpClient ?: synchronized(this) {
```
e5l, line 82: I would extract double-check with synchronized in the separate method, but I don't insist

ktor-client/ktor-client-java/jvm/src/io/ktor/client/engine/java/JavaHttpResponse.kt
```diff
@@ -0,0 +1,27 @@
...
+    requestData: HttpRequestData
+): HttpResponseData {
+    val httpRequest = requestData.convertToHttpRequest(callContext)
+    return try {
+        sendAsync(httpRequest, JavaHttpResponseBodyHandler(callContext)).await().body()
+    } catch (e: HttpConnectTimeoutException) {
+        throw ConnectTimeoutException(requestData, e)
+    } catch (e: HttpTimeoutException) {
+        throw SocketTimeoutException(requestData, e)
+    } catch (e: Exception) {
```
e5l, line 24: Consider avoiding such variable names. Please use `cause` instead.

ktor-client/ktor-client-java/jvm/src/io/ktor/client/engine/java/JavaHttpRequestBodyPublisher.kt
```diff
@@ -0,0 +1,175 @@
...
+            if (done) {
+                return
+            }
+            if (n < 1) {
+                val ex = IllegalArgumentException(
+                    "$subscriber violated the Reactive Streams rule 3.9 by requesting "
+                        + "a non-positive number of elements."
+                )
+                signalOnError(ex)
+            } else {
```
e5l, line 68: early return instead of `else`

ktor-client/ktor-client-java/jvm/src/io/ktor/client/engine/java/JavaHttpRequestBodyPublisher.kt
```diff
@@ -0,0 +1,175 @@
...
+
+                synchronized(this) {
+                    writeInProgress = false
+                }
+            }
+        }
+
+        private fun closeChannel() {
+            try {
+                inputChannel.cancel()
```
e5l, line 138: Why do we `cancel` channel instead of closing? It can lead to losing some bytes

e5l (changes requested):
Hey @dtretyakov, thanks for the PR. LGTM in general. The solution contains some problems with following Kotlin code style and general contribution guide issues, so we should fix some things to make it easier to maintain.

ktor-client/ktor-client-java/jvm/src/io/ktor/client/engine/java/JavaHttpRequestBodyPublisher.kt
```diff
@@ -0,0 +1,175 @@
...
+                } while (outstandingDemand.decrementAndGet() > 0)
+
+                if (inputChannel.isClosedForRead) {
+                    // Reached the end of the channel, notify the subscriber and cleanup
+                    signalOnComplete()
+                    closeChannel()
+                }
+
+                synchronized(this) {
+                    writeInProgress = false
```
cy6erGn0m, line 131: There is a race here: when we request, we increment the number of permits first (`outstandingDemand`) and then check for the flag. Here we do the opposite. Consider the case: we are already saw 0 count here and going to make `writeInProgress = false` but it is `true` yet and we are just before the synchronized block and lost the time quant. The "requester" comes just after and increases the count and looking at the flag and it's true so no new coroutine is launched. After that, we are putting false to `writeInProgress` and them we quit. So we have a positive `outstandingDemand` but no coroutine is handling it. The quick solution is to re-check the count before existing the launch but avoiding double-launch.

ktor-client/ktor-client-java/jvm/src/io/ktor/client/engine/java/JavaHttpWebSocket.kt
```diff
@@ -0,0 +1,204 @@
...
+import java.util.*
+import java.util.concurrent.*
+import kotlin.coroutines.*
+import kotlin.text.String
+import kotlin.text.toByteArray
+
+private val ILLEGAL_HEADERS = TreeSet(String.CASE_INSENSITIVE_ORDER).apply {
+    addAll(
+        setOf(
+            "Sec-WebSocket-Accept",
```
cy6erGn0m, line 31: Don't we have them in `HttpHeaders`?

cy6erGn0m (changes requested):

ktor-client/ktor-client-java/jvm/src/io/ktor/client/engine/java/JavaHttpEngine.kt
dtretyakov: Thanks for noticing that, now we're using `executor` in the `dispatcher`.

ktor-client/ktor-client-java/jvm/src/io/ktor/client/engine/java/JavaHttpEngine.kt
dtretyakov: I've applied changes, now each new thread has `number` in suffux.

ktor-client/ktor-client-java/jvm/src/io/ktor/client/engine/java/JavaHttpEngine.kt
dtretyakov: I tried to find a more readable way to do that, but failed. Please feel free to fix it once it'll be merged.

ktor-client/ktor-client-java/jvm/src/io/ktor/client/engine/java/JavaHttpResponse.kt
dtretyakov: Thanks, now I'm using the `cause` name for exceptions.

ktor-client/ktor-client-java/jvm/src/io/ktor/client/engine/java/JavaHttpRequestBodyPublisher.kt
dtretyakov: Ok, did that.

ktor-client/ktor-client-java/jvm/src/io/ktor/client/engine/java/JavaHttpRequestBodyPublisher.kt
dtretyakov: Since it's `ByteReadChannel` it has only `#cancel` method. Should we `#discard` remaining data?

ktor-client/ktor-client-java/jvm/src/io/ktor/client/engine/java/JavaHttpRequestBodyPublisher.kt
```diff
@@ -0,0 +1,175 @@
...
+    private val contentLength: Long = -1,
+    private val getChannel: () -> ByteReadChannel
+) : HttpRequest.BodyPublisher {
+
+    override fun contentLength(): Long {
+        return contentLength
+    }
+
+    override fun subscribe(subscriber: Flow.Subscriber<in ByteBuffer>) {
+        println("Subscribing $subscriber")
```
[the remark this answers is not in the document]
dtretyakov: Removed forgotten `println`s.

ktor-client/ktor-client-java/jvm/src/io/ktor/client/engine/java/JavaHttpRequestBodyPublisher.kt
dtretyakov: Thanks for noticing this race condition. I've switched to using optimistic locking to avoid that:
https://github.com/ktorio/ktor/blob/4f1015e3a0ca4f7ac5dcf0725d2e3a7cfa4f1313/ktor-client/ktor-client-java/jvm/src/io/ktor/client/engine/java/JavaHttpRequestBodyPublisher.kt#L115

ktor-client/ktor-client-java/jvm/src/io/ktor/client/engine/java/JavaHttpWebSocket.kt
dtretyakov: Now all of these constants are replaced by corresponding `HttpHeaders`.

dtretyakov:
Rebased all changes on top of the `1.5.0` branch.

ktor-client/ktor-client-java/jvm/test/io/ktor/client/engine/java/JavaEngineTests.kt
```diff
@@ -0,0 +1,84 @@
...
+        val engine = JavaHttpEngine(JavaHttpConfig())
+        engine.close()
+
+        assertTrue("Java HTTP dispatcher is not working.") {
+            engine.executor.isShutdown
+        }
+    }
+
+    @Test
+    fun testThreadLeak() = runBlocking {
```
e5l, line 24: this test looks flaky

e5l (approved):
lgtm

dtretyakov:
@e5l @cy6erGn0m the [tests are passing on agent with JDK 11](https://teamcity.jetbrains.com/buildConfiguration/KotlinTools_Ktor_BuildGradleJdk8Chrome/3193306?buildTab=overview), but do we need to apply some modifications in build files to ignore `:ktor-client:ktor-client-java` module build on agents with older JDK?

ktor-client/ktor-client-java/jvm/test/io/ktor/client/engine/java/JavaEngineTests.kt
dtretyakov: Is it failing on your local machine or in CI build? The tricky part in it - to wait when SelectorManager thread exists, by default it uses polling each 3 seconds and the minimal polling interval is 1 second.

dtretyakov: https://github.com/AdoptOpenJDK/openjdk-jdk11/blob/master/src/java.net.http/share/classes/jdk/internal/net/http/HttpClientImpl.java#L827

PR #2188, opened 2020-11-08
KTOR-1264 - Add UUID to DefaultConversionService

hfhbd:
**Subsystem**
Server-Core, JVM, DefaultConversionService

**Motivation**
Add UUID to DefaultConversionService  - https://youtrack.jetbrains.com/issue/KTOR-1264
```
val id: UUID by call.parameters
```
**Solution**
Added UUID

### DIFF
--- a/ktor-utils/jvm/src/io/ktor/util/converters/ConversionServiceJvm.kt
+++ b/ktor-utils/jvm/src/io/ktor/util/converters/ConversionServiceJvm.kt
@@ -5,6 +5,7 @@
 package io.ktor.util.converters
 
 import java.math.*
+import java.util.*
 import kotlin.reflect.*
 
 @OptIn(ExperimentalStdlibApi::class)
@@ -34,6 +35,7 @@ private fun convertSimpleTypes(value: String, klass: KClass<*>): Any? = when (kl
     java.lang.Character::class -> value[0]
     BigDecimal::class -> BigDecimal(value)
     BigInteger::class -> BigInteger(value)
+    UUID::class -> UUID.fromString(value)
     else -> null
 }
 
@@ -53,6 +55,7 @@ internal actual fun platformDefaultToValues(value: Any): List<String>? {
         is java.lang.Character -> listOf(value.toString())
         is BigDecimal -> listOf(value.toString())
         is BigInteger -> listOf(value.toString())
+        is UUID -> listOf(value.toString())
         else -> null
     }
 }
--- a/ktor-utils/jvm/test/io/ktor/tests/utils/converters/DataConversionTest.kt
+++ b/ktor-utils/jvm/test/io/ktor/tests/utils/converters/DataConversionTest.kt
@@ -7,6 +7,7 @@ package io.ktor.tests.utils.converters
 import io.ktor.util.converters.*
 import io.ktor.util.reflect.*
 import java.math.*
+import java.util.*
 import kotlin.test.*
 
 class DataConversionTest {
@@ -55,5 +56,15 @@ class DataConversionTest {
 
         val toValues = DefaultConversionService.toValues(converted)
         assertEquals(listOf("B"), toValues)
+    }
+
+    @Test
+    fun testDefaultConversionUUID() {
+        val id = UUID.randomUUID()
+        val converted = DefaultConversionService.fromValues(listOf(id.toString()), typeInfo<UUID>())
+        assertEquals(id, converted)
+
+        val toValues = DefaultConversionService.toValues(converted)
+        assertEquals(listOf(id.toString()), toValues)
     }
 }


cy6erGn0m:
I remember we already had similar PR in the past. The problem with UUID is that it is JVM-only. Once one starts using it on the server, it appears to be logical to have it in client. However, in the client, it only works on JVM but not on iOS. So it's preventing seamless porting. 
On the other hand, we already have support for several JVM-only types like `BigInteger` or `BigDecimal` So we potentially can try adding it.
The other solution would be to write our own multiplatform UUID implementation, but it requires additional design and more effort to introduce.

hfhbd:
@cy6erGn0m Yeah, working on a MPP with UUID (not only ktor, but exposed etc also) is very annoying, until Kotlin will finally have a common UUID implementation...

cy6erGn0m:
Ok, let me see if we can add our own UUID at some point. I need to investiage the problem and measure costs and risks.

hfhbd:
Here is a very simple UUID4 implementation using `typealias` in JVM to support older frameworks. Alternativ: Use converter function like in `Kotlinx.Datetime`.

```
// Common
expect class UUID

//JVM
actual typealias UUID = java.util.UUID

// JS
actual /* inline */ class UUID(private val uuidString: String) {

    init {
        require(uuidString matches uuid4Regex) { "$uuidString is not a UUIDv4" }
    }

    fun version(): Int = 4
    fun variant(): Int = 1

    companion object {
        private val uuid4Regex = Regex("[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}", RegexOption.IGNORE_CASE)

        fun randomUUID(): UUID {
            val randomBytes = Random.nextUBytes(16)
            val clearVersion: UByte = 0x0fu
            val version4: UByte = 0x40u
            val clearVariant: UByte = 0x3fu
            val variant: UByte = 0x80u
            randomBytes[6] = randomBytes[6] and clearVersion
            randomBytes[6] = randomBytes[6] or version4
            randomBytes[8] = randomBytes[8] and clearVariant
            randomBytes[8] = randomBytes[8] or variant
            return UUID(randomBytes.formatUUID())
        }

        private fun UByteArray.formatUUID(): String = joinToString("") {
            val char = it.toInt()
            if (char <= 15) {
                "0" + char.toString(16)
            } else {
                char.toString(16)
            }
        }.let {
            StringBuilder(it).apply {
                insert(8, "-")
                insert(12 + 1, "-")
                insert(16 + 2, "-")
                insert(20 + 3, "-")
            }.toString()
        }
    }

    override fun toString() = uuidString
}

**Comment by cy6erGn0m** [comment]
Still a lot of open questions anyway:
- why it is secure on JVM and it is not on JS?
- what if one needs a hash-based UUID?
- what is hashCode/equals contract? Should it be the same across platforms? Comparable?
- what about serialization? How does one serialize it with kotlinx.serialization?
- finally, why it is ktor who holds the type? shouldn't it be a separate 3rd party lib or even stdlib (no)?

**Comment by cy6erGn0m** [comment]
So, the first question: why do you need it? What is your use-case, why don't you use regular String and why it is so boring so that you made this PR?

**Comment by hfhbd** [comment]
Thank you for your feedback!

One basic use case is a multiplatform application, eg a Todo service.
The backend uses Ktor + Exposed on JVM.
The frontend uses Kotlin/JS, without a db, because its always online, ignoring offline web-apps...
The Android app uses Room as DB with common code and the iOS app Kotlin/Native and CoreData as DB.

The main advantage using UUID over an Integer as the identifier is the possibility creating a new Todo offline without sync conflicts. Because a UUIDv4 is "unique", the probability having the same UUID in your application scope is extremely low (see https://en.wikipedia.org/wiki/Universally_unique_identifier#Collisions). You can upload this local generated id directly without problems to you server and store them in the server DB. The CloudKit framework by Apple uses this advantage.

Creating a UUID:
- on JVM: Using Java UUID with SecureRandom
- on Android: Using Java UUID with SecureRandom
- on iOS/Native: Using Foundation with `/dev/urandom` [under the hood](https://github.com/apple/swift-corelibs-foundation/blob/66ed50829f38bce2829b80d77adaaa4d44da1d83/Sources/UUID/uuid.c#L157)
- on JS: Using insecure random generator, because https://github.com/Kotlin/KEEP/issues/184

Compareable, hashable: 
A UUID is either a 128 bit string or a 32 hex string (depending on the implementation), so comparing, hashCode/equals should always use the (bit/hex) string which has to be the same over all platforms.

Support for other UUID versions:
Yes, for a full design all versions needed to be supported.

Serialization:
This is indeed an open question. For JSON for example you might want to use the hex string as string, for photobuf you should use the bit string, see
https://github.com/Kotlin/kotlinx.serialization/pull/1184. So when creating the expected builtin support needs to keep this in mind.

Absolute, it is way lot more difficult to design this "simple" class for public use with binary compatibility than implementing a UUID class in a non-public project without binary compatibility, which needs only UUIDv4 :)

I would love to see a UUID implementation in the Standard lib instead in Ktor! (https://youtrack.jetbrains.com/issue/KT-43042)

But unfortunately a UUID is still not implemented in the std lib.
Like [base64Encoding](https://youtrack.jetbrains.com/issue/KT-9823), which is also missing in the std lib, but implement by yourself, you/I could implement a UUID for ktor.


Back to the initial "problem" of this pull request :)
As a current workaround, you could overload the `getValue` function.
```kotlin
operator fun Parameters.getValue(receiver: Any?, property: KProperty<*>): java.util.UUID =
    UUID.fromString(get(property.name))
```
I simple created this PR because in ConversionService you already have JVM only types.

cy6erGn0m:
@hfhbd See https://github.com/cy6erGn0m/kotlinx-uuid for prevew

hfhbd:
@cy6erGn0m WOW! I did not see this coming so fast.

cy6erGn0m:
@hfhbd Remember that it's not even a real kotlinx library: it's not under the organization and it's status is undecided yet. I am planning to publish it to a separate bintray repo next week but I can't guarantee anything else for now.

cy6erGn0m:
@hfhbd added exposed support to the library

hfhbd:
@cy6erGn0m Thanks! I did try to used it with my playground/experimental app [ComposeTodo](https://github.com/hfhbd/ComposeTodo), and releasing the artifacts would be super.

hfhbd:
@rsinukov Rebased on `2.0.0-eap`.

rsinukov (approved):

rsinukov:
Good job, thanks!

PR #2191, opened 2020-11-09
Rewrite ApacheRequestProducer

rsinukov:
**Subsystem**
Client, Apache

**Motivation**
[Apache client engine sometimes hits an unrecoverable socket timeout when using ChannelWriterContent](https://youtrack.jetbrains.com/issue/KTOR-1149)

**Solution**
There was a race condition in `ApacheRequestProducer`. Fixed by rewriting it to a better approach.

### DIFF
--- a/ktor-io/jvm/src/io/ktor/utils/io/ByteBufferChannel.kt
+++ b/ktor-io/jvm/src/io/ktor/utils/io/ByteBufferChannel.kt
@@ -601,6 +601,45 @@ internal open class ByteBufferChannel(
         }
     }
 
+    override fun readAvailable(min: Int, block: (ByteBuffer) -> Unit): Int {
+        require(min > 0) { "min should be positive" }
+        require(min <= BYTE_BUFFER_CAPACITY) { "Min($min) shouldn't be greater than $BYTE_BUFFER_CAPACITY" }
+
+        var result = 0
+        val read = reading { state ->
+            val locked = state.tryReadAtLeast(min)
+
+            if (locked > 0 && locked >= min) {
+                // here we have locked all available for read bytes
+                // however we don't know how many bytes will be actually read
+                // so later we have to return (locked - actuallyRead) bytes back
+
+                // it is important to lock bytes to fail concurrent tryLockForRelease
+                // once we have locked some bytes, tryLockForRelease will fail so it is safe to use buffer
+
+                val position = position()
+                val limit = limit()
+                block(this)
+                if (limit != limit()) throw IllegalStateException("buffer limit modified")
+
+                result = position() - position
+                if (result < 0) throw IllegalStateException("position has been moved backward: pushback is not supported")
+
+                bytesRead(state, result)
+
+                if (result < locked) {
+                    state.completeWrite(locked - result) // return back extra bytes (see note above)
+                    // we use completeWrite in spite of that it is read block
+                    // we don't need to resume read as we are already in read block
+                }
+                true
+            } else false
+        }
+
+        if (!read) return -1
+        return result
+    }
+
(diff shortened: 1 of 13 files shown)


cy6erGn0m:
Sorry for Russian but

> безумству храбрых поём мы песню

ktor-io/jvm/src/io/ktor/utils/io/ByteReadChannelJVM.kt
```diff
@@ -236,6 +252,9 @@ public actual interface ByteReadChannel {
         max: Long
     ): Long
 
+    @ExperimentalIoApi
```
e5l, line 255: This annotation is deprecated, please consider introducing an issue of why this API is experimental and create a new annotation for that.

ktor-io/jvm/src/io/ktor/utils/io/ByteChannelSequentialJVM.kt
```diff
@@ -83,6 +83,25 @@ public class ByteChannelSequentialJVM(initial: IoBuffer, autoFlush: Boolean) : B
         return readAvailableSuspend(dst)
     }
 
+    override fun readAvailable(min: Int, block: (ByteBuffer) -> Unit): Int {
```
e5l, line 86: why do we need `min` here?

ktor-io/jvm/src/io/ktor/utils/io/ByteBufferChannel.kt
```diff
@@ -601,6 +601,45 @@ internal open class ByteBufferChannel(
         }
     }
 
+    override fun readAvailable(min: Int, block: (ByteBuffer) -> Unit): Int {
```
e5l, line 604: Unit tests required for all public API

ktor-io/jvm/src/io/ktor/utils/io/ByteBufferChannel.kt
```diff
@@ -601,6 +601,45 @@ internal open class ByteBufferChannel(
...
+                if (result < locked) {
+                    state.completeWrite(locked - result) // return back extra bytes (see note above)
+                    // we use completeWrite in spite of that it is read block
+                    // we don't need to resume read as we are already in read block
+                }
+                true
+            } else false
+        }
+
+        if (!read) return -1
```
e5l, line 639: Why do we need to distinguish `-1` from `0` here?

e5l (changes requested):
LGTM, Please add unit tests for the new ByteChannel methods.

ktor-io/jvm/src/io/ktor/utils/io/ByteChannelSequentialJVM.kt
rsinukov: I added it to mirror API and behavior of `writeAvailable`. Nut I also think that it's useful and we can have it on its own.

ktor-io/jvm/src/io/ktor/utils/io/ByteBufferChannel.kt
rsinukov: 1. It mirrors API and behavior of `writeAvailable`.
2. As specified in docs, `@return number of consumed bytes or -1 if the block wasn't executed.`. This is behavior that we rely on to know if we had enough bytes to read or should we suspend to await data.

ktor-io/jvm/src/io/ktor/utils/io/ByteBufferChannel.kt
rsinukov: done

e5l (approved):
lgtm

PR #2192, opened 2020-11-09
KTOR-1229 Fix mapping of SocketTimeoutException in Android engine

rsinukov:
**Subsystem**
Client, Android

**Motivation** 
[Timeout﻿ feature: android engine throws Java's SocketTimeoutException instead of ConnectTimeoutException](https://youtrack.jetbrains.com/issue/KTOR-1229)

**Solution**
Not only `connect()`, but also `statusCode` and other methods on `HttpURLConnection` can throw an exception that we need to map. This PR wraps all of them in a block.

Note: Not sure how to write a proper test for it, since only some versions of Android has this behavior.


e5l (changes requested):
lgtm, please add tests

rsinukov:
@e5l unit test added

e5l (approved):

PR #2092, opened 2020-09-25
 Add freeze calls to iOS engine

Thomas-Vos:
**Subsystem**
`io.ktor:ktor-client-ios`

**Motivation**
This PR makes it possible to make network requests from `Dispatchers.Default` without getting `IncorrectDereferenceException`. See: https://kotlinlang.slack.com/archives/C0A974TJ9/p1600972836014600

**Solution**
Add some freeze calls to the iOS engine.

PR #2193, opened 2020-11-11
Fix IncorrectDereferenceException when using HttpClient on Dispatchers.Default

Thomas-Vos:
**Subsystem**
Client, native

**Motivation**
This PR fixes an `IncorrectDereferenceException` when using the HttpClient on `Dispatchers.Default`. The bug was caused by https://github.com/ktorio/ktor/pull/2110 that added a top level value not marked as `@SharedImmutable`. 

HttpClient on `Dispatchers.Default` was supposed to be fixed in Ktor 1.4.2 with https://github.com/ktorio/ktor/pull/2092 but as explained above unfortunately another issue was introduced. That is why I also added a test to prevent issues like this in the future.

https://kotlinlang.slack.com/archives/C0A974TJ9/p1605058326000100

**Solution**
Mark the top level value with `@SharedImmutable` and add a test.

### DIFF
--- a/ktor-client/ktor-client-tests/posix/test/io/ktor/client/tests/HttpClientFreezingTest.kt
+++ b/ktor-client/ktor-client-tests/posix/test/io/ktor/client/tests/HttpClientFreezingTest.kt
@@ -8,6 +8,7 @@ import io.ktor.client.*
 import io.ktor.client.request.*
 import io.ktor.client.tests.utils.*
 import io.ktor.utils.io.*
+import kotlinx.coroutines.*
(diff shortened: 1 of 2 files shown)


Thomas-Vos:
@e5l could you review this?

atripathi-chwy (approved):
Tested this locally. It fix the issue I was having https://kotlinlang.slack.com/archives/C0A974TJ9/p1605058326000100

rsinukov (approved):
Thanks for the contribution, LGTM

e5l (approved):

PR #2197, opened 2020-11-12
KTOR-800 Fix Netty HTTP/2 push to not crash on HTTP/2 reset

cy6erGn0m:
**Subsystem**
ktor-server-netty

**Motivation/Problem**
[KTOR-800](https://youtrack.jetbrains.com/issue/KTOR-800) HTTP/2 push fails with netty engine 
When the netty backend initiates a push promise, it creates several objects for streaming a sub-channel. However, one of netty's streams, an `Http2Stream` is only created after sending a push promise frame and it's not initialized in time. When a client sends reject frame to cancel the push stream, it leads to a crash on server side inside of the Netty internals. 

Notice that if a client doesn't reject a push promise, then it works well except for that there is a chance that HTTP/2 control frames didn't work.

**Soltion**
To avoid this unfortunate behaviour, we need to initialize all the stuff properly. Unfortunately, Netty itself still doesn't provide the complete push promise support. I would say that the chance of the proper implemetation on the Netty side is so poor so we have no other chice but to do some black magic to eliminate exceptions.

Even worse, we have no ability to test incoming push promises at the moment while writing an HTTP/2 client with such support is a serious work. This is why the only solution to check it is to manually test this. The existing ktor HTTP/2 client implementation (based on Jetty) does always reject (just as browser do for text/plain). So in the test suite we are doing requests for both resources and the test checks if we are failing or not in case of rejection that is the topic of the issue.

A "weak" evidence from my manual testing: notice the CSS file that is "pushed" from server:

Screenshot 2020-11-12 at 18 45 34

The original error from the ticket disappeaed and no more `IllegalStateException` complaining about stream object.

### DIFF
--- a/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/http2/NettyHttp2Handler.kt
+++ b/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/http2/NettyHttp2Handler.kt
@@ -123,7 +123,14 @@ internal class NettyHttp2Handler(
         child.setId(promisedStreamId)
 
         val promise = rootContext.newPromise()
-        codec.encoder().writePushPromise(rootContext, streamId, promisedStreamId, headers, 0, promise)
+        val childStream = connection.local().createStream(promisedStreamId, false)
+        if (!child.stream().setStreamAndProperty(codec, childStream)) {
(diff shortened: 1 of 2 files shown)


e5l (approved):
LGTM

PR #2199, opened 2020-11-16
Add build parameter to build ktor with JVM IR compiler

ALikhachev:
To be used in specific kotlinx-train CI configuration. It's needed for faster adoption of JVM IR compiler.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -29,6 +29,10 @@ buildscript {
             }
         }
     }
+    // These two flags are enabled in train builds for JVM IR compiler testing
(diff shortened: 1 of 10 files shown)


cy6erGn0m (changes requested):
Looks like we lost some tests: we had 5600+ tests but in the branch, we got 60 less tests

ALikhachev:
> Looks like we lost some tests: we had 5600+ tests but in the branch, we got 60 less tests

Tests were lost because base commit was too old. Now PR is rebased on current master.

PR #2200, opened 2020-11-16
KTOR-1247 - Fix writing race condition in ByteBufferChannel

rsinukov:
**Subsystem**
Client/Server, ByteBufferChannel

**Motivation**
[KTOR-1247](https://youtrack.jetbrains.com/issue/KTOR-1247)
Race condition:

1. writer calls `channel.writeByte` 
2. writer goes inside `setupStateForWrite` 
3. no bytes available for write -> `buffer.limit() == buffer.position()`
4. reader reads content, creates space for the writer
5. writer checks capacity in state -> success, tries to write to a preconfigured buffer that has no available space
6. crash

**Solution**
* Move setting up buffer position and limit after checking of capacity to avoid race. 
* Unified behavior of `writeByte/Short/Int/Long`
* Cleaned up some unused methods and parameters

### DIFF
--- a/ktor-io/jvm/src/io/ktor/utils/io/ByteBufferChannel.kt
+++ b/ktor-io/jvm/src/io/ktor/utils/io/ByteBufferChannel.kt
@@ -886,26 +886,26 @@ internal open class ByteBufferChannel(
         }
     }
 
-    internal fun bytesWrittenFromSesion(buffer: ByteBuffer, c: RingBufferCapacity, n: Int) {
-        buffer.bytesWritten(c, n)
+    internal fun bytesWrittenFromSesion(buffer: ByteBuffer, capacity: RingBufferCapacity, count: Int) {
+        buffer.bytesWritten(capacity, count)
     }
 
-    private fun ByteBuffer.bytesWritten(c: RingBufferCapacity, n: Int) {
-        require(n >= 0)
+    private fun ByteBuffer.bytesWritten(capacity: RingBufferCapacity, count: Int) {
+        require(count >= 0)
 
-        writePosition = carryIndex(writePosition + n)
-        c.completeWrite(n)
+        writePosition = carryIndex(writePosition + count)
+        capacity.completeWrite(count)
         @Suppress("DEPRECATION")
-        totalBytesWritten += n
+        totalBytesWritten += count
     }
 
-    private fun ByteBuffer.bytesRead(c: RingBufferCapacity, n: Int) {
-        require(n >= 0)
+    private fun ByteBuffer.bytesRead(capacity: RingBufferCapacity, count: Int) {
+        require(count >= 0)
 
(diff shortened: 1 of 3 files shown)


ktor-io/jvm/src/io/ktor/utils/io/ByteBufferChannel.kt
```diff
@@ -1129,6 +1109,16 @@ internal open class ByteBufferChannel(
...
         if (c.tryWriteExact(8)) {
+            prepareWriteBuffer(this, 8)
+            doWrite(this, l, c)
+            return true
+        }
+
+        return false
+    }
+
+    private fun doWrite(buffer: ByteBuffer, l: Long, c: RingBufferCapacity) {
```
e5l, line 1120: `c -> capacity`
`l -> limit`

ktor-io/jvm/src/io/ktor/utils/io/ByteBufferChannel.kt
```diff
@@ -1129,6 +1109,16 @@ internal open class ByteBufferChannel(
 
     private fun ByteBuffer.tryWriteLong(l: Long, c: RingBufferCapacity): Boolean {
         if (c.tryWriteExact(8)) {
```
e5l, line 1111: please invert condition and return early

e5l (changes requested):
LGTM, please make some cleanup.

It's really strange that `joining` state was unused

ktor-io/jvm/src/io/ktor/utils/io/ByteBufferChannel.kt
```diff
@@ -906,24 +882,24 @@ internal open class ByteBufferChannel(
         }
     }
 
-    internal fun bytesWrittenFromSesion(buffer: ByteBuffer, c: RingBufferCapacity, n: Int) {
-        buffer.bytesWritten(c, n)
+    internal fun bytesWrittenFromSesion(buffer: ByteBuffer, capacity: RingBufferCapacity, n: Int) {
```
e5l, line 885: `n` could be also renamed

ktor-io/jvm/src/io/ktor/utils/io/ByteBufferChannel.kt
```diff
@@ -956,31 +932,32 @@ internal open class ByteBufferChannel(
         }
     }
 
-    override suspend fun writeByte(b: Byte) {
-        joining?.let { resolveDelegation(this, it)?.let { return it.writeByte(b) } }
+    override suspend fun writeByte(value: Byte) {
```
e5l, line 935: public parameter names shouldn't be renamed

e5l (approved):
LGTM, Please fix names

cy6erGn0m (approved):

PR #2202, opened 2020-11-17
KTOR-1290 Specialize shared usage to native platform in perf

e5l:
Remove shared in perf critical io

Fixes [KTOR-1290](https://youtrack.jetbrains.com/issue/KTOR-1290)

### DIFF
--- a/ktor-io/common/src/io/ktor/utils/io/core/AbstractInput.kt
+++ b/ktor-io/common/src/io/ktor/utils/io/core/AbstractInput.kt
@@ -3,7 +3,6 @@
 package io.ktor.utils.io.core
 
 import io.ktor.utils.io.bits.*
-import io.ktor.utils.io.concurrent.*
 import io.ktor.utils.io.core.internal.*
 import io.ktor.utils.io.core.internal.require
 import io.ktor.utils.io.pool.*
@@ -17,6 +16,7 @@ public abstract class AbstractInput(
     remaining: Long = head.remainingAll(),
     public val pool: ObjectPool<ChunkBuffer> = ChunkBuffer.Pool
 ) : Input {
+    private val state = AbstractInputSharedState(head, remaining)
 
     @Suppress("DEPRECATION")
     @Deprecated("Binary compatibility.", level = DeprecationLevel.HIDDEN)
@@ -44,14 +44,13 @@ public abstract class AbstractInput(
     /**
      * Current head chunk reference
      */
-    private final var __head: ChunkBuffer by shared(head)
     private final var _head: ChunkBuffer
-        get() = __head
+        get() = state.head
         set(newHead) {
-            __head = newHead
-            headMemory = newHead.memory
-            headPosition = newHead.readPosition
-            headEndExclusive = newHead.writePosition
+            state.head = newHead
+            state.headMemory = newHead.memory
+            state.headPosition = newHead.readPosition
+            state.headEndExclusive = newHead.writePosition
         }
 
     @PublishedApi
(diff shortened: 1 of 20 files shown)


ktor-io/posix/src/io/ktor/utils/io/ByteChannelSequentialBaseSharedStateNative.kt
```diff
@@ -0,0 +1,28 @@
...
+
+    actual var totalBytesRead: Long by shared(0L)
+
+    actual var totalBytesWritten: Long by shared(0L)
+
+    actual var closedCause: Throwable? by shared(null)
+
+    actual var lastReadAvailable: Int by shared(0)
+
+    actual var lastReadView: ChunkBuffer by shared(ChunkBuffer.Empty)
```
whyoleg, line 26: Why not to use plain atomics on native, as those classes already splitted by js/jvm/native?
If it was a problem for JVM, may be it will also have perf problem for native, to create all those objects all time, and so on?

e5l: It shouldn't be a big problem for native, but let's change them to atomics

ktor-io/common/src/io/ktor/utils/io/core/AbstractInput.kt
```diff
@@ -7,6 +7,29 @@ import io.ktor.utils.io.concurrent.*
 import io.ktor.utils.io.core.internal.*
 import io.ktor.utils.io.core.internal.require
 import io.ktor.utils.io.pool.*
+import kotlin.Boolean
```
cy6erGn0m, line 10: A?

ktor-io/common/src/io/ktor/utils/io/core/AbstractInput.kt
```diff
@@ -44,14 +68,13 @@ public abstract class AbstractInput(
     /**
      * Current head chunk reference
      */
-    private final var __head: ChunkBuffer by shared(head)
     private final var _head: ChunkBuffer
-        get() = __head
+        get() = state.head
```
cy6erGn0m, line 72: I wonder why didn't we seriously slow down because of this? We added an indirection here (`by shared`) so this should be much slower, isn't it?

cy6erGn0m (changes requested):
The overall impression is "delegate every single property". Doesn't look good. The original issue is that we added both unnecessary atomic and indirection and what a shame we can only fix atomics. 
The other alternative would be to specialize classes itself instead of states but not sure how do we reuse the implementation. So, seems to be we have no big choice other than that, right?  So let's fix imports and let's go with this until we get better K/N

ktor-io/common/src/io/ktor/utils/io/core/AbstractInput.kt
e5l: thanks, will remove

ktor-io/common/src/io/ktor/utils/io/core/AbstractInput.kt
e5l: We removed allocation at least, and indirection is in the final field

cy6erGn0m (approved):

ktor-io/posix/src/io/ktor/utils/io/ByteChannelSequentialBaseSharedStateNative.kt
whyoleg: @e5l can you look at https://youtrack.jetbrains.com/issue/KTOR-1049#focus=Comments-27-4565293.0-0
Seems, that this change improves performance on native. I don't see any other code change, that can cause such perf improvement. Or do you know, what other can increase it?

PR #2203, opened 2020-11-17
KTOR-1236 Remove unnecessary escaping in the query parameter value

e5l:
Fixes [KTOR-1236](https://youtrack.jetbrains.com/issue/KTOR-1236)

### DIFF
--- a/ktor-http/common/test/io/ktor/tests/http/UrlTest.kt
+++ b/ktor-http/common/test/io/ktor/tests/http/UrlTest.kt
@@ -122,6 +122,18 @@ class UrlTest {
     }
 
     @Test
+    fun testEqualsInQueryValue() {
+        val urlString =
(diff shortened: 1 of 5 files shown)


rsinukov (approved):

ktor-http/common/test/io/ktor/tests/http/UrlTest.kt
```diff
@@ -121,6 +121,18 @@ class UrlTest {
         assertEquals("/response-headers?message=foo%25bar", url.fullPath)
     }
 
+    @Test
+    fun testEqualsInQueryValue() {
+        val urlString =
+            "https://akamai.bintray.com/22/225b067044aa56f36590ef56d41e256cd1d0887b176bfdeec123ecccc6057790?__gda__=exp=1604350711~hmac=417cbd5a97b4c499e2cf7e9eae5dfb9ad95b42cb3ff76c5fb0fae70e2a42db9c&..."
```
rsinukov, line 127: can you simplify this test value?

rsinukov (approved):

ktor-http/common/test/io/ktor/tests/http/UrlTest.kt
e5l: Yep, but I would like to keep it as it is in the origin issue

PR #2205, opened 2020-11-17
KTOR-1127 Rewrite timeout impl for CIO

rsinukov:
**Motivation**
[KTOR-1127 CIO: client engine exceptions are both logged and thrown](https://youtrack.jetbrains.com/issue/KTOR-1127)

**Solution**
The previous implementation of socket timeout was very ineffective and allocated a lot of unnecessary objects. Rewrote it to a more low level one.

### DIFF
--- a/ktor-network/jvm/src/io/ktor/network/sockets/CIOReader.kt
+++ b/ktor-network/jvm/src/io/ktor/network/sockets/CIOReader.kt
@@ -6,12 +6,12 @@ package io.ktor.network.sockets
 
 import io.ktor.network.selector.*
 import io.ktor.network.util.*
-import kotlinx.coroutines.*
 import io.ktor.utils.io.*
 import io.ktor.utils.io.ByteChannel
 import io.ktor.utils.io.core.*
 import io.ktor.utils.io.nio.*
 import io.ktor.utils.io.pool.*
+import kotlinx.coroutines.*
 import java.nio.*
 import java.nio.channels.*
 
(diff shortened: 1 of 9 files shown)


ktor-network/jvm/src/io/ktor/network/sockets/CIOReader.kt
```diff
@@ -26,19 +27,24 @@ internal fun CoroutineScope.attachForReadingImpl(
     val buffer = pool.borrow()
     return writer(Dispatchers.Unconfined + CoroutineName("cio-from-nio-reader"), channel) {
         try {
+            val timeout = if (socketOptions?.socketTimeout != null) {
```
e5l, line 30: Let's discuss why we need this coroutine at all

e5l (changes requested):

ktor-network/jvm/src/io/ktor/network/sockets/CIOReader.kt
rsinukov: discussed offline

ktor-utils/jvm/src/io/ktor/util/date/DateJvm.kt
```diff
@@ -64,3 +64,8 @@ public fun Calendar.toDate(timestamp: Long?): GMTDate {
  * Convert to [Date]
  */
 public fun GMTDate.toJvmDate(): Date = Date(timestamp)
+
+/**
+ * Gets current system time in milliseconds since certain moment in the past, only delta between two subsequent calls makes sense.
+ */
+public actual fun getTimeMillis(): Long = System.currentTimeMillis()
```
e5l, line 71: Why we can't use GMTDate.timestamp in common?

ktor-network/jvm/src/io/ktor/network/util/Utils.kt
```diff
@@ -4,23 +4,86 @@
...
+    return Timeout(name, timeoutMs, clock, this, onTimeout)
+}
+
+internal inline fun <T> Timeout?.withTimeout(block: () -> T): T {
+    if (this == null) {
+        return block()
     }
+
+    start()
+    val result = block()
```
e5l, line 86: try finally?

e5l (approved):

ktor-utils/jvm/src/io/ktor/util/date/DateJvm.kt
rsinukov: to save on allocations

PR #2207, opened 2020-11-18
Reformat ktor-http to conform to ktlint

cy6erGn0m:
**Subsystem**
ktor-http (all)

**Motivation**
This is an attempt to make `ktor-http` module conform to [ktlint](https://github.com/pinterest/ktlint). This is under question if we want such changes. Please look at whet I was forced to do.

The only disabled rule is "no-wildcard-imports" to make start imports work.

### DIFF
--- a/ktor-http/common/src/io/ktor/http/CacheControl.kt
+++ b/ktor-http/common/src/io/ktor/http/CacheControl.kt
@@ -98,14 +98,14 @@ public sealed class CacheControl(public val visibility: Visibility?) {
         }
 
         override fun equals(other: Any?): Boolean {
-            return other === this ||
-                (other is MaxAge &&
+            return other === this || (
+                other is MaxAge &&
                     other.maxAgeSeconds == maxAgeSeconds &&
                     other.proxyMaxAgeSeconds == proxyMaxAgeSeconds &&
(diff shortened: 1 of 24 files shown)


ktor-server/ktor-server-test-suites/jvm/src/io/ktor/server/testing/suites/ContentTestSuite.kt
```diff
@@ -235,6 +236,20 @@ public abstract class ContentTestSuite<TEngine : ApplicationEngine, TConfigurati
         }
     }
 
+    @Test
+    fun testReceiveBlocking() {
```
rsinukov, line 240: looks like it is not related to formatting, but ok for me

rsinukov (approved):

ktor-server/ktor-server-test-suites/jvm/src/io/ktor/server/testing/suites/ContentTestSuite.kt
cy6erGn0m: That wasn't intentional! Thanks!

e5l (approved):
please create an issue and refer it in the commit message.

PR #2211, opened 2020-11-20
KTOR 1344 - Unified pre-typed routes

hfhbd:
**Subsystem**
Server - Routing

**Motivation**
Currently, only `post` has a typed overload. [KTOR-1344](https://youtrack.jetbrains.com/issue/KTOR-1344)

**Solution**
Added a typed overload to `put` and `patch` as well

### DIFF
--- a/ktor-server/ktor-server-core/jvm/src/io/ktor/routing/RoutingBuilder.kt
+++ b/ktor-server/ktor-server-core/jvm/src/io/ktor/routing/RoutingBuilder.kt
@@ -118,11 +118,15 @@ public fun Route.post(path: String, body: PipelineInterceptor<Unit, ApplicationC
 @ContextDsl
 @JvmName("postTyped")
 public inline fun <reified R : Any> Route.post(
-    path: String,
+    path: String? = null,
     crossinline body: suspend PipelineContext<Unit, ApplicationCall>.(R) -> Unit
 ): Route {
-    return route(path, HttpMethod.Post) {
-        handle {
+    return if(path != null) {
+        post(path) {
+            body(call.receive())
+        }
+    } else {
+        post {
             body(call.receive())
         }
     }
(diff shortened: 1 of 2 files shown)


ktor-server/ktor-server-core/jvm/src/io/ktor/routing/RoutingBuilder.kt
```diff
@@ -118,11 +118,15 @@ public fun Route.post(path: String, body: PipelineInterceptor<Unit, ApplicationC
 @ContextDsl
 @JvmName("postTyped")
 public inline fun <reified R : Any> Route.post(
-    path: String,
+    path: String? = null,
     crossinline body: suspend PipelineContext<Unit, ApplicationCall>.(R) -> Unit
 ): Route {
-    return route(path, HttpMethod.Post) {
-        handle {
+    return if(path != null) {
```
cy6erGn0m, line 124: The problem here is that the function is inline so introducing `if` here you duplicate the body bytecode and duplicate user's bytecode everywhere regardless of whether they need the `path` or not.

cy6erGn0m (changes requested):
LGTM except for bytecode duplication

ktor-server/ktor-server-core/jvm/src/io/ktor/routing/RoutingBuilder.kt
hfhbd: So you need multiple overloads? One with path, one without?

cy6erGn0m:
@rsinukov This generally looks fine but I feel these function may/will clash with locations functions so I wonder if we can somehow workaround it? The functions contributed seems useful but I don't know how do we avoid ambiguity

rsinukov:
@cy6erGn0m it looks ok for me. Locations builders and these ones are in different packages, so users can always specify what function they want to use. We already have this for `post` and we never had any complaints about it.

rsinukov (changes requested):
Please run `./gradlew apiDump` and change the target branch to 1.6.0

hfhbd:
@rsinukov Changed. apiDump did not update my changes, because all functions are `inline reified`.

rsinukov (approved):

cy6erGn0m (approved):

cy6erGn0m:
@hfhbd Ok, let's proceed with merge. Could you please rebase it and skip irrelevant commits? Notice the base branch 1.6.0-eap

hfhbd:
Dont merge, one second

hfhbd:
Okay, I don't care about the username
@cy6erGn0m rebased

PR #2212, opened 2020-11-21
KTOR-730 Add baseURL option to DefaultRequest

hfhbd:
**Subsystem**
Client Core - DefaultRequest

**Motivation**
The `defaultRequest` feature is missing a `baseURL` setting, see [KTOR-730](https://youtrack.jetbrains.com/issue/KTOR-730)

**Solution**
Add a baseURL option to `HttpRequestBuilder`

- [x] ApiDump

### DIFF
--- a/ktor-client/ktor-client-core/common/test/io/ktor/client/DefaultRequestTest.kt
+++ b/ktor-client/ktor-client-core/common/test/io/ktor/client/DefaultRequestTest.kt
@@ -1,0 +1,78 @@
+/*
+ * Copyright 2014-2020 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.client
+
+import io.ktor.client.plugins.*
+import io.ktor.client.plugins.DefaultRequest.Plugin.install
+import io.ktor.client.request.*
+import io.ktor.http.*
+import kotlin.test.*
+
+class DefaultRequestTest {
+    private fun HttpRequestBuilder.defaultRequest(block: DefaultRequest.Builder.() -> Unit) = apply {
+        install(block)
+    }
+
+    @Test
+    fun testUsingBaseURLMatchingCurrentPipelineBehaviour() {
+        val todoURL = HttpRequestBuilder().apply {
+            url.takeFrom(URLBuilder("/sub").build())
+        }.defaultRequest {
+            baseURL("https://ktor.io/docs")
+        }.build().url
+        assertEquals("https://ktor.io/docs/sub", todoURL.toString())
+    }
+
+    @Test
+    fun overrideOther() {
+        val overridden = HttpRequestBuilder().apply {
+            url.takeFrom("https://kotlinlang.org/blog")
+        }.defaultRequest {
+            baseURL("https://ktor.io/docs")
+        }.build().url
+        assertEquals("https://kotlinlang.org/blog", overridden.toString())
+    }
+
+    @Test
+    fun overriddenLocalhostOnly() {
+        val localhost = HttpRequestBuilder().apply {
+            url.takeFrom("http://localhost/")
+        }.defaultRequest {
+            baseURL("https://ktor.io/docs")
+        }.build().url
+        assertEquals("http://localhost/", localhost.toString())
+    }
+
+    @Test
+    fun overriddenLocalhostWithPath() {
+        val localhost = HttpRequestBuilder().apply {
+            url.takeFrom("http://localhost/sub")
+        }.defaultRequest {
+            baseURL("https://ktor.io/docs")
+        }.build().url
+        assertEquals("http://localhost/sub", localhost.toString())
+    }
+
+    @Test
+    fun useDefaultAsBaseURL() {
+        val localhost = HttpRequestBuilder().apply {
+            url.takeFrom("http://localhost/sub")
+        }.defaultRequest {
+            baseURL(URLBuilder())
+        }.build().url
+        assertEquals("http://localhost/sub", localhost.toString())
+    }
+
+    @Test
+    fun malformedBaseURL() {
+        assertFailsWith<IllegalArgumentException> {
+            HttpRequestBuilder().apply {
+                url.takeFrom("http://localhost/sub")
+            }.defaultRequest {
+                baseURL("https://ktor.io?a")
+            }
+        }
+    }
+}
--- a/ktor-client/ktor-client-core/api/ktor-client-core.api
+++ b/ktor-client/ktor-client-core/api/ktor-client-core.api
@@ -272,6 +272,12 @@ public final class io/ktor/client/plugins/DefaultRequest {
 	public fun <init> (Lkotlin/jvm/functions/Function1;)V
 }
 
+public final class io/ktor/client/plugins/DefaultRequest$Builder : io/ktor/client/request/HttpRequestBuilder {
+	public fun <init> (Lio/ktor/client/request/HttpRequestBuilder;)V
+	public final fun baseURL (Lio/ktor/http/URLBuilder;)V
+	public final fun baseURL (Ljava/lang/String;)V
+}
+
 public final class io/ktor/client/plugins/DefaultRequest$Plugin : io/ktor/client/plugins/HttpClientPlugin {
 	public fun getKey ()Lio/ktor/util/AttributeKey;
 	public fun install (Lio/ktor/client/plugins/DefaultRequest;Lio/ktor/client/HttpClient;)V
@@ -939,7 +945,7 @@ public final class io/ktor/client/request/HttpRequest$DefaultImpls {
 	public static fun getCoroutineContext (Lio/ktor/client/request/HttpRequest;)Lkotlin/coroutines/CoroutineContext;
 }
 
-public final class io/ktor/client/request/HttpRequestBuilder : io/ktor/http/HttpMessageBuilder {
+public class io/ktor/client/request/HttpRequestBuilder : io/ktor/http/HttpMessageBuilder {
 	public static final field Companion Lio/ktor/client/request/HttpRequestBuilder$Companion;
 	public fun <init> ()V
 	public final fun build ()Lio/ktor/client/request/HttpRequestData;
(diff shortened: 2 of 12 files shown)


ktor-client/ktor-client-core/common/src/io/ktor/client/features/DefaultRequest.kt
```diff
@@ -37,3 +37,49 @@ public fun HttpClientConfig<*>.defaultRequest(block: HttpRequestBuilder.() -> Un
...
+public fun HttpRequestBuilder.baseURL(baseUrlBuilder: URLBuilder) {
+    require(baseUrlBuilder.parameters.build() == Parameters.Empty && baseUrlBuilder.fragment.isEmpty()) {
+        "The baseURL cannot have a query or a fragment"
+    }
+
+    url {
+        val defaultURLBuilder = URLBuilder()
+        val isDefault = this == defaultURLBuilder
+        val onlyPathSet = startsWith(defaultURLBuilder)
+        if(!isDefault && onlyPathSet) {
```
cy6erGn0m, line 68: Why do we do nothing if `!onlyPathSet`?

ktor-client/ktor-client-core/common/test/io/ktor/client/features/cache/CacheExpiresTest.kt
```diff
@@ -1,7 +1,8 @@
 /*
  * Copyright 2014-2020 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
  */
-package io.ktor.client.features.cache.tests
```
cy6erGn0m, line 4: Why change package name here? Why does it make difference?

ktor-http/common/src/io/ktor/http/URLBuilder.kt
```diff
@@ -22,7 +22,7 @@ public const val DEFAULT_PORT: Int = 0
  * @property fragment URL fragment (anchor name)
  * @property trailingQuery keep a trailing question character even if there are no query parameters
  */
-public class URLBuilder(
+public data class URLBuilder(
```
cy6erGn0m, line 25: This sounds good at first but notice that it makes it difficult to keep binary compatibility because of the `copy` function with a lot of parameters in the case when we need to add a property. Also, adding a property in the middle breaks components functions as they are just ordered by declaration. 
What was the intention of this? Is it because of missing `toString` or because of `clone` vs `copy`?

ktor-http/common/src/io/ktor/http/URLBuilder.kt
```diff
@@ -113,7 +113,8 @@ internal expect val URLBuilder.Companion.originHost: String?
 /**
  * Create a copy of this builder. Modifications in a copy is not reflected in the original instance and vise-versa.
  */
-public fun URLBuilder.clone(): URLBuilder = URLBuilder().takeFrom(this)
+@Deprecated("Use generated copy function instead", ReplaceWith("copy()"))
+public fun URLBuilder.clone(): URLBuilder = copy()
```
cy6erGn0m, line 117: If we decide to migrate to `copy()`, we need to ensure that it does exactly the same as `takeFrom()` and fix all usages and tests.

ktor-client/ktor-client-core/common/src/io/ktor/client/features/DefaultRequest.kt
```diff
@@ -37,3 +37,49 @@ public fun HttpClientConfig<*>.defaultRequest(block: HttpRequestBuilder.() -> Un
...
+            encodedPath += if (encodedPath.endsWith("/")) {
+                requestedPath
+            } else {
+                "/$requestedPath"
+            }
+        }
+    }
+}
+
+internal fun URLBuilder.startsWith(other: URLBuilder): Boolean =
```
cy6erGn0m, line 80: `startsWith` = `sameOrigin`?

ktor-client/ktor-client-core/common/src/io/ktor/client/features/DefaultRequest.kt
```diff
@@ -37,3 +37,49 @@ public fun HttpClientConfig<*>.defaultRequest(block: HttpRequestBuilder.() -> Un
...
+    protocol == other.protocol
+        && host == other.host
+        && port == other.port
```
cy6erGn0m, line 83: Note that port number could be "default" or specified explicitly so it may not match

cy6erGn0m (changes requested):

ktor-client/ktor-client-core/common/src/io/ktor/client/features/DefaultRequest.kt
hfhbd: Keep in mind, the default request feature is called after! the user customized the url.
If the user uses `a.com/api` as baseURL, and requests `/some`, the `url` in `HttpRequestBuilder` is `http://localhost/some`, because of the default values in `URLBuilder`. So you have to check, if the `url` is a relative URL. If the url is relative (`onlyPathSet`), the baseURL is added. If `(!onlyPathSet`) the url is "complete", has a host, so the baseURL feature is not used.

hfhbd: Maybe I should rename the variable to `isRelativeURL`

ktor-client/ktor-client-core/common/test/io/ktor/client/features/cache/CacheExpiresTest.kt
hfhbd: It does not make any differences. In some tests, you use the normal package name, which allows IDEA to switch from the class right to the test, and sometimes you don't use a package at all, or you use a `tests` package name. In the end, there should be ONE convention in ktor...

ktor-http/common/src/io/ktor/http/URLBuilder.kt
hfhbd: The intention is the problem with the default values. With the default values set inside the constructor and not inside the `build()` function, it is impossible to know, if the user wants to request the relative `/some` or the explicit url `localhost/some`, because both url strings creates the same url builder. This pr use generated `toEquals` function for the missing equality to compare two instances. Of course this is also possible with a custom `toEquals` function instead.

hfhbd:
The main problem comes up is the problem with the default values in `URLBuilder`. With the default values set inside the constructor and not inside the build() function, it is impossible to know, if the user wants to request the relative /some or the explicit url localhost/some, because both url strings creates the same url builder. See [KTOR-1345](https://youtrack.jetbrains.com/issue/KTOR-1345)

ktor-client/ktor-client-core/common/src/io/ktor/client/features/DefaultRequest.kt
hfhbd: Switched to authority, which handles this special case

ktor-client/ktor-client-core/common/src/io/ktor/client/features/DefaultRequest.kt
hfhbd: Yes, `startsWith` = `protocol` + `authority`

ktor-client/ktor-client-core/common/src/io/ktor/client/features/DefaultRequest.kt
```diff
@@ -37,3 +37,45 @@ public fun HttpClientConfig<*>.defaultRequest(block: HttpRequestBuilder.() -> Un
         block()
     }
 }
+
+/**
+ * Set the [baseUrl] to use it in sub-requests.
+ * To override the existing [baseURL] in a sub-request, the full [URL] is needed in the [HttpRequestBuilder].
```
cy6erGn0m, line 43: This function is also applicable to a single request so the question is: what will happen if one invokes it outside of call config? The other issue is that it's not clear what is a sub-request for one who doesn't know about the default request feature and out of scope. For sure this should be clarified or somehow fixed.. not sure

cy6erGn0m (changes requested):

ktor-client/ktor-client-core/common/src/io/ktor/client/features/DefaultRequest.kt
hfhbd: Yeah, ideally this call should be only available in DefaultRequest feature (hey Multi Receiver Extensions). Until this feature is available, I could subclass the HttpRequestBuilder class and add the baseURL function to the sub-class.

hfhbd: I subclassed HttpRequestBuilder and created a DefaultRequest.Builder having the baseURL feature

e5l:
Hey @hfhbd, thanks for the PR. These changes contain new API, could you rebase this branch on `1.5.0` and change the PR base?

ktor-client/ktor-client-core/common/test/io/ktor/client/DefaultRequestTest.kt
```diff
@@ -0,0 +1,67 @@
...
+ */
+
+package io.ktor.client
+
+import io.ktor.client.features.*
+import io.ktor.client.request.*
+import io.ktor.http.*
+import kotlin.test.*
+
+class DefaultRequestTest {
```
hfhbd, line 12: Add test for wrong baseURL with parameters

hfhbd:
Added last test ensuring the order of calling baseURL in defaultRequestBlock is not important

e5l (changes requested):
Hey @hfhbd, thanks for the PR. Could you rebase this branch on `1.6.0-eap` and switch base branch to `1.6.0-eap` eap as well? This PR has new API and should be added only in minor release

hfhbd:
@e5l rebased

hfhbd:
@e5l Don't say it, I fixed the test already...

hfhbd:
Currently the tests are failing due to timeouts.

e5l:
@rsinukov, could you take a look how it will work with new client API?

rsinukov:
@e5l I don't see any conflicts between new client API and this feature. Looks godd for me

PR #2215, opened 2020-11-24
KTOR-1358 Propagate parentCoroutineContext to handlers

rsinukov:
**Subsystem**
Server

**Motivation**
[Netty engine doesn't keep coroutine elements in coroutine context](https://youtrack.jetbrains.com/issue/KTOR-1358)

**Solution**
Use `parentCoroutineContext` as base context for handlers coroutines

### DIFF
--- a/ktor-server/ktor-server-servlet/jvm/src/io/ktor/server/servlet/KtorServlet.kt
+++ b/ktor-server/ktor-server-servlet/jvm/src/io/ktor/server/servlet/KtorServlet.kt
@@ -46,10 +46,17 @@ public abstract class KtorServlet : HttpServlet(), CoroutineScope {
      */
     protected abstract val upgrade: ServletUpgrade
 
-    override val coroutineContext: CoroutineContext =
+    /**
+     * Base context for call handling
+     */
+    protected abstract val parentCoroutineContext: CoroutineContext
+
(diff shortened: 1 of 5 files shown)


ktor-server/ktor-server-servlet/jvm/src/io/ktor/server/servlet/KtorServlet.kt
```diff
@@ -46,10 +46,17 @@ public abstract class KtorServlet : HttpServlet(), CoroutineScope {
      */
     protected abstract val upgrade: ServletUpgrade
 
-    override val coroutineContext: CoroutineContext =
+    /**
+     * Base context for call handling
+     */
+    protected abstract val parentCoroutineContext: CoroutineContext
```
e5l, line 52: It's a breaking change: a new protected field in a public abstract class

e5l (changes requested):

ktor-server/ktor-server-servlet/jvm/src/io/ktor/server/servlet/KtorServlet.kt
rsinukov: updated

e5l (approved):
LGTM

PR #2219, opened 2020-11-26
KTOR-1367 Handle failure in reading request body

rsinukov:
**Subsystem**
Client, engines

**Motivation**
[Handle failure in reading request body](https://youtrack.jetbrains.com/issue/KTOR-1367)


e5l (approved):
LGTM

PR #2221, opened 2020-11-27
KTOR-1349 rewrite Input.copyTo from writeSuspendSession to write

rsinukov:
**Subsystem**
Client

**Motivation**
[Client: NPE in FormDataContentKt -> Input.copyTo](https://youtrack.jetbrains.com/issue/KTOR-1349)

**Solution**
TBH I am not sure how to reproduce it, but the crash is happening [here](https://github.com/ktorio/ktor/pull/2221/files#diff-6667837f9857b90b173abc24a146f7cbfaf384d6181180c32d40384363099a12L151)
I rewrote this block to the newer approach to remove the possibility of NPE


e5l (approved):
This looks like an identical change, and it looks like we hiding the problem. But I'm ok to drop `suspend session` API usage, so LGTM

PR #2226, opened 2020-12-03
KTOR-348 Fix testErrorInWritingPropagates

cy6erGn0m:
**Subsystem**
ktor-client-java

**Motivation**
Test testErrorInWritingPropagates is failing in master

**Solution**
If the body channel is closed too quickly, a publisher quits with no signalling error so the whole engine freezes. The solution is to propagate the cause in this case instead of just quitting.

### DIFF
--- a/ktor-client/ktor-client-java/jvm/src/io/ktor/client/engine/java/JavaHttpRequestBodyPublisher.kt
+++ b/ktor-client/ktor-client-java/jvm/src/io/ktor/client/engine/java/JavaHttpRequestBodyPublisher.kt
@@ -97,6 +97,7 @@ internal class JavaHttpRequestBodyPublisher(
         private fun readData() {
             // It's possible to have another request for data come in after we've closed the channel.
             if (inputChannel.isClosedForRead) {
+                tryToSignalOnErrorFromChannel()
                 return
             }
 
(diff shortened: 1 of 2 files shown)


ktor-client/ktor-client-java/jvm/src/io/ktor/client/engine/java/JavaHttpRequestBodyPublisher.kt
```diff
@@ -151,6 +152,16 @@ internal class JavaHttpRequestBodyPublisher(
                 subscriber.onError(cause)
             }
         }
+
+        private fun tryToSignalOnErrorFromChannel() {
+            if (inputChannel is ByteWriteChannel) {
```
e5l, line 157: Please invert the condition and use early return

ktor-client/ktor-client-java/jvm/src/io/ktor/client/engine/java/JavaHttpRequestBodyPublisher.kt
```diff
@@ -151,6 +152,16 @@ internal class JavaHttpRequestBodyPublisher(
...
+                // what a shame closedCause is in WriteChannel only
+                // we need to add it in 2.0.0
```
e5l, line 159: redundant comment, please create an issue describing the problem

e5l (approved):

PR #2229, opened 2020-12-03
KTOR-1434 Remove always included native targets

hfhbd:
**Subsystem**
Gradle

**Motivation**
The native targets are always include, ignoring the variable. https://youtrack.jetbrains.com/issue/KTOR-1434


cy6erGn0m:
Thanks! Merged

PR #2230, opened 2020-12-04
KTOR-941 CIO TLSConfigBuilder JVM - Use only specific alias

hfhbd:
**Subsystem**
CIO - JVM

Fixing [KTOR-941](https://youtrack.jetbrains.com/issue/KTOR-941)

**Motivation**
This fix adds the option to use only a specific alias instead all alias for a given JVM Keystore.
This change is addictive only, no breaking change.
 
**Solution**
Added null-able default null alias which will be use if present. Otherwise, all aliases from a Keystore will be used as known behavior. 

https://youtrack.jetbrains.com/issue/KTOR-941
Please reopen the issue on youtrack, it was closed after 30 days...

### DIFF
--- a/ktor-network/ktor-network-tls/jvm/src/io/ktor/network/tls/TLSConfigBuilder.kt
+++ b/ktor-network/ktor-network-tls/jvm/src/io/ktor/network/tls/TLSConfigBuilder.kt
@@ -81,15 +81,17 @@ public fun TLSConfigBuilder.addCertificateChain(chain: Array<X509Certificate>, k
 /**
  * Add client certificates from [store] by using all certificates
  */
+@Suppress("unused") // Keep for binary compatibility
+@Deprecated("Binary compatibility", level = DeprecationLevel.HIDDEN)
 public fun TLSConfigBuilder.addKeyStore(store: KeyStore, password: CharArray) {
-    addKeyStore(store, password, null)
+    addKeyStore(store, password)
 }
 
 /**
  * Add client certificates from [store] by using the certificate with specific [alias]
  * or all certificates, if [alias] is null.
  */
-public fun TLSConfigBuilder.addKeyStore(store: KeyStore, password: CharArray, alias: String?) {
+public fun TLSConfigBuilder.addKeyStore(store: KeyStore, password: CharArray, alias: String? = null) {
     val keyManagerAlgorithm = KeyManagerFactory.getDefaultAlgorithm()!!
     val keyManagerFactory = KeyManagerFactory.getInstance(keyManagerAlgorithm)!!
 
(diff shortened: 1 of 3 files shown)


ktor-network/ktor-network-tls/jvm/src/io/ktor/network/tls/TLSConfigBuilder.kt
```diff
@@ -79,16 +79,17 @@ public fun TLSConfigBuilder.addCertificateChain(chain: Array<X509Certificate>, k
 }
 
 /**
- * Add client certificates from [store].
+ * Add client certificates from [store] by using the certificate with specific [alias]
+ * or all certificates, if [alias] is null.
  */
-public fun TLSConfigBuilder.addKeyStore(store: KeyStore, password: CharArray) {
+public fun TLSConfigBuilder.addKeyStore(store: KeyStore, password: CharArray, alias: String? = null) {
```
rsinukov, line 85: Please split this method in two with the new parameter and without it. It's required to keep the binary compatibility.

ktor-network/ktor-network-tls/jvm/test/io/ktor/network/tls/tests/TLSConfigBuilderTest.kt
```diff
@@ -0,0 +1,39 @@
...
+package io.ktor.network.tls.tests
+
+import io.ktor.network.tls.*
+import io.ktor.network.tls.certificates.*
+import io.ktor.network.tls.extensions.*
+import kotlin.test.*
+
+internal class TLSConfigBuilderTest {
+    @Test
+    fun specificAliasInKeyStore() {
```
rsinukov, line 14: Can you split this test into two? With fixed alias and without it.

rsinukov (changes requested):
Looks good, but a couple of minor things.

ktor-network/ktor-network-tls/jvm/test/io/ktor/network/tls/tests/TLSConfigBuilderTest.kt
hfhbd: Of course :D

ktor-network/ktor-network-tls/jvm/src/io/ktor/network/tls/TLSConfigBuilder.kt
hfhbd: Splitted

hfhbd: Currently this new function is addictive only. Do you want to deprecate and hide the "old" function too? Then the alias must have `null` by default, to ensure no source changes in usage

rsinukov: Both ways are ok for me. What matters is to not remove anything from public API (both source and binary) in non-major releases.

rsinukov (approved):

hfhbd:
I would prefer deprecate and hiding the old function

PR #2235, opened 2020-12-07
KTOR-1412 Move response validation to the receive pipeline

rsinukov:
**Subsystem**

**Motivation**
[KTOR-1412](https://youtrack.jetbrains.com/issue/KTOR-1412) Response validation happens only on call to `receive<T>()`

**Solution**
Move validation from `HttpResponsePipeline` to `HttpReceivePipeline`

### DIFF
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/HttpClient.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/HttpClient.kt
@@ -162,10 +162,7 @@ public class HttpClient(
 
             config += this
 
-            if (expectSuccess) {
-                // it's important to install validators last
-                config.addDefaultResponseValidation()
-            }
+            config.addDefaultResponseValidation()
 
             config.install(this@HttpClient)
         }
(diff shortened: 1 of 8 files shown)


ktor-client/ktor-client-core/common/src/io/ktor/client/HttpClient.kt
```diff
@@ -154,17 +154,17 @@ public class HttpClient(
...
-
             config.install(HttpSend)
 
             if (followRedirects) {
                 config.install(HttpRedirect)
             }
 
             config += this
+
+            // it's important to install validators last
```
e5l, line 165: Could we fix that by inserting a new phase? Please remove the comment

e5l (changes requested):

ktor-client/ktor-client-core/common/src/io/ktor/client/HttpClient.kt
rsinukov: No, we can't, because validation happens not in pipeline interceptor, but `HttpSend` interceptor.
Comment removed.

e5l (approved):

PR #2236, opened 2020-12-08
KTOR-1464 Use realm value in header if not otherwise specified by user

you-shimizu:
**Subsystem**
client-auth
 
**Motivation**
If no realm value is provided, the DigestAuthProvider returns false instead of applying the realm value present in the 401 response header. 
 
**Solution**
Use realm value in 401 header for authentication if no other realm value has been specified.


cy6erGn0m:
Rewritten since it's not even compilable

PR #2237, opened 2020-12-08
KTOR-1465 Fix socket self connect problem

arthur-zhang:
**Subsystem**
ktor-network-jvm

**Motivation**
TCP has a well known self-connect problem, which client can connect to the client itself without any program listen on the port. 

Reproduce Code :

```
fun testSelfConnect() {
    runBlocking {
        // Find a port that would be used as a local address.
        val port = getAvailablePort()

        val tcpSocketBuilder = aSocket(ActorSelectorManager(Dispatchers.IO)).tcp()
        // Try to connect to that address repeatedly.
        for (i in 0 until 100000) {
            try {
                val socket = tcpSocketBuilder.connect(InetSocketAddress("127.0.0.1", port))
                println("connect to self succeed: ${socket.localAddress} to ${socket.remoteAddress}")
                System.`in`.read()
                break
            } catch (ex: Exception) {
                // ignore
            }
        }
    }
}

// since new linux kernel version introduce a feature, new bind port number will always be odd number
// and connect port will always be even, so we find a random even port with while loop
// see https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=07f4c90062f8fc7c8c26f8f95324cbe8fa3145a5
private fun getAvailablePort(): Int {
    while (true) {
        val port = ServerSocket().apply {
            bind(InetSocketAddress("127.0.0.1", 0))
            close()
        }.localPort

        if (port % 2 == 0) {
            return port
        }

        try {
            // try bind the next even port
            ServerSocket().apply {
                bind(InetSocketAddress("127.0.0.1", port + 1))
                close()
            }
            return port + 1
        } catch (ex: Exception) {
            // ignore
        }
    }
}

```

Running the above code whill result in a tcp connection with src addr equal to dst addr and src port equal to dst port.

image

If this happens, client will behave abnormal because it connect to the wrong server，server will bind fail on this machine.

Other programming language also have this problem, eg golang: https://golang.org/issue/2690

Fix this problem is easy, just check if the local and remote socket pair  are equal.


ktor-network/jvm/src/io/ktor/network/sockets/SocketImpl.kt
```diff
@@ -48,4 +57,13 @@ internal class SocketImpl<out S : SocketChannel>(
     private fun wantConnect(state: Boolean = true) {
         interestOp(SelectInterest.CONNECT, state)
     }
+
+    private fun selfConnect(): Boolean {
+        val localHostAddress = (this.localAddress as? InetSocketAddress)?.address?.hostAddress ?: ""
+        val remoteHostAddress = (this.remoteAddress as? InetSocketAddress)?.address?.hostAddress ?: ""
```
Ololoshechkin, line 63: Is there a way for `localAddress` to be `null`?

ktor-network/jvm/test/io/ktor/network/sockets/tests/ClientSocketTest.kt
```diff
@@ -91,6 +92,53 @@ class ClientSocketTest {
...
+        runBlocking {
+            // Find a port that would be used as a local address.
+            val port = getAvailablePort()
+
+            val tcpSocketBuilder = aSocket(ActorSelectorManager(Dispatchers.IO)).tcp()
+            // Try to connect to that address repeatedly.
+            for (i in 0 until 100000) {
+                try {
+                    val socket = tcpSocketBuilder.connect(InetSocketAddress("127.0.0.1", port))
+                    assertTrue(false, "connect to self succeed: ${socket.localAddress} to ${socket.remoteAddress}")
```
Ololoshechkin, line 106: fail("connect to self succeed: ${socket.localAddress} to ${socket.remoteAddress}")

ktor-network/jvm/src/io/ktor/network/sockets/SocketImpl.kt
arthur-zhang: Yes, you are right, I will fix this

ktor-network/jvm/test/io/ktor/network/sockets/tests/ClientSocketTest.kt
arthur-zhang: ok, i will fix this

arthur-zhang:
The new code has been commited.

Ololoshechkin (approved):

PR #2238, opened 2020-12-08
KTOR-1460 Switch to org.w3c.fetch

hfhbd:
**Subsystem**
Client Core JS

**Motivation**
Currently, the JS client contains old generated definitions, because the Standard Library did not contain these definitions. Now, the Standard Library contains these definitions. [KTOR-1460](https://youtrack.jetbrains.com/issue/KTOR-1460)

**Solution**
Switch from old generated definitions to the Standard Library


cy6erGn0m:
@hfhbd  We definitely have some problem: a lot of errors `ReferenceError: Headers is not defined` on nodejs. Didn't investigated it yet

hfhbd:
Sorry!
@cy6erGn0m I did investigate and the `Headers` are defined in `org.w3c.fetch`, which are defined in browsers, but not on NodeJS. So I defined them using `external` and `node-fetch`, but interesting this external definitions only works with `js(IR)`, see https://github.com/hfhbd/NodeFetchKotlinJS and run `./gradlew run`, if you switch to `js` the run task will fail... 
I will create an issue on Youtrack.

hfhbd:
@cy6erGn0m Could you please rerun the TeamCity build? I added the definitions and on my MacBook, the full build with allTests(3563) runs fine, except 4 JVM StacktraceWalking tests...

rsinukov:
Hi @hfhbd,
Thanks for the contribution! Looks good, but can you rebase to the latest master, please?

hfhbd:
@rsinukov here you go.

rsinukov:
thanks @hfhbd 
In general, it looks good, but we can't merge it until the issue with the JS legacy compiler is fixed or JS IR becomes the default one.
Let's wait for the compiler team.

hfhbd:
> thanks @hfhbd
> In general, it looks good, but we can't merge it until the issue with the JS legacy compiler is fixed or JS IR becomes the default one.
> Let's wait for the compiler team.

@rsinukov which issue do you mean? https://youtrack.jetbrains.com/issue/KT-44238? I did implement a workaround for this issue. Otherwise, `Headers` on Node.JS would be still undefined and node.JS test would fail

rsinukov (approved):

rsinukov:
@hfhbd oh, sorry, I misread your previous comment. LGTM

ktor-client/ktor-client-core/js/src/io/ktor/client/fetch/LibDom.kt
```diff
@@ -1,392 +0,0 @@
...
- * Copyright 2014-2019 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
- */
-
-package io.ktor.client.fetch
-
-import kotlin.js.Promise
-
-// external fun fetch(input: Request, init: RequestInit? = definedExternally): Promise<Response>
-
-public external fun fetch(input: String, init: RequestInit? = definedExternally): Promise<Response>
```
rsinukov, line 11: We can't delete declarations now, only in 2.0.0. Can you keep it, but add deprecations? Then it's ok to go to the master branch.

rsinukov (changes requested):
Oh, sorry, one more small thing

hfhbd:
@rsinukov That was one of my initial questions in the ticket :D
"Will this update be a breaking change? All current definitions are public but external.
ApiDump did not changed the api file, but supports the binary validator external JS definitions?"
So, public external JS declarations are counting to the API? At the end, the Kotlin declarations are "only" wrapper declarations and the underlying JS functions are still available...

rsinukov:
Yes, public external counts, because people can reference them in their code.

hfhbd:
@rsinukov Added deprecations

rsinukov (approved):
LGTM

PR #2240, opened 2020-12-09
KTOR-1406 Throw ParseException on invalid header and handle it as 400

rsinukov:
**Subsystem**
Server

**Motivation**
https://youtrack.jetbrains.com/issue/KTOR-1406

### DIFF
--- a/ktor-http/common/src/io/ktor/http/parsing/ParseException.kt
+++ b/ktor-http/common/src/io/ktor/http/parsing/ParseException.kt
@@ -4,6 +4,12 @@
 
 package io.ktor.http.parsing
 
+/**
+ * Thrown to indicate that the application has attempted to convert
+ * a string to one of the specific types, but that the string does not
+ * have the appropriate format.
+ * Please check the message for more details on the failure.
+ * */
 public class ParseException(
     override val message: String,
     override val cause: Throwable? = null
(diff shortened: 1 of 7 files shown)


e5l:
There is a compile error on CI, please check

rsinukov:
@e5l thanks, updated.

e5l (approved):
The implementation LGTM.
This PR can affect Space, @cy6erGn0m wdyt?

ktor-http/common/src/io/ktor/http/parsing/ParseException.kt
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright 2014-2020 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.http.parsing
+
+public class ParseException(
```
cy6erGn0m, line 7: This should be either documented or internal. Also, please don't forget to run apiDump for this

cy6erGn0m (approved):

rsinukov:
@e5l new `ParseException` extends `IllegalArgumentExcpetion` for backwards compatibility. If users were handling it, it should keep the behavior. Otherwise, we will return the proper 400 status instead of 500.

ktor-http/common/src/io/ktor/http/parsing/ParseException.kt
rsinukov: thanks, done

PR #2244, opened 2020-12-10
KTOR-926 Introduce wrapper for sending cookies via HTTPRequestBuilder

Ololoshechkin:
**Subsystem**
Client, HttpRequestBuilder

**Motivation**
https://youtrack.jetbrains.com/issue/KTOR-926

**Solution**
Wrapper over `headers { ... }` extension that allows adding cookies via `Cookie` header that can be accessed on server.

### DIFF
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/request/HttpRequest.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/request/HttpRequest.kt
@@ -217,6 +217,39 @@ public class HttpResponseData constructor(
  */
 public fun HttpRequestBuilder.headers(block: HeadersBuilder.() -> Unit): HeadersBuilder = headers.apply(block)
 
+/**
+ * Executes a [block] that configures the [CookiesBuilder] associated to this request.
+ */
+public fun HttpRequestBuilder.cookies(block: CookiesBuilder.() -> Unit): HeadersBuilder =
+    CookiesBuilder(headers).apply(block).headersBuilder
+
+public class CookiesBuilder(internal val headersBuilder: HeadersBuilder) {
+    public fun append(
+        name: String,
+        value: String,
+        maxAge: Int = 0,
+        expires: GMTDate? = null,
+        domain: String? = null,
+        path: String? = null,
+        secure: Boolean = false,
+        httpOnly: Boolean = false,
+        extensions: Map<String, String?> = emptyMap()
+    ) {
+        val cookie = Cookie(
+            name = name,
+            value = value,
+            maxAge = maxAge,
+            expires = expires,
+            domain = domain,
+            path = path,
+            secure = secure,
+            httpOnly = httpOnly,
+            extensions = extensions
+        )
+        headersBuilder.append(HttpHeaders.Cookie, renderSetCookieHeader(cookie))
+    }
+}
+
 
 /**
  * Mutates [this] copying all the data from another [request] using it as base.
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/request/HttpRequest.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/request/HttpRequest.kt
@@ -217,39 +217,6 @@ public class HttpResponseData constructor(
  */
 public fun HttpRequestBuilder.headers(block: HeadersBuilder.() -> Unit): HeadersBuilder = headers.apply(block)
 
-/**
- * Executes a [block] that configures the [CookiesBuilder] associated to this request.
- */
-public fun HttpRequestBuilder.cookies(block: CookiesBuilder.() -> Unit): HeadersBuilder =
-    CookiesBuilder(headers).apply(block).headersBuilder
-
-public class CookiesBuilder(internal val headersBuilder: HeadersBuilder) {
-    public fun append(
-        name: String,
-        value: String,
-        maxAge: Int = 0,
-        expires: GMTDate? = null,
-        domain: String? = null,
-        path: String? = null,
-        secure: Boolean = false,
-        httpOnly: Boolean = false,
-        extensions: Map<String, String?> = emptyMap()
-    ) {
-        val cookie = Cookie(
-            name = name,
-            value = value,
-            maxAge = maxAge,
-            expires = expires,
-            domain = domain,
-            path = path,
-            secure = secure,
-            httpOnly = httpOnly,
-            extensions = extensions
-        )
-        headersBuilder.append(HttpHeaders.Cookie, renderSetCookieHeader(cookie))
-    }
-}
-
 
 /**
  * Mutates [this] copying all the data from another [request] using it as base.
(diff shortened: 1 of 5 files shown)


ktor-client/ktor-client-core/common/src/io/ktor/client/request/HttpRequest.kt
```diff
@@ -217,6 +217,39 @@ public class HttpResponseData constructor(
...
 public fun HttpRequestBuilder.headers(block: HeadersBuilder.() -> Unit): HeadersBuilder = headers.apply(block)
 
+/**
+ * Executes a [block] that configures the [CookiesBuilder] associated to this request.
+ */
+public fun HttpRequestBuilder.cookies(block: CookiesBuilder.() -> Unit): HeadersBuilder =
+    CookiesBuilder(headers).apply(block).headersBuilder
+
+public class CookiesBuilder(internal val headersBuilder: HeadersBuilder) {
+    public fun append(
```
e5l, line 227: Please add KDoc for all new public API

e5l (changes requested):
LGTM, please add the docs and check the API dump

ktor-client/ktor-client-core/common/src/io/ktor/client/request/HttpRequest.kt
```diff
@@ -217,6 +217,39 @@ public class HttpResponseData constructor(
  */
 public fun HttpRequestBuilder.headers(block: HeadersBuilder.() -> Unit): HeadersBuilder = headers.apply(block)
 
+/**
+ * Executes a [block] that configures the [CookiesBuilder] associated to this request.
+ */
+public fun HttpRequestBuilder.cookies(block: CookiesBuilder.() -> Unit): HeadersBuilder =
+    CookiesBuilder(headers).apply(block).headersBuilder
+
+public class CookiesBuilder(internal val headersBuilder: HeadersBuilder) {
```
Stexxe, line 226: Is this the right place for this class? Isn't it better to have it in `io.ktor.client.features.cookies` package

ktor-client/ktor-client-core/common/src/io/ktor/client/request/HttpRequest.kt
Stexxe, line 227: The signature of this method is exactly duplicated by the class `Cookie`

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/CookiesTest.kt
```diff
@@ -200,5 +200,17 @@ class CookiesTest : ClientLoader() {
         }
     }
 
+    @Test
+    fun testRequestBuilderSingleCookie() = clientTests(listOf("Js")) {
+        test { client ->
+            val result = client.get<String>("$TEST_HOST/respond-single-cookie") {
+                cookies {
```
Stexxe, line 207: To add some headers to a request one can use the `header` function. Maybe we should stick to the same convention here and use `cookie("single", value = "abacaba")`.

e5l: It's a good point!

ktor-client/ktor-client-core/common/src/io/ktor/client/request/HttpRequest.kt
e5l: I would keep it in the same package but extract in the separate file

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/CookiesTest.kt
Ololoshechkin: But on the contrary, users don't have to know that cookies are implemented via headers, and why would they know that if they are about to send cookies they have to use `headers  {...}` function? 
Is that intuitive?
And btw, on the server side we have : `context.request.header(NAME)` and `context.request.cookies[NAME]` that are separated even though cookies use Cookie header inside. So I would separate these 2 entities on a client for consistency as well.
@e5l What do you think?

ktor-client/ktor-client-core/common/src/io/ktor/client/request/HttpRequest.kt
Ololoshechkin: Yes that's correct, so what would are you suggesting?
Inside a `cookies { ... }` block we can also allow syntax like:
```
cookies {
     +Cookie(...constructor params...)
}
```
Or
```
cookies {
     append(Cookie(...constructor params...))
}
```
Does that sound better to you?

From my perspective, `+Cookie(...)` version is easy to write and handy, but not that discoverable.
While `append(Cookie(...))` version is discoverable but too verbose and visually redundant given that we are already inside a `cookies { ... }` block (and obviously what we are appending inside -- is cookies, so another word Cookie is redundant is not that good for readability).

Ololoshechkin: Sure, will add, thanks!

ktor-client/ktor-client-core/common/src/io/ktor/client/request/HttpRequest.kt
Ololoshechkin: Good point, thanks, will move.

ktor-client/ktor-client-core/common/src/io/ktor/client/request/HttpRequest.kt
e5l: I don't like the syntax with `+Cookie(...)` cause it looks unfamiliar to Ktor users. 

I would keep `HttpRequestBuilder.cookie(...)` method

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/CookiesTest.kt
e5l: The server-side request is not a builder. Let's make it similar to the request builder headers

ktor-client/ktor-client-core/common/src/io/ktor/client/request/HttpRequest.kt
Ololoshechkin: Will add, thanks

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/CookiesTest.kt
Stexxe: I didn't say we need to put the `cookie` function inside the `headers` block.

e5l (approved):

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/CookiesTest.kt
Ololoshechkin: Done

PR #2249, opened 2020-12-11
KTOR-1370 Fix CORS simple content types processing

cy6erGn0m:
**Subsystem**
ktor-server-core (CORS)

**Motivation**
[KTOR-1370 CORS denies simple request where the content type includes a character encoding](https://youtrack.jetbrains.com/issue/KTOR-1370) 

**Solution**
Discard content type parameters and use only type and subtype as specified as per the fetch spec.


e5l (approved):
lgtm, please check API dump

cy6erGn0m:
Merged

PR #2252, opened 2020-12-14
KTOR-878 Strengthen session ID generation

cy6erGn0m:
**Subsystem**
ktor-server-core

**Motivation and solution**
[KTOR-878 Possibility of birthday attach for SessionStorage key](https://youtrack.jetbrains.com/issue/KTOR-878)
Currently, the session-id strength is not that high. In spite of that, we have no particular evidence of that it's really unsafe, our competitors usually have longer identifiers. So I decided to twice the id in ktor as well.

Also, the default session-id generation function is now public.

### DIFF
--- a/ktor-server/ktor-server-core/jvm/src/io/ktor/sessions/SessionIdProvider.kt
+++ b/ktor-server/ktor-server-core/jvm/src/io/ktor/sessions/SessionIdProvider.kt
@@ -1,0 +1,12 @@
+/*
+ * Copyright 2014-2020 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.sessions
+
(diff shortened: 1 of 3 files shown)


ktor-server/ktor-server-core/jvm/src/io/ktor/sessions/SessionIdProvider.kt
```diff
@@ -0,0 +1,12 @@
...
+ */
+
+package io.ktor.sessions
+
+import io.ktor.util.*
+
+/**
+ * Generates a secure random session ID
+ */
+public fun generateSessionId(): String = generateNonce() + generateNonce()
```
e5l, line 12: so strong :)

e5l (approved):
lgtm

PR #2253, opened 2020-12-14
KTOR-368 Introduce Netty tcpKeepAlive flag

cy6erGn0m:
**Subsystem**
ktor-server-netty

**Motivation**
[KTOR-368 ktor not closing user connections](https://youtrack.jetbrains.com/issue/KTOR-368) 
Unlike other backends, the netty backend doesn't have either read timeout, IDLE timeout or TCP keep alive option. So dead clients that don't close TCP connections properly, may get stuck forever. We only have read timeout that is disabled by default and still doesn't solve the problem because it's difficult to find the best time value for it as it may break WebSockets, SSE and long polling, including old techniques like "comet".

**Solution**
Introduce a way to enable TCP keep alive for Netty via option. It is disabled by default.

**Problems**
I see no way to write a reliable and fast enough test for this. So I decided to enable that flag for one of our server test suites for netty so we can ensure at least that setting this option doesn't lead to a crash.

### DIFF
--- a/ktor-server/ktor-server-netty/api/ktor-server-netty.api
+++ b/ktor-server/ktor-server-netty/api/ktor-server-netty.api
@@ -90,6 +90,7 @@ public final class io/ktor/server/netty/NettyApplicationEngine$Configuration : i
 	public final fun getResponseWriteTimeoutSeconds ()I
 	public final fun getRunningLimit ()I
 	public final fun getShareWorkGroup ()Z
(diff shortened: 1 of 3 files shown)


e5l (approved):
lgtm, please make a PR to docs or mark the issue with `docs needed`

cy6erGn0m:
Created [KTOR-1531](https://youtrack.jetbrains.com/issue/KTOR-1531) since we have no documentation for server engines at all.

PR #2256, opened 2020-12-16
KTOR-1458 Add tunneling functionality to CIO

rsinukov:
**Subsystem**
Client

**Motivation**
[KTOR-1458](https://youtrack.jetbrains.com/issue/KTOR-1458) Support SSL over HTTP proxy for CIO

Since we need to use raw input of Socket before starting TLS handshake, we can't open it for reading inside TLS methods. That's why the new class `Connection` was introduced. It holds the socket and its input and output.

### DIFF
--- a/ktor-client/ktor-client-cio/common/src/io/ktor/client/engine/cio/Endpoint.kt
+++ b/ktor-client/ktor-client-cio/common/src/io/ktor/client/engine/cio/Endpoint.kt
@@ -4,15 +4,21 @@
 
 package io.ktor.client.engine.cio
 
+import io.ktor.client.engine.*
 import io.ktor.client.features.*
 import io.ktor.client.request.*
+import io.ktor.http.*
+import io.ktor.http.cio.*
+import io.ktor.http.content.*
 import io.ktor.network.sockets.*
+import io.ktor.network.sockets.Connection
 import io.ktor.network.tls.*
 import io.ktor.util.*
 import io.ktor.util.date.*
 import io.ktor.util.network.*
 import io.ktor.utils.io.*
 import io.ktor.utils.io.core.*
+import io.ktor.utils.io.errors.*
 import kotlinx.atomicfu.*
 import kotlinx.coroutines.*
 import kotlinx.coroutines.channels.*
@@ -22,7 +28,7 @@ import kotlin.coroutines.*
 internal class Endpoint(
     private val host: String,
     private val port: Int,
-    private val overProxy: Boolean,
+    private val proxy: ProxyConfig?,
     private val secure: Boolean,
     private val config: CIOEngineConfig,
     private val connectionFactory: ConnectionFactory,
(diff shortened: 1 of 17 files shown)


e5l (approved):
lgtm

.idea/codeStyles/Project.xml
```diff
@@ -25,4 +25,4 @@
       <option name="CODE_STYLE_DEFAULTS" value="KOTLIN_OFFICIAL" />
     </codeStyleSettings>
   </code_scheme>
-</component>
\ No newline at end of file
+</component>
```
cy6erGn0m, line 28: ?

ktor-client/ktor-client-cio/common/src/io/ktor/client/engine/cio/Endpoint.kt
```diff
@@ -178,16 +184,23 @@ internal class Endpoint(
...
 
-                if (!secure) return@connect connection
+                val connection = socket.connection()
+                if (!secure) {
+                    return@connect connection
+                }
 
                 try {
-                    return connection.tls(coroutineContext) {
+                    if (proxy?.type == ProxyType.HTTP && requestData.url.protocol == URLProtocol.HTTPS) {
```
cy6erGn0m, line 193: What about secure WebSocket `wss://`? Is it already erased to HTTPS at this point? Shouldn't we use the `secure` flag here? Also, isn't it always the case at this point, as we are doing an early return a few lines above?

ktor-client/ktor-client-cio/common/src/io/ktor/client/engine/cio/utils.kt
```diff
@@ -138,6 +138,55 @@ internal suspend fun readResponse(
...
+        val host = if (request.url.protocol.defaultPort == request.url.port) {
+            request.url.host
+        } else {
+            request.url.hostWithPort
+        }
+
+        builder.headerLine(HttpHeaders.Host, host)
+        builder.headerLine("Proxy-Connection", "Keep-Alive") // For HTTP/1.0 proxies like Squid.
+        request.headers[HttpHeaders.UserAgent]?.let {
+            builder.headerLine(HttpHeaders.UserAgent, it)
```
cy6erGn0m, line 162: I wonder, do we need to pass it? It could be considered as a security issue as it deanonymizes user. Sorry for the old link, can't quickly find the actual document: https://www.eff.org/deeplinks/2010/01/tracking-by-user-agent

.idea/codeStyles/Project.xml
rsinukov: thanks, removed

ktor-client/ktor-client-cio/common/src/io/ktor/client/engine/cio/Endpoint.kt
rsinukov: updated

PR #2257, opened 2020-12-17
KTOR-572 Retry on HttpCode or network error

cy6erGn0m:
**Subsystem**
ktor-client-retry

**Motivation**
[KTOR-572 Retry on HttpCode or network error](https://youtrack.jetbrains.com/issue/KTOR-572)
[KTOR-945 Apache: ConnectionClosedException when requesting endpoint with the same client two times with some delay between](https://youtrack.jetbrains.com/issue/KTOR-945)

Currently, when the client fails to connect or a pooled connection is dead for some reason, the request fails. Also, sometimes, an error status code could be recovered by retrying the request.

**Solution**
Provide an optional feature that does retry on the particular list of exceptions to recover from I/O errors and request validation errors.


ktor-client/ktor-client-tests/jvm/src/io/ktor/client/tests/utils/ClientLoaderJvm.kt
```diff
@@ -24,7 +24,7 @@ public actual abstract class ClientLoader {
     public lateinit var engine: HttpClientEngineContainer
 
     @get:Rule
-    public open val timeout: CoroutinesTimeout = CoroutinesTimeout.seconds(60)
+    public open val timeout: CoroutinesTimeout = CoroutinesTimeout.seconds(5)
```
rsinukov: is it intentional?

ktor-client/ktor-client-features/ktor-client-retry/common/src/io/ktor/client/fatures/retry/Exceptions.kt
```diff
@@ -0,0 +1,45 @@
...
+        collectedErrors.forEach { subCause ->
+            cause.addSuppressed(subCause)
+        }
+    }
+
+    throw cause
+}
+
+
+internal fun Result<*>.checkExceptionOrThrowIfNotWhitelisted(): Throwable? {
```
rsinukov, line 30: whitelist -> allowed or something similar

ktor-client/ktor-client-features/ktor-client-retry/common/src/io/ktor/client/fatures/retry/Retry.kt
```diff
@@ -0,0 +1,163 @@
...
+ * The number of retries is limited by [retries] property.
+ * After every failed attempt it does a fixed time delay of [retryIntervalInSeconds].
+ *
+ * When all attempts exceeded, [RequestRetriesExceededException] exception is thrown
+ * having a list of all errors in the suppressed list and/or a cause.
+ *
+ * @property retries number of retries to recover
+ * @property retryIntervalInSeconds between attempts
+ */
+public class Retry private constructor(
```
rsinukov, line 34: A couple of design questions: 
* Should the list of allowed exceptions be configurable?
* Can we enable/disable retries per request? Can be a flag in request builder that saves some flag in the request attributes

rsinukov (changes requested):

ktor-client/ktor-client-tests/jvm/src/io/ktor/client/tests/utils/ClientLoaderJvm.kt
cy6erGn0m: Sorry, accidental change, reverted

ktor-client/ktor-client-features/ktor-client-retry/common/src/io/ktor/client/fatures/retry/Retry.kt
cy6erGn0m: Don't know how to configure it as people usually need subtypes that is not possible with KClass while lambdas have problems with freezes. Any ideas?
Regarding the flag: any ideas, what use-cases? How it should look like? Only inside of the request builder block, right?

ktor-client/ktor-client-features/ktor-client-retry/common/src/io/ktor/client/fatures/retry/Exceptions.kt
cy6erGn0m: IfNotRecoverable ?

ktor-client/ktor-client-features/ktor-client-retry/api/ktor-client-retry.api
```diff
@@ -0,0 +1,38 @@
+public final class io/ktor/client/fatures/retry/Retry {
```
e5l, line 1: package name typo

ktor-client/ktor-client-features/ktor-client-retry/common/src/io/ktor/client/fatures/retry/Exceptions.kt
e5l: I would also avoid extension on `Result` type

ktor-client/ktor-client-features/ktor-client-retry/common/src/io/ktor/client/fatures/retry/Exceptions.kt
```diff
@@ -0,0 +1,45 @@
...
+        is IOException,
+        is HttpRequestTimeoutException,
+        is ResponseException,
+        is UnresolvedAddressException -> cause
+        else -> throw cause
+    }
+}
+
+internal fun Throwable.unwrapCancellation(): Throwable = when (this) {
+    is CancellationException -> cause?.unwrapCancellation() ?: this
```
e5l, line 43: Can we get an infinite loop here?

e5l (changes requested):
I would delay this feature and make the design first

cy6erGn0m:
Closing it for now. Will reopen later with 1.6.0

PR #2264, opened 2020-12-20
Add shortcut function to configure the json instance

bluefireoly:
**Subsystem**
ktor-features/ktor-serialization

**Motivation**
While with gson and jackson together with ktor you can do the following to configure the instances:
```kt
install(ContentNegotiation) {
    gson {
        // Configure Gson here
    }
}
```
... with kotlinx.serialization you have to do the following:
```kt
install(ContentNegotiation) {
    json(Json {
        // Configure kotlinx.serialization here
    })
}
```
... which most new users do not know.

**Solution**
With the new function I added, this is not the case anymore:
```kt
install(ContentNegotiation) {
    json {
        // Configure kotlinx.serialization here
    }
}
```

In my opinion, this is much more intuitive to use.

### DIFF
--- a/ktor-features/ktor-serialization/jvm/src/io/ktor/serialization/JsonSupport.kt
+++ b/ktor-features/ktor-serialization/jvm/src/io/ktor/serialization/JsonSupport.kt
@@ -36,6 +36,19 @@ public val DefaultJsonConfiguration: Json = Json {
 }
 
 /**
+ * Applies the default json configuration for Ktor to this [JsonBuilder].
+ * @see DefaultJson
+ */
+public fun JsonBuilder.setKtorDefaults() {
+    encodeDefaults = true
+    isLenient = true
+    allowSpecialFloatingPointValues = true
+    allowStructuredMapKeys = true
+    prettyPrint = false
+    useArrayPolymorphism = true
+}
+
+/**
  * The default json configuration used in [SerializationConverter]. The settings are:
  * - defaults are serialized
  * - mode is not strict so extra json fields are ignored
@@ -46,12 +59,7 @@ public val DefaultJsonConfiguration: Json = Json {
  * See [Json] for more details.
  */
 public val DefaultJson: Json = Json {
-    encodeDefaults = true
-    isLenient = true
-    allowSpecialFloatingPointValues = true
-    allowStructuredMapKeys = true
-    prettyPrint = false
-    useArrayPolymorphism = true
+    setKtorDefaults()
 }
 
 /**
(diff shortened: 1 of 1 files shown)


hhariri:
Thanks for the PR. There are a few issues here, namely two: 

1. gson/jackson is referring to library names, whereas json is referring to a format. 
2. Kotlinx serialisation provides more than just JSON and this could be confusing if we were to use this function name. 

It does make sense to provide some defaults and remove boilerplate code. Do you have a suggestion for another name for this?

hhariri (changes requested):
Thanks for the PR. There are a few issues here, namely two:

1. gson/jackson is referring to library names, whereas json is referring to a format.
2. Kotlinx serialisation provides more than just JSON and this could be confusing if we were to use this function name.

It does make sense to provide some defaults and remove boilerplate code. Do you have a suggestion for another name for this?

bluefireoly:
I was not the one who named this function json, Ktor already has a function called `json`, which just did not provide a lamda for configuring the instance.
Also, previously the `json` function for configuring kotlinx.serialization was called `serialization`, I would not change this again.

However if you really want to change the name, then I suggest `kotlinxSerialization` - that would follow the scheme of the other function names.

bluefireoly:
Here is a [link to the function](https://github.com/ktorio/ktor/blob/b2feeccbc5b9d4202673cf97ca6ae6f25d0d10b6/ktor-features/ktor-serialization/jvm/src/io/ktor/serialization/JsonSupport.kt#L88) which already exists in Ktor. That is why I named my additional function `json` aswell.

Additionally, I think that the name `json` is indeed a good name, because users should be encouraged to use kotlinx.serialization:
- it is multiplatform
- by jetbrains
- an official kotlinx library

hhariri:
@bluefireoly Yeah, it's not ideal we have that there either. But the issue is Serialization isn't limited to json. And yet this is somewhat implying that. Does that make sense?

bluefireoly:
I understand your point, but what this function does is creating a `Json` instance of kotlinx.serialization - and that instance is limited to json.

hhariri:
Very much appreciate the contribution, but as it stands we need to rethink this. I've created a design ticket to further discuss how we can do this. https://youtrack.jetbrains.com/issue/KT-46046

bluefireoly:
I really support the design ticket you created there and I am aswell of the opinion that this must be rethought, BUT would it maybe be possible to merge this PR anyways, just to make the current API more consistent UNTIL that major serialization API change is coming?

hhariri:
The issue with merging this is that any new API we provide we'll need to then move to a deprecation cycle...

bluefireoly:
So you think the lifespan of this feature would be too short?

hhariri:
Yes. The design ticket should be approved soon, and we could try and implement the functionality rather soon.

PR #2267, opened 2020-12-21
KTOR-969 Update changelog

cy6erGn0m:


hfhbd:
@cy6erGn0m How do you generate the changelog and why it misses at least KTOR-941? :D

cy6erGn0m:
@hfhbd it is from YT, the ticker you are referring is stuck for some reason and doesn't have the correct target version :(

rsinukov:
thanks @hfhbd! updated it here #2269

PR #2268, opened 2020-12-21
KTOR-1543 Fix URL path encoding using incorrect encoding function

### LINKED ISSUE #1293: Reserved characters in path is not encoded
Ktor client version: `1.2.3`

```
URLBuilder(host = "localhost").apply {
    path("?")
    parameters["?"] = "?"
}
    .buildString()
    .also(::println)
```

will produce:

```
http://localhost/??%3F=%3F
```

I would expect the following url:

```
http://localhost/%3F?%3F=%3F
```

`URLBuilder` uses `encodeURLQueryComponent` function to encode a path: https://github.com/ktorio/ktor/blob/1.2.3/ktor-http/common/src/io/ktor/http/URLBuilder.kt#L50
Possibly it should use `encodeURLPath`?

Not sure it is a bug, but a feature. Fixing it will break backward compatibility.

amayers:
KTOR-1543
Fixes #1293

**Subsystem**
`URLBuilder`

**Motivation**
The `URLBuilder` was using the wrong url encoding method for the path, which wasn't encoding some characters as they should have been.

**Solution**
Switch the encoder function call from `path.encodeURLQueryComponent()` to `path.encodeURLPath()`.

### DIFF
--- a/ktor-http/common/src/io/ktor/http/URLBuilder.kt
+++ b/ktor-http/common/src/io/ktor/http/URLBuilder.kt
@@ -54,7 +54,7 @@ public class URLBuilder(
      * Encode [components] to [encodedPath]
      */
     public fun path(components: List<String>): URLBuilder {
-        encodedPath = components.joinToString("/", prefix = "/") { it.encodeURLQueryComponent() }
+        encodedPath = components.joinToString("/", prefix = "/") { it.encodeURLPath() }
 
         return this
     }
--- a/ktor-http/common/test/io/ktor/tests/http/URLBuilderTest.kt
+++ b/ktor-http/common/test/io/ktor/tests/http/URLBuilderTest.kt
@@ -180,6 +180,17 @@ internal class URLBuilderTest {
         assertEquals("/path/%F0%9F%90%95", url.encodedPath)
     }
 
+    @Test
+    fun testPathEncoding() {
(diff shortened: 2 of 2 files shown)


amayers:
I do question if this encoding method is doing enough. It seems to me that either this function should encode all special characters, or none. For example the `+` is not encoded by this function. What if the caller wanted to actually send a `+`? They can't use this, because the server will treat it like a space. But if the caller tries to encode prior to using `path(...)` then it will end up with double encoding. So to be safe callers should avoid all use of the `path(...)` function and do all encoding manually.

cy6erGn0m:
Probably, seever shouldn't treat + as space. Isn't it form url encoded only ?

e5l:
Hey, @amayers. Thanks for the PR. Will check after CI run

amayers:
> Probably, seever shouldn't treat + as space. Isn't it form url encoded only ?

I know it is in the path, and not the query. But if you look at virtually any search engine search url, `+` is treated as a space, unless percent encoded. So it doesn't seem uncommon.

e5l (approved):
LGTM, @cy6erGn0m could you also take a look?

cy6erGn0m (approved):
We defeinitely need this fix. The only remaining issue is that "/" is still not escaped while I personally would expect it to be escaped

Issue #1256, opened 2019-08-01
ResponseException is not Serializable


Thomas-Vos:
**Subsystem**
Client, core. I think this is only relevant for the JVM.

**Is your feature request related to a problem? Please describe.**
`ResponseException` is not serializable, it throws a `NotSerializableException`.

**Describe the solution you'd like**
Please make `ResponseException` serializable. This can be done by making the response field transient.

**Motivation to include to ktor**
What I am doing is showing the user an appropriate error message depending on the type of exception. To do this, I serialize the exception so I can pass it to a different Android Fragment using a Bundle. That Android Fragment “converts” the exception to a localised error message. Maybe I should change it so I first “convert” the exception to a localized message, and put that message in the Bundle instead. However, that would mean I would need to add that to every place in my app where I use Ktor / other possible exceptions. It was way easier to pass the exception to a single place and convert it there.

See discussion on Slack: https://kotlinlang.slack.com/archives/C0A974TJ9/p1564513134083200

Thomas-Vos:
Unfortunately this broke again. See new issue [KTOR-1552](https://youtrack.jetbrains.com/issue/KTOR-1552) and PR https://github.com/ktorio/ktor/pull/2270.

PR #2270, opened 2020-12-22
KTOR-1552 Mark _response transient in ResponseException

Thomas-Vos:
**Subsystem**
ktor-client-core, JVM

**Motivation**
https://youtrack.jetbrains.com/issue/KTOR-1552
https://github.com/ktorio/ktor/issues/1256

**Solution**
Mark _response transient in ResponseException.

Note: accessing the response field of the deserialized exception will throw a NullPointerException. That is the same behaviour as it was before. This cannot be improved unless the response field is made nullable.

### DIFF
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/features/DefaultResponseValidation.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/features/DefaultResponseValidation.kt
@@ -9,6 +9,7 @@ import io.ktor.client.call.*
 import io.ktor.client.statement.*
 import io.ktor.util.*
 import io.ktor.utils.io.concurrent.*
+import kotlin.jvm.*
 import kotlin.native.concurrent.*
 
 @SharedImmutable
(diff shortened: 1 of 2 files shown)


ktor-client/ktor-client-core/common/src/io/ktor/client/features/DefaultResponseValidation.kt
```diff
@@ -57,7 +58,7 @@ public open class ResponseException(
     @Deprecated(level = DeprecationLevel.WARNING, message = DEPRECATED_EXCEPTION_CTOR)
     public constructor(response: HttpResponse): this(response, NO_RESPONSE_TEXT)
 
-    private val _response: HttpResponse? by threadLocal(response)
+    @delegate:Transient private val _response: HttpResponse? by threadLocal(response)
```
cy6erGn0m, line 61: I would split it into two lines

Ololoshechkin (approved):

cy6erGn0m (approved):

PR #2271, opened 2020-12-23
KTOR-1220 Fix terminating curl processor worker on close

cy6erGn0m:
**Subsystem**
ktor-client-curl

**Motivation**
[KTOR-1220 "Unfinished workers detected" using client on native](https://youtrack.jetbrains.com/issue/KTOR-1220)

**Solution**
Terminate worker on client close

**Known issues**
We can't force memory checker to run in tests, so no tests provided


e5l (approved):
LGTM. 
Small note here: we still have a native `Future` leak here and can't totally avoid it. It would be nice to have this class rewritten using coroutines. We had no `native-mt` at the moment of creating it and now it should be possible

PR #2272, opened 2020-12-23
Bump kotlin libs versions

cy6erGn0m:


cy6erGn0m:
Ok, looks like we need to wait for atomicfu 0.15.1 otherwise we can't upgrade even kotlin

Thomas-Vos:
Why do you need to update atomicfu? I am using atomicfu 0.14.4 with Kotlin 1.4.21 on Native in my project which works fine.

cy6erGn0m:
@Thomas-Vos because of JVM/JS IR backends and incompatible binary IR libs formats.

PR #2273, opened 2020-12-23
KTOR-1567 Fix freezing Java client

cy6erGn0m:
**Subsystem**
ktor-client-java

**Motivation**
[KTOR-1567 Java client freeze](https://youtrack.jetbrains.com/issue/KTOR-1567)

**Solution**
Stop doing runBlocking in a reactive handler. Launch a separate coroutine to push bytes from reactive stream to byte channel


e5l (approved):
I like an idea about removing `runBlocking`, so LGTM. Probably we can try to simplify it by removing the queue somehow.

PR #2274, opened 2020-12-24
KTOR-1572 Fix flaky ClassCastExceptionTest

cy6erGn0m:
**Subsystem**
ktor-server-cio/tests

**Motivation**
From time to time the test is failing because of the address already in use error.

**Solution**
Use server tests infrastructure that is almost tolerant for such kind of problems.


e5l (approved):
lgtm

e5l:
Could it also be related to `address already in used`? Maybe we have similar tests to fix

PR #2275, opened 2020-12-24
KTOR-874 Fix jetty client error handling

cy6erGn0m:
**Subsystem**
ktor-client-jetty

**Motivation**
[KTOR-874 Jetty: requests to resources, that doesn't respond with HTTP/2, lead to unexpected behaviour](https://youtrack.jetbrains.com/issue/KTOR-874)

**Solution**
Add error handler and fail requests with more detailed errors rather than EOF at writing.


e5l (approved):
Looks nice!

PR #2278, opened 2020-12-28
KTOR-1583 Fix wrong connect timeout exception with OkHttp

cy6erGn0m:
**Subsystem**
ktor-client-okhttp

**Motivation**
[KTOR-1583 HttpTimeoutTest.testConnect are flaky](https://youtrack.jetbrains.com/issue/KTOR-1583)
Sometimes, a timeout exception is not get mapped to the correct exception type because the platform exception text may be different. In particular, on Mac it may contain "Connect" so the contains check may fail

**Solution**
- Split and simplify code
- use case-insensitive contains check instead


e5l (approved):
lgtm

PR #2281, opened 2020-12-29
Fix server receive exception handling

cy6erGn0m:
**Subsystem**
ktor-http, ktor-server-cio

**Motivation**
[KTOR-1590 Server losing channel exceptions at receive](https://youtrack.jetbrains.com/issue/KTOR-1590)

**Solution**
- Fix losing exception
- Fix exception handling: we should wrap channel errors to ChannelIOException for proper logging
- Prevent call handling in this case
- Fix test to not allow receiving any value in this case

### DIFF
--- a/ktor-http/ktor-http-cio/common/src/io/ktor/http/cio/HttpBody.kt
+++ b/ktor-http/ktor-http-cio/common/src/io/ktor/http/cio/HttpBody.kt
@@ -82,7 +82,7 @@ public suspend fun parseHttpBody(
             transferEncoding.equalsLowerCase(other = "identity") -> {
                 // do nothing special
             }
-            else -> out.close(IllegalStateException("Unsupported transfer-encoding $transferEncoding"))
+            else -> throw IllegalArgumentException("Unsupported transfer-encoding $transferEncoding")
         }
     }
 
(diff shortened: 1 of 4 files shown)


ktor-http/ktor-http-cio/common/src/io/ktor/http/cio/HttpBody.kt
```diff
@@ -77,12 +82,8 @@ public suspend fun parseHttpBody(
     out: ByteWriteChannel
 ) {
     if (transferEncoding != null) {
-        when {
-            transferEncoding.equalsLowerCase(other = "chunked") -> return decodeChunked(input, out)
-            transferEncoding.equalsLowerCase(other = "identity") -> {
-                // do nothing special
-            }
-            else -> out.close(IllegalStateException("Unsupported transfer-encoding $transferEncoding"))
+        if (isTransferEncodingChunked(transferEncoding)) {
```
e5l, line 85: &&

ktor-http/ktor-http-cio/common/src/io/ktor/http/cio/HttpBody.kt
```diff
@@ -121,3 +122,32 @@ public suspend fun parseHttpBody(
...
+                // ignore this token
+            }
+            else -> throw IllegalTransferEncodingException("Unsupported transfer encoding $name")
+        }
+    }
+
+    return chunked
+}
+
+private class IllegalTransferEncodingException(override val message: String) : IllegalArgumentException(message)
```
e5l, line 153: Could you tell me why do we need a new type of exception? Do we have the intention to catch it?

e5l (changes requested):

e5l (approved):

Issue #1882, opened 2020-05-20
Status-code must be 3-digit


zuevmaxim:
**Ktor Version and Engine Used (client or server and name)**
io.ktor:ktor-server-cio:1.3.2

**Describe the bug**
Status-code must be 3-digit(see https://tools.ietf.org/html/rfc2616#section-6.1.1)

**To Reproduce**
Run code:
```
import io.ktor.http.cio.parseResponse
import io.ktor.utils.io.ByteReadChannel
import kotlinx.coroutines.runBlocking


fun main() {
    listOf(
        """
        HTTP/1.1 0 OK


        """.trimIndent(),
        """
        HTTP/1.1 10 OK


        """.trimIndent(),
        """
        HTTP/1.1 1000 OK


        """.trimIndent(),
        """
        HTTP/1.1 10000 OK


        """.trimIndent()
    ).forEach { test(it) }

}

fun test(http: String) {
    val response = runBlocking { parseResponse(ByteReadChannel(http)) }!!
    println("status=${response.status};statusText={${response.statusText}};headers=[${response.headers}];version=${response.version}")
}

```

**Expected behavior**
An error expected - exception or null result.

oleg-larshin:
Please check the <a href="https://youtrack.jetbrains.com/issue/KTOR-370">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

vladf2n:
Hey, I'll work on this!

e5l:
Fixed in 1.5.1

PR #2283, opened 2020-12-30
KTOR-370 Fix status code response accepting any numeric value (#1882)

vladf2n:
**Subsystem**
ktor-http

**Motivation**
#1882 

**Solution**
Create a conditional to check if the status-code is a valid response (between 100 and 599)

### DIFF
--- a/ktor-http/ktor-http-cio/common/src/io/ktor/http/cio/HttpParser.kt
+++ b/ktor-http/ktor-http-cio/common/src/io/ktor/http/cio/HttpParser.kt
@@ -16,7 +16,7 @@ public class ParserException(message: String) : Exception(message)
 
 private const val HTTP_LINE_LIMIT = 8192
 private const val HTTP_STATUS_CODE_MIN_RANGE = 100
-private const val HTTP_STATUS_CODE_MAX_RANGE = 599
+private const val HTTP_STATUS_CODE_MAX_RANGE = 999
 
 /**
  * Parse an HTTP request line and headers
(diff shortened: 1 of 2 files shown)


ktor-http/ktor-http-cio/common/src/io/ktor/http/cio/HttpParser.kt
```diff
@@ -204,6 +210,7 @@ private fun parseStatusCode(text: CharSequence, range: MutableRange): Int {
     return status
 }
 
+private fun statusOutOfRange(code: Int) = code < HTTP_STATUS_CODE_MIN_RANGE || code > HTTP_STATUS_CODE_MAX_RANGE
```
cy6erGn0m, line 213: Status codes higher than 599 (e.g. 6xx) aren't allowed; however, some services may use it despite that. So I would relax this limitation and apply the [HTTP syntax limitation](https://tools.ietf.org/html/rfc7230#section-3.1.2) that is defined like this:
>      status-code    = 3DIGIT

cy6erGn0m (changes requested):
Thanks for your contribution. It's fine in general except for the small note regarding the upper status code limit.

ktor-http/ktor-http-cio/common/src/io/ktor/http/cio/HttpParser.kt
vladf2n: Thank you so much for your feedback! I'll fix it now.

vladf2n: Just made a new commit to fix this limitation.

cy6erGn0m (approved):

PR #2285, opened 2021-01-02
Fix CIO client tunneling https through http proxy

PegasisForever:
**Subsystem**
CIO client

**Motivation**
Fixes wrong http `CONNECT` request when tunneling https through a http proxy.
The request line should be `CONNECT $host:$port HTTP/1.1` (e.g. `CONNECT example.com:443 HTTP/1.1`)
Not `CONNECT $url HTTP/1.1` (e.g. `CONNECT https://example.com/ HTTP/1.1`)
According to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/CONNECT)

**Related PR**
#2256

### DIFF
--- a/ktor-client/ktor-client-cio/common/src/io/ktor/client/engine/cio/utils.kt
+++ b/ktor-client/ktor-client-cio/common/src/io/ktor/client/engine/cio/utils.kt
@@ -146,9 +146,7 @@ internal suspend fun startTunnel(
     val builder = RequestResponseBuilder()
 
     try {
-        val urlString = request.url.toString()
-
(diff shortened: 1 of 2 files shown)


rsinukov (changes requested):
Thanks for the contribution and fix!
Do you mind revert [this](https://github.com/ktorio/ktor/pull/2256/files#diff-18edf3fb889a9385335519e1e6676232f6c9bf580c927935a60f407267059fe8R85) change from the original PR as well?
Otherwise LGTM

PegasisForever:
Thanks, I reverted that change

rsinukov (approved):

rsinukov:
Thanks, I will merge it after it runs on CI

PR #2287, opened 2021-01-07
KTOR-1615 fix single slash routing

rsinukov:
**Subsystem**
Server, routing

**Motivation**
[A single slash gets ignored for defining a route](https://youtrack.jetbrains.com/issue/KTOR-1615)

**Solution**
Rework solution of handling trailing slash in routing. Introduce a new selector that matches only if the trailing slash is equal in request URI and route.


ktor-server/ktor-server-core/jvm/src/io/ktor/routing/RouteSelector.kt
```diff
@@ -225,43 +223,40 @@ public data class OptionalParameterRouteSelector(val name: String) :
  * Evaluates a route against a constant path segment
  * @param value is a value of the path segment
  */
-public data class PathSegmentConstantRouteSelector(val value: String) : RouteSelector(RouteSelectorEvaluation.qualityConstant) {
-
-    private var hasTrailingSlash: Boolean = false
+public data class PathSegmentConstantRouteSelector(val value: String) :
```
e5l, line 226: codestyle

e5l (approved):

ktor-server/ktor-server-core/jvm/src/io/ktor/routing/RouteSelector.kt
rsinukov: What is the correct codestyle here? It doesn't fit in one line. Will you prefer
```
public data class PathSegmentConstantRouteSelector(val value: String) : RouteSelector(
    RouteSelectorEvaluation.qualityConstant
) {
```

e5l: Yep, it's correct according: https://kotlinlang.org/docs/reference/coding-conventions.html#class-header-formatting

There is different option:
```kotlin
public data class PathSegmentConstantRouteSelector(
val value: String
) : RouteSelector(RouteSelectorEvaluation.qualityConstant) {
```

PR #2290, opened 2021-01-11
KTOR-426 Add test to check body logging in post request with json

e5l:
Added test to verify fixed behaviour

Fixes [KTOR-426](https://youtrack.jetbrains.com/issue/KTOR-426)


cy6erGn0m (approved):
The added test has failed once but it seems to be the irrelevant issue "response body omitted"

Issue #1865, opened 2020-05-14
Non US-ASCII characters are prohibited


zuevmaxim:
**Ktor Version and Engine Used (client or server and name)**
io.ktor:ktor-server-cio:1.3.2

**Describe the bug**
Non US-ASCII characters(for example "ё") are prohibited(see https://tools.ietf.org/html/rfc3986#appendix-A)

**To Reproduce**
Run code:
```
import io.ktor.http.cio.parseRequest
import io.ktor.utils.io.ByteReadChannel
import kotlinx.coroutines.runBlocking

fun main() {
    listOf(
        """
        POST /ё.txt HTTP/1.1
        Host: www.example.com


        """.trimIndent(),
        """
        POST / HTTP/1.1
        Host: www.example.ёё


        """.trimIndent()
    ).forEach {
        test(it)
    }

}

fun test(http: String) {
    val request = runBlocking { parseRequest(ByteReadChannel(http)) }!!
    println("method=${request.method};uri={${request.uri}};headers=[${request.headers}]")
}


```

**Expected behavior**
An error expected - exception or null result.

zuevmaxim:
The same problem with `parseResponce`
```
import io.ktor.http.cio.parseResponse
import io.ktor.utils.io.ByteReadChannel
import kotlinx.coroutines.runBlocking

fun main() {
    test(
        """
        HTTP/1.1 200 OK
        Vary: Authorization,Accept
        Transfer-Encoding: ёncoding


    """.trimIndent()
    )
}

fun test(http: String) {
    val response = runBlocking { parseResponse(ByteReadChannel(http)) }!!
    println("status=${response.status};statusText={${response.statusText}};headers=[${response.headers}];version=${response.version}")
}
```

oleg-larshin:
Please check the <a href="https://youtrack.jetbrains.com/issue/KTOR-381">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

vladf2n:
I'll work on this.

PR #2294, opened 2021-01-12
KTOR-381 Non US-ASCII characters are prohibited

vladf2n:
**Subsystem**
ktor-http

**Motivation**
[KTOR-381 Non US-ASCII characters are prohibited](https://youtrack.jetbrains.com/issue/KTOR-381)
#1865  

**Solution**
Check if every character from a request/response is US-ASCII.

### DIFF
--- a/ktor-http/ktor-http-cio/common/src/io/ktor/http/cio/HttpParser.kt
+++ b/ktor-http/ktor-http-cio/common/src/io/ktor/http/cio/HttpParser.kt
@@ -16,7 +16,7 @@ public class ParserException(message: String) : Exception(message)
 
 private const val HTTP_LINE_LIMIT = 8192
 private const val HTTP_STATUS_CODE_MIN_RANGE = 100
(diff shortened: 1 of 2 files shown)


cy6erGn0m:
As far as I remember some browsers produce requests having non-ascii characters in some headers. Don't remember where exactly, perhaps it about multipart file upload or something related... Let me check

vladf2n:
Thank you for your response. Ok I'm waiting!

cy6erGn0m:
Ok, looks like we call `parseHeaders` directly in the multipart parser. However, I am reading the RFCs to ensure that we don't miss anything as I am too scary to introduce such restrictions so easily :)

cy6erGn0m:
The other reason for doubts is that other servers allow non-ASCII characters

vladf2n:
Hmm... got it. So, how do you think I should proceed?

cy6erGn0m:
I am not sure. Wait please a little bit more. We'll discuss it

vladf2n:
Okay!

PR #2295, opened 2021-01-13
KTOR-1662 Fix CORS headers processing

cy6erGn0m:
**Subsystem**
ktor-server-core

**Motivation**
[KTOR-1662 CORS doesn't reject bad headers](https://youtrack.jetbrains.com/issue/KTOR-1662)
Caused by #2034 

**Solution**
Fix accidental early return


rsinukov (approved):
nice catch! inline and return can be confusing =(

PR #2305, opened 2021-01-15
~ Make CIO selector do wakeup on close/suspend

cy6erGn0m:
**Subsystem**
ktor-client-cio

**Motivation**
Currently, CIO selector does blocking select. Closing it or adding interests don't wakeup so no new interests will be selected and close operation may hang forever.

**Solution**
Use a signal pipe and do selection on it. When a new interest is configured or close request arrived, we do signal to force abort blocking selection.

### DIFF
--- a/ktor-network/posix/src/io/ktor/network/selector/SignalPoint.kt
+++ b/ktor-network/posix/src/io/ktor/network/selector/SignalPoint.kt
@@ -1,0 +1,100 @@
+/*
+ * Copyright 2014-2021 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.network.selector
+
+import io.ktor.network.util.*
+import io.ktor.utils.io.*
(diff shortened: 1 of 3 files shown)


ktor-network/posix/src/io/ktor/network/selector/SignalPoint.kt
```diff
@@ -0,0 +1,100 @@
...
+    private val readDescriptor: Int
+    private val writeDescriptor: Int
+    private val signalCounter = atomic(0)
+
+    val selectionDescriptor: Int
+        get() = readDescriptor
+
+    init {
+        val (read, write) = memScoped {
+            val pipeFds = allocArray<IntVar>(2)
```
e5l, line 25: codestyle

e5l (approved):
LGTM, please fix minor code style issues

whyoleg:
Thx for a fix! Looks like I often got into this situation! Will retest it after release.

PR #2306, opened 2021-01-15
KTOR-1641 Implementation for subrouting feature config

rsinukov:
**Subsystem**

**Motivation**

**Solution**
Describe your solution.


e5l (changes requested):

e5l (approved):
Please prepare the docs before merging.

ktor-server/ktor-server-core/jvm/src/io/ktor/application/DynamicConfigFeature.kt
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    // dynamic feature needs to be installed into routing, because only routing will have all interceptors
+    @Suppress("UNCHECKED_CAST")
+    val installPipeline = when (this) {
+        is Application -> routing {} as P
+        else -> this
+    }
+    val installed = feature.install(installPipeline)
+    val registry = installPipeline.attributes.computeIfAbsent(featureRegistryKey) { Attributes(true) }
```
cy6erGn0m, line 81: Extract and reuse `Pipeline<**>. featureRegistry` ?

ktor-server/ktor-server-core/jvm/src/io/ktor/features/ContentNegotiation.kt
```diff
@@ -42,13 +42,17 @@ public data class ContentTypeWithQuality(val contentType: ContentType, val quali
...
- *
- * @param registrations is a list of registered converters for ContentTypes
  */
-public class ContentNegotiation(
-    public val registrations: List<ConverterRegistration>,
-    private val acceptContributors: List<AcceptHeaderContributor>
-) {
+public class ContentNegotiation internal constructor() {
+
+    @Deprecated("This will be removed", level = DeprecationLevel.ERROR)
```
cy6erGn0m, line 48: Unfortunately, we can't make it ERROR in a step, WARNING then ERROR

ktor-utils/common/src/io/ktor/util/Attributes.kt
```diff
@@ -10,7 +10,26 @@ package io.ktor.util
...
  */
 public class AttributeKey<T>(public val name: String) {
-    override fun toString(): String = if (name.isEmpty()) super.toString() else "AttributeKey: $name"
+
+    override fun toString(): String = when {
+        name.isEmpty() -> super.toString()
+        else -> "AttributeKey: $name"
+    }
+
+    override fun equals(other: Any?): Boolean {
```
cy6erGn0m, line 19: I'am not sure about this: this breaks semantics since different attribute keys with accidentally same names may start colliding. Imagine several unrelated people using different attributes with silly names like `"attribute1"` may surprisingly clash each other

cy6erGn0m (changes requested):

ktor-utils/common/src/io/ktor/util/Attributes.kt
rsinukov: thanks, fixed

ktor-server/jvm/src/io/ktor/server/application/plugins/api/ServerPlugin.kt
```diff
@@ -12,31 +12,56 @@ import io.ktor.server.request.*
...
+import kotlin.random.*
 
 /**
  * A plugin that embeds into the HTTP pipeline and extends Ktor functionality.
  **/
-public abstract class ServerPlugin<Configuration : Any> private constructor(
-    pluginFactory: ServerPluginFactory<Configuration>
+public sealed class ServerPlugin<Configuration : Any> private constructor(
+    internal val key: AttributeKey<out ServerPlugin<Configuration>>
 ) : PluginContext {
```
Ololoshechkin, line 22: What if user wants to also extend `ServerPlugin` ?

ktor-server/jvm/src/io/ktor/server/application/plugins/api/ServerPlugin.kt
```diff
@@ -12,31 +12,56 @@ import io.ktor.server.request.*
...
+
+    public abstract class ApplicationPlugin<Configuration : Any>(
+        pluginFactory: ServerApplicationPluginFactory<Configuration>
+    ) : ServerPlugin<Configuration>(pluginFactory.key) {
```
Ololoshechkin, line 26: Missing KDoc

ktor-server/jvm/src/io/ktor/server/application/plugins/api/ServerPlugin.kt
```diff
@@ -12,31 +12,56 @@ import io.ktor.server.request.*
...
 import io.ktor.util.pipeline.*
-import kotlin.random.Random
+import kotlin.random.*
 
 /**
  * A plugin that embeds into the HTTP pipeline and extends Ktor functionality.
  **/
-public abstract class ServerPlugin<Configuration : Any> private constructor(
-    pluginFactory: ServerPluginFactory<Configuration>
+public sealed class ServerPlugin<Configuration : Any> private constructor(
```
Ololoshechkin, line 20: What if a user wants to extend this class as well?

ktor-server/jvm/src/io/ktor/server/application/plugins/api/ServerPlugin.kt
```diff
@@ -12,31 +12,56 @@ import io.ktor.server.request.*
...
+        /**
+         * A configuration for the current plugin. It may hold required data and any functionality that can be used by your
+         * plugin.
+         **/
+        public abstract val pluginConfig: Configuration
+    }
+
+    public abstract class RoutingScopedPlugin<Configuration : Any>(
+        pluginFactory: ServerRoutingScopedPluginFactory<Configuration>
+    ) : ServerPlugin<Configuration>(pluginFactory.key) {
```
Ololoshechkin, line 37: Missing KDoc

ktor-server/jvm/src/io/ktor/server/application/plugins/api/ServerPlugin.kt
```diff
@@ -12,31 +12,56 @@ import io.ktor.server.request.*
...
+         **/
+        public val PipelineContext<*, ApplicationCall>.pluginConfig: Configuration
+            get() = call.attributes[configKey]
+
+        /**
+         * A configuration for the current plugin for this pipeline. It may hold required data and any
+         * functionality that can be used by your plugin.
+         **/
+        public val CallHandlingContext.pluginConfig: Configuration
+            get() = context.pluginConfig
```
Ololoshechkin, line 51: Why these extensions are declared inside `RoutingScopedPlugin`?
If they are useful maybe it's worth making them available for any plugin?

ktor-server/jvm/src/io/ktor/server/application/plugins/api/ServerPlugin.kt
```diff
@@ -12,31 +12,56 @@ import io.ktor.server.request.*
...
-import kotlin.random.Random
+import kotlin.random.*
 
 /**
  * A plugin that embeds into the HTTP pipeline and extends Ktor functionality.
  **/
-public abstract class ServerPlugin<Configuration : Any> private constructor(
-    pluginFactory: ServerPluginFactory<Configuration>
+public sealed class ServerPlugin<Configuration : Any> private constructor(
+    internal val key: AttributeKey<out ServerPlugin<Configuration>>
```
Ololoshechkin, line 21: Please, add `key` to KDoc.
Also, I'm not sure it's a good idea to require `key` from a user. `AttributeKey` seems to me as some weared wrapper class, and we decided to move to simpler abstractions.
Maybe receive `name: String` here and then construct a `key` from it? Same way it is now done in `createPlugin`

ktor-server/jvm/src/io/ktor/server/application/plugins/api/ServerPluginFactory.kt
```diff
@@ -11,9 +11,24 @@ import io.ktor.util.pipeline.*
...
+ * A factory class to be passed to the [install] function for creating a [ServerPlugin.ApplicationPlugin]
+ * instance and installing it into the current application context.
+ **/
+public abstract class ServerApplicationPluginFactory<Configuration : Any> :
+    ServerPluginFactory<Configuration, ServerPlugin.ApplicationPlugin<Configuration>>(),
+    ApplicationPlugin<ApplicationCallPipeline, Configuration, ServerPlugin.ApplicationPlugin<Configuration>>
+
+/**
+ * A factory class to be passed to the [install] function for creating a [ServerPlugin.RoutingScopedPlugin]
+ * instance and installing it into the current application context.
```
Ololoshechkin, line 27: Maybe `application context` -> `routing context` ?

ktor-server/jvm/src/io/ktor/server/application/plugins/api/ServerPluginFactory.kt
```diff
@@ -11,9 +11,24 @@ import io.ktor.util.pipeline.*
  * A factory class to be passed to the [install] function for creating a [ServerPlugin]
  * instance and installing it into the current application context.
  **/
-public abstract class ServerPluginFactory<Configuration : Any>(
-    public val name: String
-) : ApplicationPlugin<ApplicationCallPipeline, Configuration, ServerPlugin<Configuration>>
+public sealed class ServerPluginFactory<Configuration : Any, TPlugin : ServerPlugin<Configuration>> :
```
Ololoshechkin, line 14: Maybe add some information about the second parameter to KDoc?

ktor-server/jvm/src/io/ktor/server/application/plugins/api/ServerPlugin.kt
```diff
@@ -274,12 +326,68 @@ public abstract class ServerPlugin<Configuration : Any> private constructor(
...
+
+            onResponseInterceptions.forEach {
+                it.action(pipeline.sendPipeline)
+            }
+
+            afterResponseInterceptions.forEach {
+                it.action(pipeline.sendPipeline)
+            }
+        }
     }
```
Ololoshechkin, line 392: `createRoutingScopedPlugin` function with `Configuration` is missing. Is there a reason for that?

ktor-server/jvm/test/io/ktor/tests/plugins/api/ServerPluginTest.kt
```diff
@@ -294,4 +295,65 @@ class ServerPluginTest {
...
+        class Config(var data: String)
+
+        val plugin = createRoutingScopedPlugin("F", { Config("default") }) {
+            onCall {
+                context.call.respond(pluginConfig.data)
+            }
+        }
+
+        withTestApplication {
+            application.install(plugin)
```
Ololoshechkin, line 310: More of a design question.
As I see here, `RoutingScopedPlugin` can be still installed to an application.
And the only difference with a normal plugin is that it overrides settings when installed on subroutes.

1. Doesn't it make sense to make all plugins routing-scoped?
2. From a user perspective, when I'm about to create a plugin, how should I choose: should it be routing-scoped or not. Does it make sense not to make it routing-scoped for some reason?
3. Do we have routing-scoped plugins among our out-of-the-box plugins?

ktor-utils/common/src/io/ktor/util/Attributes.kt
Ololoshechkin: Please, resolve obsolete conversations if possible :)

ktor-server/jvm/src/io/ktor/server/application/plugins/api/ServerPluginFactory.kt
```diff
@@ -23,16 +38,9 @@ public abstract class ServerPluginFactory<Configuration : Any>(
...
-    return attributes[pluginRegistryKey].getOrNull(plugin.key)
-        ?: throw MissingApplicationPluginException(plugin.key)
-}
-
-internal fun <A : Pipeline<*, ApplicationCall>> A.findServerPlugin(
-    plugin: ServerPluginFactory<*>
+    plugin: ServerPluginFactory<ConfigurationT, *>
 ): ServerPlugin<*> {
-    return attributes[pluginRegistryKey].getOrNull(plugin.key)
+    return pluginRegistry.getOrNull(plugin.key)
```
Ololoshechkin, line 43: 1. Should it call `findPluginInRoute` for the case of `RoutingScopedPlugin`?
2. Could you please write tests checking this method returns different plugin instances in different subroutes?

Ololoshechkin (changes requested):

ktor-server/jvm/src/io/ktor/server/application/plugins/api/ServerPlugin.kt
rsinukov: This constructor is private and invoked only inside `creade[X]Plugin`. So users can not pass `key` here.

ktor-server/jvm/src/io/ktor/server/application/plugins/api/ServerPlugin.kt
rsinukov: Is there a reason for it? I prefer to forbid inheritance until use cases are clear, because it's something that is hard to take away or change when needed.

ktor-server/jvm/src/io/ktor/server/application/plugins/api/ServerPlugin.kt
rsinukov: It is here in this file (https://github.com/ktorio/ktor/pull/2306/files/9dd5d45bafef0d87ad0714c48fce4360dcbfc00e#diff-244208d28b7b8aa486f55d16a600eae10295b68b68029cf9c54d14af37d4ace5R286)
Or am I missing something?

ktor-server/jvm/test/io/ktor/tests/plugins/api/ServerPluginTest.kt
rsinukov: Please read the design doc because it answers your questions https://youtrack.jetbrains.com/issue/KTOR-928
TL;DR

The main difference is that routing scoped plugins can access their config only inside call context.
1. No, some plugins can't be routing scoped. Imagine a plugin that schedules some cron jon on application start. It has nothing to do with routing.
2. Usually it's clear when you look at what is the purpose of the plugin.
3. Not yet, but we will default plugins if they can be routing scoped

ktor-server/jvm/src/io/ktor/server/application/plugins/api/ServerPluginFactory.kt
rsinukov: 1. Thanks, I will
2. No, because there is only one plugin instance installed, but its configuration is changed depending on the call.

ktor-server/jvm/src/io/ktor/server/application/plugins/api/ServerPluginFactory.kt
rsinukov: No need, I removed the second parameter =)

ktor-server/jvm/src/io/ktor/server/application/plugins/api/ServerPluginFactory.kt
rsinukov: done

ktor-server/jvm/src/io/ktor/server/application/plugins/api/ServerPlugin.kt
rsinukov: done

ktor-server/jvm/src/io/ktor/server/application/plugins/api/ServerPlugin.kt
rsinukov: done

ktor-server/jvm/src/io/ktor/server/application/plugins/api/ServerPlugin.kt
Ololoshechkin: Your link says, "We went looking everywhere, but couldn’t find those commits."

ktor-server/jvm/test/io/ktor/tests/plugins/api/ServerPluginTest.kt
Ololoshechkin: Thanks!
I would personally add some example and a brief explanation to KDoc because users may ask the same questions when writing code. 
cc: @andreyaksenov ?

ktor-server/ktor-server-core/jvm/src/io/ktor/server/application/plugins/api/ServerPlugin.kt
```diff
@@ -206,58 +236,87 @@ public abstract class ServerPlugin<Configuration : Any> private constructor(
...
+         *      // This block will be executed when you call install(MyPlugin)
+         *
+         *      onCall { call ->
+         *          // Prints requested URL each time your application receives a call:
+         *          println(call.request.uri)
+         *      }
+         * }
+         *
+         * route("a") {
+         *   install(MyPlugin)
```
Ololoshechkin, line 289: I would add some example describing the semantics of the re-installation into a different subroute

Ololoshechkin (approved):

ktor-server/ktor-server-core/jvm/src/io/ktor/server/application/plugins/api/ServerPlugin.kt
Ololoshechkin: cc: @andreyaksenov

PR #2307, opened 2021-01-16
Fix bug causing headers to be omitted when connecting to websocket using K/JS (nodejs) (KTOR-1629)

DRSchlaubi:
**Subsystem**
Websocket client on Kotlin/JS (nodejs)

**Motivation**
Because the browser websocket client does not support custom headers on the WS handshake the Ktor JS engine just ignores them completely however
the used NodeJS package (ws) supports setting custom headers

Issue: [KTOR-1629](https://youtrack.jetbrains.com/issue/KTOR-1629)

**Solution**
Change the JsClientEngine.createWebsocketFunction() to also acknowledge the headers on nodejs
```diff
- private fun createWebSocket(urlString_capturingHack: String): WebSocket {
+ private fun createWebSocket(urlString_capturingHack: String, headers: Headers): WebSocket {
        return if (PlatformUtils.IS_NODE) {
            val ws_capturingHack = js("require('ws')")
+            val headers_capturingHack: dynamic = object {}
+           headers.forEach { name, values ->
+                val value = values.first()
+                headers_capturingHack[name] = value
+           }
-            js("new ws_capturingHack(urlString_capturingHack)")
+            js("new ws_capturingHack(urlString_capturingHack, { headers: headers_capturingHack })")
        } else {
            js("new WebSocket(urlString_capturingHack)")
        }
    }
```


e5l (changes requested):
Hey @DRSchlaubi, thanks for the PR.
Could you add a test that fails before and pass after the PR?

DRSchlaubi:
> Hey @DRSchlaubi, thanks for the PR.
> Could you add a test that fails before and pass after the PR?

I don't really know-how. I thought I just copy the echo test, but echo.websocket.org does not send back the headers
So I thought to check whether the headers get added to DefaultWebsocketSession.call.request but it does that either way

e5l:
Sure, you can add the test here: https://github.com/ktorio/ktor/blob/master/ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/WebSocketTest.kt similar to other tests. And here is the test server: https://github.com/ktorio/ktor/blob/master/ktor-client/ktor-client-tests/jvm/src/io/ktor/client/tests/utils/tests/WebSockets.kt. It will start automatically on the test run.

For instance, you can set up a server handler to collect header fields from a request and send them back in a text frame(and verify them on the client-side for sure :) )

DRSchlaubi:
this would work for the ws tests but what about the wss tests since the test server can't use https

e5l:
Could you tell me why do we need wss to test headers content?

DRSchlaubi:
It's just that we test the connection to ws and wss. I don't really think this matters though

e5l:
Yep, let's just have a regular ws test

DRSchlaubi:
I can't actually read the code style check so don't know what to change

ktor-client/ktor-client-core/js/src/io/ktor/client/engine/js/JsClientEngine.kt
```diff
@@ -54,12 +54,22 @@ internal class JsClientEngine(override val config: HttpClientEngineConfig) : Htt
...
+        "UNUSED_VARIABLE", "UNUSED_PARAMETER",
+    )
+    private fun createWebSocket(urlString_capturingHack: String, headers: Headers): WebSocket {
         return if (PlatformUtils.IS_NODE) {
             val ws_capturingHack = js("require('ws')")
-            js("new ws_capturingHack(urlString_capturingHack)")
+            val headers_capturingHack: dynamic = object {}
+            headers.forEach { name, values ->
+                val value = values.first()
+                headers_capturingHack[name] = value
```
hfhbd, line 70: How does `first()` work with a header having multiple values, eg 
```
Cookie: name=value; name2=value2; name3=value3
```

DRSchlaubi: the header property is a JS object. And afaik you can't set multiple properties with the same name of a JS object, right?
So I assume the WebSocket package can't do that?

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/WebSocketTest.kt
```diff
@@ -89,6 +95,54 @@ class WebSocketTest : ClientLoader() {
         }
     }
 
+    @Test
+    fun testWsHandshakeHeaders() {
+        if(PlatformUtils.IS_ BROWSER) return // browser websocket client does not support custom headers so the test get's ignored
```
hfhbd, line 100: With this space, you code does not compile

DRSchlaubi: OOOPS I didn't see that

ktor-client/ktor-client-core/js/src/io/ktor/client/engine/js/JsClientEngine.kt
hfhbd: Right, and the developers of NodeJS had the same problem: https://github.com/nodejs/node/issues/3591. According to https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2, they joined the values using a comma.
So I would use
```kotlin
val value = values.joinToString(",")
```` 
instead. But maybe, as I am not a webserver/RFC profi, I am wrong :)

Comment (DRSchlaubi) [reply ktor-client/ktor-client-core/js/src/io/ktor/client/engine/js/JsClientEngine.kt:70]: Well I am not a webserver/RFC profi either so IDK

Comment (hfhbd) [reply ktor-client/ktor-client-core/js/src/io/ktor/client/engine/js/JsClientEngine.kt:70]: One last example: https://developer.mozilla.org/en-US/docs/Web/API/Headers/append#example. Mozilla also concats the values with a comma, if the header allows multiple values...

https://github.com/ktorio/ktor/blob/21771311f3dd1723d557ab50b5f1e2621d2d47e3/ktor-client/ktor-client-core/common/src/io/ktor/client/engine/Utils.kt#L37

Comment (DRSchlaubi) [reply ktor-client/ktor-client-core/js/src/io/ktor/client/engine/js/JsClientEngine.kt]: I added another unit test and it passes with your change so it seems to work

**Comment by ahardewig** [comment]
Hey there, this issue seems to be affecting me too. A host exposes a WSS endpoint that needs a custom header to be sent in the request for validation. Despite being specified in the HttpHeaders, it's not propagated up to the actual request.  Glad to see a fix is already being worked on! Is there anything I can do to help out?

File: ktor-client/ktor-client-core/js/src/io/ktor/client/engine/js/JsClientEngine.kt [hunk 3]
```diff
@@ -54,12 +54,22 @@ internal class JsClientEngine(override val config: HttpClientEngineConfig) : Htt
...
             val ws_capturingHack = js("require('ws')")
-            js("new ws_capturingHack(urlString_capturingHack)")
+            val headers_capturingHack: dynamic = object {}
+            headers.forEach { name, values ->
+                val value = values.joinToString(",")
+                headers_capturingHack[name] = value
+            }
+            js("new ws_capturingHack(urlString_capturingHack, { headers: headers_capturingHack })")
         } else {
             js("new WebSocket(urlString_capturingHack)")
```
hfhbd, line 74: What about browser websocket support?

DRSchlaubi:
Actually this should be done already not sure if the lot team wants me to fix anything

hfhbd:
@DRSchlaubi could you rebase your PR with the latest master, which refactored the JS engine?
Eg:
````kotlin
// JsClientEngine.kt
val properties: dynamic = object { }
headers.forEach { name, values ->
    val value = values.joinToString(",")
    properties.headers[name] = value
}
NodeWebsocket(urlString, properties)
````

**Comment by DRSchlaubi** [comment]
Oh you now separated between nodeWebsocket and JS WebSocket, right? I*ll try to update the PR

**Comment by DRSchlaubi** [comment]
I couldn't run the tests after merging in master. I let them run for about an hour and Intellij still displayed "instanceiating tests"

**Comment by hfhbd** [comment]
@DRSchlaubi Did you use `.gradlew allTests` using the terminal or IntelliJ? 
Bildschirmfoto 2021-02-18 um 10 02 31

**Comment by DRSchlaubi** [comment]
I used Intellij. Will try cli later. Wouldn't be the first time that K/JS test don't work in Intellij

**Comment by DRSchlaubi** [comment]
> @DRSchlaubi Did you use `./gradlew allTests` using the terminal or IntelliJ? 
> Bildschirmfoto 2021-02-18 um 10 02 31
> 

That is not even one of my tests or related to code I changed

**Comment by hfhbd** [comment]
> > @DRSchlaubi Did you use `./gradlew allTests` using the terminal or IntelliJ?
> > Bildschirmfoto 2021-02-18 um 10 02 31
> 
> That is not even one of my tests or related to code I changed

I know, the screenshot shows not your test. It is only an example.

File: ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/WebSocketTest.kt [hunk 4]
```diff
@@ -74,6 +80,71 @@ class WebSocketTest : ClientLoader() {
...
+                assertTrue(frame is Frame.Text)
+                assertEquals("PING", frame.readText())
+            }
+        }
+    }
+
+    @Test
+    fun testWsHandshakeHeaders() {
+        if (PlatformUtils.IS_BROWSER) return // browser websocket client does not support custom headers so the test gets ignored
+        clientTests(ENGINES_WITHOUT_WEBSOCKETS + "native:CIO") {
```
hfhbd, line 102: you need `return clientTests...` see https://youtrack.jetbrains.com/issue/KT-22228

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/WebSocketTest.kt
```diff
@@ -74,6 +80,71 @@ class WebSocketTest : ClientLoader() {
...
+                val header = headers[CUSTOM_HEADER]?.first()
+                assertEquals(CUSTOM_HEADER_VALUE, header)
+            }
+        }
+    }
+
+    @Test
+    fun testWsHandshakeHeadersWithMultipleValues() {
+        if (PlatformUtils.IS_BROWSER) return // browser websocket client does not support custom headers so the test gets ignored
+        clientTests(ENGINES_WITHOUT_WEBSOCKETS + "native:CIO") {
```
hfhbd, line 126: you need `return clientTests...` see https://youtrack.jetbrains.com/issue/KT-22228

DRSchlaubi:
Also Intellij doesn't like to import ktor properly

DRSchlaubi:
Yeah now i am stuck with this again
``` > Task :ktor-client:ktor-client-tests:startTestServer
[TestServer] start
<========-----> 64% EXECUTING [15m 47s]
> IDLE
> IDLE
> IDLE
> IDLE
> IDLE
> :kotlinNpmInstall > [4/4] Building fresh packages...
> IDLE
> IDLE
> IDLE

```

DRSchlaubi:
Tests are stuck at downloading yarn deps

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/WebSocketTest.kt
```diff
@@ -74,6 +80,71 @@ class WebSocketTest : ClientLoader() {
...
+
+                val frame = session.incoming.receive()
+                assertTrue(frame is Frame.Text)
+                val headers =
+                    Json.decodeFromString<Map<String, List<String>>>(frame.readText())
+                val header = headers[CUSTOM_HEADER]?.first()
+                assertEquals(CUSTOM_HEADER_VALUE, header)
+            }
+        }
+    }
```
hfhbd, line 121: Compile error, this bracket is false

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/WebSocketTest.kt
```diff
@@ -74,6 +80,71 @@ class WebSocketTest : ClientLoader() {
...
+
+                val frame = session.incoming.receive()
+                assertTrue(frame is Frame.Text)
+                val headers =
+                    Json.decodeFromString<Map<String, List<String>>>(frame.readText())
+                val header = headers[CUSTOM_HEADER]
+                assertEquals(listOf(CUSTOM_HEADER_VALUE, CUSTOM_HEADER_VALUE), header)
+            }
+        }
+    }
```
hfhbd, line 146: Compile error, this bracket is false

hfhbd (changes requested):

ktor-client/ktor-client-core/js/src/io/ktor/client/engine/js/node/NodeDefinitions.kt
```diff
@@ -7,7 +7,7 @@ package io.ktor.client.engine.js.node
 import io.ktor.client.engine.js.compatibility.*
 import org.w3c.dom.*
 import org.w3c.fetch.*
-import kotlin.js.*
+import kotlin.js.Promise
```
hfhbd, line 10: code style, use star import

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/WebSocketTest.kt
```diff
@@ -74,6 +80,71 @@ class WebSocketTest : ClientLoader() {
...
+                    header(CUSTOM_HEADER, CUSTOM_HEADER_VALUE)
+                    header(CUSTOM_HEADER, CUSTOM_HEADER_VALUE)
+                }
+
+                val frame = session.incoming.receive()
+                assertTrue(frame is Frame.Text)
+                val headers =
+                    Json.decodeFromString<Map<String, List<String>>>(frame.readText())
+                val header = headers[CUSTOM_HEADER]
+                assertEquals(listOf(CUSTOM_HEADER_VALUE, CUSTOM_HEADER_VALUE), header)
```
hfhbd, line 143: Also your test `testWsHandshakeHeadersWithMultipleValues` failed with `AssertionError: Expected <[custom_header_value, custom_header_value]>, actual <[custom_header_value,custom_header_value]>.` because you join all same headers using `joinToString(","), so the same header would not return `listOf(headerValue1, headerValue2)` but instead `listOf("$headerValue1,$headerValue2")`.
The list in your test has a size of 1 with one String containing both headers. Fix your test or fix the response in the Testserver, eg by using flattenEntries(), each pair has a unique String key and a String value containing all headerValues or use `toMap.mapValues { it -> it.first() }` which results into Map<String, String>.

DRSchlaubi: Sorry for delaying it so much. I have no idea why but IDEA doesn't seem to like that project. now it's not even recognized as a Gradle Project anymore

DRSchlaubi: I pushed a possible fix but since importing is stuck at
> warning workspace-aggregator-6d4bb098-5ecb-4134-acb8-43e7347bc4a3 > ktor-ktor-http-jsLegacy-test > webpack > micromatch > snapdragon > source-map-resolve > resolve-url@0.2.1: https://github.com/lydell/resolve-url#deprecated
warning workspace-aggregator-6d4bb098-5ecb-4134-acb8-43e7347bc4a3 > ktor-ktor-http-jsLegacy-test > webpack > micromatch > snapdragon > source-map-resolve > urix@0.1.0: Please see https://github.com/lydell/urix#deprecated

i can't test it

cy6erGn0m:
@DRSchlaubi could you please rebase the branch? Is it ready for review at the moment?

DRSchlaubi:
I can't import ktor into intellij due to the error above

DRSchlaubi:
I rebased, but i couldn't run the tests as all tests give me an iOS error and core test complains about no module being specified

e5l (approved):

PR #2308, opened 2021-01-19
KTOR-1795 Fix serialization failure due to types mismatch

rsinukov:
**Subsystem**
Server, serialization

**Motivation**
[KTOR-1795](https://youtrack.jetbrains.com/issue/KTOR-1795)

**Solution**
Move the `try/catch` block to the one level above.

### DIFF
--- a/ktor-features/ktor-serialization/jvm/src/io/ktor/serialization/SerializationConverter.kt
+++ b/ktor-features/ktor-serialization/jvm/src/io/ktor/serialization/SerializationConverter.kt
@@ -110,16 +110,39 @@ private constructor(
         contentType: ContentType,
         value: Any
     ): Any? {
-        @Suppress("UNCHECKED_CAST")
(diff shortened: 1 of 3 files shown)


ktor-features/ktor-serialization/jvm/src/io/ktor/serialization/SerializationConverter.kt
```diff
@@ -110,16 +110,39 @@ private constructor(
...
+            // can fail due to
+            // 1. https://github.com/Kotlin/kotlinx.serialization/issues/1163)
+            // 2. mismatching between compile-time and runtime types of the response.
+            null
+        }
+        if (result != null) {
+            return result
+        }
 
+        val manualSearchSerializer = searchSerializer(value, format.serializersModule)
```
cy6erGn0m, line 127: Manual sounds like we have a human pool resolving every particular case :) The search function is trying to find or guess serializer in some cases. So I would name this thinking of this "guessing"

cy6erGn0m (approved):

PR #2312, opened 2021-01-20
~ fix floating log entry

cy6erGn0m:
**Subsystem**
ktor-client-tests

**Motivation**
Test `testLogResponseWithException` is flaky because there is a floating log entry that appears in different places.

**Solution**
Provide the ability to specify a log entry that is required but could appear later in the log

### DIFF
--- a/ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/utils/Logging.kt
+++ b/ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/utils/Logging.kt
@@ -8,6 +8,14 @@ import io.ktor.client.features.logging.*
 import io.ktor.util.collections.*
 import kotlin.test.*
 
+/**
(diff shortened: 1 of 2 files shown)


rsinukov (approved):

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/utils/Logging.kt
```diff
@@ -24,6 +32,7 @@ internal class TestLogger(private vararg val expectedLog: String) : Logger {
         var actualIndex = 0
 
         val message = StringBuilder()
+        val stashed = ArrayList<String>()
```
rsinukov, line 35: minor: can be set

cy6erGn0m: This is a list intentionally: you can specify the same text multiple times

PR #2314, opened 2021-01-21
KTOR-1797 Add check for unescaped quote inside quoted header parameter

rsinukov:
**Motivation**
https://youtrack.jetbrains.com/issue/KTOR-1797

### DIFF
--- a/ktor-http/common/src/io/ktor/http/HeaderValueWithParameters.kt
+++ b/ktor-http/common/src/io/ktor/http/HeaderValueWithParameters.kt
@@ -93,7 +93,27 @@ private fun String.checkNeedEscape(): Boolean {
     return false
 }
 
-private fun String.isQuoted(): Boolean = length > 1 && first() == '\"' && last() == '\"'
+private fun String.isQuoted(): Boolean {
+    if (length < 2) {
+        return false
+    }
+    if (first() != '"' || last() != '"') {
+        return false
+    }
+    var startIndex = 1
+    do {
+        val index = indexOf('"', startIndex)
+        if (index == -1 || index == lastIndex) {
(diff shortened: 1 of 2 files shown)


ktor-http/common/src/io/ktor/http/HeaderValueWithParameters.kt
```diff
@@ -93,7 +93,27 @@ private fun String.checkNeedEscape(): Boolean {
...
+    if (length < 2) {
+        return false
+    }
+    if (first() != '"' || last() != '"') {
+        return false
+    }
+    var startIndex = 1
+    do {
+        val index = indexOf('"', startIndex)
+        if (index == -1 || index == lastIndex) {
```
cy6erGn0m, line 106: btw, `-1` is not possible since we already checked that the last is `"`

ktor-http/common/src/io/ktor/http/HeaderValueWithParameters.kt
```diff
@@ -93,7 +93,27 @@ private fun String.checkNeedEscape(): Boolean {
...
+            break
+        }
+        if (this[index - 1] != '\\') {
```
cy6erGn0m, line 109: This is not guaranteed because the slash could be escaped as well: `\\"` I personally don't think that it's so serious here. Do you?

cy6erGn0m (approved):

ktor-http/common/src/io/ktor/http/HeaderValueWithParameters.kt
rsinukov: thanks, updated and added test

PR #2315, opened 2021-01-21
Fix client bootstrap problems

rsinukov:
**Subsystem**
Client

**Motivation**
https://youtrack.jetbrains.com/issue/KTOR-1705

**Solution**
First part with cleaning up APIs


e5l:
Hey, could you change the base branch to `1.6.0`?

ktor-client/ktor-client-core/common/src/io/ktor/client/request/RequestBody.kt
```diff
@@ -0,0 +1,40 @@
...
+@PublishedApi
+@OptIn(ExperimentalStdlibApi::class)
+internal inline fun <reified T : Any> tryGetType(ignored: T): KType? = try {
+    // We need to wrap getting type in try catch because of KT-42913
+    typeOf<T>()
+} catch (_: Throwable) {
+    null
+}
+
+public inline fun <reified T> HttpRequestBuilder.body(body: T) {
```
e5l, line 24: Maybe we need to think about this method name

ktor-client/ktor-client-core/common/src/io/ktor/client/request/builders.kt
```diff
@@ -110,18 +137,64 @@ public suspend inline fun <reified T> HttpClient.head(builder: HttpRequestBuilde
     return request(builder)
 }
 
+/**
+ * Executes a [HttpClient] GET request, with the information from the [builder]
+ */
+public suspend fun HttpClient.getRaw(builder: HttpRequestBuilder): HttpResponse = get(builder)
```
e5l, line 143: We still should add these names to the design agenda. It is simpler but still not the simplest naming

e5l (approved):
Please fix the design agenda and let's decide what we should do with naming.

ktor-client/ktor-client-core/common/src/io/ktor/client/request/builders.kt
rsinukov: updated design ticket

ktor-client/ktor-client-core/common/src/io/ktor/client/request/RequestBody.kt
rsinukov: is there anything, in particular, you don't like about this name?

ktor-client/ktor-client-core/common/src/io/ktor/client/request/buildersWithUrl.kt
```diff
@@ -5,102 +5,173 @@
...
 import io.ktor.http.*
 
 /**
  * Executes a [HttpClient] GET request, with the specified [url] as Url and
  * an optional [block] receiving an [HttpRequestBuilder] for further configuring the request.
- *
- * Tries to receive a specific type [T], if fails, an exception is thrown.
  */
-public suspend inline fun <reified T> HttpClient.get(
+public suspend inline fun HttpClient.get(
```
e5l, line 15: Why do we need inline here?

ktor-features/ktor-auth/jvm/src/io/ktor/auth/OAuth1a.kt
```diff
@@ -162,18 +163,21 @@ private suspend fun requestOAuth1aAccessToken(
...
+
+            setBody(
+                WriterContent(
+                    { params.formUrlEncodeTo(this) },
+                    ContentType.Application.FormUrlEncoded
+                )
+            )
+
+            accessTokenInterceptor?.invoke(this)
+        }.body<String>()
```
e5l, line 180: Maybe we can face the same problem with `body<T>` as with `get<T>`. Should we also have some field without generic type?

ktor-features/ktor-auth/jvm/src/io/ktor/auth/OAuth1a.kt
```diff
@@ -162,18 +163,21 @@ private suspend fun requestOAuth1aAccessToken(
...
-        // some of really existing OAuth servers don't support other accept header values so keep it
-
-        body = WriterContent(
-            { params.formUrlEncodeTo(this) },
-            ContentType.Application.FormUrlEncoded
-        )
-
-        accessTokenInterceptor?.invoke(this)
-    }
+    val body = // some of really existing OAuth servers don't support other accept header values so keep it
```
e5l, line 166: I would move the comment before `val`

ktor-client/ktor-client-core/jvm/src/io/ktor/client/call/utilsJvm.kt
```diff
@@ -13,10 +13,10 @@ import java.net.*
...
  */
 @Deprecated(
-    "Unbound [HttpClientCall] is deprecated. Consider using [request<HttpResponse>(url, block)] instead.",
+    "Unbound [HttpClientCall] is deprecated. Consider using [request(url, block)] instead.",
     level = DeprecationLevel.ERROR,
-    replaceWith = ReplaceWith("this.request<HttpResponse>(url, block)", "io.ktor.client.statement.*")
+    replaceWith = ReplaceWith("this.request(url, block)", "io.ktor.client.statement.*")
 )
 @Suppress("RedundantSuspendModifier", "unused", "UNUSED_PARAMETER")
 public suspend fun HttpClient.call(url: URL, block: HttpRequestBuilder.() -> Unit = {}): HttpClientCall =
```
e5l, line 21: We can drop level error deprecations with the major release

e5l (approved):
Nice work, LGTM in general. I should try to use it in EAP. Maybe some API can be simplified further.

ktor-client/ktor-client-core/common/src/io/ktor/client/request/buildersWithUrl.kt
rsinukov: Yes, to not allocate builder block

ktor-features/ktor-auth/jvm/src/io/ktor/auth/OAuth1a.kt
rsinukov: Added `bodyAsText` and `bodyAsChannel` extensions for visibility.

PR #2316, opened 2021-01-22
KTOR-1684 Deprecate locationOrNull

cy6erGn0m:
**Subsystem**
Ktor-locations

**Motivation**
[KTOR-1684 ApplicationCall.locationOrNull raises error](https://youtrack.jetbrains.com/issue/KTOR-1684)

**Solution**
At the first step in 1.6.0 we will simply add `location` function and deprecate `locationOrNull` pointing to the new function. This will help us to fix this in 2.0.0 when we can make `locationOrNull` return the nullable `T`


Stexxe:
Will it go to minor or patch release? According to semantic versioning spec:
> Minor version MUST be incremented if any public API functionality is marked as deprecated

https://semver.org/#spec-item-7

PR #2319, opened 2021-01-24
KTOR-1914 : Allow 256 bits Cipher Suites to be used

satyamagarwal:
**Subsystem**
Client, ktor-network-tls

**Motivation**
Issue : [1914](https://youtrack.jetbrains.com/issue/KTOR-1914)

The list contains TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384 here :
https://github.com/ktorio/ktor/blob/master/ktor-network/ktor-network-tls/common/src/io/ktor/network/tls/CipherSuites.kt#L144-L152

but it is ignored by isSupported method here :
https://github.com/ktorio/ktor/blob/master/ktor-network/ktor-network-tls/jvm/src/io/ktor/network/tls/CipherSuitesJvm.kt#L9-L12

which makes these cipher suites unavailable to be used because of :
https://github.com/ktorio/ktor/blob/master/ktor-network/ktor-network-tls/jvm/src/io/ktor/network/tls/TLSHandshakeType.kt#L70

**Solution**
This PR increases this limit from 128 bits to 256 bits, adds a test to verify it.

### DIFF
--- a/ktor-network/ktor-network-tls/jvm/src/io/ktor/network/tls/CipherSuitesJvm.kt
+++ b/ktor-network/ktor-network-tls/jvm/src/io/ktor/network/tls/CipherSuitesJvm.kt
@@ -7,6 +7,8 @@ package io.ktor.network.tls
 import io.ktor.network.tls.platform.*
 
 internal actual fun CipherSuite.isSupported(): Boolean = when (platformVersion.major) {
-    "1.8.0" -> platformVersion.minor >= 161 || keyStrength <= 256
-    else -> keyStrength <= 256
+    "1.8.0" -> platformVersion.minor >= 161 || keyStrength <= 128
+    "1.7.0" -> platformVersion.minor >= 171 || keyStrength <= 128
+    "1.6.0" -> platformVersion.minor >= 181 || keyStrength <= 128
+    else -> true
 }
--- a/ktor-network/ktor-network-tls/jvm/src/io/ktor/network/tls/CipherSuitesJvm.kt
+++ b/ktor-network/ktor-network-tls/jvm/src/io/ktor/network/tls/CipherSuitesJvm.kt
@@ -7,6 +7,6 @@ package io.ktor.network.tls
 import io.ktor.network.tls.platform.*
 
 internal actual fun CipherSuite.isSupported(): Boolean = when (platformVersion.major) {
-    "1.8.0" -> platformVersion.minor >= 161 || keyStrength <= 128
-    else -> keyStrength <= 128
+    "1.8.0" -> platformVersion.minor >= 161 || keyStrength <= 256
+    else -> keyStrength <= 256
 }
--- a/ktor-client/ktor-client-cio/jvm/test/io/ktor/client/engine/cio/CIOHttpsTest.kt
+++ b/ktor-client/ktor-client-cio/jvm/test/io/ktor/client/engine/cio/CIOHttpsTest.kt
@@ -156,7 +156,8 @@ class CIOHttpsTest : TestWithKtor() {
             "https://google.com",
             "https://facebook.com",
             "https://elster.de",
-            "https://freenode.net"
+            "https://freenode.net",
+            "https://tls-v1-2.badssl.com:1012/"
         )
 
         config {


ktor-network/ktor-network-tls/jvm/src/io/ktor/network/tls/CipherSuitesJvm.kt
```diff
@@ -7,6 +7,6 @@ package io.ktor.network.tls
 import io.ktor.network.tls.platform.*
 
 internal actual fun CipherSuite.isSupported(): Boolean = when (platformVersion.major) {
-    "1.8.0" -> platformVersion.minor >= 161 || keyStrength <= 128
-    else -> keyStrength <= 128
+    "1.8.0" -> platformVersion.minor >= 161 || keyStrength <= 256
```
e5l, line 10: This will fail with jdk `1.8.0_151`

ktor-network/ktor-network-tls/jvm/src/io/ktor/network/tls/CipherSuitesJvm.kt
```diff
@@ -7,6 +7,6 @@ package io.ktor.network.tls
...
+    else -> keyStrength <= 256
```
e5l, line 11: Probably we need to add other JDKs here instead.

e5l (changes requested):

ktor-network/ktor-network-tls/jvm/src/io/ktor/network/tls/CipherSuitesJvm.kt
satyamagarwal: Hei @e5l , 

Can you help me point to correct resources to find out which jdk supports which cipher suite ? 

I looked at : https://docs.oracle.com/javase/8/docs/technotes/guides/security/StandardNames.html#ciphersuites

and this supports all that are in the list : 

```
TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256
TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384
TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384
TLS_RSA_WITH_AES_128_CBC_SHA
TLS_RSA_WITH_AES_128_GCM_SHA256
TLS_RSA_WITH_AES_256_CBC_SHA
```

satyamagarwal: Hei @e5l, 

Can you help out here please ? I would love to get this into 1.5.2

e5l: It depends on the actual JDK build. Right now the issue is that we should allow using all suites for JDK greater than 8

e5l: Basically, we need to replace `else` with the condition on the `minor` version: (<=8) and allow any key for versions greater

ktor-network/ktor-network-tls/jvm/src/io/ktor/network/tls/CipherSuitesJvm.kt
```diff
@@ -8,5 +8,7 @@ import io.ktor.network.tls.platform.*
 
 internal actual fun CipherSuite.isSupported(): Boolean = when (platformVersion.major) {
     "1.8.0" -> platformVersion.minor >= 161 || keyStrength <= 128
-    else -> keyStrength <= 128
+    "1.7.0" -> platformVersion.minor >= 171 || keyStrength <= 128
+    "1.6.0" -> platformVersion.minor >= 181 || keyStrength <= 128
+    else -> true
```
satyamagarwal, line 13: How does this looks @e5l 

Used this https://www.oracle.com/java/technologies/javase/8u161-relnotes.html as it gives security baseline. 

I still am not able to find out what key strength is supported for these earlier versions, but from jdk 9 onwards we should be able to support any key strength right ?

e5l: Yep, it looks good

e5l (approved):
LGTM, will merge after CI build

PR #2320, opened 2021-01-25
KTOR-340 Prevent sending close to already closed outgoing ws channel

rsinukov:
**Subsystem**
Web socket

**Motivation**
[Support explicit WebSocket session close](https://youtrack.jetbrains.com/issue/KTOR-340)

**Solution**
Calling `close` on `WebSocketSession` closes `outgoing` channel. When `incoming` receives `Frame.Close` it sends close to the outgoing channel, resulting in the crash. To fix just check that the outgoing channel is open.


e5l (approved):
lgtm

PR #2321, opened 2021-01-25
KTOR-1660 Publishing EAP to Space

hhariri:

### DIFF
--- a/gradle/publish.gradle
+++ b/gradle/publish.gradle
@@ -47,8 +47,8 @@ tasks.withType(AbstractPublishToMaven).all {
     onlyIf { isAvailableForPublication(publication) }
 }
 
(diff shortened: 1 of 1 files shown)


gradle/publish.gradle
```diff
@@ -58,8 +58,8 @@ publishing {
             } else {
                 url = "https://oss.sonatype.org/service/local/staging/deploy/maven2/"
```
e5l, line 59: URL should also be from the env variable

e5l: We need to use this URL for maven and space URL for EAPs

e5l (changes requested):

e5l (approved):

PR #2326, opened 2021-01-30
KTOR-1965 Add an option to ignore trailing slash in routing

rsinukov:
**Subsystem**
Server, routing

**Motivation**
https://youtrack.jetbrains.com/issue/KTOR-1165

**Solution**
Add feature to ignore trailing slash in routing

### DIFF
--- a/ktor-server/ktor-server-core/api/ktor-server-core.api
+++ b/ktor-server/ktor-server-core/api/ktor-server-core.api
@@ -1439,8 +1439,8 @@ public final class io/ktor/routing/Routing : io/ktor/routing/Route {
 	public static final field Feature Lio/ktor/routing/Routing$Feature;
 	public fun <init> (Lio/ktor/application/Application;)V
 	public final fun getApplication ()Lio/ktor/application/Application;
+	public final fun getIgnoreTrailingSlash ()Z
 	public final fun interceptor (Lio/ktor/util/pipeline/PipelineContext;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
-	public final fun isTrailingSlashMatters ()Z
 	public final fun trace (Lkotlin/jvm/functions/Function1;)V
 }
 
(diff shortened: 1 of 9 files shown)


cy6erGn0m:
Not sure if `isTrailingSlashMatters` a good name. I personally would, in opposite, name it `ignoreTrailingSlash`. But I don't insist

ktor-server/ktor-server-core/api/ktor-server-core.api
```diff
@@ -1440,6 +1440,7 @@ public final class io/ktor/routing/Routing : io/ktor/routing/Route {
 	public fun <init> (Lio/ktor/application/Application;)V
 	public final fun getApplication ()Lio/ktor/application/Application;
 	public final fun interceptor (Lio/ktor/util/pipeline/PipelineContext;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
+	public final fun isTrailingSlashMatters ()Z
```
e5l, line 1443: We need a different name for sure

e5l (changes requested):
Let's brainstorm on the parameter name. I would also check the test coverage of the routing feature

e5l:
Agree with @cy6erGn0m, `ignoreTrailingSlash` is good. We also can try `allowTrailingSlashHandlers`

rsinukov:
i like @cy6erGn0m  suggestion of `ignoreTrailingSlash `

ktor-server/ktor-server-core/jvm/src/io/ktor/routing/IgnoreTrailingSlash.kt
```diff
@@ -0,0 +1,40 @@
...
+
+import io.ktor.application.*
+import io.ktor.util.*
+
+private val IgnoreTrailingSlashAttributeKey: AttributeKey<Unit> = AttributeKey("IgnoreTrailingSlashAttributeKey")
+
+internal var ApplicationCall.ignoreTrailingSlash: Boolean
+    get() = attributes.contains(IgnoreTrailingSlashAttributeKey)
+    private set(value) = when (value) {
+        true -> attributes.put(IgnoreTrailingSlashAttributeKey, Unit)
```
e5l, line 15: why not `if else`?

e5l (approved):

PR #2340, opened 2021-02-04
KTOR-2007 Fix running custom response validation without default one

rsinukov:
**Subsystem**
Client

**Motivation**
[Custom response validation is not running when the default is disabled](https://youtrack.jetbrains.com/issue/KTOR-2007)

### DIFF
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/features/HttpCallValidator.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/features/HttpCallValidator.kt
@@ -6,6 +6,7 @@ package io.ktor.client.features
 
 import io.ktor.client.*
 import io.ktor.client.call.*
+import io.ktor.client.features.HttpCallValidator.*
 import io.ktor.client.request.*
 import io.ktor.client.statement.*
 import io.ktor.client.utils.*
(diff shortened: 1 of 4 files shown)


Stexxe (changes requested):
Unit tests are missing

rsinukov:
@Stexxe don't think that it makes sense to write unit tests here. The current feature touches a lot of entities like pipelines, client, call, request, response, interceptors, etc. I don't see a value in a test that will mock them, but if we will use real ones, we will end up in the same integration test. What do you think?

ktor-client/ktor-client-core/common/src/io/ktor/client/features/HttpCallValidator.kt
```diff
@@ -50,8 +50,13 @@ public class HttpCallValidator internal constructor(
         callExceptionHandlers: List<CallExceptionHandler>
     ) : this(responseValidators, callExceptionHandlers, true)
 
-    private suspend fun validateResponse(response: HttpResponse) {
-        responseValidators.forEach { it(response) }
+    private suspend fun validateResponse(call: HttpClientCall) {
+        val expectSuccess = call.attributes.getOrNull(ExpectSuccessAttributeKey) ?: expectSuccess
+        val validators = when {
```
Stexxe, line 55:
```
val defaultValidators = responseValidators.filter { it == DefaultValidator }.filter { expectSuccess }
val restValidators = responseValidators.filter { it != DefaultValidator }
(defaultValidators + restValidators).forEach { it(call.response) }
```

Stexxe (approved):

e5l (approved):

PR #2341, opened 2021-02-04
Move data conversion to common code to reuse it in client

rsinukov:
**Subsystem**
Client/Server

**Motivation**
https://youtrack.jetbrains.com/issue/KTOR-1780

**Solution**
Since we can't delete the old feature in the server module, just create a new one, but without JVM dependencies. This feature is not used yet but will be required for client-side Locations in future


ktor-client/ktor-client-core/common/src/io/ktor/client/call/HttpClientCall.kt
```diff
@@ -180,9 +191,15 @@ public class DoubleReceiveException(call: HttpClientCall) : IllegalStateExceptio
 @Suppress("KDocMissingDocumentation", "unused")
 public class ReceivePipelineException(
```
cy6erGn0m, line 192: Shouldn't we deprecate the old constructor as well?

ktor-server/ktor-server-core/jvm/src/io/ktor/features/DataConversion.kt
```diff
@@ -68,7 +68,7 @@ public class DataConversion(private val converters: Map<Type, ConversionService>
             return DataConversion(configuration.converters)
```
cy6erGn0m, line 68: `ConversionService` provides function `convert` taking `ConversionService` so you need to add functions with the new type as well

ktor-server/ktor-server-core/jvm/src/io/ktor/util/Parameters.kt
```diff
@@ -49,15 +53,27 @@ public inline fun Parameters.getOrFail(name: String): String {
  */
 @OptIn(ExperimentalStdlibApi::class)
 public inline fun <reified R : Any> Parameters.getOrFail(name: String): R {
-    return getOrFailImpl(name, R::class, typeOf<R>().toJavaType())
+    return getOrFailImpl(name, typeInfo<R>())
```
cy6erGn0m, line 56: `toJavaType` invocation is lost so some edge-cases may stop working because of the missing `JavaTypeAdapter`. I wonder why no tests failed because of this...

cy6erGn0m (changes requested):

ktor-server/ktor-server-core/jvm/src/io/ktor/features/DataConversion.kt
rsinukov: I didn't deprecate his class yet, because it's coupled with the Locations feature. I will address it in the next PRs to not overload this one with changes.

ktor-utils/common/src/io/ktor/util/converters/DataConversion.kt
```diff
@@ -0,0 +1,93 @@
...
+        val type: KClass<*> = value?.let { it::class } ?: return listOf()
+        val converter = converters[type] ?: DefaultConversionService
+        return converter.toValues(value)
+    }
+
+    /**
+     * Data conversion service configuration
+     */
+    public class Configuration {
+        internal val converters = mutableMapOf<KClass<*>, ConversionService>()
```
e5l, line 32: Please add native mutation tests as well

ktor-utils/common/src/io/ktor/util/converters/DataConversion.kt
```diff
@@ -0,0 +1,93 @@
...
+         * Register and [configure] convertor for reified type [T]
+         */
+        @OptIn(ExperimentalStdlibApi::class)
+        public inline fun <reified T> convert(noinline configure: DelegatingConversionService.() -> Unit): Unit =
+            convert(typeOf<T>(), configure)
+    }
+}
+
+/**
+ * Custom convertor builder
```
e5l, line 59: .

e5l: Could you describe the purpose of this class?

e5l (changes requested):

e5l:
Check API is failing

ktor-utils/common/src/io/ktor/util/converters/DataConversion.kt
rsinukov: There is test `testInstalledConversion` that mutates this map in the `common` test module. Is there anything else to test?

ktor-server/ktor-server-core/jvm/src/io/ktor/util/Parameters.kt
rsinukov: I use default java mechanisms to get `Type`. I added tests for `WildcardType` and they pass.

ktor-utils/common/src/io/ktor/util/converters/DataConversion.kt
```diff
@@ -0,0 +1,93 @@
...
+public class DelegatingConversionService internal constructor(private val kClass: KClass<*>) : ConversionService {
+
+    private var decoder: ((values: List<String>, type: TypeInfo) -> Any?)? = null
+    private var encoder: ((value: Any?) -> List<String>)? = null
+
+    /**
+     * Configure decoder function. Only one decoder could be supplied
+     * @throws IllegalStateException
+     */
+    public fun decode(converter: (values: List<String>, type: TypeInfo) -> Any?) {
```
e5l, line 70: I'm not sure that it's a good pattern of setting fields. Why can't we have the configuration and pass this `var` in the constructor?

e5l (changes requested):
LGTM, except some API. Could you elaborate on this first?

ktor-utils/common/src/io/ktor/util/converters/DataConversion.kt
rsinukov: It's a copy-paste of an old API. But I agree with you that it can be improved while we have a chance. I will rework it.

ktor-utils/common/src/io/ktor/util/converters/DataConversion.kt
e5l: thanks, they are good

ktor-utils/common/src/io/ktor/util/converters/DataConversion.kt
rsinukov: Updated it fixing some nullability issues and adding generics. We still need to have similar functions to have user side code look like this 
```kotlin
convert<EntityID> {
    decode { values, _ -> EntityID1(values[0], values[1])}
    encode { value -> listOf(value.typeId, value.entityId) }
}
```
I find it better then 
```kotlin
convert<EntityID> {
    decoder = { values, _ -> EntityID1(values[0], values[1])}
    encoder = { value -> listOf(value.typeId, value.entityId) }
}
```

cy6erGn0m (approved):
Please notice the compilation error

e5l (approved):
lgtm

PR #2345, opened 2021-02-09
KTOR-1801 Fix Netty IllegalReferenceCountException

cy6erGn0m:
**Subsystem**
ktor-server-netty

**Motivation**
[KTOR-1801 server/netty: IllegalReferenceCountException](https://youtrack.jetbrains.com/issue/KTOR-1801)

Sometimes, the Netty engines does call termination twice (from different contexts). The underlying implementation does unsafe things so double disposal is not allowed. The double release is a very unlikely case this is why we don't face it in tests.

**Solution**
Use atomic flag to ensure we only release it once.

**Problem**
This is very time-sensitive issue and there is no way to write a stable test for this.

### DIFF
--- a/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationCall.kt
+++ b/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationCall.kt
@@ -8,6 +8,7 @@ import io.ktor.application.*
 import io.ktor.server.engine.*
 import io.netty.channel.*
 import io.netty.util.*
+import kotlinx.atomicfu.*
 import kotlinx.coroutines.*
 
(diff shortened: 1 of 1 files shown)


whyoleg:
Why not to use [ReferenceCountUtil.safeRelease](https://netty.io/4.0/api/io/netty/util/ReferenceCountUtil.html#safeRelease-java.lang.Object-) instead?

cy6erGn0m:
> Why not to use [ReferenceCountUtil.safeRelease](https://netty.io/4.0/api/io/netty/util/ReferenceCountUtil.html#safeRelease-java.lang.Object-) instead?

Because it does log error. Also, from the docs:
> It is usually recommended to use release(Object) instead, unless you absolutely need to swallow an exception.

rsinukov (approved):

PR #2347, opened 2021-02-10
KTOR-1464 Fix client Digest auth realm handling

cy6erGn0m:
Fix client digest auth realm handling to use the realm provided by the server if the realm is not configured in the client feature

**Subsystem**
ktor-client-auth

**Motivation**
[Digest authentition: using the realm of the 401 response to an initial incomplete request for an immediate second complete request including realm.](https://youtrack.jetbrains.com/issue/KTOR-1464)

This replaces PR #2236


e5l (approved):
lgtm

PR #2349, opened 2021-02-10
KTOR-2054 Fix routing with non optional trailing path parameter

rsinukov:
**Subsystem**
Server, routing

**Motivation**
[Mandatory Path Segment parameter can be empty, if no explicit route with trailing / is defined](https://youtrack.jetbrains.com/issue/KTOR-2054)

### DIFF
--- a/ktor-server/ktor-server-tests/jvm/test/io/ktor/tests/server/routing/RoutingResolveTest.kt
+++ b/ktor-server/ktor-server-tests/jvm/test/io/ktor/tests/server/routing/RoutingResolveTest.kt
@@ -1042,6 +1042,125 @@ class RoutingResolveTest {
     }
 
     @Test
+    fun testRoutingWithWildcardTrailingPathParameter() = withTestApplication {
+        application.routing {
+            get("test/*") {
+                call.respondText("test")
+            }
+        }
+        on("making /test request") {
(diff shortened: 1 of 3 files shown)


ktor-server/ktor-server-tests/jvm/test/io/ktor/tests/server/routing/RoutingResolveTest.kt
```diff
@@ -1041,6 +1041,125 @@ class RoutingResolveTest {
         }
     }
 
+    @Test
+    fun testRoutingWithWildcardTrailingPathParameter() = withTestApplication {
+        application.routing {
+            get("test/*") {
```
e5l, line 1047: I'm not sure that `*` is a mandatory parameter. What is a current behavior?

e5l (changes requested):

ktor-server/ktor-server-tests/jvm/test/io/ktor/tests/server/routing/RoutingResolveTest.kt
rsinukov: from the doc
> * (wildcard) matches any path segment and can't be missing.

e5l (approved):

PR #2353, opened 2021-02-11
Cy/micrometer defaults

cy6erGn0m:
**Subsystem**
Ktor-metrics-micrometer

**Motivation**
[KTOR-1781 Set Default Registry for MicrometerMetrics feature](https://youtrack.jetbrains.com/issue/KTOR-1781)
There is no default for micrometer registry so the empty configuration does always crash.

**Solution**
Make `LoggingMeterRegistry` the default registry with it's default options so it will log metrics every minute using an slf4j logger.

I believe it is not a breaking change so we can release this change in 1.5.2

### DIFF
--- a/ktor-features/ktor-metrics-micrometer/jvm/src/io/ktor/metrics/micrometer/MicrometerMetrics.kt
+++ b/ktor-features/ktor-metrics-micrometer/jvm/src/io/ktor/metrics/micrometer/MicrometerMetrics.kt
@@ -17,6 +17,7 @@ import io.micrometer.core.instrument.binder.jvm.*
 import io.micrometer.core.instrument.binder.system.*
 import io.micrometer.core.instrument.config.*
 import io.micrometer.core.instrument.distribution.*
+import io.micrometer.core.instrument.logging.*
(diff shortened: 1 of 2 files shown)


ktor-features/ktor-metrics-micrometer/jvm/src/io/ktor/metrics/micrometer/MicrometerMetrics.kt
```diff
@@ -87,12 +88,12 @@ public class MicrometerMetrics private constructor(
...
     public class Configuration {
-
-        public lateinit var registry: MeterRegistry
+        public var registry: MeterRegistry = LoggingMeterRegistry()
 
         public var distinctNotRegisteredRoutes: Boolean = true
 
-        internal fun isRegistryInitialized() = this::registry.isInitialized
+        @Deprecated("Will be removed because there is always registry")
+        internal fun isRegistryInitialized() = true
```
e5l, line 96: Could you tell me why we need deprecation for internal?

e5l (approved):

e5l:
Please check public API

PR #2354, opened 2021-02-11
NettyApplicationEngine: providing a configureBootstrap in the configuration throws IllegalStateException: group set already

Stexxe:
https://youtrack.jetbrains.com/issue/KTOR-2078

### DIFF
--- a/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationEngine.kt
+++ b/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationEngine.kt
@@ -84,25 +84,31 @@ public class NettyApplicationEngine(
     /**
      * [EventLoopGroupProxy] for accepting connections
      */
-    private val connectionEventGroup: EventLoopGroupProxy by lazy {
-        EventLoopGroupProxy.create(configuration.connectionGroupSize)
+    private val connectionEventGroup: EventLoopGroup by lazy {
+        customBootstrap.config().group() ?: EventLoopGroupProxy.create(configuration.connectionGroupSize)
     }
 
     /**
      * [EventLoopGroupProxy] for processing incoming requests and doing engine's internal work
      */
-    private val workerEventGroup: EventLoopGroupProxy by lazy {
-        if (configuration.shareWorkGroup) {
+    private val workerEventGroup: EventLoopGroup by lazy {
+        val defaultGroup = if (configuration.shareWorkGroup) {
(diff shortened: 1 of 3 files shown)


ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationEngine.kt
```diff
@@ -125,7 +125,12 @@ public class NettyApplicationEngine(
         environment.connectors.map { connector ->
             ServerBootstrap().apply {
                 configuration.configureBootstrap(this)
-                group(connectionEventGroup, workerEventGroup)
+
+                val conf = config()
```
e5l, line 129: I would use different name.

e5l (approved):
LGTM. @cy6erGn0m, could you also take a look?

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationEngine.kt
Stexxe: What's wrong with this name?

e5l: It's a shortcut

Stexxe: So what?

e5l (approved):

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationEngine.kt
```diff
@@ -125,7 +125,11 @@ public class NettyApplicationEngine(
         environment.connectors.map { connector ->
             ServerBootstrap().apply {
                 configuration.configureBootstrap(this)
-                group(connectionEventGroup, workerEventGroup)
+
+                if (config().group() == null && config().childGroup() == null) {
+                    group(connectionEventGroup, workerEventGroup)
+                }
+
                 channel(connectionEventGroup.channel.java)
```
cy6erGn0m, line 133: If we already have an event group configured, we overwrite a channel class here that may be wrong because different event groups may have different channel classes. Not sure if it actually dangerous but there is a chance of incompatible channels produced from here.

cy6erGn0m (changes requested):

cy6erGn0m (approved):

PR #2356, opened 2021-02-12
Fix flaky ClientSocketTest.testSelfConnect[jvm]

Stexxe:
https://youtrack.jetbrains.com/issue/KTOR-2060

### DIFF
--- a/ktor-network/jvm/test/io/ktor/network/sockets/tests/ClientSocketTest.kt
+++ b/ktor-network/jvm/test/io/ktor/network/sockets/tests/ClientSocketTest.kt
@@ -7,14 +7,17 @@ package io.ktor.network.sockets.tests
 import io.ktor.network.selector.*
 import io.ktor.network.sockets.*
 import io.ktor.network.sockets.Socket
+import io.ktor.network.sockets.SocketImpl
 import io.ktor.utils.io.*
+import io.mockk.*
 import kotlinx.coroutines.*
 import kotlinx.coroutines.debug.junit4.*
 import org.junit.*
 import org.junit.rules.*
-import java.net.InetSocketAddress
+import java.net.*
 import java.net.ServerSocket
 import java.nio.*
+import java.nio.channels.*
 import java.util.concurrent.*
 import kotlin.concurrent.*
 import kotlin.test.*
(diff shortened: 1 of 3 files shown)


ktor-network/jvm/test/io/ktor/network/sockets/tests/ClientSocketTest.kt
```diff
@@ -95,45 +99,35 @@ class ClientSocketTest {
...
-            val tcpSocketBuilder = aSocket(ActorSelectorManager(Dispatchers.IO)).tcp()
-            // Try to connect to that address repeatedly.
-            for (i in 0 until 100000) {
-                try {
-                    val socket = tcpSocketBuilder.connect(InetSocketAddress("127.0.0.1", port))
-                    fail("connect to self succeed: ${socket.localAddress} to ${socket.remoteAddress}")
-                } catch (ex: Exception) {
-                    // ignore
+            selector.buildOrClose({ openSocketChannel() }) {
+                val clientAddress = InetSocketAddress("0.0.0.0", 3333)
```
cy6erGn0m, line 103: The fixed port will introduce one more instability point

ktor-network/jvm/test/io/ktor/network/sockets/tests/ClientSocketTest.kt
```diff
@@ -95,45 +99,35 @@ class ClientSocketTest {
...
-                ServerSocket().apply {
-                    bind(InetSocketAddress("127.0.0.1", port + 1))
-                    close()
+                try {
+                    server { }
+                    nonBlocking()
+                    SocketImpl(
+                        this,
+                        clientSocket,
+                        ActorSelectorManager(Dispatchers.IO)
```
cy6erGn0m, line 121: This should be closed explicitly. I know we had it before, but would be nice to fix once you are changing the test

ktor-network/jvm/test/io/ktor/network/sockets/tests/ClientSocketTest.kt
```diff
@@ -95,45 +99,35 @@ class ClientSocketTest {
...
+                    ).connect(server!!.first.localSocketAddress)
+                } catch (e: ClosedChannelException) {
```
cy6erGn0m, line 123: `e` -> `cause` or `ignored` or even `_`

cy6erGn0m (changes requested):
The main question is why we stopped looking for a random port?

Ololoshechkin (changes requested):
Please, try using multiple ports but not a single one that may occasionally be taken.

Stexxe:
We could use some mocking framework to avoid writing boilerplate code.

### Note about the fix

There are two problems with the code that fixes the self-connection problem:
1. The author assumes that the client passes a socket created from the channel and therefore closes the socket instead of the channel but it's possible to pass any socket to the `SocketImpl` constructor. If the socket isn't bound to the channel and the self-connection condition is true then the `while` loop inside `connect` method will go forever.
2. If a client encounters a self-connection problem the `ClosedChannelException` will be thrown without the actual reason why that's happened.

cy6erGn0m (approved):

e5l (approved):

Ololoshechkin (approved):
Please, try using multiple ports but not a single one that may occasionally be taken.

PR #2358, opened 2021-02-16
Check if there are multiple unsafe headers in request

jltimm:
**Subsystem**
ktor-client

**Motivation**
If a user adds multiple unsafe headers to a request, then an UnsafeHeadersException will only be thrown on the first unsafe header it finds. The user will then remove that header from their request, try again, and then see another error with the other unsafe header that it finds.

**Solution**
All of the unsafe headers in a request should be identified and then returned to the user, rather than just the first one.


e5l (approved):
LGTM, @cy6erGn0m  what do you think?

cy6erGn0m:
Makes sense. The only disadvantage is that the exception class has a property for the header name but we are putting multiple names instead.

PR #2361, opened 2021-02-17
KTOR-2162: Fix JettyKtorHandler dispatcher grow problem.

spand:
**Subsystem**
Jetty engine

**Motivation**
From the issue:
JettyKtorHandler has the following code
```
    private val queue: BlockingQueue<Runnable> = LinkedBlockingQueue()
    private val executor = ThreadPoolExecutor(
        configuration.callGroupSize,
        configuration.callGroupSize * 8,
        THREAD_KEEP_ALIVE_TIME,
        TimeUnit.MINUTES,
        queue
    ) { r ->
        Thread(r, "ktor-jetty-$environmentName-${JettyKtorCounter.incrementAndGet()}")
    }
```
I believe the intention is to have an executor of size `configuration.callGroupSize`, burstable to `configuration.callGroupSize * 8` and then queue the tasks once that has been reached. However, this is not the case since `ThreadPoolExecutor` will only increase number of threads above core size if `queue.offer()` returns `false` which it never will since it is unbounded. See https://stackoverflow.com/a/19528305/422924 . However I am not a fan of the "solution" described.

**Solution**
I will instead suggest that queue is replaced with `SynchronousQueue<Runnable>()` which is what `Executors` uses in the non-fixed thread count cases. The new resulting behavior is the same except that, after growing to `configuration.callGroupSize * 8` threads (which is quite a fair amount) then instead of queuing tasks it will block and provide much needed back pressure to Jetty.

### DIFF
--- a/ktor-server/ktor-server-jetty/jvm/src/io/ktor/server/jetty/JettyKtorHandler.kt
+++ b/ktor-server/ktor-server-jetty/jvm/src/io/ktor/server/jetty/JettyKtorHandler.kt
@@ -33,13 +33,12 @@ internal class JettyKtorHandler(
     configuration: JettyApplicationEngineBase.Configuration
 ) : AbstractHandler(), CoroutineScope {
     private val environmentName = environment.connectors.joinToString("-") { it.port.toString() }
-    private val queue: BlockingQueue<Runnable> = LinkedBlockingQueue()
     private val executor = ThreadPoolExecutor(
         configuration.callGroupSize,
         configuration.callGroupSize * 8,
         THREAD_KEEP_ALIVE_TIME,
         TimeUnit.MINUTES,
-        queue
+        SynchronousQueue<Runnable>(),
     ) { r ->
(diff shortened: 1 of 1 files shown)


cy6erGn0m:
The only potential issue I can imagine is that blocking on overflow may cause coroutines to block and therefore get stuck. Unsure if it is dangerous or not in case of the queue overflow

spand:
> The only potential issue I can imagine is that blocking on overflow may cause coroutines to block and therefore get stuck. Unsure if it is dangerous or not in case of the queue overflow

Do you happen to know why it creates its own Dispatcher and not just use the default dispatcher?

cy6erGn0m:
We may try to replace it with the coroutine's experimental dispatcher instead. We definitely need to create a separate to provide the proper shutdown

spand:
Why is a separate dispatcher needed for shutdown ? Is it not enough to kill a scope ?

cy6erGn0m:
To ensure that no alive threads running after server shutdown. Unfortunately, we have Dispatchers.IO in few places so it's not true at the moment, but it should be so

PR #2365, opened 2021-02-20
KTOR-2143 Fixed CIO server wrong requested URL

LichtHund:
**Subsystem**
CIO Server

**Motivation**
CIO server would use wrong request URL, causing OAuth to return `redirect_uri_mismatch`. [KTOR-2143](https://youtrack.jetbrains.com/issue/KTOR-2143).

**Solution**
Reverse `host`'s get condition logic to be like `Netty` and `Jetty`.


Stexxe:
@ipsk please add a test.

rsinukov:
Will be fixed in #3122 3122

PR #2367, opened 2021-02-22
KTOR-778 Rewrite url builder

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-778

The current API for `UrlBuilder` does not allow using already encoded url parts.  To fix it, this PR adds properties or methods to set encoded or decoded url parts. Internally, `UrlBuilder` always holds properties in encoded format. 

Also, revert #1352 since this PR provides a better API for the same problem.

It's impossible to have a compatible API, because the current builder exposes its internal implementation as an API. That's why this PR goes to the 2.0.0 version.


ktor-client/ktor-client-mock/jvm/test/io/ktor/client/tests/mock/MockEngineTests.kt
```diff
@@ -38,9 +38,7 @@ class MockEngineTests {
 
         assertEquals(byteArrayOf(1, 2, 3).toList(), client.get<ByteArray>("/").toList())
         assertEquals("My Value", client.request<HttpResponse>("/").headers["X-MyHeader"])
-        assertEquals("Not Found other/path", client.get<String>("/other/path"))
-
-        Unit
+        assertEquals("Not Found /other/path", client.get("/other/path"))
```
e5l, line 41: Unrelated to current PR

e5l, line 41: Unrelated change

ktor-http/common/src/io/ktor/http/Query.kt
```diff
@@ -40,27 +40,42 @@ private fun ParametersBuilder.parse(query: String, startIndex: Int, limit: Int)
...
 
         if (spaceEndIndex > spaceNameIndex) {
-            val name = query.decodeURLQueryComponent(spaceNameIndex, spaceEndIndex)
+            val name = when {
+                decode -> query.decodeURLQueryComponent(spaceNameIndex, spaceEndIndex)
+                else -> query.substring(spaceNameIndex, spaceEndIndex)
+            }
             appendAll(name, emptyList())
         }
     } else {
```
e5l, line 64: I would put early return instead of else

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/http2/NettyHttp2Handler.kt
```diff
@@ -177,7 +172,11 @@ internal class NettyHttp2Handler(
         get() = javaClass.findIdField()
 
     private tailrec fun Class<*>.findIdField(): Field {
-        val f = try { getDeclaredField("id") } catch (t: NoSuchFieldException) { null }
+        val f = try {
```
e5l, line 175: consider using different name

ktor-server/ktor-server-tests/jvm/test/io/ktor/tests/server/http/URLBuilderTest.kt
```diff
@@ -156,7 +137,7 @@ class URLBuilderTest {
 
     @Test
     fun testParametersSpaceInParamName() {
-        assertEquals("http://localhost/?p1+space=v1", url { parameters.append("p1 space", "v1") })
+        assertEquals("http://localhost/?p1%20space=v1", url { parameters.append("p1 space", "v1") })
```
e5l, line 140: Could we also have a test with `+`?

ktor-http/common/src/io/ktor/http/URLBuilder.kt
```diff
@@ -240,27 +209,19 @@ public val URLBuilder.authority: String
...
     }
-    encodedPath += paths
-
-    return this
-}
-
-/**
- * Adds [components] to current [encodedPath]
- */
-public fun URLBuilder.pathComponents(vararg components: String): URLBuilder {
```
e5l, line 264: Could you tell me why we removing this method and do we have the migration?

e5l (changes requested):

ktor-http/common/src/io/ktor/http/Query.kt
rsinukov: This is the unchanged code. I don't want to do a clean up of the old code here to not bloat PR, since it's very big already

ktor-http/common/src/io/ktor/http/URLBuilder.kt
rsinukov: Added it back with deprecation in favor of new method

e5l (approved):

e5l:
Please check the CI build

.idea/codeStyles/Project.xml
```diff
@@ -6,15 +6,7 @@
...
-      <option name="CALL_PARAMETERS_WRAP" value="5" />
-      <option name="CALL_PARAMETERS_LPAREN_ON_NEXT_LINE" value="true" />
-      <option name="CALL_PARAMETERS_RPAREN_ON_NEXT_LINE" value="true" />
-      <option name="METHOD_PARAMETERS_WRAP" value="5" />
-      <option name="METHOD_PARAMETERS_LPAREN_ON_NEXT_LINE" value="true" />
-      <option name="METHOD_PARAMETERS_RPAREN_ON_NEXT_LINE" value="true" />
-      <option name="EXTENDS_LIST_WRAP" value="1" />
-      <option name="METHOD_CALL_CHAIN_WRAP" value="1" />
-      <option name="ASSIGNMENT_WRAP" value="1" />
+      <option name="CODE_STYLE_DEFAULTS" value="KOTLIN_OFFICIAL" />
```
cy6erGn0m, line 9: ?

rsinukov: thanks, removed

cy6erGn0m (changes requested):

ktor-http/api/ktor-http.api
```diff
@@ -915,11 +912,15 @@ public final class io/ktor/http/URLBuilderJvmKt {
 
 public final class io/ktor/http/URLBuilderKt {
 	public static final field DEFAULT_PORT I
+	public static final fun appendPathSegments (Lio/ktor/http/URLBuilder;Ljava/util/List;)Lio/ktor/http/URLBuilder;
+	public static final fun appendPathSegments (Lio/ktor/http/URLBuilder;[Ljava/lang/String;)Lio/ktor/http/URLBuilder;
 	public static final fun clone (Lio/ktor/http/URLBuilder;)Lio/ktor/http/URLBuilder;
 	public static final fun getAuthority (Lio/ktor/http/URLBuilder;)Ljava/lang/String;
-	public static final fun getAuthority (Lio/ktor/http/Url;)Ljava/lang/String;
```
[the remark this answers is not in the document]
rsinukov: It was moved to the `Url.kt` file

ktor-http/api/ktor-http.api
```diff
@@ -988,24 +990,17 @@ public final class io/ktor/http/UnsafeHeaderException : java/lang/IllegalArgumen
...
-	public final fun component1 ()Lio/ktor/http/URLProtocol;
-	public final fun component2 ()Ljava/lang/String;
-	public final fun component3 ()I
-	public final fun component4 ()Ljava/lang/String;
-	public final fun component5 ()Lio/ktor/http/Parameters;
-	public final fun component6 ()Ljava/lang/String;
-	public final fun component7 ()Ljava/lang/String;
-	public final fun component8 ()Ljava/lang/String;
-	public final fun component9 ()Z
-	public final fun copy (Lio/ktor/http/URLProtocol;Ljava/lang/String;ILjava/lang/String;Lio/ktor/http/Parameters;Ljava/lang/String;Ljava/lang/String;Ljava/lang/String;Z)Lio/ktor/http/Url;
```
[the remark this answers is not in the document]
rsinukov: Updated, keeping source compatibility

cy6erGn0m (approved):

TimothyEarley:
I recently updated to 1.6.1 (from 1.2.6) and was confused why the URL encoding broke for me. My URLs are already encoded and get passed to KTOR (`client.get<ByteArray>(url)`). Turns out https://github.com/ktorio/ktor/pull/2418 turned off the URL encoding which basically un-encoded my URLs. I added `this.url.parameters.urlEncodingOption = UrlEncodingOption.DEFAULT` to re-encode them which seems to work.

What is the new direction on this? Will URLs that are already encoded be kept encoded?
Thanks for your time and sorry if it is already answered somewhere in this thread.

rsinukov:
Hi @TimothyEarley,
This PR does exactly this, it keeps urls encoding when constructing them from strings. It also adds separate methods to compose urls from encoded or decoded parts, such as `appendParam`/`appendEncodedParam`, etc. 
#2418 was needed because it was the only possible way to solve issues with redirects while keeping the binary compatibility.

ymanse:
I was using UrlEncodingOption.NO_ENCODING in ktor 1.6, that feature is gone in this fix.

https://youtrack.jetbrains.com/issue/KTOR-553
https://github.com/ktorio/ktor/issues/1351

PR #2372, opened 2021-03-01
Unexpected exception when using Session feature: "Using blocking primitives on this dispatcher is not allowed"

Stexxe:
https://youtrack.jetbrains.com/issue/KTOR-1452

### DIFF
--- a/ktor-features/ktor-server-sessions/jvm/src/io/ktor/sessions/CacheStorage.kt
+++ b/ktor-features/ktor-server-sessions/jvm/src/io/ktor/sessions/CacheStorage.kt
@@ -6,13 +6,13 @@ package io.ktor.sessions
 
 import io.ktor.util.*
 import io.ktor.utils.io.*
-import io.ktor.utils.io.jvm.javaio.*
+import io.ktor.utils.io.core.*
 
 @Suppress("KDocMissingDocumentation")
 @InternalAPI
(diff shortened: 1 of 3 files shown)


ktor-features/ktor-server-sessions/jvm/src/io/ktor/sessions/CacheStorage.kt
```diff
@@ -6,13 +6,16 @@ package io.ktor.sessions
...
 import io.ktor.utils.io.*
-import io.ktor.utils.io.jvm.javaio.*
 
 @Suppress("KDocMissingDocumentation")
 @InternalAPI
 public class CacheStorage(public val delegate: SessionStorage, idleTimeout: Long) : SessionStorage {
     private val referenceCache = SoftReferenceCache<String, ByteArray> { id ->
-        delegate.read(id) { input -> input.toInputStream().readBytes() }
+        delegate.read(id) { input ->
+            val result = ByteArray(input.availableForRead)
```
cy6erGn0m, line 15: So we only read small amount of bytes instead of the full content. Available are bytes available for read at the moment, while the complete size could be unknown

cy6erGn0m (changes requested):

cy6erGn0m (approved):

PR #2373, opened 2021-03-01
KTOR-2194 Force ContentNegotiation to verify Accept and ContentType header compliance

rsinukov:
**Subsystem**
Server, ContentNegotiation

**Motivation**
[call.respond() will not check or apply ContentNegotiation for some types](https://youtrack.jetbrains.com/issue/KTOR-2194)

**Solution**
Add a flag to `ContentNegotiation` feature that will check accept header of response and content-type header of the request

### DIFF
--- a/ktor-server/ktor-server-core/jvm/src/io/ktor/features/ContentNegotiation.kt
+++ b/ktor-server/ktor-server-core/jvm/src/io/ktor/features/ContentNegotiation.kt
@@ -71,9 +71,11 @@ public class ContentNegotiation internal constructor(
             return true
         }
         return acceptItems.any {
-            it.contentType.contentType == "*"
-                || (it.contentType.contentType == contentType.contentType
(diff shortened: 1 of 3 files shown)


ktor-server/ktor-server-core/jvm/src/io/ktor/features/ContentNegotiation.kt
```diff
@@ -45,17 +45,49 @@ public data class ContentTypeWithQuality(val contentType: ContentType, val quali
...
+        if (!checkAcceptHeaderCompliance) {
+            return true
+        }
+        if (acceptItems.isEmpty()) {
+            return true
+        }
+        if (contentType == null) {
+            return true
+        }
+        return acceptItems.any {
```
cy6erGn0m, line 73: Isn't it `ContentType.match`?

rsinukov: thanks, didn't know about this one

e5l (approved):

PR #2380, opened 2021-03-04
Fix the high load generator

cy6erGn0m:
**Subsystem**
ktor server stress tests

**Motivation**
THis is fix for [KTOR-2081 NettyStressTest.testLongResponse](https://youtrack.jetbrains.com/issue/KTOR-2081) and [KTOR-2080 JettyStressTest.highLoadStressTest](https://youtrack.jetbrains.com/issue/KTOR-2080)

**Solution**
Fix the load generator to close channels and don't fail in the cases when it shouldn't just because of already closed channel.
The second fix is to introduce a delay between the high pressure and a regular request otherwise we get connection timeout.

### DIFF
--- a/ktor-server/ktor-server-test-host/api/ktor-server-test-host.api
+++ b/ktor-server/ktor-server-test-host/api/ktor-server-test-host.api
@@ -69,6 +69,7 @@ public final class io/ktor/server/testing/HighLoadHttpGenerator {
 	public static final fun main ([Ljava/lang/String;)V
 	public final fun mainLoop ()V
 	public final fun shutdown ()V
+	public final fun stat ()Ljava/lang/String;
 	public final fun stop ()V
 }
 
--- a/ktor-server/ktor-server-test-host/api/ktor-server-test-host.api
+++ b/ktor-server/ktor-server-test-host/api/ktor-server-test-host.api
@@ -69,7 +69,6 @@ public final class io/ktor/server/testing/HighLoadHttpGenerator {
 	public static final fun main ([Ljava/lang/String;)V
 	public final fun mainLoop ()V
 	public final fun shutdown ()V
-	public final fun stat ()Ljava/lang/String;
 	public final fun stop ()V
 }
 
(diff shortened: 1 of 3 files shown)


ktor-server/ktor-server-test-host/api/ktor-server-test-host.api
```diff
@@ -69,6 +69,7 @@ public final class io/ktor/server/testing/HighLoadHttpGenerator {
 	public static final fun main ([Ljava/lang/String;)V
 	public final fun mainLoop ()V
 	public final fun shutdown ()V
+	public final fun stat ()Ljava/lang/String;
```
rsinukov, line 72: should it go to the minor release then? or this module is not published?

ktor-server/ktor-server-test-suites/jvm/src/io/ktor/server/testing/suites/EngineStressSuite.kt
```diff
@@ -194,6 +198,8 @@ public abstract class EngineStressSuite<TEngine : ApplicationEngine, TConfigurat
 
         HighLoadHttpGenerator.doRun("/", "localhost", port, 8, 50, 10, true, gracefulMillis, timeMillis)
 
+        Thread.sleep(10000)
```
rsinukov, line 201: do we need it in other tests as well?

ktor-server/ktor-server-test-host/jvm/src/io/ktor/server/testing/HighLoadHttpGenerator.kt
```diff
@@ -449,14 +451,18 @@ public class HighLoadHttpGenerator(
                         }
                         connectionsCount++
                     } catch (t: Throwable) {
+                        ch.close()
                         connectErrors.incrementAndGet()
+                        connectFailureInRowCount++
```
rsinukov, line 456: should you set it back to zero on a successful connection?

rsinukov (changes requested):

rsinukov (approved):

PR #2386, opened 2021-03-09
Fix testThreadLeak

cy6erGn0m:
**Subsystem**
ktor-client-java

**Motivation**
[KTOR-2098 Fix flaky JavaEngineTests.testThreadLeak[jvm] ](https://youtrack.jetbrains.com/issue/KTOR-2098)

**Solution**
Filter threads more precisely by filtering them by name.

### DIFF
--- a/ktor-client/ktor-client-java/jvm/test/io/ktor/client/engine/java/JavaEngineTests.kt
+++ b/ktor-client/ktor-client-java/jvm/test/io/ktor/client/engine/java/JavaEngineTests.kt
@@ -7,6 +7,7 @@ package io.ktor.client.engine.java
 import io.ktor.client.*
 import io.ktor.client.request.*
 import kotlinx.coroutines.*
+import java.util.concurrent.*
 import kotlin.test.*
 
 class JavaEngineTests {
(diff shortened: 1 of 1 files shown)


ktor-client/ktor-client-java/jvm/test/io/ktor/client/engine/java/JavaEngineTests.kt
```diff
@@ -26,29 +27,38 @@ class JavaEngineTests {
 
         val initialNumberOfThreads = Thread.getAllStackTraces().size
         val repeats = 25
+        val executors = ArrayList<ExecutorService>()
```
rsinukov, line 30: `ArrayList` -> `mutableListOf`?

rsinukov (approved):

ktor-client/ktor-client-java/jvm/test/io/ktor/client/engine/java/JavaEngineTests.kt
cy6erGn0m: I actually see no point in using this function without arguments. What is the benefit?

PR #2350, opened 2021-02-10
KTOR-539 JS Set fetch parameters

hfhbd:
**Subsystem**
Client, JS Engine

**Motivation**
https://youtrack.jetbrains.com/issue/KTOR-539

**Solution**
Set the options for the JS fetch engine

PR #2388, opened 2021-03-10
KTOR-2155 Split JS module to node and browser

rsinukov:
**Subsystem**
Client

**Motivation**
https://youtrack.jetbrains.com/issue/KTOR-2155

**Solution**
Introduce two new modules and make old `client-js` dependant on them for backward compatibility


ktor-client/ktor-client-js-node/js/src/io/ktor/client/engine/js/Node.kt
```diff
@@ -0,0 +1,83 @@
...
+
+    override fun fetch(url: String): Promise<Response> = nodeFetch(url)
+
+    override fun fetch(url: String, init: RequestInit): Promise<Response> = nodeFetch(url, init)
+
+    override fun createAbortController(): AbortController = NodeAbortController()
+
+    override fun createHeaders(): Headers = NodeFetch.Headers()
+
+    override fun createWebSocket(urlString: String): WebSocket = NodeWebsocket(urlString)
```
hfhbd, line 29: To support Websockets with headers on nodeJS #2307 someday, this function should take an optional option parameter: `options: dynamic` as well. Alternative #2307 must provide this change.

ktor-client/ktor-client-js-node/js/src/io/ktor/client/engine/js/Node.kt
```diff
@@ -0,0 +1,83 @@
...
+import org.w3c.dom.*
+import org.w3c.fetch.*
+import kotlin.js.Promise
+
+/**
+ * Node JS API wrapper
+ */
+public object Node : JsPlatformApi {
+
+    override fun fetch(url: String): Promise<Response> = nodeFetch(url)
```
hfhbd, line 21: Do you need this function? Currently, only the second one with the `init: RequestInit` is called

ktor-client/ktor-client-core/js/src/io/ktor/client/engine/js/JsPlatformApi.kt
```diff
@@ -0,0 +1,31 @@
...
+import org.w3c.dom.*
+import org.w3c.fetch.*
+import kotlin.js.*
+
+/**
+ * Platform specific API for JS
+ */
+public interface JsPlatformApi {
+
+    public fun fetch(url: String): Promise<Response>
```
hfhbd, line 20: Do you need this function? Currently, only the second one with the `init: RequestInit` is called

ktor-client/ktor-client-js-browser/js/src/io/ktor/client/engine/js/Js.kt
```diff
@@ -0,0 +1,22 @@
+/*
+ * Copyright 2014-2021 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.client.engine.js
+
+import io.ktor.client.engine.*
+
+private val initHook = JsBrowser
```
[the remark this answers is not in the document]
rsinukov: This is used in tests and for running with default engine.

ktor-client/ktor-client-js-node/js/src/io/ktor/client/engine/js/Node.kt
rsinukov: I think it's better to have related change in #2307

ktor-client/ktor-client-core/js/src/io/ktor/client/engine/js/JsPlatformApi.kt
rsinukov: Since this is a public API, it's good to provide an API to fetch with the default request init args. Unfortunately, we can't use `dynamic` value in interfaces. That's why there are two methods.

hfhbd:
Hey @rsinukov, could you please rebase it? I would love to implement #2350 KTOR-539, and use it as a forked version to test the draft.

rsinukov:
Hi @hfhbd. It looks like we can't make it work until https://youtrack.jetbrains.com/issue/KT-44332 is fixed. Otherwise, default engine will not work for JS.

hfhbd:
@rsinukov When do you want to release 2.0.0? If https://youtrack.jetbrains.com/issue/KT-44332 will be still open, it would be possible to remove the common js module at all and provide only 2 modules. What do you think about this breaking change?

rsinukov:
@hhariri Moving the JS engine from the `core` module into two separate modules will break the automatic selection of default engine for JS., so we can't do it as well.

PR #2392, opened 2021-03-13
Fix for KTOR-2128.

sergeknystautas:
**Subsystem**
ktor-features in oauth2 parameters.

**Motivation**
This exposes extraParameters in OAuth2 access token from settings to allow providers who need this to supply refresh tokens.  See the KTOR-2128 for examples of Google and another provider of what extra parameters they need for refresh tokens towork.

**Solution**
There were already parameters being passed in but were unused by anything AFAICT, so felt ok changing the parameter type from List<Pair<String, String>> to Map<String, String> since that's how the URL builder wanted.

### DIFF
--- a/ktor-features/ktor-auth/jvm/src/io/ktor/auth/OAuth.kt
+++ b/ktor-features/ktor-auth/jvm/src/io/ktor/auth/OAuth.kt
@@ -109,6 +109,7 @@ public sealed class OAuthServerSettings(public val name: String, public val vers
      * @property nonceManager to be used to produce and verify nonce values
      * @property authorizeUrlInterceptor an interceptor function to customize authorization URL
      * @property accessTokenInterceptor an interceptor function to customize access token request
+     * @property extraParameters extra parameters to send provider during authentication
      */
     public class OAuth2ServerSettings(
         name: String,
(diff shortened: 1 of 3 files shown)


ktor-features/ktor-auth/jvm/src/io/ktor/auth/OAuth.kt
```diff
@@ -125,7 +126,8 @@ public sealed class OAuthServerSettings(public val name: String, public val vers
 
         public val authorizeUrlInterceptor: URLBuilder.() -> Unit = {},
         public val passParamsInURL: Boolean = false,
-        public val accessTokenInterceptor: HttpRequestBuilder.() -> Unit = {}
+        public val accessTokenInterceptor: HttpRequestBuilder.() -> Unit = {},
+        public val extraParameters: Map<String, String> = emptyMap()
```
kostrse, line 130: nit: I would add the `extraParameters` before `accessTokenInterceptor`. Not sure whether it would break existing users though.

TheBestPessimist:
Bumping this as it would be really nice to not have to add the interceptor for google oauth

e5l:
@rsinukov could you take a look?

PR #2393, opened 2021-03-14
KTOR-2311 Fix the assertion that the key exists in the key store

svilen-ivanov:
**Subsystem**
Server

**Motivation**
The assertion that the TLS key exists in the key store will trigger properly. Otherwise it fails with NPE somewhere in Netty pipepline during startup

**Solution**
Proper null check for absence of the key


rsinukov (approved):
Good catch, thanks!

PR #2396, opened 2021-03-19
Fix packet release

e5l:
Fixes [KTOR-960](https://youtrack.jetbrains.com/issue/KTOR-960)


ktor-io/common/src/io/ktor/utils/io/core/Buffers.kt
```diff
@@ -17,6 +17,8 @@ public expect class IoBuffer : Input, Output, ChunkBuffer {
     @Suppress("ConvertSecondaryConstructorToPrimary")
     public constructor(memory: Memory, origin: ChunkBuffer?)
```
rsinukov, line 18: Please deprecate old constructor so that we can remove it in future releases

ktor-io/common/src/io/ktor/utils/io/core/Buffers.kt
```diff
@@ -78,13 +80,18 @@ public fun Buffer.readBytes(count: Int = readRemaining): ByteArray {
 
 @Suppress("DEPRECATION")
 internal fun IoBuffer.releaseImpl(pool: ObjectPool<IoBuffer>) {
```
rsinukov, line 82: please create another method without param and deprecate this one

rsinukov (approved):
LGTM, with minor comments

ktor-io/common/src/io/ktor/utils/io/core/Buffers.kt
e5l: We have no public alternative available, it's possible to do only in 1.6.0

PR #2401, opened 2021-03-22
Io error message logging

arkanovicz:
**Subsystem**

Server

**Motivation**

Logging pipeline i/o errors stacktraces was too verbose (since they can often be due to client errors), and they have been filtered so that a single line error message is logged. But this line only states "io failed" whereas it should at least log the exception error message.

In my case, it was a permission problem on the server side (on the persistent sessions directory). It took me a painful debugging session to figure it out, whereas it's a trivial error that should be logged.

**Solution**

Have the engine log the i/o exception message.

### DIFF
--- a/ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/DefaultEnginePipeline.kt
+++ b/ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/DefaultEnginePipeline.kt
@@ -104,7 +104,7 @@ private fun ApplicationEnvironment.logFailure(call: ApplicationCall, cause: Thro
             is CancellationException -> log.info("$status: $logString, cancelled")
             is ClosedChannelException -> log.info("$status: $logString, channel closed")
             is ChannelIOException -> log.info("$status: $logString, channel failed")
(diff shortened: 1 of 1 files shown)


arkanovicz:
Note: I *could not* test this change (my test app keeps using the original `DefaultEnginePipeline.kt`), I posted a question on slack in the #intellij channel:

> Does intellij idea use some ktor classes internally? Whenever I try to build a custom ktor version, under a custom version name and on a private maven repo, there are some ktor source files (and corresponding modifications) that are never seen by idea (even after totally emptying local maven and gradle caches and disabling mavenLocal from build.gradle). See the attached picture for files in the `ktor-server-host-common-jvm/io.ktor.server.engine/` package : idea shows a mix of source and class files in the same package! And my modifications on `DefaultEnginePipeline` never reach the runtime, I don't understand where idea fetches the old version.

> (same behavior with ultimate 2020.1, ultimate 2020.3 and ultimate EAP 2021.1 Beta, each time refreshing indexes)

dependencies

e5l (approved):
Hey @arkanovicz, thanks for the PR! LGTM, will be merged after a CI run

PR #2403, opened 2021-03-27
ContentNegotiation feature for client

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-1689

Extract `ContentNegatiaion` feature into common module to share it between client and server.
* All serializers are shared now.
* `JsonFeature` was removed in favour of new client-side `ContentNegatitaion`
* Deleted depraceted API, related to this change


settings.gradle
```diff
@@ -101,3 +95,8 @@ include ':ktor-network:ktor-network-tls'
 include ':ktor-network:ktor-network-tls:ktor-network-tls-certificates'
 include ':ktor-bom'
 include ':ktor-test-dispatcher'
+include ':ktor-common'
```
e5l, line 98: `ktor-common` is really confusing name here. I can't even guess what should be there

e5l (changes requested):
The PR is really hard to review cause there are many changes inside:
- type info refactoring
- extraction of common in serialization
- utils refactoring

Could you split the PR to make changes gradually?

settings.gradle
rsinukov: This is for sharing code between server and client, and it's too small to be in a top level module. Do you have a better naming ideas?

rsinukov:
@e5l there were some not relevant changes after rebasing 2.0.0. Now this PR is smaller. I can try to separate it into different PRs, but since most of the changes are in tests and most of the new code is just moving of an old feature, it's hard to split to make it cosiderably smaller.

settings.gradle
e5l: Let's discuss how are we going to have such modules. I'm not sure that we should put all common between all features in a single common module. Maybe it's better to have ktor-serialization-common

e5l: btw we have a name conflict here: there are 2 ktor-serialization modules(cause all modules start being plain during publishing)

rsinukov:
@e5l I split this PR into three parts. Please review #2441, #2442 and #2443

PR #2407, opened 2021-03-31
Adding application startup time log

jgabrielfreitas:
**Subsystem**
Server - core

**Motivation**
I've been using Ktor in production for a few months and it is a simple thing that I miss.
I saw a issue (#2006) and that sounds good for me too.
There is a open pull requests from Sep/2020 (#2093) by @pavelgordon that looks like it has been left out, so I basically remade the solution, breaking it down into smaller variables and removing from `forEach` loop.

**Solution**
Calculating the time difference between the Engine `init` and receiving the `ApplicationStarted` event.

```
2021-03-30 21:47:20.370 [main] INFO  ktor.application - Responding at http://0.0.0.0:8090
2021-03-30 21:47:20.370 [main] INFO  ktor.application - Started application in 1.517 seconds
```
> example

--- 
PS: this is my first contribution to an open source project, so sorry for any inconvenience.

### DIFF
--- a/ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/BaseApplicationEngine.kt
+++ b/ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/BaseApplicationEngine.kt
@@ -29,6 +29,7 @@ public abstract class BaseApplicationEngine(
     public open class Configuration : ApplicationEngine.Configuration()
 
     init {
+        val initializedAt = System.currentTimeMillis()
         BaseApplicationResponse.setupSendPipeline(pipeline.sendPipeline)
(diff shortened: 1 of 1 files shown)


cy6erGn0m (changes requested):
If the application reloaded due to the hot-reload feature, the startup time will be equal to the total uptime instead of reload time.

Also, notice the codestyle errors

BOOMeranGG:
Hi! I want to get my first experience in open source. Nobody has been solving this problem for a long time, can I take it?

If so, I will immediately ask questions:
- Is there a need to display the time when auto-loading? (If so, is it correct to initialize the startInitializedAt variable when the ApplicationStarting event is received?)

PR #2410, opened 2021-04-06
Remove change that had wrong release set

rsinukov:
YT ticket had wrong release version set


hfhbd:
@rsinukov KTOR 940 as well :P

e5l (approved):

PR #2418, opened 2021-04-12
KTOR-2057 fix http redirect with encoded URL

rsinukov:
Disable query encoding/decoding when creating url from string

Proper fix is comming in 2.0.0 here #2367


e5l (approved):
lgtm

PR #2419, opened 2021-04-12
KTOR-2427 fix trailing slash toString

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-2427

### DIFF
--- a/ktor-server/ktor-server-core/jvm/src/io/ktor/routing/RouteSelector.kt
+++ b/ktor-server/ktor-server-core/jvm/src/io/ktor/routing/RouteSelector.kt
@@ -265,7 +265,7 @@ internal object TrailingSlashRouteSelector : RouteSelector(RouteSelectorEvaluati
         else -> RouteSelectorEvaluation.Failed
     }
 
-    override fun toString(): String = "/"
+    override fun toString(): String = "<trailing_slash>"
 }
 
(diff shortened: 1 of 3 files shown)


ktor-server/ktor-server-core/jvm/src/io/ktor/routing/RouteSelector.kt
```diff
@@ -265,7 +265,7 @@ internal object TrailingSlashRouteSelector : RouteSelector(RouteSelectorEvaluati
         else -> RouteSelectorEvaluation.Failed
     }
 
-    override fun toString(): String = "/"
+    override fun toString(): String = "<trailing_slash>"
```
cy6erGn0m, line 268: Not sure about it. What if someone already has a custom 'Route.toString' implementation based on RouteSelector.toString implementation? It may accidentally stop working

cy6erGn0m (changes requested):

ktor-server/ktor-server-core/jvm/src/io/ktor/routing/RouteSelector.kt
rsinukov: ok, reverted

cy6erGn0m (approved):

PR #2422, opened 2021-04-14
KTOR-1094 Drop obsolete base classes for packets

e5l:


ktor-io/jvm/src/io/ktor/utils/io/core/ByteBuffers.kt
```diff
@@ -154,31 +144,16 @@ internal fun AbstractInput.nioBuffer(size: Int): ByteBuffer? {
...
 @Deprecated("Binary compatibility.", level = DeprecationLevel.HIDDEN)
 internal fun ByteReadPacket.afterNioBufferUsed(read: Int) {
-    (this as ByteReadPacketBase).afterNioBufferUsed(read)
+    (this as AbstractInput).afterNioBufferUsed(read)
 }
 
 @PublishedApi
-@Suppress("unused", "DEPRECATION", "DEPRECATION_ERROR")
-@Deprecated("Binary compatibility.", level = DeprecationLevel.HIDDEN)
+@Suppress("DEPRECATION_ERROR")
```
cy6erGn0m, line 155: What error are we suppressing here? Isn't it redundant now?

cy6erGn0m (approved):

PR #2425, opened 2021-04-15
KTOR-2468 New routing resolve algorithm

rsinukov:
Clean up and make routing resolving algorithm smarter. Compare not only direct children's priorities, but the whole successful path.
Ex: 
Route 1:  path, q=1 -> missing parameter, q=0.2
Route 2: path, q=1 -> parameter, q=0.8
If both of these routes match, the old algorithm would choose route 1, even though it has lower quality overall. It happens because it compares only siblings' quality, and when they are similar, it will choose the first one based on order. The new algorithm will compare the whole paths.

### DIFF
--- a/ktor-server/ktor-server-core/jvm/src/io/ktor/routing/RouteSelector.kt
+++ b/ktor-server/ktor-server-core/jvm/src/io/ktor/routing/RouteSelector.kt
@@ -93,6 +93,12 @@ public data class RouteSelectorEvaluation(
             RouteSelectorEvaluation(true, RouteSelectorEvaluation.qualityConstant)
 
         /**
+         * Route evaluation succeeded for a transparent value
+         */
+        public val Transparent: RouteSelectorEvaluation =
+            RouteSelectorEvaluation(true, RouteSelectorEvaluation.qualityTransparent)
+
+        /**
          * Route evaluation succeeded for a single path segment with a constant value
          */
         public val ConstantPath: RouteSelectorEvaluation =
@@ -111,7 +117,13 @@ public data class RouteSelectorEvaluation(
  *
  * @param quality indicates how good this selector is compared to siblings
  */
-public abstract class RouteSelector(public val quality: Double) {
+public abstract class RouteSelector
+@Deprecated("quality property is not used anymore and will be removed", replaceWith = ReplaceWith("RouteSelector()"))
+constructor(@Deprecated("This property is not used anymore and will be removed") public val quality: Double) {
+
+    @Suppress("Deprecation")
+    public constructor() : this(0.0)
+
     /**
      * Evaluates this selector against [context] and a path segment at [segmentIndex]
      */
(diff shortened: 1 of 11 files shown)


ktor-server/ktor-server-core/jvm/src/io/ktor/routing/RouteSelector.kt
```diff
@@ -92,6 +92,12 @@ public data class RouteSelectorEvaluation(
         public val Constant: RouteSelectorEvaluation =
             RouteSelectorEvaluation(true, RouteSelectorEvaluation.qualityConstant)
 
+        /**
+         * Route evaluation succeeded for a transparent value
```
cy6erGn0m, line 96: This needs a clarification what exactly "transparent" means, an example or something

ktor-server/ktor-server-core/jvm/src/io/ktor/routing/RoutingResolve.kt
```diff
@@ -23,8 +23,16 @@ public sealed class RoutingResolveResult(public val route: Route) {
...
      */
-    public class Success(route: Route, override val parameters: Parameters) : RoutingResolveResult(route) {
-        override fun toString(): String = "SUCCESS${if (parameters.isEmpty()) "" else "; $parameters"} @ $route)"
+    public class Success internal constructor(
+        route: Route,
+        override val parameters: Parameters,
+        internal val quality: Double
+    ) : RoutingResolveResult(route) {
+
+        @Deprecated("Binary compatibility")
```
cy6erGn0m, line 32: Fix comment to describe that it will become internal and instantiation this class will be no longer allowed

ktor-server/ktor-server-core/jvm/src/io/ktor/routing/RoutingResolve.kt
```diff
@@ -103,126 +111,119 @@ public class RoutingResolveContext(
...
+                        return@maxWithOrNull compareValues(quality1, quality2)
+                    }
+                    index1++
+                    index2++
+                }
+                compareValues(result1.size, result2.size)
+            }!!
+
+        val parameters = bestPath
+            .fold(ParametersBuilder()) { acc, result -> acc.apply { appendAll(result.parameters) } }
```
cy6erGn0m, line 218:
```suggestion
            .fold(ParametersBuilder()) { builder, result -> builder.appendAll(result.parameters) }
```

cy6erGn0m (changes requested):
Looks good in general. We introduce a little bit more overhead because of extra allocations for lists but it seems to be difficult to eliminate.

e5l (changes requested):
Do we have the issue with the algorithm described first? I'm not sure that I can verify it directly in the source code. I would like to see the specification first

rsinukov:
@e5l The main issue is described in the PR description. It started to cause problems with the addition of a `TrailingSlashSelector` that can be in the middle of the route and matches almost everything with constant value.

e5l (approved):
LGTM as discussed, should we also make a topic in docs with an explanation of different cases?

rsinukov:
@e5l yes, I have a ticket to document the final algorithm

cy6erGn0m (approved):

PR #2428, opened 2021-04-20
KTOR-2512 Require forces webpack to resolve dependency

ilgonmic:
**Subsystem**
Client

**Motivation**
Even when it under `if` condition and standard Node.JS module, in webpack 4, it polyfill it into final bundle.
Since webpack 5, webpack not longer polyfill standard Node.JS modules, and report error.
So for WA this problem, need to use `eval('require')` instead of `require`

https://youtrack.jetbrains.com/issue/KT-46082

**Solution**
Webpack will not resolve `eval('require')('crypto')`, and it will be false branch of `if`


e5l:
Hey @ilgonmic, could you check the code-style failing on CI?

ilgonmic:
Ready

e5l (approved):

PR #2429, opened 2021-04-20
KTOR-1094 Merge abstract base with interfaces in IO

e5l:


ktor-io/jvm/src/io/ktor/utils/io/charsets/CharsetJVM.kt
```diff
@@ -63,7 +63,7 @@ public actual fun CharsetEncoder.encodeUTF8(input: ByteReadPacket, dst: Output)
 
     try {
         tmp.writeDirect(0) { tmpBb ->
-            val cb = tmpBb.asCharBuffer()!!
+            val cb = tmpBb.asCharBuffer()
```
cy6erGn0m, line 66: This actually makes difference as removing it causes compiled to add assert to every every variable access. This is why I do like this to make this check only once.

ktor-io/jvm/src/io/ktor/utils/io/charsets/CharsetJVM.kt
```diff
@@ -147,7 +147,7 @@ internal actual fun CharsetDecoder.decodeBuffer(
     var charactersCopied = 0
     input.readDirect { bb ->
         val tmpBuffer = ChunkBuffer.Pool.borrow()
-        val cb = tmpBuffer.memory.buffer.asCharBuffer()!!
+        val cb = tmpBuffer.memory.buffer.asCharBuffer()
```
cy6erGn0m, line 150: same here

ktor-io/jvm/src/io/ktor/utils/io/charsets/CharsetJVM.kt
```diff
@@ -339,5 +339,5 @@ actual constructor(message: String) : java.nio.charset.MalformedInputException(0
         get() = _message
 }
 
-private val EmptyCharBuffer = CharBuffer.allocate(0)!!
+private val EmptyCharBuffer = CharBuffer.allocate(0)
```
cy6erGn0m, line 342: same

cy6erGn0m (approved):
please check `!!` impact: most likely we get slowdown because of this

ktor-io/jvm/src/io/ktor/utils/io/charsets/CharsetJVM.kt
e5l: It marked with `@NotNull`, there is no platform types any more

PR #2432, opened 2021-04-21
KTOR-756 Fix exception swallow in html builder

e5l:
Fix [KTOR-756](https://youtrack.jetbrains.com/issue/KTOR-756)

### DIFF
--- a/ktor-features/ktor-html-builder/jvm/src/io/ktor/html/RespondHtml.kt
+++ b/ktor-features/ktor-html-builder/jvm/src/io/ktor/html/RespondHtml.kt
@@ -12,6 +12,7 @@ import io.ktor.util.cio.*
 import io.ktor.utils.io.*
 import kotlinx.html.*
 import kotlinx.html.stream.*
+import java.io.*
 
 /**
  * Responds to a client with a HTML response, using specified [block] to build an HTML page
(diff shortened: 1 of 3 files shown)


ktor-features/ktor-html-builder/jvm/src/io/ktor/html/RespondHtml.kt
```diff
@@ -32,9 +33,18 @@ public class HtmlContent(
...
-            it.appendHTML().html(block = builder)
+        val content = ByteArrayOutputStream()
+
+        try {
+            content.bufferedWriter().use {
+                it.append("<!DOCTYPE html>\n")
+                it.appendHTML().html(block = builder)
+            }
+        } catch (cause: Throwable) {
+            channel.close(cause)
```
cy6erGn0m, line 44: Don't we already do that on the call side of the `writeTo`?

cy6erGn0m (approved):

ktor-features/ktor-html-builder/jvm/src/io/ktor/html/RespondHtml.kt
e5l: Nope, but we probably should

PR #2433, opened 2021-04-21
KTOR-2516 fix ByteChannelSequential.readPacket freeze on closed channel

whyoleg:
Fix [KTOR-2516](https://youtrack.jetbrains.com/issue/KTOR-2516)


whyoleg:
@e5l Is it ok, that exception between platforms is different?
So in ByteChannelSequential we receive EOFException on readPacket/readFully if channel is closed without cause
But in ByteBufferChannel we receive ClosedReceiveChannelException
it was before like this, but should it be fixed, or at least create an issue?

ktor-io/common/test/io/ktor/utils/io/ByteChannelCloseTest.kt
```diff
@@ -0,0 +1,114 @@
...
+
+    private inline fun assertMayFail(block: () -> Unit) {
+        when (expectedExceptionClass) {
+            null -> block()
+            else -> assertFailsWith(expectedExceptionClass, block)
+        }
+    }
+
+    @Test
+    fun testImmediateFailurePropagation() = runTest {
```
rsinukov, line 42: This test and the ones below will run for every subclass. Maybe extract them to a separate class to avoid it?

rsinukov (changes requested):
Thanks for your contribution! In general LGTM, but I have couple of comments.

rsinukov:
@whyoleg good find. Yes, exceptions should be the same. We can't fix it until 2.0.0, because it will break semantics of the method. Please file an issue to not forget to fix it later.

whyoleg:
created https://youtrack.jetbrains.com/issue/KTOR-2562 - Align exceptions between ByteBufferChannel and ByteChannelSequential

ktor-io/common/test/io/ktor/utils/io/ByteChannelCloseTest.kt
whyoleg: That's what I want :)
Those tests are a little modified tests from removed ReadRemainingTest.kt. Class contains several ways on how channel can be closed, and I thought, that it will be good base to check, that behavior of `readRemaining`/`readPacket` after channel closing (with or without exception) is more or less aligned.

ktor-io/common/src/io/ktor/utils/io/ByteChannelSequential.kt
```diff
@@ -342,6 +342,13 @@ public abstract class ByteChannelSequentialBase(
         }
     }
 
+    private fun checkClosed(n: Int, closeable: BytePacketBuilder) {
+        if (closed) {
```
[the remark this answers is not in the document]
whyoleg: Ok, I will write test for it, and adapt method

ktor-io/common/test/io/ktor/utils/io/ByteChannelCloseTest.kt
rsinukov: Oh, I see it now.

rsinukov (approved):
LGTM
Will merge after CI runs

PR #2434, opened 2021-04-21
KTOR-2519 fix ByteBufferChannel.readRemaining suspend when should not

whyoleg:
Fix [KTOR-2519](https://youtrack.jetbrains.com/issue/KTOR-2519)


rsinukov (approved):
Thanks, looks good. I triggered the tests and will merge this PR after they run.

e5l (approved):

PR #2435, opened 2021-04-22
Corrects the port range check for Url.

Klurige:
The port should be a 16-bit positive int, which is 0..65535.

**Subsystem**
Urls

**Motivation**
The port number was required to be 1..65536, which misses the default port and allows for an illegal port number (65536 is 17 bits, should be limited to 16 bits).

**Solution**
Changes the range check to 0..65535


Thomas-Vos:
Actually the default port (0) was already allowed because the if statement also checks for `DEFAULT_PORT` (which is `0`). So this PR only fixes the illegal port number 65536.

Thomas-Vos:
So I think it is best to change the 0 back to 1 as port 0 is already checked separately.

Klurige:
> So I think it is best to change the 0 back to 1 as port 0 is already checked separately.

But then the error message should state "1 to 65535 _or_ DEFAULT_PORT if not set" if it should match the code. But it is possible to set it to 0. 
I would rather change the condition to only check the range, and remove the check for default port, as the default port is within the range.

PR #2441, opened 2021-04-29
KTOR-1689 Use TypeInfo in client and server request/response pipeline

rsinukov:
Pre requirement for `ContentNegotiation` client feature. 
To unify serializers API, they need to receive type in the same way.


ktor-client/ktor-client-core/common/src/io/ktor/client/request/RequestBody.kt
```diff
@@ -7,36 +7,25 @@ package io.ktor.client.request
...
-} catch (_: Throwable) {
-    null
-}
+internal val BodyTypeAttributeKey: AttributeKey<TypeInfo> = AttributeKey("BodyTypeAttributeKey")
 
 public inline fun <reified T> HttpRequestBuilder.setBody(body: T) {
     when (body) {
         null -> {
             this.body = EmptyContent
+            this.bodyType = null
```
e5l, line 20: obsolete `this`

ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/DefaultTransform.kt
```diff
@@ -44,7 +44,7 @@ public fun ApplicationReceivePipeline.installDefaultTransformations() {
     intercept(ApplicationReceivePipeline.Transform) { query ->
         val channel = query.value as? ByteReadChannel ?: return@intercept
 
-        val transformed: Any? = when (query.typeInfo.jvmErasure) {
+        val transformed: Any? = when (query.typeInfo.kotlinType!!.jvmErasure) {
```
e5l, line 47: maybe we can use `?` here

e5l (approved):

PR #2442, opened 2021-04-29
KTOR-1689 Extract ContentConverter to shared module

rsinukov:
Part 2 of `ContentNegotiation` client feature.
Extract all content converters to a shared module to reuse them in the client.

All new classes are extracted from the server module with minor changes in APIs.


e5l (approved):

e5l:
Could you please check `Allocations tests` if the difference in allocations is critical?

unoexperto:
@rsinukov Rustam, you mentioned today in https://youtrack.jetbrains.com/issue/KTOR-444 that it's fixed. Could you please show where you pass `KType` in `convertForSend` ?

It seems like `ContentConverter` is unchanged in
https://github.com/ktorio/ktor/blob/main/ktor-server/ktor-server-core/jvm/src/io/ktor/features/ContentNegotiation.kt#L255

rsinukov:
Hi @unoexperto! 
As you can see [here](https://github.com/ktorio/ktor/pull/2442/files#diff-2c71ca64dd886eeebc8a5378a453d33635be015b0ea8caab44f07f5b2f146c35R40) new methods `serialize` and `deserialize` have parameter with type `TypeInfo`, that contains `KType`. Indeed, it's not fixed in the `main` branch, because it's a breaking change and will be released in 2.0.0.

PR #2448, opened 2021-05-10
KTOR-367 Deprecate uninstallFeature method

rsinukov:
**Subsystem**
Server

**Motivation**
This method is misleading. It only closes open resources, but doesn't remove feature interceptors.
Right now, there is no way to properly uninstall features, but also no clear use case for that.


cy6erGn0m (approved):
Generally, it's fine but the text.. I would prefer to point to an existing ticket instead. Could you please open a design ticket describing the problem and refer to it in the deprecation text?

PR #2454, opened 2021-05-12
Deprecate KtorExperimentalAPI

cy6erGn0m:
**Subsystem**
All

**Motivation**
[KTOR-2204  Deprecate old experimental annotations ](https://youtrack.jetbrains.com/issue/KTOR-2204)

**Solution**
Stabilize and split into new separate experimental annotations, open corresponding tickets, refer to the tickets from the new annotations.


e5l (approved):
thanks for the PR, lgtm

PR #2458, opened 2021-05-13
KTOR-1094 Add missing deprecations for 2.0.0

e5l:
Fixes [KTOR-1094](http://youtrack.jetbrains.com/issue/KTOR-1094)

### DIFF
--- a/ktor-io/common/src/io/ktor/utils/io/core/AbstractInput.kt
+++ b/ktor-io/common/src/io/ktor/utils/io/core/AbstractInput.kt
@@ -11,6 +11,11 @@ import io.ktor.utils.io.pool.*
  * The default abstract base class implementing [Input] interface.
  * @see [AbstractInput.fill] and [AbstractInput.closeSource].
  */
+@Deprecated(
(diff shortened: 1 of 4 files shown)


ktor-io/common/src/io/ktor/utils/io/core/AbstractInput.kt
```diff
@@ -6,11 +6,40 @@ import io.ktor.utils.io.bits.*
 import io.ktor.utils.io.core.internal.*
 import io.ktor.utils.io.core.internal.require
 import io.ktor.utils.io.pool.*
+import kotlin.Boolean
```
rsinukov, line 9: is it intended?

rsinukov (approved):

ktor-io/common/src/io/ktor/utils/io/core/AbstractInput.kt
e5l: nope, I guess it's auto import issue

e5l: thanks for the notice

PR #2466, opened 2021-05-19
KTOR-2709 Use kotlin.reflect.javaType instead of type token

cliffred:
**Subsystem**
JVM Utils

**Motivation**
Fix https://youtrack.jetbrains.com/issue/KTOR-2709
Using kotlin.reflect.javaType doesn't produce an anonymous class for each call site. Also, this prevents Ktor from being affected by this compiler bug: https://youtrack.jetbrains.com/issue/KT-46797.

**Solution**
Use kotlin.reflect.javaType instead of type token pattern.

### DIFF
--- a/ktor-utils/jvm/src/io/ktor/util/reflect/TypeInfoJvm.kt
+++ b/ktor-utils/jvm/src/io/ktor/util/reflect/TypeInfoJvm.kt
@@ -8,6 +8,10 @@ import kotlin.reflect.*
 
 public actual typealias Type = java.lang.reflect.Type
 
+@PublishedApi
+@Deprecated("Will be removed")
+internal open class TypeBase<T>
+
 @OptIn(ExperimentalStdlibApi::class)
 public actual inline fun <reified T> typeInfo(): TypeInfo {
     val kType = typeOf<T>()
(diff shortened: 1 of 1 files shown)


rsinukov:
Hi, thanks for the PR!
I will run the tests to see if we are not affected by any corner cases.

ktor-utils/jvm/src/io/ktor/util/reflect/TypeInfoJvm.kt
```diff
@@ -4,22 +4,15 @@
...
 package io.ktor.util.reflect
 
-import java.lang.reflect.*
-import java.lang.reflect.Type
 import kotlin.reflect.*
 
 public actual typealias Type = java.lang.reflect.Type
 
-@PublishedApi
-internal open class TypeBase<T>
```
rsinukov, line 14: Since this class is annotated with `@PublishedApi`, it's part of the public API and can't be removed in minor release. Please keep it, but add a deprecation to it.

rsinukov (changes requested):
LGTM in general and tests look good, but small comment

rsinukov (approved):
LGTM

PR #2467, opened 2021-05-20
KTOR-737 Reply with proper status code on routing mismatch

rsinukov:
[Should return 405 when route exists but not for given method instead of 404](https://youtrack.jetbrains.com/issue/KTOR-737)

* Make `RouteSelectorEvaluation` a sealed class with two children: `Success` and `Failure`
* `Failure` has `errorStatusCode` property to reply if the route doesn't match. For example, the `HttpMethodSelector` status code is 405
* The routing resolve algorithm keeps track of the best failed route and uses it to respond if no matches are found.

### DIFF
--- a/ktor-features/ktor-auth/jvm/src/io/ktor/auth/Authentication.kt
+++ b/ktor-features/ktor-auth/jvm/src/io/ktor/auth/Authentication.kt
@@ -327,7 +327,7 @@ public fun Route.authenticate(
  */
 public class AuthenticationRouteSelector(public val names: List<String?>) : RouteSelector() {
     override fun evaluate(context: RoutingResolveContext, segmentIndex: Int): RouteSelectorEvaluation {
-        return RouteSelectorEvaluation(true, RouteSelectorEvaluation.qualityTransparent)
+        return RouteSelectorEvaluation.Transparent
     }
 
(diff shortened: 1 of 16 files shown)


ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/BodyProgressTest.kt
```diff
@@ -180,7 +180,7 @@ class BodyProgressTest : ClientLoader() {
             client.preparePost("$TEST_SERVER/content/echo") {
                 body = channel
                 onDownload(listener)
-            }.body<ByteReadChannel>()
+            }.body<ByteReadChannel, Unit> { it.readRemaining() }
```
e5l, line 183: could you tell me why do we need second type here?

e5l (changes requested):

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/BodyProgressTest.kt
rsinukov: The first type is for parameters, the second is for the return type of a block. It's an existing API and it wasn't changed in this PR.

e5l (approved):

PR #2469, opened 2021-05-20
KTOR-2712 deprecate TestApplicationCall::requestHandled property

rsinukov:
And replace all usages in tests

https://youtrack.jetbrains.com/issue/KTOR-2712


ktor-server/ktor-server-tests/jvm/test/io/ktor/tests/server/routing/RoutingProcessingTest.kt
```diff
@@ -44,7 +41,8 @@ class RoutingProcessingTest {
                 method = HttpMethod.Post
             }
             it("should not be handled") {
-                assertFalse(result.requestHandled)
+                val responseHandled = result.response.status()?.isSuccess() ?: false
```
e5l, line 44: maybe we can introduce the extension: there are lots of similar checks

e5l (approved):
lgtm

PR #2479, opened 2021-05-26
KTOR-400 Fix hanging upload observer

rsinukov:
* Fix `InvalidMutabilityException` on observing `WriteChannelContent`
* FIx hanging of observer
* Add more tests


e5l (approved):

e5l:
Please check public API and docs

PR #2487, opened 2021-05-28
KTOR-379 Add Host header verification

e5l:
Fix [KTOR-379](https://youtrack.jetbrains.com/issue/KTOR-379)

### DIFF
--- a/ktor-server/ktor-server-test-suites/jvm/src/io/ktor/server/testing/suites/SustainabilityTestSuite.kt
+++ b/ktor-server/ktor-server-test-suites/jvm/src/io/ktor/server/testing/suites/SustainabilityTestSuite.kt
@@ -800,4 +800,34 @@ abstract class SustainabilityTestSuite<TEngine : ApplicationEngine, TConfigurati
             assertEquals(10000 * 13L, result.second)
         }
     }
+
+    @Test
+    fun testDoubleHost() {
+        createAndStartServer {
+            get("/") {
+                call.respond("OK")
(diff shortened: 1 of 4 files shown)


ktor-server/ktor-server-test-suites/jvm/src/io/ktor/server/testing/suites/SustainabilityTestSuite.kt
```diff
@@ -800,4 +805,34 @@ abstract class SustainabilityTestSuite<TEngine : ApplicationEngine, TConfigurati
...
+            outputStream.bufferedWriter().apply {
+                write(content)
+                flush()
+            }
+
+            val response = inputStream.bufferedReader()
+            val status = response.readLine()
+
+            assertTrue(status.startsWith("HTTP/1.1 400"))
+            outputStream.close()
```
rsinukov, line 835: I know that it doesn't matter in this case, but it's a bit painful to see that `inputStream` is not closed =)

rsinukov (approved):
LGTM

ktor-server/ktor-server-test-suites/jvm/src/io/ktor/server/testing/suites/SustainabilityTestSuite.kt
e5l: Socket is already closed at that moment, so `inputStream.close()` will fail with an exception

PR #2488, opened 2021-05-28
fix youtrack links

andreyaksenov:
Added `https://` to Youtrack links since GitHub adds `https://github.com/ktorio/ktor/blob/main/` before the address and links lead to 404.


hhariri (approved):
Thanks. Not sure how I missed this :(

PR #2494, opened 2021-06-02
KTOR-2626 Remove internal require function

hfhbd:
**Subsystem**
Server, IO Internal

**Motivation**
https://youtrack.jetbrains.com/issue/KTOR-2626

**Solution**
Switch to stdlib require function


rsinukov (changes requested):
Please commit `apiDump` result

hfhbd:
@rsinukov Okay, but removing a internal functions won't change the api, BUT there are a lot of missing apiDumps from other changes..

rsinukov:
Since this function was annotated with `@PublishedApi` it was effectively public, so it was part of binary API

hfhbd:
@rsinukov Yes, but other`@PublishedApi` annotated functions are also not printed in the .api file as well (https://github.com/ktorio/ktor/blob/9c902b3c001f40bafe744d8d27573487d54c670c/ktor-features/ktor-auth/jvm/src/io/ktor/auth/SessionAuth.kt#L30). Don't know if it is a wrong configuration or a bug in the validator itself.

rsinukov:
@hfhbd hm, thanks for pointing out, they should be in the dump. 
In the dump that you published there are things that were deleted in 2.0.0. Can you do apiDump with clean build, or even deleted `.gradle` folder? Somtimes it caches too much and you have dumps from different branches.

hfhbd:
@rsinukov Done :D

rsinukov (approved):

hfhbd:
So, for some reasons `apiDump` does not remove the included lines https://github.com/ktorio/ktor/blob/ec9bd2fada9b149f69b758bd1eb474e7ef3a6e01/ktor-io/api/ktor-io.api#L1753

a) I remove the files manually.
b) Fix the binary validator settings, if possible. I did remove all custom settings regarding the plugin, update it to `0.5.0`, but it still did not delete the lines.

rsinukov:
Please rebase and I will merge it. I'll fix apiDump later myself.

hfhbd:
@rsinukov rebased

PR #2497, opened 2021-06-04
KTOR-2318 add keys function to ApplicationConfig

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-2318


e5l (approved):
Good job, LGTM!

PR #2501, opened 2021-06-08
KTOR-333 Fix CIO hanging when `call.respond()` was not called

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-333

### DIFF
--- a/ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/BaseApplicationEngine.kt
+++ b/ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/BaseApplicationEngine.kt
@@ -6,7 +6,9 @@ package io.ktor.server.engine
 
 import io.ktor.application.*
 import io.ktor.http.*
+import io.ktor.http.content.*
 import io.ktor.response.*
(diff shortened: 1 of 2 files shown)


ktor-server/ktor-server-cio/jvm/src/io/ktor/server/cio/CIOApplicationResponse.kt
```diff
@@ -31,6 +31,9 @@ internal class CIOApplicationResponse(
     @Volatile
     private var chunkedJob: Job? = null
 
+    @Volatile
+    private var sent: Boolean = false
```
e5l, line 35: Could you tell me why the `responded` flag from BaseApplicationEngine can't be reused?

e5l (changes requested):

ktor-server/ktor-server-cio/jvm/src/io/ktor/server/cio/CIOApplicationResponse.kt
rsinukov: reworked solution

e5l (approved):

PR #2504, opened 2021-06-10
KTOR-2797 Fix client request stuck in refreshToken

e5l:
Fix [KTOR-2797](https://youtrack.jetbrains.com/issue/KTOR-2797)


ktor-client/ktor-client-features/ktor-client-auth/common/src/io/ktor/client/features/auth/providers/AuthTokenHolder.kt
```diff
@@ -5,47 +5,31 @@
...
-
-        try {
-            val token = block()
-            deferred.complete(token)
-            return token
-        } catch (cause: Throwable) {
-            deferred.completeExceptionally(cause)
-            throw cause
-        }
+        val token = block()
```
rsinukov, line 31: If we removed blocking here, then `testSetTokenCalledOnce` would fail.

rsinukov (changes requested):

ktor-client/ktor-client-features/ktor-client-auth/common/src/io/ktor/client/features/auth/providers/AuthTokenHolder.kt
e5l: Discussed, agreed to drop the test and approve the behaviour change

PR #2506, opened 2021-06-11
KTOR-2691 Fix ConnectException message, change level to debug

e5l:
Fix [KTOR-2691](https://youtrack.jetbrains.com/issue/KTOR-2691)

### DIFF
--- a/ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/DefaultEnginePipeline.kt
+++ b/ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/DefaultEnginePipeline.kt
@@ -102,15 +102,15 @@ private fun ApplicationEnvironment.logFailure(call: ApplicationCall, cause: Thro
             "(request error: $cause)"
         }
 
+        val infoString = "$status: $logString. Exception ${cause::class}: ${cause.message}]"
         when (cause) {
-            is CancellationException -> log.debug("$status: $logString, cancelled")
-            is ClosedChannelException -> log.debug("$status: $logString, channel closed")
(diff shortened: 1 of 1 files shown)


orangy (commented):
Why don't you just handle base IOException, but include the type of the exception in the message? Currently you are missing messages in some cases.

ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/DefaultEnginePipeline.kt
```diff
@@ -101,10 +103,11 @@ private fun ApplicationEnvironment.logFailure(call: ApplicationCall, cause: Thro
...
         when (cause) {
-            is CancellationException -> log.info("$status: $logString, cancelled")
-            is ClosedChannelException -> log.info("$status: $logString, channel closed")
-            is ChannelIOException -> log.info("$status: $logString, channel failed")
-            is IOException -> log.info("$status: $logString, io failed: ${cause.message ?: "unknown error"}")
+            is CancellationException -> log.debug("$status: $logString, cancelled")
+            is ClosedChannelException -> log.debug("$status: $logString, channel closed")
+            is ChannelIOException -> log.debug("$status: $logString, channel failed")
+            is ConnectException -> log.debug("$status: $logString, connection failed with: ${cause.message}")
+            is IOException -> log.debug("$status: $logString, io failed with ${cause.message ?: "unknown error"}")
```
rsinukov, line 110: Maybe add an exception class to the message as well?

rsinukov (approved):
LGTM, but minor comment

PR #2510, opened 2021-06-14
KTOR-686 Add local network address function to engine

rsinukov:
Fixes https://youtrack.jetbrains.com/issue/KTOR-686

### DIFF
--- a/ktor-server/ktor-server-cio/jvm/src/io/ktor/server/cio/CIOApplicationEngine.kt
+++ b/ktor-server/ktor-server-cio/jvm/src/io/ktor/server/cio/CIOApplicationEngine.kt
@@ -9,6 +9,7 @@ import io.ktor.http.cio.*
 import io.ktor.server.cio.backend.*
 import io.ktor.server.engine.*
 import io.ktor.util.*
+import io.ktor.util.network.*
 import io.ktor.util.pipeline.*
 import kotlinx.coroutines.*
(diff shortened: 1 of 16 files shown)


ktor-server/ktor-server-cio/jvm/src/io/ktor/server/cio/CIOApplicationEngine.kt
```diff
@@ -48,6 +49,8 @@ public class CIOApplicationEngine(
     private val startupJob: CompletableDeferred<Unit> = CompletableDeferred()
     private val stopRequest: CompletableJob = Job()
 
+    override val networkAddresses: Deferred<List<NetworkAddress>> = CompletableDeferred()
```
e5l, line 52: I would add the `suspend` method instead: deferred expose `cancel` and other public calls

e5l (changes requested):

ktor-server/ktor-server-cio/jvm/src/io/ktor/server/cio/CIOApplicationEngine.kt
rsinukov: good point, updated

e5l (approved):
Thanks for the PR, lgtm

e5l:
Please also check the CI

Issue #1833, opened 2020-04-28
parameterOf() should have a variant that takes in a Map<String, List<String>>


rock3r:
**Subsystem**
Server

**Is your feature request related to a problem? Please describe.**
There is currently no way to build a Parameters instance from a Map, even though the internal representation of the values is a map. `Parameters` can only be constructed using:
 * `parametersOf()`
 * `parametersOf(name: String, value: String)`
 * `parametersOf(name: String, values: List<String>)`
 * `parametersOf(vararg pairs: Pair<String, List<String>>)`

`ParametersImpl` is internal and cannot be constructed directly.

**Describe the solution you'd like**
I'd like there to be a `parametersOf(parameters: Map<String, List<String>>)`:

```kotlin
fun parametersOf(vararg parameters: Map<String, List<String>>): Parameters =
    ParametersImpl(parameters)
```

**Motivation to include to ktor**
Seems like it should be there. There are many cases where you're already working with a parameters map, and having to transform that into a list of pairs just to have it retransformed back under the hood seems wasteful.

oleg-larshin:
Please check the <a href="https://youtrack.jetbrains.com/issue/KTOR-399">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

PR #2515, opened 2021-06-15
Ktor-399 Add a new variant of parametersOf() that takes Map<String, List<String>>

BOOMeranGG:
**Subsystem**
Core

**Solution**
Implementation was suggested in issue #1833, but not implemented in the project


BOOMeranGG:
How can I understand why the code style verify failed? Can I run it locally?

rsinukov:
Hi @BOOMeranGG.
Codestyle is broken in `main` branch, fix is comming in #2517.

ktor-http/common/src/io/ktor/http/Parameters.kt
```diff
@@ -85,6 +85,11 @@ public fun parametersOf(name: String, values: List<String>): Parameters = Parame
  */
 public fun parametersOf(vararg pairs: Pair<String, List<String>>): Parameters = ParametersImpl(pairs.asList().toMap())
 
+/**
+ * Creates a parameters instance from the specified map [parameters]
+ */
+public fun parametersOf(parameters: Map<String, List<String>>): Parameters = ParametersImpl(parameters)
```
rsinukov, line 91: I think we should have defensive copy here. Someone can pass `MutableMap` here and then change its content from outside.

rsinukov (changes requested):
Thanks for your PR.
We can't add any new API in patch releases. The next release with the new API will be 2.0.0. Can you please rebase to `2.0.0-eap` branch, run `./gradlew apiDump` locally and commit results? This will update public API.

BOOMeranGG:
I'm not sure I did the right thing. If this is not what is needed, please write what was expected

rsinukov:
You need to apply your changes on top of the `2.0.0-eap` branch. Basically, do this
```
git fetch
git rebase -i origin/2.0.0-eap
<remove all the commits that are not relevant to this change by replacing `pick` with `drop` (or `d`)>
git rebase --continue
./gradlew apiDump
git push -f
```

PR #2516, opened 2021-06-15
KTOR-343 Fix exception in template can't be handled in StatusPages

rsinukov:
`StatusPages` can't handle errors that happen in writing bodies, because response was already sent at this point. To fix this, move all the templates rendering before responding.


e5l (approved):
The problem is fixed, so lgtm. Please note, by changing the semantic from channel to array we can hide the origin problem with the exception handling.

PR #2518, opened 2021-06-15
KTOR-2788 Add support for X-Forwarded-Port

rsinukov:
Added without public API changes. A proper config will be available in the next minor release.


e5l (approved):
lgtm

PR #2522, opened 2021-06-16
KTOR-2822 Change `HttpReceivePipeline` context to Unit

rsinukov:
Otherwise, it's impossible to modify response, since context has its own instance of response.


e5l (changes requested):
please check compilation failures

e5l (changes requested):
It looks like the JVM `numberTest` is failing without flaky here. Could you take a look?

rsinukov:
@e5l thanks, fixed. I decreased the test timeout for debugging purposes and forgot to change it back.

e5l (approved):
Please check if we need to make documentation for that and make new ticket if we need

PR #2523, opened 2021-06-16
KTOR-1345 Default values in URLBuilder

hfhbd:
**Subsystem**
Core

**Motivation**
https://youtrack.jetbrains.com/issue/KTOR-1345

**Use case**
BaseURL feature: https://youtrack.jetbrains.com/issue/KTOR-730 (#2212)

```kotlin
// current hack
private fun URLBuilder.isRelativeURL(): Boolean {
  val defaultURLBuilder = URLBuilder()
  val isDefault = this == defaultURLBuilder
  val isRelativeURL = protocol == defaultURLBuilder.protocol && authority == defaultURLBuilder.authority
  return !isDefault && isRelativeURL
}

// with nullable host when not set
private fun URLBuilder.isRelativeURL(): Boolean {
  return host == null
}

```


ktor-http/common/src/io/ktor/http/URLBuilder.kt
```diff
@@ -23,15 +23,15 @@ public const val DEFAULT_PORT: Int = 0
...
-    user: String? = null,
-    password: String? = null,
-    pathSegments: List<String> = mutableListOf(""),
-    parameters: Parameters = Parameters.Empty,
-    fragment: String = "",
-    public var trailingQuery: Boolean = false
+    public var protocol: URLProtocol? = null,
+    public var host: String? = null,
+    public var port: Int? = null,
+    public var user: String? = null,
```
rsinukov, line 29: It's essential that these parameters are not properties. Otherwise encode/decode process gets broken.

ktor-http/common/src/io/ktor/http/URLBuilder.kt
```diff
@@ -86,33 +86,41 @@ public class URLBuilder(
...
-        specifiedPort = port,
-        pathSegments = pathSegments,
-        parameters = parameters.build(),
-        fragment = fragment,
-        user = user,
-        password = password,
-        trailingQuery = trailingQuery,
-        urlString = buildString()
+    public fun build(
+        defaultProtocol: URLProtocol = UrlProtocol.Http,
```
rsinukov, line 93: This looks very counterintuitive that the `build` function accepts parameters, especially the ones that should be default. What does the usage of it look like?

ktor-http/common/src/io/ktor/http/Url.kt
```diff
@@ -27,14 +27,15 @@ public class Url internal constructor(
     public val fragment: String,
     public val user: String?,
     public val password: String?,
-    public val trailingQuery: Boolean,
-    private val urlString: String
+    public val trailingQuery: Boolean
```
rsinukov, line 30: `urlString` should be a separate parameter and it can't be created from the rest, because of encoding/decoding.

rsinukov (changes requested):

ktor-http/common/src/io/ktor/http/URLBuilder.kt
hfhbd: My idea was to prevent hardcoding the values, but it looks strange, thats right. Honestly, I can't imagining a usage with given parameters. I will move them into the build function body.

ktor-http/common/src/io/ktor/http/URLBuilder.kt
```diff
@@ -23,15 +23,15 @@ public const val DEFAULT_PORT: Int = 0
...
+    public var protocol: URLProtocol? = null,
+    public var host: String? = null,
+    public var port: Int? = null,
     user: String? = null,
     password: String? = null,
-    pathSegments: List<String> = mutableListOf(""),
-    parameters: Parameters = Parameters.Empty,
-    fragment: String = "",
-    public var trailingQuery: Boolean = false
+    pathSegments: List<String>? = null,
```
rsinukov, line 31: This will make usage of builder harder, but probably there is no better way to do separate default value vs non-set value.
Do you want to proceed with this PR? Otherwise I can take it.

hfhbd: @rsinukov You are free to take it :D

e5l:
The alternative fix is merged in 2.0.0-eap: https://github.com/ktorio/ktor/commit/373d33f760a93d9b83f9841f12f8cda02b469d1f

PR #2527, opened 2021-06-22
KTOR-2360 Extract plugins from server module

rsinukov:


e5l (approved):
@Ololoshechkin, could you tell me if we can have an assist for package renaming?

unoexperto:
@rsinukov @e5l Folks, is there a way for me to be up to date on 2.0 design changes ? Perhaps there is some epic ticket, design document or even Telegram chat I can be part of ?

PR #2531, opened 2021-06-25
crypto is undefined in IE11

Stexxe:
https://youtrack.jetbrains.com/issue/KTOR-409


ktor-utils/js/test/io.ktor.util/CryptoTest.kt
```diff
@@ -0,0 +1,20 @@
...
+ */
+
+package io.ktor.util
+
+import kotlin.test.*
+
+class CryptoTest {
+
+    @Test
+    fun expectCryptoToBeDefinedInIE11() {
```
rsinukov, line 12: nit: The test name looks disconnected from what the test does

rsinukov (approved):

PR #2532, opened 2021-06-28
KTOR-2555 Drop kotlin-test-junit dependency

e5l:
Fix [KTOR-2555](https://youtrack.jetbrains.com/issue/KTOR-2555)


ktor-server/ktor-server-test-host/jvm/src/io/ktor/server/testing/Utils.kt
```diff
@@ -4,35 +4,48 @@
...
  * Return parsed content type from the test response
  */
-public fun TestApplicationResponse.contentType(): ContentType {
+fun TestApplicationResponse.contentType(): ContentType {
     val contentTypeHeader = requireNotNull(headers[HttpHeaders.ContentType])
     return ContentType.parse(contentTypeHeader)
 }
+
+
+inline fun <reified T : Throwable> assertFailsWith(block: () -> Unit) {
```
ilya-g, line 41: Is it a public function? 
It could create confusion among the users of ktor-server-test-host who also use kotlin-test.

ilya-g (commented):
So, what is the intention here? Does this change makes it easier for users of `ktor-server-test-host` to use different testing frameworks in their projects?

ktor-server/ktor-server-test-host/jvm/src/io/ktor/server/testing/Utils.kt
e5l: sure, let me move it to the use places and make private

ktor-io/common/src/io/ktor/utils/io/core/Use.kt
```diff
@@ -1,17 +0,0 @@
-package io.ktor.utils.io.core
-
-public inline fun <I : Input, R> I.use(block: (I) -> R): R {
```
rsinukov, line 3: isn't it breaks API?

ktor-io/jvm/test/io/ktor/utils/io/BlockingAdaptersOnProhibitedThreadTest.kt
```diff
@@ -13,6 +13,7 @@ class BlockingAdaptersOnProhibitedThreadTest {
 
     @AfterTest
     fun cleanup() {
+        fail()
```
rsinukov, line 16: Can you explain why you fail tests here?

rsinukov (changes requested):

ktor-io/common/src/io/ktor/utils/io/core/Use.kt
e5l: Nope, Input is closeable

ktor-io/jvm/test/io/ktor/utils/io/BlockingAdaptersOnProhibitedThreadTest.kt
e5l: Thanks for catch, was added by mistake

PR #2534, opened 2021-06-29
KTOR-2865 Move server code to io.ktor.server.* package

rsinukov:


e5l (approved):
lgtm, please make sure that we will have all replacements in place

PR #2541, opened 2021-07-01
NPE in ApacheRequestProducer when "http://" is requested

Stexxe:
https://youtrack.jetbrains.com/issue/KTOR-1405


e5l (approved):

ktor-client/ktor-client-apache/jvm/src/io/ktor/client/engine/apache/ApacheRequestProducer.kt
```diff
@@ -32,7 +33,9 @@ internal class ApacheRequestProducer(
 ) : HttpAsyncRequestProducer, CoroutineScope {
 
     private val request: HttpUriRequest = setupRequest()
-    private val host = URIUtils.extractHost(request.uri)!!
+
+    private val host = URIUtils.extractHost(request.uri)
+        ?: throw IllegalStateException("Cannot extract host from URL ${request.uri}")
```
e5l, line 38: I would use `IllegalArgumentException` here

PR #2543, opened 2021-07-01
KTOR-2795 Rename Features to Plugins in Client API

rsinukov:


gradle/compatibility.gradle
```diff
@@ -8,7 +8,7 @@ apiValidation {
         'ktor-client-json-tests',
         'ktor-client-content-negotiation-tests',
         'ktor-client',
-        'ktor-client-features',
+        'ktor-client-plugins',
```
e5l, line 11: consider keeping old modules(empty with transitive dependency to the renamed module)

e5l (changes requested):

gradle/compatibility.gradle
rsinukov: Will be added in the next PR after all moving will be ready

e5l (approved):
lgtm, the same note about replacements

PR #2546, opened 2021-07-02
"ApplicationEngineEnvironment was not started" when accessing application before server is started

Stexxe:
https://youtrack.jetbrains.com/issue/KTOR-1854

### DIFF
--- a/ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/ApplicationEngineEnvironmentReloading.kt
+++ b/ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/ApplicationEngineEnvironmentReloading.kt
@@ -54,7 +54,8 @@ public class ApplicationEngineEnvironmentReloading(
         parentCoroutineContext, rootPath, developmentMode = true
     )
 
-    private var _applicationInstance: Application? = null
+    private var _applicationInstance: Application? = Application(this)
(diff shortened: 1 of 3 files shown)


ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/ApplicationEngineEnvironmentReloading.kt
```diff
@@ -289,7 +290,13 @@ public class ApplicationEngineEnvironmentReloading(
     }
 
     private fun instantiateAndConfigureApplication(currentClassLoader: ClassLoader): Application {
-        val newInstance = Application(this)
+        val newInstance = if (recreateInstance || _applicationInstance == null) {
```
e5l, line 293: What will happen if I call `application.coroutineContext[Job]?.invokeOnCompletion { println("foo") }` before application actually starts? The same question for `intercept` method

e5l (changes requested):

ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/ApplicationEngineEnvironmentReloading.kt
Stexxe: I've added tests to check these cases.

e5l (approved):

PR #2547, opened 2021-07-02
corsCheckRequestHeaders false

Stexxe:
https://youtrack.jetbrains.com/issue/KTOR-445

### DIFF
--- a/ktor-server/ktor-server-core/jvm/src/io/ktor/features/CORS.kt
+++ b/ktor-server/ktor-server-core/jvm/src/io/ktor/features/CORS.kt
@@ -138,10 +138,7 @@ public class CORS(configuration: Configuration) {
     }
 
     private suspend fun ApplicationCall.respondPreflight(origin: String) {
-        val requestHeaders =
-            request.headers.getAll(HttpHeaders.AccessControlRequestHeaders)?.flatMap { it.split(",") }?.map {
-                it.trim().toLowerCasePreservingASCIIRules()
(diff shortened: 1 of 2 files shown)


ktor-server/ktor-server-core/jvm/test/io/ktor/tests/features/CORSTest.kt
```diff
@@ -40,4 +48,43 @@ class CORSTest {
         every { point.port } returns port
         return point
     }
+
+    @Test
+    fun expectRequestHeadersCheckPassIfControlRequestHeadersHeaderHasEmptyValue(): Unit = runBlocking {
```
e5l, line 53: please make sure that the test starts with word `test`, consider also making short name.

I would use: `testCorsIsEmpty`

ktor-server/ktor-server-core/jvm/src/io/ktor/features/CORS.kt
```diff
@@ -164,6 +161,15 @@ public class CORS(configuration: Configuration) {
         respond(HttpStatusCode.OK)
     }
 
+    private fun ApplicationCall.controlRequestHeaders(): List<String> {
+        val rawValue = request.headers[HttpHeaders.AccessControlRequestHeaders]
```
e5l, line 165: Consider reusing value from the `headers` object.

e5l (changes requested):

PR #2549, opened 2021-07-06
Save TypeInfo implementation class in HttpResponseContainer

Prototik:
**Subsystem**
Client core

**Motivation**
In my project I wrote feature plugin which deserializes response bytes with using new `TypeInfo` interface, but it uses custom class extended with some additional info required for that process. But I noticed that `HttpResponseContainer` always re-creates type info with old deprecated data class and there is no any way to preserve original typeinfo class.

**Solution**
I rewrote `HttpResponseContainer` with binary-compatibility way and forced it to save original type info and create old one only by request from old code.
The only ABI breaking change in `component1` fun, but it's source-compatible if end-user not used old typeinfo class directly. Likelly, almost everyone uses it like these:
```kotlin
client.responsePipeline.intercept(HttpResponsePipeline.Transform) { (info, body) ->
  // Implicit usage of component1() here -----------------------------^
  // It will work after recompiling
  ...
}
```

### DIFF
--- a/ktor-client/ktor-client-core/api/ktor-client-core.api
+++ b/ktor-client/ktor-client-core/api/ktor-client-core.api
@@ -1034,14 +1034,17 @@ public abstract class io/ktor/client/statement/HttpResponse : io/ktor/http/HttpM
 }
 
 public final class io/ktor/client/statement/HttpResponseContainer {
-	public fun <init> (Lio/ktor/client/call/TypeInfo;Ljava/lang/Object;)V
+	public synthetic fun <init> (Lio/ktor/client/call/TypeInfo;Ljava/lang/Object;)V
 	public fun <init> (Lio/ktor/util/reflect/TypeInfo;Ljava/lang/Object;)V
-	public final fun component1 ()Lio/ktor/client/call/TypeInfo;
+	public final fun component1 ()Lio/ktor/util/reflect/TypeInfo;
 	public final fun component2 ()Ljava/lang/Object;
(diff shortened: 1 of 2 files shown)


rsinukov (changes requested):
Thanks for your contribution!
This PR contains breaking API changes, so it can't be merged into `main` branch, but can go into `2.0.0-eap`. Do you mind rebasing it?
Second, you can just replace it and completely drop usage of deprecated class. It's OK in a major release.

Prototik:
@rsinukov so you requested to rework it to completely drop deprecated class for `2.0.0-eap`? It's ok for me, I designed a workaround for `1.6`. Please confirm if I got it correctly.

Prototik:
Quickly browsed `2.0.0-eap` branch: seems like it's already removed deprecated class. So we can just close this pr?

rsinukov:
Yes, you got it right. We can't have API changes in a patch release. The next release that allows them is 2.0.0. If it's already deleted there, then this PR can be closed. 

Did I understand correctly that you have a solution for your problem in the current version? If not, what is your use case?

Prototik:
Yeah, i have a solution for current stable and proper migration plan for 2.0 branch.

My use case to use kotlinx.serialization in reflection-less environment, so I need to pass serializer explicitly for every request. The proper way to do that is extending `TypeInfo` and attach serializer to it, but it's possible only in 2.0 branch as current stable erases any custom `TypeInfo` and replaces it with deprecated dataclass. So I'll live with workaround until 2.0 released, it's okay for me

PR #2552, opened 2021-07-13
Don't report websocket extensions when there is nothing to report

Prototik:
**Subsystem**
Client

**Motivation**
In case there is no any extensions for websocket client, ktor sends empty `Sec-WebSocket-Extensions` header. Some servers (CloudFlare in my case) responds with 400 bad request when sees that empty header.

**Solution**
Disable sending `Sec-WebSocket-Extensions` when there is no extensions

https://youtrack.jetbrains.com/issue/KTOR-2388


Prototik:
Fixed by #2556

PR #2556, opened 2021-07-15
KTOR-2388 Fix empty WebSocket extension header

e5l:
Fix [KTOR-2388](https://youtrack.jetbrains.com/issue/KTOR-2388)


rsinukov (approved):

rsinukov (changes requested):
New test fails on native

rsinukov (approved):

PR #2560, opened 2021-07-18
Update changelog with some missing breaking changes

TWiStErRob:
**Subsystem**
Documentation

**Motivation**
Updating Kotlin / coroutines / serialization dependencies is a breaking change. It's hard to find the compatible version of ktor with the corresponding Kotlin version as is, because of the disparate versioning. Including them in the changelog along with any other behavior change is the minimum, because updating ktor might pull any project's Kotlin version over to a new major version.

Version 1.2.3 and 1.4.1 gave the right significance to these updates, it would be nice if that convention was kept.

**Solution**
Updated version history based on Git Blame of gradle.properties.

Also included the suggested change by https://youtrack.jetbrains.com/issue/KTOR-372#focus=Comments-27-4951326.0-0


rsinukov (approved):
Thanks for your contribution!

rsinukov (approved):

PR #2569, opened 2021-08-02
KTOR-2977 Add parameters to refreshTokens

rsinukov:


ktor-client/ktor-client-plugins/ktor-client-auth/common/src/io/ktor/client/plugins/auth/providers/BearerAuthProvider.kt
```diff
@@ -24,17 +25,25 @@ public class BearerTokens(
     public val refreshToken: String
 )
 
+private typealias RefreshTokens =
+    suspend (client: HttpClient, response: HttpResponse, tokens: BearerTokens?) -> BearerTokens?
```
e5l, line 29: Could you tell me why do we have the client in parameters and not in the receiver?

e5l: 3 parameters is quite long, should we try to group it inside some context-like class?

e5l (changes requested):

ktor-client/ktor-client-plugins/ktor-client-auth/common/src/io/ktor/client/plugins/auth/providers/BearerAuthProvider.kt
rsinukov: 1. `client.get("update-token").body<Token>()` looks more ideomatic than just `get("update-token").body<Token>()`
2. Yes, let's create a class.

e5l: 1. agree

e5l (approved):

PR #2571, opened 2021-08-03
KTOR-2876 Add shortcut functions for basic and bearer A

hfhbd:
authorization

**Subsystem**
Client RequestBuilder Common

**Motivation**
https://youtrack.jetbrains.com/issue/KTOR-2876

**Solution**
Added shortcuts

### DIFF
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/request/utils.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/request/utils.kt
@@ -5,6 +5,7 @@
 package io.ktor.client.request
 
 import io.ktor.http.*
+import io.ktor.util.*
 import io.ktor.util.date.*
 
 /**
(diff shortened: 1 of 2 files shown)


ktor-client/ktor-client-core/common/src/io/ktor/client/request/utils.kt
```diff
@@ -77,3 +78,15 @@ public fun HttpRequestBuilder.parameter(key: String, value: Any?): Unit =
  */
 public fun HttpRequestBuilder.accept(contentType: ContentType): Unit =
     headers.append(HttpHeaders.Accept, contentType.toString())
+
+/**
+ * Sets the [HttpHeaders.Authorization] to Basic Authorization with the provided [username] and [password]
+ */
+public fun HttpRequestBuilder.basicAuth(username: String, password: String): Unit =
```
rsinukov, line 85: Would it be better to put it in the same place with `Bearer/BasicAuthProvider`?

hfhbd: This would require adding the `auth` artifact too. 
My idea was to provide a shortcut function for simple one-time usage. For more configs/options (like default auth on every request) you should use the `auth` artifact.

rsinukov:
Looks good. Please rebase on 2.0.0-eap and commit API changes

hfhbd:
@rsinukov I totally forgot the `apiDump`.

rsinukov (approved):

PR #2572, opened 2021-08-05
KTOR-3008 Fixed Client Bearer Feature 

den-gap:
**Subsystem**
Client

**Motivation**
After clearing the tokens, we could not start loading the tokens again because the initialized flag was not reset
e.g., log out of one account and log in to another was impossible

**Solution**
just set initialized to null when clearing tokens

### DIFF
--- a/ktor-client/ktor-client-features/ktor-client-auth/common/src/io/ktor/client/features/auth/providers/AuthTokenHolder.kt
+++ b/ktor-client/ktor-client-features/ktor-client-auth/common/src/io/ktor/client/features/auth/providers/AuthTokenHolder.kt
@@ -15,6 +15,7 @@ internal class AuthTokenHolder<T>(
 
     internal fun clearToken() {
         cachedBearerTokens.value = null
+        initialized.value = false
     }
 
     internal suspend fun loadToken(): T? {
--- a/ktor-client/ktor-client-features/ktor-client-auth/common/test/io/ktor/client/features/auth/AuthTest.kt
+++ b/ktor-client/ktor-client-features/ktor-client-auth/common/test/io/ktor/client/features/auth/AuthTest.kt
@@ -354,4 +354,28 @@ class AuthTest : ClientLoader() {
             assertEquals("OK", second)
         }
     }
+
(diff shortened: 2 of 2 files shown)


e5l (approved):
Hey @den-gap, LGTM. Could you also add a test for this case?

den-gap:
> Hey @den-gap, LGTM. Could you also add a test for this case?

Unfortunately no, something goes wrong and I can not check the tests

rsinukov:
Hi @den-gap, thanks for the PR.
Please add a test to `AuthTest.kt` file. Something like this should work
```kotlin
    @Test
    fun testLoadTokenAfterClear() = clientTests {
        var loadCount = 0
        config {
            install(Auth) {
                bearer {
                    refreshTokens { null }
                    loadTokens {
                        loadCount++
                        BearerTokens("valid", "refresh")
                    }
                }
            }
        }

        test { client ->
            client.get("$TEST_SERVER/auth/bearer/test-refresh")
            client[Auth].providers.filterIsInstance<BearerAuthProvider>().first().clearToken()
            client.get("$TEST_SERVER/auth/bearer/test-refresh")

            assertEquals(2, loadCount)
        }
    }
```

den-gap:
> Hi @den-gap, thanks for the PR.
> Please add a test to `AuthTest.kt` file. Something like this should work
> 
> ```kotlin
>     @Test
>     fun testLoadTokenAfterClear() = clientTests {
>         var loadCount = 0
>         config {
>             install(Auth) {
>                 bearer {
>                     refreshTokens { null }
>                     loadTokens {
>                         loadCount++
>                         BearerTokens("valid", "refresh")
>                     }
>                 }
>             }
>         }
> 
>         test { client ->
>             client.get("$TEST_SERVER/auth/bearer/test-refresh")
>             client[Auth].providers.filterIsInstance<BearerAuthProvider>().first().clearToken()
>             client.get("$TEST_SERVER/auth/bearer/test-refresh")
> 
>             assertEquals(2, loadCount)
>         }
>     }
> ```

done, thanks

PR #2573, opened 2021-08-09
KTOR-2605 Add THIRDPARTY.md

hhariri:
KTOR-2605 Add list of third-party libraries used


e5l (changes requested):
What about Netty, Tomcat, Jetty, GSON and CURL? Probably it should be some way to get the whole list from the Gradle

e5l (approved):

PR #2577, opened 2021-08-12
Force eager initialization for initHook on Native

SvyatoslavScherbina:
**Subsystem**
Client

**Motivation**
Prepare for lazy top-level property initialization strategy in Kotlin/Native.
We consider enabling this strategy by default with new memory manager.

**Solution**
`@EagerInitialization` annotation forces a property to be initialized eagerly on program startup, like it happens for any top-level property in current non-lazy init strategy.
If lazy init strategy is not enabled, this annotation doesn't have any effect, so it is safe to merge it to `main` branch. But the annotation is available since Kotlin 1.5.30-RC, so the version update would be required.


e5l (approved):
lgtm. will merge after CI run

PR #2585, opened 2021-08-21
KTOR-2584 Remove invalid session id

Maxr1998:
**Subsystem**
Server/Sessions

**Motivation**
Found this apparent bug when looking at the code. There was already an issue opened [here](https://youtrack.jetbrains.com/issue/KTOR-2584).

**Solution**
Actually remove the invalid session as described in the comment above.


e5l (approved):

e5l:
Thanks for the PR, lgtm

PR #2586, opened 2021-08-23
KTOR-3055 Add @ThreadLocal annotation to top level constants

Evgeny-Gusev:
**Subsystem**
Ktor Client, Auth Feature

**Motivation**
Auth feature uses global constants during token refresh process which are not marked with @ThreadLocal annotation. This makes it impossible to perform refresh from non-main thread in Kotlin/Native targets.
I attached a stack trace to [the YouTrack issue](https://youtrack.jetbrains.com/issue/KTOR-3055).

**Solution**
In order to make it work on iOS I annotated 2 constants as ThreadLocal.

### DIFF
--- a/ktor-client/ktor-client-features/ktor-client-auth/common/src/io/ktor/client/features/auth/Auth.kt
+++ b/ktor-client/ktor-client-features/ktor-client-auth/common/src/io/ktor/client/features/auth/Auth.kt
@@ -82,5 +82,5 @@ public fun HttpClientConfig<*>.Auth(block: Auth.() -> Unit) {
  * be removed after that.
  */
 @PublicAPICandidate("1.6.0")
-@ThreadLocal
+@SharedImmutable
 internal val AuthHeaderAttribute = AttributeKey<HttpAuthHeader>("AuthHeader")
--- a/ktor-client/ktor-client-features/ktor-client-auth/common/src/io/ktor/client/features/auth/Auth.kt
+++ b/ktor-client/ktor-client-features/ktor-client-auth/common/src/io/ktor/client/features/auth/Auth.kt
@@ -10,6 +10,7 @@ import io.ktor.client.request.*
 import io.ktor.http.*
 import io.ktor.http.auth.*
 import io.ktor.util.*
+import kotlin.native.concurrent.*
 
 /**
  * Client authentication feature.
@@ -81,4 +82,5 @@ public fun HttpClientConfig<*>.Auth(block: Auth.() -> Unit) {
  * be removed after that.
  */
 @PublicAPICandidate("1.6.0")
+@ThreadLocal
 internal val AuthHeaderAttribute = AttributeKey<HttpAuthHeader>("AuthHeader")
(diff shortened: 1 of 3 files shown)


Stexxe:
The `testRefreshOnBackgroundThread` test passes even without your changes on Linux.

ktor-client/ktor-client-features/ktor-client-auth/common/src/io/ktor/client/features/auth/Auth.kt
```diff
@@ -81,4 +82,5 @@ public fun HttpClientConfig<*>.Auth(block: Auth.() -> Unit) {
  * be removed after that.
  */
 @PublicAPICandidate("1.6.0")
+@ThreadLocal
```
rsinukov, line 85: Since `AttrubuteKey` does not overwrite `equals `, it is important to have a single reference to this property. Please use `@SharedImmutable` instead.

rsinukov (changes requested):
Thanks for your contribution! One small change is needed.

ktor-client/ktor-client-features/ktor-client-auth/common/src/io/ktor/client/features/auth/Auth.kt
Evgeny-Gusev: Thank you for the comment. I changed it to `@SharedImmutable`

Evgeny-Gusev:
@Stexxe you're right. Apparently for some reason I can't make any test fail in native targets in `AuthTests.kt`. I tried to add `assertTrue { false }` and run `./gradlew :ktor-client:ktor-client-features:ktor-client-auth:allTests`, but it shows errors only for jvm and js targets. Am I missing something?

Screenshot 2021-08-24 at 23 16 54

Stexxe:
> @Stexxe you're right. Apparently for some reason I can't make any test fail in native targets in `AuthTests.kt`. I tried to add `assertTrue { false }` and run `./gradlew :ktor-client:ktor-client-features:ktor-client-auth:allTests`, but it shows errors only for jvm and js targets. Am I missing something?
> 
> edit: I'm running it on macOS
> 
> Screenshot 2021-08-24 at 23 16 54

The following simplest test fails on my machine when running mentioned Gradle task:
```kt
@Test
fun testSome() {
    assertEquals(0, 1)
}
```

Evgeny-Gusev:
> The following simplest test fails on my machine when running mentioned Gradle task:
> 
> ```kotlin
> @Test
> fun testSome() {
>     assertEquals(0, 1)
> }
> ```

I've tried with 2 simple tests:
```kotlin
    @Test
    fun testSomePlain() {
        assertEquals(0, 1)
    }

    @Test
    fun testSomeClient() = clientTests {
        test { _ ->
            assertEquals(0, 1)
        }
    }
```

`testSomePlain` fails on all targets, but `testSomeClient` fails only on jvm and js.

Stexxe:
@Evgeny-Gusev you are right. @rsinukov please take a look.

rsinukov:
Looks like a problem on our side. I'll take a look on it ASAP. In the meantime, this PR is good to be merged after CI completes.

rsinukov (approved):

rsinukov:
I created this ticket to track issues with tests not running https://youtrack.jetbrains.com/issue/KTOR-3069

PR #2588, opened 2021-08-25
KTOR-3069 Register native engines in tests

rsinukov:
Native engines were not registered for tests outside of `ktor-client-tests` modules.

### DIFF
--- a/ktor-client/ktor-client-tests/build.gradle.kts
+++ b/ktor-client/ktor-client-tests/build.gradle.kts
@@ -98,41 +98,6 @@ kotlin.sourceSets {
             api(project(":ktor-client:ktor-client-js"))
         }
     }
-
-    if (rootProject.ext.get("native_targets_enabled") as Boolean) {
-        if (!ideaActive) {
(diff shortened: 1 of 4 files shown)


ktor-client/ktor-client-tests/build.gradle.kts
```diff
@@ -166,10 +131,53 @@ if (!ideaActive) {
...
+
+                            if (!hostname.startsWith("win")) {
+                                api(project(":ktor-client:ktor-client-cio"))
+                            }
+                        }
+                    }
+                }
+                return@sourceSets
+            }
+            if (name in listOf("macosX64Test", "linuxX64Test")) {
```
e5l, line 159: `when` could be used

e5l (approved):
lgtm

ktor-client/ktor-client-tests/build.gradle.kts
rsinukov: yes, but it adds one more level of nesting

PR #2593, opened 2021-09-06
KTOR-3116 Fix closing installed features

john5f35:
**Subsystem**
Client

**Motivation**
KTOR-3116
It seems that the original intention of closing installed `Closeable` features are not working properly due to an added layer of indirection of `FEATURE_INSTALLED_LIST` key at https://github.com/ktorio/ktor/blob/92f4e96b6f863ffbfc3f6afae52a8ac584a2be3b/ktor-client/ktor-client-core/common/src/io/ktor/client/HttpClientConfig.kt#L75
The demo test fails because the only key in `attributes` is this `FEATURE_INSTALLED_LIST`, and all features are under the `Attributes` value of this key.

**Solution**
This PR retrieves the correct "map" for installed features to invoke the `close()` function on them as originally intended.

### DIFF
--- a/ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/CommonHttpClientTest.kt
+++ b/ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/CommonHttpClientTest.kt
@@ -8,6 +8,7 @@ import io.ktor.client.*
 import io.ktor.client.engine.mock.*
 import io.ktor.client.features.*
 import io.ktor.util.*
+import io.ktor.utils.io.concurrent.*
 import io.ktor.utils.io.core.*
 import kotlinx.coroutines.*
 import kotlin.test.*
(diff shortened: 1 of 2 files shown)


ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/CommonHttpClientTest.kt
```diff
@@ -36,4 +36,26 @@ class CommonHttpClientTest {
...
+    fun testHttpClientClosesInstalledFeatures() {
+        val client = HttpClient(MockEngine) {
+            engine { addHandler { respond("") } }
+            install(TestFeature)
+        }
+        client.close()
+        assertTrue(client.feature(TestFeature)!!.closed)
+    }
+    class TestFeature : Closeable {
+        var closed = false
```
rsinukov, line 50: Native tests are failing on mutating this property. You can use `var closed by shared(false)` to fix it.

rsinukov (changes requested):
Hi!
Thanks for the PR, good catch! LGTM in general, but one small issue in tests.

PR #2594, opened 2021-09-06
KTOR-3105 Fix HttpCookies feature overwriting request cookies

john5f35:
**Subsystem**
Client

**Motivation**
KTOR-3105
`HttpCookies` forcefully overwrite the cookie header with its cookie storage. This causes additional cookies explicitly included in request builder to be overwritten.
This holds true for dropping the cookie header when the storage is empty.

The current implementation effectively asserts that the storage in `HttpCookies` feature holds *all* cookies ever needed by a potential client. However this is inflexible and not necessarily true for all use cases.

**Solution**
Preserve the currently cookie header if it exists; append the rendered storage cookies after the header.


ktor-client/ktor-client-core/common/test/CookiesTest.kt
```diff
@@ -28,4 +28,38 @@ class CookiesTest {
...
+        val builder = HttpRequestBuilder()
+
+        builder.cookie("test", "value")
+        feature.sendCookiesWith(builder)
+
+        assertEquals("test=value; SOMECOOKIE=somevalue;", builder.headers[HttpHeaders.Cookie])
+    }
+
+    @Test
+    fun testNoCookieHeaderWhenEmptyStorageAndNoRequestCookies() = testSuspend {
```
john5f35, line 57: This is just to highlight the desired behaviour that we don't want to send an empty cookie header (#2008)

e5l (commented):
Hey @john5f35, thanks for the PR. LGTM, will merge after CI run

rsinukov:
Hi @john5f35.
These 2 tests are failing on CI. Can you check, please?
```
SessionAuthTest.testSessionOnly[jvm] io.ktor.tests.auth
SessionAuthTest.testSessionOnlyDeprecated[jvm] io.ktor.tests.auth
```

ktor-http/common/src/io/ktor/http/Cookie.kt
```diff
@@ -133,22 +133,10 @@ public fun renderSetCookieHeader(cookie: Cookie): String = with(cookie) {
 }
 
 /**
- * Format `Set-Cookie` header value
+ * Format `Cookie` header value
  */
 public fun renderCookieHeader(cookie: Cookie): String = with(cookie) {
```
john5f35, line 138: This function seems only being used by https://github.com/ktorio/ktor/blob/14c687426471ea854a4cf96b4c91de38ebf4fd37/ktor-client/ktor-client-core/common/src/io/ktor/client/request/utils.kt#L34-L47
But the implementation was not appropriate; it didn't conform to `Cookie` header spec in [MDN doc](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cookie) and [RFC6265 spec](https://httpwg.org/specs/rfc6265.html#cookie) by potentially containing attribute pairs such as `Max-Age=...; Expires=...;`. This could causes a problem when trying to parse cookies from such string, as `Max-Age`, `Expires` will be treated as cookies themselves.

john5f35:
Hi @e5l, @rsinukov! Sorry for the long delay.
I looked into those failed tests, and the problem was due to interaction between `HttpCookies` and `HttpRedirect` features. But the fix also involves some complexity around `HttpRequestBuilder.cookie()` function. 

I added a `CookiesAndRedirectMockedTest` to illustrate the various combinations of these three things to demonstrate the resulting behaviours. I've also reworked the `HttpCookies` changes to cover the original intent and make everything work. 

The PR is ready for review again. Thanks!

PR #2601, opened 2021-09-10
KTOR-3120 Add colors to CallLogging logs

Ololoshechkin:
https://youtrack.jetbrains.com/issue/KTOR-3120


ktor-server/ktor-server-core/jvm/src/io/ktor/features/CallLogging.kt
```diff
@@ -221,7 +223,34 @@ private class MDCSurvivalElement(mdc: Map<String, String>) : ThreadContextElemen
     private object Key : CoroutineContext.Key<MDCSurvivalElement>
 }
 
-private fun defaultFormat(call: ApplicationCall): String = when (val status = call.response.status() ?: "Unhandled") {
-    HttpStatusCode.Found -> "$status: ${call.request.toLogString()} -> ${call.response.headers[HttpHeaders.Location]}"
-    else -> "$status: ${call.request.toLogString()}"
+private fun defaultFormat(call: ApplicationCall): String {
+    if (!AnsiConsole.isInstalled()) {
+        AnsiConsole.systemInstall()
```
e5l, line 228: In some cases, this method can throw an exception. Could we fall back to the regular log in such cases? 
It could be part of the `colored` method for instance

e5l: We also think about flag to opt-out the feature

e5l (changes requested):

e5l (approved):

PR #2602, opened 2021-09-11
KTOR-399 Add variant of parametersOf() that takes in Map<String, List…

rahul29112002:
…<String>>

**Subsystem**
ktor-http

**Motivation**
parametersOf() function returns a Parameters instance that internally stores a map. Yet there existed no way to instantiate a Parameters instance using a map directly. See: [Issue 399](https://youtrack.jetbrains.com/issue/KTOR-399)

**Solution**
Added a new function parametersOf(Map<String, List<String>>) that aims to solve the problem


BOOMeranGG:
I started doing this in the past (# 2515) but didn’t finish it. Pay attention to the comments from rsinukov (about `2.0.0-eap` branch and MutableMap)

PR #2603, opened 2021-09-11
KTOR-781 Add unix domain sockets support

ptitjes:
This is a PR to fix [KTOR-781](https://youtrack.jetbrains.com/issue/KTOR-781).

**Subsystem**
ktor-network, ktor-client-cio, ktor-server-cio

**Motivation**

[KTOR-781 Support for Unix domain sockets for both server and client](https://youtrack.jetbrains.com/issue/KTOR-781)

**Solution**

- Implement a `SocketAddress` hierarchy inside `ktor-network` itself
- Remove references to `NetworkAddress` inside `ktor-network`
- Target Java 1.7 in the JVM code and remove useless reflection in favor of newer 1.7 NIO APIs
- Fix bug with Native INET address resolution (port numbers are in network byte order)
- Implement Native INET address IP to string
- Update `ktor-network` API file
- Update uses of `ktor-network` in CIO client and server engines

**Problems**

The Native code now has a test for unix sockets.
I **manually tested the JVM code**, but I don't know how we could test it as part of the build as this requires Java 16+.


AndrewWestberg:
This is some really nice code. Much better than the hack I am using currently. Hopefully this gets reviewed and merged soon! Thanks for contributing!

ptitjes:
@e5l I am wondering if I should also add a commit to rename `SocketBuilder.tcp()` and `SocketBuilder.udp()` to `SocketBuilder.byteStream()` and `SocketBuilder.datagram()`? Indeed, `tcp` and `udp` don't make much sense anymore.

Also, just some notes about the choice to make a new `SocketAddress` hierarchy inside `ktor-network`. I tried two different designs to make the hierarchy around `NetworkAddress` without success due to:
- the difference between ways to instantiate `java.net.InetSocketAddress` and `java.net.UnixDomainSocketAddress`
- the fact that `java.net.UnixDomainSocketAddress` is only in Java 16+
- I believe `java.net.InetSocketAddress` is resolved early, so now resolution is made at `connect` and `bind` which is consistent with what is made for Native
- this will be easier to extract when `ktor-network` moves to `kotlinx-io`

ptitjes:
Finally, do you think this could make it for Ktor 2.0?

ptitjes:
@e5l I've just rebased on main.

Here is the final set of implemented changes:
- Implement a `SocketAddress` hierarchy inside `ktor-network` itself
- Remove references to `NetworkAddress` inside `ktor-network`
- Target Java 1.7 in the JVM code and remove useless reflection in favor of newer 1.7 NIO APIs
- Fix bug with Native INET address resolution (port numbers are in network byte order)
- Implement Native INET address IP to string
- Update `ktor-network API` file
- Update uses of `ktor-network` in CIO client and server engines

Some notes:
- I only ran `jvmTest` (on JVM 11 and 16) and `linuxX64Test`, but this should cover all my changes.
- I added a conditional on the test (in `jvmAndNix`) for unix domain sockets, so that it only runs if the feature is available in the JVM running the test. So it is possible to add a Java 16+ build in TeamCity to test this feature.
- I have another branch with an additional commit (https://github.com/ptitjes/ktor/commit/f0580cf338ea57ed4e54d396e9ccf344951e23c6) that renames and deprecates `Builders.tcp()` and `Builders.udp()` to `Builders.stream()` and `Builders.datagram()`. Indead, The `tcp` and `udp` protocol names are usually used when using network sockets, and the more general `stream` and `datagram` are used when the using either network or unix domain sockets. I can cherry-pick this commit in this PR if you want.
- Maybe the `SocketAddress` hierarchy should be moved to `ktor-utils` to completely replace the `NetworkAddress` class.
- I believe the code in the `nixMain` part could be further simplified.
- This PR doesn't fully fix [KTOR-781](https://youtrack.jetbrains.com/issue/KTOR-781) as I believe the unix domain sockets support could be extended to the `ktor-server` and `ktor-client` APIs. However, I don't know what shape should this take in terms of API design, so maybe this can be done in a followup.

ptitjes:
Rebased again to have kotlin 1.6.0 on my local install.

ptitjes:
Rebase again to have https://github.com/ktorio/ktor/pull/2698 on my local install.

ptitjes:
Rebased again.

e5l:
Merged, thank you!

ptitjes:
> Merged, thank you!

Thank You!

Thomas-Vos:
FYI this PR broke sockets on Android: https://youtrack.jetbrains.com/issue/KTOR-3659

PR #2604, opened 2021-09-14
KTOR-3091 Verify nonce in OAuth2 authentication

rsinukov:


e5l (approved):
LGTM, @hhariri could you check public API docs?

e5l:
@rsinukov, could you also check public API?

rsinukov:
I changed PR so that there is no new API, so nothing to review in terms of docs.

PR #2611, opened 2021-09-20
Disable HMPP

e5l:


rsinukov (changes requested):

ktor-bom/build.gradle
```diff
@@ -2,7 +2,8 @@ plugins {
     id 'java-platform'
 }
 
-apply from: rootProject.file('gradle/publish.gradle')
+// TODO
+// apply from: rootProject.file('gradle/publish.gradle')
```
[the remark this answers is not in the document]
e5l: thanks, fixed

ktor-client/ktor-client-tests/build.gradle.kts
```diff
@@ -180,16 +180,16 @@ rootProject.allprojects {
             if (!(rootProject.ext.get("native_targets_enabled") as Boolean)) {
                 return@sourceSets
             }
-            if (KtorBuildProperties.ideaActive) {
+            if (IDEA_ACTIVE) {
                 if (name == "posixTest") {
                     getByName(name) {
                         dependencies {
-                            val hostname: String by project.ext
+//                            val hostname: String by project.ext
```
rsinukov, line 187: please uncomment or delete this code

ktor-client/ktor-client-tests/build.gradle.kts
```diff
@@ -126,13 +126,13 @@ kotlin.sourceSets {
         } else {
             val posixTest by getting {
                 dependencies {
-                    val hostname: String by project.ext
+//                    val hostname: String by project.ext
```
rsinukov, line 129: same here

ktor-client/ktor-client-tests/build.gradle.kts
e5l: Thanks, fixed

ktor-client/ktor-client-tests/build.gradle.kts
e5l: Thanks, fixed!

e5l:
Sure, will check it along with the code-style

buildSrc/src/main/kotlin/JvmConfig.kt
```diff
@@ -0,0 +1,111 @@
...
+                    }
+
+                    api("org.slf4j:slf4j-api:$slf4j_version")
+                }
+            }
+
+            val jvmTest by getting {
+                dependencies {
+                    api("org.jetbrains.kotlin:kotlin-test")
+                    api("org.jetbrains.kotlin:kotlin-test-junit")
```
olme04, line 58: junit artifacts added twice.
Also, does it make sense to add them, as now test dependencies added automatically when adding `kotlin-test` to common sourceset? same as with adding `stdlib` (in all targets)

buildSrc/src/main/kotlin/JvmConfig.kt
```diff
@@ -0,0 +1,111 @@
...
+import org.jetbrains.kotlin.gradle.targets.jvm.tasks.*
+
+val JDK_16: String? = System.getProperty("JDK_16")
+val JDK_17: String? = System.getProperty("JDK_17")
+val JDK_18: String? = System.getProperty("JDK_18")
+
+val JDK_11: String? = System.getProperty("JDK_11")
+
+fun Project.configureJvm() {
+    val jdk = when (name) {
```
olme04, line 17: `KtorBuildProperties.projectJdk` do the same (duplicate code)

buildSrc/src/main/kotlin/TargetsConfig.kt
```diff
@@ -0,0 +1,178 @@
...
+    val hasDesktop = hasPosix || files.any { it.name == "desktop" }
+    val hasNix = hasPosix || files.any { it.name == "nix" }
+    val hasDarwin = hasNix || files.any { it.name == "darwin" }
+    val hasJs = hasCommon || files.any { it.name == "js" }
+    val hasJvm = hasCommon || files.any { it.name == "jvm" }
+
+    kotlin {
+        if (hasJvm) {
+            jvm()
+            configureJvm()
```
olme04, line 28: `configureJvm` also installs `jvm` target. for consistency may be it's better to leave target installing in one place?

buildSrc/src/main/kotlin/TargetsConfig.kt
```diff
@@ -0,0 +1,178 @@
...
+        if (hasJs) {
+            js {
+                nodejs()
+                browser()
+            }
+
+            configureJs()
+        }
+
+        if (hasPosix) {
```
olme04, line 40: is it expected that those targets will be created even if `IDEA_ACTIVE` is false?

buildSrc/src/main/kotlin/TargetsConfig.kt
```diff
@@ -0,0 +1,178 @@
...
+                    val desktopTest by creating
+                }
+            }
+        }
+
+        sourceSets {
+            if (hasCommon) {
+                val commonMain by getting {
+                    dependencies {
+                        api("org.jetbrains.kotlin:kotlin-stdlib-common")
```
olme04, line 77: same comment about stdlib/test dependencies as in `JvmConfig` file

buildSrc/src/main/kotlin/TargetsConfig.kt
```diff
@@ -0,0 +1,178 @@
...
+
+                extra.set("commonStructure", true)
+            }
+            if (hasJvmAndNix) {
+                val jvmAndNixMain by creating {
+                    findByName("commonMain")?.let { dependsOn(it) }
+
+                    dependencies {
+                        api("org.jetbrains.kotlin:kotlin-stdlib-common")
+                        api("org.jetbrains.kotlinx:kotlinx-coroutines-core:$coroutines_version")
```
olme04, line 96: is it needed, as `common` already contains it?

buildSrc/src/main/kotlin/Versions.kt
```diff
@@ -0,0 +1,28 @@
+/*
+ * Copyright 2014-2021 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+val jdk8Modules = setOf(
```
olme04, line 5: same exists in `KtorBuildProperties` - is it really needed to duplicate it?

olme04 (commented):
minor comments.
overall, Im just curios, why at all HMPP need to be disabled now, as soon HMPP will be enabled by default? :)

e5l:
Hey @olme04, thanks for the review. I will take a look

buildSrc/src/main/kotlin/TargetsConfig.kt
olme04: minor: I see that `posixTargets` also called down the file (so similar as with JVM double install of targets, even if should work, better to remove it from here for consistency, as this file is already hard to understand in detail

buildSrc/src/main/kotlin/JvmConfig.kt
e5l: Thank you, haven't noticed. Fixed

buildSrc/src/main/kotlin/TargetsConfig.kt
e5l: Yep, it should work. Fixed

buildSrc/src/main/kotlin/TargetsConfig.kt
e5l: `posixTargets()` is just an accessor POSIX targets(which create them if they don't exist).

I would say `posixTargets()` is called here for consistency: we create all targets before getting down to the source set configuration

buildSrc/src/main/kotlin/TargetsConfig.kt
e5l: Thanks, fixed

buildSrc/src/main/kotlin/Versions.kt
e5l: Thanks, removed

PR #2612, opened 2021-09-20
KTOR-3166 Fix watcher access on Android

e5l:
Fix [KTOR-3166](https://youtrack.jetbrains.com/issue/KTOR-3166)

@Stexxe please let me know if you have an idea we could test it


Stexxe:
> Fix [KTOR-3166](https://youtrack.jetbrains.com/issue/KTOR-3166)
> 
> @Stexxe please let me know if you have an idea we could test it

If we can substitute a class loader then we could create one for testing where some classes like `java.nio.file.WatchService` are not found.

rsinukov (approved):

e5l:
@Stexxe actually this is not a bad idea!

e5l:
@rsinukov, nope it looks like the remote host is not responding, will check the test

PR #2610, opened 2021-09-19
CborFeature

mpetuska:
[KTOR-3174 [Ktor Client] CborFeature](https://youtrack.jetbrains.com/issue/KTOR-3174)
**Subsystem**
Client
* [new] ktor-client-cbor
* [new] ktor-client-cbor-serialization

**Motivation**
Add support for CBOR serialization format to unlock ktor client usage with CBOR servers

**Solution**
Base solution on existing JSON format and implement new serializer module based on kotlinx-serialization-cbor (MPP). Note that due to trying to keep existing API surface unaffected, a lot of code duplication was introduced between JsonFeature and CborFeature. Long term I'd recommend unifying various different serializatrion formats under a `ContentNegotiation` feature for the client (just like we have on the server). I'm happy to raise a PR for that if you'd accept breaking changes.

PR #2613, opened 2021-09-20
generic client KotlinxSerializer for ContentNegotiation feature

mpetuska:
[KTOR-3174 [Ktor Client] CborFeature](https://youtrack.jetbrains.com/issue/KTOR-3174)
Superset of #2610
**Subsystem**
Client

**Motivation**
Add support for CBOR serialization format to unlock ktor client usage with CBOR servers. Also allow for easy registration of other serialization formats mapped to ContentType

**Solution**
Replicate the `serialization()` function behaviour we have for server-ContentNegotiation feature and use it to implement shortcuts for json and cbor.

To free up the name `ktor-client-serialization` for new wrapper module to encapsulate such implementations, existing json modules were renamed to follow this schema: `ktor-client-json-<<backingSerializer>>`. e.g.: `ktor-client-json-gson`, `ktor-client-json-serialization`...

**Usage**
```kotlin
install(ContentNegotiation) {
  cbor(Cbor) {
     // Allows matching custom serializers in the pipeline. Already being used to inject JsonObject serializer for json implementation based on this
     match(MyCustomSerializer()) {value: Any ->
        value == someCondition
     }
  }
  // equvivalent to
  serialization(ContentType.Application.Cbor, Cbor) {
     match(MyCustomSerializer()) {value: Any ->
        value == someCondition
     }
  }
}
```


mpetuska:
@rsinukov over to you. Had to reshuffle quite a few modules, but seeing as it's targeting 2.0.0, I hope it's acceptable. I honestly believe this new structure should make artefacts clearer and easier to both, manage and consume.

mpetuska:
I can do you one better. [Here's](https://github.com/mpetuska/ktor/blob/2.0.0-eap-extended-content-negotiation/ktor-client/ktor-client-plugins/ktor-client-serialization/ktor-client-serialization-json/common/src/io/ktor/client/plugins/serialization/json.kt#L35) an example usage of it. It's then being consumed [here](https://github.com/mpetuska/ktor/blob/2.0.0-eap-extended-content-negotiation/ktor-client/ktor-client-plugins/ktor-client-serialization/common/src/io/ktor/client/plugins/SerializationConverter.kt#L62) to make "guessing" serializers more lenient.

In short, the idea is that if everything goes to shit, this allows extending "last-resort" handling of things. Similar pattern is being used on the server ContentNegotiation as well, so this `match()` escape hatch just allows to make it even more flexible and generic.

rsinukov:
Can you point to the code that uses this pattern on the server?

In case of `JsonElement` [this](https://github.com/mpetuska/ktor/blob/2.0.0-eap-extended-content-negotiation/ktor-client/ktor-client-plugins/ktor-client-serialization/common/src/io/ktor/client/plugins/SerializationConverter.kt#L62) will not be reached, because serializer will be found on the previous step [here](https://github.com/mpetuska/ktor/blob/2.0.0-eap-extended-content-negotiation/ktor-client/ktor-client-plugins/ktor-client-serialization/common/src/io/ktor/client/plugins/SerializationConverter.kt#L49).

The other question is why it's used only inside `serialize`, but not in `deserialize`?

mpetuska:
Same thing is going on in the existing client [implementation](https://github.com/mpetuska/ktor/blob/2.0.0-eap-extended-content-negotiation/ktor-client/ktor-client-plugins/ktor-client-json/ktor-client-json-serialization/common/src/io/ktor/client/plugins/json/serializer/KotlinxSerializer.kt#L58) as well.

rsinukov:
I see, thanks. I am a bit hesistant to introduce new powerful public API without clear design and use cases to the end users. Maybe it can be useful to add some extension points to the authors of `ContentConverter`, since they are more advanced and motivated to learn new API users. 
Also, in the 2.0.0 there are no client or server converters, they are reused between them.

mpetuska:
Which code exactly?

mpetuska:
As for your concerns about exposing this new public api surface, we could require an optin for it or even mark it as internal ktor api

mpetuska:
BTW, should I try moving these new converters to shared module instead?

rsinukov:
The converters themselves should be in a shared module (`ktor-shared:ktor-shared-serialization-gson`, `ktor-shared:ktor-shared-serialization-jackson`, `ktor-shared:ktor-shared-serialization-kotlinx`). But install and register functions are client specific and can't be moved.

mpetuska:
Got it. Are you still happy with new `ktor-client-serialization-*` modules, though?

rsinukov:
We are deprecating `InternalKtorApi` since it doesn't really solve the problem with binary compatibility. So it's not an option to use it. Same goes for `OptIn`.

What do you think about such naming for modules:
```
ktor-client-gson
ktor-client-jackson
ktor-client-kotlinx-serialization // common 
ktor-client-kotlinx-serialization-json
ktor-client-kotlinx-serialization-cbor
```
Since `gson` and `jackson` do not support other formats, it's not neccesary to specify that they are `json`.

rsinukov:
Oh, sorry I got lost in the modules myself =)

So, we don't need anything specific for client. It's enough to add new format to the shared serialization modules, somethinkg like this:
```
ktor-shared-serialization // common for all serializers. keep it
ktor-shared-serialization-gson // keep as it is
ktor-shared-serialization-jackson // keep as it is
ktor-shared-serialization-kotlinx // keep everything but remove `JsonSupport.kt`
ktor-shared-serialization-kotlinx-json // put `JsonSupport.kt` here
ktor-shared-serialization-kotlinx-cbor // add new format here
```

mpetuska:
Well jackson does support CBOR at least ;) But in general I quite like `kotlinx-serialization` vs `serialization` naming conventions. Makes things a lot clearer. Mind if I introduce the following modules then? 
```
ktor-shared-serialization
ktor-shared-serialization-gson
ktor-shared-serialization-jackson // <--- common stuff for jackson
ktor-shared-serialization-jackson-json // <--- json stuff for jackson
ktor-shared-serialization-jackson-cbor // <--- cbor stuff for jackson
ktor-shared-serialization-kotlinx // <--- common stuff for kotlinx-serialization
ktor-shared-serialization-kotlinx-json // <--- json stuff for kotlinx-serialization
ktor-shared-serialization-kotlinx-cbor // <--- cbor stuff for kotlinx-serialization
```

rsinukov:
Yes, I thinks this is the way to go. Thank you!

rsinukov:
Regarding the special handling of `JsonElement`, we can add extention point inside the `ktor-shared-serialization-kotlinx` and use it in `ktor-shared-serialization-kotlinx-json`. This way we avoid exposing users to this API, while still being able to add format specific classes.

rsinukov:
Yes, sorry for the confusion. I mean this API will be public, but the end users will not need to know about it and use it. It will be used only inside modules that provide specific formats.

mpetuska:
@rsinukov back to you. I've decided to drop jackson-cbor as it turned out to be more involved than I hoped and focus on kotlinx-serialization in this PR (might raise another PR later for jackson).

In this iteration I've abstracted all the tests you had for kotlinx json and made them reusable to make it easier to add even more formats in the future. As for that `matching()` API, I've left it as-is in the ContentConverter, but removed `configuration` argument from `serialization()` builder, so now the only way you can access it is if you use `register()` directly (like it's currently done for `json()` builder.

On a side note, once this PR is merged and settled, do you guys have any other formats in mind that you think would be useful to add support for? I quite enjoyed implementing CBOR support and with the current framework I believe adding any extra kotlinx.serialization formats would be a trivial matter.

rsinukov:
Thanks for your work!
I briefly checked it and it looks good in general. I will properly review it later this week.

mpetuska:
@rsinukov a gentle nudge to review this. Would really like to see it in 2.0.0

ktor-client/ktor-client-plugins/ktor-client-content-negotiation/common/src/io/ktor/client/plugins/JsonContentTypeMatcher.kt
```diff
@@ -1,21 +0,0 @@
-/*
```
rsinukov, line 1: Please keep this class for backwards compatibility.

ktor-shared/ktor-shared-serialization-kotlinx/common/src/io/ktor/shared/serialization/kotlinx/KotlinxSerializationConverter.kt
```diff
@@ -49,7 +60,10 @@ public class KotlinxSerializationConverter(private val format: SerialFormat = De
             return result
         }
 
-        val guessedSearchSerializer = guessSerializer(value, format.serializersModule)
+        val matchedSerializer = extraSerializerMatchers.entries.firstOrNull { (_, matcher) ->
```
rsinukov, line 63: Am I right that this is used only for `JsonElement`? If yes, can you test that serialization works without this code? Then we can delete it and simplify the logic here. My guess that it was required for some old version of the library, but now it's redundant. 
I have some other concerns about it, but lets see if we can just delete it first.

rsinukov (changes requested):
Thanks for the great work. Just a couple of small comments.

ktor-shared/ktor-shared-serialization-kotlinx/common/src/io/ktor/shared/serialization/kotlinx/KotlinxSerializationConverter.kt
mpetuska: Yes and no. At the moment it's only an escape hatch for `json()` and is not used by `cbor()`. I can check if `json()` works without it when receiving JsonObject. Will report back.

ktor-client/ktor-client-plugins/ktor-client-content-negotiation/common/src/io/ktor/client/plugins/JsonContentTypeMatcher.kt
mpetuska: But it was never released, was it? ContentNegotiation feature was only added in 2.0.0

mpetuska: Happy to add it back, just want to make sure you don;t take up on unnecessary deprecation cycle.

rsinukov: you are right, thanks for correcting me

ktor-shared/ktor-shared-serialization-kotlinx/common/src/io/ktor/shared/serialization/kotlinx/KotlinxSerializationConverter.kt
mpetuska: Sorry for the delay - had a busy week. I finally got a chance to test JsonElement serialization without `match` API and it works just fine. Removed that API altogether now.

mpetuska:
@rsinukov ready for another (hopefully final) review. Due to force-push earlier this week I had quite a few merge conflicts to resolve, so it'd be great if you could keep that in mind when reviewing to make sure I didn't break anything unrelated with my conflict resolution.

mpetuska:
@e5l , @rsinukov does the above means that the changes will not be accepted for 2.0.0 afterall?

e5l:
Sorry, I assumed that GitHub will forward all PRs. Let me fix it

mpetuska:
@rsinukov, @e5l  merge conflicts from `main` solved. Ready for review again.

rsinukov:
@mpetuska can you please rebase again to keep only the relevant changes here? Sorry for the extra work for you.

mpetuska:
Due to numerous conflict resolutions rebasing is now a mess and it was quicker for me to just replicate changes on fresh branch. Closing as superseeded by #2628

PR #2617, opened 2021-09-22
KTOR-1686 Provide API for Intellij debugger to determine plugin information in runtime

Ololoshechkin:
**Subsystem**
Server


ktor-utils/common/src/io/ktor/util/debug/IdeDebugChecker.kt
```diff
@@ -0,0 +1,13 @@
...
+ */
+
+package io.ktor.util.debug
+
+internal expect object IntellijIdeaDebugDetector {
+    /**
+     * Checks whether Intellij Idea debugger is conected to the current Ktor server.
+     * May return true only for JVM debugger.
+     * */
+    val isDebuggerConnected: Boolean
```
Ololoshechkin, line 12: Note: in order to execute this code once the pipeline.execute is called, I have to make this common module.
But on the other hand:
a) idk if it's possible to easily check debugger connection on JS or Native
b) currently, there seems to be not much use of it for non-jvm

PR #2625, opened 2021-10-01
KTOR-730 Support merging urls in default request

rsinukov:


e5l:
@hhariri, @andreyaksenov could you proofread the public API docs?

PR #2631, opened 2021-10-05
KTOR-3144 Add Apple Silicon targets

hfhbd:
**Subsystem**
Client

**Motivation**
https://youtrack.jetbrains.com/issue/KTOR-3144
All Apple Silicon targets are missing.

**Solution**
Add all Apple Silicon targets.


hfhbd:
Currently, `macosX64` is very often hardcoded. `HostManager` already has `enabledByHost` property. Maybe it could be used instead, without checking the host again in many modules.

hfhbd:
I will split this PR anyway, focusing on `simulator` targets only. We could discuss other changes in the new PR after opening :)

PR #2637, opened 2021-10-11
KTOR-489 Add XML Support for ContentNegotiation Feature

e5l:
Fix [KTOR-489](https://youtrack.jetbrains.com/issue/KTOR-489)


e5l:
Hey @pdvrieze, it would be great if you also take a look. Thanks for providing a good XML library!

pdvrieze:
One question I have is, how is this different/better than `io.github.pdvrieze.xmlutil:ktor`. That library does mostly the same, except that it actually supports using streams/readers/writers as sources/destinations, rather than delegate it to the kotlinx.serialization support that doesn't support streams.

I have no problem with moving that module to ktor.

pdvrieze:
Btw. I've updated the dev branch with an (for now experimental) cross-platform parser (ported/adapted from kxml2). This means native is now supported (experimentally). It should work on the same targets as kotlinx.serialization (list copied from the library), but as konan doesn't do cross-compilation to different host systems, I can't create macos targeting libraries (mingw should be possible, but I haven't tried yet).

PR #2649, opened 2021-10-15
KTOR-3257 Add type info and refactor transformBody methods

Ololoshechkin:
https://youtrack.jetbrains.com/issue/KTOR-3257


e5l:
Please check tests, code style and API dump before merging

PR #2653, opened 2021-10-19
KTOR-472 Update Netty to 4.1.69.Final

e5l:
Fix [KTOR-472](https://youtrack.jetbrains.com/issue/KTOR-472)


e5l:
@rsinukov, thanks! will do

PR #2659, opened 2021-10-24
KTOR-3290 Add support for testing Server-Sent Events (SSE)

goncalossilva:
**Subsystem**
Server test host.

**Motivation**
Ktor does not facilitate SSE testing, and doing it all on the application side is cumbersome. [KTOR-3290](https://youtrack.jetbrains.com/issue/KTOR-3290) goes into a bit more detail.

**Solution**
The solution is inspired by `handleWebSocketConversation`. In SSE's case the data flow is unidirectional, so only one channel is setup and used, but most other details, such as keeping the connection alive, or executing server and client side in parallel, remain.

I am using a variant of this successfully [here](https://github.com/goncalossilva/ffs/blob/359c015606680e48d08e9e32d393e0e485c0cedd/ffs-server/src/test/kotlin/ext/TestApplicationEngineExt.kt#L16-L60), albeit with more difficulty due to not having access to `call.response.responseChannelDeferred` on the application side.

**Question**
Something that is not obvious to me is where to add tests for this. `handleWebSocketConversation` is being tested from within the `ktor-server-websockets` module, which also provides the plugin. But there is no plugin for SSE.

Would `TestApplicationEngineTest` inside the `ktor-server-tests` module be the best place?

### DIFF
--- a/ktor-server/ktor-server-test-host/jvm/src/io/ktor/server/testing/TestApplicationEngine.kt
+++ b/ktor-server/ktor-server-test-host/jvm/src/io/ktor/server/testing/TestApplicationEngine.kt
@@ -265,6 +265,59 @@ class TestApplicationEngine(
         return call
     }
 
+    private fun createServerSentEventsCall(uri: String, setup: TestApplicationRequest.() -> Unit): TestApplicationCall =
+        createCall(closeRequest = false) {
+            this.uri = uri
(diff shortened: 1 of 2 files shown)


e5l:
@marychatte, could you check if it's useful and rebase if needed?

marychatte:
@goncalossilva, thank you for issue. We added support for SSE in Ktor, and we also added it for testing, you can see some examples in [PR](https://github.com/ktorio/ktor/pull/3930)

PR #2660, opened 2021-10-25
KTOR-3346 ContentNegotiation plugins don't accept null-responses from ContentConverts

T-Fowl:
**Subsystem**
Client

**Motivation**
The `ContentConverter` documentation as well as the server implementation of `ContentNegotiation` both specify that `ContentConverter::(de)serialize` should return null when the converter is not suitable for the current content-type/body. This is to allow other registrations / converters to (de)serialize the content.

The client implementation of `ContentNegotiation` only tried (de)serialization with the first matching registration.

**Solution**
Instead of only (de)serializing with the first matching registration, this PR changes the behaviour to filter out matching registrations and try (de)serialization on each until one succeeds.

### DIFF
--- a/ktor-client/ktor-client-plugins/ktor-client-content-negotiation/common/src/io/ktor/client/plugins/ContentNegotiation.kt
+++ b/ktor-client/ktor-client-plugins/ktor-client-content-negotiation/common/src/io/ktor/client/plugins/ContentNegotiation.kt
@@ -120,11 +120,14 @@ public class ContentNegotiation internal constructor(
 
                 val contentType = context.response.contentType() ?: return@intercept
                 val registrations = plugin.registrations
-                val registration = registrations
-                    .firstOrNull { it.contentTypeMatcher.contains(contentType) } ?: return@intercept
+                val matchingRegistrations = registrations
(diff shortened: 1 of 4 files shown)


T-Fowl:
Added your requests and rebased off main

> 1. Add test.

I added unit tests to cover both my changes and some of the other converter selection procedures in the client plugin

> 3. I noticed that the server side plugin has correct behaviour for sending, but for receiving. Do you mind updating it as well?

Changed however I did not write extra tests for the server plugin

Let me know if anything more needs to be changed - such as the test function names which I was not sure if they fit your code style guidelines

rsinukov:
I see that test are not compiling in `:ktor-client:ktor-client-plugins:ktor-client-content-negotiation:compileTestKotlinJvm`. Can you check it please?

T-Fowl:
Looks like I forgot to removed the `shared` package prefix when rebasing off main. Hopefully this should fix it

rsinukov (approved):
Hi @T-Fowl!
Me and @lolilofit are working on the same area, so we rebased and fixed some small issues for you here, to merge it faster and avoid conflicts.
The code LGTM, thanks for your work!

PR #2670, opened 2021-10-27
Support other http methods than POST for OAuth 1a

drampelt:
**Subsystem**
Server -> auth -> oauth

**Motivation**
Some OAuth 1a providers require HTTP methods or than POST for the authorization flow. Specifically, [Schoology](https://developers.schoology.com/api-documentation/authentication) (one of the providers my company needs to integrate with) only supports GET requests and fails with POST requests.

**Solution**
Add an `httpMethod` setting to control what method is used for the requests


drampelt:
I'm going to put this on hold for now, after further testing the provider I'm trying to work with violates the OAuth 1a spec in multiple other ways. I don't want to make these changes directly in ktor just to support a single provider that doesn't follow the spec.

rsinukov:
Sounds good, thanks anyway!

Issue #1252, opened 2019-07-29
List of all routes


sksamuel:
https://github.com/ktorio/ktor/issues/788**Subsystem**
Server

**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
Things like spring actuator are able to list all the endpoints in the current system. Something like this for ktor would be very helpful.

**Motivation to include to ktor**
A very common usage so makes sense to build directly into ktor.

CarrotCodes:
I was interested logging all computed routes on service startup, in a cheap way, so did a small investigation.

You can traverse the tree of `Route`s, picking out leaves by filtering on a property, such as the route having a method selector. Assuming there can't be cycles:

```kotlin
fun Application.ktorMain() {
    // set up omitted
    val root = feature(Routing)
    val allRoutes = allRoutes(root)
    val allRoutesWithMethod = allRoutes.filter { it.selector is HttpMethodRouteSelector }
    allRoutesWithMethod.forEach {
        logger.info("route: $it")
    }
}

fun allRoutes(root: Route): List<Route> {
    return listOf(root) + root.children.flatMap { allRoutes(it) }
}
```

Example output:
```
22:41:42.148 [main] INFO  Service - route: /(authenticate adoptanimals-session-cookie)/shelter/{shelterVisualId}/(method:GET)
22:41:42.148 [main] INFO  Service - route: /private/healthcheck/(method:GET)
```

I've no doubt there'll be a better way built in eventually, but this did what we needed for now.

oleg-larshin:
Please check the <a href="https://youtrack.jetbrains.com/issue/KTOR-581">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

PR #2673, opened 2021-10-29
KTOR-581 Add API to get list of all method routes

BOOMeranGG:
**Subsystem**
ktor-server-core

**Motivation**
See: #1252 or [YouTrack Issue](https://youtrack.jetbrains.com/issue/KTOR-581)

**Solution**
Find of all routes of the http method using recursion DFS for a tree.
Usage example:

```
val root = plugin(Routing)
val allMethodRoutes = root.getAllHttpMethodRoutes()
```
This can be used on any node of routing tree.


marychatte:
Duplicated by [pull request #3287](https://github.com/ktorio/ktor/pull/3287)

PR #2682, opened 2021-11-08
KTOR-3217 Add support for the new native memory model

e5l:
Fix [KTOR-3217](https://youtrack.jetbrains.com/issue/KTOR-3217)


ktor-network/nix/src/io/ktor/network/selector/SelectUtils.kt
```diff
@@ -95,7 +95,7 @@ internal class SelectorHelper {
             processSelectedEvents(watchSet, closeSet, completed, readSet, writeSet, errorSet)
         }
 
-        val exception = CancellationException("Selector closed").freeze()
+        val exception = CancellationException("Selector closed").apply { freeze() }
```
rsinukov, line 98: should it be `makeShared`?

rsinukov (approved):
LGTM

rsinukov:
Please check CI build

olme04 (commented):
Minor: we can use check for experimental MM in [shared collections](https://github.com/ktorio/ktor/blob/4a4f0cd584c948a24c79c2809d521e79c0c93a62/ktor-utils/common/src/io/ktor/util/collections/CollectionUtils.kt#L10), to use simple maps instead of concurrentMap based on atomics

e5l:
@olme04, nice observation. Will replace! Thank you

olme04:
BTW, `IS_PLATFORM_BIG_ENDIAN` seems can be replaced by [Platform.isLittleEndian](https://github.com/JetBrains/kotlin/blob/master/kotlin-native/runtime/src/main/kotlin/kotlin/native/Platform.kt#L63) (with inverting of check of course)

ktor-network/nix/src/io/ktor/network/selector/SelectUtils.kt
e5l: Thanks, fixed

PR #2684, opened 2021-11-09
KTOR-383 Request may not include Host parameter

aschugunov:
**Subsystem**
Core

**Motivation**
[KTOR-383](https://youtrack.jetbrains.com/issue/KTOR-383)

**Solution**
The parser should check the version of HTTP and throw parse exception accordingly lacking host header and HTTP 1.1


e5l:
Hey @aschugunov, we're not going to restrict queries of this type.

PR #2686, opened 2021-11-10
KTOR-3364 Flush Header With Netty

lolilofit:
**Subsystem**
Server

**Motivation**
Headers are only flushed after first byte is written [KTOR-3364](https://youtrack.jetbrains.com/issue/KTOR-3364). We need the same behavior for `Jetty`, `Netty` and other engines, but `Netty` is different now.

**Solution**
Flush respond headers right after writing them to the channel.


e5l (approved):
lgtm

PR #2688, opened 2021-11-11
KTOR-3464 Set missing simple properties in Netty EngineMain

trond-oydna:
**Subsystem**
Server

**Motivation**
Ability to configure all of the simple Netty engine properties via `application.conf`.
I was surprised to discover after reading the docs that `runningLimit` was not set in `EngineMain`.

**Solution**
Like with the other simple properties, read from application configuration and assign to `NettyApplicationEngine` if present and valid.


rsinukov (commented):
Hi, and thanks for the PR!
The code looks good to me, but do you mind creating a ticket in YouTrack and mentioning it in the PR?

trond-oydna:
Great!
No problem, I've created an issue now and updated commit message and PR title now.

rsinukov (approved):

PR #2693, opened 2021-11-16
KTOR-3422 Upgrade kotlin to 1.6.0

e5l:
Fix [KTOR-3422](https://youtrack.jetbrains.com/issue/KTOR-3422)


lolilofit (approved):
lgtm

hfhbd:
What about coroutines, especially for new MM with 1.6.0?

e5l:
@hfhbd, the new MM support is merged. All you need to do is take `kotlinx.coroutines` version with support(when it will be released) and provide it as a dependency. You also need to specify the flag to enable a new memory model.

PR #2697, opened 2021-11-19
KTOR-524 Session cookie with BASE64 encoding fails to set correct cookie

lolilofit:
**Subsystem**
Server

**Motivation**
Session cookie with BASE64 encoding fails to set correct cookie. [KTOR-524](https://youtrack.jetbrains.com/issue/KTOR-524)

**Solution**
Pass cookie encoding to `RequestCookies.get` call


e5l (changes requested):
Hey @lolilofit, thanks for the PR. Could you check the CI?

lolilofit:
> Hey @lolilofit, thanks for the PR. Could you check the CI?

Sure, CI failed with test with old testing API. I've rewrited it to the new one.

e5l (approved):

PR #2700, opened 2021-11-23
KTOR-3459 Add check if plugin is installed for `WebSocket` builders

rsinukov:


ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/MissingClientPluginException.kt
```diff
@@ -0,0 +1,11 @@
+/*
+ * Copyright 2014-2021 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.client.plugins
+
+/**
+ * Thrown when trying to access a plugin that wasn't installed
+ */
+public class MissingClientPluginException(plugin: HttpClientPlugin<*, *>) :
```
e5l, line 10: Could you tell me why do we need a new type of exception? It looks like a runtime diagnostic, not a thing you want to catch. I would use illegal state instead

e5l (changes requested):

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/MissingClientPluginException.kt
rsinukov: It was simpler to test that the correct exception is thrown and not some internal error. I updated the test and removed the new class.

e5l (approved):

PR #2707, opened 2021-11-26
KTOR-3476 Don't install jansi if log colours disabled

Ololoshechkin:


e5l (changes requested):
Please check the CI build

rsinukov (approved):

PR #2709, opened 2021-11-30
Fix HttpTimeoutTest

e5l:


lolilofit (changes requested):
Please check CI, it fails with Java 11 on macOS

lolilofit (approved):
lgtm

PR #2710, opened 2021-11-30
KTOR-3492 KTOR-3503 Native CIO server fixes

rsinukov:
* Fix flush in `copyTo`
* Fix chunked response in CIO

### DIFF
--- a/ktor-http/ktor-http-cio/common/src/io/ktor/http/cio/HttpHeadersMap.kt
+++ b/ktor-http/ktor-http-cio/common/src/io/ktor/http/cio/HttpHeadersMap.kt
@@ -5,7 +5,8 @@
 package io.ktor.http.cio
 
 import io.ktor.http.cio.internals.*
-import io.ktor.util.*
+import io.ktor.util.collections.*
+import io.ktor.utils.io.concurrent.*
 import io.ktor.utils.io.pool.*
 import kotlin.native.concurrent.*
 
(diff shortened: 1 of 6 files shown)


ktor-http/ktor-http-cio/common/src/io/ktor/http/cio/HttpHeadersMap.kt
```diff
@@ -29,17 +30,17 @@ private const val HEADER_SIZE = 8
 private const val HEADER_ARRAY_POOL_SIZE = 1000
 
 @ThreadLocal
-private val EMPTY_INT_ARRAY = IntArray(0)
+private val EMPTY_INT_LIST = mutableListOf<Int>()
```
e5l, line 33: should we use `emptyList()` from stdlib instead?

e5l (changes requested):
LGTM in general, please check a small comment

ktor-http/ktor-http-cio/common/src/io/ktor/http/cio/HttpHeadersMap.kt
rsinukov: The instance of `emptyList()` is `List`, but we need `MutableList` here

e5l:
Oh, sorry. I'm not sure how I close it

ktor-http/ktor-http-cio/common/src/io/ktor/http/cio/HttpHeadersMap.kt
e5l: Should we create a new list every time? I'm afraid it can be accidentally freezed

rsinukov:
It wasn't you =) I merged PR that was a target branch of this PR, so GH closed this one automatically.

ktor-http/ktor-http-cio/common/src/io/ktor/http/cio/HttpHeadersMap.kt
rsinukov: It's thread local and should not be modified. If it is modified, it will crash with `IndexOutOfBoundException`, since there is no adding elements to it, only setting

e5l (approved):

PR #2708, opened 2021-11-30
KTOR-3510 Fix `testErrorHandling` with JS

lolilofit:
**Motivation**
`assertFailsWith` fails with `IllegalStateException` in `testErrorHandling`.

**Solution**
Remove `assertFailsWith` usage in websocket test

PR #2712, opened 2021-12-01
Fix failing tests

e5l:


ktor-client/ktor-client-plugins/ktor-client-websockets/common/test/io/ktor/client/tests/plugins/WebSocketRemoteTest.kt
```diff
@@ -121,21 +121,27 @@ class WebSocketRemoteTest : ClientLoader() {
...
     @Test
-    fun testErrorHandling() = clientTests(listOf("Android", "Apache", "Curl")) {
+    fun testErrorHandling() = clientTests(skipEngines) {
         config {
             install(WebSockets)
         }
 
         test { client ->
-            assertFailsWith<CustomException> {
+            try {
```
olme04, line 130: https://github.com/ktorio/ktor/pull/2708#discussion_r759910292
> BTW, this should work with assertFailsWith(CustomException::class) instead of using reified version (had the same issue in rsocket-kotlin repository

lolilofit: @olme04 Thanks for your help!

rsinukov (approved):

PR #2714, opened 2021-12-01
KTOR-3487 Drop client.get(plugin) operator

rsinukov:


ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/HttpClientPlugin.kt
```diff
@@ -32,24 +32,18 @@ public interface HttpClientPlugin<out TConfig : Any, TPlugin : Any> {
 }
 
 /**
- * Try to get the [plugin] installed in this client. Returns `null` if the plugin was not previously installed.
+ * Returns a [plugin] installed in this client. Returns `null` if the plugin was not previously installed.
  */
 public fun <B : Any, F : Any> HttpClient.plugin(plugin: HttpClientPlugin<B, F>): F? =
```
olme04, line 37: Does it make sense to break it in 2.0 and do `plugin` and `pluginOrNull` naming?
same naming strategy is used in server plugins

e5l: +1

e5l (changes requested):
let's make it the same naming as in the server

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/HttpClientPlugin.kt
rsinukov: done

e5l (approved):

PR #2716, opened 2021-12-03
KTOR-536 Support for adding values to the MDC later on in the pipeline.

rsinukov:
Call `withMdcBlock` multiple times, to add new MDC entries to the context.
Also, migrate test to the new API


e5l (approved):
lgtm

PR #2728, opened 2021-12-09
KTOR-3544 Improve HttpRequestRetry plugin

rsinukov:


ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/HttpRequestRetry.kt
```diff
@@ -130,37 +182,52 @@ public class HttpRequestRetry internal constructor(configuration: Configuration)
...
          * and should return the number of milliseconds to wait before retrying.
          */
-        public fun delayMillis(block: (retry: Int) -> Long) {
-            delayMillis = block
+        public fun delayMillis(
+            respectRetryAfterHeader: Boolean = true,
+            block: DelayContext.(retry: Int) -> Long
+        ) {
+            delayMillis = {
+                val retryAfter = response?.headers?.get(HttpHeaders.RetryAfter)?.toLong()
```
apatrida, line 190: if `RetryAfter` is the Date format instead of numeric this will cause an exception.  I think it is ok to not handle the Date format of the header (see: https://www.geeksforgeeks.org/http-headers-retry-after/) but the `toLong()` needs protection.

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/HttpRequestRetry.kt
```diff
@@ -130,37 +182,52 @@ public class HttpRequestRetry internal constructor(configuration: Configuration)
...
+                if (respectRetryAfterHeader && retryAfter != null) {
```
apatrida, line 191: This should be `maxOf(block(), retryAfter ?: 0)` or similar.  The server might not know you are in exponential fall-off and it is try to pause clients or slow them down, but at the same time your logic of falling off should allow to be even slower since the client has more context about repeated calls than the server.

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/HttpRequestRetry.kt
```diff
@@ -81,11 +123,18 @@ public class HttpRequestRetry internal constructor(configuration: Configuration)
             shouldRetryOnException = { _, _ -> false }
         }
 
+        /**
+         * Modifies request before retrying.
```
andreyaksenov, line 127: > Modifies a request ...

andreyaksenov (changes requested):
Several small notes

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/HttpRequestRetryTest.kt
```diff
@@ -56,6 +66,39 @@ class HttpRequestRetryTest {
...
+                    assertEquals("2", it.headers["X-RETRY-COUNT"])
+                    respondError(HttpStatusCode.ServiceUnavailable)
+                }
+                addHandler {
+                    assertEquals("3", it.headers["X-RETRY-COUNT"])
+                    respondOk()
+                }
+            }
+            install(HttpRequestRetry) {
+                modifyRequest { it.headers.append("X-RETRY-COUNT", retryCount.toString()) }
```
e5l, line 91: Should we make a retry handler with the request builder instead? I'm not sure that it's a most comfort way of modifying the request

e5l: We can make a small design discussion about that

e5l (changes requested):

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/HttpRequestRetry.kt
```diff
@@ -31,13 +33,52 @@ import kotlin.random.*
...
  *      delayMillis { retry -> retry * 3000 } // will retry in 3, 6, 9, etc. seconds
+ *      modifyRequest { it.headers.append("X_RETRY_COUNT", retryCount.toString()) }
  * }
  * ```
  */
 public class HttpRequestRetry internal constructor(configuration: Configuration) {
 
     /**
-     * Data for the [HttpRequestRetryEvent] event. Contains non-null [response] or [cause], but not both.
+     * Context for [HttpRequestRetry.Configuration.shouldRetry]
```
andreyaksenov, line 43: > A context ...

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/HttpRequestRetry.kt
```diff
@@ -31,13 +33,52 @@ import kotlin.random.*
...
+     * and [HttpRequestRetry.Configuration.shouldRetryOnException]
+     */
+    public class ShouldRetryContext(
+        /**
+         * Retry count starting from 1
```
andreyaksenov, line 48: > A retry count that starts from 1

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/HttpRequestRetry.kt
```diff
@@ -94,7 +144,10 @@ public class HttpRequestRetry internal constructor(configuration: Configuration)
          * Specifies retry logic for failed requests. The [block] accepts [HttpRequestBuilder]
          * and an [Throwable] and should return true if this request should be retried.
```
andreyaksenov, line 145: an [Throwable] -> [Throwable] (without the article)

andreyaksenov (changes requested):

andreyaksenov (approved):

PR #2729, opened 2021-12-10
Fix TestLocal repo URL is incorrect.

docee:
**Subsystem**
Maven publication

**Motivation**
The testLocal repo url in Publication.kt is incorrectly written, which will result in publishing to the `$rootProject.buildDir/m2` folder of all subprojects.

**Solution**
Replace `$rootProject.buildDir/m2` with `${rootProject.buildDir}/m2`


docee:
@e5l

PR #2731, opened 2021-12-10
KTOR-3538 AttributeKey instance is identified by its identity instead of its name

lolilofit:
Ticket: [KTOR-3538](https://youtrack.jetbrains.com/issue/KTOR-3538)

**Solution**
Add `equals` and `hashCode` to the `AttributeKey` class.
Deprecate `EquatableAttributeKey`


ktor-utils/common/src/io/ktor/util/Attributes.kt
```diff
@@ -11,13 +11,30 @@ package io.ktor.util
...
+    }
 }
 
 /**
  * Version of [AttributeKey] that overrides [equals] and [hashCode] using [name]
  * @param T is type of the value stored in the attribute
  * @param name is a name of the attribute
  */
+@Suppress("DEPRECATION")
+@Deprecated("Please use `AttributeKey` class instead")
```
rsinukov, line 37: Please replace internal usages of this class

rsinukov (changes requested):

olme04:
Maybe it's better to make `EquatableAttributeKey` a deprecated typealias for `AttributeKey`?
And as we a here, I would propose to add check in `init` that name isn't blank (`require(name.isNotBlank())`) and adapt `toString` according this change. Now, as all keys are the same based on key, there is no any sense to have possibility to specify blank name

lolilofit:
> Maybe it's better to make `EquatableAttributeKey` a deprecated typealias for `AttributeKey`? And as we a here, I would propose to add check in `init` that name isn't blank (`require(name.isNotBlank())`) and adapt `toString` according this change. Now, as all keys are the same based on key, there is no any sense to have possibility to specify blank name

@olme04 
Thanks for your comment! I've added name check in init and referenced to the `AttributeKey` in `EquatableAttributeKey` with replaceWith

olme04:
@lolilofit and what about typealias instead of `open class` and inheriting? Now, both classes are fully identical in implementation.

ktor-utils/common/src/io/ktor/util/Attributes.kt
```diff
@@ -7,17 +7,43 @@ package io.ktor.util
 /**
  * Specifies a key for an attribute in [Attributes]
  * @param T is type of the value stored in the attribute
- * @param name is a name of the attribute for diagnostic purposes
+ * @param name is a name of the attribute for diagnostic purposes. Couldn't be blank
  */
 public open class AttributeKey<T>(public val name: String) {
-    override fun toString(): String = if (name.isEmpty()) super.toString() else "AttributeKey: $name"
+    init {
+        if(name.isEmpty()) {
```
rsinukov, line 14: Please fix formatting here

ktor-utils/common/src/io/ktor/util/Attributes.kt
```diff
@@ -7,17 +7,43 @@ package io.ktor.util
 /**
  * Specifies a key for an attribute in [Attributes]
  * @param T is type of the value stored in the attribute
- * @param name is a name of the attribute for diagnostic purposes
+ * @param name is a name of the attribute for diagnostic purposes. Couldn't be blank
  */
 public open class AttributeKey<T>(public val name: String) {
```
rsinukov, line 12: While you are here, can you please change the type of `<T>` to `<T : Any>`? `Attributes` class uses this restriction

lolilofit:
> @lolilofit and what about typealias instead of `open class` and inheriting? Now, both classes are fully identical in implementation.

@olme04 
I've added typealias with new commits. Thanks for the help!

lolilofit:
@andreyaksenov, please could you check public API?

ktor-utils/common/src/io/ktor/util/Attributes.kt
```diff
@@ -7,24 +7,22 @@ package io.ktor.util
 /**
  * Specifies a key for an attribute in [Attributes]
  * @param T is type of the value stored in the attribute
- * @param name is a name of the attribute for diagnostic purposes
+ * @param name is a name of the attribute for diagnostic purposes. Couldn't be blank
```
andreyaksenov, line 10: Couldn't be blank -> Can't be blank

ktor-utils/common/src/io/ktor/util/Attributes.kt
```diff
@@ -36,6 +34,18 @@ public class EquatableAttributeKey<T>(name: String) : AttributeKey<T>(name) {
     }
 }
 
+/**
+ * Version of [AttributeKey] that overrides [equals] and [hashCode] using [name]
```
andreyaksenov, line 38: Version -> A version

ktor-utils/common/src/io/ktor/util/Attributes.kt
```diff
@@ -36,6 +34,18 @@ public class EquatableAttributeKey<T>(name: String) : AttributeKey<T>(name) {
...
+ * @param T is type of the value stored in the attribute
```
andreyaksenov, line 39: is type -> is a type

ktor-utils/common/src/io/ktor/util/Attributes.kt
```diff
@@ -36,6 +34,18 @@ public class EquatableAttributeKey<T>(name: String) : AttributeKey<T>(name) {
...
+ */
+@Suppress("DEPRECATION")
+@Deprecated(
+    "Please use `AttributeKey` class instead",
+    replaceWith = ReplaceWith("AttributeKey")
+)
+public typealias EquatableAttributeKey<T> = AttributeKey<T>
+
 /**
  * Create attributes instance suitable for the particular platform
```
andreyaksenov, line 50: > Creates an attributes instance ...

andreyaksenov (changes requested):

andreyaksenov (approved):

rsinukov (approved):

PR #2732, opened 2021-12-10
Fix port order

e5l:


ptitjes:
@e5l Hi Leonid. Just so you know, I fixed all this in https://github.com/ktorio/ktor/pull/2603. :D

PR #2743, opened 2021-12-16
KTOR-3497 HttpCookies: parse / in the name of a cookie

lolilofit:
Ticket: [KTOR-3497](https://youtrack.jetbrains.com/issue/KTOR-3497)

**Motivation**
Current cookie implementation supports [rfc6265](https://datatracker.ietf.org/doc/html/rfc6265#page-13), but browsers support old [netscape spec](https://curl.se/rfc/cookie_spec.html) in cookie names

**Solution**
Support netscape spec: allow characters ()<>@:/\"[]? in cookie names. 
Also browsers support comma and white space as [stackowerflow question](https://stackoverflow.com/questions/1969232/what-are-allowed-characters-in-cookies) says, so this changes allow them too.

### DIFF
--- a/ktor-http/common/src/io/ktor/http/Cookie.kt
+++ b/ktor-http/common/src/io/ktor/http/Cookie.kt
@@ -92,7 +92,7 @@ public fun parseServerSetCookieHeader(cookiesHeader: String): Cookie {
 }
 
 @ThreadLocal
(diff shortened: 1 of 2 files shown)


rsinukov (approved):
Please rebase on latest main

PR #2748, opened 2021-12-21
KTOR-1619 Use placeholder text if response body decoding fails in default validator

arthurljones:
**Subsystem**
Client

**Motivation**
Resolves [KTOR-1619](https://youtrack.jetbrains.com/issue/KTOR-1619)
HTTP responses with non-2XX status codes and non-text bodies cause an unexpected exception in default response validation. This is because the validator attempts to decode the body as text to generate the exception message. This causes issues with endpoints that respond with a generic error image when requesting a non-existent record.

**Solution**
Attempt to decode the body in a `try` block, then on `MalformedInputException`, use a placeholder string to represent the non-decodable body in the resulting exception message. Also includes unit test written by original issue creator Takahiro Menju.


rsinukov (approved):
LGTM
Sorry for the delay, this PR got lost =( Next time please feel free to ping one of the maintainers. 
I rebased and updated this PR to not ask you to come back to it after half a year. Will merge after CI run

PR #2757, opened 2021-12-24
KTOR-1435: add support for client request single line logging

cbrown184:
**Subsystem**
Logging plugin

**Motivation**
The client logging plugin does not allow logging to a single line easily - this can be a problem when using elastic search etc. making centralised logs very messy. https://youtrack.jetbrains.com/issue/KTOR-1435

**Solution**
Add a flush method to the Logger interface (defaulting to nothing) which is called by Logging.kt.doneLogging() - this way the user can supply their own Logger implementation and gather the logs in a client request before Flushing to a single log line.


rsinukov (changes requested):
Thanks for the PR!
But there are couple things missing.
1. Please provide API doc for the new method
2. Please add tests.
3. Optional. It would be great to have provide logger single line logger out of the box. Can you imoplement it

rsinukov:
Closing this PR because of inactivity

PR #2762, opened 2021-12-30
KTOR-3634 Retry-after header value should be in seconds

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-3634

### DIFF
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/HttpRequestRetry.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/HttpRequestRetry.kt
@@ -189,7 +189,7 @@ public class HttpRequestRetry internal constructor(configuration: Configuration)
         ) {
             delayMillis = {
                 if (respectRetryAfterHeader) {
-                    val retryAfter = response?.headers?.get(HttpHeaders.RetryAfter)?.toLongOrNull()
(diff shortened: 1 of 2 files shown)


ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/HttpRequestRetry.kt
```diff
@@ -189,7 +189,7 @@ public class HttpRequestRetry internal constructor(configuration: Configuration)
         ) {
             delayMillis = {
                 if (respectRetryAfterHeader) {
-                    val retryAfter = response?.headers?.get(HttpHeaders.RetryAfter)?.toLongOrNull()
+                    val retryAfter = response?.headers?.get(HttpHeaders.RetryAfter)?.toLongOrNull()?.times(1000)
```
Stexxe, line 192: What if it's a date?

e5l (changes requested):
please check the CI build

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/HttpRequestRetry.kt
rsinukov: We don't support a date for now, because there is no implementation of Http Date parser available for multiplatform.

e5l (approved):
lgtm

PR #2768, opened 2022-01-07
KTOR-3650 Fixed `appendIfNameAbsent` and `appendIfNameAndValueAbsent`

rsinukov:


ktor-utils/common/src/io/ktor/util/StringValues.kt
```diff
@@ -395,15 +395,15 @@ public fun StringValuesBuilder.appendAll(builder: StringValuesBuilder): StringVa
  * Appends [name] [value] pair if there are no values associated with [name]
  */
 public fun StringValuesBuilder.appendIfNameAbsent(name: String, value: String): StringValuesBuilder = apply {
-    if (!contains(name)) return@apply
+    if (contains(name)) return@apply
```
Stexxe, line 398: Could you please just get rid of `return` since there is only one if statement?

e5l:
Hey, could you check failing test?

e5l (changes requested):
please check the CI build

e5l (approved):
lgtm

PR #2769, opened 2022-01-10
Initializer should be initialized eagerly

ilgonmic:
**Subsystem**
Client

**Motivation**
Since 1.6.20 there is lazy initialization of properties. So now using of properties as side effect is not possible.
For compatibility annotation `@EagerInitialization` can be used.

**Solution**
Using of `@EagerInitialization` on side-effect property


ilgonmic:
cc @e5l this commit needs to be in master after upgrade to Kotlin 1.6.20

PR #2772, opened 2022-01-11
Use DEFAULT_COMPRESSION for java.util.zip.Deflater

evestera:
Use DEFAULT_COMPRESSION for java.util.zip.Deflater.
Rather than BEST_COMPRESSION. BEST_COMPRESSION is generally many times
slower than DEFAULT_COMPRESSION for miniscule improvements to compressed
size, making BEST_COMPRESSION almost always the wrong choice when
compression time is part of the equation.

This should probably be configurable, since the optimal choice depends
on expected network speed, CPU speed, and which implementation the JDK
uses, but DEFAULT_COMPRESSION is a pretty good default.

**Subsystem**
ktor-utils

**Motivation**
Started down the rabbit hole that led me here because I noticed a ktor server being significantly slower to transfer a large response than pretty much any other server I tried (if using `Accept-Encoding: deflate, gzip`), and eventually narrowed down the main culprit to this.

**Solution**

```diff
-    val deflater = Deflater(Deflater.BEST_COMPRESSION, true)
+    val deflater = Deflater(Deflater.DEFAULT_COMPRESSION, true)
```

### DIFF
--- a/ktor-utils/jvm/src/io/ktor/util/Deflater.kt
+++ b/ktor-utils/jvm/src/io/ktor/util/Deflater.kt
@@ -64,7 +64,7 @@ public fun ByteReadChannel.deflated(
     coroutineContext: CoroutineContext = Dispatchers.Unconfined
 ): ByteReadChannel = GlobalScope.writer(coroutineContext, autoFlush = true) {
     val crc = CRC32()
-    val deflater = Deflater(Deflater.BEST_COMPRESSION, true)
+    val deflater = Deflater(Deflater.DEFAULT_COMPRESSION, true)
     val input = pool.borrow()
     val compressed = pool.borrow()
 


kibertoad:
Should this even be done on application level? Isn't it a best practice to do compression on reverse proxy level?

evestera:
> Should this even be done on application level? Isn't it a best practice to do compression on reverse proxy level? Hence being configurable from app configuration would be the best, but I agree that default should be switched to `DEFAULT_COMPRESSION`

This is only in use if you do `install(Compression)`, so when I mention that it should be configurable, I'm referring to the compression level. If you use a reverse proxy, doing compression there instead may be a good idea, particularly since `java.util.zip.Deflater` is not necessarily that fast.

evestera:
> LGTM, thanks for the PR Do you mind creating a ticket in YouTrack and mentioning it in the PR title?

Of course. Was a bit unsure if I should make one or not since CONTRIBUTING.md said to "Ensure that the description [...] refers to an existing ticket/bug **if applicable**" (emphasis mine), but made one now.

Issue here: https://youtrack.jetbrains.com/issue/KTOR-3680

PR #2773, opened 2022-01-11
KTOR-3659 Fix sockets on Android

Thomas-Vos:
https://youtrack.jetbrains.com/issue/KTOR-3659

Since https://github.com/ktorio/ktor/pull/2603 was merged the sockets no longer work on Android API level < 24.

This PR fixes the issue by only calling Java 1.7 APIs if they are available, and fall back to older APIs otherwise. 

I have not tested this PR yet on Android, will do that soon.

### DIFF
--- a/ktor-network/jvm/src/io/ktor/network/sockets/CIOReader.kt
+++ b/ktor-network/jvm/src/io/ktor/network/sockets/CIOReader.kt
@@ -63,7 +63,11 @@ internal fun CoroutineScope.attachForReadingImpl(
             pool.recycle(buffer)
             if (nioChannel is SocketChannel) {
                 try {
-                    nioChannel.shutdownInput()
+                    if (java7NetworkApisAvailable) {
+                        nioChannel.shutdownInput()
+                    } else {
+                        nioChannel.socket().shutdownInput()
+                    }
                 } catch (ignore: ClosedChannelException) {
                 }
             }
@@ -115,7 +119,11 @@ internal fun CoroutineScope.attachForReadingDirectImpl(
     } finally {
         if (nioChannel is SocketChannel) {
             try {
-                nioChannel.shutdownInput()
+                if (java7NetworkApisAvailable) {
+                    nioChannel.shutdownInput()
+                } else {
+                    nioChannel.socket().shutdownInput()
+                }
             } catch (ignore: ClosedChannelException) {
             }
         }
--- a/ktor-network/jvm/src/io/ktor/network/sockets/CIOWriter.kt
+++ b/ktor-network/jvm/src/io/ktor/network/sockets/CIOWriter.kt
@@ -62,7 +62,11 @@ internal fun CoroutineScope.attachForWritingImpl(
             pool.recycle(buffer)
             if (nioChannel is SocketChannel) {
                 try {
-                    nioChannel.shutdownOutput()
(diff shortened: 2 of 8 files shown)


Thomas-Vos:
@ptitjes if you have time, please review if this does not break anything for you. Java 1.7 APIs are called if available so I think all should work fine.

lolilofit (approved):
LGTM

olme04 (commented):
may be it's better to somehow extract those checks under some extension functions in one place, to not repeat same code in many different places ?
like this
```kotlin
val localAddress = if (java7NetworkApisAvailable) {
    channel.localAddress
} else {
    channel.socket().localSocketAddress
}
```

lolilofit:
@Thomas-Vos, please, also write a message here when you will check this code on Android

Thomas-Vos:
@olme04 sounds good, but I don't think that is really possible here. The `socket()` function is not shared between the classes. For example there is `DatagramChannel.socket()`, `ServerSocketChannel.socket()`, and `SocketChannel.socket()`. Not sure how to share code for those calls. If you have any suggestions, let me know,

Will get back with testing results in a moment.

olme04:
@Thomas-Vos even so, we can extract this for at least in one place, but for those 3 classes. Even if we not be able to share everything, we will have this `compatibility` things in one place

Thomas-Vos:
@lolilofit, just tested my project which uses uses Ktor UDP sockets and requires options like broadcast.

I built this PR locally and included it as a dependency. Ran the app on a Samsung Galaxy S5 mini with Android 6, and all is working fine. So the PR is working as expected.

Thomas-Vos:
@olme04 so if I understand you correctly, you want me to extract all the calls I added in this PR into one file? That is a lot of extra code. That would mean I need to create extension functions for all of the available socket options, too, instead of just a simple if else condition.

If you want me to do that, any suggestions for naming the functions? `localAddressCompatibility`, `localAddressSafe`, etc

olme04:
BTW, have you checked, that using `nioChannel.socket().shutdownInput()` will not work on both android and unix (there tests for unix in ktor network module)? 
May be we just need to revert those specific changes of skiping `socket()` call there?

Thomas-Vos:
@olme04 `SocketChannel.shutdownInput()` is only available since Android API 24, see: https://developer.android.com/reference/java/nio/channels/SocketChannel#shutdownInput()

That is the reason I added an if condition to call `socket().shutdownInput()` instead, if the newer API is unavailable. That's how it was before #2603.

olme04:
nevermind, looks like it will give almost no sense to extract those checks in one place. 
May be in future we can somehow improve this unix(jdk 16)/android(jdk 6)/common(jdk 8) support.

ptitjes:
@Thomas-Vos Just seen the notification. Sorry I broke support for Android < 24, I did not know it supported Java 6 only.
I don't see why your changes would break things in my use-case, but I'll test it anyway tomorrow evening.

lolilofit:
> @lolilofit, just tested my project which uses uses Ktor UDP sockets and requires options like broadcast.
> 
> I built this PR locally and included it as a dependency. Ran the app on a Samsung Galaxy S5 mini with Android 6, and all is working fine. So the PR is working as expected.

@Thomas-Vos, thanks, that's great

PR #2776, opened 2022-01-17
KTOR-3656 Use secure random for nonce generation

rsinukov:


e5l (changes requested):
Could you check the compilation error for benchmarks?

https://ktor.teamcity.com/buildConfiguration/Ktor_AllocationTests/62594?showRootCauses=false&expandBuildChangesSection=true&expandBuildProblemsSection=true

rsinukov:
Hi @danwallach 
Thanks for your comments. This PR is still in draft and I opened it just for checking that it works on different platform and OS. I'll address your comments before merging.

e5l (approved):
lgtm

ktor-utils/js/src/io/ktor/util/CryptoJs.kt
```diff
@@ -8,7 +8,7 @@ import kotlinx.coroutines.*
 import org.khronos.webgl.*
 import kotlin.js.*
 
-private const val NONCE_SIZE_IN_BYTES = 8
+private const val NONCE_SIZE_IN_BYTES = 16
```
rsinukov, line 11: please extract to common internal constant

danwallach:
Feature request: expose the `secureRandom` method (currently `internal expect fun secureRandom(bytes: ByteArray)`) as a public part of the API. This would make it useful for applications, like mine, that need a portable way to get secure random bytes, for their own cryptographic needs.

PR #2782, opened 2022-01-19
KTOR-3669 Rewrite ContentNegotiation to New Plugins API

e5l:
Fix [KTOR-3669](https://youtrack.jetbrains.com/issue/KTOR-3669)


ktor-server/ktor-server-plugins/ktor-server-content-negotiation/jvm/src/io/ktor/server/plugins/contentnegotiation/ContentNegotiationConfig.kt
```diff
@@ -0,0 +1,62 @@
...
+package io.ktor.server.plugins.contentnegotiation
+
+import io.ktor.http.*
+import io.ktor.serialization.*
+import kotlin.reflect.*
+
+/**
+ * Configuration type for [ContentNegotiation] plugin
+ */
+public class ContentNegotiationConfig : Configuration {
```
olme04, line 14: may be not the right place, but: In all ktor codebase (specifically plugins) there is 2 namings for plugins configuration: *Config and *Configuration. Even in plugins API `createRouteScopedPlugin` has parameter called `createConfiguration` but has type parameter called `PluginConfigT`, what's more, it's create an instance of `RouteScopedPlugin<PluginConfigT>` which has declaration RouteScopedPlugin<TConfiguration>.

Looks like, this should be somehow generalized, so renamed, to always using or `*Config` or `*Configuration` namings.

Other thing: here `ContentNegotiationConfig` extends `Configuration` - for me it's a little confusing.

And the last thing, may be it's better to rename `Configuration` from `ktor-serialization` to something more specific? like `ContentConverterConfiguration` or anything else. Name is too short and widely used.

e5l: Hey @olme04, thank you for the notice! We agree to rename `Configuration` to `Config` everywhere and put prefixes with a clear name.

olme04: ah, several more questions why we are here:)
* what is the reason of naming of shared module as `ktor-serialization` but client/server plugins as  `ktor-server/client-content-negotiation` and the only thing exists inside `ktor-serialization` is `ContentConverter`s related declarations?
* why not to use or `content-negotiation` or `serialization` or other consistent naming everywhere? 
* may be it's better to merge `ktor-serialization` and `ktor-websocket-serialization` modules, as `WebsocketContentConverter` and dependency on `ktor-websockets` is already inside `ktor-serialization`?
* if not, one more proposal for naming alignment would be to use in modules names or `websocket` or `websockets` :)

ktor-server/ktor-server-plugins/ktor-server-content-negotiation/jvmAndNix/src/io/ktor/server/plugins/contentnegotiation/RequestConverter.kt
```diff
@@ -0,0 +1,58 @@
...
+                    )
+                } ?: return@transformBody body
+            } catch (convertException: ContentConvertException) {
+                throw BadRequestException(
+                    convertException.message ?: "Can't convert parameters",
+                    convertException.cause
+                )
+            }
+
+            return@transformBody ApplicationReceiveRequest(
```
rsinukov, line 51: Why do you need to return `ApplicationReceiveRequest` here?

ktor-server/ktor-server-tests/jvmAndNix/test/io/ktor/tests/server/plugins/ContentNegotiationTest.kt
```diff
@@ -578,32 +578,6 @@ class ContentNegotiationTest {
         }
     }
 
-    @Test
```
rsinukov, line 581: why this test is removed?

rsinukov (changes requested):

ktor-server/ktor-server-plugins/ktor-server-content-negotiation/jvmAndNix/src/io/ktor/server/plugins/contentnegotiation/RequestConverter.kt
e5l: to set `reusableValue = true`, it will be dropped with new `DoubleReceive`

ktor-server/ktor-server-plugins/ktor-server-content-negotiation/jvmAndNix/src/io/ktor/server/plugins/contentnegotiation/RequestConverter.kt
```diff
@@ -0,0 +1,58 @@
...
+                typeInfo = requestedType,
+                value = converted,
+                reusableValue = true
```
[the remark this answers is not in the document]
e5l: `DoubleReceive` is not merged, will be dropped after merge

ktor-server/ktor-server-tests/jvmAndNix/test/io/ktor/tests/server/plugins/ContentNegotiationTest.kt
e5l: The test was checking the generic `accept` header verifying, which is not responsibility of `ContentNegotiation` anymore

rsinukov (approved):

ktor-server/ktor-server-plugins/ktor-server-content-negotiation/jvm/src/io/ktor/server/plugins/contentnegotiation/ContentNegotiationConfig.kt
olme04: @e5l ping

PR #2786, opened 2022-01-19
Convert DefaultHeaders plugin to new API

hhariri:
Convert DefaultHeaders plugin to new API

### DIFF
--- a/ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/application/CreatePluginUtils.kt
+++ b/ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/application/CreatePluginUtils.kt
@@ -85,8 +85,13 @@ public fun <PluginConfigT : Any> createRouteScopedPlugin(
         pipeline: ApplicationCallPipeline,
         configure: PluginConfigT.() -> Unit
     ): PluginInstance {
-        check(pipeline is Route)
-        return createPluginInstance(pipeline.application, pipeline, body, createConfiguration, configure)
+        val application = when (pipeline) {
+            is Route -> pipeline.application
+            is Application -> pipeline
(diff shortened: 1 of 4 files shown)


e5l (approved):
lgtm

e5l:
Please rebase on the main branch and run `./gradlew apiDump` with committing changes before merging

ktor-server/ktor-server-plugins/ktor-server-default-headers/jvm/src/io/ktor/server/plugins/DefaultHeaders.kt
```diff
@@ -7,93 +7,72 @@ package io.ktor.server.plugins
...
-        public var clock: () -> Long = { System.currentTimeMillis() }
+/**
+ * Adds the standard `Date` and `Server` HTTP headers and provides the ability
+ * to add additional default headers into each response.
+ */
+public val DefaultHeaders: RouteScopedPlugin<DefaultHeadersConfig, PluginInstance> = createRouteScopedPlugin(
+    "DefaultHeaders",
+    createConfiguration = {
+        DefaultHeadersConfig()
+    }
```
Ololoshechkin, line 44: Nit: I would just simply write `:: DefaultHeadersConfig`

Ololoshechkin (approved):

ktor-server/ktor-server-plugins/ktor-server-default-headers/jvm/src/io/ktor/server/plugins/DefaultHeaders.kt
hhariri: Actually the docs in the issue was using this style...I can change.

PR #2792, opened 2022-01-25
Vbr/metrics ktor 3666

Ololoshechkin:
**Subsystem**

**Motivation**

**Solution**
Describe your solution.


ktor-server/ktor-server-plugins/ktor-server-metrics/jvm/src/io/ktor/server/metrics/dropwizard/DropwizardMetrics.kt
```diff
@@ -14,129 +14,112 @@ import io.ktor.util.pipeline.*
...
-                val routingMetrics = call.attributes.take(routingMetricsKey)
-                val status = call.response.status()?.value ?: 0
-                val statusMeter = plugin.registry.meter(name(plugin.baseName, routingMetrics.name, status.toString()))
-                statusMeter.mark()
-                routingMetrics.context.stop()
-            }
+private object BeforeMonitoring : Hook<(ApplicationCall) -> Unit> {
+    override fun install(application: ApplicationCallPipeline, handler: (ApplicationCall) -> Unit) {
+        val phase = PipelinePhase("MetricsBeforeMonitoring")
+        application.insertPhaseBefore(ApplicationCallPipeline.Monitoring, phase)
```
rsinukov, line 47: why not use `Monitoring`?

ktor-server/ktor-server-plugins/ktor-server-metrics/jvm/src/io/ktor/server/metrics/dropwizard/DropwizardMetrics.kt
```diff
@@ -14,129 +14,112 @@ import io.ktor.util.pipeline.*
...
+        }
+    }
+}
 
-            return plugin
+private object AfterMonitoring : Hook<(ApplicationCall) -> Unit> {
+    override fun install(application: ApplicationCallPipeline, handler: (ApplicationCall) -> Unit) {
+        application.intercept(ApplicationCallPipeline.Monitoring) {
+            try {
+                proceed()
```
rsinukov, line 58: that's not after monitoring, but after whole pipeline is executed

ktor-server/ktor-server-plugins/ktor-server-metrics/jvm/test/io/ktor/server/metrics/dropwizard/DropwizardMetricsTests.kt
```diff
@@ -78,10 +78,12 @@ class DropwizardMetricsTests {
     @Test
     fun `should prefix all metrics with baseName`(): Unit = withTestApplication {
         val prefix = "foo.bar"
-        val registry = application.install(DropwizardMetrics) {
+        var registry: MetricRegistry? = null
+        application.install(DropwizardMetrics) {
             baseName = prefix
             registerJvmMetricSets = false
-        }.registry
+            registry = this.registry
```
rsinukov, line 85: maybe reverse this statement? initialize the registry outside and set it in config block

ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/routing/RoutingHooks.kt
```diff
@@ -0,0 +1,22 @@
...
+ */
+public object CallRouted : Hook<(RoutingApplicationCall) -> Unit> {
+    override fun install(application: ApplicationCallPipeline, handler: (RoutingApplicationCall) -> Unit) {
+        application.environment?.monitor?.subscribe(Routing.RoutingCallStarted) { call -> handler(call) }
+    }
+}
+
+/**
+ * Hook that will be triggered after a call processing was completely finished (i.e. all handlers were
+ * executed and response was sent)
```
rsinukov, line 16: It's not guaranteed the response was sent here. It's possible that the call was not handled by routing and the engine will reply with 404 later in the pipeline

rsinukov (changes requested):

ktor-server/ktor-server-plugins/ktor-server-metrics/jvm/src/io/ktor/server/metrics/dropwizard/DropwizardMetrics.kt
Ololoshechkin: What do you mean? Intercept `Monitoring` instead of intercepting before `Monitoring` ?

rsinukov: Yes

ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/routing/RoutingHooks.kt
Ololoshechkin: Nice point, thanks!
Not sure about the "later in the pipeline" thing, though. As from what I can see, `RoutingCallFinished` event will be triggered after the whole pipeline was execcuted, in the `finally` block.

Ololoshechkin: Correct me if I'm wrong

ktor-server/ktor-server-plugins/ktor-server-metrics/jvm/src/io/ktor/server/metrics/dropwizard/DropwizardMetrics.kt
Ololoshechkin: Am I correct that these 2 blocks are equivalent?
```kotlin
  application.intercept(ApplicationCallPipeline.Monitoring) {
      try {
          proceed()
      } finally {
          HANDLE_ACTION_FIRST()
      }
}
```
and
```kotlin
  application.intercept(ApplicationCallPipeline.Plugins) {
      try {
          proceed()
      } finally {
          HANDLE_ACTION_SECOND()
      }
}
```
The only difference is that if it breaks before `Plugins` phase then `HANDLE_ACTION_SECOND` will not be executed, which is not the case with `HANDLE_ACTION_FIRST` although it's a rare case.

Correct?

ktor-server/ktor-server-plugins/ktor-server-metrics/jvm/test/io/ktor/server/metrics/dropwizard/DropwizardMetricsTests.kt
Ololoshechkin: Will do, thanks

ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/routing/RoutingHooks.kt
Ololoshechkin: I would say:
```
Hook that will be triggered after a call processing was completely finished (i.e. all handlers were executed and response was sent or an error happened)
 ```

File: ktor-server/ktor-server-plugins/ktor-server-metrics/jvm/src/io/ktor/server/metrics/dropwizard/DropwizardMetrics.kt (hunk 2 above)
Comment (Ololoshechkin) [reply ktor-server/ktor-server-plugins/ktor-server-metrics/jvm/src/io/ktor/server/metrics/dropwizard/DropwizardMetrics.kt:58]: Btw, not it seems that there is no actual difference between `on(CallFinished)`(which happens after the full pipeline execution if I'm not mistaken) and `on(AfterMonitoring)` (which is not "after monitoring" but after everything)

File: ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/routing/RoutingHooks.kt (hunk 4 above)
Comment (rsinukov) [reply ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/routing/RoutingHooks.kt:16]: It can be confusing, but we have 2 pipelines here: application and routing. Routing pipeline is executed inside `Call` phase of the application pipeline. So, this `finally` block is only at the end of the routing pipeline, but is still inside `Call` phase of the application pipeline. 
So, the correct doc will say that this hook is executed after routing was processed or error in handler happened.

File: ktor-server/ktor-server-plugins/ktor-server-metrics/jvm/src/io/ktor/server/metrics/dropwizard/DropwizardMetrics.kt (hunk 2 above)
Comment (rsinukov) [reply ktor-server/ktor-server-plugins/ktor-server-metrics/jvm/src/io/ktor/server/metrics/dropwizard/DropwizardMetrics.kt:58]: Yes, it's correct

**Review by rsinukov — APPROVED** [review]
LGTM, but please rebase

File: ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/routing/RoutingHooks.kt [hunk 5]
```diff
@@ -0,0 +1,21 @@
+package io.ktor.server.routing
+
+import io.ktor.server.application.*
+
+/**
+ * Hook that will be triggered when a call was routed and before any other handlers.
+ */
+public object CallRouted : Hook<(RoutingApplicationCall) -> Unit> {
```
e5l, line 8: consider making all hooks internal and move to the plugin scope for a while. Let's have separate discussion about hooks and naming

ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/routing/RoutingHooks.kt
```diff
@@ -0,0 +1,21 @@
...
+    override fun install(application: ApplicationCallPipeline, handler: (RoutingApplicationCall) -> Unit) {
+        application.environment?.monitor?.subscribe(Routing.RoutingCallStarted) { call -> handler(call) }
```
e5l, line 10: I'm not sure that monitoring is a proper way of doing that, cause it would be really handy to throw an exception here in some cases

ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/routing/RoutingHooks.kt
```diff
@@ -0,0 +1,21 @@
...
+    }
+}
+
+/**
+ * Hook that will be triggered after a call was successfully routed and it's processing has completely finished.
+ */
+public object RoutedCallProcessed : Hook<(ApplicationCall) -> Unit> {
+    override fun install(application: ApplicationCallPipeline, handler: (ApplicationCall) -> Unit) {
+        application.environment?.monitor?.subscribe(Routing.RoutingCallFinished) { call -> handler(call) }
```
e5l, line 19: same note here

ktor-server/ktor-server-plugins/ktor-server-metrics/jvm/src/io/ktor/server/metrics/dropwizard/DropwizardMetrics.kt
```diff
@@ -10,133 +10,113 @@ import com.codahale.metrics.jvm.*
...
+            } finally {
+                handler(call)
+            }
         }
     }
+}
 
-    private data class CallMeasure(val timer: Timer.Context)
+public val DropwizardMetrics: ApplicationPlugin<Application, DropwizardMetricsConfig, PluginInstance> =
```
e5l, line 63: KDoc is missing

ktor-server/ktor-server-plugins/ktor-server-metrics/jvm/src/io/ktor/server/metrics/dropwizard/DropwizardMetrics.kt
```diff
@@ -10,133 +10,113 @@ import com.codahale.metrics.jvm.*
...
-    private val measureKey = AttributeKey<CallMeasure>("metrics")
+        on(CallFailed) { _, _ ->
+            exceptions.mark()
+        }
 
-    private fun before(call: ApplicationCall) {
-        active.inc()
-        call.attributes.put(measureKey, CallMeasure(duration.time()))
-    }
+        on(CallRouted) { call ->
```
e5l, line 86: I'm not sure why do we need these hooks when can use monitoring directly

ktor-server/ktor-server-plugins/ktor-server-metrics/jvm/src/io/ktor/server/metrics/dropwizard/DropwizardMetrics.kt
```diff
@@ -10,133 +10,113 @@ import com.codahale.metrics.jvm.*
...
+            val statusMeter =
+                pluginConfig.registry.meter(name(pluginConfig.baseName, routingMetrics.name, status.toString()))
+            statusMeter.mark()
+            routingMetrics.context.stop()
         }
-        meter.mark()
-        call.attributes.getOrNull(measureKey)?.apply {
-            timer.stop()
+
+        on(Monitoring) { call ->
```
e5l, line 107: It's not clear when monitoring is happening, consider to rename it

e5l (changes requested):

ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/routing/RoutingHooks.kt
Ololoshechkin: We have `CallFailed` for handling exceptions while call routed is just an event that happens at the same time as `RoutingCallStarted` (basically it is just a wrapper over the later that is new-api-friendly)

Ololoshechkin: I'll make it internal anyway, so that we can describe later

ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/routing/RoutingHooks.kt
Ololoshechkin: ditto

ktor-server/ktor-server-plugins/ktor-server-metrics/jvm/src/io/ktor/server/metrics/dropwizard/DropwizardMetrics.kt
Ololoshechkin: Just because these hoks encapsulate the boilerplate code and are much handier and consistend with new API

ktor-server/ktor-server-plugins/ktor-server-metrics/jvm/src/io/ktor/server/metrics/dropwizard/DropwizardMetrics.kt
Ololoshechkin: Would `BeforeCall` work here? 
Or maybe `TracingCall`

e5l (approved):
LGTM

v1ctor:
Hey all, we recently found an issue when DropwizardMetricsPlugin doesn't log a correct http status when status is set by the StatusPages plugin (https://youtrack.jetbrains.com/issue/KTOR-5420/DropwizardMetricsPlugin-logs-status-code-incorrectly-when-is-used-together-with-StatusPages-plugin.)

I believe that this PR fixes this issue, as it changes what hooks DropwizardMetrics plugin is using, however I can see that branch `new-plugins-eap` was never merged to the main. What was the reason for that?

PR #2793, opened 2022-01-25
KTOR-3740 Support routing scoped plugins in hooks

Ololoshechkin:
https://youtrack.jetbrains.com/issue/KTOR-3740


ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/application/CreatePluginUtils.kt
```diff
@@ -198,5 +198,6 @@ private fun <Configuration : Any, Plugin : ApplicationPluginBuilder<Configuratio
         it.action(pipeline.sendPipeline)
     }
 
-    hooks.forEach { it.install(application) }
+    println("   installing hooks: ${hooks.size}")
```
e5l, line 201: please drop println

e5l (approved):
please update API dump as well

PR #2797, opened 2022-01-25
KTOR-3730 Flush unread bytes after readAvailable

jhead:
**Subsystem**

- ktor-io (jvm)
- loosely related to Apache client engine

**Motivation**

https://youtrack.jetbrains.com/issue/KTOR-3730

**Solution**

This `readAvailable` variant provides `block` with a buffer and tracks how many bytes were read. After executing `block`, any unread bytes are moved back to `pendingToFlush`. This change flushes those _unread_ bytes back to `availableForRead` so that subsequent `readAvailable` calls are able to read them.

Also adds tests, including one that would have caught this issue.


ktor-io/jvm/test/io/ktor/utils/io/ByteBufferChannelTest.kt
```diff
@@ -97,6 +97,72 @@ class ByteBufferChannelTest {
...
+            }
+            assertEquals(dataSize, secondRead)
+        }
+
+        try {
+            withTimeout(2500) { job.join() }
+        } catch (e: TimeoutCancellationException) {
+            fail("All bytes should be written to and read from the channel")
+        }
+    }
```
jhead, line 130: this new test fails without the fix

e5l (approved):

e5l:
Hey @jhead, thanks for the PR. LGTM, merged!

PR #2812, opened 2022-01-27
KTOR-3661 Refactor Compression to new API

rsinukov:


ktor-server/ktor-server-plugins/ktor-server-compression/jvm/src/io/ktor/server/plugins/compression/Compression.kt
```diff
@@ -15,65 +15,53 @@ import io.ktor.util.cio.*
...
-     * Conditions for the encoder
-     */
-    val conditions: List<ApplicationCall.(OutgoingContent) -> Boolean>,
-    /**
-     * Priority of the encoder
-     */
-    val priority: Double
-)
+    class Context(private val pipelineContext: PipelineContext<Any, ApplicationCall>) {
+        fun transformBody(block: (OutgoingContent) -> OutgoingContent?) {
```
Ololoshechkin, line 27: What happens if ` OutgoingContent?` is `null`?

Ololoshechkin (approved):

ktor-server/ktor-server-plugins/ktor-server-compression/jvm/src/io/ktor/server/plugins/compression/Compression.kt
rsinukov: It means that the body could not be transformed and the plugin will keep the old body

PR #2818, opened 2022-01-28
KTOR-789 Use UTF-8 as default charset for text and forms

rsinukov:
According to  https://www.rfc-editor.org/rfc/rfc2046.html#section-4.1.2 the default charset to `Text/*` is ASCII, so using `Charsets.ISO_8859_1` doesn't make sense here. `Application/json` and other popular types have `UTF-8` as default charset.


e5l (approved):
lgtm

PR #2819, opened 2022-01-31
KTOR-574 Add support for different application profiles

Moluram:
**Subsystem**
Server-Host-Common

**Motivation**
It is useful to have the ability to change profiles via env parameters.

Solves [#KTOR-574](https://youtrack.jetbrains.com/issue/KTOR-574)

**Solution**
Added additional "-profile" env parameter which if present adds a postfix to the "application.conf" file.

### DIFF
--- a/ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/EnvironmentUtilsJvm.kt
+++ b/ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/EnvironmentUtilsJvm.kt
@@ -19,7 +19,8 @@ internal actual fun buildApplicationConfig(args: Array<String>): ApplicationConf
         .filterKeys { it.startsWith("-P:") }
         .mapKeys { it.key.removePrefix("-P:") }
 
-    val configFile = argumentsPairs["-config"]?.let { File(it) }
+    val profile = argumentsPairs["-profile"]
+    val configFile = argumentsPairs["-config"]?.let { File(it) } ?: profile?.let { File("application-${profile}.conf") }
 
     val argConfig = ConfigFactory.parseMap(commandLineProperties, "Command-line options")
 
--- a/ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/CommandLineJvm.kt
+++ b/ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/CommandLineJvm.kt
@@ -29,7 +29,10 @@ public fun commandLineEnvironment(
             else -> File(it).toURI().toURL()
         }
     }
-    val configFile = argsMap["-config"]?.let { File(it) }
+
+    val profile = argsMap["-profile"]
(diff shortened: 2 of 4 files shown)


ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/EnvironmentUtilsJvm.kt
```diff
@@ -19,7 +19,8 @@ internal actual fun buildApplicationConfig(args: Array<String>): ApplicationConf
         .filterKeys { it.startsWith("-P:") }
         .mapKeys { it.key.removePrefix("-P:") }
 
-    val configFile = argumentsPairs["-config"]?.let { File(it) }
+    val profile = argumentsPairs["-profile"]
+    val configFile = argumentsPairs["-config"]?.let { File(it) } ?: profile?.let { File("application-${profile}.conf") }
```
Stexxe, line 23: How to read profile configs from the local file system if the are stored not in the current working directory?

Moluram: Hello @Stexxe, what do you think about such a solution to a stated problem?
```
    val profile = argumentsPairs["-profile"]
    val configFilePath = argumentsPairs["-config"]?.let {
        profile?.let { profileName ->
            if (it.endsWith(".conf")) {
                return@let "${it.removeSuffix(".conf")}-${profileName}.conf";
            }
            it
        } ?: it
    }

    val configFile = configFilePath?.let { File(it) } ?: profile?.let { File("application-${it}.conf") }
```

rsinukov: @Stexxe but isn't this case will be executed when user hasn't specified a custom path to config file, which means Ktor should load one from `resources` folder?

Stexxe: @rsinukov according to the new logic it reads a config with a path provided by `-config` option or a profile config in the current working directory.

Stexxe: @Moluram I don't like it because it relies on a config extension.

rsinukov: @Stexxe exactly. So user either specifies full path, or profile. In case of profile, we should only read from the current directory. Does it make your original comment irrelevant or am I missing something?

mfarkan:
Is there any update for this feature ?

rsinukov:
Closing this PR because the original ticket was declined. Please see [this comment](https://youtrack.jetbrains.com/issue/KTOR-574/support-application.conf-with-env-type#focus=Comments-27-6392578.0-0) for details

PR #2820, opened 2022-02-01
KTOR-3351 Rewrite CallLogging

e5l:
Fix [KTOR-3351](https://youtrack.jetbrains.com/issue/KTOR-3351)

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -29,7 +29,7 @@ kotlin.mpp.enableCompatibilityMetadataVariant=true
 serialization_version=1.3.0
 coroutines_version=1.6.0
 atomicfu_version=0.17.0
-validator_version=0.3.0
+validator_version=0.8.0
 kotlinx_html_version=0.7.3
 
 # server
--- a/ktor-server/ktor-server-plugins/ktor-server-call-logging/build.gradle.kts
+++ b/ktor-server/ktor-server-plugins/ktor-server-call-logging/build.gradle.kts
@@ -12,5 +12,11 @@ kotlin {
                 implementation("org.fusesource.jansi:jansi:$jansi_version")
             }
         }
+
+        val jvmTest by getting {
+            dependencies {
+                implementation(project(":ktor-server:ktor-server-plugins:ktor-server-call-id"))
+            }
+        }
     }
 }
--- a/ktor-server/ktor-server-plugins/ktor-server-call-logging/jvm/src/io/ktor/server/plugins/callloging/CallLogging.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-call-logging/jvm/src/io/ktor/server/plugins/callloging/CallLogging.kt
@@ -5,228 +5,24 @@
 package io.ktor.server.plugins.callloging
 
 import io.ktor.events.*
-import io.ktor.http.*
 import io.ktor.server.application.*
 import io.ktor.server.logging.*
-import io.ktor.server.request.*
 import io.ktor.util.*
 import io.ktor.util.pipeline.*
-import kotlinx.coroutines.*
-import org.fusesource.jansi.*
-import org.slf4j.*
 import org.slf4j.event.*
-import kotlin.coroutines.*
 
 /**
  * Logs application lifecycle and call events.
  */
-public class CallLogging private constructor(
-    private val log: Logger,
-    private val monitor: Events,
-    private val level: Level,
-    private val filters: List<(ApplicationCall) -> Boolean>,
-    private val mdcEntries: List<MDCEntry>,
-    private val formatCall: (ApplicationCall) -> String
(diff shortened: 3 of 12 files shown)


ktor-server/ktor-server-plugins/ktor-server-call-logging/jvm/src/io/ktor/server/plugins/callloging/CallLoggingConfig.kt
```diff
@@ -0,0 +1,101 @@
...
+
+    internal fun ApplicationRequest.toLogStringWithColors(): String =
+        "${colored(httpMethod.value, Ansi.Color.CYAN)} - ${path()}"
+
+    private fun colored(status: HttpStatusCode): String {
+        try {
+            if (!AnsiConsole.isInstalled()) {
+                AnsiConsole.systemInstall()
+            }
+        } catch (cause: Throwable) {
```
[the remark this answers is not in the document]
e5l: @sonatype-lift ignore

gradle.properties
```diff
@@ -29,7 +29,7 @@ kotlin.mpp.enableCompatibilityMetadataVariant=true
 serialization_version=1.3.0
 coroutines_version=1.6.0
 atomicfu_version=0.17.0
-validator_version=0.3.0
+validator_version=0.8.0
```
Ololoshechkin, line 32: Is it connected to this PR about call logging?

ktor-server/ktor-server-plugins/ktor-server-call-logging/jvm/src/io/ktor/server/plugins/callloging/CallLogging.kt
```diff
@@ -5,265 +5,92 @@
...
-            else value.toString() // ignore color
+public val CallLogging: ApplicationPlugin<Application, CallLoggingConfig, PluginInstance> = createApplicationPlugin(
+    "CallLogging",
+    ::CallLoggingConfig
+) {
+    val log = pluginConfig.logger ?: application.log
+    val filters = pluginConfig.filters
+    val formatCall = pluginConfig.formatCall
+
+    fun log(message: String) = when (pluginConfig.level) {
```
Ololoshechkin, line 25: Nit: maybe just private methods would be better here?

e5l: they are inside lambda and internal to the plugin. Can't make outside because of capturing

ktor-server/ktor-server-plugins/ktor-server-call-logging/jvm/src/io/ktor/server/plugins/callloging/CallLogging.kt
```diff
@@ -5,265 +5,92 @@
...
-     */
-    private suspend inline fun withMDC(call: ApplicationCall, crossinline block: suspend () -> Unit) {
-        withContext(MDCSurvivalElement(setupMdc(call))) {
-            try {
-                block()
-            } finally {
-                cleanupMdc()
-            }
-        }
+    on(MDCHook(ApplicationCallPipeline.Monitoring)) { call, block ->
```
Ololoshechkin, line 62: Nit: here and below --  this code will be executed each time one installs your plugin.
Hence, new objects will be created. Maybe creating static objects inside `MDCHook` will be just a bit more efficient :)

ktor-server/ktor-server-plugins/ktor-server-call-logging/build.gradle.kts
```diff
@@ -12,5 +12,11 @@ kotlin {
                 implementation("org.fusesource.jansi:jansi:$jansi_version")
             }
         }
+
+        val jvmTest by getting {
+            dependencies {
+                implementation(project(":ktor-server:ktor-server-plugins:ktor-server-call-id"))
+            }
+        }
```
Ololoshechkin, line 20: Is it used anywhere? Did you add any tests?

ktor-server/ktor-server-plugins/ktor-server-call-logging/jvm/src/io/ktor/server/plugins/callloging/CallLogging.kt
Ololoshechkin: You can pass parameters cuz there's only a few of then required

ktor-server/ktor-server-plugins/ktor-server-call-logging/build.gradle.kts
e5l: Yep, it is used in moved tests

PR #2832, opened 2022-02-03
KTOR-3783 Use global DslMarker

rsinukov:
Fixes https://youtrack.jetbrains.com/issue/KTOR-3333


e5l (approved):
Good job, lgtm

hfhbd:
This only works, iff a 3rd party plugin also uses `KtorDSL`, doesn't it?

e5l:
@hfhbd all receivers should be marked

hfhbd:
Yeah sure. I like the solution, but only want to notice a 3rd party plugin author has to add the annotation. So this should be written in the docs.

rsinukov:
@hfhbd Yes, it should be. 
@andreyaksenov, When you will work on docs for custom plugins, please consider adding that config classes should be annotated with `@KtorDsl`

PR #2833, opened 2022-02-04
KTOR-3516 Add auth from single provider if no WWW-Auth header is present

rsinukov:


ktor-client/ktor-client-plugins/ktor-client-auth/common/src/io/ktor/client/plugins/auth/AuthProvider.kt
```diff
@@ -24,14 +24,16 @@ public interface AuthProvider {
...
     public fun isApplicable(auth: HttpAuthHeader): Boolean
 
     /**
-     * Add authentication method headers and creds.
+     * Adds authentication method headers and credentials.
+     * @param authHeader value of `WWW-Authenticate` header from failed response, if exists
+     * @param request builder for an authenticated request
      */
-    public suspend fun addRequestHeaders(request: HttpRequestBuilder)
+    public suspend fun addRequestHeaders(authHeader: HttpAuthHeader?, request: HttpRequestBuilder)
```
e5l, line 36: I would add the header as the last parameter with a default `null` value

e5l (approved):

PatrykKleczkowski:
Hello @e5l 

Do you have any idea when it will be released?

Best regards!

e5l:
Soon after kotlin 1.6.20!

PatrykKleczkowski:
Thank You @e5l !
Is it related with this issue?
https://youtrack.jetbrains.com/issue/KTOR-2806

e5l:
@rsinukov, could you take a look?

rsinukov:
@PatrykKleczkowski Yes, it's the same issue

PR #2835, opened 2022-02-04
Add optionalRoute

goncalossilva:
**Subsystem**
Server.

**Motivation**
Optional route allows a path segment to optionally exist.

One example where this is useful is in API versioning and rerouting “missing version” to the latest version without a redirect:

```kotlin
fun Application.installRoutes() = routing {
    optionalRoute("v1") {
        // ...
    }
}
```

In this case, both `api.domain.com/v1/some-route` and `api.domain.com/some-route` would point to the same place.

**Solution**
This is a draft PR, lifted from [Doist/ffs](https://github.com/Doist/ffs/blob/cfd7c757f2d8205ac485551cbc5b507ca7ae7d8a/ffs-server/src/main/kotlin/ext/RouteExt.kt). I'm happy to add a ticket in YT, review the structure for a tighter integration, add tests, and so forth, but would prefer determining if there's interest in this first.


rsinukov:
Hi @goncalossilva 
We already have functionaluty for optional path parameters. But as I understand, you need optional constant path segments, right? 
For path parameters you can use this code 
```kotlin
    @Test
    fun testOptionalPathParameter() = testApplication {
        routing {
            route("{test?}") {
                handle { call.respond(call.parameters["test"] ?: "missing") }
            }
        }

        assertEquals("asd", client.get("/asd").bodyAsText())
        assertEquals("missing", client.get("/").bodyAsText())
    }
```

goncalossilva:
Exactly. The issue with optional path parameters is that they can only be used at the end of a path, while an optional route can be set anywhere in the path.

Thinking back to the API example I alluded to in the original post, both of these could work and point to the same route:

```
api.domain.com/v1/organizations/1/edit
api.domain.com/organizations/1/edit
```

A test could be:

```kotlin
@Test
    fun testOptionalRouteParameter() = testApplication {
        routing {
            optionalRoute("optional") {
				route("test") {
                	handle { call.respond("here") }
				}
            }
        }

        assertEquals("here", client.get("/optional/test").bodyAsText())
        assertEquals("here", client.get("/test").bodyAsText())
    }
```

e5l (approved):

e5l:
Hey @goncalossilva, thanks for the PR. Could you add some tests and run `./gradlew apiDump` and commit changes?

goncalossilva:
Hey @e5l! I've been meaning to make time for this but I'm unfortunately not actively using ktor, so it's difficult to prioritize. I'll close the ticket to reduce entropy for you guys. I'll pick it up if I return to using ktor and need this functionality again.

PR #2838, opened 2022-02-08
KTOR-3336 Fixes invalidating cookies cache on every request

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-3336

Simplify cookies' cache to store strings and compare value before invalidating cache.

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-sessions/jvm/src/io/ktor/server/sessions/SessionStorage.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-sessions/jvm/src/io/ktor/server/sessions/SessionStorage.kt
@@ -4,19 +4,14 @@
 
 package io.ktor.server.sessions
 
-import io.ktor.utils.io.*
-
 /**
  * Represents a way to [write], [read] and [invalidate] session bits.
  */
 public interface SessionStorage {
     /**
-     * Writes a session [id] using a specific [provider].
-     *
-     * This method calls the [provider] with a [ByteWriteChannel] and it is in charge of the channel's lifecycle.
-     * [provider] is in charge of writing session bits to the specified [ByteWriteChannel].
+     * Writes a session [value] for [id].
      */
-    public suspend fun write(id: String, provider: suspend (ByteWriteChannel) -> Unit)
+    public suspend fun write(id: String, value: String)
 
     /**
      * Invalidates session [id].
@@ -28,14 +23,9 @@ public interface SessionStorage {
     public suspend fun invalidate(id: String)
 
     /**
-     * Reads session [id] using a [consumer] as [R]
+     * Reads session by [id]
      *
-     * This method calls the [consumer] with a [ByteReadChannel] and it is in charge of the channel's lifecycle,
-     * and returns the object [R] produced by the [consumer].
-     * [consumer] should read the content of the specified [ByteReadChannel] and return an object of type [R].
-     *
-     * @return instance of [R] representing the session object returned from the [consumer].
      * @throws NoSuchElementException when session [id] is not found.
      */
-    public suspend fun <R> read(id: String, consumer: suspend (ByteReadChannel) -> R): R
+    public suspend fun read(id: String): String
 }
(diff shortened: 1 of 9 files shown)


e5l (approved):
LGTM.

I'm still thinking if it's a good idea to keep the same coroutine context if the `SessionStorage` is implemented on top of an async database.

ktor-server/ktor-server-plugins/ktor-server-sessions/jvm/src/io/ktor/server/sessions/SessionStorage.kt
```diff
@@ -4,19 +4,14 @@
...
 public interface SessionStorage {
     /**
-     * Writes a session [id] using a specific [provider].
-     *
-     * This method calls the [provider] with a [ByteWriteChannel] and it is in charge of the channel's lifecycle.
-     * [provider] is in charge of writing session bits to the specified [ByteWriteChannel].
+     * Writes a session [value] for [id].
      */
-    public suspend fun write(id: String, provider: suspend (ByteWriteChannel) -> Unit)
+    public suspend fun write(id: String, value: String)
```
artem-zinnatullin, line 14: I like this change because it reads better, but it's going to be a breaking change for users...

ktor-server/ktor-server-plugins/ktor-server-sessions/jvm/src/io/ktor/server/sessions/SessionStorage.kt
```diff
@@ -28,14 +23,9 @@ public interface SessionStorage {
...
-     * This method calls the [consumer] with a [ByteReadChannel] and it is in charge of the channel's lifecycle,
-     * and returns the object [R] produced by the [consumer].
-     * [consumer] should read the content of the specified [ByteReadChannel] and return an object of type [R].
+     * Reads session by [id]
      *
-     * @return instance of [R] representing the session object returned from the [consumer].
      * @throws NoSuchElementException when session [id] is not found.
      */
-    public suspend fun <R> read(id: String, consumer: suspend (ByteReadChannel) -> R): R
+    public suspend fun read(id: String): String
```
artem-zinnatullin, line 30: Same, this will break existing user implementations :(

ktor-server/ktor-server-plugins/ktor-server-sessions/jvm/src/io/ktor/server/sessions/SessionStorage.kt
artem-zinnatullin: tbh current interface and reference implementations in Ktor sources and docs https://ktor.io/docs/storages.html#custom_storage (SimplifiedSessionStorage) are overcomplicated for such a simple concept, I found myself just completely lost multiple times while reading sources and trying to implement mine

ktor-server/ktor-server-plugins/ktor-server-sessions/jvm/src/io/ktor/server/sessions/SessionStorage.kt
e5l: that's ok, we're going to make it in `2.0.0` with other breaking changes.

@rsinukov please make sure we have migrations updated

ktor-server/ktor-server-plugins/ktor-server-sessions/jvm/src/io/ktor/server/sessions/SessionStorage.kt
ikorennoy: It might be better to replace String with ByteArray for value

ktor-server/ktor-server-plugins/ktor-server-sessions/jvm/src/io/ktor/server/sessions/SessionStorage.kt
ikorennoy: Same here, ByteArray in API looks better

ikorennoy (approved):

ktor-server/ktor-server-plugins/ktor-server-sessions/jvm/src/io/ktor/server/sessions/SessionStorage.kt
rsinukov: @artem-zinnatullin I agree that it's overcomplicated and that's why we are takeing opportunity of major release to simplify it.

@ikorennoy Why do you prefer a byte array? Cookie is a string, all users of this API read and write strings. Are there cases when we need to use byte array?

ikorennoy: @rsinukov, I got this idea because I was thinking about a use case where the session is a complex object that is serialized and stored in a database, using a string probably reduces the number of serialization algorithms that can be used.

Also, this suggestion is based on this example [Custom storage](https://ktor.io/docs/storages.html#custom_storage)

rsinukov: Currently, our API for serializer supports only serialization to and from a string. We may change it in the future if there would be requests for this.

PR #2839, opened 2022-02-08
KTOR-3803 Drop before and after from new plugins API

e5l:
Fix [KTOR-3803](https://youtrack.jetbrains.com/issue/KTOR-3803)


lolilofit (approved):
LGTM

PR #2844, opened 2022-02-10
KTOR-3825 multipart/form-data requests: No way of streaming data asynchronously

Stexxe:
https://youtrack.jetbrains.com/issue/KTOR-3825


e5l (changes requested):

e5l (approved):
lgtm

PR #2847, opened 2022-02-11
KTOR-3831 Fix InvalidPathException in ApplicationEngineEnvironmentRel…

spand:
…oading.kt

**Subsystem**
Server

**Motivation**
https://youtrack.jetbrains.com/issue/KTOR-3831

**Solution**
Keeping consistent with the "keep on trucking" approach for the rest of the code I have added more error handling. The patch applied to the 1.6.7 version fixed the exceptions I encountered as described in the issue.


e5l (approved):
LGTM

PR #2855, opened 2022-02-15
KTOR-3873 Introduce BaseApplicationPlugin

e5l:
Fix [KTOR-3873](https://youtrack.jetbrains.com/issue/KTOR-3873)

### DIFF
--- a/ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/application/ApplicationPlugin.kt
+++ b/ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/application/ApplicationPlugin.kt
@@ -39,7 +39,6 @@ public interface Plugin<
  * @param TConfiguration is the configuration object type for this Plugin
  * @param TPlugin is the instance type of the Plugin object
  */
-@Suppress("AddVarianceModifier")
 public interface BaseApplicationPlugin<
     in TPipeline : Pipeline<*, ApplicationCall>,
     out TConfiguration : Any,
@@ -47,11 +46,8 @@ public interface BaseApplicationPlugin<
 
 /**
  * Defines a Plugin that is installed into Application
- * @param TPipeline is the type of the pipeline this plugin is compatible with
  * @param TConfiguration is the configuration object type for this Plugin
- * @param TPlugin is the instance type of the Plugin object
  */
-@Suppress("AddVarianceModifier")
 public interface ApplicationPlugin<out TConfiguration : Any> :
     BaseApplicationPlugin<Application, TConfiguration, PluginInstance>
 
@@ -91,7 +87,7 @@ public fun <P : Pipeline<*, ApplicationCall>, B : Any, F : Any> P.install(
     plugin: Plugin<P, B, F>,
     configure: B.() -> Unit = {}
 ): F {
-    if (this is Route && plugin is RouteScopedPlugin) {
+    if (this is Route && plugin is BaseRouteScopedPlugin) {
         return installIntoRoute(plugin, configure)
     }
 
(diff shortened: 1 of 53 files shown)


ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/application/ApplicationPlugin.kt
```diff
@@ -34,17 +34,27 @@ public interface Plugin<
...
 @Suppress("AddVarianceModifier")
-public interface ApplicationPlugin<
+public interface BaseApplicationPlugin<
     in TPipeline : Pipeline<*, ApplicationCall>,
     out TConfiguration : Any,
     TPlugin : Any> : Plugin<TPipeline, TConfiguration, TPlugin>
 
+/**
+ * Defines a Plugin that is installed into Application
+ * @param TPipeline is the type of the pipeline this plugin is compatible with
```
rsinukov, line 50: please update kdoc

ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/application/ApplicationPlugin.kt
```diff
@@ -34,17 +34,27 @@ public interface Plugin<
...
+ * @param TConfiguration is the configuration object type for this Plugin
+ * @param TPlugin is the instance type of the Plugin object
+ */
+@Suppress("AddVarianceModifier")
+public interface ApplicationPlugin<out TConfiguration : Any> :
```
rsinukov, line 55: should it be type alias or do we need a real type?

ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/application/ApplicationPlugin.kt
```diff
@@ -34,17 +34,27 @@ public interface Plugin<
...
 /**
- * Defines a Plugin that is installed into Application
+ * Defines a Plugin that is installed into Application.
  * @param TPipeline is the type of the pipeline this plugin is compatible with
  * @param TConfiguration is the configuration object type for this Plugin
  * @param TPlugin is the instance type of the Plugin object
  */
 @Suppress("AddVarianceModifier")
-public interface ApplicationPlugin<
+public interface BaseApplicationPlugin<
```
rsinukov, line 43: what about route scoped plugin?

rsinukov (changes requested):

ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/application/ApplicationPlugin.kt
e5l: I will check if the type alias appears in the IDE actions and if so we can use it

ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/application/ApplicationPlugin.kt
e5l: Sure, thanks

ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/application/ApplicationPlugin.kt
e5l: Sure, let me fix this as well

ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/application/ApplicationPlugin.kt
e5l: Checked. Typealias are not used when you specify an explicit type. So interface should be used

rsinukov (approved):
LGTM, but api check is failed/

PR #2857, opened 2022-02-15
KTOR-3879 Update node_fetch_version to 2.6.7, remove kotlin-js-store/yarn.lock from .gitignore

SebastianAigner:
This PR updates node-fetch to 2.6.7 (ref. https://youtrack.jetbrains.com/issue/KTOR-3879)

It also removes kotlin-js-store/yarn.lock from .gitignore to allow for proper version pinning and for Dependabot to monitor the repository for potential future vulnerabilities, and adds a current yarn.lock to be used.

**Subsystem**
JS Client

**Motivation**
ref. https://youtrack.jetbrains.com/issue/KTOR-3879

**Solution**
Incremented version number, built locally to generate yarn.lock


e5l (changes requested):
Could you tell me why `kotlin-js-store` is dropped from `.gitignore`?

e5l (approved):

PR #2858, opened 2022-02-16
KTOR-3832 Introduce file cache; Prevent blocking when receiving conte…

e5l:
Fix [KTOR-3832](https://youtrack.jetbrains.com/issue/KTOR-3832)

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-double-receive/jvm/src/io/ktor/server/plugins/doublereceive/FileCacheJvm.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-double-receive/jvm/src/io/ktor/server/plugins/doublereceive/FileCacheJvm.kt
@@ -1,0 +1,59 @@
+/*
+ * Copyright 2014-2022 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.server.plugins.doublereceive
+
+import io.ktor.util.cio.*
+import io.ktor.utils.io.*
+import kotlinx.coroutines.*
+import java.io.*
+import java.nio.*
+import kotlin.coroutines.*
+
+internal actual class FileCache actual constructor(
+    private val body: ByteReadChannel,
+    bufferSize: Int,
+    context: CoroutineContext
+) : DoubleReceiveCache {
+    private val file = File.createTempFile("ktor-double-receive-cache", ".tmp")
+
+    @OptIn(DelicateCoroutinesApi::class)
+    private val saveJob = GlobalScope.launch(context + Dispatchers.IO) {
+        val buffer = ByteBuffer.allocate(bufferSize)
+
+        @Suppress("BlockingMethodInNonBlockingContext")
+        FileOutputStream(file).use { stream ->
+            stream.channel.use { out ->
+                out.truncate(0L)
+                buffer.position(buffer.limit())
+
+                while (true) {
+                    while (buffer.hasRemaining()) {
+                        out.write(buffer)
+                    }
+                    buffer.clear()
+
+                    if (body.readAvailable(buffer) == -1) break
+                    buffer.flip()
+                }
+            }
+        }
+    }
+
+    actual override fun read(): ByteReadChannel = file.readChannel()
+
+    actual override fun dispose() {
+        runCatching {
+            saveJob.cancel()
+        }
+        runCatching {
+            file.delete()
+        }
+        if (!body.isClosedForRead) {
+            runCatching {
+                body.cancel()
+            }
+        }
+    }
+}
--- a/ktor-server/ktor-server-plugins/ktor-server-double-receive/jvm/src/io/ktor/server/plugins/doublereceive/FileCacheJvm.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-double-receive/jvm/src/io/ktor/server/plugins/doublereceive/FileCacheJvm.kt
@@ -1,3 +1,4 @@
+// ktlint-disable filename
 /*
  * Copyright 2014-2022 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
  */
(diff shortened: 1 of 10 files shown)


ktor-server/ktor-server-plugins/ktor-server-double-receive/jvm/src/io/ktor/server/plugins/doublereceive/FileCacheJvm.kt
```diff
@@ -0,0 +1,59 @@
...
+import java.io.*
+import java.nio.*
+import kotlin.coroutines.*
+
+internal actual class FileCache actual constructor(
+    private val body: ByteReadChannel,
+    bufferSize: Int,
+    context: CoroutineContext
+) : DoubleReceiveCache {
+    private val file = File.createTempFile("ktor-double-receive-cache", ".tmp")
```
InsanusMokrassar, line 19: Add `deleteOnExit()`?

e5l: I would say it should be in `call.afterFinish`.

ktor-server/ktor-server-plugins/ktor-server-double-receive/jvmAndNix/src/io/ktor/server/plugins/doublereceive/DoubleReceiveConfig.kt
```diff
@@ -24,6 +24,13 @@ public class DoubleReceiveConfig {
      */
     public var cacheRawRequest: Boolean = true
 
+    /**
+     * Indicates if a file should be used to cache raw requests. Otherwise, requests will be stored in memory.
+     *
+     * If [cacheRawRequest] is `false` then this setting is ignored.
+     */
+    public var useFileForCache: Boolean = PlatformUtils.IS_JVM
```
olme04, line 32: isn't it better to make it configurable per request, and not via one boolean flag?
similar as with `excludeFromCache`?
So `File` cache will be used only for those requests, that can contain big body. If not, using file cache will decrease performance, as writing to File is almost always slower than writing to Memory

ktor-server/ktor-server-plugins/ktor-server-double-receive/jvm/src/io/ktor/server/plugins/doublereceive/FileCacheJvm.kt
InsanusMokrassar: Yes, that too

ktor-server/ktor-server-plugins/ktor-server-double-receive/jvmAndNix/src/io/ktor/server/plugins/doublereceive/DoubleReceiveConfig.kt
e5l: You can install the plugin only for the specific route which is supposed to have a big body. I'm not sure how it should be possible to configure for a specific request.

olme04: i mean use the same declaration as with `excludeFromCache`, so like:
```kotlin
fun useFileForCache(block: (call: ApplicationCall, body: Any) -> Boolean)
```
and in this case we will be able to f.e. do `file` cache when `contentLength` is big, or for specific content type, and not to apply this config for multiple routes

rsinukov: I agree, it should be configurable per request. Also, default should be opposite, IMO.

ktor-server/ktor-server-plugins/ktor-server-double-receive/jvm/src/io/ktor/server/plugins/doublereceive/FileCacheJvm.kt
rsinukov, line 19: Do you use the same file for all requests?

rsinukov (changes requested):

ktor-server/ktor-server-plugins/ktor-server-double-receive/jvm/src/io/ktor/server/plugins/doublereceive/FileCacheJvm.kt
e5l: Yep, `createTempFile` will append random number to the provided prefix

rsinukov: oh, sorry, missed this part

ktor-server/ktor-server-plugins/ktor-server-double-receive/jvmAndNix/src/io/ktor/server/plugins/doublereceive/DoubleReceiveConfig.kt
e5l: @olme04  
`fun useFileForCache(block: (call: ApplicationCall, body: Any) -> Boolean)` is actually a good idea. Thank you for the hint.

@rsinukov the file default seems to be not much slower, but much safer. I would keep it as a default, but we can discuss.

e5l: Yep. And I would even put default here:`fun useFileForCache(block: (call: ApplicationCall) -> Boolean = { true })`

rsinukov (approved):

PR #2703, opened 2021-11-24
KTOR-3461 KTOR-3460 Fix websocket session

rsinukov:

PR #2861, opened 2022-02-17
KTOR-2829 Add WebSocketSession hangs back

e5l:
Check for [KTOR-2829](https://youtrack.jetbrains.com/issue/KTOR-2829)

@rsinukov could you check, I can't reproduce it.


rsinukov:
It was fixed in #2703

e5l:
ok, will close with obsolete

PR #2862, opened 2022-02-18
KTOR-3652 handleResponseException should have access to request to provide valuable information in exceptions

lolilofit:
[KTOR-3652](https://youtrack.jetbrains.com/issue/KTOR-3652)

**Solution**
Create new method `handleResponseExceptionWithRequest` providing request

### DIFF
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/HttpCallValidator.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/HttpCallValidator.kt
@@ -5,13 +5,14 @@
 package io.ktor.client.plugins
 
 import io.ktor.client.*
+import io.ktor.client.call.*
+import io.ktor.client.call.HttpClientCall
 import io.ktor.client.plugins.HttpCallValidator.*
 import io.ktor.client.request.*
 import io.ktor.client.statement.*
 import io.ktor.client.utils.*
 import io.ktor.util.*
 import io.ktor.util.pipeline.*
-import kotlin.native.concurrent.*
 
 /**
  * Response validator method.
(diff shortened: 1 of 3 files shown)


ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/HttpCallValidator.kt
```diff
@@ -142,3 +164,9 @@ public var HttpRequestBuilder.expectSuccess: Boolean
     set(value) = attributes.put(ExpectSuccessAttributeKey, value)
 
 internal val ExpectSuccessAttributeKey = AttributeKey<Boolean>("ExpectSuccessAttributeKey")
+
+internal interface HandlerWrapper
```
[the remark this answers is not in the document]
lolilofit: Thanks for the help! I've made sealed class, but left it with `internal` because `HandlerWrapper` used in `HttpCallValidator` and `Config`

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/HttpCallValidator.kt
```diff
@@ -67,7 +78,15 @@ public class HttpCallValidator internal constructor(
          * Last added handler executes first.
          */
         public fun handleResponseException(block: CallExceptionHandler) {
-            responseExceptionHandlers += block
+            responseExceptionHandlers += ExceptionHandlerWrapper(block)
```
rsinukov, line 81: Maybe deprecate old method?

rsinukov (approved):

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/HttpCallValidator.kt
lolilofit: Thanks, I've deprecated old method

PR #2863, opened 2022-02-19
KTOR-728 Use conditional headers from response too

rsinukov:
When replying with data class, there is no way to set headers on `OutgoingContent` and headers set on response are ignored
https://youtrack.jetbrains.com/issue/KTOR-728


ktor-server/ktor-server-tests/jvm/test/io/ktor/server/plugins/ConditionalHeadersTests.kt
```diff
@@ -109,7 +135,7 @@ class LastModifiedTest(@Suppress("UNUSED_PARAMETER") name: String, zone: ZoneId)
     fun testIfModifiedSinceEqZoned() = testApplication {
         application {
             install(ConditionalHeaders) {
-                version { listOf(LastModifiedVersion(date)) }
+                version { _, _ -> listOf(LastModifiedVersion(date)) }
```
e5l, line 138: we can leave old version as well: it will keep source compatibility

e5l (approved):
lgtm

PR #2864, opened 2022-02-19
KTOR-3795 Break infinite refresh token cycle on 401

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-3795

When refresh token request fails with 401 it tries to refresh token again, resulting in an infinite cycle. Since token refreshing in a client's code, they need to mark such requests, so we can have a special case for them.  
Marking function is only available inside `refreshToken {}` block.

### DIFF
--- a/ktor-client/ktor-client-plugins/ktor-client-auth/common/test/io/ktor/client/plugins/auth/AuthTest.kt
+++ b/ktor-client/ktor-client-plugins/ktor-client-auth/common/test/io/ktor/client/plugins/auth/AuthTest.kt
@@ -13,7 +13,6 @@ import io.ktor.client.statement.*
 import io.ktor.client.tests.utils.*
 import io.ktor.http.*
 import io.ktor.util.*
-import io.ktor.utils.io.concurrent.*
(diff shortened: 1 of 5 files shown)


ktor-client/ktor-client-plugins/ktor-client-auth/common/test/io/ktor/client/plugins/auth/AuthTest.kt
```diff
@@ -419,6 +418,33 @@ class AuthTest : ClientLoader() {
...
+                            val token = clientWithAuth!!.get("$TEST_SERVER/auth/bearer/token/refresh-401") {
+                                markAsRefreshTokenRequest()
+                            }.bodyAsText()
+                            BearerTokens(token, token)
+                        }
+                    }
+                }
+            }
+
+            val result = clientWithAuth!!.get("$TEST_SERVER/auth/bearer/second")
```
e5l, line 443: maybe `lateinit` can drop `!!` assertions

e5l (approved):
lgtm

PR #2866, opened 2022-02-21
KTOR-3758 Add call to CachingHeadersConfig.options callback

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-3758


e5l (approved):
Good job, lgtm

PR #2867, opened 2022-02-21
KTOR-2435 Fix MDC context in client response body logging

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-2435


e5l:
Could you please check CI first?

e5l:
please also consider a test scenario, when user is receiving with logging on already finishing call

rsinukov:
@e5l thanks, I updated the implementation to use the same scope as before

e5l (approved):
LGTM!

PR #2869, opened 2022-02-23
KTOR-3940 Add isNotEmpty() and size methods to PlaceholderList

ustits:
**Subsystem**
ktor-server-html-builder

**Motivation**
[KTOR-3940](https://youtrack.jetbrains.com/issue/KTOR-3940)

**Solution**
Added `size` property and `isNotEmpty` method that reuse existing methods of `PlaceholderList`

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-html-builder/jvm/src/io/ktor/server/html/Template.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-html-builder/jvm/src/io/ktor/server/html/Template.kt
@@ -36,6 +36,11 @@ public open class Placeholder<TOuter> {
 public open class PlaceholderList<TOuter, TInner> {
     private var items = ArrayList<PlaceholderItem<TInner>>()
 
+    public val size: Int
+        get() {
+            return items.size
+        }
+
     public operator fun invoke(meta: String = "", content: TInner.(Placeholder<TInner>) -> Unit = {}) {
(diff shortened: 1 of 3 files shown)


ktor-server/ktor-server-plugins/ktor-server-html-builder/jvm/src/io/ktor/server/html/Template.kt
```diff
@@ -36,6 +36,11 @@ public open class Placeholder<TOuter> {
 public open class PlaceholderList<TOuter, TInner> {
     private var items = ArrayList<PlaceholderItem<TInner>>()
 
+    public val size: Int
+        get() {
+            return items.size
```
rsinukov, line 41: maybe convert to expression body?

rsinukov (approved):
LGTM, but please run `./gradlew apiDump` and commit changes.

ktor-server/ktor-server-plugins/ktor-server-html-builder/jvm/src/io/ktor/server/html/Template.kt
ustits: 👍

ustits:
Generated api, but now build fails on codestyle which doesn't seem to be connected with this PR

rsinukov (approved):

rsinukov:
Thanks for your contribution!

PR #2873, opened 2022-02-28
KTOR-3799 The "charset=UTF-8" part is added for the Content-Type header

lolilofit:
[KTOR-3799](https://youtrack.jetbrains.com/issue/KTOR-3799)


ktor-http/common/src/io/ktor/http/ContentTypes.kt
```diff
@@ -279,7 +279,19 @@ public class BadContentTypeFormatException(value: String) : Exception("Bad Conte
...
-public fun ContentType.withCharset(charset: Charset): ContentType = withParameter("charset", charset.name)
+public fun ContentType.withCharset(charset: Charset): ContentType =
+    withParameter("charset", charset.name)
+
+/**
+ * Creates a copy of `this` type with the added charset parameter with [charset] value
+ * if charset is not ignored by this [ContentType]
+ */
+public fun ContentType.withCharsetIfNeeded(charset: Charset): ContentType =
+    if (charset.name == "UTF-8" && contentType == "application" && contentSubtype == "json") {
```
roomscape, line 290: I don't know if it's correct to single out UTF-8 here.

The `application/json` content type does not have _any_ charset parameter, regardless of value. [RFC 4627](https://datatracker.ietf.org/doc/html/rfc4627#section-3) specifies that the character encoding of a JSON text is always auto-detected by the client (from among a limited set of options).

So `application/json; charset=UTF-16` would incorrect just as `application/json; charset=UTF-8` would be. Neither is a valid content type.

lolilofit: You are right, thanks for the help! 
I've ignored only UTF-8 because I thought Ktor ignored only this charset with application/json in 1.x.x versions. But I've checked the 1.6.7 version now and I see I was wrong.

ktor-client/ktor-client-tests/jvm/test/io/ktor/client/tests/ExceptionsJvmTest.kt
```diff
@@ -62,4 +71,22 @@ class ExceptionsJvmTest {
...
+
+    @Test
+    fun testContentNegotioationMediaType() = runBlocking {
+        val client = HttpClient(Apache) {
+            install(ContentNegotiation) {
+                json()
+            }
+        }
+
+        val response = client.post("https://httpbin.org/post") {
```
rsinukov, line 83: better to avoid 3rd party services and real network in tests

ktor-server/ktor-server-plugins/ktor-server-content-negotiation/jvmAndNix/src/io/ktor/server/plugins/contentnegotiation/ResponseConverter.kt
```diff
@@ -38,11 +39,15 @@ internal fun PluginBuilder<ContentNegotiationConfig>.convertResponseBody() = onC
...
         }
 
+        val acceptCharset = call.request.headers.suitableCharsetOrNull()
+
         // Pick the first one that can convert the subject successfully
         val converted: OutgoingContent? = suitableConverters.firstNotNullOfOrNull {
             it.converter.serialize(
-                contentType = it.contentType,
-                charset = call.request.headers.suitableCharset(),
+                contentType = acceptCharset?.let { charset ->
```
rsinukov, line 47: please extract to `val`

rsinukov (approved):

ktor-client/ktor-client-tests/jvm/test/io/ktor/client/tests/ExceptionsJvmTest.kt
lolilofit: Thanks for the review! I've removed this service from the test.

PR #2874, opened 2022-02-28
KTOR-3194 - enable customize default tags in micrometer

LucasMilotich:
Subsystem
Ktor-metrics-micrometer

Motivation
Feature: https://youtrack.jetbrains.com/issue/KTOR-3194

Sometimes you need to override default tags and the only way was using this sentence in MicrometerMetrics installation:
metrics.config().meterFilter(MeterFilter.ignoreTags($tag))

Solution
I've created a default tags field which allows to be overwritten in the configuration class.
Some fields/classes have to be internal in order to allow the client to access and customize tags using the call.


e5l:
Rebased in #5210

PR #2878, opened 2022-03-02
KTOR-3966 add links to a migration guide

andreyaksenov:
Added links to a migration guide for some deprecated members.
Note that these links will work AFTER releasing v2.0.0 as currently a migration guide is only available under https://ktor.io/docs/eap.


e5l (approved):
LGTM, please check code style before merging

andreyaksenov:
> LGTM, please check code style before merging

thanks, fixed

PR #2885, opened 2022-03-08
Release 1.6.8

e5l:


gradle/common.gradle
```diff
@@ -1,12 +1,17 @@
 
-kotlin.sourceSets {
-    commonMain.dependencies {
-        api "org.jetbrains.kotlin:kotlin-stdlib-common"
-        api group: "org.jetbrains.kotlinx", name: "kotlinx-coroutines-core", version: coroutines_version
+kotlin {
+    jvm {
```
olme04, line 3: should be in jvm.gradle, no?

gradle/libs.versions.toml
```diff
@@ -0,0 +1,105 @@
+[versions]
+kotlin-version = "1.6.10"
```
olme04, line 2: may be remove `version` suffix?

build.gradle.kts
```diff
@@ -53,15 +53,13 @@ buildscript {
     val kotlin_version: String by extra
     val atomicfu_version: String by extra
     val validator_version: String by extra
-    val android_gradle_version: String by extra
     val ktlint_version: String by extra
 
     dependencies {
         classpath("org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version")
```
olme04, line 59: version catalogs can be used here also

gradle/libs.versions.toml
olme04, line 2: does it make sense to store `kotlin` version and `kotlin` dependencies here, as they are overridable by in root `build.gradle` file?
Looks like it can cause unpredictable issues

olme04 (commented):
Looks like, adopting `version catalogs` should be a separate PR not coupled with 1.6.8 release :)

gradle/common.gradle
e5l: yep

build.gradle.kts
e5l: Thanks, will replace

gradle/libs.versions.toml
e5l: They are not used directly, only as refs and this file. I'm trying to keep short names for the libs themself

e5l: I'm going to drop version overriding in `1.6.*` branch

ktor-client/ktor-client-curl/build.gradle.kts
```diff
@@ -3,10 +3,9 @@ import org.jetbrains.kotlin.gradle.targets.native.tasks.KotlinNativeTest
 import org.jetbrains.kotlin.gradle.plugin.mpp.KotlinNativeTarget
 
 val ideaActive: Boolean by project.extra
-val serialization_version: String by project.extra
 
 plugins {
-    id("kotlinx-serialization")
+    kotlin("plugin.serialization") version "1.6.10"
```
rsinukov, line 8: can you not hardcore the version here and below for `serialization`?

rsinukov (approved):
lgtm, but one comment

PR #2893, opened 2022-03-11
OverridingClassLoader now delegates to parent for resources also.

spand:
Previously it was (presumably) only able to find resources on its own classpath.

KTOR-4004

**Subsystem**
Server

**Motivation**
Classes were unable to load resources from the classpath when loaded in OverridingClassLoader

**Solution**
A breaking test demonstrates the problem and I have submitted a fix.

### DIFF
--- a/ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/OverridingClassLoader.kt
+++ b/ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/OverridingClassLoader.kt
@@ -4,8 +4,9 @@
 
 package io.ktor.server.engine
 
-import io.ktor.utils.io.core.*
+import java.io.*
 import java.net.*
+import java.util.*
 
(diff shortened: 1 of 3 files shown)


e5l (approved):
LGTM, will merge after review

e5l:
Hey @spand, could you check the CI build?

spand:
@e5l
Is it good now ? I think I saw another problem but all is green. Do you need to trigger a full test suite manually ?

e5l:
Sure, in a moment

spand:
I reworked it a bit so maybe it is fixed but I cannot reproduce the problem locally.

spand:
@e5l I placed an ignore on the failing test. It seems pretty doomed for me to try to fix this test without access to the environment. Cant we agree that the changes are simple and makes sense enough to get this fix merged?

e5l:
@spand, sure

PR #2895, opened 2022-03-14
KTOR-3945 Fix OAuth scopes encoded to %2B

ishitatsuyuki:
**Subsystem**
Server, OAuth2 module.

**Motivation**
See KTOR-3945

**Solution**
Separate scopes with space. These are encoded to "+" by the URL encoder. (Generally encoding to "%20" is also acceptable, so I'm not going to test that here.)

Also fix the tests that were failing to catch this due to accidental double-decoding of query string.

Question: Can we catch these double-decoding errors systematically? IDE inspections? Type-level guards?


rsinukov (approved):
Great catch, thanks. Will merge after CI check

PR #2899, opened 2022-03-15
Cio native websockets

olme04:
**Subsystem**
Client/Server, WebSockets

**Motivation**
[[client] MPP WebSockets client](https://youtrack.jetbrains.com/issue/KTOR-751)
[Native websocket client support](https://youtrack.jetbrains.com/issue/KTOR-599)
No CIO native WebSockets implementation

**Solution**

Build preparation:
* cleanup JNA workaround
* use java 8 for buildSrc via toolchain
* minor cleanup of kotlin stdlib/test dependencies
* make nix depends on posix and jvmAndNix depends on common in tests
* cleanup server plugins dependencies setup

Common websockets prepration:
* align frame actuals between platforms
* move js ws code to correct folder
* remove expects from WebSocketSession - masking is already not supported in all engines (to be correct, it's mutable only in CIO, and immutable in Java and OkHttp engines)
* remove expects from DefaultWebSocketSession - ping-pong is not supported only on JS, and in okhttp it's not mutable
* commonize basic DefaultWebSocketSession and RawWebSocket
* make ping/pong internal

CIO native websockets:
* fix charsets for native
* copy SHA1 impl for native from okio
* make Server websockets plugin working on nix
* adapt server WS tests to work on both jvm and nix
* make WebSocketRemoteTest work for native:CIO


ktor-client/ktor-client-core/js/src/io/ktor/client/plugins/websocket/JsWebSocketSession.kt
```diff
@@ -30,9 +30,21 @@ internal class JsWebSocketSession(
...
+        set(_) = throw WebSocketException("Websocket timeout is not supported in JS engine.")
+
+    override var masking: Boolean
+        get() = true
+        set(_) = throw WebSocketException("Masking switch is not supported in JS engine.")
+
     override var maxFrameSize: Long
         get() = Long.MAX_VALUE
-        set(_) {}
+        set(_) = throw WebSocketException("Max frame size switch is not supported in OkHttp engine.")
```
e5l, line 47: in Js engine

ktor-client/ktor-client-core/js/src/io/ktor/client/plugins/websocket/JsWebSocketSession.kt
```diff
@@ -42,21 +54,17 @@ internal class JsWebSocketSession(
...
+                    else -> {
+                        val error = IllegalStateException("Unknown frame type: ${event.type}")
+                        _closeReason.completeExceptionally(error)
+                        throw error
                     }
-
-                    _incoming.trySend(frame).isSuccess
                 }
+
+                _incoming.trySend(frame)
```
e5l, line 67: Should we check the operation result here?

ktor-client/ktor-client-core/js/src/io/ktor/client/plugins/websocket/JsWebSocketSession.kt
```diff
@@ -73,19 +81,15 @@ internal class JsWebSocketSession(
...
-                    val reason = CloseReason(event.code as Short, event.reason as String)
-                    _closeReason.complete(reason)
-                    _incoming.send(Frame.Close(reason))
-                    _incoming.close()
-
-                    _outgoing.cancel()
-                }
+                val reason = CloseReason(event.code as Short, event.reason as String)
+                _closeReason.complete(reason)
+                _incoming.trySend(Frame.Close(reason))
```
e5l, line 86: and here

ktor-server/ktor-server-plugins/ktor-server-websockets/jvmAndNix/src/io/ktor/server/websocket/WebSocketUpgrade.kt
```diff
@@ -96,8 +96,11 @@ public class WebSocketUpgrade(
 
         webSocket.launch(WebSocketHandlerCoroutineName) {
             try {
-                webSocket.start(handle)
+                webSocket.handle()
+                webSocket.flush()
             } catch (cause: Throwable) {
+            } finally {
+                webSocket.cancel()
```
e5l, line 103: could you explain why we need cancel here?

ktor-shared/ktor-websockets/common/src/io/ktor/websocket/DefaultWebSocketSession.kt
```diff
@@ -153,6 +199,7 @@ public class DefaultWebSocketSessionImpl(
             ponger.close()
             filtered.close(cause)
         } finally {
+            println("DEFAULT_INCOMING")
```
e5l, line 202: Please remove all printlns

ktor-server/ktor-server-plugins/ktor-server-websockets/jvmAndNix/test/io/ktor/tests/websocket/RawWebSocketTest.kt
```diff
@@ -5,39 +5,34 @@
...
-import org.junit.rules.*
-import java.io.*
 import kotlin.reflect.*
 import kotlin.test.*
 
 @OptIn(ExperimentalCoroutinesApi::class)
 class RawWebSocketTest {
-    @get:Rule
-    val timeout: CoroutinesTimeout = CoroutinesTimeout.seconds(10, true)
+//    @get:Rule
```
e5l, line 20: Could you make `expect`-`actual`? We need to keep timeout at least for jvm

ktor-server/ktor-server-plugins/ktor-server-websockets/jvmAndNix/test/io/ktor/tests/websocket/RawWebSocketTest.kt
```diff
@@ -202,8 +197,9 @@ class RawWebSocketTest {
     }
 
     private fun runTest(block: suspend CoroutineScope.() -> Unit) {
-        runBlocking(CoroutineName("test-${test.methodName}")) {
+        runBlocking(CoroutineName("test-")) {
```
e5l, line 200: Could we find a decent replacement?

ktor-server/ktor-server-test-host/nix/src/EngineTestBaseNix.kt
```diff
@@ -29,7 +29,7 @@ actual abstract class EngineTestBase<TEngine : ApplicationEngine, TConfiguration
     actual val applicationEngineFactory: ApplicationEngineFactory<TEngine, TConfiguration>
 ) : CoroutineScope {
     private val testJob = Job()
-    actual override val coroutineContext: CoroutineContext = testJob + Dispatchers.Main
+    actual override val coroutineContext: CoroutineContext = testJob + Dispatchers.Unconfined
```
e5l, line 32: Please consider making `expect`-`actual` method for dispatcher. I would also avoid using `Unconfined` here

ktor-utils/posix/src/io/ktor/util/CryptoNative.kt
```diff
@@ -23,4 +23,207 @@ public actual fun Digest(name: String): Digest = error("[Digest] is not supporte
 /**
  * Compute SHA-1 hash for the specified [bytes]
  */
-public actual fun sha1(bytes: ByteArray): ByteArray = error("sha1 currently is not supported in ktor-native")
+public actual fun sha1(bytes: ByteArray): ByteArray {
```
e5l, line 26: We need to make formating and add implementation tests

ktor-shared/ktor-websockets/js/src/io/ktor/websocket/RawWebSocket.kt
```diff
@@ -0,0 +1,17 @@
...
+ * Copyright 2014-2022 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.websocket
+
+import io.ktor.utils.io.*
+import kotlin.coroutines.*
+
+@Suppress("FunctionName")
+public actual fun RawWebSocket(
```
e5l, line 11: Consider adding KDoc for public declarations

e5l (changes requested):
Hey @olme04, awesome start. There is some stuff to fix, but in general, it's almost ready. Thanks for the PR!

ktor-client/ktor-client-core/js/src/io/ktor/client/plugins/websocket/JsWebSocketSession.kt
olme04: looks like no, as channel is unlimited and closed based on events from socket, so if channel is closed, we are already failing. Looks like it was before `offer` and just replaced with auto migration from coroutines.

But, if channel is closed externally (by user), we can't subscribe to it, but we can handle it here. 
Not sure, if such operations are supported in other WebSocketSession implementation. Will check it

ktor-server/ktor-server-plugins/ktor-server-websockets/jvmAndNix/src/io/ktor/server/websocket/WebSocketUpgrade.kt
olme04: it's inlined `start` function:
https://github.com/ktorio/ktor/pull/2899/files/7ec0dc58a9ecdf887347ee55b037d3e04cc7817a#diff-b3700a64a6cb2f618bcaf420986cddf505e8fc5de9aff09b175b3f6bc3cd0ff6L85-L92

Current behavior is that after `handler` lambda is finished, we should close websocket, so call `cancel`. I didn't want to change behavior here.

ktor-server/ktor-server-plugins/ktor-server-websockets/jvmAndNix/test/io/ktor/tests/websocket/RawWebSocketTest.kt
olme04: I think, that the only possible way is to use Throwable stacktrace or some compiler plugin.
For now I return it back via expect-actual

ktor-server/ktor-server-test-host/nix/src/EngineTestBaseNix.kt
olme04: Main is available only on `darwin` targets on `nix`. May be we can use `Default` on all `nix` targets?
I was used `Unconfined` here instead of `Default`, as `ktor-network` sockets closing somehow failing the whole process, but I have not found the root cause of it.
I will try to investigate more

e5l: Didn't notice `Main` here. We should replace it with `IO` or `Default`

olme04: FYI, it's `nix` specific code.
There is no `IO` on native on current moment, so `Default` is the best choice, as it's also used as JVM replacement of `IO` on `nix` in `ktor`

ktor-client/ktor-client-core/js/src/io/ktor/client/plugins/websocket/JsWebSocketSession.kt
olme04: Looks like no other implementations except `test-host` implementation handles channels close events, so let's leave it like this for now in this PR

olme04:
Hey, @e5l can you take a look on `BaseTest` file for a proposal of improving and generalizing tests for MPP cases
https://github.com/ktorio/ktor/pull/2899/commits/d7116bcb1cf27980966e3d3f44b9f3ac1b563682#diff-56fa798e59914472f0d7dec474f350c4df146bf696a9919951c905cf3f5e2690

> it supports:
    * timeout - for JVM via junit rule, on nix(and possibly all other targets) via `withTimeout`
    * test name - for JVM via junit rule, on nix via block qualified name (for that, should be called like `fun test() = runTest {}`
    * collecting and reporting unhandled errors - for JVM via junit rule, on nix via synchronized list and assertion in `AfterTest`

ktor-server/ktor-server-test-host/nix/src/EngineTestBaseNix.kt
olme04: replaced with Default, but anyway I need to use `Unconfined` in one test (testWebSocketDisconnectDuringSending) for using ktor-network socket in it. If not to do it, then it will fail process with no exception thrown.
For now, I can't find the root cause of it :(

olme04:
Ready for review

e5l:
Hey @olme04, could you check compilation issues?

olme04:
@e5l Done

ktor-server/ktor-server-test-host/jvmAndNix/src/io/ktor/server/testing/BaseTest.kt
```diff
@@ -0,0 +1,19 @@
+/*
+ * Copyright 2014-2022 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.server.testing
+
+import kotlinx.coroutines.*
+import kotlin.time.*
+
+// TODO: can be extracted from `server-testing` module, fully commonized(and js and mingw targets) and used in all ktor-tests by default
```
e5l, line 10: could you log issues for TODO? and drop todo

e5l (changes requested):
LGTM in general.

Could you check native tests? It looks like it's related.
The code-style build is also failing

ktor-server/ktor-server-test-host/jvmAndNix/src/io/ktor/server/testing/BaseTest.kt
olme04: https://youtrack.jetbrains.com/issue/KTOR-4038 done

olme04:
* lint error fixed
* TODOs removed
* increased timeout for running tests in `WebSocketEngineSuite` - on my machine it's ~10 seconds for longest test
* CallIdTest failure on macos looks like not related to current PR

e5l (approved):
lgtm

PR #2900, opened 2022-03-15
KTOR-3979 CORS `allowHeadersPrefixed` should ignore case

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-3979


e5l (approved):
good catch, lgtm

PR #2905, opened 2022-03-16
KTOR-4025 Move ApplicationCallPlugin to companion object

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-4025


e5l (approved):
Please check API dump

PR #2906, opened 2022-03-16
KTOR-4028 Fix Compression plugin memory leak

flore77:
**Subsystem**
Utils

**Motivation**
The Deflater implementation is causing memory leaks when the response channel gets closed (for example when a peer resets the connection) and the gzipped content is larger than the ByteBufferChannel backing buffer (4088). More details in the YT ticket: https://youtrack.jetbrains.com/issue/KTOR-4028

**Solution**
Refactored the `ByteWriteChannel.deflated` method to not use an intermediary channel, but instead to write the compressed body directly into the response channel. This will fix the memory leak and maybe at the same time bring also a small performance improvement by not needing to spawn an extra coroutine and an extra copy across buffers.

Another solution could have been to pass the coroutineContext of the first coroutine to the second one, instead of the coroutineContext received as parameter: 

```kotlin
public fun ByteWriteChannel.deflated(
    ..,
    coroutineContext: CoroutineContext = Dispatchers.Unconfined
): ByteWriteChannel = GlobalScope.reader(coroutineContext, autoFlush = true) {
    // Use this.coroutineContext to not break parent-child relationship
    channel.deflated(gzip, pool, this.coroutineContext).copyTo(this@deflated)
}.channel
```
Through this when the coroutine started by `GlobalScope.reader` would have finished it would have cancelled the one started by `channel.deflated(...)`, but I think the refactor is a better approach.

### DIFF
--- a/ktor-utils/jvm/src/io/ktor/util/Deflater.kt
+++ b/ktor-utils/jvm/src/io/ktor/util/Deflater.kt
@@ -54,15 +54,14 @@ private suspend fun ByteWriteChannel.deflateWhile(deflater: Deflater, buffer: By
 }
 
 /**
- * Launch a coroutine on [coroutineContext] that does deflate compression
+ * Does deflate compression
  * optionally doing CRC and writing GZIP header and trailer if [gzip] = `true`
  */
-@OptIn(DelicateCoroutinesApi::class)
-public fun ByteReadChannel.deflated(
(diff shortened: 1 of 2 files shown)


e5l (commented):
Hey @flore77, thanks for the PR!

Good Job, LGTM

e5l (approved):

PR #2911, opened 2022-03-18
[PROTOTYPE] CIO server HTTPS support

olme04:
**Subsystem**
CIO, Server, network-tls

**Motivation**
No support for TLS in CIO

**Solution**
On current moment, there is only CIO client JVM TLS implementation, which uses handwritten TLS handshake mechanism. This is for now also don't support TLS 1.3. 
I propose to switch CIO SSL to `SSLEngine` (netty also uses it under the hood if not using native openssl).
For now PR contains just a POC of integration. But all server tests(except one) works with SSL enabled.
Also, I'm in progress of adding SSL support to `native(nix)` via native openssl for both server and client.

Questions:
* What is the best way to integrate/replace CIO client SSL configuration?
* SSL configuration
  * As for now, servers HTTPS configuration is done via connectors, so no engine-specific API exists to configure it, but there are several configuration properties, and only part of them are used in different engines
  * On the other side, on client side, all https configuration happened in engine specific configuration
  * May be we can somehow improve it?
* It's also possible to do a secure UDP connection, via `SSLEngine`, do we need it?

### DIFF
--- a/ktor-client/ktor-client-cio/jvmAndNix/src/io/ktor/client/engine/cio/Endpoint.kt
+++ b/ktor-client/ktor-client-cio/jvmAndNix/src/io/ktor/client/engine/cio/Endpoint.kt
@@ -166,6 +166,7 @@ internal class Endpoint(
         pipeline.pipelineContext.invokeOnCompletion { releaseConnection() }
     }
 
+    @OptIn(InternalAPI::class)
     private suspend fun connect(requestData: HttpRequestData): Connection {
         val connectAttempts = config.endpoint.connectAttempts
         val (connectTimeout, socketTimeout) = retrieveTimeouts(requestData)
@@ -204,11 +205,7 @@ internal class Endpoint(
                     if (proxy?.type == ProxyType.HTTP) {
                         startTunnel(requestData, connection.output, connection.input)
                     }
-                    val tlsSocket = connection.tls(coroutineContext) {
-                        takeFrom(config.https)
-                        serverName = serverName ?: address.hostname
-                    }
-                    return tlsSocket.connection()
+                    return connection.ssl(coroutineContext + CoroutineName("CLIENT"), SslContext(config.https).createClientEngine())
                 } catch (cause: Throwable) {
                     try {
                         socket.close()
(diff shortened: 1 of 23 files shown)


olme04:
UPDATE:
* implemented nix(linux only for now) client and server SSL
* minimal hardcoded certificates for tests (public API for providing certificates on nix is still WIP)

Notes:
* Current cinterop for openssl uses linker options to directory inside `ktor` project for libssl and libcrypto. If use paths from `/usr/lib/`, `/usr/lib64/` and similars, we will have link issue, because kotlin-native uses it's own GCC toolchain [TODO: link to kotlin issue]
* On other hand, curl uses `/usr/lib` and others for linker, which is bad, as we link to unspecified libraries, which could not work with kotlin-native, and this should be migrated [TODO: link to kotlin issue with better description of the issue].
* As for now, cinterop/linkDebug for both `openssl` and `curl` fails with error, that it can't link binary, looks like because of `curl` client uses wrong libs, 
* I have no way now to test/debug darwin targets, so this PR will contain only linux part of native ssl support
*  Darwin targets: there is 2 ways to do ssl there:
   1. same as with linux by linking openssl
   2. using rich darwin platform and use something like [Secure transport](https://developer.apple.com/documentation/security/secure_transport) or [Network](https://developer.apple.com/documentation/network)
* current nix ssl implementation don't handle all possible SSL errors, so more tests need to be provided

Questions:
* are there in ktor some common client tests, which checks, that SSL is working?
* I'm thinking of extracting implemented ssl things in new module called `ktor-network-ssl` as it uses both for jvm and nix targets much more general approach, that will be possible to use for both TCP and UDP with bigger amount of available algorithms and protocol versions (f.e. `network-tls` is TLS 1.2 only, but both openssl 1.1 and SSLEngine supports TLS 1.1 - 1.3). And leave `ktor-network-tls` untouched, at least for backward compatibility. WDYT?
* After polishing current API I would want to also implement nodejs sockets support (both raw and with SSL, which is provided out of the box in node) - As on current moment we can't create separate targets for browser and node, it will be in a separate artifact, like `ktor-network-node`. But in future it can be integrated inside `ktor-network`. Question is, that looks like nodejs sockets don't need selector provided at all, and as so, may be we also need to somehow change API for socket creation to support both ways?

olme04:
more or less working POC is here: https://github.com/olme04/ktor/tree/network-security-with-libs

PR #2924, opened 2022-03-25
KTOR-4058 Use request timeout instead of socket timeout in Java client

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-4058


ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/HttpTimeoutTest.kt
```diff
@@ -504,7 +504,7 @@ class HttpTimeoutTest : ClientLoader() {
     }
 
     @Test
-    fun testSocketTimeoutWriteFailOnWrite() = clientTests(listOf("Js", "Android", "native:CIO")) {
+    fun testSocketTimeoutWriteFailOnWrite() = clientTests(listOf("Js", "Android", "native:CIO", "Java")) {
```
e5l, line 507: should we drop this test?

e5l (approved):
lgtm

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/HttpTimeoutTest.kt
rsinukov: why? it works for engines that use `socketTimeout` from config

PR #2927, opened 2022-03-29
KTOR-506 Fix selector manager

e5l:
Fix [KTOR-506](https://youtrack.jetbrains.com/issue/KTOR-506)

### DIFF
--- a/ktor-network/jvm/src/io/ktor/network/selector/ActorSelectorManager.kt
+++ b/ktor-network/jvm/src/io/ktor/network/selector/ActorSelectorManager.kt
@@ -30,7 +30,7 @@ public class ActorSelectorManager(context: CoroutineContext) : SelectorManagerSu
     @Volatile
     private var closed = false
 
-    private val mb = LockFreeMPSCQueue<Selectable>()
+    private val selectionQueue = LockFreeMPSCQueue<Selectable>()
 
     override val coroutineContext: CoroutineContext = context + CoroutineName("selector")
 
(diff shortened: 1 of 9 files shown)


olme04:
@e5l can you please describe, where is the main fix happen? Because, while looking on diff, it's hard to understand where the fix is, as there is a lot of renamings :)

e5l:
@olme04, thanks for the notice. Let me check if I pushed all changes

e5l:
@olme04 yep, all is in place. ActorSelectorManager.kt:143 line

olme04:
@e5l thx!

ktor-network/jvm/src/io/ktor/network/selector/ActorSelectorManager.kt
```diff
@@ -30,34 +30,32 @@ public class ActorSelectorManager(context: CoroutineContext) : SelectorManagerSu
...
     init {
         launch {
             val selector = provider.openSelector() ?: error("openSelector() = null")
             selectorRef = selector
-            try {
+            selector.use {
                 try {
-                    process(mb, selector)
-                } catch (t: Throwable) {
+                    process(selectionQueue, it)
```
rsinukov, line 43: please extract `it` to named parameter

rsinukov (approved):
lgtm

ktor-network/jvm/src/io/ktor/network/selector/ActorSelectorManager.kt
e5l: thanks, sure

PR #2929, opened 2022-03-30
CIO server HTTPS support on JVM

olme04:
**Subsystem**
Server, CIO, Network

**Motivation**
Fixes https://youtrack.jetbrains.com/issue/KTOR-694

**Solution**
Sses JDK SSLEngine for TLS implementation
Also, supports client TLS, but not used for now

TODO: 
* one test failed - HttpServerCommonTestSuite.testProxyHeaders - not sure what is the reason

### DIFF
--- a/ktor-network/ktor-network-tls/jvm/src/io/ktor/network/tls/SSLEngineBufferAllocator.kt
+++ b/ktor-network/ktor-network-tls/jvm/src/io/ktor/network/tls/SSLEngineBufferAllocator.kt
@@ -59,5 +59,4 @@ internal class SSLEngineBufferAllocator(private val engine: SSLEngine) {
         new.put(buffer)
         return new
     }
-
 }
--- a/ktor-network/ktor-network-tls/jvm/src/io/ktor/network/tls/SSLEngineBufferAllocator.kt
+++ b/ktor-network/ktor-network-tls/jvm/src/io/ktor/network/tls/SSLEngineBufferAllocator.kt
@@ -1,0 +1,63 @@
+/*
+ * Copyright 2014-2022 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.network.tls
+
+import java.nio.*
+import javax.net.ssl.*
+
+internal class SSLEngineBufferAllocator(private val engine: SSLEngine) {
+    private var packetBufferSize = 0
+    private var applicationBufferSize = 0
+
+    fun allocatePacket(length: Int): ByteBuffer = allocate(
+        length,
+        get = { packetBufferSize },
+        set = { packetBufferSize = it },
+        new = { engine.session.packetBufferSize }
+    )
+
+    fun allocateApplication(length: Int): ByteBuffer = allocate(
+        length,
+        get = { applicationBufferSize },
+        set = { applicationBufferSize = it },
+        new = { engine.session.applicationBufferSize }
+    )
+
+    fun reallocatePacket(buffer: ByteBuffer, flip: Boolean): ByteBuffer =
+        reallocate(buffer, flip, ::allocatePacket)
+
+    fun reallocateApplication(buffer: ByteBuffer, flip: Boolean): ByteBuffer =
+        reallocate(buffer, flip, ::allocateApplication)
+
+    private inline fun allocate(
+        length: Int,
+        get: () -> Int,
+        set: (Int) -> Unit,
+        new: () -> Int
+    ): ByteBuffer = synchronized(this) {
+        if (get() == 0) {
+            set(new())
+        }
+        if (length > get()) {
+            set(length)
+        }
+        return ByteBuffer.allocate(get())
+    }
+
+    private inline fun reallocate(
+        buffer: ByteBuffer,
+        flip: Boolean,
+        allocate: (length: Int) -> ByteBuffer
+    ): ByteBuffer {
+        val newSize = buffer.capacity() * 2
+        val new = allocate(newSize)
+        if (flip) {
+            new.flip()
+        }
+        new.put(buffer)
+        return new
+    }
+
+}
--- a/ktor-network/ktor-network-tls/jvm/src/io/ktor/network/tls/SSLEngineSocket.kt
+++ b/ktor-network/ktor-network-tls/jvm/src/io/ktor/network/tls/SSLEngineSocket.kt
@@ -1,0 +1,168 @@
+/*
+ * Copyright 2014-2022 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.network.tls
+
+import io.ktor.network.sockets.*
+import io.ktor.utils.io.*
+import kotlinx.atomicfu.*
+import kotlinx.coroutines.*
+import kotlinx.coroutines.sync.*
+import javax.net.ssl.*
+import kotlin.coroutines.*
+
+internal class SSLEngineSocket(
+    override val coroutineContext: CoroutineContext,
+    private val engine: SSLEngine,
+    connection: Connection
+) : Socket, CoroutineScope {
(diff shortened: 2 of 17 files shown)


e5l:
Please change the target branch to 2.1

ktor-network/ktor-network-tls/jvm/src/io/ktor/network/tls/SSLEngineUnwrapper.kt
```diff
@@ -0,0 +1,107 @@
...
+
+    private fun resumeUnwrapContinuation(result: SSLEngineResult?) {
+        synchronized(this) {
+            unwrapResultCont?.resume(result)
+            unwrapResultCont = null
+        }
+    }
+
+    private suspend fun readData(): Boolean {
+        val read = input.readAvailable(unwrapSource)
```
e5l, line 103: Could we feed SSLEngine directly from the channel without a buffer? (method providing a block).

We also should be careful if the channel gets canceled with an exception.

e5l: I would try to split up several extensions on ByteReadChannel and ByteWriteChannel feeding SSLEngine with the available data

ktor-network/ktor-network-tls/jvm/src/io/ktor/network/tls/SSLEngineUnwrapper.kt
```diff
@@ -0,0 +1,107 @@
...
+    fun cancel(cause: Throwable?) {
+        input.cancel(cause)
+    }
+
+    @Suppress("BlockingMethodInNonBlockingContext")
+    suspend inline fun readAndUnwrap(
+        initialUnwrapDestination: ByteBuffer,
+        updateUnwrapDestination: (ByteBuffer) -> Unit
+    ): SSLEngineResult? {
+        if (!unwrapLock.tryLock()) return suspendCancellableCoroutine {
```
e5l, line 36: is it possible to directly wrap

ktor-network/ktor-network-tls/jvm/test/io/ktor/network/tls/tests/ConnectionTest.kt
```diff
@@ -53,8 +55,93 @@ class ConnectionTest {
...
+                    .use { socket ->
+                        socket.openWriteChannel().apply {
+                            writeStringUtf8("GET / HTTP/1.1\r\n")
+                            writeStringUtf8("Host: www.google.com\r\n")
+                            writeStringUtf8("Connection: close\r\n")
+                            flush()
+                        }
+                        val reader = socket.openReadChannel()
+                        repeat(3) {
+                            println("CCC: ${assertNotNull(reader.readUTF8Line())}")
```
e5l, line 139: please drop printlns

ktor-network/ktor-network-tls/jvm/src/io/ktor/network/tls/SSLEngineSocket.kt
```diff
@@ -0,0 +1,145 @@
...
+        }
+    }
+
+    private suspend fun doHandshake(initialStatus: SSLEngineResult.HandshakeStatus) = lock.withLock {
+        var temp = bufferAllocator.allocateApplication(0)
+        var status = initialStatus
+        while (true) {
+            when (status) {
+                SSLEngineResult.HandshakeStatus.NEED_TASK -> {
+                    coroutineScope {
```
e5l, line 121: coroutineScope will create a new job for every `NEED_TASK` status.

ktor-network/ktor-network-tls/jvm/src/io/ktor/network/tls/SSLEngineSocket.kt
```diff
@@ -0,0 +1,145 @@
...
+                        while (true) {
+                            val task = engine.delegatedTask ?: break
+                            launch(Dispatchers.IO) { task.run() }
```
e5l, line 124: same here

ktor-network/ktor-network-tls/jvm/src/io/ktor/network/tls/SSLEngineSocket.kt
```diff
@@ -0,0 +1,145 @@
...
+            doClose(null)
+            return
+        }
+        if (result.handshakeStatus.needHandshake) {
+            doHandshake(result.handshakeStatus)
+        }
+    }
+
+    private suspend fun doClose(cause: Throwable?) = lock.withLock {
+        if (closed.compareAndSet(expect = false, update = true)) {
```
e5l, line 108: if (!closed.compareAndSet(expect = false, update = true)) return

ktor-network/ktor-network-tls/jvm/src/io/ktor/network/tls/SSLEngineBufferAllocator.kt
```diff
@@ -0,0 +1,62 @@
...
+            set(length)
+        }
+        return ByteBuffer.allocate(get())
+    }
+
+    private inline fun reallocate(
+        buffer: ByteBuffer,
+        flip: Boolean,
+        allocate: (length: Int) -> ByteBuffer
+    ): ByteBuffer {
```
e5l, line 53: Could we just make large enough buffers upfront and use a regular pool? the maximum packet size is limited by the protocol

e5l (changes requested):

ktor-network/ktor-network-tls/jvm/src/io/ktor/network/tls/SSLEngineBufferAllocator.kt
whyoleg: sslengine max buffer size is 16K, and we need up to 4 such buffers for every TLS socket implementation, so 64K per socket.
For client side - it's ok, as there will not so much active sockets at the same time, but for server side it can cause memory issues with a lot of clients.
Note: most of the time, those 16K are not fully used, and buffer of smaller size is enough to perform TLS operations

ktor-network/ktor-network-tls/jvm/src/io/ktor/network/tls/SSLEngineSocket.kt
whyoleg: we can remove this, but `delegatedTask` is potentially blocking task, and then, we should make sure, that when creating TLS socket, we don't provide Dispatchers.Default or other non IO dispatcher to it constructor.

ktor-network/ktor-network-tls/jvm/src/io/ktor/network/tls/SSLEngineUnwrapper.kt
```diff
@@ -0,0 +1,107 @@
...
+                    else -> break
+                }
+            }
+            unwrapRemaining = unwrapSource.remaining()
+            if (unwrapDestination !== initialUnwrapDestination) updateUnwrapDestination(unwrapDestination)
+            resumeUnwrapContinuation(result)
+            return result
+        } catch (cause: Throwable) {
+            synchronized(this) {
+                unwrapResultCont?.resumeWithException(cause)
```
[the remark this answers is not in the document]
whyoleg: should not be an issue, as no `suspend` functions are called under `synchronization`
But, this can be done via simple atomic, can change to it

ktor-network/ktor-network-tls/jvm/src/io/ktor/network/tls/SSLEngineUnwrapper.kt
whyoleg: > SSLEngine directly from the channel without a buffer

Not really, as SSLEngine can say, that it needs more data to unwrap data, but buffer from channel f.e. is of some fixed size, while SSLEngine need bigger buffer (BUFFER_UNDERFLOW case)

PR #2932, opened 2022-03-31
KTOR-4091 Keep original content type when no registrations match in ContentNegotiation

rsinukov:


ktor-client/ktor-client-plugins/ktor-client-content-negotiation/common/src/io/ktor/client/plugins/contentnegotiation/ContentNegotiation.kt
```diff
@@ -93,15 +93,16 @@ public class ContentNegotiation internal constructor(
...
-                context.headers.remove(HttpHeaders.ContentType)
 
                 if (payload is Unit || payload is EmptyContent) {
+                    context.headers.remove(HttpHeaders.ContentType)
                     proceedWith(EmptyContent)
                     return@intercept
                 }
 
                 val matchingRegistrations = registrations.filter { it.contentTypeMatcher.contains(contentType) }
                     .takeIf { it.isNotEmpty() } ?: return@intercept
```
efemoney, line 104: Unrelated to this PR but this line can be simplified to `ifEmpty { return@intercept }`.

Just seems more readable that way.

e5l (approved):
lgtm

PR #2933, opened 2022-04-03
KTOR-4105 Fixed ContentType.parser crash in Darwin and Browserjs in some case

roths:
…ome case

**Subsystem**
ktor-client, related modules : ktor-client-core.jsMain, ktor-client-darwin.darwinMain

**Motivation**
https://youtrack.jetbrains.com/issue/KTOR-4105

**Solution**
split header value with ',' before apend them into HttpHeadersBuilder

### DIFF
--- a/ktor-client/ktor-client-darwin/darwin/src/io/ktor/client/engine/darwin/DarwinResponseReader.kt
+++ b/ktor-client/ktor-client-darwin/darwin/src/io/ktor/client/engine/darwin/DarwinResponseReader.kt
@@ -36,7 +36,11 @@ internal class DarwinResponseReader(
 
         val status = HttpStatusCode.fromValue(response.statusCode.toInt())
         val headers = buildHeaders {
-            headersDict.mapKeys { (key, value) -> append(key, value) }
+            headersDict.mapKeys { entry ->
+                entry.value.split(",").forEach{ valueItem ->
+                    append(entry.key, valueItem.trim())
+                }
+            }
         }
 
         val responseBody = GlobalScope.writer(callContext + Dispatchers.Unconfined, autoFlush = true) {
--- a/ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/ContentTest.kt
+++ b/ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/ContentTest.kt
@@ -291,6 +291,17 @@ class ContentTest : ClientLoader(5 * 60) {
         }
     }
 
+    @Test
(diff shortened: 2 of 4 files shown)


rsinukov (changes requested):
Thanks for you PR.
Can you please add a test for it inside `ktor-client-tests` module?

roths:
> Thanks for you PR. Can you please add a test for it inside `ktor-client-tests` module?

Already added

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/ContentTest.kt
```diff
@@ -291,6 +291,17 @@ class ContentTest : ClientLoader(5 * 60) {
         }
     }
 
+    @Test
+    fun testDuplicatedContentType() = clientTests(listOf("CIO", "Android", "Apache", "Curl", "Java", "Jetty", "Okhttp")) {
```
rsinukov, line 295: Is there a reason to ignore all these engines?

ktor-client/ktor-client-darwin/darwin/src/io/ktor/client/engine/darwin/DarwinResponseReader.kt
```diff
@@ -36,7 +36,11 @@ internal class DarwinResponseReader(
 
         val status = HttpStatusCode.fromValue(response.statusCode.toInt())
         val headers = buildHeaders {
-            headersDict.mapKeys { (key, value) -> append(key, value) }
+            headersDict.mapKeys { entry ->
+                entry.value.split(",").forEach{ valueItem ->
```
rsinukov, line 40: please fix codestyle here

rsinukov (changes requested):
Thanks! Couple of comments

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/ContentTest.kt
roths: because only browserjs and Darwin engine have this bug

rsinukov: But it's still good to verify this behaviour for other engines.

rsinukov:
@e5l is it obsolete now?

e5l:
Yep, it should be. Moreover, our parser can handle multiple content-type in one header line

rsinukov:
Closing this PR as obsolete.

PR #2934, opened 2022-04-04
Update kotlin to 1.6.20

e5l:
Fix [KTOR-4107](https://youtrack.jetbrains.com/issue/KTOR-4107)


gradle.properties
```diff
@@ -19,52 +19,16 @@ org.gradle.vfs.watch=true
...
-jwks_rsa_version=0.17.0
-tomcat_version=9.0.56
-
-# utility
-logback_version=1.2.10
-junit_version=4.13.2
-dokka_version=1.4.32
-nodejs_version=14.17.1
-npm_version=6.4.1
 slf4j_version=1.7.36
```
olme04, line 32: if those are used inside `buildSrc` or just in build script, they can be replaced with something like this, to avoid storing versions in `gradle.properties`:
```kotlin
val versionCatalog = extensions.getByType<VersionCatalogsExtension>().named("libs")
api(versionCatalog.findLibrary("slf4j-api").get()) //or implementation
```

also, it's possible to get version in the same way, look at: https://docs.gradle.org/current/javadoc/org/gradle/api/artifacts/VersionCatalog.html

build.gradle.kts
```diff
@@ -103,6 +90,8 @@ apply(from = "gradle/compatibility.gradle")
 
 plugins {
     id("org.jetbrains.dokka") version "1.6.10"
```
olme04, line 92: you can create `[plugins]` config in version catalog TOML file: 
```
[plugins]
jmh = { id = "me.champeau.jmh", version = "0.6.5" }
```
and use it like this: `alias(libs.plugins.jmh)`
more info: https://docs.gradle.org/current/userguide/platforms.html#sec:plugins

build.gradle.kts
```diff
@@ -49,19 +49,6 @@ buildscript {
...
-    val validator_version: String by extra
-
-    dependencies {
-        classpath("org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version")
-        classpath("org.jetbrains.kotlinx:atomicfu-gradle-plugin:$atomicfu_version")
-        classpath("org.jetbrains.kotlin:kotlin-serialization:$kotlin_version")
-        classpath("org.jetbrains.kotlinx:binary-compatibility-validator:$validator_version")
-    }
-
-    CacheRedirector.configureBuildScript(rootProject, this)
```
Ololoshechkin, line 64: Why don't we need atomicfu/serialization and stuff now?

buildSrc/src/main/kotlin/CacheRedirector.kt
```diff
@@ -1,136 +0,0 @@
...
-
-private val cacheRedirectorAvaialble: Boolean = isCacheRedirectorAvailable()
-
-/**
- * Enabled via environment variable, so that it can be reliably accessed from any piece of the build script,
- * including buildSrc within TeamCity CI.
- */
-private val cacheRedirectorEnabled =
-    cacheRedirectorAvaialble && (System.getenv("CACHE_REDIRECTOR_ENABLED")?.toBoolean() == true)
-
```
Ololoshechkin, line 19: Is this change connected with kotlin 1.6.20?

build.gradle.kts
e5l: Because we replaced it with version catalog

buildSrc/src/main/kotlin/CacheRedirector.kt
e5l: Nope, it's dropped unused code

olme04 (changes requested):

settings.gradle.kts
```diff
@@ -10,8 +10,8 @@ pluginManagement {
     }
     resolutionStrategy {
         eachPlugin {
-            if (requested.id.id == "kotlin2js") {
-                useModule("org.jetbrains.kotlin:kotlin-gradle-plugin:${requested.version}")
+            if (requested.id.id == "kotlinx-atomicfu") {
+                useModule("org.jetbrains.kotlinx:atomicfu-gradle-plugin:${requested.version}")
```
Ololoshechkin, line 14: Why is it connected with 1.6.20?

Ololoshechkin (approved):

build.gradle.kts
e5l: It also need changes in version injection for build train configuration, so we decided to postpone this

PR #2938, opened 2022-04-09
Ktor-3854 Fix: separate HttpCookie rendered cookie pairs with space

john5f35:
**Subsystem**
Client

**Motivation**
KTOR-3854

Cookie header rendered by `HttpCookies` should separate cookies with `"; "` rather than just `";"`.

**Solution**
Update `renderClientCookies()` function to join rendered cookies with `"; "`.


e5l (approved):
lgtm

PR #2939, opened 2022-04-09
Native Sockets TLS Client/Server support for linux

olme04:
Based on: https://github.com/ktorio/ktor/pull/2929

**Subsystem**
Network, CIO

**Motivation**
partially fixes https://youtrack.jetbrains.com/issue/KTOR-1181 (no darwin support yet)

**Solution**
* Use cinterop with openssl on linux to handle TLS
* Support both client/server authentication and verification
* Added support for multiplatform configuration of TLS via PKCS12 certificates
* Update JVM implementation of TLS to use SSLEngine with new API (server TLS, authentication for client and so on), but use old implementation if new API isn't used.

Note:
* I have no ability to make it working on darwin, as I have no mac, so implementation is fully in linux sourceset
* on Darwin, we can use SecureTransport api, which is out of the box coming in kotlin-native via Foundation. openssl isn't needed, which will simplify adopting it. SecureTransport has similar API to both openssl and JVM SSLEngine

TODO:
* support certificate generation via openssl - for now, I created 2 certificates with 1000 days validity for tests via `tls-certificates` module
* Implementations of both SSLEngine and via openssl is behaving differently when working with certificates - f.e. JVM don't check certificate validity and server name validity.
* May be we can introduce such an API with `ExperimentalTLSApi` annotation, because of possible bugs, and get more feedback on cases, where it can have unexpected behavior?
* More tests for TLS specifics is needed: different certificates (expired, wrong data), passwords validation, TLS session graceful closing
* For now, cinterop on linux points to `brew` location, because, at least on my machine, if to use same paths as in curl, it will fail too link binary because of GLIBC issues (related: https://youtrack.jetbrains.com/issue/KT-43501, https://youtrack.jetbrains.com/issue/KT-47061#focus=Comments-27-4947040.0-0)
* Need to restore binary compatibility on JVM - need to add some hidden declarations, from source perspective, changes are compatible


ktor-network/ktor-network-tls/build.gradle.kts
```diff
@@ -1,14 +1,23 @@
-kotlin.sourceSets {
-    jvmAndNixMain {
-        dependencies {
-            api(project(":ktor-network"))
-            api(project(":ktor-utils"))
+kotlin {
+    linuxX64 {
```
e5l, line 2: could we try adding other platofrms?

ktor-network/ktor-network-tls/darwin/src/io/ktor/network/tls/PKCS12Certificate.kt
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright 2014-2022 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.network.tls
+
+public class PKCS12Certificate(
```
e5l, line 7: kdocs are missing

ktor-network/ktor-network-tls/darwin/src/io/ktor/network/tls/PKCS12Certificate.kt
```diff
@@ -0,0 +1,10 @@
...
+    public val path: String,
+    public val password: (() -> CharArray)?
```
e5l, line 9: consider adding default if password really can be null

ktor-network/ktor-network-tls/darwin/src/io/ktor/network/tls/TLSConfigDarwin.kt
```diff
@@ -0,0 +1,28 @@
...
+import io.ktor.utils.io.core.*
+
+public actual class TLSConfig(
+    public actual val isClient: Boolean,
+    public actual val serverName: String?,
+    public actual val authentication: TLSAuthenticationConfig?,
+    public actual val verification: TLSVerificationConfig?,
+) : Closeable {
+    override fun close() {
+        //NOOP
```
e5l, line 17: obsolete

ktor-network/ktor-network-tls/jvmAndNix/src/io/ktor/network/tls/TLSConfig.kt
```diff
@@ -4,4 +4,21 @@
 
 package io.ktor.network.tls
 
-public expect class TLSConfig
+import io.ktor.utils.io.core.*
+
+//TODO: make it closeable
+public expect class TLSConfig: Closeable {
```
e5l, line 10: Could you tell me why the config is closeable?

ktor-network/ktor-network-tls/jvmAndNix/test/io/ktor/network/tls/TlsClientSocketTest.kt
```diff
@@ -0,0 +1,40 @@
...
+                    authentication({ "".toCharArray() }) {} //forces using of SSLEngine
+                }
+                .use { socket ->
+                    socket.openWriteChannel().run {
+                        writeStringUtf8("GET / HTTP/1.1\r\n")
+                        writeStringUtf8("Host: www.google.com\r\n")
+                        writeStringUtf8("Connection: close\r\n\r\n")
+                        flush()
+                    }
+                    println(socket.openReadChannel().readRemaining().readText())
```
e5l, line 36: please consider dropping println

Tiebe:
@olme04 are you still working on this?

whyoleg:
@Tiebe just in case, while using openssl here could be a good idea, the main problem with this approach is how to distribute it. I mean, we can do or dynamic or static linking to openssl or 2 separate artefacts for each, but it will be not so straightforward for end-users and library developers what to choose. Also there could be issues with how cinterop for `curl` (client engine) works with this cinterop for `openssl`. While this could be now implemented, there are a lot of questions and issues which should be done to make this production ready. As you see, there are a lot of other things in TODO column.

Tiebe:
@whyoleg Fair enough. Would just really love to have a websocket client with TLS support in Kotlin Native. Would you have any recommendations?

whyoleg:
@Tiebe I think, that for Apple platforms you can use `darwin` engine, for Windows - `winhttp`, for linux - there is really nothing now, because there are 2 ways: write TLS impl from scratch (you will still need cryptography support, from scratch or library, which have similar limitations) or use openssl/borinssl etc and leave with inconveniences :(
So, if you need linux support - you have no good ways for now

morki:
@whyoleg Maybe we can use [your multiplatform library](https://github.com/whyoleg/cryptography-kotlin) here, which solves choice of static/dynamic OpenSSL linking by artifact?

morki:
Thank you very much for your time answering this :) looking forward to some day it will maybe be implemented, cause i started creating multiplatform email / smtp lib, and tls sockets are the most common connection for this. But testing with pure sockets is the way forward for now :)

e5l:
Hey @olme04, @morki, thanks for the bump. The last time I investigated, it was tough to provide dependency for OpenSSL on the user side - in the application setup, the user had to specify the path to OpenSSL headers directly.

Could you tell me if you know about changes there? If we could avoid forcing users writing such config files, we can check this and deliver with 3.0.0

morki:
I am not an expert in the native field, i just use @whyoleg´s library and I think he solved it somehow, bacause the only thing I need as a consumer is choose openssl-shared or openssl-static dependency and everything is sorted out on all systems automatically

PR #2941, opened 2022-04-11
KTOR-4130 Only load default config if custom is not set in tests

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-4130

I am not sure how to add a test for it, since there is no side effect of loading a config, unless it's invalid and fails on crashing. But then we lose the ability to test proper default config, which is more important.


e5l (approved):
Good catch, lgtm

PR #2942, opened 2022-04-12
CIO server HTTPS support on Native

olme04:
Based on: https://github.com/ktorio/ktor/pull/2939

**Subsystem**
Server, CIO

**Motivation**
Fixes https://youtrack.jetbrains.com/issue/KTOR-694 for native

**Solution**
* New backward-compatible API for multiplatform SSL connectors (similar to TLSConfig from `ktor-network-tls`)
* CIO HTTPS support on native
* Make current JVM engines support both old and new configuration

TODO:
* looks like it will better to deprecate current APIs, and remove them with next MAJOR release
* Tests with client certificate verification is now disabled at all: looks like it's possible to create one more SSL connector in `EngineTestBase` to test also for it


ktor-server/ktor-server-cio/nix/src/io/ktor/server/cio/HttpServerSettingsNix.kt
```diff
@@ -4,13 +4,22 @@
...
-internal actual fun HttpServerSettings(
-    connectionIdleTimeoutSeconds: Long,
-    connectorConfig: EngineConnectorConfig
-): HttpServerSettings = HttpServerSettings(
-    host = connectorConfig.host,
-    port = connectorConfig.port,
-    connectionIdleTimeoutSeconds = connectionIdleTimeoutSeconds
-)
+internal actual fun TLSConfigBuilder.takeFromConnector(connectorConfig: EngineSSLConnectorConfig) {
+    connectorConfig.authentication?.let { config ->
```
e5l, line 11:
```
a?b.let { b -> b?.clet { c -> ... }
``` 
is hard to read. Could be replaced with:
```
val c = a?.b?.c
``` 
instead

ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/EngineConnectorConfigJvm.kt
```diff
@@ -23,22 +23,132 @@ public inline fun ApplicationEngineEnvironmentBuilder.sslConnector(
...
-    override var keyAlias: String,
-    override var keyStorePassword: () -> CharArray,
+public actual class EngineSSLConnectorBuilder actual constructor() :
+    EngineConnectorBuilder(ConnectorType.HTTPS), EngineSSLConnectorConfig {
+    public constructor(
+        keyStore: KeyStore,
+        keyAlias: String,
+        keyStorePassword: () -> CharArray,
+        privateKeyPassword: () -> CharArray
+    ) : this() {
```
e5l, line 33: we can make this constructor primary(with all `val` parameters) and make actual constructor secondary

PR #2943, opened 2022-04-13
KTOR-4157 CORS plugin should be route scoped

rsinukov:


ktor-server/ktor-server-tests/jvmAndNix/test/io/ktor/tests/server/plugins/CORSTest.kt
```diff
@@ -114,6 +114,58 @@ class CORSTest {
         }
     }
 
+    @Test
+    fun testSimpleRequestSubrouteInstall() {
+        withTestApplication {
```
e5l, line 119: Please consider using new `testApplication` API

ktor-server/ktor-server-plugins/ktor-server-cors-common/jvmAndNix/src/io/ktor/server/plugins/cors/CORSCommon.kt
```diff
@@ -0,0 +1,561 @@
...
+    val requestMethod = request.header(HttpHeaders.AccessControlRequestMethod)?.let { HttpMethod(it) }
+    return requestMethod != null && requestMethod in methods
+}
+
+private suspend fun ApplicationCall.respondCorsFailed() {
+    respond(HttpStatusCode.Forbidden)
+}
+
+private fun isValidOrigin(origin: String): Boolean {
+    if (origin.isEmpty()) {
```
e5l, line 503: `if (origin.isEmpty()) return`

e5l: and other lines

ktor-server/ktor-server-plugins/ktor-server-cors-common/jvmAndNix/src/io/ktor/server/plugins/cors/CORSCommon.kt
```diff
@@ -0,0 +1,561 @@
...
+     * doesn't handle method overridden by `X-Http-Method-Override`.
+     */
+    public fun allowMethod(method: HttpMethod) {
+        if (method !in CorsDefaultMethods) {
+            methods.add(method)
+        }
+    }
+}
+
+internal enum class OriginCheckResult {
```
e5l, line 358: consider extracting internal API to a separate file. It can make code easier to read

ktor-server/ktor-server-plugins/ktor-server-cors-common/jvmAndNix/src/io/ktor/server/plugins/cors/CORSCommon.kt
```diff
@@ -0,0 +1,561 @@
...
+            call.response.header(HttpHeaders.AccessControlExposeHeaders, exposedHeaders)
+        }
+    }
+}
+
+/**
+ * A configuration for the [CORS] plugin.
+ */
+@KtorDsl
+public class CORSConfig {
```
e5l, line 128: Consider extracting this to a separate file

ktor-server/ktor-server-plugins/ktor-server-cors-common/jvmAndNix/src/io/ktor/server/plugins/cors/CORSCommon.kt
```diff
@@ -0,0 +1,561 @@
...
+import io.ktor.http.*
+import io.ktor.server.application.*
+import io.ktor.server.plugins.*
+import io.ktor.server.request.*
+import io.ktor.server.response.*
+import io.ktor.server.routing.*
+import io.ktor.util.*
+import io.ktor.util.pipeline.*
+
+public fun PluginBuilder<CORSConfig>.buildCORSPlugin() {
```
e5l, line 16: Consider adding KDoc for public API

e5l (approved):
LGTM, please address minor comments

olme04:
Hey, what is the reason to have both route scoped and application plugin for the same feature, and in different modules?
As route scoped plugin can be installed in routing root, and as so do the same as application?
And also instead of 1 CORS dependency, there are 3 different now...

rsinukov:
@olme04 We can't change existing plugin now, because it would be a breaking change. But some of our customers need to install `CORS` into routing. 
We put it into a separate module because we don't want to publish 2.1.0 yet. The new module can be published as 2.0.0.

olme04:
ok, but then, you have already done breaking change:
1. `CORSConfig` package changed
2. binary compatibility is broken, as you moved several classes from one package to another

That's why Im asking

rsinukov:
You are right, I'll fix it. Thanks.

olme04:
BTW, Im still thinking, that adding just one public field `CORSRouteScoped` to current module in patch version is much better, than creating additional module, and somehow sharing INTERNAL implementation details between 2 modules.
Even if it doesn't match semver, it will be less intrusive change for all: users of CORS plugin and developers of ktor

rsinukov:
I opened PR #2945 with everything in one module

PR #2946, opened 2022-04-14
KTOR-3916 Fix combination of HttpCache and Logging plugins

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-3916


e5l (approved):
LGTM

PR #2954, opened 2022-04-19
KTOR-4163 Fix double header in ConditionalHeaders

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-4163


e5l (approved):
LGTM, docs should be adjusted as well

PR #2955, opened 2022-04-19
KTOR-4187 Select most specific exception in StatusPages

e5l:
Fix [KTOR-4187](https://youtrack.jetbrains.com/issue/KTOR-4187)


ktor-server/ktor-server-plugins/ktor-server-status-pages/jvmAndNix/src/io/ktor/server/plugins/statuspages/StatusPages.kt
```diff
@@ -29,7 +29,14 @@ public val StatusPages: ApplicationPlugin<StatusPagesConfig> = createApplication
     val statuses = HashMap(pluginConfig.statuses)
 
     fun findHandlerByValue(cause: Throwable): HandlerFunction? {
-        val key = exceptions.keys.find { cause.instanceOf(it) } ?: return null
+        val keys = exceptions.keys.filter { cause.instanceOf(it) }
+        if (keys.isEmpty()) return null
+
+        if (keys.isNotEmpty() && keys.size == 1) {
```
rsinukov, line 35: `keys.isNotEmpty()` is always false here

rsinukov (approved):

ktor-server/ktor-server-plugins/ktor-server-status-pages/jvmAndNix/src/io/ktor/server/plugins/statuspages/StatusPages.kt
e5l: thanks, will fix

PR #2958, opened 2022-04-20
KTOR-4166 Prevent raw web socket session from freeze

e5l:
Fix [KTOR-4166](https://youtrack.jetbrains.com/issue/KTOR-4166/Raw-Web-Socket-Connection-Suspending-Forever)


rsinukov (approved):

rsinukov:
Please check codestyle

olme04:
May be it will be good to make added test common, both for platforms (native/jvm) and engines (CIO, Netty, etc)?

e5l:
@olme04, raw session is supported by CIO only

PR #2959, opened 2022-04-20
KTOR-836 Fix Location href stripping trailing slash.

spand:
**Subsystem**
Locations feature/plugin

**Motivation**
Fixes KTOR-836

**Solution**
Restore trailing slash if it was present prior to all the `.split("/")` calls. It is kind of similar to how it always prefixes a "/" except that is always required.


rsinukov (approved):
LGTM, thanks for the PR

rsinukov (approved):

PR #2966, opened 2022-04-25
KTOR-4217 Fix sha1 argument

e5l:
Fix [KTOR-4217](https://youtrack.jetbrains.com/issue/KTOR-4217/%60io.ktor.util.sha1()%60-always-returns-da39a3ee5e6b4b0d3255bfef956)

### DIFF
--- a/ktor-utils/posix/src/io/ktor/util/CryptoNative.kt
+++ b/ktor-utils/posix/src/io/ktor/util/CryptoNative.kt
@@ -23,4 +23,4 @@ public actual fun Digest(name: String): Digest = error("[Digest] is not supporte
 /**
  * Compute SHA-1 hash for the specified [bytes]
  */
-public actual fun sha1(bytes: ByteArray): ByteArray = Sha1().digest()
+public actual fun sha1(bytes: ByteArray): ByteArray = Sha1().digest(bytes)
--- a/ktor-utils/common/test/io/ktor/util/HashFunctionTest.kt
+++ b/ktor-utils/common/test/io/ktor/util/HashFunctionTest.kt
@@ -4,6 +4,7 @@
 
 package io.ktor.util
 
+import kotlin.random.*
 import kotlin.test.*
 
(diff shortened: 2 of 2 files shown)


ktor-utils/posix/src/io/ktor/util/CryptoNative.kt
```diff
@@ -23,4 +23,6 @@ public actual fun Digest(name: String): Digest = error("[Digest] is not supporte
 /**
  * Compute SHA-1 hash for the specified [bytes]
  */
-public actual fun sha1(bytes: ByteArray): ByteArray = Sha1().digest()
+public actual fun sha1(bytes: ByteArray): ByteArray = Sha1().apply {
+    update(bytes)
+}.digest()
```
Lukellmann, line 28: I think there is already an extension called `digest` that takes `bytes` directly and does both the `update` and `digest` calls.

ktor-utils/posix/src/io/ktor/util/CryptoNative.kt
```diff
@@ -23,4 +23,6 @@ public actual fun Digest(name: String): Digest = error("[Digest] is not supporte
 /**
  * Compute SHA-1 hash for the specified [bytes]
  */
-public actual fun sha1(bytes: ByteArray): ByteArray = Sha1().digest()
+public actual fun sha1(bytes: ByteArray): ByteArray = Sha1().apply {
```
olme04, line 26: `digest(bytes)` can be used here https://github.com/ktorio/ktor/blob/b1ab7659065ee0013a57210df77027677489de4e/ktor-utils/common/src/io/ktor/util/HashFunction.kt#L12)

ktor-utils/posix/src/io/ktor/util/CryptoNative.kt
e5l: thanks, fixed

rsinukov (approved):

PR #2971, opened 2022-04-26
HTTP Parser Fixes

e5l:
Fix [KTOR-380](https://youtrack.jetbrains.com/issue/KTOR-380)
Fix [KTOR-382](https://youtrack.jetbrains.com/issue/KTOR-382)
Fix [KTOR-369](https://youtrack.jetbrains.com/issue/KTOR-369)

### DIFF
--- a/ktor-http/ktor-http-cio/common/test/io/ktor/tests/http/cio/HttpParserTestUtils.kt
+++ b/ktor-http/ktor-http-cio/common/test/io/ktor/tests/http/cio/HttpParserTestUtils.kt
@@ -1,0 +1,27 @@
+/*
+* Copyright 2014-2021 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+*/
+
+package io.ktor.tests.http.cio
+
+import kotlinx.coroutines.*
+import kotlinx.coroutines.intrinsics.*
+import kotlin.coroutines.*
+import kotlin.test.*
+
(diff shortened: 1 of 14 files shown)


ktor-http/ktor-http-cio/common/test/io/ktor/tests/http/cio/HttpParserTestUtils.kt
```diff
@@ -0,0 +1,27 @@
...
+
+package io.ktor.tests.http.cio
+
+import kotlinx.coroutines.*
+import kotlinx.coroutines.intrinsics.*
+import kotlin.coroutines.*
+import kotlin.test.*
+
+@OptIn(InternalCoroutinesApi::class)
+internal fun test(block: suspend () -> Unit) {
```
olme04, line 13: why not just use `testSuspend`?

rsinukov:
Please check CI

ktor-http/ktor-http-cio/common/test/io/ktor/tests/http/cio/HttpParserTestUtils.kt
olme04: IMO, it will be better to use `testSuspend`, to be more consistent. I think, that there is no reason to keep different suspend test runner functions all over the codebase.

e5l:
@rsinukov, sorry for the delay. Ready for review

PR #2978, opened 2022-04-28
Return correct route in Resources (KTOR-4239)

utybo:
**Subsystem**
Server (`ktor-server-resources`)

**Motivation**
This PR aims to solve KTOR-4239, see that ticket for more information.

**Solution**
Simply retrieve the result of the `method()` call and return it. There may be a more idiomatic way to do this, but I could not figure out a simpler solution.


rsinukov (approved):

rsinukov:
Thanks for the PR. LGTM

hfhbd:
BTW you could use contracts and `val`, but it is already merged.
```kt
@OptIn(ExperimentalContracts::class)
public inline fun <reified T : Any> Route.resource(noinline body: Route.() -> Unit): Route {
    contract {
        callsInPlace(body, InvocationKind.EXACTLY_ONCE)
    }
    val serializer = serializer<T>()
    return resource(serializer, body)
}
```

PR #2981, opened 2022-05-02
KTOR-4267 Fixed routing returns 405 even for not completely matched paths

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-4267


e5l (approved):
lgtm

PR #2984, opened 2022-05-05
KTOR-4280 Routes with tailcard should not count for specific http error codes

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-4280


e5l (approved):
lgtm

PR #2985, opened 2022-05-05
KTOR-4250 Fix missing routing params in auth interceptors

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-4250

This part of the app should be redesigned to remove circular reference between `ApplicationCall` and `AuthenticationContext`. But for now, to keep binary compatibility, we can only update reference every time context is accessed.


e5l (approved):
lgtm

PR #2986, opened 2022-05-05
KTOR-4281 Use only protocol from `DefaultRequest` for requests with host set

rsinukov:


ktor-client/ktor-client-core/common/test/DefaultRequestTest.kt
```diff
@@ -111,14 +112,34 @@ class DefaultRequestTest {
         val defaultUrl = Url(URLBuilder.Companion.origin)
         if (defaultUrl.port == defaultUrl.protocol.defaultPort) {
             assertEquals("https://localhost", client.get {}.bodyAsText())
-            assertEquals("ws://other.host:443/", client.get("ws://other.host/").bodyAsText())
```
e5l, line 114: could you tell me why do we drop these checks?

e5l, line 114: could you update these checks to present values instead of removing them?

e5l (approved):
lgtm

ktor-client/ktor-client-core/common/test/DefaultRequestTest.kt
rsinukov: It was not dropped, but moved outside the `if`.

PR #2991, opened 2022-05-07
KTOR-4267 Fixed routing returns 405 for not constant matches

rsinukov:


ktor-server/ktor-server-tests/jvmAndNix/test/io/ktor/tests/server/routing/RoutingProcessingTest.kt
```diff
@@ -887,6 +887,29 @@ class RoutingProcessingTest {
         }
     }
 
+    @Test
+    fun testRoutingSpecificErrorStatusCodeOnlyForConstantQualityPath() = testApplication {
```
e5l, line 891: Could you also add a test with a prefix(like `/api/v1/`)?

e5l (approved):
lgtm

ktor-server/ktor-server-tests/jvmAndNix/test/io/ktor/tests/server/routing/RoutingProcessingTest.kt
rsinukov: There are other tests in this class that test route with complex path responding with 405 status code

PR #2992, opened 2022-05-09
KTOR-4160 Fixed original exception is swallowed by "No request transformation found"

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-4160


ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/CallValidatorTest.kt
```diff
@@ -414,6 +413,58 @@ class CallValidatorTest {
             }
         }
     }
+
+    @Test
+    fun testThrowsOriginalExceptionWhenBodyIsNotSerialized(): Unit = testWithEngine(MockEngine) {
```
e5l, line 418: We should avoid explicit `Unit` in suspend MPP tests

e5l (approved):
LGTM, please drop Units

PR #2993, opened 2022-05-09
Tiny typo fix in documentation

sematnic:
The default for `useArrayPolymorphism` changed from `true` to `false` in version 2.0 but the documentation wasn't updated.

**Subsystem**

**Motivation**

**Solution**
Describe your solution.


rsinukov:
Good catch, thanks!

PR #2994, opened 2022-05-10
KTOR-3953 Encode slash in path parameter of UrlBuilder

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-3953

### DIFF
--- a/ktor-client/ktor-client-plugins/ktor-client-resources/common/test/ResourcesTest.kt
+++ b/ktor-client/ktor-client-plugins/ktor-client-resources/common/test/ResourcesTest.kt
@@ -125,4 +125,27 @@ class ResourcesTest {
             assertEquals(HttpStatusCode.OK, response.status)
         }
     }
+
+    @Serializable
+    @Resource("{path}")
+    class ParametersEncoded(val path: String, val query: String, val queryList: List<String>)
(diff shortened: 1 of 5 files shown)


ktor-server/ktor-server-tests/jvmAndNix/test/io/ktor/tests/server/http/URLBuilderTest.kt
```diff
@@ -90,12 +90,12 @@ class URLBuilderTest {
...
     @Test
     fun testPathComponentsWithTrailingSlashes() {
-        assertEquals("http://localhost/asd///", url { appendPathSegments("asd///") })
+        assertEquals("http://localhost/asd%2F%2F%2F", url { appendPathSegments("asd///") })
     }
 
     @Test
     fun testPathComponentsWithLeadingSlashes() {
-        assertEquals("http://localhost///asd", url { appendPathSegments("///asd") })
+        assertEquals("http://localhost/%2F%2F%2Fasd", url { appendPathSegments("///asd") })
```
e5l, line 98: could you test space with this change?

e5l (approved):
LGTM, but should be checked with space as well.

PR #2995, opened 2022-05-11
KTOR-4304 update API docs

andreyaksenov:


ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/cookies/HttpCookies.kt
```diff
@@ -15,11 +15,9 @@ import io.ktor.utils.io.core.*
 import kotlinx.coroutines.*
 
 /**
- * [HttpClient] plugin that handles sent `Cookie`, and received `Set-Cookie` headers,
- * using a specific [storage] for storing and retrieving cookies.
+ * A plugin that allows you to keep cookies between calls in an in-memory storage.
```
rsinukov, line 18: Storage is not necessary in-memory. Users can provide implementation for persistent `CookiesStorage`

rsinukov (approved):

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/cookies/HttpCookies.kt
andreyaksenov: thank you, will fix

PR #3000, opened 2022-05-15
KTOR-4323: WebSockets use custom serializer

hfhbd:
**Subsystem**
Client and Server, Websockets with Kotlinx serialization

**Motivation**
It is not possible to use WebSockets with a custom KSerializer

**Solution**
Added a KSerializer as a parameter


ktor-shared/ktor-serialization/common/src/SerializableWebSocketSession.kt
```diff
@@ -0,0 +1,14 @@
+/*
+ * Copyright 2014-2022 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.serialization
+
+import io.ktor.websocket.*
+
+public interface SerializableWebSocketSession : WebSocketSession {
```
rsinukov, line 9: the name is misleading. the session itself is not serializable, but has a serializer

ktor-server/ktor-server-plugins/ktor-server-websockets/jvmAndNix/test/io/ktor/tests/websocket/CIOWebSocketTest.kt
```diff
@@ -4,11 +4,52 @@
...
+import io.ktor.server.routing.*
+import io.ktor.server.websocket.*
+import io.ktor.util.*
+import io.ktor.utils.io.*
+import io.ktor.websocket.*
+import kotlinx.serialization.Serializer
+import kotlinx.serialization.json.*
+import kotlin.test.*
 
+@InternalAPI
```
rsinukov, line 19: why this annotation is needed here?

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/WebSocketTest.kt
```diff
@@ -142,6 +146,29 @@ class WebSocketTest : ClientLoader(100000) {
...
+            WebSockets {
+                contentConverter = KotlinxWebsocketSerializationConverter(Json)
+            }
+        }
+
+        test { client ->
+            client.webSocket("$TEST_WEBSOCKET_SERVER/websockets/echo") {
+                repeat(TEST_SIZE) {
+                    val originalData = Data("hello")
+                    sendSerialized(originalData, DataSerializer)
```
rsinukov, line 164: what is the use case of passing serializer here? can't it be registered in module or through `@UseSerializers` annotation?

rsinukov (changes requested):

ktor-server/ktor-server-plugins/ktor-server-websockets/jvmAndNix/test/io/ktor/tests/websocket/CIOWebSocketTest.kt
hfhbd: Because I use `writeFrame(Frame.Close())`, which I copied from other tests.

ktor-shared/ktor-serialization/common/src/SerializableWebSocketSession.kt
hfhbd: Yeah, that's true. I didn't found a better name through, maybe `WebSocketSessionWithSerializer`?

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/WebSocketTest.kt
hfhbd: I will add it in another PR and focus this PR on Websockets only.

whyoleg: BTW: have you tried to register external serializer for 3rd party class via `SerializersModule` via `contextual` call?. According to code it should just work by fetching serialiaer by `KClass`
Or can you at least provide an example of what you want to archive, and it fails

hfhbd: I want to archive this:
```
private val json = Json {
    SerializersModule {
        contextual(Data::class, DataSerializer)
    }
}

fun main() {
    val main = Data(42)
    val mainJson = json.encodeToString(main)
    println(mainJson)
}

data class Data(val value: Int)

@Serializer(forClass = Data::class)
object DataSerializer
```
And adding this to contextual does not work.

hfhbd: I deleted my old reproducer and at the moment reproducing the case does not work. But adding all possibilities via `contextual` or `polymorphic` to the serializer module is annoying, so passing the serializer directly at usage is nicer IMHO. And the implementation is quite forward for web sockets.

whyoleg (changes requested):
some more questions

ktor-shared/ktor-serialization/common/src/WebSocketSessionWithContentConverter.kt
```diff
@@ -0,0 +1,14 @@
+/*
+ * Copyright 2014-2022 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.serialization
+
+import io.ktor.websocket.*
+
+public interface WebSocketSessionWithContentConverter : WebSocketSession {
```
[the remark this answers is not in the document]
hfhbd: I agree, but the problem is the classpath. KSerializer is defined in kotlinx-serialiation, which is only added as dependency in ktor-shared/ktor-serialization-kotlinx. To create an extension function in the client web sockets it needs to resolve KSerializer. But the actual serialization lib (kotlinx.serialization) is an implementation detail, and not available in the core module (this is the same problem with `client.get { body(foo, FooSerializer)` etc.)

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/WebSocketTest.kt
whyoleg: I think that both ways are little annoying :)
But, even it's quite forward for websockets, looks reasonable to add the same to http endpoints, but there it will be not such easy.

And the main my concern about, why it's not good - calling this method will fail, if there is non kx.serialization converter.
Or f.e. if there will be some custom converter, which delegates to kotlinx.serialization converter, but uses another encoding for some types. So using `serializersModule` will work in both cases, but providing explicit serializer will work only in case of plain kx.serialization converter - such an inconsistency isn't so good in my opionion.

And in this specific use case, may be it will be much easier to not use this plugin at all, and do serialization handling manually, or write a simple plugin for your use case? :)

ktor-shared/ktor-serialization/common/src/WebSocketSessionWithContentConverter.kt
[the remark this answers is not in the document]
whyoleg: Ah, now I see.

Now I think, that may be some additional serialization plugin should be created for kotlinx.serialization specific usage, which will support all it's cool things, like explicit serializers provider, and reduced overhead of usage reflection over typeOf serialization retrieval. But not sure if it's really needed.

e5l:
The pr is hard to rebase, closed

PR #3001, opened 2022-05-17
KTOR-3900 Make Darwin engine work in runBlocking

e5l:
Fix [KTOR-3900](https://youtrack.jetbrains.com/issue/KTOR-3900/A-native-application-with-the-Darwin-engine-doesn-t-make-a-reque)


ktor-client/ktor-client-darwin/darwin/src/io/ktor/client/engine/darwin/DarwinClientEngine.kt
```diff
@@ -95,7 +94,7 @@ internal class DarwinClientEngine(override val config: DarwinClientEngineConfig)
         val session = NSURLSession.sessionWithConfiguration(
             configuration,
             responseReader,
-            delegateQueue = NSOperationQueue.currentQueue()
+            delegateQueue = requestQueue
```
rsinukov, line 97: should you do the same for WS requests?

rsinukov (approved):

PR #3005, opened 2022-05-17
KTOR-457 Support setting caching options on call

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-457

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-caching-headers/jvmAndNix/src/io/ktor/server/plugins/cachingheaders/CachingHeaders.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-caching-headers/jvmAndNix/src/io/ktor/server/plugins/cachingheaders/CachingHeaders.kt
@@ -18,6 +18,7 @@ public class CachingHeadersConfig {
     internal val optionsProviders = mutableListOf<(ApplicationCall, OutgoingContent) -> CachingOptions?>()
 
     init {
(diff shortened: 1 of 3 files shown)


ktor-server/ktor-server-plugins/ktor-server-caching-headers/jvmAndNix/src/io/ktor/server/plugins/cachingheaders/CachingHeaders.kt
```diff
@@ -79,3 +80,13 @@ public val CachingHeaders: RouteScopedPlugin<CachingHeadersConfig> = createRoute
         }
     }
 }
+
+/**
+ * Gets or sets the [CacheControl] for this call.
+ */
+public var ApplicationCall.caching: CachingOptions?
```
e5l, line 87: should we name it `cachingOptions`?

e5l (approved):
lgtm

ktor-server/ktor-server-plugins/ktor-server-caching-headers/jvmAndNix/src/io/ktor/server/plugins/cachingheaders/CachingHeaders.kt
rsinukov: I named it the same way as in `OutgoingContent` for consistency.

PR #3006, opened 2022-05-17
KTOR-4105 Consider only last content-type header

e5l:
Fix [KTOR-4105](https://youtrack.jetbrains.com/issue/KTOR-4105/Darwin-and-Kotlin-JS-List-has-more-than-one-element-error-when-h)


e5l:
> Why last but not the first? What's the logic behind picking last?

I checked several backends and noticed that in some cases after transforming the body the content-type header is appended in the end instead of replacing it

rsinukov (approved):

PR #3015, opened 2022-05-20
Jetty11

JonathanLennox:
**Subsystem**
This adds support for Jakarta EE servlets and Jetty 11, separately from the existing Java EE and Jetty 9 support. 

**Motivation**
As descripted in KTOR-1474, this adds support for Jakarta EE servlets, and Jetty 11 which depends on it.  (For trademark reasons, the name "Java EE" was changed to "Jakarta EE", and the Java namespaces of the packages changed with it.)

This makes it possible to use ktor in a project which uses Jetty 11.  (My specific motivating project was Jitsi's [jibri](https://github.com/jitsi/jibri).)

**Solution**
I clone the existing ktor-client-jetty, ktor-server-servlet, and ktor-servert-jetty directories, as ktor-client-jetty11, ktor-server-servlet5, and ktor-server-jetty11.

The client directory needed no code changes; the servlet directory needed only javax -> jakarta renames; and the jetty11 directory also needed some minor code changes to reflect other changes in Jetty between versions 9 and 11.

(It may be possible to avoid duplicating the code for ktor-client-jetty?  But I'm not enough of a Gradle expert to know how to do that.)

This should also make it possible to support Tomcat 10 but I haven't done that work.


marychatte:
Duplicated by [pull request #3339](https://github.com/ktorio/ktor/pull/3339)

PR #3016, opened 2022-05-22
Fixes for receiving and sending plain json string

rsinukov:
Default transform of string should happen after `ContentNegotiation`


e5l (approved):
lgtm

PR #3020, opened 2022-05-24
Little bit optimized code and refactoring code style

Tihon-Ustinov:
**Subsystem**
http

**Motivation**
I accidentally saw and decided that I could do better

**Solution**
I guess it's solve is better, because make array with nulls is bad practice and when was "find" for byValue is bad too

### DIFF
--- a/ktor-http/common/src/io/ktor/http/HttpStatusCode.kt
+++ b/ktor-http/common/src/io/ktor/http/HttpStatusCode.kt
@@ -4,8 +4,6 @@
 
 package io.ktor.http
 
-import kotlin.native.concurrent.*
-
 /**
  * Represents an HTTP status code and description.
  * @param value is a numeric code.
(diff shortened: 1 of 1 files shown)


ktor-http/common/src/io/ktor/http/HttpStatusCode.kt
```diff
@@ -103,21 +101,17 @@ public data class HttpStatusCode(val value: Int, val description: String) {
...
+        private val statusCodesMap: Map<Int, HttpStatusCode> = allStatusCodes().associateBy { it.value }
         /**
          * All known status codes
          */
-        public val allStatusCodes: List<HttpStatusCode> = allStatusCodes()
-
-        private val byValue: Array<HttpStatusCode?> = Array(1000) { idx ->
-            allStatusCodes.firstOrNull { it.value == idx }
-        }
+        public val allStatusCodes: List<HttpStatusCode> get() = statusCodesMap.values.toList()
```
rsinukov, line 108: Please remove `get()`. Otherwise, every call to this property will create a new list.

rsinukov (changes requested):
Thanks for the PR. One comment

ktor-http/common/src/io/ktor/http/HttpStatusCode.kt
Tihon-Ustinov: Done

PR #3023, opened 2022-05-25
KTOR-316 Add custom validation support for origin in CORS plugin

Ribesg:
**Subsystem**
`ktor-server-cors` plugin

**Motivation**
https://youtrack.jetbrains.com/issue/KTOR-316
Ktor Server has been lacking proper CORS support for years.
In our specific case, we validate CORS origins using a regex defined externally and injected into various API projects by the CI/CD. Note that these projects are developed using various languages and frameworks, hence why a single, simple regex was chosen to implement CORS check globally.

Until now, our way to implement this in our Ktor backends was using an altered copy of the CORS plugin made like 2 or 3 years ago, largely unmaintained. I'm in the process of upgrading a very old (more than 4 years) important Kotlin API project to the latest and greatest, including Ktor 2. During this process I checked on the linked issue above and realized it was improperly closed and still not implemented, and I actually could spent the time doing it.

**Solution**
My specific case is matching the origin against a single `Regex`, but I decided to make something a little more open and close to what is already available for headers as `config.headerPredicates` (populated using `config.allowHeaders(predicate: (String) -> Boolean)`).

With the proposed solution, the user can configure multiple predicates used to check the origin of the CORS request. If any matches, the origin is allowed. Basically what is done for headers.

Check the implemented tests to see the usage. I made a simple usage and one with a regex.

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-cors/jvmAndNix/src/io/ktor/server/plugins/cors/CORS.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-cors/jvmAndNix/src/io/ktor/server/plugins/cors/CORS.kt
@@ -43,6 +43,7 @@ internal fun PluginBuilder<CORSConfig>.buildPlugin() {
         (pluginConfig.headers + CORSConfig.CorsSimpleRequestHeaders).let { headers ->
             if (pluginConfig.allowNonSimpleContentTypes) headers else headers.minus(HttpHeaders.ContentType)
         }
+    val originPredicates: List<(String) -> Boolean> = pluginConfig.originPredicates
     val headerPredicates: List<(String) -> Boolean> = pluginConfig.headerPredicates
     val methods: Set<HttpMethod> = HashSet(pluginConfig.methods + CORSConfig.CorsDefaultMethods)
     val allHeadersSet: Set<String> = allHeaders.map { it.toLowerCasePreservingASCIIRules() }.toSet()
@@ -91,6 +92,7 @@ internal fun PluginBuilder<CORSConfig>.buildPlugin() {
             allowsAnyHost,
             hostsNormalized,
             hostsWithWildcard,
+            originPredicates,
             numberRegex
         )
         when (checkOrigin) {
@@ -153,6 +155,7 @@ private fun checkOrigin(
     allowsAnyHost: Boolean,
     hostsNormalized: Set<String>,
     hostsWithWildcard: Set<Pair<String, String>>,
+    originPredicates: List<(String) -> Boolean>,
     numberRegex: Regex
 ): OriginCheckResult =
     when {
@@ -163,6 +166,7 @@ private fun checkOrigin(
             allowsAnyHost,
             hostsNormalized,
             hostsWithWildcard,
+            originPredicates,
             numberRegex
         ) -> OriginCheckResult.Failed
         else -> OriginCheckResult.OK
@@ -251,12 +255,13 @@ private fun corsCheckOrigins(
     allowsAnyHost: Boolean,
     hostsNormalized: Set<String>,
     hostsWithWildcard: Set<Pair<String, String>>,
+    originPredicates: List<(String) -> Boolean>,
     numberRegex: Regex
 ): Boolean {
     val normalizedOrigin = normalizeOrigin(origin, numberRegex)
     return allowsAnyHost || normalizedOrigin in hostsNormalized || hostsWithWildcard.any { (prefix, suffix) ->
         normalizedOrigin.startsWith(prefix) && normalizedOrigin.endsWith(suffix)
-    }
+    } || originPredicates.any { it(origin) }
 }
 
 private fun corsCheckRequestHeaders(
--- a/ktor-server/ktor-server-plugins/ktor-server-cors/jvmAndNix/src/io/ktor/server/plugins/cors/CORSConfig.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-cors/jvmAndNix/src/io/ktor/server/plugins/cors/CORSConfig.kt
@@ -97,6 +97,11 @@ public class CORSConfig {
     public var allowCredentials: Boolean = false
 
     /**
+     * If present allows any origin matching any of the predicates.
+     */
+    internal val originPredicates: MutableList<(String) -> Boolean> = mutableListOf()
+
+    /**
      * If present represents the prefix for headers which are permitted in CORS requests.
      */
     public val headerPredicates: MutableList<(String) -> Boolean> = mutableListOf()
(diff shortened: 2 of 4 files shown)


ktor-server/ktor-server-plugins/ktor-server-cors/jvmAndNix/src/io/ktor/server/plugins/cors/CORSConfig.kt
```diff
@@ -96,6 +96,11 @@ public class CORSConfig {
      */
     public var allowCredentials: Boolean = false
 
+    /**
+     * If present allows any origin matching any of the predicates.
+     */
+    public val originPredicates: MutableList<(String) -> Boolean> = mutableListOf()
```
rsinukov, line 102: can be private

ktor-server/ktor-server-plugins/ktor-server-cors/jvmAndNix/src/io/ktor/server/plugins/cors/CORSConfig.kt
```diff
@@ -193,6 +198,13 @@ public class CORSConfig {
         allowHeader(HttpHeaders.XHttpMethodOverride)
     }
 
+    /**
+     * Allows using an origin matching [predicate] for the actual [CORS] request.
+     */
+    public fun allowOrigins(predicate: (String) -> Boolean) {
```
rsinukov, line 204: Can you rename it to `allowHosts`?. It's confusing to use different terminology for one thing, even if `origin` fits better here.

rsinukov (changes requested):
LGTM, but some comments. 
Also, since this PR contains adding new API, we can't merge it to patch release. Do you mind rebasing it into `2.1.0-eap` branch? And please run `./gradlew apiDump` and commit changes here.

ktor-server/ktor-server-plugins/ktor-server-cors/jvmAndNix/src/io/ktor/server/plugins/cors/CORSConfig.kt
Ribesg: I really mirrored what I saw for headers, do you see any reason why `headerPredicates` is public?

ktor-server/ktor-server-plugins/ktor-server-cors/jvmAndNix/src/io/ktor/server/plugins/cors/CORSConfig.kt
Ribesg: Actually I feel like in `https://example.org`, `origin` would be the entire thing while `host` would be just `example.org`, that's why I eventually renamed it to `allowOrigins` as I felt it was misleading to name it `allowHosts`

Ribesg: Initially, I wanted to add both `allowHosts` and `allowOrigins` functions but did not continue this way as `allowHosts` would be a little more complicated (it would have `host` but also `scheme` as parameters at least, maybe more) and I personally don't have a use case for it.

Ribesg:
> LGTM, but some comments. Also, since this PR contains adding new API, we can't merge it to patch release. Do you mind rebasing it into `2.1.0-eap` branch? And please run `./gradlew apiDump` and commit changes here.

Done! 👍

ktor-server/ktor-server-plugins/ktor-server-cors/jvmAndNix/src/io/ktor/server/plugins/cors/CORSConfig.kt
rsinukov: No, there is no reason, and it should be internal. I will clean up it in the next major release. We now try to avoid exposing mutable collections in public API.

ktor-server/ktor-server-plugins/ktor-server-cors/jvmAndNix/src/io/ktor/server/plugins/cors/CORSConfig.kt
rsinukov: Ok, this makes sense.

ktor-server/ktor-server-plugins/ktor-server-cors/jvmAndNix/src/io/ktor/server/plugins/cors/CORS.kt
```diff
@@ -251,12 +255,13 @@ private fun corsCheckOrigins(
...
     hostsNormalized: Set<String>,
     hostsWithWildcard: Set<Pair<String, String>>,
+    originPredicates: List<(String) -> Boolean>,
     numberRegex: Regex
 ): Boolean {
     val normalizedOrigin = normalizeOrigin(origin, numberRegex)
     return allowsAnyHost || normalizedOrigin in hostsNormalized || hostsWithWildcard.any { (prefix, suffix) ->
         normalizedOrigin.startsWith(prefix) && normalizedOrigin.endsWith(suffix)
-    }
+    } || originPredicates.any { it(origin) }
```
rsinukov, line 264: checkstyle fails here. `}` should be on a new line

Ribesg: Unless I'm mistaken, that's not what the build says https://ktor.teamcity.com/viewLog.html?buildId=150645&tab=Inspection&buildTypeId=Ktor_KtorCodeStyleVerifyKtLint

I fixed the "issue" reported there. I also strongly disagree with code style that reduces readability, but whatever. It should pass now.

ktor-server/ktor-server-plugins/ktor-server-cors/jvmAndNix/src/io/ktor/server/plugins/cors/CORSConfig.kt
Ribesg: Would you rather have `originPredicates` still be `public` here for coherence with the current major or make it `private` immediately to prevent any futur user migration?

rsinukov: I'll prefer it to be private (internal, actually)

Ribesg: Fixed. I also ran `./gradlew apiDump` again before committing.

rsinukov (approved):
Great job, thanks!

PR #3024, opened 2022-05-25
KTOR-4373 External services should use config from environment

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-4373


e5l (approved):
lgtm

PR #3029, opened 2022-05-27
Allow overriding HSTS settings per host

rescribet:
**Subsystem**
io.ktor:ktor-server-hsts

**Motivation**
When serving multiple sites though a single ktor instance, it can be useful to allow different HSTS configurations per site.

**Solution**
The current solution allows to set HSTS for specific hosts, falling back to the previous behaviour if no match was found.

Though some multi-tenant solutions might need dynamic configuration, the current solution simply implements a precomputed map to minimise runtime impact to an additional map lookup.


ktor-server/ktor-server-plugins/ktor-server-hsts/jvmAndNix/src/io/ktor/server/plugins/hsts/HSTS.kt
```diff
@@ -39,6 +40,18 @@ public class HSTSConfig {
...
+
+    /**
+     * @see [withHost]
+     */
+    internal val hostSpecific: MutableMap<String, HSTSConfig> = HashMap()
+
+    /**
+     * Set specific configuration for a [host].
+     */
+    public fun withHost(host: String, configure: HSTSConfig.() -> Unit) {
```
rsinukov, line 52: this allows writing code like 
```
install(HSTS) {
  withHost("a") { 
    withHost("b") { 
      ...
    }
  }
}
```
Can you fix this, please? Extracting base config should work.

ktor-server/ktor-server-plugins/ktor-server-hsts/jvmAndNix/src/io/ktor/server/plugins/hsts/HSTS.kt
```diff
@@ -81,9 +91,22 @@ public val HSTS: RouteScopedPlugin<HSTSConfig> = createRouteScopedPlugin("HSTS",
         }
     }
 
+    /**
+     * A constructed default `Strict-Transport-Security` header value.
+     */
+    val headerValue: String = constructHeaderValue(pluginConfig)
+
+    /**
+     * Constructed `Strict-Transport-Security` header value for given hosts
```
rsinukov, line 100: no need to document local values unless it's not clear what they do

rsinukov (changes requested):
Thanks for the PR!
Some comments. And since this PR has a new API, we can't merge it to a patch release. Please rebase to `2.1.0-eap` branch, run `./gradlew apiDump` and commit changes here.

ktor-server/ktor-server-plugins/ktor-server-hsts/jvmAndNix/src/io/ktor/server/plugins/hsts/HSTS.kt
rescribet: This has been changed and rebased.

Not sure what would be the most appropriate name for the base class, but though that `HSTSHostConfig` would be more indicative of function than something like`HSTSBaseConfig`

rescribet:
Rebased

rsinukov (approved):
LGTM!
Thanks, will merge after CI build

PR #3035, opened 2022-06-01
KTOR-4406 Simplify build configuration for IDE

e5l:


buildSrc/build.gradle.kts
```diff
@@ -29,6 +27,22 @@ dependencies {
 
     val ktlint_version = libs.versions.ktlint.version.get()
     implementation("org.jmailen.gradle:kotlinter-gradle:$ktlint_version")
+
+    implementation("io.ktor:ktor-server-default-headers:2.0.2")
```
rsinukov, line 31: is it possible/needed to update this version with every new release?

whyoleg: also, why not to use version catalog as with other dependencies in buildSrc build script?

e5l: We can put `*` here, but I'm not sure why do we need this

e5l: It will be actually good to migrate manually with IDE plugin

e5l: > why not to use version catalog as with other dependencies in buildSrc build script?

we could, but it would be listed as ktor_version and confused with actual project version

whyoleg: ktor_build_version or it can be named some else way :)

BTW IMO it will be better to extract this test server into additional composite build, and don't put it inside buildSrc

rsinukov (approved):

PR #3036, opened 2022-06-01
KTOR-4411 Fixed Resources plugin doesn't respect default values for Enum

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-4411


ktor-shared/ktor-resources/common/src/io/ktor/resources/serialization/Decoders.kt
```diff
@@ -28,8 +28,10 @@ internal class ParametersDecoder(
         while (parameterNames.hasNext()) {
             currentName = parameterNames.next()
             val elementIndex = descriptor.getElementIndex(currentName)
-            val isPrimitive = descriptor.getElementDescriptor(elementIndex).kind is PrimitiveKind
-            if (!isPrimitive || parameters.contains(currentName)) {
+            val elementDescriptorKind = descriptor.getElementDescriptor(elementIndex).kind
```
e5l, line 31: I would extract it to extension

e5l (approved):
lgtm

ktor-shared/ktor-resources/common/src/io/ktor/resources/serialization/Decoders.kt
rsinukov: that's the only place where it's used, so not much point in it

PR #3042, opened 2022-06-07
KTOR-620 Allow to set ContentType header

e5l:
Fix [KTOR-620](https://youtrack.jetbrains.com/issue/KTOR-620/Content-Type-in-a-request-headers)


rsinukov:
Related to #2884

rsinukov (approved):

PR #3043, opened 2022-06-07
KTOR-4423 Fix UDP response delay

Thomas-Vos:
https://youtrack.jetbrains.com/issue/KTOR-4423


ktor-network/jvm/src/io/ktor/network/sockets/JavaSocketAddress.kt
```diff
@@ -0,0 +1,115 @@
+/*
+ * Copyright 2014-2022 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.network.sockets
+
+import java.lang.reflect.*
+
+public actual sealed class SocketAddress {
+    internal abstract val address: java.net.SocketAddress
```
whyoleg, line 10: may be it's better to rename it to `jvmAddress` or `jdkAddress`?

ktor-network/jvm/src/io/ktor/network/sockets/JavaSocketAddress.kt
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    actual override fun hashCode(): Int {
+        return address.hashCode()
+    }
+
+    public actual override fun toString(): String = address.toString()
+
+    private companion object {
+        private fun checkSupportForUnixDomainSockets() {
```
whyoleg, line 105: may be it's better to create an underlying variable to init this at once + accessor which returns `Class` instance or throw an error ? this will also remove double call to it in `constructor` which receives `path: String` parameter

ktor-network/jvmAndNix/src/io/ktor/network/sockets/SocketAddress.kt
```diff
@@ -4,8 +4,31 @@
 
 package io.ktor.network.sockets
 
-public sealed class SocketAddress
+public expect sealed class SocketAddress
```
whyoleg, line 7: just for information: this change isn't source compatible for code written in `common` source set where exhaustive when is used

ktor-network/jvm/src/io/ktor/network/sockets/JavaSocketAddress.kt
whyoleg: and also, may be add import aliases to `java.net.*` classes?
(similarly to how kotlinx.datetime do it https://github.com/Kotlin/kotlinx-datetime/blob/master/core/jvm/src/Instant.kt#L22)

Thomas-Vos: Thanks for the review. The `address` name was suggested by @e5l.

ktor-network/jvm/src/io/ktor/network/sockets/JavaSocketAddress.kt
Thomas-Vos: Makes sense. I just moved the current implementation of this function and did not change the behaviour. I can change this.

ktor-network/jvmAndNix/src/io/ktor/network/sockets/SocketAddress.kt
Thomas-Vos: How exactly is this incompatible? The implementations of the sealed classes are also declared in the same file.

whyoleg: still, if `sealed class` is `expect` - it will be not exhausted in common code, as platforms can declare additional implementations, but common code doesn't know about them. 
So this code in common sourceset will not compile now after this changes:
```
val address: SocketAddress  = TODO()
val something = when(address) {
  is InetSocketAddress -> 1
  is UnixSocketAddress -> 2
}
```

whyoleg: https://kotlinlang.org/docs/sealed-classes.html#sealed-classes-and-when-expression
specifically info in green box :)

whyoleg: it can be used in kotlin code base in platform specific code, like jvm, or nix - no error will be there, the only issue is with common code

Thomas-Vos: Thanks for the links, I see. Initially I tried to declare only InetSocketAddress as an expect/actual but that fails with the screenshot above. Do you happen to have any suggestions how to get this working retaining compatibility?

whyoleg: as far as I know, it's not possible to do it without breaking of source or binary compatibility...
now I think it depends on ktor team, deciding if it's acceptable to not allow to do exhaustive when on this class in common code or not

Thomas-Vos: Yes I will await their response I guess. In my opinion a minor source breaking change is less important than the UDP issue that this solves on Windows. Anyway, thanks for the review, appreciate it.

whyoleg: Im always happy to help :)

ktor-network/jvm/src/io/ktor/network/sockets/JavaSocketAddress.kt
Thomas-Vos: improved it a bit

ktor-network/jvm/src/io/ktor/network/sockets/JavaSocketAddress.kt
e5l: I would rename the file to `SocketAddressJvm` to follow the general pattern and keep `address` as it is

ktor-network/jvm/src/io/ktor/network/sockets/JavaSocketAddress.kt
```diff
@@ -0,0 +1,118 @@
...
+
+public actual sealed class SocketAddress {
+    internal abstract val address: java.net.SocketAddress
+}
+
+public actual class InetSocketAddress internal constructor(
+    override val address: java.net.InetSocketAddress
+) : SocketAddress() {
+
+    // May trigger a name service reverse lookup when called.
```
e5l, line 17: Please consider writing a regular KDoc(`/** */`) with docs

ktor-network/jvmAndNix/src/io/ktor/network/sockets/SocketAddress.kt
e5l: For me, it looks like an MPP issue: all implementations are available at compile time. 

I also think it's ok to make this change. @rsinukov what do you think?

e5l (changes requested):

ktor-network/jvmAndNix/src/io/ktor/network/sockets/SocketAddress.kt
rsinukov: I agree.

ktor-network/jvm/src/io/ktor/network/sockets/JavaSocketAddress.kt
Thomas-Vos: @e5l thanks, fixed, also renamed the native file to keep it consistent.

ktor-network/jvm/src/io/ktor/network/sockets/JavaSocketAddress.kt
Thomas-Vos: @e5l added some documentation. What do you think about it?

ktor-network/jvmAndNix/src/io/ktor/network/sockets/SocketAddress.kt
Thomas-Vos: Okay thanks, keeping it as is.

Thomas-Vos:
Not sure why the Native on macOS test failed, it seems unrelated to this PR. Other platforms did succeed.

ktor-network/jvm/src/io/ktor/network/sockets/JavaSocketAddress.kt
e5l: @andreyaksenov, could you check?

andreyaksenov: @e5l looks good

e5l (approved):
LGTM! will merge after CI

e5l:
Thanks a lot,  good job! Merged.

ktor-network/jvmAndNix/src/io/ktor/network/sockets/SocketAddress.kt
ptitjes: > still, if `sealed class` is `expect` - when will be not exhausted in common code, as platforms can declare additional implementations, but common code doesn't know about them. So this code in common sourceset will not compile after this changes:
> 
> ```
> val address: SocketAddress  = TODO()
> val something = when(address) {
>   is InetSocketAddress -> 1
>   is UnixSocketAddress -> 2
> }
> ```

Just got bitten by that while bumping to Ktor 2.0.3... :D

PR #2592, opened 2021-09-05
KTOR-1460 Switch to org.w3c.fetch

hfhbd:
2nd PR after [KTOR-2230](https://youtrack.jetbrains.com/issue/KTOR-2230)

**Subsystem**
Client, JS

**Motivation**
[KTOR-1460](https://youtrack.jetbrains.com/issue/KTOR-1460)
- browser: Reduce code size by using std lib provided fetch api/classes
- nodeJS: added minimal defined classes manually

**Solution**
Switch to org.w3c.fetch or use manually defined classes on nodeJS

PR #3053, opened 2022-06-13
KTOR-326 Allow to set followRedirect property for js client engine

e5l:
Fix [KTOR-326](https://youtrack.jetbrains.com/issue/KTOR-326/JS-client-Cannot-change-redirect-policy-by-followRedirectsfalse)

### DIFF
--- a/ktor-client/ktor-client-core/js/src/io/ktor/client/engine/js/JsClientEngine.kt
+++ b/ktor-client/ktor-client-core/js/src/io/ktor/client/engine/js/JsClientEngine.kt
@@ -18,7 +18,9 @@ import org.w3c.dom.*
 import org.w3c.dom.events.*
 import kotlin.coroutines.*
 
-internal class JsClientEngine(override val config: HttpClientEngineConfig) : HttpClientEngineBase("ktor-js") {
+internal class JsClientEngine(
+    override val config: HttpClientEngineConfig
+) : HttpClientEngineBase("ktor-js") {
 
     override val dispatcher = Dispatchers.Default
 
@@ -31,13 +33,14 @@ internal class JsClientEngine(override val config: HttpClientEngineConfig) : Htt
     @OptIn(InternalAPI::class)
     override suspend fun execute(data: HttpRequestData): HttpResponseData {
         val callContext = callContext()
+        val clientConfig = data.attributes[CLIENT_CONFIG]
 
         if (data.isUpgradeRequest()) {
             return executeWebSocketRequest(data, callContext)
         }
 
         val requestTime = GMTDate()
-        val rawRequest = data.toRaw(callContext)
+        val rawRequest = data.toRaw(clientConfig, callContext)
         val rawResponse = commonFetch(data.url.toString(), rawRequest)
 
         val status = HttpStatusCode(rawResponse.status.toInt(), rawResponse.statusText)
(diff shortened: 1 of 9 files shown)


ktor-client/ktor-client-core/js/src/io/ktor/client/engine/js/JsClientEngine.kt
```diff
@@ -31,13 +33,14 @@ internal class JsClientEngine(override val config: HttpClientEngineConfig) : Htt
     @OptIn(InternalAPI::class)
     override suspend fun execute(data: HttpRequestData): HttpResponseData {
         val callContext = callContext()
+        val clientConfig = data.attributes[CLIENT_CONFIG]
```
rsinukov, line 36: why pass it through data? there is a property `config` in class

ktor-client/ktor-client-core/js/src/io/ktor/client/engine/js/JsClientEngine.kt
```diff
@@ -129,7 +134,7 @@ private fun Event.asString(): String = buildString {
     append(JSON.stringify(this@asString, arrayOf("message", "target", "type", "isTrusted")))
 }
 
-private fun io.ktor.client.fetch.Headers.mapToKtor(): Headers = buildHeaders {
+private fun org.w3c.fetch.Headers.mapToKtor(): Headers = buildHeaders {
```
hfhbd, line 137: Are you sure this works on nodeJS? See also https://github.com/ktorio/ktor/pull/2592

e5l: In this case, we replace some types, and the actual implementation stays the same.

e5l: The tests are also fine

hfhbd: Unit tests? AFAIK they does not work with nodeJS production, at least 1 year ago, which results into the same problem at runtime: https://github.com/ktorio/ktor/pull/2385 so you needed to create a test project and use the artifact.

ktor-client/ktor-client-core/js/src/io/ktor/client/engine/js/JsClientEngine.kt
e5l: cause we need client config, not engine

ktor-client/ktor-client-core/js/src/io/ktor/client/engine/js/JsClientEngine.kt
e5l: I will try to set up a sample

ktor-client/ktor-client-core/js/src/io/ktor/client/engine/js/JsClientEngine.kt
rsinukov: Ok. Let's create a ticket to pass client config in client factory `HttpClientEngineFactory.create` method in 3.0.0 for a proper fix.

rsinukov (approved):
LGTM, but please check sample with node

e5l:
Sure, https://youtrack.jetbrains.com/issue/KTOR-4484/Pass-Client-Config-to-HttpClientEngine

PR #3054, opened 2022-06-14
 KTOR-4390 Fix Query Parameters for UrlBuilder.set method

e5l:
Fix [KTOR-4390](https://youtrack.jetbrains.com/issue/KTOR-4390/CIO-Websockets-request-doesnt-include-query-parameters)

### DIFF
--- a/ktor-http/common/src/io/ktor/http/Url.kt
+++ b/ktor-http/common/src/io/ktor/http/Url.kt
@@ -55,14 +55,12 @@ public class Url internal constructor(
     }
 
     public val encodedQuery: String by lazy {
-        if (parameters.isEmpty()) {
-            return@lazy ""
-        }
         val queryStart = urlString.indexOf('?') + 1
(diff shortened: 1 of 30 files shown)


ktor-http/common/src/io/ktor/http/Url.kt
```diff
@@ -55,10 +55,9 @@ public class Url internal constructor(
     }
 
     public val encodedQuery: String by lazy {
-        if (parameters.isEmpty()) {
-            return@lazy ""
-        }
         val queryStart = urlString.indexOf('?') + 1
+        if (queryStart == 0) return@lazy ""
```
rsinukov, line 59: can you do the same change for fragment?

rsinukov (approved):

ktor-http/common/src/io/ktor/http/Url.kt
e5l: sure

PR #3063, opened 2022-06-21
KTOR-4495 Ignore Converter Failures for ContentNegotiation Plugin

e5l:
Fix [KTOR-4495](https://youtrack.jetbrains.com/issue/KTOR-4495/)


e5l:
Replaced with:
https://youtrack.jetbrains.com/issue/KTOR-4511/Ignore-ByteReadChannel-as-receive-type-in-ContentNegotiation
https://youtrack.jetbrains.com/issue/KTOR-4513/Serialization-Converters-Throws-When-It-Fails-to-Handle-Type-Instead-of-Returning-Null

PR #3065, opened 2022-06-21
KTOR-4475 Limit current running requests

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-4475


ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationEngine.kt
```diff
@@ -43,7 +43,7 @@ public class NettyApplicationEngine(
         /**
          * Number of concurrently running requests from the same http pipeline
          */
-        public var runningLimit: Int = 10
+        public var runningLimit: Int = 30
```
e5l, line 46: I would increase it at least to 32 or 64

e5l (approved):

PR #3067, opened 2022-06-22
KTOR-4511 Ignore ByteReadChannel for server response and client request

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-4511

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-content-negotiation/jvm/src/DefaultIgnoredTypesJvm.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-content-negotiation/jvm/src/DefaultIgnoredTypesJvm.kt
@@ -1,0 +1,13 @@
+/*
+ * Copyright 2014-2022 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.server.plugins.contentnegotiation
+
+import io.ktor.http.*
+import io.ktor.utils.io.*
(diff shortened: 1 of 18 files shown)


ktor-server/ktor-server-plugins/ktor-server-content-negotiation/jvm/src/DefaultIgnoredTypesJvm.kt
```diff
@@ -0,0 +1,13 @@
...
+
+package io.ktor.server.plugins.contentnegotiation
+
+import io.ktor.http.*
+import io.ktor.utils.io.*
+import java.io.*
+import kotlin.reflect.*
+
+internal actual val DefaultIgnoredTypes: Set<KClass<*>> =
+    mutableSetOf(HttpStatusCode::class, ByteReadChannel::class, InputStream::class)
```
e5l, line 13: Should we also ignore `ByteArray`, `ByteBuffer` and etc?

e5l (approved):
lgtm

ktor-server/ktor-server-plugins/ktor-server-content-negotiation/jvm/src/DefaultIgnoredTypesJvm.kt
rsinukov: In a server transforming to/from `ByteArray` happens before `ContentNegotiation`

PR #3085, opened 2022-06-28
KTOR-4572 Fixed concurrent requests in TestApplication

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-4572


e5l (approved):
Could you also add a test to make sure that requests are actually going in parallel? For instance we can send response body of the first request as the request body for the second.

PR #3086, opened 2022-06-28
KTOR-4593 Add support to automatically use the correct translation file for Pebble based on Accept Language header

TreyBastian:
**Subsystem**
Ktor-Server-Pebble

**Motivation**
This PR adds support for improving the ergonomics of using pebbles built in I18N support. It essentially makes it so you do not need to match Accept Headers to Locales in every route.

**Solution**
This uses a similar approach as the #1632 using `call.request.acceptLanguages` and validating it against a list of supported translations to either set the Locale or keep it `null` which is the default state.

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-pebble/jvm/src/io/ktor/server/pebble/Pebble.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-pebble/jvm/src/io/ktor/server/pebble/Pebble.kt
@@ -23,7 +23,7 @@ public class PebbleConfiguration : PebbleEngine.Builder() {
      * Allows you to define currently available language translations
      * must follow IETF BCP 47 language tag string specification
      */
-    public var availableLanguages: MutableList<String>? = null
+    public var availableLanguages: List<String>? = null
 }
 
 /**
@@ -50,16 +50,16 @@ public class PebbleContent(
  */
 public val Pebble: ApplicationPlugin<PebbleConfiguration> = createApplicationPlugin("Pebble", ::PebbleConfiguration) {
     val engine = pluginConfig.build()
+    val availableLanguages: List<String>? = pluginConfig.availableLanguages?.toList()
 
     fun process(content: PebbleContent, call: ApplicationCall): OutgoingContent = with(content) {
         val writer = StringWriter()
-        val availableLanguages: List<String>? = pluginConfig.availableLanguages?.toList()
         var locale = locale
 
         if (availableLanguages != null && locale == null) {
-            locale = call.request.acceptLanguageItems().firstOrNull {
-                pluginConfig.availableLanguages!!.contains(it.value)
-            }?.value?.let { Locale.forLanguageTag(it) }
+            locale = call.request.acceptLanguageItems()
+                .firstOrNull { pluginConfig.availableLanguages!!.contains(it.value) }
+                ?.value?.let { Locale.forLanguageTag(it) }
         }
         engine.getTemplate(content.template).evaluate(writer, model, locale)
 
--- a/ktor-server/ktor-server-plugins/ktor-server-pebble/jvm/src/io/ktor/server/pebble/Pebble.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-pebble/jvm/src/io/ktor/server/pebble/Pebble.kt
@@ -5,16 +5,18 @@
 package io.ktor.server.pebble
 
 import com.mitchellbosecke.pebble.*
-import com.mitchellbosecke.pebble.template.*
 import io.ktor.http.*
 import io.ktor.http.content.*
 import io.ktor.server.application.*
-import io.ktor.server.response.*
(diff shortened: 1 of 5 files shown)


TreyBastian:
What would the best way to update the docs for this be for once this is merged? There would be code snippets so would it be after a release?

ktor-server/ktor-server-plugins/ktor-server-pebble/jvm/src/io/ktor/server/pebble/Pebble.kt
```diff
@@ -5,17 +5,19 @@
...
+import io.ktor.server.request.*
 import io.ktor.util.*
-import io.ktor.util.cio.*
-import io.ktor.utils.io.*
 import java.io.*
 import java.util.*
 
+@KtorDsl
+public class PebbleConfiguration : PebbleEngine.Builder() {
+    public var availableLanguages: Array<String>? = null
```
rsinukov, line 18: Can you replace an array with a mutable list, please?

ktor-server/ktor-server-plugins/ktor-server-pebble/jvm/src/io/ktor/server/pebble/Pebble.kt
```diff
@@ -38,15 +40,21 @@ public class PebbleContent(
...
+    fun process(content: PebbleContent, call: ApplicationCall): OutgoingContent = with(content) {
         val writer = StringWriter()
-        engine.getTemplate(content.template).evaluate(writer, model, locale)
+
+        if (pluginConfig.availableLanguages != null && content.locale == null) {
+            val locale = call.request.acceptLanguageItems().firstOrNull {
+                pluginConfig.availableLanguages!!.contains(it.value)
+            }?.value?.let { Locale.forLanguageTag(it) }
+
+            engine.getTemplate(content.template).evaluate(writer, model, locale)
```
rsinukov, line 54: can you extract locale to `if/else` block and this line outside of it?

ktor-server/ktor-server-plugins/ktor-server-pebble/jvm/src/io/ktor/server/pebble/Pebble.kt
rsinukov: Please add an API doc

rsinukov (changes requested):
Thanks for the PR.
It's almost good to go, but minor comments. 
Regarding the docs, please create a ticket for this change in YouTrack, and we will make sure that it has proper docs after it's merged.

TreyBastian:
> 

Thanks for reviewing. I made the changes I'll submit the YouTrack ticket in the morning. Let me know if there's anything else that should change. I've tried to stick with conventions I've seen in other plugins.

TreyBastian:
https://youtrack.jetbrains.com/issue/KTOR-4593/Allow-Pebble-to-use-Accepted-Language-header-for-built-in-i18n-support

ktor-server/ktor-server-plugins/ktor-server-pebble/jvm/src/io/ktor/server/pebble/Pebble.kt
```diff
@@ -5,17 +5,27 @@
...
+ * Configuration for the [Pebble] plugin.
+ */
+@KtorDsl
+public class PebbleConfiguration : PebbleEngine.Builder() {
+
+    /**
+     * Allows you to define currently available language translations
+     * must follow IETF BCP 47 language tag string specification
+     */
+    public var availableLanguages: MutableList<String>? = null
```
rsinukov, line 26: Actually, it can be just `List` and not `MutableList`. Sorry =)

ktor-server/ktor-server-plugins/ktor-server-pebble/jvm/src/io/ktor/server/pebble/Pebble.kt
```diff
@@ -38,14 +48,19 @@ public class PebbleContent(
...
-    "Pebble",
-    { PebbleEngine.Builder() }
-) {
+public val Pebble: ApplicationPlugin<PebbleConfiguration> = createApplicationPlugin("Pebble", ::PebbleConfiguration) {
     val engine = pluginConfig.build()
 
-    fun process(content: PebbleContent): OutgoingContent = with(content) {
+    fun process(content: PebbleContent, call: ApplicationCall): OutgoingContent = with(content) {
         val writer = StringWriter()
+        val availableLanguages: List<String>? = pluginConfig.availableLanguages?.toList()
```
rsinukov, line 56: please extract it to `createApplicationPlugin` to save the reference at install time to ignore possible mutations of config later

rsinukov (changes requested):
Couple more changes.

TreyBastian:
Sorry for the delay there those changes are in.

rsinukov (approved):
LGTM!
Will merge after CI run

PR #3088, opened 2022-06-30
KTOR-4597 Fixed awaitSuspend wakes up early in closed ByteChannelSequential

rsinukov:
If reading happens between the channel was closed but before content was flushed, `awaitSuspend` wakes up to early.


e5l (approved):
lgtm

PR #3091, opened 2022-07-05
KTOR-4610 HttpCache plugin does not support max-stale directive

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-4610


ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/cache/HttpCache.kt
```diff
@@ -83,35 +83,54 @@ public class HttpCache private constructor(
...
                 val cache = plugin.findResponse(context, content)
-                if (cache != null && !cache.shouldValidate()) {
-                    finish()
-                    val call = cache.produceResponse().call
-
-                    scope.monitor.raise(HttpResponseFromCache, call.response)
-                    proceedWith(call)
-
+                if (cache == null) {
+                    val header = parseHeaderValue(context.headers[HttpHeaders.CacheControl])
```
e5l, line 87: Could you extract this to a separate method with a valuable name?

e5l (approved):
lgtm

PR #3093, opened 2022-07-07
PoC New API for Client Plugins

rsinukov:


e5l:
Could you name the branch with `eap` suffix? I would like to try it out as dependency

rsinukov:
reopened in #3094

PR #3094, opened 2022-07-08
Client plugins API 

rsinukov:


e5l:
I played a lot and it is great.

There are 2 points I found:
- I missed the simple way of creating `HttpClientCall` and `HttpResponse` from `HttpRequestBuilder` without making a network call.

- I would also replace `HttpClientCall` with `HttpResponse` in the public API.

rsinukov:
@e5l Nice! Can you add these points to the design ticket please, so that we have discussion in one place?

e5l:
@rsinukov Sure. Done!

PR #3097, opened 2022-07-11
KTOR-2128 add extra parameters property for auth0 authorisation and receiving token

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-2128


ktor-server/ktor-server-plugins/ktor-server-auth/jvm/src/io/ktor/server/auth/OAuth.kt
```diff
@@ -80,8 +82,42 @@ public sealed class OAuthServerSettings(public val name: String, public val vers
 
         public val authorizeUrlInterceptor: URLBuilder.() -> Unit = {},
         public val passParamsInURL: Boolean = false,
+        public val extraAuthParameters: List<Pair<String, String>> = emptyList(),
```
e5l, line 85: Maybe we should make some DSL for OAuth config

e5l (approved):
lgtm

PR #3101, opened 2022-07-13
KTOR-4637 Override isOneShot for streaming body types in OkHttp

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-4637


e5l (approved):
lgtm

PR #3105, opened 2022-07-18
KTOR-4193 Log CallLogging MDC values in StatusPages exception handlers.

luke-stead-sonocent:
**Subsystem**
Server Call Logging & Server Status Pages

**Motivation**
https://youtrack.jetbrains.com/issue/KTOR-4193/CallLogging-configured-MDC-entries-are-not-passed-to-StatusPages-exception-handlers

**Solution**
Fix `mdcProvider`, and use it to wrap the exception handlers in the mdc block.
I'm open to other solutions, may be possible to change CallLogging to not remove MDC entries once the try/finally has finished, but instead to remove them in a dedicated phase after fallback, but that feels more brittle.
Perhaps KtorMDCProvider should use a AttributeKey accessible in core, instead of the current "Look through all the plugins for the one we want" method.

### DIFF
--- a/ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/logging/Logging.kt
+++ b/ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/logging/Logging.kt
@@ -34,5 +34,5 @@ private object EmptyMDCProvider : MDCProvider {
 public val Application.mdcProvider: MDCProvider
     @Suppress("UNCHECKED_CAST")
     get() = pluginRegistry.allKeys
-        .firstNotNullOfOrNull { attributes.getOrNull(it as AttributeKey<Any>) as? MDCProvider }
+        .firstNotNullOfOrNull { pluginRegistry.getOrNull(it as AttributeKey<Any>) as? MDCProvider }
         ?: EmptyMDCProvider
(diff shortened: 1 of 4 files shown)


LukasForst:
Thanks @luke-stead-sonocent for doing this.

@rsinukov can you look at this PR as well please? It's especially inconvenient as for example we use MDC to store tracking information (like call ID and other properties) and when the exception is caught and logged, they're not present in the logs thanks to this.

ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/logging/Logging.kt
```diff
@@ -34,5 +34,5 @@ private object EmptyMDCProvider : MDCProvider {
 public val Application.mdcProvider: MDCProvider
     @Suppress("UNCHECKED_CAST")
     get() = pluginRegistry.allKeys
-        .firstNotNullOfOrNull { attributes.getOrNull(it as AttributeKey<Any>) as? MDCProvider }
+        .firstNotNullOfOrNull { pluginRegistry.getOrNull(it as AttributeKey<Any>) as? MDCProvider }
```
rsinukov, line 37: why is this change needed?

rsinukov: oh, I see. thanks for fixing it

rsinukov (approved):
LGTM. Will merge after CI run

PR #3106, opened 2022-07-19
KTOR-4658 Fixed Client resources plugin miss builders for PATCH method

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-4658


e5l (approved):
Could we track this methods in the API dump as well?

rsinukov:
No, because they are inline with reified parameter

PR #3109, opened 2022-07-20
KTOR-735 Support WinHTTP engine

### LINKED ISSUE #671: Support native windows HTTP client

dtretyakov:
This change brings support of new Windows engine built on top of WinHTTP library. It supports HTTP 1.x/2 and WebSocket protocols.

**Subsystem**
Client, native windows engine.

**Motivation**
Fixes #671 WinHTTP API allows using native HTTP client without need to include additional DLLs. Also, it allows using certificate stored in Windows Certificates store.

**Solution**

We’re using WinHTTP in asynchronous working mode. In such mode notifications about operation results are delivered via status callback. The library supports both HTTP request processing as well as WebSockets protocol.

In HTTP engine we are using the following abstractions:
* `WinHttpSession` - create once per HTTP engine and allows configuring common settings like timeouts.
* `WinHttpConnect` - represents connection to HTTP host. In our case it also stores list of callbacks for coroutine continuations.
* `WinHttpRequest` - allows writing/reading request and response messages.
* `WinHttpRequestProducer` - handles request body writing. Allows using chunked body encoding.
* `WinHttpWebSocket` - represents WebSocket session. It's created from WinHttpRequest by closing original request and keeping `WinHttpConnect`ion open.

WinHTTP allows specifying connect and socket (send/receive) timeouts. Request timeout is handled by corresponding Ktor plugin.

**Notes**:
WinHTTP is strict regarding `Content-Length` header validation in HTTP 1.x requests and requires using `Transfer-Encoding: chunked` request body if body length is unknown, e.g. when it's stream or channel based. In such cases we’re trying to use system encoding (modern Windows 10+ OS) or we're using fallback to manual chunks writing.

In case of failures we're converting error codes into human readable messages by using `FormatMessage` function.

### DIFF
--- a/ktor-client/ktor-client-winhttp/build.gradle.kts
+++ b/ktor-client/ktor-client-winhttp/build.gradle.kts
@@ -1,0 +1,38 @@
+import org.jetbrains.kotlin.gradle.targets.native.tasks.*
+
+apply<test.server.TestServerPlugin>()
+
+val WIN_LIBRARY_PATH = "c:\\msys64\\mingw64\\bin;c:\\tools\\msys64\\mingw64\\bin;C:\\Tools\\msys2\\mingw64\\bin"
+
+plugins {
+    id("kotlinx-serialization")
+}
+
+kotlin {
+    fastTarget()
+
+    createCInterop("winhttp", windowsTargets()) {
+        defFile = File(projectDir, "windows/interop/winhttp.def")
+    }
+
+    sourceSets {
+        windowsMain {
+            dependencies {
+                api(project(":ktor-client:ktor-client-core"))
+                api(project(":ktor-http:ktor-http-cio"))
+            }
+        }
+        windowsTest {
+            dependencies {
+                api(project(":ktor-client:ktor-client-plugins:ktor-client-logging"))
+                api(project(":ktor-client:ktor-client-plugins:ktor-client-json"))
+            }
+        }
+    }
+
+    afterEvaluate {
+        if (HOST_NAME != "windows") return@afterEvaluate
+        val winTests = tasks.findByName("mingwX64Test") as? KotlinNativeTest? ?: return@afterEvaluate
+        winTests.environment("PATH", WIN_LIBRARY_PATH)
+    }
+}
--- a/ktor-client/ktor-client-winhttp/windows/interop/winhttp.def
+++ b/ktor-client/ktor-client-winhttp/windows/interop/winhttp.def
@@ -1,27 +1,12 @@
-package = platform.windows
-headers = minwindef.h minwinbase.h winhttp.h
-headerFilter = *
+package = ktor.cinterop.winhttp
+headers = wtypes.h winhttp.h
+headerFilter = winhttp.h
 
 compilerOpts = -DUNICODE
 linkerOpts = -lwinhttp
-
 ---
-static unsigned int GetHResultFromError(unsigned int errorCode) {
-    int result;
-    if (errorCode < 0) {
-        result = errorCode;
-    }
-    else {
-        result = (int)(((unsigned int)errorCode & 0x0000FFFF) | (7 << 16) | 0x80000000);
-    }
-
-    return result;
-}
-
-static unsigned int MakeLanguageId(unsigned int primaryLanguageId, unsigned int subLanguageId) {
-    return (((subLanguageId) << 10) | (primaryLanguageId));
-}
-
+// These WebSocket declarations are not present in our sysroots,
+// so we add them here.
 #define WINHTTP_OPTION_UPGRADE_TO_WEB_SOCKET            114
 
 typedef enum _WINHTTP_WEB_SOCKET_OPERATION
--- a/ktor-client/ktor-client-winhttp/windows/interop/winhttp.def
+++ b/ktor-client/ktor-client-winhttp/windows/interop/winhttp.def
@@ -1,0 +1,142 @@
+package = platform.windows
+headers = minwindef.h minwinbase.h winhttp.h
+headerFilter = *
+
+compilerOpts = -DUNICODE
+linkerOpts = -lwinhttp
+
+---
+static unsigned int GetHResultFromError(unsigned int errorCode) {
+    int result;
+    if (errorCode < 0) {
+        result = errorCode;
+    }
+    else {
+        result = (int)(((unsigned int)errorCode & 0x0000FFFF) | (7 << 16) | 0x80000000);
+    }
+
+    return result;
+}
+
+static unsigned int MakeLanguageId(unsigned int primaryLanguageId, unsigned int subLanguageId) {
+    return (((subLanguageId) << 10) | (primaryLanguageId));
+}
+
+#define WINHTTP_OPTION_UPGRADE_TO_WEB_SOCKET            114
+
+typedef enum _WINHTTP_WEB_SOCKET_OPERATION
+{
+    WINHTTP_WEB_SOCKET_SEND_OPERATION                   = 0,
+    WINHTTP_WEB_SOCKET_RECEIVE_OPERATION                = 1,
+    WINHTTP_WEB_SOCKET_CLOSE_OPERATION                  = 2,
+    WINHTTP_WEB_SOCKET_SHUTDOWN_OPERATION               = 3
+} WINHTTP_WEB_SOCKET_OPERATION;
+
+typedef enum _WINHTTP_WEB_SOCKET_BUFFER_TYPE
+{
+    WINHTTP_WEB_SOCKET_BINARY_MESSAGE_BUFFER_TYPE       = 0,
+    WINHTTP_WEB_SOCKET_BINARY_FRAGMENT_BUFFER_TYPE      = 1,
+    WINHTTP_WEB_SOCKET_UTF8_MESSAGE_BUFFER_TYPE         = 2,
+    WINHTTP_WEB_SOCKET_UTF8_FRAGMENT_BUFFER_TYPE        = 3,
+    WINHTTP_WEB_SOCKET_CLOSE_BUFFER_TYPE                = 4
+} WINHTTP_WEB_SOCKET_BUFFER_TYPE;
+
+typedef enum _WINHTTP_WEB_SOCKET_CLOSE_STATUS
+{
+    WINHTTP_WEB_SOCKET_SUCCESS_CLOSE_STATUS                = 1000,
+    WINHTTP_WEB_SOCKET_ENDPOINT_TERMINATED_CLOSE_STATUS    = 1001,
+    WINHTTP_WEB_SOCKET_PROTOCOL_ERROR_CLOSE_STATUS         = 1002,
+    WINHTTP_WEB_SOCKET_INVALID_DATA_TYPE_CLOSE_STATUS      = 1003,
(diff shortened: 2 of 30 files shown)


dtretyakov:
At first we need to check tests in CI build.

ktor-client/ktor-client-winhttp/windows/src/io/ktor/client/engine/winhttp/WinHttpClientEngine.kt
```diff
@@ -0,0 +1,44 @@
...
+            session.executeRequest(data, callContext)
+        } catch (cause: Throwable) {
+            callContext.cancel(CancellationException("Failed to execute request", cause))
+            throw cause
+        }
+    }
+}
+
+@Suppress("KDocMissingDocumentation")
+public class WinHttpIllegalStateException(cause: String) : IllegalStateException(cause)
```
whyoleg, line 44: why it's needed?

ktor-client/ktor-client-winhttp/windows/interop/winhttp.def
```diff
@@ -0,0 +1,142 @@
+package = platform.windows
```
whyoleg, line 1: is it ok to use `platform` package here?

ktor-client/ktor-client-winhttp/windows/src/io/ktor/client/engine/winhttp/internal/WinHttpRequest.kt
```diff
@@ -0,0 +1,366 @@
...
+        }
+    }
+
+    /**
+     * Enables requested HTTP protocols.
+     *
+     * @param protocolVersion is required protocol version.
+     */
+    private fun enableHttpProtocols(protocolVersion: HttpProtocolVersion) = memScoped {
+        println("request: enableHttpProtocols $protocolVersion")
```
whyoleg, line 257: please remove all printlns

gradle.properties
```diff
@@ -4,7 +4,7 @@
 
 # ktor
 # Makes Ktor import and resolve
-ktor.ide.jvmAndCommonOnly=true
+ktor.ide.jvmAndCommonOnly=false
```
whyoleg, line 7: please revert

whyoleg (commented):
low signal review, will do more in depth review later
Overall, looks cool! even WebSockets are supported!

ktor-client/ktor-client-winhttp/windows/src/io/ktor/client/engine/winhttp/WinHttpClientEngine.kt
dtretyakov: It's used in the same cases as a similar `CurlIllegalStateException` in curl engine, e.g. https://github.com/ktorio/ktor/search?q=CurlIllegalStateException

ktor-client/ktor-client-winhttp/windows/interop/winhttp.def
dtretyakov: Naturally WinHTTP is a part of Windows platform library. But kotlin native team cautiously adds new parts in it: https://youtrack.jetbrains.com/issue/KT-39747

ktor-client/ktor-client-winhttp/windows/src/io/ktor/client/engine/winhttp/WinHttpClientEngineConfig.kt
```diff
@@ -0,0 +1,28 @@
...
+    /**
+     * A value that allows to set required security protocol versions.
+     * By default will be used system setting.
+     */
+    public var securityProtocols: WinHttpSecurityProtocol = WinHttpSecurityProtocol.Default
+
+    /**
+     * A value that disables TLS verification for outgoing requests.
+     */
+    public var sslVerify: Boolean = true
```
dtretyakov, line 27: It was used only for tests to check connection to web sites with self signed certificates. Could we expose it for end users?

ktor-client/ktor-client-winhttp/windows/src/io/ktor/client/engine/winhttp/internal/WinHttpExceptions.kt
```diff
@@ -0,0 +1,127 @@
...
+
+/**
+ * Creates an exception from WinAPI error code.
+ */
+internal fun getWinHttpException(message: String, errorCode: UInt): Exception {
+    val hResult = GetHResultFromError(errorCode)
+    val errorMessage = getErrorMessage(errorCode).trimEnd('.')
+    val cause = "$message: $errorMessage. Error $errorCode (0x${hResult.toString(16)})"
+
+    return if (errorCode == ERROR_WINHTTP_TIMEOUT) {
```
dtretyakov, line 36: I failed to find a way to distinguish which timeout may trigger this error during runtime and using `ConnectTimeoutException`. But maybe we need to use another exception class.

ktor-client/ktor-client-winhttp/windows/interop/winhttp.def
whyoleg: The problem will arise, if they decide to add it someday
I would prefer to use some other package to prevent future issues in such case

sbogolepov: Hi. I suggest something like this: https://gist.github.com/sbogolepov/276ce9218e6939165362af251fe4301c. The exact package name is not important, but it shouldn't be `platform.windows`.
1. `platform.windows` is enormous which is bad for many things, including IDE experience. We can't split it easily because of backward API&ABI compatibility, but at least we can stop adding new declarations to it.
2. `winhttp.h` and `wininet.h` have constants with the same name but different value (`INTERNET_SCHEME_HTTP `). Putting them into one package will result into shadowing of one of them.  
3. `GetHResultFromError` and `MakeLanguageId` can be implemented in Kotlin, thus keeping interop package clean.
My suggestion is very close to something that I'm going to implement in 1.8.0. So after its release, it will be possible to switch to `platform.winhttp` without major changes.

ktor-client/ktor-client-winhttp/windows/src/io/ktor/client/engine/winhttp/WinHttpClientEngineConfig.kt
ArcticLampyrid: Why not? [CURL exposes `sslVerify` too.](https://api.ktor.io/ktor-client/ktor-client-curl/io.ktor.client.engine.curl/-curl-client-engine-config/ssl-verify.html) Users may use it for self-host services.

ktor-client/ktor-client-winhttp/windows/src/io/ktor/client/engine/winhttp/WinHttpClientEngine.kt
ArcticLampyrid: Prefer delicated exceptions. `CurlIllegalStateException` is deprecated (though there are no alternatives available)

ktor-client/ktor-client-winhttp/windows/src/io/ktor/client/engine/winhttp/internal/WinHttpExceptions.kt
ArcticLampyrid: It did actually the same things as Curl.
https://github.com/ktorio/ktor/blob/24c5d4654f6bdb68b28f9d4515953ef8b9d0b24b/ktor-client/ktor-client-curl/desktop/src/io/ktor/client/engine/curl/internal/CurlMultiApiHandler.kt#L293

ArcticLampyrid:
Any progress?

Zxilly:
Any update on this?

e5l (changes requested):
LGTM in general! Please address minor notes, check CI and rebase on 2.2.0-eap branch. We're going to merge and release this PR with 2.2.0

rsinukov:
Please reopen it into `main` branch.

Zxilly:
Any further update on this?

ktor-client/ktor-client-winhttp/build.gradle.kts
```diff
@@ -0,0 +1,38 @@
...
+        }
+        windowsTest {
+            dependencies {
+                api(project(":ktor-client:ktor-client-plugins:ktor-client-logging"))
+                api(project(":ktor-client:ktor-client-plugins:ktor-client-json"))
+            }
+        }
+    }
+
+    afterEvaluate {
```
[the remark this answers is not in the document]
dtretyakov: It was copied from the `ktor-client-curl`: https://github.com/ktorio/ktor/blob/main/ktor-client/ktor-client-curl/build.gradle.kts#L64

ktor-client/ktor-client-winhttp/windows/src/io/ktor/client/engine/winhttp/internal/WinHttpCallback.kt
```diff
@@ -0,0 +1,41 @@
...
+
+internal fun winHttpCallback(
+    @Suppress("UNUSED_PARAMETER")
+    hInternet: HINTERNET?,
+    dwContext: DWORD_PTR,
+    dwStatus: DWORD,
+    statusInfo: LPVOID?,
+    statusInfoLength: DWORD
+) {
+    val contextPtr = dwContext.toLong().toCPointer<COpaque>() ?: run {
```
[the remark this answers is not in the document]
dtretyakov: I suspect in this `run` block was debug logging. I'll eliminate it.

ktor-client/ktor-client-winhttp/windows/src/io/ktor/client/engine/winhttp/internal/WinHttpCallback.kt
```diff
@@ -0,0 +1,41 @@
...
+        return
+    }
+
+    val connect = contextPtr.asStableRef<WinHttpConnect>().get()
+    if (connect.isClosed) {
```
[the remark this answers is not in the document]
dtretyakov: We're creating this stable ref in the `WinHttpRequest` constructor: https://github.com/dtretyakov/ktor/blob/client/winhttp/ktor-client/ktor-client-winhttp/windows/src/io/ktor/client/engine/winhttp/internal/WinHttpRequest.kt#L37

And releasing it in `WinhttpRequest#close()`: https://github.com/dtretyakov/ktor/blob/client/winhttp/ktor-client/ktor-client-winhttp/windows/src/io/ktor/client/engine/winhttp/internal/WinHttpRequest.kt#L374

PR #3113, opened 2022-07-28
KTOR-4695 Fix UnixSocketAddress.path

ptitjes:
**Subsystem**
ktor-network

**Motivation**
Fixes a regression in the JVM version of UnixSocketAddress (cf [KTOR-4695](https://youtrack.jetbrains.com/issue/KTOR-4695/Regression-UnixSocketAddresspath-fails-on-JVM))

**Solution**
Simply give the right object to the call to `getPath.invoke()`.

### DIFF
--- a/ktor-network/jvm/src/io/ktor/network/sockets/SocketAddressJvm.kt
+++ b/ktor-network/jvm/src/io/ktor/network/sockets/SocketAddressJvm.kt
@@ -65,7 +65,7 @@ public actual class UnixSocketAddress internal constructor(
     public actual val path: String
         get() {
             val getPath: Method = checkSupportForUnixDomainSockets().getMethod("getPath")
-            return getPath.invoke(this).toString()
+            return getPath.invoke(address).toString()
         }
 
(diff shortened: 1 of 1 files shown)


whyoleg:
I think, that may be ktor-team need to setup testing all ktor codebase on JDK 8, 11 and 17 (not sure, if it's possible on infra). ping @e5l

e5l:
Thanks, will take a look!

ptitjes:
@Thomas-Vos Well, it happens! :)

There is a coarse `testEchoOverUnixSockets` test in `ktor-network/jvmAndNix/test/io/ktor/network/sockets/tests/TcpSocketTest.kt` that only runs on JDK 16+ (as it is guarded by an `if (!supportsUnixDomainSockets())`). But it doesn't seem to run on the CI.

Also, there are no simple unit tests for the creation of `SocketAddress` objects, nor for the access of their properties. @e5l should I add some?

e5l (approved):
lgtm, will merge after CI

PR #3114, opened 2022-07-29
KTOR-4697 Consider wss to be a secure context

LukasForst:
**Subsystem**
`ktor-server-host-common`

**Motivation**
It solves https://youtrack.jetbrains.com/issue/KTOR-4697 

**Solution**
The application now considers `wss` secure context and allows setting `Secure` cookies even  `wss` is used as a request scheme (for example in case when reverse proxy uses `X-Forwarded-Proto: wss` as described in  KTOR-4697)

### DIFF
--- a/ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/BaseApplicationResponseJvm.kt
+++ b/ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/BaseApplicationResponseJvm.kt
@@ -30,7 +30,7 @@ public actual abstract class BaseApplicationResponse actual constructor(
         get() = responded
 
     actual override val cookies: ResponseCookies by lazy {
-        ResponseCookies(this, call.request.origin.scheme == "https")
+        ResponseCookies(this, call.request.origin.scheme == "https" || call.request.origin.scheme == "wss")
     }
 
     actual override fun status(): HttpStatusCode? = _status
--- a/ktor-server/ktor-server-host-common/nix/src/io/ktor/server/engine/BaseApplicationResponseNative.kt
+++ b/ktor-server/ktor-server-host-common/nix/src/io/ktor/server/engine/BaseApplicationResponseNative.kt
@@ -27,7 +27,7 @@ public actual abstract class BaseApplicationResponse actual constructor(
         get() = responded
 
     actual override val cookies: ResponseCookies by lazy {
-        ResponseCookies(this, call.request.origin.scheme == "https")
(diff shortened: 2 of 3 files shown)


rsinukov (commented):
Hi and thanks for the PR
Can you please add a test for this change?

LukasForst:
@rsinukov I wasn't sure where to exactly put it, but this test fails if `wss` is not considered to be secure context, now it works.

ktor-server/ktor-server-tests/jvmAndNix/test/io/ktor/tests/server/plugins/CookiesTest.kt
```diff
@@ -377,3 +383,49 @@ class ParserServerSetCookieTest {
         assertEquals("/", parsed.path)
     }
 }
+
+class SecureCookiesTest {
+    @Test
+    fun `consider wss and https to be secure contexts`() {
+        testApplication {
```
rsinukov, line 390: can be inlined as `fun test() = testApplication { ... } `

ktor-server/ktor-server-tests/jvmAndNix/test/io/ktor/tests/server/plugins/CookiesTest.kt
```diff
@@ -377,3 +383,49 @@ class ParserServerSetCookieTest {
         assertEquals("/", parsed.path)
     }
 }
+
+class SecureCookiesTest {
```
rsinukov, line 387: This test should be in `jvm` folder, because Ktor for `native` doesn't support session

rsinukov (changes requested):
Some comments

ktor-server/ktor-server-tests/jvmAndNix/test/io/ktor/tests/server/plugins/CookiesTest.kt
LukasForst: inlined

ktor-server/ktor-server-tests/jvmAndNix/test/io/ktor/tests/server/plugins/CookiesTest.kt
LukasForst: moved to `jvm` folder

LukasForst:
@rsinukov I addressed all comments, is there anything else to do? Or can we merge this?

rsinukov (approved):
LGTM
Will merge after we publish 2.1.0

PR #3116, opened 2022-08-03
KTOR-4715  Fix crash in browser web workers where window is undefined

baconz:
**Subsystem**
Clients

**Motivation**
In web workers `window` will be undefined, and `window === undefined` will crash because `window` is not dynamic.

**Solution**
I'm sure there is a more elegant way to accomplish this, but it seems like the best solution for now might be to just execute raw JS to check for `undefined`.

### DIFF
--- a/ktor-http/js/src/io/ktor/http/URLBuilderJs.kt
+++ b/ktor-http/js/src/io/ktor/http/URLBuilderJs.kt
@@ -16,7 +16,7 @@ import org.w3c.workers.*
  */
 public actual val URLBuilder.Companion.origin: String
     get() = when {
-        PlatformUtils.IS_BROWSER -> if (window !== undefined) {
+        PlatformUtils.IS_BROWSER -> if (js("typeof window !== 'undefined'") as Boolean) {
             window.location.origin
         } else {
             js("self.location.origin") as String


rsinukov (commented):
LGTM in general. Do you think it's possible to add a test for this change?

baconz:
It seems like in order to write a good test for this we'd need a way to run the test in a JS runtime. I didn't see a way to do that based on the existing tests. I could just mock `js()`, I suppose. WDYT?

rsinukov:
I see that there is similar code in `CryptoJS.kt` file, can you update it as well, please?
Yes, creating test is not easy here, you need to mess with gradle config to use worker.

baconz:
It looks like the code in [`CryptoJs.kt`](https://github.com/ktorio/ktor/blob/main/ktor-utils/js/src/io/ktor/util/CryptoJs.kt#L51) is already doing the right thing:

```
        js("(window ? (window.crypto ? window.crypto : window.msCrypto) : self.crypto)")
```

In a worker `self.crypto` should exist, and behave the same way as `window.crypto`. The issue would only arise if that was kotlin code.

rsinukov (approved):
LGTM

PR #3119, opened 2022-08-08
Fix breaking changes before 2.1.0

e5l:


ktor-client/ktor-client-plugins/ktor-client-content-negotiation/common/src/io/ktor/client/plugins/contentnegotiation/ContentNegotiation.kt
```diff
@@ -77,11 +92,77 @@ public class ContentNegotiation internal constructor(
             registrations.add(registration)
         }
 
+        /**
+         * Adds a type to the list of types that should be ignored by [ContentNegotiation].
+         *
+         * The list contains the [HttpStatusCode] type by default.
```
rsinukov, line 98: not only `HttpStatusCode`

ktor-client/ktor-client-plugins/ktor-client-content-negotiation/common/src/io/ktor/client/plugins/contentnegotiation/ContentNegotiation.kt
```diff
@@ -77,11 +92,77 @@ public class ContentNegotiation internal constructor(
...
+         */
+        public inline fun <reified T> ignoreType() {
+            ignoredTypes.add(T::class)
+        }
+
+        /**
+         * Clear all configured ignored types including defaults.
```
rsinukov, line 105: This api is not really usefull. If you want to have all default types, but not string, you need to copy the code from sources and keep it up to date.

ktor-client/ktor-client-plugins/ktor-client-content-negotiation/common/src/io/ktor/client/plugins/contentnegotiation/ContentNegotiation.kt
```diff
@@ -9,14 +9,24 @@ import io.ktor.client.plugins.*
...
 import io.ktor.serialization.*
 import io.ktor.util.*
+import io.ktor.util.reflect.*
 import io.ktor.utils.io.*
 import io.ktor.utils.io.charsets.*
 import kotlin.reflect.*
 
+internal val DefaultCommonIgnoredTypes: Set<KClass<*>> = setOf(
+    ByteArray::class,
+    String::class,
```
rsinukov, line 24: It's annoying that we change default behaviour two times in a row =(

ktor-client/ktor-client-plugins/ktor-client-content-negotiation/common/src/io/ktor/client/plugins/contentnegotiation/ContentNegotiation.kt
```diff
@@ -9,14 +9,24 @@ import io.ktor.client.plugins.*
...
+    HttpStatusCode::class,
+    ByteReadChannel::class,
+    OutgoingContent::class
```
rsinukov, line 27: Outgoing content is excluded in the parsing anyway

ktor-client/ktor-client-plugins/ktor-client-content-negotiation/common/test/io/ktor/client/plugins/IgnoreTypesTest.kt
```diff
@@ -0,0 +1,87 @@
...
+
+        message = assertFails {
+            plugin.convertResponse(typeInfo<String>(), ByteReadChannel("FOO"), ContentType.Application.Json)
+        }.message
+
+        assertEquals("This should not be called", message)
+    }
+
+    @Test
+    fun testIgnoresString() = testSuspend {
```
rsinukov, line 70: can you write more user-facing test here?

rsinukov (changes requested):

gradle.properties
```diff
@@ -4,7 +4,7 @@
 
 # ktor
 # Makes Ktor import and resolve
-ktor.ide.jvmAndCommonOnly=true
+ktor.ide.jvmAndCommonOnly=false
```
rsinukov, line 7: please revert

ktor-server/ktor-server-core/jvm/src/io/ktor/server/config/HoconApplicationConfig.kt
```diff
@@ -5,22 +5,36 @@
...
-        val file = File(resolvedPath)
-        if (file.exists()) {
-            return HoconApplicationConfig(ConfigFactory.parseFile(file).resolve())
-        }
-        return null
+        val config = when {
+            resource != null -> ConfigFactory.load(resolvedPath)
+            else -> {
+                val file = File(resolvedPath)
+                ConfigFactory.parseFile(file)
```
rsinukov, line 33: please add a check that file exists

rsinukov (changes requested):

ktor-client/ktor-client-plugins/ktor-client-content-negotiation/common/src/io/ktor/client/plugins/contentnegotiation/ContentNegotiation.kt
e5l: Let's add remove with type as well

gradle.properties
e5l: thanks, done

ktor-server/ktor-server-core/jvm/src/io/ktor/server/config/HoconApplicationConfig.kt
e5l: thanks, done

ktor-client/ktor-client-plugins/ktor-client-content-negotiation/common/test/io/ktor/client/plugins/IgnoreTypesTest.kt
e5l: Sure

ktor-client/ktor-client-plugins/ktor-client-content-negotiation/common/src/io/ktor/client/plugins/contentnegotiation/ContentNegotiation.kt
e5l: fixed

ktor-client/ktor-client-plugins/ktor-client-content-negotiation/common/src/io/ktor/client/plugins/contentnegotiation/ContentNegotiation.kt
e5l: Yep, but we have no other options

rsinukov (approved):

ktor-client/ktor-client-plugins/ktor-client-content-negotiation/common/src/io/ktor/client/plugins/contentnegotiation/ContentNegotiation.kt
```diff
@@ -77,11 +92,84 @@ public class ContentNegotiation internal constructor(
             registrations.add(registration)
         }
 
+        /**
+         * Adds a type to the list of types that should be ignored by [ContentNegotiation].
+         *
+         * The list contains the [HttpStatusCode], [ByteArray], [String] and streaming types by default.
+         */
+        public inline fun <reified T> ignoreType() {
```
whyoleg, line 100: does it make sense to have API with `KClass`? (this will also allow us to make `ignoredTypes` not `PublishedApi`

ktor-client/ktor-client-plugins/ktor-client-json/common/src/io/ktor/client/plugins/json/JsonPlugin.kt
```diff
@@ -111,6 +129,29 @@ public class JsonPlugin internal constructor(
         public fun receive(matcher: ContentTypeMatcher) {
             _receiveContentTypeMatchers += matcher
         }
+
+        /**
+         * Adds a type to the list of types that should be ignored by [ContentNegotiation].
+         *
+         * The list contains the [HttpStatusCode], [ByteArray], [String] and streaming types by default.
+         */
+        public inline fun <reified T> ignoreType() {
```
whyoleg, line 138: does it make sense at all to add new functionality to deprecated plugin?

e5l: Yep, we have internal breaking changes because of that

ktor-client/ktor-client-plugins/ktor-client-content-negotiation/common/src/io/ktor/client/plugins/contentnegotiation/ContentNegotiation.kt
e5l: Could you tell me why we need KClass API and why PublishedApi is not ok?

whyoleg: 1. KClass (or even KType) Api is sometimes needed, when you want to dynamically configure ignored types. Even internally `DefaultCommonIgnoredTypes` and `DefaultIgnoredTypes` are sets of types, which you want to ignore in not reified manner. So I would expect, that users will also want at some day do something like this. There were already several times issue raised in youtrack in routing, that ktor only has reified functions.
2. `PublishedApi` is tracked in binary compatibility. So if later `ignoreType` underlying representation will change and f.e. `ignoredTypes` will be not Set, or we will add additional parameter to `ignoreType` which we will need to store, we should keep `ignoredTypes` there, working, as it's can be referenced from `public` api because of inline functions, that were already in user code and compiled. So it makes internal representation effectively public, which is not good.

e5l: Agree! Let's add KClass API

PR #3122, opened 2022-08-15
KTOR-4141 KTOR-2143 Fixed inconsistency between server engines when determining port/host of an incoming request

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-430
https://youtrack.jetbrains.com/issue/KTOR-2501
https://youtrack.jetbrains.com/issue/KTOR-4141
https://youtrack.jetbrains.com/issue/KTOR-2143

### DIFF
--- a/ktor-http/common/src/io/ktor/http/RequestConnectionPoint.kt
+++ b/ktor-http/common/src/io/ktor/http/RequestConnectionPoint.kt
@@ -11,7 +11,7 @@ package io.ktor.http
  */
 public interface RequestConnectionPoint {
     /**
-     * Request scheme, for example "http" or "https"
+     * Scheme, for example "http" or "https"
      */
     public val scheme: String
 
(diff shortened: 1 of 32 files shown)


ktor-http/common/src/io/ktor/http/RequestConnectionPoint.kt
```diff
@@ -23,13 +23,40 @@ public interface RequestConnectionPoint {
     /**
      * Request port, for example 80 or 443
      */
+    @Deprecated("Use localPort or serverPort instead")
     public val port: Int
 
+    /**
+     * Request port on which the request was received, for example 80 or 443
```
e5l, line 30: I'm not sure that I understand the doc. I would say it's `The port used to accept the client connection. It's usually specified in the client request`

ktor-http/common/src/io/ktor/http/RequestConnectionPoint.kt
```diff
@@ -23,13 +23,40 @@ public interface RequestConnectionPoint {
...
     /**
      * Request host, useful for virtual hosts routing
      */
+    @Deprecated("Use localHost or serverHost instead")
     public val host: String
 
+    /**
+     * Request host on which the request was received, is useful for virtual hosts routing
+     */
+    public val localHost: String
```
e5l, line 48: is this hostname only?

e5l (approved):
LGTM 

@andreyaksenov, could you check docs?
@rsinukov, please run the space build after merging

ktor-http/common/src/io/ktor/http/RequestConnectionPoint.kt
rsinukov: I took the same terminology and naming as in Jetty

ktor-http/common/src/io/ktor/http/RequestConnectionPoint.kt
rsinukov: yes, port is separate property

PR #3130, opened 2022-08-22
KTOR-4736 Replace fail with warning in blocking stream adapters

e5l:
Fix [KTOR-4736](https://youtrack.jetbrains.com/issue/KTOR-4736)

### DIFF
--- a/ktor-io/jvm/src/io/ktor/utils/io/jvm/javaio/Blocking.kt
+++ b/ktor-io/jvm/src/io/ktor/utils/io/jvm/javaio/Blocking.kt
@@ -8,10 +8,12 @@ import io.ktor.utils.io.*
 import kotlinx.atomicfu.*
 import kotlinx.coroutines.*
 import kotlinx.coroutines.CancellationException
+import org.slf4j.LoggerFactory
 import java.io.*
-import java.util.concurrent.locks.*
 import kotlin.coroutines.*
 import kotlin.coroutines.intrinsics.*
+
+private val ADAPTER_LOGGER by lazy { LoggerFactory.getLogger(BlockingAdapter::class.java) }
 
 /**
  * Create blocking [java.io.InputStream] for this channel that does block every time the channel suspends at read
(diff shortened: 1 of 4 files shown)


ktor-io/jvm/src/io/ktor/utils/io/jvm/javaio/Blocking.kt
```diff
@@ -58,7 +56,7 @@ private class InputAdapter(parent: Job?, private val channel: ByteReadChannel) :
         val buffer = single ?: ByteArray(1).also { single = it }
         val rc = loop.submitAndAwait(buffer, 0, 1)
         if (rc == -1) return -1
-        if (rc != 1) error("rc should be 1 or -1 but got $rc")
+        if (rc != 1) error("internal error: read count should be 1 or -1 but got $rc")
```
Stexxe, line 59: > Expected a single byte or nothing. Got $rc bytes.

ktor-io/jvm/src/io/ktor/utils/io/jvm/javaio/Blocking.kt
```diff
@@ -246,6 +244,21 @@ private abstract class BlockingAdapter(val parent: Job? = null) {
     @OptIn(InternalCoroutinesApi::class)
     private fun parkingLoop(thread: Thread) {
         if (state.value !== thread) return
+        if (!isParkingAllowed()) {
+            ADAPTER_LOGGER.warn(
+                """
+                Blocking network thread detected. 
+                Please make sure you're using blocking IO primitives like InputStream and OutputStream only only in
```
Stexxe, line 251: only only

ktor-io/jvm/src/io/ktor/utils/io/jvm/javaio/Blocking.kt
```diff
@@ -246,6 +244,21 @@ private abstract class BlockingAdapter(val parent: Job? = null) {
...
+                """
+                Blocking network thread detected. 
+                Please make sure you're using blocking IO primitives like InputStream and OutputStream only only in 
+                the context of Dispatchers.IO:
+                ```
+                withContext(Dispatchers.IO) {
+                    myInputStream.read()
+                }
+                ```
+                It can possible lead to a performance decline or even a deadlock.
```
Stexxe, line 258: This line is confusing because it is placed after the solution to the problem (feels like the solution can lead to a decline or a deadlock). It's better to structure it this way: problem description -> possible consequences -> the solution

ktor-io/jvm/test/io/ktor/utils/io/BlockingAdaptersOnProhibitedThreadTest.kt
```diff
@@ -42,12 +40,10 @@ class BlockingAdaptersOnProhibitedThreadTest {
     }
 
     @Test
-    fun testOutputStreamProhibited() {
+    fun testOutputStreamProhibitedNotFails() {
```
Stexxe, line 43: It's better to name a test without negation like `canConvertChannelToOutputStreamInNetworkThread`

Stexxe (changes requested):

Stexxe (approved):

rsinukov (approved):

PR #3133, opened 2022-08-22
KTOR-4771 Fix Receive Non-Nullable null Message

e5l:
Fix [KTOR-4771](https://youtrack.jetbrains.com/issue/KTOR-4771)

### DIFF
--- a/ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/request/ApplicationReceiveFunctions.kt
+++ b/ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/request/ApplicationReceiveFunctions.kt
@@ -58,7 +58,8 @@ public suspend inline fun <reified T : Any> ApplicationCall.receiveOrNull(): T? 
  * @return instance of [T] received from this call.
  * @throws ContentTransformationException when content cannot be transformed to the requested type.
  */
-public suspend inline fun <reified T : Any> ApplicationCall.receive(): T = receiveNullable(typeInfo<T>())!!
+public suspend inline fun <reified T : Any> ApplicationCall.receive(): T = receiveNullable(typeInfo<T>())
+    ?: throw CannotTransformContentToTypeException(typeInfo<T>().kotlinType!!)
 
 /**
  * Receives content for this request.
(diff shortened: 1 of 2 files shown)


ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/request/ApplicationReceiveFunctions.kt
```diff
@@ -58,7 +58,8 @@ public suspend inline fun <reified T : Any> ApplicationCall.receiveOrNull(): T?
  * @return instance of [T] received from this call.
  * @throws ContentTransformationException when content cannot be transformed to the requested type.
  */
-public suspend inline fun <reified T : Any> ApplicationCall.receive(): T = receiveNullable(typeInfo<T>())!!
+public suspend inline fun <reified T : Any> ApplicationCall.receive(): T = receiveNullable(typeInfo<T>())
+    ?: throw CannotTransformContentToTypeException(typeInfo<T>().kotlinType!!)
```
Stexxe, line 62: Also `CannotTransformContentToTypeException` will be thrown if transformation result is `null` and it's successful.

Stexxe: Why do need to throw the same exception when it's raised already in `ApplicationCall.receiveNullable`? The only case it covers is when a result is `NullBody` so the exception is misleading.

ktor-server/ktor-server-core/jvmAndNix/test/io/ktor/server/application/ApplicationCallReceiveTest.kt
```diff
@@ -0,0 +1,47 @@
...
+                    NullBody
+                }
+            }
+        }
+
+        install(nullPlugin)
+
+        routing {
+            post("/") {
+                val result: String = try {
```
Stexxe, line 34: Could be simplified with the `assertFailsWith` inside route's hanlder

ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/request/ApplicationReceiveFunctions.kt
e5l: We throw `ContentTransformationException` if content cannot be transformed to the requested type. If content is converted to null for non nullable type it's `ContentTransformationException`

rsinukov, line 62: maybe extract `typeInfo<T>()` to val?

rsinukov (approved):

ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/request/ApplicationReceiveFunctions.kt
e5l: I would keep it as one-liner cause `typeInfo` is compile time constant

PR #3136, opened 2022-08-23
KTOR-4784 Provide access to the original call from RoutingApplicationCall

rsinukov:


ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/routing/RoutingApplicationCall.kt
```diff
@@ -14,28 +14,30 @@ import kotlin.coroutines.*
 
 /**
  * An application call handled by [Routing].
+ * @property call original call from [ApplicationEngine]
  * @property route is the selected route
  */
 public class RoutingApplicationCall(
-    private val call: ApplicationCall,
+    public val originalCall: ApplicationCall,
```
e5l, line 21: If it's the call from an engine should we name it `engineCall` then?

e5l (approved):
lgtm

PR #3137, opened 2022-08-25
KTOR-621 Add extensions to check if the Url is absolute or relative

e5l:
Fix [KTOR-621](https://youtrack.jetbrains.com/issue/KTOR-621/The-beginning-character-of-endodedPath-fieldUrl-class-is-wrong-when-relative-path)

### DIFF
--- a/ktor-http/common/src/io/ktor/http/URLUtils.kt
+++ b/ktor-http/common/src/io/ktor/http/URLUtils.kt
@@ -132,22 +132,22 @@ public fun Appendable.appendUrlFullPath(
 /**
  * Checks if [Url] has absolute path.
  */
-public val Url.isAbsolute: Boolean get() = pathSegments.firstOrNull() == ""
+public val Url.isAbsolutePath: Boolean get() = pathSegments.firstOrNull() == ""
 
 /**
  * Checks if [Url] has absolute path.
  */
-public val Url.isRelative: Boolean get() = !isAbsolute
+public val Url.isRelativePath: Boolean get() = !isAbsolutePath
 
 /**
  * Checks if [Url] has absolute path.
  */
-public val URLBuilder.isAbsolute: Boolean get() = pathSegments.firstOrNull() == ""
+public val URLBuilder.isAbsolutePath: Boolean get() = pathSegments.firstOrNull() == ""
 
 /**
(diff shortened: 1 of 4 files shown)


ktor-http/common/src/io/ktor/http/URLUtils.kt
```diff
@@ -129,6 +129,26 @@ public fun Appendable.appendUrlFullPath(
         }
 }
 
+/**
+ * Checks if [Url] has absolute path.
+ */
+public val Url.isAbsolute: Boolean get() = pathSegments.firstOrNull() == ""
```
rsinukov, line 135: What if you build URL like `Url("/some/path")`? Then `path` is absolute, but url itself is relative. 
Maybe rename this extension to `isAbsolutePath` and introduce `val Url.isAbsolute get() = localhost.isEmpty()`?

rsinukov (changes requested):

ktor-http/common/src/io/ktor/http/URLUtils.kt
e5l: let's rename it about path

ktor-http/common/test/io/ktor/tests/http/UrlTest.kt
```diff
@@ -302,4 +302,18 @@ class UrlTest {
         val url = Url(urlString)
         assertEquals(urlString, "$url")
     }
+
+    @Test
+    fun testIsAbsolute() {
```
rsinukov, line 307: `testIsAbsolutePath`

ktor-http/common/test/io/ktor/tests/http/UrlTest.kt
```diff
@@ -302,4 +302,18 @@ class UrlTest {
...
+        assertTrue(Url("https://ktor.io/").isAbsolutePath)
+        assertTrue(Url("/").isAbsolutePath)
+        assertTrue(Url("/hello").isAbsolutePath)
+    }
+
+    @Test
+    fun testIsRelative() {
```
rsinukov, line 314: `testIsRelativePath`

rsinukov (approved):

Issue #1903, opened 2020-05-28
Add Url capability: parsing/validating url strings


edenman:
**Subsystem**
ktor-http

**Is your feature request related to a problem? Please describe.**
My app gives me unexpected results on `Url(String)`

**Describe the solution you'd like**
Url should have a factory method that returns an optional Url, something like:
```
class Url {
  companion object {
    fun parse(url: String): Url? { ... }
  }
}
```
This method should return null if the url is not valid.  It should not throw exceptions.

**Motivation to include to ktor**
User-submitted content often has urls in it and it would be useful to validate them and extract the various url parts.

Specific urls that are currently broken:

`Url("bogus").toString()` returns `http://localhost/bogus` and I think it should return null from this new method.
`Url("http://localhost:7000Pragma").toString()` throws an exception and I think it should return null from this new method.

edenman:
For now I'm just working around this on my end by doing:
```
fun Url.Companion.parse(str: String): Url? {
  try {
    return Url(str)
  } catch (e: Throwable) {
    return null
  }
}
```
but obviously this doesn't address the "bogus" case.

oleg-larshin:
Please check the <a href="https://youtrack.jetbrains.com/issue/KTOR-360">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

victorcwai:
> `Url("bogus").toString()` returns `http://localhost/bogus` and I think it should return null from this new method.

Currently the `takeFrom(String)` function called by `Url(urlString)` allow parsing relative URLs (see https://github.com/ktorio/ktor/pull/849).
However if no base URL is provided in `Url(urlString)`, `protocol` and `host` will default to `http` and `localhost` respectively.

I am not sure if the `Url(urlString)` function should return a default `protocol` and `host`. Maybe it should just set `protocol` and `host` to null or throw an exception if no `protocol` is provided?

chetankokil:
can i work on this @edenman , @victorcwai ?

chetankokil:
@e5l team, please review the PR when you get a chance.

J-Swift:
This also got me. Its extremely unexpected to have a relative url get parsed as `localhost` relative.

jsoberg:
Would it make sense to return a `kotlin.Result` here? The consumer could then fold it to handle each case, or call `getOrNull()` on the result if they want it to have the functionality described here (`null` for invalid).

ablx:
[...]

PR #3139, opened 2022-08-28
Issue #1903. Added code to parse the url if valid else return null

chetankokil:
**Subsystem**
ktor-Http

**Motivation**
**issue #1903**. Currently the Url parse return valid url when a bogus url is passed.

**Solution**
Added a function to the Url Companion object, which parse the urlstring and return the valid url if present else returns null instead of appending `http://localhost` to the invalid url

### DIFF
--- a/ktor-http/common/src/io/ktor/http/Url.kt
+++ b/ktor-http/common/src/io/ktor/http/Url.kt
@@ -116,7 +116,11 @@ public class Url internal constructor(
         return urlString.hashCode()
     }
 
-    public companion object
+    public companion object {
+        public fun parse(urlString: String) : Url? {
+            return Url(urlString)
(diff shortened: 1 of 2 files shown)


rsinukov (changes requested):
Hi and sorry for a long time to reply. I see that one of the tests that you added does not pass. Can you fix it, please?
Also, name `parse` is confusing, because it doesn't highlight intended behaviour.

chetankokil:
> Hi and sorry for a long time to reply. I see that one of the tests that you added does not pass. Can you fix it, please? Also, name `parse` is confusing, because it doesn't highlight intended behaviour.

Sir , somehow i am not able to run the test. can you please suggest, how to se

PR #3143, opened 2022-08-29
KTOR-550 Fix MDC exception crash request

e5l:
Fix [KTOR-550](https://youtrack.jetbrains.com/issue/KTOR-550/)

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-call-logging/jvm/src/io/ktor/server/plugins/callloging/MDCEntryUtils.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-call-logging/jvm/src/io/ktor/server/plugins/callloging/MDCEntryUtils.kt
@@ -29,7 +29,8 @@ internal fun List<MDCEntry>.setup(call: ApplicationCall): Map<String, String> {
     val result = HashMap<String, String>()
 
     forEach { entry ->
-        entry.provider(call)?.let { mdcValue ->
+        val provider = runCatching { entry.provider(call) }.getOrNull()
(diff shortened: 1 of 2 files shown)


ktor-server/ktor-server-plugins/ktor-server-call-logging/jvm/src/io/ktor/server/plugins/callloging/MDCEntryUtils.kt
```diff
@@ -29,7 +29,8 @@ internal fun List<MDCEntry>.setup(call: ApplicationCall): Map<String, String> {
     val result = HashMap<String, String>()
 
     forEach { entry ->
-        entry.provider(call)?.let { mdcValue ->
+        val provider = runCatching { entry.provider(call) }.getOrNull()
```
rsinukov, line 32: maybe at least log error here?

rsinukov (changes requested):

ktor-server/ktor-server-plugins/ktor-server-call-logging/jvm/src/io/ktor/server/plugins/callloging/MDCEntryUtils.kt
e5l: Nope, we shouldn't.

We introduced this problem by intercepting several phases to put MDC in, and in the earliest phase session access will fail because they were not handled at that moment. It leads to an exception and a whole request failure.

rsinukov (approved):

PR #3144, opened 2022-08-29
KTOR-4032 Fix netty timeout

e5l:
[KTOR-4032](https://youtrack.jetbrains.com/issue/KTOR-4032)


rsinukov (approved):
But please update the commit message and mention CIO there as well

PR #3145, opened 2022-08-29
KTOR-4533 Allow to use configuration in plugins

e5l:
[KTOR-4533](https://youtrack.jetbrains.com/issue/KTOR-4533)


ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/application/CreatePluginUtils.kt
```diff
@@ -4,10 +4,51 @@
...
+ * application.install(RequestLoggingPlugin)
+ * ```
+ *
+ * You can learn more from [Custom plugins](https://ktor.io/docs/custom-plugins.html).
+ *
+ * @param name A name of a plugin that is used to get its instance.
+ * @param createConfiguration Defines how the initial [PluginConfigT] of your new plugin can be created.
+ * Note that it may be modified later when a user of your plugin calls [Application.install].
+ * @param body Allows you to define handlers ([onCall], [onCallReceive], [onCallRespond] and so on) that
+ * can modify the behaviour of an [Application] where your plugin is installed.
```
rsinukov, line 32: `configurationPath` is not in the doc

ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/application/CreatePluginUtils.kt
```diff
@@ -94,6 +134,57 @@ public fun <PluginConfigT : Any> createRouteScopedPlugin(
...
+ * ```
+ *
+ * You can learn more from [Custom plugins](https://ktor.io/docs/custom-plugins.html).
+ *
+ * @param name A name of a plugin that is used to get its instance
+ * when it is installed to [io.ktor.server.routing.Routing].
+ * @param createConfiguration Defines how the initial [PluginConfigT] of your new plugin can be created. Please
+ * note that it may be modified later when a user of your plugin calls [install].
+ * @param body Allows you to define handlers ([onCall], [onCallReceive], [onCallRespond] and so on) that
+ * can modify the behaviour of an [Application] where your plugin is installed.
```
rsinukov, line 160: same here, `configurationPath` is not in the doc

rsinukov (changes requested):

ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/application/CreatePluginUtils.kt
e5l: Thanks! Fixed

ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/application/CreatePluginUtils.kt
e5l: Thanks! Fixed

rsinukov (approved):

PR #3151, opened 2022-09-02
KTOR-1647 Extract autoreload in the separate module

e5l:
Fix [KTOR-1647](https://youtrack.jetbrains.com/issue/KTOR-1647/comsunniofileSensitivityWatchEventModifier-Move-the-reflection-call-of-this-modifier-out-from-the-Ktor-Core)


rsinukov:
Please check CI

ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/ApplicationEngineEnvironmentJvm.kt
```diff
@@ -72,8 +72,69 @@ public actual class ApplicationEngineEnvironmentBuilder {
...
+                connectors,
+                modules,
+                watchPaths,
+                parentCoroutineContext,
+                rootPath,
+                developmentMode
+            )
+
+            if (result != null) return result
+            val message = "Failed to initialize reloading environment. " +
```
whyoleg, line 89: why not to throw exception here, as anyway it's `dev` mode?
And in this case also, f.e. autoreload can be explicitly disabled if `watchPaths` is empty or via some flag?

ktor-server/ktor-server-plugins/ktor-server-autoreload/api/ktor-server-autoreload.api
```diff
@@ -0,0 +1,23 @@
...
+	public fun getLog ()Lorg/slf4j/Logger;
+	public fun getMonitor ()Lio/ktor/events/Events;
+	public fun getParentCoroutineContext ()Lkotlin/coroutines/CoroutineContext;
+	public fun getRootPath ()Ljava/lang/String;
+	public final fun reload ()V
+	public fun start ()V
+	public fun stop ()V
+}
+
+public final class io/ktor/server/autoreload/ApplicationEngineEnvironmentReloading$Companion {
```
whyoleg, line 21: while we anyway brake binary compatibility here, may be just remove empty companion?

ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/StaticApplicationEngineEnvironment.kt
```diff
@@ -0,0 +1,61 @@
...
+package io.ktor.server.engine
+
+import io.ktor.events.*
+import io.ktor.events.EventDefinition
+import io.ktor.server.application.*
+import io.ktor.server.config.*
+import io.ktor.util.logging.*
+import kotlin.coroutines.*
+
+internal class StaticApplicationEngineEnvironment(
```
whyoleg, line 14: looks like it can be shared with nix implementation, except ClassLoader I think. But still then expect/actual can be just cleaner IMO

whyoleg (changes requested):

ktor-server/ktor-server-host-common/jvm/src/io/ktor/server/engine/ApplicationEngineEnvironmentJvm.kt
e5l: `dev` mode is not only auto-reloading, it's stack traces. 

Yep, let's make it disabled on empty watchPaths

ktor-server/ktor-server-plugins/ktor-server-autoreload/api/ktor-server-autoreload.api
e5l: I'm going to fix binary compatibility

e5l:
It looks like extracting is not possible for now, will try a different solution

PR #3158, opened 2022-09-06
KTOR-4809 Replace val with fun to fix init order

e5l:


rsinukov (approved):
pls check codestyle build

PR #3160, opened 2022-09-07
KTOR-2572 Kotlinx serialization support in sessions

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-2572


ktor-server/ktor-server-plugins/ktor-server-sessions/jvm/src/io/ktor/server/sessions/KotlinxSessionSerializer.kt
```diff
@@ -0,0 +1,397 @@
...
+        if (index == CompositeDecoder.UNKNOWN_NAME) {
+            throw IllegalStateException(
+                "${enumDescriptor.serialName} does not contain element with name '$enumName'"
+            )
+        }
+        return index
+    }
+}
+
+@OptIn(ExperimentalSerializationApi::class, InternalSerializationApi::class)
```
whyoleg, line 315: why `InternalSerializationApi` is used here?
Not sure it's a good idea. If it's not possible, IMO it would be better to reduce scope of this annotation per declaration, instead of class

whyoleg (changes requested):

ktor-server/ktor-server-plugins/ktor-server-sessions/jvm/src/io/ktor/server/sessions/KotlinxSessionSerializer.kt
rsinukov: `NamedValueDecoder` is marked as `InternalSerializationApi`. It simplifies a lot parsing data into a map. Another option would be to replicate it in our codebase, which I would like to avoid. `MapDecoder` is private, so it's relatively safe to have it here.

ktor-server/ktor-server-plugins/ktor-server-sessions/jvm/src/io/ktor/server/sessions/KotlinxSessionSerializer.kt
```diff
@@ -0,0 +1,397 @@
...
+    override fun serialize(session: T): String = format.encodeToString(serializer, session)
+    override fun deserialize(text: String): T = format.decodeFromString(serializer, text)
+}
+
+/**
+ * Special format for [KotlinxSessionSerializer] that is backward compatible with previous default serializer.
+ * In general, it's discouraged to use this format, and it's recommended to migrate your sessions to another format,
+ * such as JSON
+ */
+public class SessionsBackwardCompatibleFormat(override val serializersModule: SerializersModule) : StringFormat {
```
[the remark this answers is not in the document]
rsinukov: I thought about it, but I'm not sure that this is a good idea. This class has a valid use case, especially in multiplatform. I think we can keep it as it is for some time and deprecate it in 2.3.0 or later, but before 3.0.0

ktor-server/ktor-server-plugins/ktor-server-sessions/build.gradle.kts
```diff
@@ -4,8 +4,17 @@
 
 description = ""
 
+plugins {
+    id("kotlinx-serialization")
+}
+
 kotlin {
     sourceSets {
+        jvmMain {
```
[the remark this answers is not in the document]
rsinukov: It's planned, but in a separate PR

ktor-server/ktor-server-plugins/ktor-server-sessions/jvm/src/io/ktor/server/sessions/KotlinxSessionSerializer.kt
whyoleg: the main issue is if `NamedValueDecoder` will be changed in not backward compatible maner (like adding some function without default implementation, which is safe for kx.serialization, as it's internal API) it can cause backward compatibility issues for ktor.

ktor-server/ktor-server-plugins/ktor-server-sessions/jvm/src/io/ktor/server/sessions/KotlinxSessionSerializer.kt
```diff
@@ -0,0 +1,397 @@
...
+import kotlinx.serialization.*
+import kotlinx.serialization.descriptors.*
+import kotlinx.serialization.encoding.*
+import kotlinx.serialization.internal.*
+import kotlinx.serialization.modules.*
+
+/**
+ * Returns a [SessionSerializer] based on `kotlinx.serialization` library.
+ */
+public inline fun <reified T : Any> kotlinxSessionSerializer(
```
e5l, line 17: Could we make it scoped? I would avoid adding session specific top-level symbols

e5l: it looks like the constructor function `KotlinxSessionSerializer`

ktor-server/ktor-server-plugins/ktor-server-sessions/jvm/src/io/ktor/server/sessions/KotlinxSessionSerializer.kt
```diff
@@ -0,0 +1,397 @@
...
+        is Boolean -> "#bo${value.toString().first()}"
+        is Char -> "#ch$value"
+        is String -> "#s$value"
+        is Enum<*> -> "#s${value.name}"
+        else -> null
+    }
+}
+
+@OptIn(ExperimentalSerializationApi::class)
+private class SessionsBackwardCompatibleDecoder(
```
e5l, line 165: Please consider splitting this file into multiple shorts. It's hard to maintain and review

ktor-server/ktor-server-plugins/ktor-server-sessions/jvm/src/io/ktor/server/sessions/KotlinxSessionSerializer.kt
e5l: Btw I can't find any public usages of it

ktor-server/ktor-server-plugins/ktor-server-sessions/jvm/src/io/ktor/server/sessions/KotlinxSessionSerializer.kt
```diff
@@ -0,0 +1,397 @@
...
+    serializer: KSerializer<T>,
+    format: StringFormat
+): KotlinxSessionSerializer<T> {
+    return KotlinxSessionSerializer(format, serializer)
+}
+
+/**
+ * Implementation of [SessionSerializer] based on `kotlinx.serialization` library.
+ */
+public class KotlinxSessionSerializer<T : Any> internal constructor(
```
whyoleg, line 36: BTW, why `KotlinxSessionSerializer` and `SessionsBackwardCompatibleFormat` as classes are public?
isn't it better for reducing public API to:
1. make `fun KotlinxSessionSerializer(...)` to just return `SessionSerializer`
2. create `fun SessionsBackwardCompatibleFormat(module): StringFormat` or even just `fun KotlinxBackwardCompatibleSessionSerializer(...): SessionSerializer`

ktor-server/ktor-server-plugins/ktor-server-sessions/jvm/src/io/ktor/server/sessions/KotlinxSessionSerializer.kt
rsinukov: This class is relatively stable and it should be ok to rely on it here in the private API.

ktor-server/ktor-server-plugins/ktor-server-sessions/jvm/src/io/ktor/server/sessions/KotlinxSessionSerializer.kt
rsinukov: Yes, good point. I hide everything behind functions.

ktor-server/ktor-server-plugins/ktor-server-sessions/api/ktor-server-sessions.api
```diff
@@ -85,6 +85,15 @@ public final class io/ktor/server/sessions/KotlinTimeJvmKt {
 	public static final fun setMaxAge-6Au4x4Y (Lio/ktor/server/sessions/CookieConfiguration;Lkotlin/time/Duration;)V
 }
 
+public final class io/ktor/server/sessions/KotlinxSessionSerializer : io/ktor/server/sessions/SessionSerializer {
```
whyoleg, line 88: looks like this is something old, only changes in the end of the file should be there
need to just regenerate API dump?

rsinukov: Yep, updating it. `apiDump` relies to much on gradle caches.

e5l (approved):
lgtm

PR #3161, opened 2022-09-12
KTOR-4849 Routing: Wrong content-type should result in 415

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-4849


e5l (approved):
We need to check this with space build

PR #3162, opened 2022-09-15
Allow `respondTemplate` set HTTP status

delitescere:
**Subsystem**
Server, Thymeleaf plugin

**Motivation**
The default HTTP status in `io.ktor.server.engine.BaseApplicationResponse#commitHeaders` is `HttpStatusCode.OK` (200). This is almost always applicable when responding with a template. However, in the case of templates being used along with the `StatusPages` plugin or for response entities for HTTP methods, it is necessary to respond with an appropriate HTTP status code.

**Solution**
This change allows for an optional parameter for the caller to set the status code along with responding with a template. IIUC, it should be binary compatible.

It does not address other templating library plugins. With a little effort all the template plugins could be harmonized under an interface for the `xContent` type and an abstract implementation for the `respondTemplate` function. The only differences I've noticed are for JTE with an additional `varargs params: Pair` signature (which delegates to the `model: Map` implementation), Moustache with any `model: Any?` signature (which doesn't obviate the otherwise common `model: Map`) , a slight difference in parameter order for Pebble, and the additional `fragments: Set` for Thymeleaf (which could be converted to a `fragments: Set?` to harmonize all the other plugins.


rsinukov (changes requested):
Hi and that for the PR.

LGTM in general, but some changes regarding the compatibility are necessary. Can you please:
1. rebase and reopen this PR into `2.2.0-eap` branch? We can't add new API in patch releases
2. instead of updating an existing function, deprecate it with `level = DeprecationLevel.HIDDEN` and add a new one? Otherwise, this change is not binary compatible.
3. run `./gradlew apiDump` and commit changes
4. create a ticket and link it to the PR

e5l (approved):
Hey @delitescere, thanks for the PR.

I've rebased it on the main branch. LGTM

PR #3163, opened 2022-09-15
KTOR-4009 Add flag to respond with 500 on exceptions in tests

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-4009


e5l (approved):
lgtm

PR #3170, opened 2022-09-28
KTOR-4652 Add flag to enable retrying on request timeout

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-4652

### DIFF
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/HttpRequestRetry.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/HttpRequestRetry.kt
@@ -13,7 +13,6 @@ import io.ktor.http.*
 import io.ktor.util.*
 import kotlinx.coroutines.*
 import kotlin.math.*
-import kotlin.native.concurrent.*
 import kotlin.random.*
 
 /**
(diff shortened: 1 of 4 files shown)


e5l (approved):
LGTM. Could you reopen this PR for 2.2.0 and update API dump?

@andreyaksenov please check the docs

andreyaksenov (approved):

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/HttpRequestRetryTest.kt
```diff
@@ -361,4 +360,57 @@ class HttpRequestRetryTest {
...
+                retryOnException(3, retryOnTimeout = true)
+                delay { }
+            }
+            install(HttpTimeout) {
+                requestTimeoutMillis = 1000
+            }
+        }
+
+        test { client ->
+            assertFailsWith<HttpRequestTimeoutException> {
```
yassenb, line 411: Shouldn't this pass? How's the expected behaviour different than the previous test?

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/HttpRequestRetry.kt
```diff
@@ -157,8 +156,25 @@ public class HttpRequestRetry internal constructor(configuration: Configuration)
...
+         * Enables retrying a request if an exception is thrown during the [HttpSend] phase
+         * and specifies the number of retries.
+         * By default, [HttpRequestTimeoutException] is not retried. Set [retryOnTimeout] to `true` to retry on timeout.
+         * Note, that in this case, [HttpTimeout] plugin should be installed after [HttpRequestRetry].
+         */
+        public fun retryOnException(maxRetries: Int = -1, retryOnTimeout: Boolean = false) {
+            retryOnExceptionIf(maxRetries) { _, cause ->
+                when {
+                    cause is CancellationException -> false
+                    cause is HttpRequestTimeoutException && retryOnTimeout -> true
```
yassenb, line 174: Should be `cause is HttpRequestTimeoutException -> retryOnTimeout`? Otherwise if `retryOnTimeout` is false it defaults to the `else` below which is `true`. See comment on tests as well

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/HttpRequestRetryTest.kt
MaaxGr: Seems to be correct in master :)
https://github.com/ktorio/ktor/blob/main/ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/HttpRequestRetryTest.kt#L458

PR #3172, opened 2022-09-29
KTOR-4722 renderSetCookieHeader should not encode extensions

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-4722


e5l (approved):
lgtm

PR #3175, opened 2022-09-30
KTOR-4904 Add option to support requests with Content-Length header in Jackson

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-4904

### DIFF
--- a/ktor-shared/ktor-serialization/ktor-serialization-jackson/jvm/src/JacksonConverter.kt
+++ b/ktor-shared/ktor-serialization/ktor-serialization-jackson/jvm/src/JacksonConverter.kt
@@ -18,7 +18,23 @@ import io.ktor.utils.io.jvm.javaio.*
 import kotlinx.coroutines.*
 import kotlin.text.Charsets
 
-public class JacksonConverter(private val objectmapper: ObjectMapper = jacksonObjectMapper()) : ContentConverter {
+/**
+ * A content converter that uses [Jackson]
+ *
+ * @param mapper a configured instance of [ObjectMapper]
+ * @param streamRequestBody if set to true, will stream request body, without keeping it whole in memory.
+ * This will set `Transfer-Encoding: chunked` header.
+ */
+public class JacksonConverter(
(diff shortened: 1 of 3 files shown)


ktor-shared/ktor-serialization/ktor-serialization-jackson/jvm/src/JacksonConverter.kt
```diff
@@ -41,27 +57,33 @@ public class JacksonConverter(private val objectmapper: ObjectMapper = jacksonOb
...
-                    InputStream instead of a writer to let Jackson do it's thing.
-                     */
-                    objectmapper.writeValue(this, value)
-                } else {
-                    objectmapper.writeValue(this.writer(charset = charset), value)
-                }
-            },
-            contentType.withCharsetIfNeeded(charset)
-        )
+        return when (streamRequestBody) {
```
e5l, line 60:
```kotlin
if (!streamRequestBody) {
    return TextContent(...)
}


return OutputStreamContent(...)
```

ktor-shared/ktor-serialization/ktor-serialization-jackson/jvm/src/JacksonConverter.kt
```diff
@@ -80,8 +102,27 @@ public class JacksonConverter(private val objectmapper: ObjectMapper = jacksonOb
...
+ * You can learn more from [Content negotiation and serialization](https://ktor.io/docs/serialization.html).
+ *
+ * @param contentType the content type to send with request
+ * @param streamRequestBody if set to true, will stream request body, without keeping it whole in memory.
+ * This will set `Transfer-Encoding: chunked` header.
+ * @param block a configuration block for [ObjectMapper]
+ */
 public fun Configuration.jackson(
     contentType: ContentType = ContentType.Application.Json,
+    streamRequestBody: Boolean = true,
```
e5l, line 125: Should we think about adding a configuration class instead? so In the future, we can add more properties without deprecations

e5l (changes requested):

e5l (approved):

PR #3177, opened 2022-10-03
KTOR-4925 Fix CIO tunneling.

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-4925

Adding test is complicated, because we need a real proxy server running on another host.


e5l (approved):
lgtm.

We have a regular TCP server in tests for this purpose, you can communicate text and assert messages directly

PR #3183, opened 2022-10-04
KTOR-4521 Add Logging for Compression Plugin

e5l:
Fix [KTOR-2521](https://youtrack.jetbrains.com/issue/KTOR-4521/Add-Debug-Logging-to-Compression-Plugin)

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-compression/jvm/src/io/ktor/server/plugins/compression/Compression.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-compression/jvm/src/io/ktor/server/plugins/compression/Compression.kt
@@ -17,7 +17,7 @@ import io.ktor.util.pipeline.*
 import io.ktor.utils.io.*
 import kotlinx.coroutines.*
 
-internal val LOGGER = KtorSimpleLogger("io.ktor.server.plugins.compression.ContentEncoding")
+internal val LOGGER = KtorSimpleLogger("io.ktor.server.plugins.compression.Compression")
 
 /**
  * The default minimal content size to compress.
--- a/ktor-server/ktor-server-plugins/ktor-server-compression/jvm/src/io/ktor/server/plugins/compression/Compression.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-compression/jvm/src/io/ktor/server/plugins/compression/Compression.kt
@@ -12,9 +12,12 @@ import io.ktor.server.request.*
 import io.ktor.server.response.*
 import io.ktor.util.*
 import io.ktor.util.cio.*
+import io.ktor.util.logging.*
 import io.ktor.util.pipeline.*
 import io.ktor.utils.io.*
 import kotlinx.coroutines.*
+
+internal val LOGGER = KtorSimpleLogger("io.ktor.server.plugins.compression.ContentEncoding")
 
 /**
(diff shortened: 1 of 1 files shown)


ktor-server/ktor-server-plugins/ktor-server-compression/jvm/src/io/ktor/server/plugins/compression/Compression.kt
```diff
@@ -12,10 +12,13 @@ import io.ktor.server.request.*
 import io.ktor.server.response.*
 import io.ktor.util.*
 import io.ktor.util.cio.*
+import io.ktor.util.logging.*
 import io.ktor.util.pipeline.*
 import io.ktor.utils.io.*
 import kotlinx.coroutines.*
 
+internal val LOGGER = KtorSimpleLogger("io.ktor.server.plugins.compression.ContentEncoding")
```
whyoleg, line 20: why it's `ContentEncoding` and not `Compression`?

e5l: It's a symbol name to lookup. Could you tell me if you think it's better to use Compression here?

whyoleg: As far as I see from other PRs on adding Logging, all of them follows the idea that this symbol is the same as name of plugin - which is really useful for understanding on from where it comes from.
F.e. if I would see logger name as `io.ktor.server.plugins.XXXpackage.YYYplugin` and see some issues there, it will be easy to find in which plugin I have an issue.

e5l: Sure, let me fix this

whyoleg: Just in case, what about adding `logger` property just inside plugin builder block?
So it will be super easy for plugin writters to do logs with plugin name out of the box somehow?

e5l: I'm thinking about this, but it can be confused with the application log in the environment. We also cannot obtain the package name there.

whyoleg: Yes, but what if we will provide fully qualified name for plugin name?
Yes it will reduce simplicity, but it can also improve on how attributes work.
I mean now if external plugin will have `Compression` name and is applied in the same scope as `ktor` `Compression` plugin - looks like on of them will replace another

e5l: I would leave the renaming for a separate task. Not sure if it should be solved along with logging. 
The logger name is fixed

whyoleg: yes, of course, sorry, it was just an idea while we are here, not a call to action

PR #3185, opened 2022-10-05
KTOR-4729 KTOR-4926 Disable auto reload in tests

rsinukov:
Using empty `watchPaths` disables creating special class loader. In the future, we may add a new property to enable the development mode, but disable auto reload explicitly. But it needs design on how `developmentMode`, `autoReload` and `watchPaths` properties should work together

https://youtrack.jetbrains.com/issue/KTOR-4729
https://youtrack.jetbrains.com/issue/KTOR-4926


e5l (approved):
lgtm

PR #3186, opened 2022-10-06
KTOR-4943 NettyHttp2ApplicationResponse should not expose pseudo headers

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-4943


e5l (approved):
lgtm

PR #3190, opened 2022-10-10
KTOR-4524 Add logging for Routing

e5l:
Fix [KTOR-4524](https://youtrack.jetbrains.com/issue/KTOR-4524/Add-Debug-Logging-to-Routing)

### DIFF
--- a/ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/routing/Routing.kt
+++ b/ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/routing/Routing.kt
@@ -10,10 +10,13 @@ import io.ktor.server.application.*
 import io.ktor.server.request.*
 import io.ktor.server.response.*
 import io.ktor.util.*
+import io.ktor.util.logging.*
 import io.ktor.util.pipeline.*
 
 @InternalAPI
 public val RoutingFailureStatusCode: AttributeKey<HttpStatusCode> = AttributeKey("RoutingFailureStatusCode")
+
+internal val LOGGER = KtorSimpleLogger("io.ktor.routing.Routing")
 
 /**
  * A root routing node of an [Application].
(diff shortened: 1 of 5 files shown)


whyoleg:
Just wanted to mention that on current moment we don't have possibility to filter logs on nix server. On JVM we have slf4j and so we can have debug/trace configuration overall and based on class/tag name.
So adding a lot of logs, even TRACE, will significantly reduce usability of nix server.

e5l:
@whyoleg we're going to add logs in 2.2.0 along with the `log level` option for native.

whyoleg:
@e5l Is there any info on how it will work?

e5l:
It will be possible to set with ENV variable or configuration file

ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/routing/RoutingResolveContext.kt
```diff
@@ -170,6 +175,7 @@ public class RoutingResolveContext(
         val finalResolve = resolveResult
 
         if (finalResolve.isEmpty()) {
+            LOGGER.trace("No route matched for ${call.request.uri}")
```
rsinukov, line 178: why do we need both logging and tracing in routing? can they be merged?

rsinukov (changes requested):

ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/routing/RoutingResolveContext.kt
rsinukov: Tracing has more information and is better structured. Maybe we can enable it automatically, if log level is trace?

e5l: Done!

PR #3195, opened 2022-10-11
KTOR-4954 CallFailed hook consumes exception in writing body

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-4954

And merging native and jvm `BaseApplicationResponse` into common, because there is no difference between the two


ktor-server/ktor-server-host-common/jvmAndNix/src/io/ktor/server/engine/BaseApplicationResponse.kt
```diff
@@ -1,63 +1,226 @@
...
+            try {
+                withContext(Dispatchers.IOBridge) {
+                    content.writeTo(this@use)
+                }
+            } catch (closed: ClosedWriteChannelException) {
+                throw ChannelWriteException(exception = closed)
+            }
+
+            // TODO currently we can't ensure length like that
+            // because a joined channel doesn't increment totalBytesWritten
```
e5l, line 181: could you please drop the comment?

e5l: nope, we likely shouldn't

e5l (approved):
lgtm

PR #3200, opened 2022-10-13
KTOR-3504 KTOR-2430 Fix WS timeout

rsinukov:
Timeout should send close frame and not wait for close acknowledge from the other side. Instead, directly close channels and provide close reason.

https://youtrack.jetbrains.com/issue/KTOR-2430
https://youtrack.jetbrains.com/issue/KTOR-3504


e5l (approved):
lgtm, good catch

PR #3211, opened 2022-10-18
KTOR-4993 Fix UrlBuilder origin in iframe

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-4993

Adding test is problematic, because bug is reproduced only inside iframe


e5l (approved):
lgtm

PR #3224, opened 2022-10-28
KTOR-5059 Digest auth: Support returning custom Principal

rsinukov:
And move some common tests from jvm only

https://youtrack.jetbrains.com/issue/KTOR-5059/Digest-auth-Support-returning-any-objects-which-implement-Principal-interface


e5l (approved):
lgtm

PR #3225, opened 2022-10-29
KTOR-3788 : Serialization json support kotlinx Flow

pull-vert:
**Subsystem**
Client/Server, all ktor-serialization modules

**Motivation**
Fix [KTOR-3788](https://youtrack.jetbrains.com/issue/KTOR-3788) that exposes more details.
As coroutines are heavilly used and supported in Ktor, I think ktor-serialization should support kotlinx.coroutines `Flow<T>` Serialization natively, as it is the official asynchronous Stream on the Kotlin coroutines world.
Following the same logic, I think deserialization to `Sequence<T>` should be suported too.

**Solution**
* Allow to serialize a `Flow<t>`
* Allow to deserialize to `Sequence<T>`

**Additional content**
* Bring some performance optimizations on JVM on Gson, Jackson and kotlinx-serialization that use `java.io.OutputStream`, `java.io.InputStream`, `java.io.Writer` and `java.io.Reader`.
* Add more Unit test for Serialization, for both UTF-8 and UTF-16


pull-vert:
Hello @rsinukov , I think I fixed the build errors and rebased from upstream.

I consider that this feature is finished and ready to be reviewed

ktor-shared/ktor-serialization/ktor-serialization-gson/jvm/src/GsonConverter.kt
```diff
@@ -8,18 +8,34 @@ import com.google.gson.*
...
+ * A content converter that uses [Gson]
+ *
+ * @param gson a configured instance of [Gson]
+ * @param streamRequestBody if set to true, will stream request body, without keeping it whole in memory.
+ * This will set `Transfer-Encoding: chunked` header.
  */
-public class GsonConverter(private val gson: Gson = Gson()) : ContentConverter {
+public class GsonConverter(
+    private val gson: Gson = Gson(),
+    private val streamRequestBody: Boolean = true
```
rsinukov, line 31: I have multiple concerns about this:
1. Changing default behaviour
2. This is app level config, applied to all responses, but is useful only for responses that return flow.

ktor-shared/ktor-serialization/ktor-serialization-gson/jvm/src/GsonConverter.kt
```diff
@@ -36,29 +52,84 @@ public class GsonConverter(private val gson: Gson = Gson()) : ContentConverter {
...
+    @OptIn(InternalAPI::class)
     override suspend fun deserialize(charset: Charset, typeInfo: TypeInfo, content: ByteReadChannel): Any? {
         if (gson.isExcluded(typeInfo.type)) {
             throw ExcludedTypeGsonException(typeInfo.type)
         }
 
         try {
             return withContext(Dispatchers.IO) {
+                // specific behavior for Sequence : collect it into a List
+                var isSequence = false
```
rsinukov, line 92: What's the benefit of `call.receive<Sequence<Type>>()` over `call.receive<List<Type>>().asSequence()` . Yes, it's a little bit shorter, but it brings a lot of very unintuitive complexities to the code. Maybe receiving it as a `Flow` would be worth it, to support streamed parsing. But with sequences, there are no benefits, as far as I can see.

ktor-shared/ktor-serialization/ktor-serialization-gson/jvm/src/GsonConverter.kt
```diff
@@ -36,29 +52,84 @@ public class GsonConverter(private val gson: Gson = Gson()) : ContentConverter {
...
+                    decoded
+                }
             }
         } catch (deserializeFailure: JsonSyntaxException) {
             throw JsonConvertException("Illegal json parameter found", deserializeFailure)
         }
     }
+
+    private companion object {
+        private const val beginArrayCharCode = '['.code
```
rsinukov, line 114: This for sure will break pretty print and maybe other configs.

rsinukov (changes requested):
Hi and thanks for the PR. I have some questions about it. I added them to gson converter, but they are relevant to the whole PR.

ktor-shared/ktor-serialization/ktor-serialization-gson/jvm/src/GsonConverter.kt
pull-vert: I add a new behavior based on OutputStreamContent for all types, not only Flow, in GsonConverter and in KotlinxSerializationJsonJvmConverter (that previously used only a TextContent without streaming).
So I adopted the exact same solution to allow the user to bypass streaming (`Transfer-Encoding: chunked`) : this new `streamRequestBody` parameter to mimic the recent change you made in JacksonConverter ([here](https://github.com/ktorio/ktor/commit/3595ddff009720f3e00a59fe59897839c75fbc12)).

ktor-shared/ktor-serialization/ktor-serialization-gson/jvm/src/GsonConverter.kt
pull-vert: Thias is true, I missed it !

I can add the pretty print support for Flow serialization (with a new Unit test that verifies this use-case)

ktor-shared/ktor-serialization/ktor-serialization-gson/jvm/src/GsonConverter.kt
rsinukov: I think it is ok to use native support where it exists and throw exceptions if the underlying library doesn't support it. Especially, if the alternative is doing some hacks with types.

ktor-shared/ktor-serialization/ktor-serialization-gson/jvm/src/GsonConverter.kt
pull-vert: > we should not change behaviour, unless it is a bug

The chunked response was not available at all with current Ktor + Gson or kotlinx-coroutines, I thought that changing the default behavior of Gson and kotlinx-serialization could be a nice improvement by making Jackson, Gson and kotlinx-serialization now all 3 consistent about chunked response inside Ktor.
I don't think a future user would understand why default chunked response is `true` with Jackson and `false` in Gson and kotlinx-serialization in Ktor, now that all 3 allow it.
But this would require a proper changelog to document this change, an this should only be included in let's say the upcoming 2.2.0 release, maybe not a minor bugfix release.

This is my opinion, but I will of course respect your decision and turn to `false` this param if you ask me to preserve the default behavior on Gson and kotlinx-serialization is better !

> Actually, why you need this flag completely here?

I think if Jackson provides this flag to turn off chunked response, then Gson and kotlinx-serialization have to provide it as well, that's why I added it in my last commit.

> What is the flag set to false and user respond with Flow?

Nice catch, I forgot to handle this case for Gson !
I can fix it and add a new unit test for chunked flag to `false` + Flow for Gson, Jackson and kotlinx-serialization.

ktor-shared/ktor-serialization/ktor-serialization-gson/jvm/src/GsonConverter.kt
pull-vert: > I think it is ok to use native support where it exists and throw exceptions if the underlying library doesn't support it. Especially, if the alternative is doing some hacks with types.

Noted, I will only keep the Sequence deserialization support in the kotlinx-serialization + UTF-8 case, and remove it everywhere else. This will indeed simplify a lot the code !

ktor-shared/ktor-serialization/ktor-serialization-gson/jvm/src/GsonConverter.kt
rsinukov: I prefer to keep defaults. There are way more people who bump versions and expect the same behavior than people who switch serialization libraries and expect the same behavior =)

Jackson converter has this flag for historical reasons, and other converters do not need to have it. So far, there have been no requests for it. What do you think about not having this flag at all and just use `chunked` where user respond with `Flow` and keep the rest of the behaviour as it is?

pull-vert: I understand and respect this choice, so I will remove chunked support (= the `streamRequestBody` parameter) for Gson and kotlinx-serialization, and as you suggest keep `chunked` only for the `Flow` scenario.

But from my external point of view, as a recent Ktor user, I don't see why in Ktor Jackson would be the only framework to provide `streamRequestBody` option whereas Gson and kotlinx-serialization can support it too.
For Gson and kotlinx-serialization we could just set the `streamRequestBody` parameter with a `false` default value to maintain backward behaviour.

rsinukov: We can do it, of course, but let's keep it a separate task. This PR is already big and complex enough.

pull-vert:
Hi @rsinukov ,
I made all the requested changes, you can check it again when you have some time :)

ktor-shared/ktor-serialization/ktor-serialization-gson/jvm/src/GsonConverter.kt
```diff
@@ -59,6 +84,33 @@ public class GsonConverter(private val gson: Gson = Gson()) : ContentConverter {
             throw JsonConvertException("Illegal json parameter found", deserializeFailure)
         }
     }
+
+    private val prettyPrinting by lazy {
+        val field = Gson::class.java.getDeclaredField("prettyPrinting")
```
rsinukov, line 89: This can easily break if users use other versions of `Gson`. As an alternative, you can wrap it into `try/catch` or maybe just ignore pretty print for flows?

ktor-shared/ktor-serialization/ktor-serialization-gson/jvm/test/ClientGsonTest.kt
```diff
@@ -20,7 +24,28 @@ class ClientGsonTest : AbstractClientContentNegotiationTest() {
         register(contentType, converter)
     }
 
+    @Test
+    fun testNotChunkedEncodingByDefault() = testWithEngine(CIO) {
```
rsinukov, line 28: this test is redundant now

ktor-shared/ktor-serialization/ktor-serialization-gson/jvm/src/GsonConverter.kt
```diff
@@ -59,6 +84,33 @@ public class GsonConverter(private val gson: Gson = Gson()) : ContentConverter {
...
+     * Guaranteed to be called inside a [Dispatchers.IO] context, see [OutputStreamContent]
+     */
+    @Suppress("BlockingMethodInNonBlockingContext")
+    private suspend fun <T> Flow<T>.serializeJson(writer: Writer) {
+        writer.write(beginArrayCharCode)
+        collectIndexed { index, value ->
+            if (index > 0) {
+                writer.write(objectSeparator)
+            }
+            gson.toJson(value, writer)
```
rsinukov, line 110: what do you think about flushing after you wrote value and separator, so that the receiver side can start processing the item?

ktor-shared/ktor-serialization/ktor-serialization-jackson/jvm/src/JacksonConverter.kt
```diff
@@ -95,6 +131,61 @@ public class JacksonConverter(
...
+        private const val objectSeparator = ','.code
+    }
+
+    /**
+     * Guaranteed to be called inside a [Dispatchers.IO] context, see [OutputStreamContent]
+     */
+    @Suppress("BlockingMethodInNonBlockingContext")
+    private suspend fun <T> Flow<T>.serializeJson(outputStream: OutputStream) {
+        val jfactory = JsonFactory()
+        // cannot use ObjectMapper write to Stream because it flushes the OutputStream
```
rsinukov, line 151: can you elaborate on flushing here and why is it bad?

ktor-shared/ktor-serialization/ktor-serialization-kotlinx/ktor-serialization-kotlinx-json/jvm/src/KotlinxSerializationJsonJvmConverter.kt
```diff
@@ -0,0 +1,166 @@
...
+/**
+ * Creates a JVM converter serializing with the specified string [format] and
+ * [defaultCharset] (optional, usually it is UTF-8).
+ */
+@OptIn(ExperimentalSerializationApi::class, InternalAPI::class)
+public class KotlinxSerializationJsonJvmConverter(
+    private val json: Json
+) : AbstractKotlinxSerializationConverter(json) {
+
+    override suspend fun serializeNullable(
```
rsinukov, line 34: can it be done through `expect/actual` instead of an abstract class and overriding?

rsinukov (changes requested):
This gets better and better! Thanks for your work.
Still some comments, though. I'm also not sure about pretty print. It's very counterintuitive that changing it can change response. I'm thinking that it can be safe just to break pretty print in case if users respond with `Flow`. What do you think?

ktor-shared/ktor-serialization/ktor-serialization-gson/jvm/src/GsonConverter.kt
pull-vert: > It's very counterintuitive that changing it can change response.

I agree with you, I chose to just ignore pretty print for the Flow case (this is mentioned in code comments keep track of it)

ktor-shared/ktor-serialization/ktor-serialization-jackson/jvm/src/JacksonConverter.kt
pull-vert: I changed it to a manual flush after each value + separator (or endChar for the last value) in Flow, as you suggested for Gson, and made this behavior consistent in all 3 tools : Gson, Jackson and kotlinx-serialization

ktor-shared/ktor-serialization/ktor-serialization-kotlinx/ktor-serialization-kotlinx-json/jvm/src/KotlinxSerializationJsonJvmConverter.kt
pull-vert: @rsinukov I chose to add a `AbstractKotlinxSerializationConverter` only to provide this util function that I need to call in `KotlinxSerializationJsonJvmConverter`
```kotlin
protected fun serializerFromTypeInfo(typeInfo: TypeInfo): KSerializer<*> =
        serializerFromTypeInfo(typeInfo, format.serializersModule)
```
Because I didn't want to have to make `serializerFromTypeInfo(typeInfo: TypeInfo, module: SerializersModule): KSerializer<*>` public : I do not like at all the idea to expose a new public "root" function !

I don't think `expect/actual` would allow me to solve this issue.

I can make `KotlinxSerializationConverter` open so that `KotlinxSerializationJsonJvmConverter` extends it, this would allow me to get rid of `AbstractKotlinxSerializationConverter` and remove the code below and just call "super" for fallback instead
```kotlin
private val fallbackConverter by lazy {
    KotlinxSerializationConverter(json)
}
```
Does this seem ok to you ?

rsinukov: Do I understand it right that you extract it to have special handling for JSON? What do you think about keeping all logic internal and check if a format is `Json` inside `KotlinxSerializationConverter` instead of creating a separate class for it? I haven't tried it, but I think it should work.
In the pseudocode, it will be something like this:
```
class KotlinxSerializationConverter {
   override fun serializeNullable(...) {
    if (format !is Json) doBaseSerialization(...)
    else serializeJson(...)
  }
}
```
Btw, serialization can be multiplatform, if instead of `OutputStreamContent` you would use `ChannelWriterContent`. Something like this (simplified):
```
internal fun serializeJson(...) {
  return ChannelWriterContent {
     channel.write("[")
     flow.collect { channel.write(it.serializeToString()); channel.write(","); channel.flush() }
     channel.write("]")
  }
}
``` 

Decoding to a sequence can't be multiplatform, but there you can use `expect/actual` functions
```
override fun deserialize(...) {
    if (format !is Json) doBaseDeserialization(...)
    else deserializeJson(...)
}

internal expect fun deserializeJson(...)

// in jvm 
internal actual fun deserializeJson() { /* do sequence deserialization if needed */ }
// in posix/js
internal actual fun deserializeJson() { /* do base deserialization */ }
```

Right now, with introducing of abstract classes, such a simple thing as converter gets really complex. But what's more important, it breaks binary compatibility, so it can't be merged earlier than 3.0.0, which is not going to be soon.
With the approach of internal check, you also don't need to split `JsonSupport` in between platforms.

rsinukov (changes requested):
Another round. Almost there! Thanks again for your work.

ktor-shared/ktor-serialization/ktor-serialization-kotlinx/ktor-serialization-kotlinx-json/jvm/src/KotlinxSerializationJsonJvmConverter.kt
pull-vert: > Do I understand it right that you extract it to have special handling for JSON?

No, the reason I added an abstract class was simply because I needed to access the method `serializerFromTypeInfo(typeInfo, format.serializersModule)` (currently internal)
 that is in **ktor-serialization-kotlinx** in my JVM specific JSON implementation in the other module **ktor-serialization-kotlinx-json**.

> Right now, with introducing of abstract classes, such a simple thing as converter gets really complex. But what's more important, it breaks binary compatibility, so it can't be merged earlier than 3.0.0, which is not going to be soon.
With the approach of internal check, you also don't need to split JsonSupport in between platforms.

Your suggestion cannot work for this same "between separate modules" problem : `KotlinxSerializationConverter` is in the general **ktor-serialization-kotlinx** module, that does not have a dependency on `kotlinx-serialization-json`, so I cannot make this statement `if (format !is Json)` (Json is not accessible), and cannot access all the optimized Flow / Sequence functions based on InputStream / OutputStream that are also in `kotlinx-serialization-json`.

This not breaking binary compatibility of **ktor-serialization-kotlinx** + my intention to offer Json optimised behavior that is only available from `kotlinx-serialization-json` seems pretty complex :disappointed: 
Maybe I should just offer `Flow<T>` streaming for Gson and Jackson, and stick to the naive Flow support for kotlinx-serialization as below ?
```
// specific behavior for kotlinx.coroutines.flow.Flow : collect it into a List
val resolvedValue = if (typeInfo.type == Flow::class) {
    (value as Flow<*>).toList()
} else {
    value
}
```

pull-vert: I could add a `compileOnly` dependency on `kotlinx-serialization-json` in **ktor-serialization-kotlinx** , and replace the `if (format is Json)` check with something like `if (format::class.qualifiedName == "kotlinx.serialization.json.Json")`, and then in this if case I could use `kotlinx-serialization-json`.

Or I can change this method signature, but this would be a binary compatibility change 
```kotlin
public fun Configuration.serialization(
    contentType: ContentType,
    format: StringFormat,
    isJson: Boolean = false // <- add this new parameter
)
```

pull-vert:
> Another round. Almost there! Thanks again for your work.

Thanks for your time on this review ! Sadly I am not entirely sure we are almost there anymore.

I replied to your comment on `kotlinx-serialization` binary compatibility issue, I think this is the GO / NO GO central point for this whole Pull request.

I will wait until we find (or not) a solution to this point before making the other requested changes.

ktor-shared/ktor-serialization/ktor-serialization-jackson/jvm/src/JacksonConverter.kt
```diff
@@ -95,6 +121,59 @@ public class JacksonConverter(
...
+    /**
+     * Guaranteed to be called inside a [Dispatchers.IO] context, see [OutputStreamContent]
+     */
+    @Suppress("BlockingMethodInNonBlockingContext")
+    private suspend fun <T> Flow<T>.serializeJson(writer: Writer) {
+        val jfactory = JsonFactory()
+        // cannot use ObjectMapper write to Stream because it flushes the OutputStream on each write
+        val jGenerator = jfactory.createGenerator(writer)
+        jGenerator.prettyPrinter = MinimalPrettyPrinter("") // avoid single space between items
+        jGenerator.configure(JsonGenerator.Feature.FLUSH_PASSED_TO_STREAM, false)
```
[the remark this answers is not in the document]
pull-vert: Because the normal behavior is to flush every time a Jackson `write*` that targets the IO Writer is called (= the value in the Flow), that's not what I want, I want to flush only after the separator is written.

ktor-shared/ktor-serialization/ktor-serialization-jackson/jvm/src/JacksonConverter.kt
```diff
@@ -95,6 +121,59 @@ public class JacksonConverter(
...
+        private const val objectSeparator = ','.code
+    }
+
+    /**
+     * Guaranteed to be called inside a [Dispatchers.IO] context, see [OutputStreamContent]
+     */
+    @Suppress("BlockingMethodInNonBlockingContext")
+    private suspend fun <T> Flow<T>.serializeJson(outputStream: OutputStream) {
+        val jfactory = JsonFactory()
+        // cannot use ObjectMapper write to Stream because it flushes the OutputStream on each write
```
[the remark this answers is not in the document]
pull-vert: Because the normal behavior of `ObjectMapper` is to flush every time a Jackson write* that targets the IO Writer or OutputStream is called (= the value in the Flow), that's not what I want, I want to flush only after the separator is written.

rsinukov:
Sorry for the delay, I had a sick leave. I'll get back to this PR this week.

ktor-shared/ktor-serialization/ktor-serialization-kotlinx/ktor-serialization-kotlinx-json/jvm/src/KotlinxSerializationJsonJvmConverter.kt
rsinukov: You are absolutely right about modules, and that it's impossible to check if the format is `Json` without `compileOnly` dependency. I missed this, sorry.
Unfortunately, breaking binary compatibility is not an option and we need to find a solution that keeps it. I will experiment with it this week.

pull-vert:
> Sorry for the delay, I had a sick leave. I'll get back to this PR this week.

Welcome back, I wish you a good recovery :)

rsinukov:
@pull-vert Sorry that it took me much longer than I expected. I think I've found a solution to keep binary compatibility. I pushed some changes, please take a look and say what you think about them.
0. Rebase our work on the latest main branch
1. Did some more unification of tests
2. Add extension points to `KotlinxSerializationConverter` to be able to add format-specific logic. 
3. Use this extension point to add handling of `Sequence` and `Flow` if json converter.

pull-vert:
@rsinukov thanks a lot for this great work !

I just pushed a simple commit to support non UTF-8 charsets for Flow serialization with kotlinx-serialization.
I think everything is looking nice :slightly_smiling_face:

ktor-shared/ktor-serialization/ktor-serialization-kotlinx/ktor-serialization-kotlinx-json/common/src/io/ktor/serialization/kotlinx/json/KotlinxSerializationJsonExtensions.kt
```diff
@@ -75,15 +75,22 @@ internal class KotlinxSerializationJsonExtensions(private val format: Json) : Ko
         charset: Charset,
         channel: ByteWriteChannel
     ) {
-        val jsonArraySymbols = jsonArraySymbolsMap.getOrPut(charset) { JsonArraySymbols(charset) }
+        // Use Big Endian charsets without byte-order mark (BOM)
+        val charsetWithoutBom = when(charset) {
+            Charsets.UTF_16 -> Charsets.UTF_16BE
```
rsinukov, line 80: I don't think that these charsets are supported on native =(

rsinukov (changes requested):

ktor-shared/ktor-serialization/ktor-serialization-jackson/jvm/test/ClientJacksonTest.kt
```diff
@@ -39,13 +39,6 @@ class ClientJacksonTest : AbstractClientContentNegotiationTest() {
                 ContentType.Application.Json
             )
         }
-        post("/headers") {
```
rsinukov, line 42: Oops, this should not be removed. Do you mind rolling back this change?

ktor-shared/ktor-serialization/ktor-serialization-kotlinx/ktor-serialization-kotlinx-json/common/src/io/ktor/serialization/kotlinx/json/KotlinxSerializationJsonExtensions.kt
pull-vert: Understood, I removed this commit !

pull-vert:
@rsinukov I made the 2 requested changes.

One suggestion : Should we make `serializeFlow` feature, that by nature only supports UTF-8, multiplatform with expect/actual, like `deserializeSequence` is today ?
JS and Native would use the exact current multiplatform implementation, based on `ByteWriteChannel`, but JVM could write to this ByteWriteChannel's `OutputStream` through kotlinx-serialization optimized function `Json.encodeToStream` ?
I think direct write in the OutputStream should be more efficient than the current serialization of each item of the Flow to String, then encode String to ByteArray and finally write these bytes to the ByteWriteChannel, especially for a Flow of large items.

rsinukov:
@pull-vert 
Thanks!
The problem with `OutputStream` is that it is blocking, while the rest of the framework is not. You introduce a lot of overhead trying to connect these two together. We are working on a new IO library, hopefully `kotlinx.serialization` will natively support it in the future.

pull-vert:
@rsinukov Thanks for your reponse.

I understand your point about `OutputStream` being blocking, but as serialization frameworks currently all only provide optimizations based on `OutputStream`, without benchmarks I am not able to be sure if the overhead of using a `OutputStream` inside Ktor is worse than the one introduced by serializing to a String and then converting the String to a ByteArray, especially for very large items.
This is just a point of view, but I understand your answer :)

I follow with a lot of attention what is done here in the `io-playground` branch, and if kotlinx-coroutines natively supports it when this IO with supend support will be available, then even the `Flow` serialization (fluently flushing item by item) could be added on the kotlinx-coroutines side, this would be really nice !

I think this branch is now ready, what do you think ? Should I squash all commits ?

rsinukov:
@pull-vert It's hard to write a single good benchmark because use cases are very different. Writing one huge JSON will be faster with `OutputStream`, but writing a lot of normal size JSON at the same time is faster when it doesn't block. 

I also think that it's ready. I'll squash commits before merging this PR.

pull-vert:
Thank you very much @rsinukov 

I learned a lot by writing the code for this PR, and from talking to you about all these subjects !

rsinukov (approved):

PR #3235, opened 2022-11-03
KTOR-5099 Fix memory leak in HttpRequestRetry plugin

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-5099


ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/HttpRequestLifecycle.kt
```diff
@@ -26,7 +26,7 @@ internal class HttpRequestLifecycle private constructor() {
 
         override fun install(plugin: HttpRequestLifecycle, scope: HttpClient) {
             scope.requestPipeline.intercept(HttpRequestPipeline.Before) {
-                val executionContext = Job(context.executionContext)
+                val executionContext = SupervisorJob(context.executionContext)
```
e5l, line 29: Please be aware that SupervisorJob still bypasses cancellation

e5l (commented):
lgtm

PR #3239, opened 2022-11-07
KTOR-5112 Add the ability to access the route inside a route-scoped plugin

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-5112


e5l (approved):
lgtm

PR #3240, opened 2022-11-08
KTOR-5143 Fix merging config lists

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-5143


ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/config/MergedApplicationConfig.kt
```diff
@@ -34,7 +34,8 @@ internal class MergedApplicationConfig(
         MergedApplicationConfig(first.config(path), second.config(path))
 
     override fun configList(path: String): List<ApplicationConfig> =
-        first.configList(path) + second.configList(path)
+        (if (firstKeys.contains(path)) first.configList(path) else emptyList()) +
```
e5l, line 37: Could you please make it with a block body and extractions? so it can be easier to parse

e5l (approved):
lgtm

PR #3242, opened 2022-11-10
KTOR-2579 Add file based http cache storage for client

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-2579

Everything with suffix `Legacy` is moving of old storage related things to new files.

### DIFF
--- a/ktor-client/ktor-client-core/jvm/src/io/ktor/client/plugins/cache/storage/FileCacheStorage.kt
+++ b/ktor-client/ktor-client-core/jvm/src/io/ktor/client/plugins/cache/storage/FileCacheStorage.kt
@@ -1,0 +1,144 @@
+/*
+ * Copyright 2014-2022 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.client.plugins.cache.storage
+
+import io.ktor.http.*
+import io.ktor.util.*
+import io.ktor.util.collections.*
+import io.ktor.util.date.*
+import io.ktor.utils.io.*
+import io.ktor.utils.io.core.*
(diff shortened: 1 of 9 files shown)


ktor-client/ktor-client-core/jvm/src/io/ktor/client/plugins/cache/storage/FileCacheStorage.kt
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    private fun key(url: Url) = hex(url.toString().encodeToByteArray())
+
+    private suspend fun writeCache(urlHex: String, caches: List<CachedResponseData>) = coroutineScope {
+        val mutex = mutexes.computeIfAbsent(urlHex) { Mutex() }
+        mutex.withLock {
+            val channel = ByteChannel()
+            File(directory, urlHex).outputStream().buffered().use { output ->
+                launch {
```
e5l, line 50: The file could be closed before the launch is finished if the `launch` block suspends. Could you please add a test with a large body and move the file `use` inside?

e5l (changes requested):
lgtm, please check the launch block

ktor-client/ktor-client-core/jvm/src/io/ktor/client/plugins/cache/storage/FileCacheStorage.kt
rsinukov: `channel.copyTo(output)` is outside of `launch` block and it should ensure that file should be open as long as the channel is open. The channel is closed at the end of `launch` block. Do I miss something here?

e5l (approved):
lgtm

PR #3243, opened 2022-11-11
KTOR-735 Support WinHTTP engine

### LINKED ISSUE #671: Support native windows HTTP client

dtretyakov:
This change brings support of new Windows engine built on top of WinHTTP library. It supports HTTP 1.x/2 and WebSocket protocols.

**Subsystem**
Client, native windows engine.

**Motivation**
Fixes #671 WinHTTP API allows using native HTTP client without need to include additional DLLs. Also, it allows using certificate stored in Windows Certificates store.

**Solution**

We’re using WinHTTP in asynchronous working mode. In such mode notifications about operation results are delivered via status callback. The library supports both HTTP request processing as well as WebSockets protocol.

In HTTP engine we are using the following abstractions:
* `WinHttpSession` - create once per HTTP engine and allows configuring common settings like timeouts.
* `WinHttpConnect` - represents connection to HTTP host. In our case it also stores list of callbacks for coroutine continuations.
* `WinHttpRequest` - allows writing/reading request and response messages.
* `WinHttpRequestProducer` - handles request body writing. Allows using chunked body encoding.
* `WinHttpWebSocket` - represents WebSocket session. It's created from WinHttpRequest by closing original request and keeping `WinHttpConnect`ion open.

WinHTTP allows specifying connect and socket (send/receive) timeouts. Request timeout is handled by corresponding Ktor plugin.

**Notes**:
WinHTTP is strict regarding `Content-Length` header validation in HTTP 1.x requests and requires using `Transfer-Encoding: chunked` request body if body length is unknown, e.g. when it's stream or channel based. In such cases we’re trying to use system encoding (modern Windows 10+ OS) or we're using fallback to manual chunks writing.

In case of failures we're converting error codes into human readable messages by using `FormatMessage` function.

### DIFF
--- a/buildSrc/src/main/kotlin/KotlinExtensions.kt
+++ b/buildSrc/src/main/kotlin/KotlinExtensions.kt
@@ -72,3 +72,13 @@ fun NamedDomainObjectContainer<KotlinSourceSet>.desktopTest(block: KotlinSourceS
     block(sourceSet)
 }
 
+fun NamedDomainObjectContainer<KotlinSourceSet>.windowsMain(block: KotlinSourceSet.() -> Unit) {
+    val sourceSet = findByName("windowsMain") ?: return
+    block(sourceSet)
+}
+
+fun NamedDomainObjectContainer<KotlinSourceSet>.windowsTest(block: KotlinSourceSet.() -> Unit) {
+    val sourceSet = findByName("windowsTest") ?: return
+    block(sourceSet)
+}
+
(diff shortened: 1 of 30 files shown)


dtretyakov:
Originally from https://github.com/ktorio/ktor/pull/3109

PR #3244, opened 2022-11-11
KTOR-5161 New client plugins API

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-5161


e5l (changes requested):
please add tests

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/api/ClientPluginBuilder.kt
```diff
@@ -0,0 +1,269 @@
...
+     */
+    public fun <HookHandler> on(
+        hook: ClientHook<HookHandler>,
+        handler: HookHandler
+    ) {
+        hooks.add(HookHandler(hook, handler))
+    }
+}
+
+private object RequestHook :
```
e5l, line 132: please consider extracting hooks to files with internal visitibility

e5l (approved):
lgtm

PR #3249, opened 2022-11-16
JWT: JWTPayloadHolder.getListClaim() throws NPE when specified claim is absent

Stexxe:
https://youtrack.jetbrains.com/issue/KTOR-5098

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-auth-jwt/jvm/src/io/ktor/server/auth/jwt/JWTAuth.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-auth-jwt/jvm/src/io/ktor/server/auth/jwt/JWTAuth.kt
@@ -8,13 +8,11 @@ import com.auth0.jwk.*
 import com.auth0.jwt.*
 import com.auth0.jwt.algorithms.*
 import com.auth0.jwt.exceptions.*
-import com.auth0.jwt.impl.*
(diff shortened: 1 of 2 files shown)


ktor-server/ktor-server-plugins/ktor-server-auth-jwt/jvm/src/io/ktor/server/auth/jwt/JWTAuth.kt
```diff
@@ -110,7 +108,8 @@ public abstract class JWTPayloadHolder(
      */
     public fun <T : Any> getListClaim(name: String, clazz: KClass<T>): List<T> {
         return try {
-            payload.getClaim(name).asList(clazz.javaObjectType)
+            val claim = payload.getClaim(name)
```
rsinukov, line 111: can it be simplified to `payload.getClaim(name).asList(...) ?: null`?

Stexxe: No, because `claim` is always not null.

rsinukov: But `asList` returns `null`, right?

rsinukov (approved):

PR #3250, opened 2022-11-16
Java client engine doesn't handle HttpTimeout.INFINITE_TIMEOUT_MS properly

Stexxe:
https://youtrack.jetbrains.com/issue/KTOR-2814

### DIFF
--- a/ktor-client/ktor-client-java/jvm/src/io/ktor/client/engine/java/JavaHttpEngine.kt
+++ b/ktor-client/ktor-client-java/jvm/src/io/ktor/client/engine/java/JavaHttpEngine.kt
@@ -14,6 +14,7 @@ import kotlinx.coroutines.*
 import java.net.*
 import java.net.http.*
 import java.time.*
+import java.time.temporal.*
 import java.util.concurrent.*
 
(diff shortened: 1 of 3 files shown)


ktor-client/ktor-client-java/jvm/src/io/ktor/client/engine/java/JavaHttpEngine.kt
```diff
@@ -79,7 +80,13 @@ public class JavaHttpEngine(override val config: JavaHttpConfig) : HttpClientEng
 
                 data.getCapabilityOrNull(HttpTimeout)?.let { timeoutAttribute ->
                     timeoutAttribute.connectTimeoutMillis?.let {
-                        connectTimeout(Duration.ofMillis(it))
+                        try {
+                            // Check that timeout end date as the number of milliseconds can fit Long type
+                            Instant.now().plus(it, ChronoUnit.MILLIS).toEpochMilli()
```
rsinukov, line 85: What's the benefit of checking it this way instead of just comparing with `INFINITE_TIMEOUT_MS`? Is it realistic that users will set timeout duration higher then `Long.MaxValue - now()`?

rsinukov (changes requested):

rsinukov (approved):

PR #3251, opened 2022-11-17
KTOR-5134 Fix setting custom NSURLSession in Darwin

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-5134

* Merge `DarwinResponseReaded` and `WebSocketDelegate` into `KtorNSURLSessionDelegate`
* Make `KtorNSURLSessionDelegate` public, so that users can register it in their custom `UrlSession`


e5l (approved):
please also check the CI

PR #3276, opened 2022-11-29
Update Gradle and internal Kotlin version

e5l:


buildSrc/settings.gradle.kts
```diff
@@ -4,11 +4,12 @@
 pluginManagement {
     val build_snapshot_train: String? by settings
     repositories {
-        maven("https://plugins.gradle.org/m2")
         if (build_snapshot_train?.toBoolean() == true) {
             mavenLocal()
         }
 
+        maven("https://plugins.gradle.org/m2")
```
hfhbd, line 11: I would move the pluginPortal after mavenCentral: https://github.com/gradle/gradle/issues/15406

PR #3280, opened 2022-11-30
KTOR-5006: Close WS when reading fragmented control frames

Sebmaster:
Fragmented control frames are explicitly disallowed by the websocket spec.

**Subsystem**
Server, Websockets

**Motivation**

Ktor's websocket implementation is currently pretty non-compliant. We see a lot of "unknown opcode" errors in our logs, so I'm going on an endevaour to start fixing all the protocol violations. This one specifically fixes https://youtrack.jetbrains.com/issue/KTOR-5006/Websockets-Connection-should-failed-immediately-1002-Protocol-Error-when-control-frame-has-a-payload-with-more-than-125-octets.

**Solution**

For this initial PR we just need to add some Frame validation before forwarding the Frame to the queues. In an earlier version I had the validation directly in the frame, however this'd prevent testing this solution within ktor since invalid frames could not be constructed.

I'm not sure how to add a test for this. @Stexxe had a great standalone test in the issue, but I can't find anything comparable within the Websocket package. Is it "allowed" to create a whole Ktor server in there, or is it better to test this more granularly?

I'm a little uncertain about my choice of `IllegalStateException` as exception type. Let me know if there's a better fit/we should create a new exception type. There'll be a couple more validation like this - maybe create a `ProtocolViolationException` as an aggregator (maybe with an attached close code for unified handling within the framework)?

### DIFF
--- a/ktor-shared/ktor-websockets/jvm/test/WebSocketReaderTest.kt
+++ b/ktor-shared/ktor-websockets/jvm/test/WebSocketReaderTest.kt
@@ -1,0 +1,33 @@
+/*
+ * Copyright 2014-2022 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.tests.websocket
+
+import io.ktor.utils.io.*
+import io.ktor.utils.io.core.*
+import io.ktor.websocket.*
+import kotlinx.coroutines.*
+import kotlin.test.*
+
+class WebSocketReaderTest {
+    @Test
+    fun testFailFragmentedControl() {
+        val serializer = Serializer()
+        val channel = ByteChannel(true)
+        runBlocking {
+            val reader = WebSocketReader(channel, coroutineContext, 1000)
+            val frame = Frame.Ping(ByteArray(126))
+            serializer.enqueue(frame)
+
+            while (serializer.hasOutstandingBytes) {
+                channel.write {
+                    serializer.serialize(it)
+                }
+            }
+            assertFailsWith<ProtocolViolationException> {
+                reader.incoming.receive()
+            }
+        }
+    }
+}
--- a/ktor-shared/ktor-websockets/api/ktor-websockets.api
+++ b/ktor-shared/ktor-websockets/api/ktor-websockets.api
@@ -171,6 +171,14 @@ public final class io/ktor/websocket/FrameType$Companion {
 	public final fun get (I)Lio/ktor/websocket/FrameType;
 }
 
+public final class io/ktor/websocket/ProtocolViolationException : java/lang/Exception, kotlinx/coroutines/CopyableThrowable {
+	public fun <init> (Ljava/lang/String;)V
+	public fun createCopy ()Lio/ktor/websocket/ProtocolViolationException;
+	public synthetic fun createCopy ()Ljava/lang/Throwable;
+	public fun getMessage ()Ljava/lang/String;
+	public final fun getViolation ()Ljava/lang/String;
+}
+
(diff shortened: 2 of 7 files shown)


ktor-shared/ktor-websockets/jvm/test/WebSocketReaderTest.kt
```diff
@@ -0,0 +1,33 @@
...
+    @Test
+    fun testFailFragmentedControl() {
+        val serializer = Serializer()
+        val channel = ByteChannel(true)
+        runBlocking {
+            val reader = WebSocketReader(channel, coroutineContext, 1000)
+            val frame = Frame.Ping(ByteArray(126))
+            serializer.enqueue(frame)
+
+            while (serializer.hasOutstandingBytes) {
```
Sebmaster, line 23: Would be nice to launch {} this (since it somewhat relies on large buffer sizes right now), however there's some packaging weirdness around that and I'm not able to get that working.

whyoleg:
Please keep in mind, that on current moment, there are 2 implementations of `RawWebSocket`:
* JVM-only - which uses `WebSocketReader`, and affected by this PR
* Common (works for Native/JS and used in CIO) - which is not affected by this PR

Sebmaster:
@whyoleg Makes sense - found the other part of the code. Should we duplicate these validations there or is there some guidance on better code sharing between these classes?

rsinukov:
I think adding validation to a `Frame` is a good idea. And I prefer to have specific `ProtocolViolationException`.  `IllegalStateException` is to broad and happen due to logic error.

Sebmaster:
> I think adding validation to a Frame is a good idea.

I looked at that, but I think that may be too late in the process to get access to all the input data that's needed. However I did extend the implementation to non-JVM in `RawWebSocket`.

> And I prefer to have specific ProtocolViolationException. 

Done.

Sebmaster:
> LGTM, but can you please run ./gradlew apiDump and commit changes, please?

Done and moved the length check.

ktor-shared/ktor-websockets/jvm/test/WebSocketReaderTest.kt
```diff
@@ -0,0 +1,33 @@
...
+
+import io.ktor.utils.io.*
+import io.ktor.utils.io.core.*
+import io.ktor.websocket.*
+import kotlinx.coroutines.*
+import kotlin.test.*
+
+class WebSocketReaderTest {
+    @Test
+    fun testFailFragmentedControl() {
```
rsinukov, line 15: can you also add a test for non final control frame, please?

rsinukov (approved):
LGTM.
Unfortunately, we missed 2.2.0 release, where we could add a new API. I will merge it to `2.3.0-eap` version.

ktor-shared/ktor-websockets/jvm/test/WebSocketReaderTest.kt
Sebmaster: Unfortunately I'm not sure that's actually possible due to `Frame`'s design. I can't create raw frame or subclass it due to `sealed` and the existing control frames don't allow me to set `fin`. Any advice how to make that work?

rsinukov: You are right. It would be possible to create a serialized representation of the frame and send it to parser. But not worth it.

rsinukov: You are right. It would be possible to create a serialized representation of the frame and send it to parser. But not worth it.

PR #3282, opened 2022-12-01
KTOR-5225 Add callbacks to save application state for OAuth2

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-5225

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-auth/jvm/src/io/ktor/server/auth/OAuth.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-auth/jvm/src/io/ktor/server/auth/OAuth.kt
@@ -84,7 +84,8 @@ public sealed class OAuthServerSettings(public val name: String, public val vers
         public val passParamsInURL: Boolean = false,
         public val extraAuthParameters: List<Pair<String, String>> = emptyList(),
         public val extraTokenParameters: List<Pair<String, String>> = emptyList(),
-        public val accessTokenInterceptor: HttpRequestBuilder.() -> Unit = {}
+        public val accessTokenInterceptor: HttpRequestBuilder.() -> Unit = {},
+        public val onStateCreated: suspend (call: ApplicationCall, state: String) -> Unit = { _, _ -> }
     ) : OAuthServerSettings(name, OAuthVersion.V20) {
 
         @Deprecated("This constructor will be removed", level = DeprecationLevel.HIDDEN)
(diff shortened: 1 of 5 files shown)


ktor-server/ktor-server-plugins/ktor-server-auth/jvm/src/io/ktor/server/auth/OAuth.kt
```diff
@@ -161,15 +196,31 @@ public sealed class OAuthAccessTokenResponse : Principal {
...
         val accessToken: String,
         val tokenType: String,
         val expiresIn: Long,
         val refreshToken: String?,
-        val extraParameters: Parameters = Parameters.Empty
-    ) : OAuthAccessTokenResponse()
+        val extraParameters: Parameters = Parameters.Empty,
+    ) : OAuthAccessTokenResponse() {
+
+        public var state: String? = null
```
whyoleg, line 210: just a question: is it expected, that while `OAuth2` is `data class` this **new** property will not appear in `copy`/`componentN`/`equals`/`hashCode`?

ktor-server/ktor-server-plugins/ktor-server-auth/jvm/src/io/ktor/server/auth/OAuth.kt
```diff
@@ -117,6 +118,40 @@ public sealed class OAuthServerSettings(public val name: String, public val vers
             emptyList(),
             accessTokenInterceptor
         )
+
+        @Deprecated("This constructor will be removed", level = DeprecationLevel.HIDDEN)
```
whyoleg, line 122: minor: "This constructor will be removed" -> "Binary compatibility with 2.x"?

e5l (approved):
lgtm

ktor-server/ktor-server-plugins/ktor-server-auth/jvm/src/io/ktor/server/auth/OAuth.kt
rsinukov: It's really difficult to add it without breaking binary compatibility. In practice, I don't expect users to do `copy`/`componentN` on principals. And for equals and hashcode, having `accessToken` there should be enough.

ktor-server/ktor-server-plugins/ktor-server-auth/jvm/src/io/ktor/server/auth/OAuth.kt
rsinukov: IMO, deprecation messages should be for users explaining why this is deprecated, not for us explaining why it is kept.

whyoleg: AFAIK, declarations with `level=HIDDEN` will be not visible to user, until they open source code. They even should feel this change at all, as new parameter has default value, but ok.

rsinukov: Yes, but for consistency, I prefer it this way.

PR #3283, opened 2022-12-02
Build scripts: Bump the toolchain version to 11

martinbonnin:
Looks like Adoptium did not release Java8 versions of their M1 JDKs leading to errors like below:

```
> Unable to download toolchain matching the requirements ({languageVersion=8, vendor=any, implementation=vendor-specific}) from 'https://api.adoptium.net/v3/binary/latest/8/ga/mac/aarch64/jdk/hotspot/normal/eclipse'.
  > Could not read 'https://api.adoptium.net/v3/binary/latest/8/ga/mac/aarch64/jdk/hotspot/normal/eclipse' as it does not exist.
```

Bump to Java 11 which exists. This fixes building on a M1 mac without JDK8 installed. Given this is used to compile buildSrc, it shouldn't impact the published artifacts. 

Tests fail on my `:ktor-client:ktor-client-tests:iosSimulatorArm64TestCompileKlibraries` but this fails no matter if I have a JDK8 installed or not so it seems unrelated ?

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -176,15 +176,23 @@ fun configureDokka() {
 configureDokka()
 
 fun Project.setupJvmToolchain() {
-    val jdk = when (project.name) {
-        in jdk11Modules -> 11
-        else -> 8
-    }
-
     kotlin {
         jvmToolchain {
             check(this is JavaToolchainSpec)
-            languageVersion.set(JavaLanguageVersion.of(jdk))
+            languageVersion.set(JavaLanguageVersion.of(11))
+        }
+    }
+
+    tasks.withType(KotlinCompile::class.java).configureEach {
+        kotlinOptions {
+            when (this) {
+                is KotlinJvmOptions -> {
+                    jvmTarget = when (project.name) {
+                        in jdk11Modules -> "11"
+                        else -> "1.8"
+                    }
+                }
+            }
         }
     }
 }
--- a/buildSrc/build.gradle.kts
+++ b/buildSrc/build.gradle.kts
@@ -48,6 +48,6 @@ dependencies {
 kotlin {
     jvmToolchain {
         check(this is JavaToolchainSpec)
-        languageVersion.set(JavaLanguageVersion.of(8))
+        languageVersion.set(JavaLanguageVersion.of(11))
     }
 }


e5l (changes requested):
We're using `8` jdk to avoid problems with `ByteBuffer` overloads with primitive types: https://stackoverflow.com/questions/61267495/exception-in-thread-main-java-lang-nosuchmethoderror-java-nio-bytebuffer-flip. 

To fix this we also need to specify -Xjdk-release

martinbonnin:
> We're using 8 jdk to avoid problems with ByteBuffer overloads with primitive types

I agree in general but this change is for `buildSrc` so it shouldn't affect production code? And mandating Java11 for development sounds reasonnable these days?

e5l:
It does affect the toolchain used to compile classes for release

martinbonnin:
I'm confused. This is `buildSrc`. It generates the build logic classes, not the published ktor classes. Since it applies the `kotlin-dsl`, [it will default to 1.8](https://github.com/gradle/gradle/blob/3fed7d8caceaa13474c99e21b9cb113cac660697/subprojects/kotlin-dsl-plugins/src/main/kotlin/org/gradle/kotlin/dsl/plugins/dsl/KotlinDslPluginOptions.kt#L40).

You can double check by checking the class file version:

```
$ file buildSrc/build/classes/kotlin/main/CodestyleKt.class                                            
buildSrc/build/classes/kotlin/main/CodestyleKt.class: compiled Java class data, version 52.0 (Java 1.8)
```

Even if it were generating Java11 bytecode (which is possible with `kotlinDslPluginOptions.jvmTarget.set("11")` in `buildSrc/build.gradle.kts`), the released JVM artifacts would still generate 1.8 bytecode because [other parts of the main build change the toolchain](https://github.com/ktorio/ktor/blob/4f5954c2b9f5e5e20ea75e7e12454e4cb8c43e25/build.gradle.kts#L178). You can double check that too:

```
$ file ktor-client/ktor-client-okhttp/build/classes/kotlin/jvm/main/io/ktor/client/engine/okhttp/OkHttpEngine.class 
ktor-client/ktor-client-okhttp/build/classes/kotlin/jvm/main/io/ktor/client/engine/okhttp/OkHttpEngine.class: compiled Java class data, version 52.0 (Java 1.8)
```

Do you have a specific artifact/class in mind that should be 1.8 bytecode?

martinbonnin:
> other parts of the main build change the toolchain

Alright, I just realized I need to change it these places too then. Not sure why I was the impression that it was working there already. I'll update this PR in a bit.

martinbonnin:
Should be all good now, let me know what you think!

hbmartin:
Is this what's blocking iosarm64 EAP artifacts after the Kotlin 1.8.10 upgrade?

PR #3284, opened 2022-12-02
Build scripts: bump Gradle to 7.6

martinbonnin:
This fixes compiling with Java 19


martinbonnin:
Never mind, I just saw this is taken care of in https://github.com/ktorio/ktor/pull/3276. Sorry for the noise.

martinbonnin:
I changed my mind. Gradle version should be updateable independently from Kotlin and there's no need to gate Gradle 7.6 (with Java 19 support) on Kotlin updates (that sound like something more involved)

hfhbd:
Please also update the wrapper jar. (Gradle requires calling the upgrade task twice, first updating the properties files, and the second run actually updates the wrapper file)

martinbonnin:
Right 👍 Gradle wrapper should be updated now

e5l:
fixed with 2.3.0

PR #3285, opened 2022-12-05
KTOR-5252 Fix EOFException in read and readUtf8Line

e5l:


Stexxe (approved):

ktor-io/jvm/src/io/ktor/utils/io/ByteBufferChannel.kt
```diff
@@ -1631,6 +1631,7 @@ internal open class ByteBufferChannel(
 
     override suspend fun read(min: Int, consumer: (ByteBuffer) -> Unit) {
         require(min >= 0) { "min should be positive or zero" }
+        if (min == 0) return
```
whyoleg, line 1634: is it really correct?
f.e. if we call it with `min=0` and we have 5 bytes in channel, I think, that those should be consumed, no?
And in case there will be no bytes in channel I don't know what is expected behaviour - may be return or call consumer with empty buffer, any way, no suspension should happen to await more bytes.
So in my opinion, it should work like non-suspend `readAvailable`, no?

ktor-io/jvm/src/io/ktor/utils/io/ByteBufferChannel.kt
```diff
@@ -1949,41 +1950,44 @@ internal open class ByteBufferChannel(
...
-                consumed += decoded
+                    consumed += decoded
 
-                if (limit != Int.MAX_VALUE && consumed >= limit && !newLine) {
-                    throw TooLongLineException("Line is longer than limit")
+                    if (limit != Int.MAX_VALUE && consumed >= limit && !newLine) {
+                        throw TooLongLineException("Line is longer than limit")
+                    }
                 }
+            } catch (_: EOFException) {
```
whyoleg, line 1990: Im not an expert of course, but I see this as something easy to break - is it possible to avoid it? or at least add a comment, on why it's ignored?

ktor-io/jvm/src/io/ktor/utils/io/ByteBufferChannel.kt
```diff
@@ -2325,19 +2329,23 @@ internal open class ByteBufferChannel(
...
+                    val oldLimit = view.limit()
+                    val canCopyToDestination = minOf(max, destination.size - destinationOffset)
+                    val newLimit = minOf(view.limit().toLong(), canCopyToDestination + offset)
+                    view.limit(newLimit.toInt())
+                    bytesCopied = view.remaining()
+                    view.copyTo(destination, destinationOffset.toInt())
+                    view.limit(oldLimit)
+                }
             }
+        } catch (_: EOFException) {
```
whyoleg, line 2347: same here

ktor-io/jvm/src/io/ktor/utils/io/ByteBufferChannel.kt
e5l: Thanks, will check

ktor-io/jvm/src/io/ktor/utils/io/ByteBufferChannel.kt
e5l: sure, it's ignored by the contract of the method

rsinukov (approved):

PR #3287, opened 2022-12-06
KTOR-581 Implement getAllRoutes method

marychatte:
**Subsystem**
Server

**Motivation**
[KTOR-581](https://youtrack.jetbrains.com/issue/KTOR-581/List-of-all-routes)

**Solution**
dfs with adding all routes that have handlers

### DIFF
--- a/ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/routing/Route.kt
+++ b/ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/routing/Route.kt
@@ -133,3 +133,19 @@ public open class Route(
         }
     }
 }
+
+/**
+ * Return list of endpoints with handlers under this route.
+ */
+public fun Route.getAllRoutes(): List<Route> {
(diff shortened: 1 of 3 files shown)


ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/routing/Route.kt
```diff
@@ -133,3 +133,19 @@ public open class Route(
         }
     }
 }
+
+/**
+ * Return list of endpoints under this route.
```
rsinukov, line 138: Maybe `Return list of endpoints with handlers under this route.`?

rsinukov (approved):

ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/routing/Route.kt
marychatte: yes, done

PR #3308, opened 2022-12-09
KTOR-5311 Fix font content type and gzip

marychatte:
**Subsystem**
Server

**Motivation**
Bug

**Solution**
Fix and add [Font types](https://www.iana.org/assignments/media-types/media-types.xhtml#font)


e5l (approved):
lgtm

PR #3309, opened 2022-12-09
Fix SwaggerUI and OpenAPI endpoints issues

e5l:


marychatte (approved):
lgtm, but jvm tests failure seems to be related

e5l:
thanks, will check

ktor-server/ktor-server-plugins/ktor-server-openapi/jvm/src/io/ktor/server/plugins/openapi/OpenAPI.kt
```diff
@@ -49,13 +50,25 @@ public fun Routing.openAPI(
...
+    val file = File(swaggerFile)
     if (!file.exists()) {
         throw FileNotFoundException("Swagger file not found: $swaggerFile")
     }
 
-    return file
+    return file.readText()
+}
+
+@Deprecated( "Replaced with the extension on [Route]", level = DeprecationLevel.HIDDEN)
```
rsinukov, line 67: please remove extra space

rsinukov:
Please change base branch to 2.3.0

ktor-server/ktor-server-plugins/ktor-server-openapi/jvm/src/io/ktor/server/plugins/openapi/OpenAPI.kt
```diff
@@ -49,13 +50,25 @@ public fun Routing.openAPI(
     }
 }
 
-internal fun Routing.resolveOpenAPIFile(swaggerFile: String): File {
-    val resource = application.environment.classLoader.getResource(swaggerFile)
-    val file = if (resource != null) File(resource.toURI()) else File(swaggerFile)
+internal fun Route.resolveOpenAPIFileContent(swaggerFile: String): String {
```
Stexxe, line 53: The extension of the `Route` class is misleading because this method only reads a resource or a file and has nothing to do with routing. This class is only needed to get a class loader.

ktor-server/ktor-server-plugins/ktor-server-openapi/jvm/src/io/ktor/server/plugins/openapi/OpenAPI.kt
```diff
@@ -19,16 +19,17 @@ import java.io.*
...
  * See [OpenAPIConfig] for more details.
  */
-public fun Routing.openAPI(
+public fun Route.openAPI(
     path: String,
     swaggerFile: String = "openapi/documentation.yaml",
     block: OpenAPIConfig.() -> Unit = {}
 ) {
-    val file = resolveOpenAPIFile(swaggerFile)
+    val api = resolveOpenAPIFileContent(swaggerFile)
```
Stexxe, line 27: In the OpenAPI specification, they call it a document, so `openAPIDocument` or `document` would be a better name for the variable.

ktor-server/ktor-server-plugins/ktor-server-openapi/jvm/src/io/ktor/server/plugins/openapi/OpenAPI.kt
Stexxe: I'm not sure that resolving means looking at different locations and reading.

e5l: ok, let me name it `readOpenAPIFile` instead

PR #3313, opened 2022-12-12
KTOR-4547 Add Apache 5 client engine

rsinukov:


e5l (changes requested):
Could you add this module as the dependency to `ktor-client-tests`?

e5l:
We also can setup some http2 tests with a server

rsinukov:
Hm, strange, it was in dependencies and I even fixed some tests there. Probably rebase issue.

PR #3315, opened 2022-12-12
KTOR-5300 Disable body length check for JS Browser target

rsinukov:
Some browsers do not expose `Content-Encoding` header 
https://youtrack.jetbrains.com/issue/KTOR-5300


marychatte (approved):
lgtm

PR #3317, opened 2022-12-13
KTOR-5014/KTOR-5018 Fix treament of improper continuation frames

Sebmaster:
**Subsystem**
Server, websockets

**Motivation**

If a continuation is started (Fin=false), a future continuation must send Opcode=0. Previously we accepted frames with a regular opcode, even when they were of a completely different type (never finishing the original continuation)!
    
This also fixes the case where we "resume" an already finished frame (via Opcode=0 some time after a Fin=true data frame) which is also disallowed.

See also:
* https://youtrack.jetbrains.com/issue/KTOR-5014/Websockets-Connection-should-be-failed-immediately-since-all-data-frames-after-the-initial-data-frame-must-have-opcode-0
* https://youtrack.jetbrains.com/issue/KTOR-5018/Websockets-connection-should-be-failed-immediately-when-no-continuation-frame-goes-after-a-fragmented-text-frame

**Solution**

We essentially just do a bit more checking on the currently tracked state and see if it aligns with the incoming frame data. When it doesn't we throw the our previously introduced `ProtocolViolationException` which already returns the right status code.


ktor-shared/ktor-websockets/jvm/test/WebSocketReaderTest.kt
```diff
@@ -11,21 +11,43 @@ import kotlinx.coroutines.*
...
+        runBlocking {
+            val reader = WebSocketReader(channel, coroutineContext, 1000)
+            writeFrame(channel, Frame.Text(false, "Foo".toByteArray()))
+            writeFrame(channel, Frame.Ping("Ping!".toByteArray()))
+            writeFrame(channel, Frame.Text(true, "Foo".toByteArray()))
+            assertIs<Frame.Text>(reader.incoming.receive())
+            assertIs<Frame.Ping>(reader.incoming.receive())
+
             assertFailsWith<ProtocolViolationException> {
+                // serializer will re-send the opcode
```
rsinukov, line 50: Should we fix this and make sure that opcode is 0 for non final frames?
And can you please add a success path test for handling control frame in the middle of fragmented ones?

rsinukov (changes requested):
Thanks for the PR!
Some minor comments.

ktor-shared/ktor-websockets/jvm/test/WebSocketReaderTest.kt
Sebmaster: > Should we fix this and make sure that opcode is 0 for non final frames?

That's not quite right. The opcode has to be 0 for frames _following_ non-final frames. This means that we'd need to add additional state tracking to the serializer (similar to the parser). I think the serializer currently doesn't handle non-fin frames correctly.

>  And can you please add a success path test for handling control frame in the middle of fragmented ones?

That's the same issue. Since the Serializer can't/doesn't handle fragmentation, there's currently not a great way to intersperse control frames in between fragmented frames. Most of the testing I've done with this has been with the Autobahn test suite instead.

rsinukov:
I see failing tests in `io.ktor.tests.websocket.WebSocketTest`. Can you look at it, please?

Sebmaster:
> I see failing tests in io.ktor.tests.websocket.WebSocketTest. Can you look at it, please?

Fixed one of the tests, the others are running into the Serializer doing improper things issue, so I think we'll have to get Serializer do the state tracking after all. I'll try to add that as well.

Sebmaster:
I added continuation-frame awareness to serializer, however wanted to bring up a completely different idea:

Could we introduce a FrameType.Continuation instead? I think that'd break backwards compat for RawWebsocket usage, but I think we could hide it well enough from higher level usage since we'll concat the frames together in FrameCollector anyways.

This'd allow much better representation of incoming and outgoing frames, both in Serializer as well as in WebsocketReader and properly expose the socket behavior up the stack, instead of being weirdly hidden inside of FrameParser.

rsinukov (approved):
LGTM

PR #3321, opened 2022-12-15
KTOR-5337 Conflict between `ContentNegotiation` and  template plugins

marychatte:
**Subsystem**
Server

**Motivation**
Bug

**Solution**
Invoke template plugins (`Mustache`, `Jte` etc) in the new `Content` phase which is before `ContentNegotiation`

### DIFF
--- a/ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/application/KtorCallContexts.kt
+++ b/ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/application/KtorCallContexts.kt
@@ -86,11 +86,3 @@ public class OnCallRespondContext<PluginConfig : Any> internal constructor(
         context.subject = transformContext.transform(context.subject)
     }
 }
-
-public suspend fun <T : Any> PipelineContext<T, ApplicationCall>.transformBody(
-    transform: suspend TransformBodyContext.(body: Any) -> T
-) {
-    val transformContext = TransformBodyContext(call.response.responseType)
-
-    subject = transformContext.transform(subject)
-}
--- a/ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/application/KtorCallContexts.kt
+++ b/ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/application/KtorCallContexts.kt
@@ -86,3 +86,11 @@ public class OnCallRespondContext<PluginConfig : Any> internal constructor(
         context.subject = transformContext.transform(context.subject)
     }
 }
+
+public suspend fun <T : Any> PipelineContext<T, ApplicationCall>.transformBody(
+    transform: suspend TransformBodyContext.(body: Any) -> T
+) {
+    val transformContext = TransformBodyContext(call.response.responseType)
+
+    subject = transformContext.transform(subject)
+}
--- a/ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/application/hooks/CommonHooks.kt
+++ b/ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/application/hooks/CommonHooks.kt
@@ -135,7 +135,7 @@ public object ReceiveRequestBytes : Hook<(call: ApplicationCall, body: ByteReadC
  * Useful for some plugins which used for templates as views within application.
  */
 @InternalAPI
-public class ResponseBeforeTransform<T : Any>(private val clazz: KClass<T>) :
+public class BeforeResponseTransform<T : Any>(private val clazz: KClass<T>) :
     Hook<suspend (call: ApplicationCall, body: T) -> Any> {
     override fun install(
         pipeline: ApplicationCallPipeline,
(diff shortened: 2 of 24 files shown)


ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/response/ApplicationSendPipeline.kt
```diff
@@ -12,7 +12,9 @@ import io.ktor.util.pipeline.*
  */
 public open class ApplicationSendPipeline(
     override val developmentMode: Boolean = false
-) : Pipeline<Any, ApplicationCall>(Before, Transform, Render, ContentEncoding, TransferEncoding, After, Engine) {
+) : Pipeline<Any, ApplicationCall>(
+    Before, Content, Transform, Render, ContentEncoding, TransferEncoding, After, Engine
```
rsinukov, line 16: No need to create a new public phase. You can create it locally and use `pipeline.insertPhaseBefore(...)` API inside your hook.

rsinukov (changes requested):

ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/response/ApplicationSendPipeline.kt
marychatte: done, thanks

ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/application/hooks/CommonHooks.kt
```diff
@@ -127,3 +127,20 @@ public object ReceiveRequestBytes : Hook<(call: ApplicationCall, body: ByteReadC
         }
     }
 }
+
+/**
+ * A hook that is invoked before `Transform` phase.
+ * Useful for some plugins which used for templates as views within application.
+ */
+public object ResponseBeforeTransform : Hook<suspend PipelineContext<Any, ApplicationCall>.(body: Any) -> Any> {
```
rsinukov, line 135: a couple of things
1. One of the goals to design new plugins API was to hide all the `Pipeline` and `PipelineContext` related entities from the users. This hook exposes them.
2. I am not sure if it should be public even. I'm ok with duplicating it in every related plugin to keep in internal. If you decide to keep it public, we need a better name for it.

rsinukov (changes requested):

ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/application/hooks/CommonHooks.kt
marychatte: okay, I think it's better not to expose `Pipeline` and `PipelineContext`, I'll rewrite it

ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/application/KtorCallContexts.kt
```diff
@@ -86,3 +86,11 @@ public class OnCallRespondContext<PluginConfig : Any> internal constructor(
         context.subject = transformContext.transform(context.subject)
     }
 }
+
+public suspend fun <T : Any> PipelineContext<T, ApplicationCall>.transformBody(
```
rsinukov, line 90: does this function needs to be public?

ktor-server/ktor-server-plugins/ktor-server-freemarker/jvm/src/io/ktor/server/freemarker/FreeMarker.kt
```diff
@@ -45,11 +47,24 @@ public val FreeMarker: ApplicationPlugin<Configuration> = createApplicationPlugi
...
+    on(ResponseBeforeTransform) { message ->
         if (message is FreeMarkerContent) {
             transformBody {
                 process(message)
             }
         }
     }
 }
+
+internal object ResponseBeforeTransform : Hook<suspend PipelineContext<Any, ApplicationCall>.(body: Any) -> Any> {
```
rsinukov, line 59: A hook can have generic of `suspend (body: Any, type: TypeInfo) -> Any`. Then you won't need `tranformBody` block and can just return your value from `on(MyHook) { body, type -> process(body) }`

rsinukov (changes requested):

ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/application/KtorCallContexts.kt
marychatte: is it mean that I can annotate it with `@InternalAPI` and it in plugins write `@OptIn(InternalAPI::class)`?

marychatte: oh, nevermind, I read the next comment :)

ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/application/hooks/CommonHooks.kt
```diff
@@ -127,3 +129,32 @@ public object ReceiveRequestBytes : Hook<(call: ApplicationCall, body: ByteReadC
...
+            if (body.instanceOf(this@ResponseBeforeTransform.clazz)) {
+                @Suppress("UNCHECKED_CAST")
+                subject = handler(call, body as T)
+            }
+        }
+    }
+}
+
+@InternalAPI
+public inline fun <reified T : Any> PluginBuilder<*>.onResponseBeforeTransform(
```
rsinukov, line 156: IMO this function is redundant. Just use `on(Hook) { ... }` API inside the plugins, especially because it's internal API.

ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/application/hooks/CommonHooks.kt
```diff
@@ -127,3 +129,32 @@ public object ReceiveRequestBytes : Hook<(call: ApplicationCall, body: ByteReadC
         }
     }
 }
+
+/**
+ * A hook that is invoked before `Transform` phase.
+ * Useful for some plugins which used for templates as views within application.
+ */
+@InternalAPI
+public class ResponseBeforeTransform<T : Any>(private val clazz: KClass<T>) :
```
rsinukov, line 138: Maybe rename `ResponseBeforeTransform` -> `BeforeResponseTransform`?

rsinukov (approved):
LGTM but couple of minor comments

PR #3324, opened 2022-12-19
KTOR-5359 Add engine shutdown configuration

marychatte:
**Subsystem**
Server

**Motivation**
[KTOR-5359](https://youtrack.jetbrains.com/issue/KTOR-5359)

### DIFF
--- a/ktor-server/ktor-server-cio/jvmAndNix/src/io/ktor/server/cio/EngineMain.kt
+++ b/ktor-server/ktor-server-cio/jvmAndNix/src/io/ktor/server/cio/EngineMain.kt
@@ -20,9 +20,11 @@ public object EngineMain {
         val applicationEnvironment = commandLineEnvironment(args)
         val engine = CIOApplicationEngine(applicationEnvironment) { loadConfiguration(applicationEnvironment.config) }
         val gracePeriod =
-            engine.environment.config.propertyOrNull("ktor.deployment.shutdownGracePeriod")?.getString()?.toLong() ?: 50
+            engine.environment.config.propertyOrNull("ktor.deployment.shutdownGracePeriod")?.getString()?.toLong()
+                ?: 50
         val timeout =
-            engine.environment.config.propertyOrNull("ktor.deployment.shutdownTimeout")?.getString()?.toLong() ?: 5000
+            engine.environment.config.propertyOrNull("ktor.deployment.shutdownTimeout")?.getString()?.toLong()
+                ?: 5000
         engine.addShutdownHook {
             engine.stop(gracePeriod, timeout)
         }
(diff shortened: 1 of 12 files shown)


ktor-server/ktor-server-host-common/jvmAndNix/src/io/ktor/server/engine/ApplicationEngine.kt
```diff
@@ -37,6 +37,16 @@ public interface ApplicationEngine {
          * Specifies the minimum size of a thread pool used to process application calls.
          */
         public var callGroupSize: Int = parallelism
+
+        /**
+         * Specifies the maximum amount of time for activity to cool down in milliseconds.
+         */
+        public var gracePeriodMillis: Long = 500
```
rsinukov, line 44: I don't see any usages of it. Why make it public property?

ktor-server/ktor-server-cio/jvmAndNix/src/io/ktor/server/cio/EngineMain.kt
```diff
@@ -19,8 +19,12 @@ public object EngineMain {
     public fun main(args: Array<String>) {
         val applicationEnvironment = commandLineEnvironment(args)
         val engine = CIOApplicationEngine(applicationEnvironment) { loadConfiguration(applicationEnvironment.config) }
+        val gracePeriod =
+            engine.environment.config.propertyOrNull("ktor.deployment.gracePeriod")?.getString()?.toLong() ?: 50
+        val timeout =
+            engine.environment.config.propertyOrNull("ktor.deployment.timeout")?.getString()?.toLong() ?: 5000
```
rsinukov, line 25: `ktor.deployment.timeout` is to generic name. can you specify that it's the timeout for stopping a server? same for `gracePeriod`

rsinukov (changes requested):

ktor-server/ktor-server-host-common/jvmAndNix/src/io/ktor/server/engine/ApplicationEngine.kt
marychatte: It is used in `public fun BaseApplicationEngine.Configuration.loadCommonConfiguration`

rsinukov: But it's only set there and not read anywhere

marychatte: But why is it bad that it is public?

rsinukov: I don't understand why it should be. There is nothing users can do with it, and we don't use it as well, only set a value to it.

marychatte: Understand, I agree that it is unnecessary

rsinukov (approved):

PR #3325, opened 2022-12-20
KTOR-5100 Fix command line array properties

marychatte:
**Subsystem**
Server

**Motivation**
[KTOR-5100](https://youtrack.jetbrains.com/issue/KTOR-5100)

### DIFF
--- a/ktor-server/ktor-server-host-common/jvm/test/io/ktor/tests/hosts/CommandLineTest.kt
+++ b/ktor-server/ktor-server-host-common/jvm/test/io/ktor/tests/hosts/CommandLineTest.kt
@@ -89,22 +89,46 @@ class CommandLineTest {
     }
 
     @Test
-    fun testListPropertiesConfig() {
+    fun testListPropertiesHocon() {
         val args = arrayOf(
             "-P:array.first.0=first", "-P:array.first.1=second", "-P:array.first.2=third",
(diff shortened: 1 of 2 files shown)


ktor-server/ktor-server-host-common/jvm/test/io/ktor/tests/hosts/CommandLineTest.kt
```diff
@@ -88,6 +88,25 @@ class CommandLineTest {
         assertEquals("8081", port)
     }
 
+    @Test
+    fun testListPropertiesConfig() {
```
rsinukov, line 92: Please add test for config lists such as 
```
users:
    - name: test
      password: asd
    - name: other
      password: qwe
```
or in string representation
```
users.0.name=test
users.0.password=asd
users.1.name=other
users.1.password=qwe
```

rsinukov (changes requested):

ktor-server/ktor-server-host-common/jvm/test/io/ktor/tests/hosts/CommandLineTest.kt
marychatte: Didn't know about this case, thanks for spotting

rsinukov (approved):
Great job!

PR #3336, opened 2023-01-04
KTOR-5405 allow booleans as multipart values

Rattenkrieg:
**Subsystem**
Client multipart/form-data enhancement

**Motivation**
As per [KTOR-5405](https://youtrack.jetbrains.com/issue/KTOR-5405) there is desire to encode `Boolean` values the same way as `Number` values are encoded.

**Solution**
Wrapping `Boolean`s in `PartData.FormItem` which effectively puts quotes around string representation of the value.

### DIFF
--- a/ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/ContentTest.kt
+++ b/ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/ContentTest.kt
@@ -165,7 +165,6 @@ class ContentTest : ClientLoader(5 * 60) {
     }
 
     @Test
-    @Ignore
     fun testMultipartFormData() = clientTests(listOf("Js")) {
         val data = {
             formData {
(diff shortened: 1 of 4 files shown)


ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/ContentTest.kt
```diff
@@ -184,6 +184,7 @@ class ContentTest : ClientLoader(5 * 60) {
                     }
                 }
                 append("hello", 5)
+                append("world", true)
```
Rattenkrieg, line 187: This test is actually annotated with `@Ignore`, I did a local run with annotation removed though.

rsinukov (changes requested):
Thanks for the PR!
LGTM in general, but since this PR has a new API, can you please rebase it on top of `2.3.0-eap` branch, run `./gradlew apiDump` task and commit changes here?

Rattenkrieg:
@rsinukov thanks for the feedback, addressed.

rsinukov (approved):
LGTM
Will merge after CI check

PR #3347, opened 2023-01-06
Minor API improvements

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-419/Allow-passing-multiple-acceptable-content-types-to-accept-route-selector
https://youtrack.jetbrains.com/issue/KTOR-627/Update-Parameters-and-Headers-DSL-to-be-consistent-with-stdlib

### DIFF
--- a/ktor-http/common/src/io/ktor/http/Headers.kt
+++ b/ktor-http/common/src/io/ktor/http/Headers.kt
@@ -77,6 +77,12 @@ public fun headersOf(name: String, values: List<String>): Headers = HeadersSingl
  */
 public fun headersOf(vararg pairs: Pair<String, List<String>>): Headers = HeadersImpl(pairs.asList().toMap())
 
+/**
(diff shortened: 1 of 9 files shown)


ktor-http/common/src/io/ktor/http/Headers.kt
```diff
@@ -77,6 +77,12 @@ public fun headersOf(name: String, values: List<String>): Headers = HeadersSingl
  */
 public fun headersOf(vararg pairs: Pair<String, List<String>>): Headers = HeadersImpl(pairs.asList().toMap())
 
+/**
+ * Builds a [Headers] instance with the given [builder] function
+ * @param builder specifies a function to build a map
+ */
+public fun headers(builder: HeadersBuilder.() -> Unit): Headers = Headers.build(builder)
```
e5l, line 84: Could you tell me why we can't use `buildHeaders { }` instead?

e5l (approved):

PR #3348, opened 2023-01-06
KTOR-1250 Log HTTP request time on server

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-1250/Log-HTTP-request-time

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-call-logging/jvm/test/CallLoggingTest.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-call-logging/jvm/test/CallLoggingTest.kt
@@ -50,7 +50,9 @@ class CallLoggingTest {
     }
     private val environment: ApplicationEngineEnvironmentBuilder.() -> Unit = {
         module {
-            install(CallLogging)
+            install(CallLogging) {
+                clock { 0 }
+            }
         }
         log = logger
     }
@@ -95,7 +97,7 @@ class CallLoggingTest {
 
         createClient { expectSuccess = false }.get("/")
 
-        assertTrue("INFO: ${red("404 Not Found")}: ${cyan("GET")} - /" in messages)
+        assertTrue("INFO: ${red("404 Not Found")}: ${cyan("GET")} - / in 0ms" in messages)
     }
 
     @Test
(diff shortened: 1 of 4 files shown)


ktor-server/ktor-server-plugins/ktor-server-call-logging/jvm/test/CallLoggingTest.kt
```diff
@@ -118,8 +120,9 @@ class CallLoggingTest {
             module {
                 install(CallLogging) {
                     format { call ->
-                        "${call.request.uri} -> ${call.response.status()}"
+                        "${call.request.uri} -> ${call.response.status()}, took ${call.processingTimeMillis { 3 }}ms"
```
marychatte, line 123: don't we want to have access to the stored `clock`? In this case, a user doesn't have the ability to change the format without changing the default clock. In short, maybe add a getter for clock too?

ktor-server/ktor-server-plugins/ktor-server-call-logging/jvm/test/CallLoggingTest.kt
```diff
@@ -148,8 +173,9 @@ class CallLoggingTest {
         client.get("/")
         client.get("/avoid")
 
-        assertTrue("INFO: ${red("404 Not Found")}: ${cyan("GET")} - /" in messages)
-        assertFalse("INFO: ${red("404 Not Found")}: ${cyan("GET")} - /avoid" in messages)
+        println(messages)
```
marychatte, line 176: debug code left (?)

rsinukov: Thanks, fixed

ktor-server/ktor-server-plugins/ktor-server-call-logging/jvm/test/CallLoggingTest.kt
rsinukov: I added default parameter to `processingTimeMillis` so that users don't need to specify it when they use the default one.

marychatte: This test now falls on `Java 8 Windows` and `Java 8 Linux` because `call.processingTimeMillis()` uses not the same clock that was set

rsinukov: FIxed

marychatte (approved):
LGTM

PR #3349, opened 2023-01-09
[QUIC] added basic QUIC socket API

Mr3zee:
**Subsystem**
ktor-network

**Motivation**
These classes describe basic functionality around QUIC usage as transport protocol.  

**Solution**
`QUICServer` exposes `QUICStream` input and output channels. (`QUICStream` is a stub for now, as we plan to work around it at last inside QUIC itself).  `QUICServer` and `QUICSocketBase` are base classes, which will contain most of the QUIC logic we plan to implement.


ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/streams/QUICStreams.kt
```diff
@@ -0,0 +1,52 @@
...
+    /**
+     * Receive a QUIC stream.
+     */
+    public suspend fun receive(): QUICStream = incoming.receive()
+}
+
+/**
+ * A channel for sending and receiving QUIC streams
+ */
+public interface QUICStreamReadWriteChannel : QUICStreamWriteChannel, QUICStreamReadChannel
```
e5l, line 47: Can be internal

e5l (approved):
lgtm for API

ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/streams/QUICStreams.kt
Mr3zee: These interfaces are exposed via `aScoket().quic()`, do we want to hide it and allow only `HTTP/3` to use `QUIC`?

e5l: it's ok, let's keep it.

Feel free to merge after aprove

PR #3359, opened 2023-01-10
KTOR-5420 Fix DropwizardMetricsPlugin when using with StatusPages plugin

marychatte:
**Subsystem**
Server

**Motivation**
[KTOR-5420](https://youtrack.jetbrains.com/issue/KTOR-5420/DropwizardMetricsPlugin-logs-status-code-incorrectly-when-is-used-together-with-StatusPages-plugin)

**Solution**
Invoke phases for DropwizardMetricsPlugin after phase `CallFalled`

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-metrics/jvm/src/io/ktor/server/metrics/dropwizard/DropwizardMetrics.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-metrics/jvm/src/io/ktor/server/metrics/dropwizard/DropwizardMetrics.kt
@@ -9,6 +9,7 @@ import com.codahale.metrics.MetricRegistry.*
 import com.codahale.metrics.jvm.*
 import io.ktor.server.application.*
 import io.ktor.server.application.hooks.*
+import io.ktor.server.request.*
 import io.ktor.server.routing.*
 import io.ktor.util.*
 import java.util.concurrent.*
@@ -100,10 +101,17 @@ public val DropwizardMetrics: ApplicationPlugin<DropwizardMetricsConfig> =
     }
 
 private fun PluginBuilder<DropwizardMetricsConfig>.callFinished(call: ApplicationCall) {
-    val routingMetrics = call.attributes.take(routingMetricsKey)
+    val routingMetrics = call.attributes.takeOrNull(routingMetricsKey)
+    val name = routingMetrics?.name ?: call.request.routeName
     val status = call.response.status()?.value ?: 0
     val statusMeter =
-        pluginConfig.registry.meter(name(pluginConfig.baseName, routingMetrics.name, status.toString()))
+        pluginConfig.registry.meter(name(pluginConfig.baseName, name, status.toString()))
     statusMeter.mark()
-    routingMetrics.context.stop()
+    routingMetrics?.context?.stop()
 }
+
+private val ApplicationRequest.routeName: String
+    get() {
+        val metricUri = uri.ifEmpty { "/" }.let { if (it.endsWith('/')) it else "$it/" }
+        return "$metricUri(method:${httpMethod.value})"
+    }
--- a/ktor-server/ktor-server-plugins/ktor-server-metrics/jvm/src/io/ktor/server/metrics/dropwizard/DropwizardMetrics.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-metrics/jvm/src/io/ktor/server/metrics/dropwizard/DropwizardMetrics.kt
@@ -79,22 +79,15 @@ public val DropwizardMetrics: ApplicationPlugin<DropwizardMetricsConfig> =
             )
         }
 
-        on(MonitoringEvent(Routing.RoutingCallFinished)) { call ->
-            val routingMetrics = call.attributes.take(routingMetricsKey)
-            val status = call.response.status()?.value ?: 0
-            val statusMeter =
-                pluginConfig.registry.meter(name(pluginConfig.baseName, routingMetrics.name, status.toString()))
-            statusMeter.mark()
-            routingMetrics.context.stop()
-        }
-
         @OptIn(InternalAPI::class)
         on(Metrics) { call ->
             active.inc()
             call.attributes.put(measureKey, CallMeasure(duration.time()))
         }
 
-        on(AfterCall) { call ->
+        on(ResponseSent) { call ->
+            callFinished(call)
+
             active.dec()
             val meter = httpStatus.computeIfAbsent(call.response.status()?.value ?: 0) {
                 pluginConfig.registry.meter(name(pluginConfig.baseName, "status", it.toString()))
@@ -105,3 +98,12 @@ public val DropwizardMetrics: ApplicationPlugin<DropwizardMetricsConfig> =
             }
         }
     }
+
+private fun PluginBuilder<DropwizardMetricsConfig>.callFinished(call: ApplicationCall) {
+    val routingMetrics = call.attributes.take(routingMetricsKey)
+    val status = call.response.status()?.value ?: 0
+    val statusMeter =
+        pluginConfig.registry.meter(name(pluginConfig.baseName, routingMetrics.name, status.toString()))
+    statusMeter.mark()
+    routingMetrics.context.stop()
+}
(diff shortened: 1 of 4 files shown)


ktor-server/ktor-server-plugins/ktor-server-metrics/jvm/src/io/ktor/server/metrics/dropwizard/DropwizardMetricsUtils.kt
```diff
@@ -5,23 +5,10 @@
 package io.ktor.server.metrics.dropwizard
 
 import com.codahale.metrics.*
-import io.ktor.server.application.*
 import io.ktor.util.*
 
 internal class RoutingMetrics(val name: String, val context: Timer.Context)
 internal val routingMetricsKey = AttributeKey<RoutingMetrics>("metrics")
```
v1ctor, line 11: One thing we noticed is that name of this 2 keys are the same. 

Because of this looking up in the attributes map fails with the class cast exception in some cases.

ktor-server/ktor-server-plugins/ktor-server-metrics/jvm/src/io/ktor/server/metrics/dropwizard/DropwizardMetrics.kt
```diff
@@ -105,3 +98,12 @@ public val DropwizardMetrics: ApplicationPlugin<DropwizardMetricsConfig> =
             }
         }
     }
+
+private fun PluginBuilder<DropwizardMetricsConfig>.callFinished(call: ApplicationCall) {
+    val routingMetrics = call.attributes.take(routingMetricsKey)
```
v1ctor, line 103: It looks like 
```kotlin 
on(MonitoringEvent(Routing.RoutingCallStarted)) 
```
Isn't called for OPTIONS method http calls.

But
```kotlin
on(ResponseSent)
```
is called

Because of that there is always an IllegalStateException on every OPTIONS method call in this line.

marychatte: Hello, can you please provide a code as I tried to reproduce and there was no exception

v1ctor: Looks like the issue happens when this plugin is used with CORS plugin:

Test
```kotlin
@Test
    fun `with Options call`() = testApplication {
        val testRegistry = MetricRegistry()
        install(DropwizardMetrics) {
            registry = testRegistry
            baseName = ""
        }
        install(CORS) {
            allowMethod(HttpMethod.Get)
            anyHost()
        }

        routing {
            get("/ok") {
                call.respond("OK")
            }
        }

        val response = client.options("/ok") {
            header("Access-Control-Request-Method", "GET")
            header("Origin", "https://ktor.io")
        }

        assertEquals(HttpStatusCode.OK, response.status)
        assertEquals(1, testRegistry.meter("/ok/(method:OPTIONS).200").count)
    }
```

DropwizardMetrics.kt
```kotlin
private fun PluginBuilder<DropwizardMetricsConfig>.callFinished(call: ApplicationCall) {
    try {
        val routingMetrics = call.attributes.take(routingMetricsKey)
        val status = call.response.status()?.value ?: 0
        val statusMeter =
            pluginConfig.registry.meter(name(pluginConfig.baseName, routingMetrics.name, status.toString()))
        statusMeter.mark()
        routingMetrics.context.stop()
    } catch (e: Exception) {
        println(e)
    }
}
```

Error:
```
java.lang.IllegalStateException: No instance for key AttributeKey: routing
```

It seems to me that CORS returns response here https://github.com/ktorio/ktor/blob/c1dbbcbed8ddf314aa33b4cb61457f3361ad6331/ktor-server/ktor-server-plugins/ktor-server-cors/jvmAndNix/src/io/ktor/server/plugins/cors/CORS.kt#L126
before Routing plugin had a chance to start.

ktor-server/ktor-server-plugins/ktor-server-metrics/jvm/src/io/ktor/server/metrics/dropwizard/DropwizardMetricsUtils.kt
marychatte: yes, fixed

ktor-server/ktor-server-plugins/ktor-server-metrics/jvm/src/io/ktor/server/metrics/dropwizard/DropwizardMetrics.kt
marychatte: Added fallback in DropWizardMetric, if there was not called `on(MonitoringEvent(Routing.RoutingCallStarted))`

ktor-server/ktor-server-plugins/ktor-server-metrics/jvm/src/io/ktor/server/metrics/dropwizard/DropwizardMetrics.kt
```diff
@@ -105,3 +99,19 @@ public val DropwizardMetrics: ApplicationPlugin<DropwizardMetricsConfig> =
...
     }
+
+private fun PluginBuilder<DropwizardMetricsConfig>.callFinished(call: ApplicationCall) {
+    val routingMetrics = call.attributes.takeOrNull(routingMetricsKey)
+    val name = routingMetrics?.name ?: call.request.routeName
+    val status = call.response.status()?.value ?: 0
+    val statusMeter =
+        pluginConfig.registry.meter(name(pluginConfig.baseName, name, status.toString()))
+    statusMeter.mark()
+    routingMetrics?.context?.stop()
```
v1ctor, line 110: It seems to me that in the case of missing `routingMetrics` timer measurement is going to be missing too.

v1ctor:
Overall it looks great to me. I'm curious if it's possible to capture OPTIONS timer measurements too?

Anyway, I think it works much better that it was before. I will leave it to someone from ktor team to review and accept. 

Thanks for you work @marychatte !

ktor-server/ktor-server-plugins/ktor-server-metrics/jvm/src/io/ktor/server/metrics/dropwizard/DropwizardMetrics.kt
marychatte: If routing didn't start then timer measurement wouldn't show anything useful. So I'm not sure but I think it's not critical

v1ctor: But http call using method OPTIONS also have latency and it might be useful to measure.

I think that the problem that CORS plugin work around Routing plugin which is a bit weird. 

Seems like a low pri to me, but someone make think otherwise.

ktor-server/ktor-server-plugins/ktor-server-metrics/jvm/src/io/ktor/server/metrics/dropwizard/DropwizardMetrics.kt
```diff
@@ -79,22 +80,15 @@ public val DropwizardMetrics: ApplicationPlugin<DropwizardMetricsConfig> =
...
-
         @OptIn(InternalAPI::class)
         on(Metrics) { call ->
             active.inc()
             call.attributes.put(measureKey, CallMeasure(duration.time()))
         }
 
-        on(AfterCall) { call ->
+        on(ResponseSent) { call ->
+            callFinished(call)
```
rsinukov, line 90: can you please do a small cleanup here and extract all `pluginConfig` properties into local vals? config is a mutable object, and it's better to capture all its properties during the plugin installation.

ktor-server/ktor-server-plugins/ktor-server-metrics/jvm/src/io/ktor/server/metrics/dropwizard/DropwizardMetrics.kt
rsinukov: CORS plugin replies to OPTIONS request before any route is matched. That's why there is no routing measure yet.

rsinukov (approved):
LGTM. Thanks for the fix and @v1ctor for the review.

PR #3361, opened 2023-01-11
KTOR-3455 Fix reading request body back pressure in Netty

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-3455
https://youtrack.jetbrains.com/issue/KTOR-4475


ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/cio/RequestBodyHandler.kt
```diff
@@ -12,12 +12,14 @@ import io.netty.util.*
 import kotlinx.coroutines.*
 import kotlinx.coroutines.channels.Channel
 import java.lang.Integer.*
+import java.util.concurrent.atomic.AtomicInteger
```
e5l, line 15: In the result, it would be better using atomicfu - one allocation less

e5l (approved):
lgtm

PR #3363, opened 2023-01-12
KTOR-5327 CIO should use Connection: Close value as default in HTTP/1.0

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-5327


e5l:
Could you check if the failed tests related?

rsinukov:
@e5l They are broken on 2.3.0-eap branch, but not related to this PR

e5l (approved):
lgtm

PR #3364, opened 2023-01-14
KTOR-5740 Add opportunity to pass type info into receiveDeserializedBase and other

InsanusMokrassar:
**Subsystem**

Client, Websockets

**Motivation**

Sometimes it is not possible to use reified parameters when creating websockets and use deserialization of result

**Solution**

In that case it is reasonable to extract type info into parameters and provide API to use websockets with the prepared type info

### DIFF
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/websocket/ClientSessions.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/websocket/ClientSessions.kt
@@ -68,6 +68,8 @@ public suspend inline fun <reified T> DefaultClientWebSocketSession.sendSerializ
  * In this case, [WebsocketDeserializeException.frame] contains the received frame.
  * May throw [ClosedReceiveChannelException] if a channel was closed
  *
+ * @param typeInfo Type info of [T]. Can be retrieved with [typeInfo] fun and will be used to work with result [T] type
+ *
  * @throws WebsocketConverterNotFoundException if no [contentConverter] is found for the [WebSockets] plugin
  * @throws WebsocketDeserializeException if the received frame can't be deserialized to type [T]
  */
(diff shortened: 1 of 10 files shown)


ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/websocket/ClientSessions.kt
```diff
@@ -68,15 +68,32 @@ public suspend inline fun <reified T> DefaultClientWebSocketSession.sendSerializ
  * In this case, [WebsocketDeserializeException.frame] contains the received frame.
  * May throw [ClosedReceiveChannelException] if a channel was closed
  *
+ * @param typeInfo Type info of [T]. Can be retrieved with [typeInfo] fun and will be used to work with result [T] type
```
rsinukov, line 71: fun -> function

ktor-shared/ktor-websocket-serialization/common/src/io/ktor/websocket/serialization/WebsocketChannelSerialization.kt
```diff
@@ -39,13 +39,15 @@ public suspend inline fun <reified T> WebSocketSession.sendSerializedBase(
  * In this case, [WebsocketDeserializeException.frame] contains the received frame.
  * May throw [ClosedReceiveChannelException] if a channel was closed
  *
+ * @param typeInfo Type info of [T]. Can be retrieved with [typeInfo] fun and will be used to work with result [T] type
```
rsinukov, line 42: fun -> function

rsinukov (changes requested):
Hi and thanks for the PR.
This PR has a new API, so it can only be merged into `2.3.0-eap` branch. Do you mind rebasing it, running `./gradlew apiDump` and committing changes here?
Also, can you add a test, please?

rsinukov (approved):
I made the relevant changes here. Ready to be merged now.

PR #3370, opened 2023-01-18
KTOR-3799 Add charset only for text/* content types

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-3799

### DIFF
--- a/ktor-http/common/src/io/ktor/http/ContentTypes.kt
+++ b/ktor-http/common/src/io/ktor/http/ContentTypes.kt
@@ -286,7 +286,7 @@ public fun ContentType.withCharset(charset: Charset): ContentType =
  * if [ContentType] is not ignored
  */
 public fun ContentType.withCharsetIfNeeded(charset: Charset): ContentType =
-    if (contentType.lowercase() == "application" && contentSubtype.lowercase() == "json") {
+    if (contentType.lowercase() != "text") {
         this
     } else {
         withParameter("charset", charset.name)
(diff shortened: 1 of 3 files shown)


e5l:
Please check compilation errors on linux

ktor-http/common/src/io/ktor/http/ContentTypes.kt
```diff
@@ -286,7 +286,7 @@ public fun ContentType.withCharset(charset: Charset): ContentType =
  * if [ContentType] is not ignored
  */
 public fun ContentType.withCharsetIfNeeded(charset: Charset): ContentType =
-    if (contentType.lowercase() == "application" && contentSubtype.lowercase() == "json") {
+    if (contentType.lowercase() != "text") {
```
Stexxe, line 289: This is a workaround replaced with a more general workaround.

rsinukov: No, it's not. Only text content type should have charset parameter.

rsinukov:
@e5l thanks, fixed. It was an imports problem

e5l (approved):

ktor-http/common/src/io/ktor/http/ContentTypes.kt
Stexxe: You right. Although, I don't like the method's name. `withCharsetIfApplicable` would be better.

PR #3371, opened 2023-01-18
KTOR-5443 Fix FileNotFoundException for long file names in FileStorage

marychatte:
**Subsystem**
Client

**Motivation**
[KTOR-5443](https://youtrack.jetbrains.com/issue/KTOR-5443/FileStorage-throws-java.io.FileNotFoundException-File-name-too-long-when-request-path-is-long)

**Solution**
Use a hash of URLs  instead URLs for keys


e5l (approved):
lgtm

PR #3380, opened 2023-01-27
[QUIC] Frames and errors

Mr3zee:
**Subsystem**
ktor-network

**Description**
This PR contains implementation for handlers that work with QUIC frames on lowest level (encoding/decoding)


ktor-network/ktor-network-quic/jvmAndNix/test/io.ktor.network.quic/frames/FrameReadWriteTest.kt
```diff
@@ -0,0 +1,680 @@
+/*
```
whyoleg, line 1: package file location is wrong - should be separated by `/` and not by `.`(dots)

ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/errors/Error_v1.kt
```diff
@@ -0,0 +1,106 @@
...
+            return when {
+                code > 0x10 -> null
+                else -> array[code]
+            }
+        }
+    }
+}
+
+@JvmInline
+internal value class CryptoHandshakeError_v1(val tlsAlertCode: UByte) : QUICTransportError_v1 {
```
whyoleg, line 96: this `value` class will be always boxed now

ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/frames/FrameReader.kt
```diff
@@ -0,0 +1,450 @@
...
+package io.ktor.network.quic.frames
+
+import io.ktor.network.quic.bytes.*
+import io.ktor.network.quic.errors.*
+import io.ktor.network.quic.packets.*
+import io.ktor.network.quic.util.*
+import io.ktor.utils.io.core.*
+
+internal object FrameReader {
+    suspend inline fun readFrame(
```
whyoleg, line 14: just want to mention, that making this method `inline` is dangerous, as it will increase amount of bytecode produced if it's will be called in a lot of places

whyoleg (commented):
just some comments from outside :)

ktor-network/ktor-network-quic/build.gradle.kts
```diff
@@ -10,6 +10,11 @@ kotlin {
                 api(project(":ktor-utils"))
             }
         }
+        jvmAndNixTest {
+            dependencies {
+                implementation(libs.kotlin.test.junit)
```
e5l, line 15: there is no junit in nix

e5l, line 15: There are no junit in nix.

ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/bytes/ByteUtils.kt
```diff
@@ -0,0 +1,13 @@
...
+
+package io.ktor.network.quic.bytes
+
+import io.ktor.utils.io.core.*
+
+/**
+ * Reads an unsigned byte from a [ByteReadPacket]. Returns null if EOF
+ */
+@ExperimentalUnsignedTypes
+internal fun ByteReadPacket.readUByteOrNull() = if (isNotEmpty) readUByte() else null
```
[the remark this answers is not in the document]
e5l: It would be better if required amount of bytes will be checked before all read calls

ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/errors/Error_v1.kt
```diff
@@ -0,0 +1,106 @@
...
+    fun writeToFrame(packetBuilder: BytePacketBuilder)
+}
+
+/**
+ * Application error codes that are produced and managed by application layer protocol
+ *
+ * Used in CONNECTION_CLOSED frames with the type of 0x1d
+ */
+@JvmInline
+internal value class AppError_v1(val intCode: Long) : Error_v1 {
```
e5l, line 28: Could you explain in KDoc why suffixes _v1 are needed?

ktor-network/ktor-network-quic/jvmAndNix/test/io.ktor.network.quic/frames/base/TestPacketTransportParameters.kt
```diff
@@ -0,0 +1,11 @@
+/*
+ * Copyright 2014-2023 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.network.quic.frames.base
+
+import io.ktor.network.quic.packets.*
+
+data class TestPacketTransportParameters(
```
e5l, line 9: the package name is broken: `io.ktor.network.quic`, it should be `io/ktor/network/quic` instead

e5l (changes requested):

ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/bytes/ByteUtils.kt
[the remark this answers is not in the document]
Mr3zee: what do you think about this?
```
@ExperimentalUnsignedTypes
internal inline fun ByteReadPacket.readUByteOrElse(elseBlock: () -> Nothing): UByte {
    if (isEmpty) {
        elseBlock()
    }
    return readUByte()
} 
``` 
```
val ubyte = readUByteOrElse { return null }
```

ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/errors/Error_v1.kt
Mr3zee: You mean because of inheritance?

ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/bytes/VarInt.kt
```diff
@@ -0,0 +1,77 @@
...
+import io.ktor.network.quic.util.*
+import io.ktor.utils.io.core.*
+
+/**
+ * Reads variable-length non-negative integer value. Returns null if EOF.
+ *
+ * [RFC Reference](https://www.rfc-editor.org/rfc/rfc9000.html#name-variable-length-integer-enc)
+ */
+@OptIn(ExperimentalUnsignedTypes::class)
+internal fun ByteReadPacket.readVarIntOrNull(): Long? {
```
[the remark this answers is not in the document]
Mr3zee: I tried to do it with value class in the first place, but it seemed redundant, as it is only used in the decoding and encoding of the frames. It this particular case there can be -1 as the `null` value, but maybe the approach as with  `readUByteOrIfEmpty` be even better

ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/bytes/VarInt.kt
```diff
@@ -0,0 +1,77 @@
...
+ * [RFC Reference](https://www.rfc-editor.org/rfc/rfc9000.html#name-variable-length-integer-enc)
+ */
+internal fun BytePacketBuilder.writeVarInt(value: Long) {
+    require(value >= 0) { "Variable-length integer cannot be negative, actual: $value" }
+    require(value < POW_2_62) { "Max value of variable-length integer value is ${POW_2_62 - 1}, actual: $value" }
+
+    var shift: Int
+    val length: Long
+    when {
+        value < POW_2_06 -> {
```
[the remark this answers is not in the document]
Mr3zee: well yes, but is that performance impact will be that significant?

ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/frames/FrameReader.kt
Mr3zee: as I see it now, it should be called only in one-two places but many many times and I placed the `inline` modifier to inline the lambda argument

ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/bytes/VarInt.kt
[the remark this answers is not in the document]
Mr3zee: but i don't want to make it inline, so maybe something like this would work, wdyt?
```
internal inline fun ByteReadPacket.readVarIntOrElse(elseBlock: () -> Nothing): Long {
    val value = readVarInt()
    if (value == -1L) {
        elseBlock()
    }
    return value
}

internal fun ByteReadPacket.readVarInt(): Long {
    val value = readUByteOrElse { return -1 }.toInt()
    ...
```

ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/bytes/VarInt.kt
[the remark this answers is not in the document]
whyoleg: performance isn't my main concern. readability is.
but it's minor remark, not sure, that this function will be edited a lot :)

ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/bytes/ByteUtils.kt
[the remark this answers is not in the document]
whyoleg: from what I see in code, may be even just calling `isEmpty` in place will be better, for readability at least as there is not a lot of calls to this function comparing to `readVarInt

ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/bytes/VarInt.kt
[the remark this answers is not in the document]
whyoleg: Yes, that's the right approach IMO (considering how it's used in code), `private no-inline` function, which returns -1 and `internal inline` which check for `null`

ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/errors/Error_v1.kt
whyoleg: Not specifically, `AppError_v1` also extends interface, but it's used as direct class in code (at least, I saw only such usage). In comparison `CryptoHandshakeError_v1` is always used by it parent `QUICTransportError_v1`

On a side note, it's an error, so not something, that will be instantiated frequently, so may be it's even not an issue to make it ordinary class.

ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/frames/FrameWriter.kt
```diff
@@ -0,0 +1,544 @@
...
+        writeUByte(typeV1.typeValue)
+    }
+
+    @OptIn(ExperimentalUnsignedTypes::class)
+    private fun BytePacketBuilder.writeConnectionId(id: ByteArray) {
+        writeUByte(id.size.toUByte())
+        writeFully(id)
+    }
+
+    private fun BytePacketBuilder.writeErrorCode(errorCode: Error_v1) {
```
whyoleg, line 541: BTW, this will also involve boxing for inherited `inline` classes. May be just making this function `inline` will allow to not box it, but Im not sure

ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/bytes/ByteUtils.kt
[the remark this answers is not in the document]
Mr3zee: I think for consistency it would be better to  make it the same as `readVarIntOrElse`, as where are a lot of usages of the last one and it wont be a problem for reader to understand the meaning quickly. Besides that, i will also use it to read packet headers, so there will be more occurrences of this function

ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/bytes/VarInt.kt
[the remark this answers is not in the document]
Mr3zee: well, yeah, I kinda did not like the way function looks too, guess I'm not the only one so I'll rewrite it

ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/frames/FrameWriter.kt
Mr3zee: I decided to get rid of Error_v1 interface at all, as it is the only place it is used, and app and transport error are processed separately

ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/bytes/VarInt.kt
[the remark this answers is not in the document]
Mr3zee: wdyt about the new one?

ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/frames/FrameReader.kt
Mr3zee: I'll leave it as it is for now, and will change the approach if the future if it would be needed

ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/bytes/ByteUtils.kt
```diff
@@ -0,0 +1,18 @@
...
+
+import io.ktor.utils.io.core.*
+
+/**
+ * Reads an unsigned byte from a [ByteReadPacket]. Returns null if EOF
+ */
+@ExperimentalUnsignedTypes
+internal inline fun ByteReadPacket.readUByteOrElse(elseBlock: () -> UByte): UByte {
+    if (isEmpty) {
+        elseBlock()
```
whyoleg, line 15: `elseBlock` return value is discarded
or use `return elseBlock()` or change type of `elseBlock` to `() -> Nothing`

(Same for `readVarIntOrElse`)

ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/bytes/VarInt.kt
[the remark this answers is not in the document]
whyoleg: I like it more!
minor remark: now I see, that may be we can replace `writeByte` with `writeShort/Int/Long` for different lengths?

ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/bytes/VarInt.kt
```diff
@@ -0,0 +1,86 @@
...
+ * [RFC Reference](https://www.rfc-editor.org/rfc/rfc9000.html#name-variable-length-integer-enc)
+ */
+@OptIn(ExperimentalUnsignedTypes::class)
+private fun ByteReadPacket.readVarInt(): Long {
+    val value = readUByteOrElse { return -1 }.toInt()
+    val length = 1 shl (value ushr 6)
+
+    var varInt = (value and 0x3f).toLong()
+    var i = 0
+    while (i < length - 1) {
```
whyoleg, line 32: may be we can unroll this loop here? It will be then symmetric to `writeVarInt` and again, easier to read/understand :)

Mr3zee: Actually this one from specification https://www.rfc-editor.org/rfc/rfc9000.html#name-sample-variable-length-inte

ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/bytes/VarInt.kt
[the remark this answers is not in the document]
Mr3zee: Hm, I like the idea! Will try it

ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/errors/Error_v1.kt
```diff
@@ -0,0 +1,100 @@
...
+}
+
+internal class CryptoHandshakeError_v1(val tlsAlertCode: UByte) : QUICTransportError_v1 {
+    @OptIn(ExperimentalUnsignedTypes::class)
+    override fun writeToFrame(packetBuilder: BytePacketBuilder) {
+        // 0b01000001, where 0100 prefix - varint length,
+        //                   0001 byte - prefix of crypto error (reserved range: 0x0100..0x01ff)
+        //
+        // RFC Reference: https://www.rfc-editor.org/rfc/rfc9001#name-tls-errors
+        packetBuilder.writeUByte(0x41u)
```
whyoleg, line 97: minor: may be extract 0x41 to constant with meaningful name (used both in reading and writing)

ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/bytes/VarInt.kt
whyoleg: Yeah, I see it now
But, IMO, It looks better (easier to debug/understand) without loop
Look for example on `quiche` impl - https://github.com/cloudflare/quiche/blob/1aa5c702332bf0c90984f5470e8cb750d2dd027f/octets/src/lib.rs#L406-L469

e5l (approved):

ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/bytes/VarInt.kt
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ * Reads variable-length non-negative integer value.
+ * Returns -1 if EOF.
+ *
+ * [RFC Reference](https://www.rfc-editor.org/rfc/rfc9000.html#name-variable-length-integer-enc)
+ */
+@OptIn(ExperimentalUnsignedTypes::class)
+private fun ByteReadPacket.readVarInt(): Long {
+    val value = tryPeek().let { if (it == -1) return -1 else it }
```
whyoleg, line 36: minor: I would unroll this one:
```kotlin
val value = tryPeek()
if (value == -1) return -1
```

whyoleg (approved):

ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/bytes/VarInt.kt
Mr3zee: done. Thanks for your comments and suggestions!

whyoleg: thanks for working on it!
Im willing to try working implementation in combination with rsocket-kotlin when it will be ready (even in some snapshot version).
If everything goes well, it will also allow me to do cross test it with netty-quic (via quiche) implementation!

PR #3384, opened 2023-01-30
Add Turbo Streams content type "text/vnd.turbo-stream.html"

morki:
**Subsystem**
Client, Server, Http

**Motivation**
Using content type for Turbo Stream from standard location `ContentType.Text.TurboStream`

**Solution**
Add Turbo Streams content type `text/vnd.turbo-stream.html`
See https://turbo.hotwired.dev/handbook/streams#streaming-from-http-responses

**Alternative**
Create empty `companion object` in `ContentType.Text` to allow add this in ktor Turbo plugin (which i now work on to contribute to server templating plugins)


rsinukov (changes requested):
Hello and thanks for the PR. 

I think it's better to use your alternative solution with an empty companion object added to a `Text` and other content types, so that users can extend with their custom ones.
If you do that, please rebase your PR onto `2.3.0-eap` branch, run `./gradlew apiDump` and commit changes here.

PR #3388, opened 2023-02-04
KTOR-5580 Add "resource" to supported protocols for classpath resource resolution

morki:
**Subsystem**
Server, Static Resources

**Motivation**
It is needed to enable serving static files inside GraalVM native image.

**Solution**
Added "resource" protocol to supported protocols.


rsinukov (changes requested):
Hello and thanks for the PR.
Do you mind creating a ticket in YouTrack and adding a test for this change, please?

rsinukov (approved):
Indeed, adding a test here is not trivial. LGTM

PR #3392, opened 2023-02-06
KTOR-5456 Reduce severity level to avoid flooding console

Stexxe:
https://youtrack.jetbrains.com/issue/KTOR-5456


e5l:
Could you tell me why we should have different log levels for different platforms?

PR #3393, opened 2023-02-06
KTOR-5448 Unmute legit SerializationException

e5l:


sandwwraith:
Don't you want to use `serializerOrNull` in `serializerFromTypeInfo`?

e5l:
@sandwwraith, could you clarify what `serializerOrNull` we should use? We already using some `serializerOrNull` https://github.com/ktorio/ktor/blob/bad1c2da472646318a402342a249cc6300bc0940/ktor-shared/ktor-serialization/ktor-serialization-kotlinx/common/src/io/ktor/serialization/kotlinx/SerializerLookup.kt#L21 in the described case

sandwwraith:
Ah, sorry, I didn't see that. Disregard my comment then.

marychatte (approved):
lgtm

PR #3395, opened 2023-02-07
KTOR-5492 Javadoc for Resources.kt cannot be compiled

helpermethod:
**Subsystem**
- Client Resources Plugin
- Server Resources Plugin

**Motivation**
The JavaDoc in its current form does not compile as the `Users` data class must have at least one primary constructor parameter, which it doesn't. See [KTOR-5492](https://youtrack.jetbrains.com/issue/KTOR-5492/Javadoc-for-Resources.kt-cannot-be-compiled) for more details.

**Solution**
Replaces the data classes in the example with normal classes analogue to the examples in the documentation.


rsinukov (approved):

rsinukov:
Thanks for the PR!

PR #3399, opened 2023-02-09
KTOR-5510 Fix isSent when content is ReadChannelContent

marychatte:
**Motivation**
[KTOR-5510](https://youtrack.jetbrains.com/issue/KTOR-5510/Compressing-the-response-will-result-in-unexpected-ERROR-log-output-after-processing-in-the-StatusPages) 

**Solution**
Flag `isSent` wasn't set because of an accident `return` call


e5l (dismissed):
Nice catch, could we have a test?

rsinukov (approved):

PR #3400, opened 2023-02-09
KTOR-5529 Add option to CIO HttpServer to reuse the address

chedabob:
**Subsystem**
 ktor-server-cio

**Motivation**
There is a feature request open on YouTrack for this https://youtrack.jetbrains.com/issue/KTOR-5527

Related to that is this bug, however I wouldn't consider this PR a permanent fix to it https://youtrack.jetbrains.com/issue/KTOR-5529

This also extends the work done in this PR to make it configurable https://github.com/ktorio/ktor/pull/3387

**Solution**
Bubble up the socket option `reuseAddress` to CIOApplicationEngine's `Configuration` class

### DIFF
--- a/ktor-server/ktor-server-cio/api/ktor-server-cio.api
+++ b/ktor-server/ktor-server-cio/api/ktor-server-cio.api
@@ -13,7 +13,9 @@ public final class io/ktor/server/cio/CIOApplicationEngine : io/ktor/server/engi
 public final class io/ktor/server/cio/CIOApplicationEngine$Configuration : io/ktor/server/engine/BaseApplicationEngine$Configuration {
 	public fun <init> ()V
 	public final fun getConnectionIdleTimeoutSeconds ()I
+	public final fun getReuseAddress ()Z
 	public final fun setConnectionIdleTimeoutSeconds (I)V
+	public final fun setReuseAddress (Z)V
 }
 
 public final class io/ktor/server/cio/EngineMain {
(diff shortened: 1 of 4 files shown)


EvGem7:
Allowing it to be configured is better approach, yeah. Probably should close my PR.

SamC-Apadmi:
Any movement on getting this in? I could really do with this feature :D

89jd:
> Any movement on getting this in? I could really do with this feature :D

+1

e5l:
Hey, @chedabob! Thanks for the PR. 
Could you please run an `./gradlew apiDump` to update API declarations?

chedabob:
> Hey, @chedabob! Thanks for the PR. Could you please run an `./gradlew apiDump` to update API declarations?

Done, and have rebased onto latest from Main.

EvGem7:
What do you think about making the reuseAddress true by default? It's weird if we have to keep in mind, that in order to make it work, we must set the flag. The working case should be default case, shouldn't it?

chedabob:
> What do you think about making the reuseAddress true by default? It's weird if we have to keep in mind, that in order to make it work, we must set the flag. The working case should be default case, shouldn't it?

It felt safer to allow people having issues to opt into it, rather than make a potentially breaking change for everyone else. I don't fully understand what's going on at the kernel level, but presumably if it didn't have side effects, it would also be default there (although it's equally likely it's just like that for historical reasons).

e5l:
@chedabob, thanks for the update - will be merged after CI run. 
We can't merge it in the `main`, cause of API changes, it will be released with 2.3.0 in March.

The default behavior change can be done only with a major 3.0.0 release.

chedabob:
> @chedabob, thanks for the update - will be merged after CI run. We can't merge it in the `main`, cause of API changes, it will be released with 2.3.0 in March.
> 
> The default behavior change can be done only with a major 3.0.0 release.

Ah didn't realise there was an EAP branch open. Do you want me to rebase onto that branch? It looks like this PR will accidentally bring in some extra changes that are on Main.

e5l:
Yep, it would be great

chedabob:
@e5l Done. Apologies if this pinged you multiple times, I'm not as familiar with Github's PR flow so wasn't sure if the force-push marked the PR as updated.

EvGem7:
Is it closed by accident? :eyes: 
Just need to change target branch?

e5l:
Yep, @chedabob  could you please reopen?

PR #3401, opened 2023-02-12
Correct private max age key to s-maxage

nbransby:
**Subsystem**
Client

**Motivation**
Fixing incorrect key name causing cache not to respect max age on private caches


rsinukov (changes requested):
Thanks for the PR. Do you mind adding a test for this change, please?

nbransby:
I spent enough of my weekend debugging and tracking this one down - I'll leave the test to someone else, perhaps the person who wrote this code without the test in the first place could volunteer 🤣 

I am only joking but I am serious that I don't have the time, apologies for that!

rsinukov:
I reimplemented this change along with other fixed in the #3406. Closing this PR.

PR #3402, opened 2023-02-13
KTOR-5467 Add default sessions auth challenge behaviour

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-5467

Merging to 2.3.0 because default behaviour changes


e5l (approved):
lgtm. We should check the Space on this change

PR #3403, opened 2023-02-13
KTOR-5523 Add ability to filter sensitive header values in logging

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-5523


e5l (approved):
lgtm

PR #3405, opened 2023-02-14
KTOR-5560 Add shutdown configuration for engine

marychatte:
**Subsystem**
Server

**Motivation**
[KTOR-5560](https://youtrack.jetbrains.com/issue/KTOR-5560/Add-shutdown-configuration-for-engine-in-stop-method)

### DIFF
--- a/ktor-server/ktor-server-host-common/jvmAndNix/src/io/ktor/server/engine/CommandLine.kt
+++ b/ktor-server/ktor-server-host-common/jvmAndNix/src/io/ktor/server/engine/CommandLine.kt
@@ -130,6 +130,12 @@ public fun BaseApplicationEngine.Configuration.loadCommonConfiguration(deploymen
     deploymentConfig.propertyOrNull("workerGroupSize")?.getString()?.toInt()?.let {
         workerGroupSize = it
     }
+    deploymentConfig.propertyOrNull("shutdownGracePeriod")?.getString()?.toLong()?.let {
+        shutdownGracePeriod = it
+    } ?: { shutdownGracePeriod = 1000 }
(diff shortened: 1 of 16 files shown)


ktor-server/ktor-server-host-common/jvmAndNix/src/io/ktor/server/engine/CommandLine.kt
```diff
@@ -130,6 +130,12 @@ public fun BaseApplicationEngine.Configuration.loadCommonConfiguration(deploymen
     deploymentConfig.propertyOrNull("workerGroupSize")?.getString()?.toInt()?.let {
         workerGroupSize = it
     }
+    deploymentConfig.propertyOrNull("shutdownGracePeriod")?.getString()?.toLong()?.let {
+        shutdownGracePeriod = it
+    } ?: { shutdownGracePeriod = 1000 }
```
e5l, line 135: Could we extract default values to a global constant and use the same value everywhere?

e5l (approved):
LGTM, please check the comment

ktor-server/ktor-server-host-common/jvmAndNix/src/io/ktor/server/engine/CommandLine.kt
marychatte: Thanks, I deleted the setting of default values here because `shutdownGracePeriod` and `shutdownTimeout` already have default values in [Configuration](https://github.com/ktorio/ktor/blob/a08b7c8f520ca41f588e00f8c0cee9a83d1e346e/ktor-server/ktor-server-host-common/jvmAndNix/src/io/ktor/server/engine/ApplicationEngine.kt#L49)

PR #3406, opened 2023-02-14
KTOR-5570 Fixed incorrect handling of private cache directive in HttpCachePlugin

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-5570


e5l (approved):
lgtm

PR #3407, opened 2023-02-15
Small cleanup of kotlinx serializer

rsinukov:


marychatte (approved):
Good, looks clearer now

PR #3411, opened 2023-02-16
KTOR-5525 Keep external coroutineContext in testApplication

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-5525


e5l (approved):
lgtm

PR #3414, opened 2023-02-19
[QUIC] QUIC packetezation

Mr3zee:
**Subsystem**
`ktor-network`

**Solution**
Handlers for QUIC packets read and write operations

### DIFF
--- a/ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/bytes/ByteUtils.kt
+++ b/ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/bytes/ByteUtils.kt
@@ -3,6 +3,7 @@
  */
 
 @file:Suppress("NOTHING_TO_INLINE", "unused")
+@file:OptIn(ExperimentalUnsignedTypes::class)
 
 package io.ktor.network.quic.bytes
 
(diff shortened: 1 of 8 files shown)


ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/bytes/ByteUtils.kt
```diff
@@ -2,17 +2,69 @@
...
+internal inline fun UInt8.toUInt16(): UInt16 = toUShort()
+
+internal inline fun UInt64.toUInt8(): UInt8 = toUByte()
+internal inline fun UInt32.toUInt8(): UInt8 = toUByte()
+internal inline fun UInt16.toUInt8(): UInt8 = toUByte()
+internal inline fun UInt64.toUInt16(): UInt16 = toUShort()
+internal inline fun UInt32.toUInt16(): UInt16 = toUShort()
+internal inline fun UInt64.toUInt32(): UInt32 = toUInt()
+
+@OptIn(ExperimentalUnsignedTypes::class)
```
e5l, line 31: Could be replaced with file:
@file:OptIn(ExperimentalUnsignedTypes::class)

ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/consts/NumericConsts.kt
```diff
@@ -2,10 +2,13 @@
  * Copyright 2014-2023 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
  */
 
-package io.ktor.network.quic.util
+package io.ktor.network.quic.consts
```
e5l, line 5: Could you tell me why a separate packet is needed? I would keep it in `quic`

e5l (approved):
some end-to-end tests would be great at some point

ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/bytes/ByteUtils.kt
Mr3zee: I'm not sure about compiling, but IDE shows me that this approach does not work (it still requires opt-in for separate functions)

ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/consts/NumericConsts.kt
Mr3zee: I just that I like to keep everything in separate packages and not mix files, easier to navigate for me

Mr3zee: but maybe for this particular file `util` directory is indeed better

ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/bytes/ByteUtils.kt
Mr3zee: checked, guess it's IDE bug

Mr3zee:
> some end-to-end tests would be great at some point

yes, as soon as I fill all stubs, I will cover this functionality with tests

PR #3415, opened 2023-02-20
KTOR-5586 Fix DefaultRequest overwriting explicitly provided http protocol

martingalda:
Fix for [KTOR-5586](https://youtrack.jetbrains.com/issue/KTOR-5586).

The solution is to move the source of truth for the protocol into a nullable property (`protocolOrNull`) while delegating the original (`protocol`) to it, defaulting to `URLProtocol.HTTP` as before. This way the original behavior remains unchanged, but if necessary it's now possible to tell whether or not http came about as the default, or if it was provided explicitly: In the first case `protocolOrNull == null` while in the latter `protocolOrNull == URLProtocol.HTTP`.

I believe the only change to the public api brought about by this (aside from the new `protocolOrNull` property) is that `protocol` is now nullable in the `URLBuilder` constructor, which shouldn't be a problem i believe?


e5l (approved):
Hey @martingalda, sorry for the long delay.

Thanks for the PR! LGTM. Will be merged after CI check

PR #3416, opened 2023-02-21
KTOR-5535 Use the longest successful routing trait on mismatch

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-5535


e5l (approved):
lgtm

PR #3417, opened 2023-02-22
KTOR-5591 Fix takeFrom(URL) method with underscore in host

rsinukov:
java `URI` does not support `_` in host, so it becomes `null`

https://youtrack.jetbrains.com/issue/KTOR-5591


e5l (approved):
LGTM, maybe we should use `toString` as the default behavior.

PR #3418, opened 2023-02-22
KTOR-5466 Connect timeout is not respected when using the HttpRequestRetry plugin

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-5466


e5l (approved):
lgtm

PR #3419, opened 2023-02-27
[QUIC] Connections

Mr3zee:
**Subsystem**
ktor-network

**Solution**
Foundation for usage of QUIC connections


ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/connections/ConnectionID.kt
```diff
@@ -0,0 +1,28 @@
+/*
+ * Copyright 2014-2023 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.network.quic.connections
+
+internal typealias ConnectionID = ByteArray
```
e5l, line 7: Maybe value class could be better

ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/connections/ConnectionID.kt
```diff
@@ -0,0 +1,28 @@
...
+
+internal infix fun ConnectionID?.neq(other: ConnectionID?): Boolean {
+    return !eq(other)
+}
+
+internal infix fun ConnectionID?.eq(other: ConnectionID?): Boolean {
+    if (this == null && other == null) return true
```
e5l, line 14: ByteArray.contentEquals: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/content-equals.html

e5l (approved):
lgtm in general. I'm not happy with many `TODO` stubs, but the PR is big, so it should be OK.

Mr3zee:
> I'm not happy with many `TODO` stubs, but the PR is big, so it should be OK.

Yeah, most of them are stubs, some are just remainders for future me in the places, where I can overlook something important, I'll get rid of them in the upcoming reviews

ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/connections/ConnectionID.kt
Mr3zee: oops

PR #3420, opened 2023-02-27
KTOR-2677 Support preCompressed with static resources

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-2677


ktor-server/ktor-server-core/jvm/src/io/ktor/server/http/content/PreCompressed.kt
```diff
@@ -0,0 +1,137 @@
...
+ * Supported pre compressed file types and associated extensions
+ *
+ * **See Also:** [Accept-Encoding](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Encoding)
+ */
+public enum class CompressedFileType(public val extension: String, public val encoding: String = extension) {
+    // https://www.theregister.co.uk/2015/10/11/googles_bro_file_format_changed_to_br_after_gender_politics_worries/
+    BROTLI("br"),
+    GZIP("gz", "gzip");
+
+    @Deprecated("This will become internal")
```
e5l, line 26: Should we have a replacement or link to the docs?

e5l (approved):

ktor-server/ktor-server-core/jvm/src/io/ktor/server/http/content/PreCompressed.kt
rsinukov: No, it's used only internally and there is no need for it to be public

ktor-server/ktor-server-core/jvm/src/io/ktor/server/http/content/PreCompressed.kt
```diff
@@ -0,0 +1,137 @@
...
+            appendFiltered(original.headers) { name, _ -> !name.equals(HttpHeaders.ContentLength, true) }
+            append(HttpHeaders.ContentEncoding, encoding)
+        }
+    }
+
+    override fun <T : Any> getProperty(key: AttributeKey<T>) = original.getProperty(key)
+    override fun <T : Any> setProperty(key: AttributeKey<T>, value: T?) = original.setProperty(key, value)
+}
+
+internal fun bestCompressionFit(
```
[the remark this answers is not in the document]
rsinukov: I just moved it from another file. It's internal anyway, so should be fine.

PR #3426, opened 2023-03-03
KTOR-5614 Support for CURLOPT_CAINFO and CURLOPT_CAPATH in ktor-client-curl

rsinukov:
Adding test is not easy, but the change is self-explanatory.

### DIFF
--- a/ktor-client/ktor-client-curl/desktop/src/io/ktor/client/engine/curl/internal/CurlMultiApiHandler.kt
+++ b/ktor-client/ktor-client-curl/desktop/src/io/ktor/client/engine/curl/internal/CurlMultiApiHandler.kt
@@ -109,6 +109,8 @@ internal class CurlMultiApiHandler : Closeable {
                 option(CURLOPT_SSL_VERIFYPEER, 0L)
                 option(CURLOPT_SSL_VERIFYHOST, 0L)
             }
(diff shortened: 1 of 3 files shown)


ktor-client/ktor-client-curl/desktop/src/io/ktor/client/engine/curl/internal/CurlMultiApiHandler.kt
```diff
@@ -109,6 +109,8 @@ internal class CurlMultiApiHandler : Closeable {
                 option(CURLOPT_SSL_VERIFYPEER, 0L)
                 option(CURLOPT_SSL_VERIFYHOST, 0L)
             }
+            request.caPath?.let { option(CURLOPT_CAPATH, it) }
```
e5l, line 112: Could you tell me if we can expose some block on `request` as well?

e5l (approved):
lgtm

ktor-client/ktor-client-curl/desktop/src/io/ktor/client/engine/curl/internal/CurlMultiApiHandler.kt
rsinukov: we can in theory, but an API would not be nice. `easyHandle` here is a `COpaquePointer` and giving it back to the user can be confusing.

PR #3427, opened 2023-03-03
KTOR-5532 ByteBufferChannel throws exception when HttpCache, ContentEncoding and onDownload are used

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-5532


e5l (approved):
lgtm

PR #3408, opened 2023-02-15
Make HttpStatusCode comparable

InsanusMokrassar:
## Subsystem

Client/Server

## Motivation

It is much more useful to compare status codes between each other. For example:

```kotlin
httpResponse.status < HttpStatusCode.OK
```

Currently, in case, if I want to compare status codes I should use next comparison:

```kotlin
httpResponse.status.value < HttpStatusCode.OK.value
```

## Solution

In my PR `HttpStatusCode` will implements `Comparable` interface and realize `compareTo` with any other `HttpStatusCode`

PR #3430, opened 2023-03-06
KTOR-5629 Make HttpStatusCode comparable

InsanusMokrassar:
https://youtrack.jetbrains.com/issue/KTOR-5629/Comparable-HttpStatusCode

## Subsystem

Client/Server

## Motivation

It is much more useful to compare status codes between each other. For example:

```kotlin
httpResponse.status < HttpStatusCode.OK
```

Currently, in case, if I want to compare status codes I should use next comparison:

```kotlin
httpResponse.status.value < HttpStatusCode.OK.value
```

## Solution

In my PR `HttpStatusCode` will implements `Comparable` interface and realize `compareTo` with any other `HttpStatusCode`


InsanusMokrassar:
https://github.com/ktorio/ktor/pull/3408#issuecomment-1456121781

rsinukov (approved):

PR #3432, opened 2023-03-08
KTOR-5502 Fixed request in Darwin engine deadlocks on macOS since 2.2.2

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-5502

### DIFF
--- a/ktor-client/ktor-client-darwin/darwin/src/io/ktor/client/engine/darwin/DarwinClientEngine.kt
+++ b/ktor-client/ktor-client-darwin/darwin/src/io/ktor/client/engine/darwin/DarwinClientEngine.kt
@@ -18,7 +18,11 @@ import kotlin.coroutines.*
 
 @OptIn(InternalAPI::class)
 internal class DarwinClientEngine(override val config: DarwinClientEngineConfig) : HttpClientEngineBase("ktor-darwin") {
-    private val requestQueue: NSOperationQueue? = NSOperationQueue.currentQueue()
+
+    private val requestQueue: NSOperationQueue? = when (val queue = NSOperationQueue.currentQueue()) {
+        NSOperationQueue.mainQueue -> NSOperationQueue()
+        else -> queue
+    }
 
     override val dispatcher = Dispatchers.Unconfined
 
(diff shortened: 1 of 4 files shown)


e5l (changes requested):
Should we apply the same fixes for a legacy?

ktor-client/ktor-client-darwin/darwin/src/io/ktor/client/engine/darwin/DarwinClientEngine.kt
```diff
@@ -18,7 +18,11 @@ import kotlin.coroutines.*
 
 @OptIn(InternalAPI::class)
 internal class DarwinClientEngine(override val config: DarwinClientEngineConfig) : HttpClientEngineBase("ktor-darwin") {
-    private val requestQueue: NSOperationQueue? = NSOperationQueue.currentQueue()
+
+    private val requestQueue: NSOperationQueue? = when (val queue = NSOperationQueue.currentQueue()) {
```
e5l, line 22: Maybe we should always use new queue by default

e5l:
Please also check codestyle

e5l (approved):
lgtm

ktor-client/ktor-client-darwin/darwin/test/DarwinEngineTest.kt
```diff
@@ -20,10 +19,36 @@ import kotlin.test.*
 
 class DarwinEngineTest {
 
-    val testCoroutineContext: CoroutineContext = Dispatchers.Default
+    @Test
+    fun testRequestInRunBlockingDispatchersDefault() = runBlocking(Dispatchers.Default) {
```
wkornewald, line 23: Why is Dispatchers.Default needed here and why can't Ktor internally do the right thing, so it's safe to be called from any dispatcher/thread?

rsinukov: What do you mean by why is it needed? This tests that request can be made on default dispatcher. Below are the tests for other dispatchers, and no dispatcher.

wkornewald: Sorry, I didn't see the one without dispatcher. But in our unit tests everything just blocks. Only Default doesn't block.

rsinukov: Which version are you using? This fixed will be released in 2.3.0. You can try [EAP version](https://ktor.io/eap/), if you want to check the fix.

wkornewald: I first tried with a local build, but with 2.3.0-eap-631 the issue is indeed fixed and apparently my repo had an outdated remote for the `main` branch. 🤦  Sorry for the confusion. Everything works as expected.

rsinukov: Great, thanks for confirming.

PR #3436, opened 2023-03-09
KTOR-4515 set realm before it's used to create credential

sproctor:
**Subsystem**
ktor-client-auth 

**Motivation**
https://youtrack.jetbrains.com/issue/KTOR-4514/DigestAuthProvider-realm-sent-from-the-server-doesnt-participate-in-the-computation-of-response

Connecting to Axis cameras fails without this fix

**Solution**
Prior to this change, the realm used in the `credential` was always the class property. If the realm is unknown, it should be read from the headers. This was happening, but after the `credential` was created.

This change uses the constructed `realm` in the `credential` instead of the class property.

### DIFF
--- a/ktor-client/ktor-client-plugins/ktor-client-auth/common/src/io/ktor/client/plugins/auth/providers/DigestAuthProvider.kt
+++ b/ktor-client/ktor-client-plugins/ktor-client-auth/common/src/io/ktor/client/plugins/auth/providers/DigestAuthProvider.kt
@@ -163,7 +163,7 @@ public class DigestAuthProvider(
         val realm = realm ?: authHeader?.let { auth ->
             (auth as? HttpAuthHeader.Parameterized)?.parameter("realm")
         }
-        
+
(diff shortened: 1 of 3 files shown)


rsinukov:
And I see that CodeStyle task fails. Please run `./gradlew formatKotlin` to fix it.

sproctor:
I'm not sure how to write that test. there's an example at the youtrack report, but I don't think hitting an external server is ok. A bit of direction would be nice and I'll try to write the test next week.

sproctor:
Sorry about the noise in `HttpStatusCode`. I have no idea where that came from, I didn't edit that file.

The bug is that the token is incorrect when the server specifies a realm, but the client doesn't provide one. I've added a weak test for this. It just asserts that the response token matches what was produced locally for me. This doesn't test the path where `qop` is set because that is much more complicated.

ninja-:
I just checked it also fixed Dahua support. thanks :)

PR #3438, opened 2023-03-13
KTOR-5699 Fix reading response of HEAD request

yassenb:
**Subsystem**
client

**Motivation**
A lot of code has the same implementation handling all HTTP requests and this includes reading the body of a response. In the case of a HEAD request this can break when the Content-Length header has a non-zero value

**Solution**
Do not check the response length matches the Content-Length header for HEAD requests


rsinukov (changes requested):
LGTM, but can you please add a test?

yassenb:
I thought about adding a test but I don't see where the tests for `DefaultTransform.kt` are, nor do I see an existing test for the functionality I'm changing (there should be a test that an `IllegalArgumentException` is raised when content length doesn't match the body length). My familiarity with the project is not good enough to dig deeper into the tests infrastructure so I would suggest either:
  * someone more familiar add a test
  * merge without a test

rsinukov:
There is a test for this transformation in `ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/DefaultTransformTest.kt`. You can add it there.

rsinukov:
I reopened this PR with the test added in #3465

PR #3439, opened 2023-03-13
Fix freezing of Apache5 engine

rsinukov:
Sometimes `channel.availableForWrite` is 0 and in this case Apache would not trigger a new read


e5l (approved):
wow, LGTM

PR #3440, opened 2023-03-14
KTOR-5652 Add proper cause to WebSocketException in Java engine

joffrey-bion:
**Subsystem**
Client, Web Sockets

**Motivation**
The web socket client throws `WebSocketException` without cause in `onError` in the Java engine.
See https://youtrack.jetbrains.com/issue/KTOR-5652/Missing-cause-for-WebSocketException-in-JavaHttpWebSocket

**Solution**
Add the cause to the exception, so the stacktraces can show the underlying problem when the message is not sufficient (or absent).

### DIFF
--- a/ktor-client/ktor-client-java/jvm/src/io/ktor/client/engine/java/JavaHttpWebSocket.kt
+++ b/ktor-client/ktor-client-java/jvm/src/io/ktor/client/engine/java/JavaHttpWebSocket.kt
@@ -183,7 +183,7 @@ internal class JavaHttpWebSocket(
     }.asCompletableFuture()
 
     override fun onError(webSocket: WebSocket, error: Throwable) {
-        val cause = WebSocketException("${error.message}")
+        val cause = WebSocketException(error.message ?: "web socket failed", error)
         _incoming.close(cause)
         _outgoing.cancel()
         socketJob.complete()
--- a/ktor-client/ktor-client-core/api/ktor-client-core.api
+++ b/ktor-client/ktor-client-core/api/ktor-client-core.api
@@ -900,6 +900,7 @@ public final class io/ktor/client/plugins/websocket/WebSocketCapability : io/kto
 
 public final class io/ktor/client/plugins/websocket/WebSocketException : java/lang/IllegalStateException {
 	public fun <init> (Ljava/lang/String;)V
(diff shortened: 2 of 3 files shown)


e5l (approved):
lgtm

joffrey-bion:
I'm a bit confused by the CI tests, it looks like `@JvmOverloads` fails on K/Native compilations, although this annotation is supposed to be available to `common` code 🤔 I can just turn this into a second constructor, but I believe the annotation is supposed to work.

e5l:
Could you also log issue for K/Native?

ktor-client/ktor-client-java/jvm/src/io/ktor/client/engine/java/JavaHttpWebSocket.kt
```diff
@@ -183,7 +183,7 @@ internal class JavaHttpWebSocket(
     }.asCompletableFuture()
 
     override fun onError(webSocket: WebSocket, error: Throwable) {
-        val cause = WebSocketException("${error.message}")
+        val cause = WebSocketException("$error.message}", error)
```
rsinukov, line 186: `"$error.message}"` -> `"${error.message}"` or maybe add WebSocket-specific message? Something generic like `WebSocket failed`

joffrey-bion: Oh... thanks for spotting this 🤦 I had actually put `error.message ?: "generic message here"` and then I reverted to keep the PR focused, but reverting manually is a bad idea as we can see!

I like the `web socket failed` generic message. Do you usually spell it in one word in Ktor's material?

joffrey-bion: I think we would benefit from adding the cause's message to that new exception message, so if some code just prints the message of Ktor's exception (and not the cause) then the error still contains the information.

And since the generic message doesn't bring any useful information on top of the exception name, we could actually only use it in case the cause's message is absent. Something like:

```kotlin
val cause = WebSocketException(error.message ?: "web socket failed", error)
```

rsinukov: Yes, sounds good.

PR #3442, opened 2023-03-14
KTOR-5658 Add possibility to use multiple configuration files

lkurcius:
**Subsystem**

**Motivation**
This PR concerns ticket KTOR-5658 and gives the possibility to use more than one configuration file. This code will make it possible to run an application like this: `java -jar simple-app.jar -config=application.conf -config=applicationFoo.conf -config=applicationBar.conf`

This solution would be useful for deployment flows where basically configuration is the same for all environments (dev/stage/prod), but there is a need to change some properties. For example, it would look like this:

- for dev env: `java -jar simple-app.jar -config=application.conf -config=application-dev.conf`
- for stage env: `java -jar simple-app.jar -config=application.conf -config=application-stage.conf`
- for prod env: `java -jar simple-app.jar -config=application.conf -config=application-prod.conf`

**Solution**
The mechanism of the solution is described in the ticket KTOR-5658, but basically, it merges configuration files (.conf) from left to right where the most-left file is a base which is modified by merging following configuration files.


rsinukov:
Hi and thanks for the PR. The file `CommandLineJvm.kt` used an older version of the code and didn't support reading `YAML` and other formats. I updated your implementation and pushed my changes here.

rsinukov (approved):
LGTM.
@marychatte please take another look

PR #3444, opened 2023-03-14
KTOR-5265 Static files generic modifier

rsinukov:


ktor-server/ktor-server-core/jvm/src/io/ktor/server/http/content/StaticContent.kt
```diff
@@ -72,22 +73,32 @@ public class StaticContentConfig<Resource : Any> internal constructor() {
...
     public fun cacheControl(block: (Resource) -> List<CacheControl>) {
         cacheControl = block
     }
 
+    /**
+     * Configures modification of a call for requested content.
+     * Useful to add headers to the response, such as [HttpHeaders.ETag]
+     * For files, [Resource] is a requested [File].
+     * For resources, [Resource] is a [URL] to requested resource.
+     */
```
marychatte, line 97: Shouldn't there be articles here?
Configures modification of a call for **the** requested content.
For resources, [Resource] is a [URL] to **the** requested resource.

marychatte (approved):
lgtm

PR #3450, opened 2023-03-19
KTOR-5672 Fix erroneous trace log about expired websocket pings

nesk:
**Subsystem**
Server, websocket

**Motivation**
[KTOR-5672](https://youtrack.jetbrains.com/issue/KTOR-5672) The trace log was written even when the `PONG` frame was received in time after the `PING` one.

**Solution**
Shouldn't we put the log in the condition body? Currently, this is confusing.


rsinukov (approved):
Thanks for the catch. LGTM

PR #3452, opened 2023-03-20
KTOR-5684 AutoHead should dispose response body

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-5684/AutoHead-should-dispose-response-body

### DIFF
--- a/ktor-server/ktor-server-core/jvm/src/io/ktor/server/http/content/StaticContent.kt
+++ b/ktor-server/ktor-server-core/jvm/src/io/ktor/server/http/content/StaticContent.kt
@@ -33,6 +33,7 @@ private val StaticContentAutoHead = createRouteScopedPlugin("StaticContentAutoHe
 
     on(ResponseBodyReadyForSend) { call, content ->
         check(call.request.local.method == HttpMethod.Head)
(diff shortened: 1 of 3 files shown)


ktor-server/ktor-server-core/jvm/src/io/ktor/server/http/content/StaticContent.kt
```diff
@@ -33,6 +33,7 @@ private val StaticContentAutoHead = createRouteScopedPlugin("StaticContentAutoHe
 
     on(ResponseBodyReadyForSend) { call, content ->
         check(call.request.local.method == HttpMethod.Head)
+        if (content is OutgoingContent.ReadChannelContent) content.readFrom().cancel(null)
```
marychatte, line 36: Why is it only for `ReadChannelContent`, shouldn't it also be for others `OutgoingContent` such as `ByteArrayContent`?

rsinukov: Other types of content do not hold resources that need to be closed.

marychatte (approved):
lgtm

PR #3453, opened 2023-03-20
KTOR-5665 CallLogging: logs caused by an exception are suppressed when mdc provider is configured

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-5665

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-call-logging/jvm/src/io/ktor/server/plugins/callloging/MDCHook.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-call-logging/jvm/src/io/ktor/server/plugins/callloging/MDCHook.kt
@@ -5,6 +5,7 @@
 package io.ktor.server.plugins.callloging
 
 import io.ktor.server.application.*
+import io.ktor.server.response.*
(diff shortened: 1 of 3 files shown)


ktor-server/ktor-server-plugins/ktor-server-call-logging/jvm/src/io/ktor/server/plugins/callloging/MDCHook.kt
```diff
@@ -20,3 +21,12 @@ internal fun MDCHook(phase: PipelinePhase) = object : Hook<suspend (ApplicationC
         }
     }
 }
+
+internal object ResponseSent : Hook<suspend (ApplicationCall) -> Unit> {
```
marychatte, line 25: Why we can't add `suspend` to the existing `ResponseSent` [hook](https://github.com/ktorio/ktor/blob/c333fb95143cbb9dc94857e42295bc0d0aeaea23/ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/application/hooks/CommonHooks.kt#L108)? Can it break something??

marychatte (approved):
lgtm

PR #3455, opened 2023-03-21
KTOR-5636 Fixed nmap crashes CIO server

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-5636

TODO: test

### DIFF
--- a/ktor-server/ktor-server-cio/jvmAndNix/src/io/ktor/server/cio/backend/HttpServer.kt
+++ b/ktor-server/ktor-server-cio/jvmAndNix/src/io/ktor/server/cio/backend/HttpServer.kt
@@ -13,6 +13,7 @@ import io.ktor.server.engine.internal.*
 import io.ktor.util.*
 import io.ktor.util.logging.*
 import io.ktor.utils.io.core.*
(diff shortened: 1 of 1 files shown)


ktor-server/ktor-server-cio/jvmAndNix/src/io/ktor/server/cio/backend/HttpServer.kt
```diff
@@ -60,7 +60,11 @@ public fun CoroutineScope.httpServer(
 
             try {
                 while (true) {
-                    val client: Socket = server.accept()
+                    val client: Socket = try {
+                        server.accept()
```
e5l, line 64: Could you tell me why we should ignore any IOExceptions?

I would add some small random delay and counter with limit to prevent dead loop

rsinukov: Because now clients can connect and immediately drop the connection, crashing the whole server, like the case with nmap.

e5l (approved):
lgtm

PR #3456, opened 2023-03-21
KTOR-5650 Prefer Node to browser behaviour

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-5650

### DIFF
--- a/ktor-client/ktor-client-core/js/src/io/ktor/client/engine/js/compatibility/Utils.kt
+++ b/ktor-client/ktor-client-core/js/src/io/ktor/client/engine/js/compatibility/Utils.kt
@@ -23,10 +23,9 @@ internal suspend fun commonFetch(
         controller.abort()
     }
 
-    val promise: Promise<org.w3c.fetch.Response> = if (PlatformUtils.IS_BROWSER) {
-        fetch(input, init)
-    } else {
-        jsRequireNodeFetch()(input, init)
+    val promise: Promise<org.w3c.fetch.Response> = when (PlatformUtils.platform) {
+        Platform.Browser -> fetch(input, init)
+        else -> jsRequireNodeFetch()(input, init)
     }
 
     promise.then(
@@ -40,21 +39,21 @@ internal suspend fun commonFetch(
 }
 
 internal fun AbortController(): AbortController {
-    return if (PlatformUtils.IS_BROWSER) {
-        js("new AbortController()")
-    } else {
-        @Suppress("UNUSED_VARIABLE")
-        val controller = js("eval('require')('abort-controller')")
-        js("new controller()")
+    return when (PlatformUtils.platform) {
+        Platform.Browser -> js("new AbortController()")
+        else -> {
+            @Suppress("UNUSED_VARIABLE")
+            val controller = js("eval('require')('abort-controller')")
+            js("new controller()")
+        }
     }
 }
 
 internal fun CoroutineScope.readBody(
     response: org.w3c.fetch.Response
-): ByteReadChannel = if (PlatformUtils.IS_BROWSER) {
-    readBodyBrowser(response)
-} else {
-    readBodyNode(response)
+): ByteReadChannel = when (PlatformUtils.platform) {
+    Platform.Node -> readBodyNode(response)
+    else -> readBodyBrowser(response)
 }
 
 private fun jsRequireNodeFetch(): dynamic = try {
--- a/ktor-utils/common/src/io/ktor/util/PlatformUtils.kt
+++ b/ktor-utils/common/src/io/ktor/util/PlatformUtils.kt
@@ -4,13 +4,22 @@
 
 package io.ktor.util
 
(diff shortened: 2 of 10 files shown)


ktor-client/ktor-client-core/js/src/io/ktor/client/engine/js/compatibility/Utils.kt
```diff
@@ -23,10 +23,10 @@ internal suspend fun commonFetch(
         controller.abort()
     }
 
-    val promise: Promise<org.w3c.fetch.Response> = if (PlatformUtils.IS_BROWSER) {
-        fetch(input, init)
-    } else {
+    val promise: Promise<org.w3c.fetch.Response> = if (PlatformUtils.IS_NODE) {
```
wkornewald, line 26: As described in my bug report, it would be even better if you had a

```kotlin
when (PlatformUtils.platform) {
    Node -> ...
    Browser -> ...
}
```

Then the logic for this decision would live in exactly one place and the next developer won't be able to re-introduce this bug by checking `IS_BROWSER` first again.

ktor-utils/common/src/io/ktor/util/PlatformUtils.kt
```diff
@@ -4,13 +4,22 @@
...
-public expect object PlatformUtils {
-    public val IS_BROWSER: Boolean
-    public val IS_NODE: Boolean
-    public val IS_JVM: Boolean
-    public val IS_NATIVE: Boolean
+public object PlatformUtils {
+    public val IS_BROWSER: Boolean = platform == Platform.Browser
+    public val IS_NODE: Boolean = platform == Platform.Node
+    public val IS_JVM: Boolean = platform == Platform.Jvm
+    public val IS_NATIVE: Boolean = platform == Platform.Native
```
wkornewald, line 11: Since these lead to incorrect code, shouldn't they be deprecated?

Also, a more general comment but I don't suggest that it should be changed in this PR: I'm surprised that you can check for `IS_JVM` etc. in common code at all and that in JS code the `Platform` enum also contains `Jvm` which is an impossible case. It feels like `expect`/`actual` are the cleaner solution for those cases and the only real place where you need a runtime check is for Node vs. Browser.

e5l: We can't drop fields from `enum` based on expect/actual model: we need to refer to all platforms in the common code. There is also a case when we need to check if the platform is native or JVM in the common code, for instance, to use UDP or raw sockets.

wkornewald: What I meant was to remove the enum in all other platforms and only have it in JS. And for those cases where you need to use a UDP or raw socket you refer to an `expect fun` which does the platform-specific thing. Is there anything preventing you from just doing it this way?

e5l: We initially started with the proposed approach. It led to copy-paste expect-actual implementations for many cases, from exception mapping to concurrency. It's hard to maintain.

Making a new platform for different target subsets would be perfect, but it works not so well with tooling.

rsinukov: I don't want to deprecate them, because writing `if (PlatformUtils.IS_JVM)` is better then `if (PlatformUtils.platform == Platform.Jvm)` and in the end, we will end up with similar extensions anyway. Also, now `IS_NODE` and `IS_BROWSER` can't be true at the same time, so the problem with flags should be fixed.

e5l (approved):
lgtm

ktor-utils/js/src/io/ktor/util/PlatformUtilsJs.kt
```diff
@@ -5,17 +5,16 @@
...
 
-    public actual val IS_JVM: Boolean = false
-    public actual val IS_NATIVE: Boolean = false
-    public actual val IS_DEVELOPMENT_MODE: Boolean = false
-    public actual val IS_NEW_MM_ENABLED: Boolean = true
-}
+public actual val PlatformUtils.platform: Platform
+    get() {
+        val hasNodeApi = js(
+            "typeof process !== 'undefined' && process.versions != null && process.versions.node != null"
```
wkornewald, line 17: After testing this in Electron the check turned out to return false even though we have the Node API. In Electron at this point there can be two `process` objects and the correct one is defined on `window.process`. So we'd have to first detect which process object to use and after that the remaining `.versions.node` check can be done.

rsinukov: Do you mind creating a ticket so that it doesn't get lost?

wkornewald: Yes sure: https://youtrack.jetbrains.com/issue/KTOR-5906/Electron-Node.js-detection-doesnt-work-correctly

PR #3463, opened 2023-03-28
KTOR-829 Support `100 Continue` on server side

marychatte:
Server
[KTOR-829](https://youtrack.jetbrains.com/issue/KTOR-829/Support-100-Continue)
[rfc7231](https://www.rfc-editor.org/rfc/rfc7231#section-5.1.1)

### DIFF
--- a/ktor-server/ktor-server-cio/build.gradle.kts
+++ b/ktor-server/ktor-server-cio/build.gradle.kts
@@ -7,7 +7,6 @@ kotlin.sourceSets {
             api(project(":ktor-http:ktor-http-cio"))
             api(project(":ktor-shared:ktor-websockets"))
             api(project(":ktor-network"))
-            implementation(project(":ktor-server:ktor-server-core"))
         }
     }
     jvmAndNixTest {
--- a/ktor-server/ktor-server-cio/build.gradle.kts
+++ b/ktor-server/ktor-server-cio/build.gradle.kts
@@ -7,6 +7,7 @@ kotlin.sourceSets {
             api(project(":ktor-http:ktor-http-cio"))
             api(project(":ktor-shared:ktor-websockets"))
             api(project(":ktor-network"))
+            implementation(project(":ktor-server:ktor-server-core"))
         }
     }
     jvmAndNixTest {
--- a/ktor-server/ktor-server-cio/jvmAndNix/src/io/ktor/server/cio/CIOApplicationEngine.kt
+++ b/ktor-server/ktor-server-cio/jvmAndNix/src/io/ktor/server/cio/CIOApplicationEngine.kt
@@ -129,10 +129,9 @@ public class CIOApplicationEngine(
         call.request.pipeline.intercept(expectedHeaderPhase) {
             val request = call.request
             val version = HttpProtocolVersion.parse(request.httpVersion)
-            val alreadyRead = request.receiveChannel().totalBytesRead > 0
-            val noBody = request.headers[HttpHeaders.ContentLength]?.let { it.toInt() == 0 } ?: true
+            val hasBody = hasBody(request)
 
-            if (expectHeader == null || version == HttpProtocolVersion.HTTP_1_0 || noBody || alreadyRead) {
+            if (expectHeader == null || version == HttpProtocolVersion.HTTP_1_0 || !hasBody) {
                 return@intercept
             }
 
@@ -155,6 +154,12 @@ public class CIOApplicationEngine(
             val connection = if (closeConnection) "close" else "keep-alive"
             call.response.headers.append(HttpHeaders.Connection, connection)
         }
+    }
+
+    private fun hasBody(request: CIOApplicationRequest): Boolean {
+        val contentLength = request.headers[HttpHeaders.ContentLength]?.toInt()
+        val transferEncoding = request.headers[HttpHeaders.TransferEncoding]
+        return transferEncoding != null || (contentLength != null && contentLength > 0)
     }
 
     private suspend fun ServerRequestScope.handleRequest(request: Request) {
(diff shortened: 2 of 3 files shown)


ktor-server/ktor-server-cio/jvmAndNix/src/io/ktor/server/cio/CIOApplicationEngine.kt
```diff
@@ -113,6 +117,46 @@ public class CIOApplicationEngine(
...
+
+        val expectHeader = call.request.headers[HttpHeaders.Expect]?.lowercase()
+        var closeConnection = true
+
+        val expectedHeaderPhase = PipelinePhase("ExpectedHeaderPhase")
+        call.request.pipeline.insertPhaseBefore(ApplicationReceivePipeline.Before, expectedHeaderPhase)
+        call.request.pipeline.intercept(expectedHeaderPhase) {
+            val request = call.request
+            val version = HttpProtocolVersion.parse(request.httpVersion)
+            val alreadyRead = request.receiveChannel().totalBytesRead > 0
```
rsinukov, line 132: Isn't `receiveChannel` calls `receive()` that triggers `ApplicationReceivePipeline`?

ktor-server/ktor-server-cio/jvmAndNix/src/io/ktor/server/cio/CIOApplicationEngine.kt
```diff
@@ -113,6 +117,46 @@ public class CIOApplicationEngine(
...
+            val noBody = request.headers[HttpHeaders.ContentLength]?.let { it.toInt() == 0 } ?: true
```
rsinukov, line 133: `Content-Length` can be null even when there is a body. For example, in chunked requests

ktor-server/ktor-server-cio/jvmAndNix/src/io/ktor/server/cio/CIOApplicationEngine.kt
rsinukov: Also, how body can be read before this pipeline is executed?

rsinukov (changes requested):
Great job! Some comments

ktor-server/ktor-server-cio/jvmAndNix/src/io/ktor/server/cio/CIOApplicationEngine.kt
marychatte: Thanks for the comments. I wanted to check if a server has already received some or all of the message body for the request. Delete it now because it's okay to send 100-continue response in such case and I don't know a better solution now

ktor-server/ktor-server-cio/build.gradle.kts
```diff
@@ -7,6 +7,7 @@ kotlin.sourceSets {
             api(project(":ktor-http:ktor-http-cio"))
             api(project(":ktor-shared:ktor-websockets"))
             api(project(":ktor-network"))
+            implementation(project(":ktor-server:ktor-server-core"))
```
rsinukov, line 10: why is this dependency needed?

ktor-server/ktor-server-cio/jvmAndNix/src/io/ktor/server/cio/CIOApplicationEngine.kt
```diff
@@ -113,6 +117,51 @@ public class CIOApplicationEngine(
...
+            }
+        }
+
+        call.response.pipeline.intercept(ApplicationSendPipeline.Transform) {
+            if (expectHeader == null) {
+                return@intercept
+            }
+
+            val connection = if (closeConnection) "close" else "keep-alive"
+            call.response.headers.append(HttpHeaders.Connection, connection)
```
rsinukov, line 155: why do you need to modify the connection header here?

ktor-server/ktor-server-cio/jvmAndNix/src/io/ktor/server/cio/CIOApplicationEngine.kt
```diff
@@ -113,6 +117,51 @@ public class CIOApplicationEngine(
...
+        }
+    }
+
+    private fun hasBody(request: CIOApplicationRequest): Boolean {
```
rsinukov, line 159: @e5l Can you check please if this method is correct? Are there any other cases that we miss here?

ktor-server/ktor-server-cio/build.gradle.kts
marychatte: legacy, thanks for spotting

ktor-server/ktor-server-cio/jvmAndNix/src/io/ktor/server/cio/CIOApplicationEngine.kt
marychatte: I tried to implement the next case:

> A server that responds with a final status code before reading the
entire message body SHOULD indicate in that response whether it
intends to close the connection or continue reading and discarding
the request message (see [Section 6.6 of [RFC7230]](https://www.rfc-editor.org/rfc/rfc7230#section-6.6)).

rsinukov: I think it's redundant here, because handling of connection will happen anyway in the call handling.

ktor-server/ktor-server-cio/jvmAndNix/src/io/ktor/server/cio/CIOApplicationEngine.kt
marychatte:
```
The presence of a message-body in a request is signaled by the
inclusion of a Content-Length or Transfer-Encoding header field in
the request's message-headers.
```
https://www.rfc-editor.org/rfc/rfc2616.html#section-4.3

rsinukov (approved):
LGTM
@e5l please also take a look at it.

e5l (approved):
LGTM. We need to add some integrations tests with an update of internal Ktor version

PR #3470, opened 2023-04-01
Fix percentEncode to use UTF-8

ay4toh5i:
**Subsystem**
ktor-http

**Motivation**
The Content-Disposition header filename* content does not handle multi byte characters properly.
This is because `percentEncode` function used within `encodeContentDispositionAttribute`, does not encode the string to UTF-8.

**Solution**
Encode the string using UTF-8 representation, and add a test that handles multi byte characters.

### DIFF
--- a/ktor-http/common/test/io/ktor/tests/http/ContentDispositionTest.kt
+++ b/ktor-http/common/test/io/ktor/tests/http/ContentDispositionTest.kt
@@ -19,6 +19,16 @@ class ContentDispositionTest {
     }
 
     @Test
+    fun testMultiByteFilename() {
+        val attachment = ContentDisposition.Attachment.withParameter(
+            ContentDisposition.Parameters.FileNameAsterisk,
+            "a©あ𠮷.txt"
+        ).toString()
+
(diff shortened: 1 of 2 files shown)


ktor-http/common/test/io/ktor/tests/http/ContentDispositionTest.kt
```diff
@@ -18,6 +18,16 @@ class ContentDispositionTest {
         assertEquals("attachment; filename*=utf-8''%27%27malicious.sh%2500%27normal.txt", attachment)
     }
 
+    @Test
+    fun testMultiByteFilename() {
+        val attachment = ContentDisposition.Attachment.withParameter(
+            ContentDisposition.Parameters.FileNameAsterisk,
+            "a©あ𠮷.txt"
```
ay4toh5i, line 25: This is 1 to 4 bytes UTF-8 characters.

```
[0x61]                   // a
[0xC2, 0xA9]             // ©
[0xE3, 0x81, 0x82]       // あ
[0xF0, 0xA0, 0xAE, 0xB7] // 𠮷
```

e5l (approved):
lgtm

PR #3476, opened 2023-04-06
KTOR-2932 Fix cancelled and failed JS websockets not closing

Sejsel:
**Subsystem**
Client, JS, Websockets

**Motivation**
Currently, websockets opened by the JS implementation in the browser do not get properly closed in case of cancellation or failure.

This is caused by browser `websocket.close()` throwing `InvalidAccessError` with codes that are not 1000 or 3000-4999 (see documentation [here](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/close#exceptions)), the implementation used 1011.

This is also mentioned in KTOR-2932. This PR does not change the meaning of cancellations which is the main focus of KTOR-2932, it only addresses the major bug.

The websockets staying open can cause critical performance issues (and has caused them for us, exhausting websockets in nginx).

**Solution**
Use 1000 as the closing reason code in the JS implementation, it is likely the only reasonable option as using something from the 3000-4999 range would be unexpected.

### DIFF
--- a/ktor-client/ktor-client-core/js/src/io/ktor/client/plugins/websocket/JsWebSocketSession.kt
+++ b/ktor-client/ktor-client-core/js/src/io/ktor/client/plugins/websocket/JsWebSocketSession.kt
@@ -127,7 +127,9 @@ internal class JsWebSocketSession(
             if (cause == null) {
                 websocket.close()
             } else {
-                websocket.close(CloseReason.Codes.INTERNAL_ERROR.code, "Client failed")
+                // We cannot use INTERNAL_ERROR similarly to other WebSocketSession implementations here
+                // as sending it is not supported by browsers.
+                websocket.close(CloseReason.Codes.NORMAL.code, "Client failed")
             }
         }
     }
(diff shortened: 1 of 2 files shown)


rsinukov (changes requested):
Thanks for the PR. Can you please add test for this fix?

Sejsel:
> Thanks for the PR. Can you please add test for this fix?

I am sorry, I do not currently have the time to implement tests for this.

Running any tests locally always fails with the default configuration. There also do not seem to be any related tests that I could easily modify for this, so I do not even know where tests for this would go.

I could not even find how to enable highlighting in the changed file for a while (changing `ktor.ide.jvmAndCommonOnly` to false in `gradle.properties` worked, but it doesn't seem to be documented anywhere, the CONTRIBUTING.md guide is fairly limited and I could not find any other documentation on the build process for this project).

rsinukov:
That's fair. I'll take it from here. Thanks again!

rsinukov (approved):

Sejsel:
@rsinukov, thank you for finishing this!

PR #3480, opened 2023-04-10
Fix curl publishing

e5l:


rsinukov (approved):

rsinukov:
Not sure if CI failure is relevant here.

PR #3487, opened 2023-04-17
Yaml config improvement

rsinukov:
KTOR-5797 YAML config does not support reading variables from itself 
KTOR-5796 Support optional properties in YAML

### DIFF
--- a/ktor-server/ktor-server-config-yaml/jvmAndNix/src/io/ktor/server/config/yaml/YamlConfig.kt
+++ b/ktor-server/ktor-server-config-yaml/jvmAndNix/src/io/ktor/server/config/yaml/YamlConfig.kt
@@ -33,7 +33,19 @@ public expect fun YamlConfig(path: String?): YamlConfig?
  * Implements [ApplicationConfig] by loading a configuration from a YAML file.
  * Values can reference to environment variables with `$ENV_VAR` or `"$ENV_VAR:default_value"` syntax.
  */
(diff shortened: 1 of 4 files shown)


ktor-server/ktor-server-config-yaml/jvmAndNix/src/io/ktor/server/config/yaml/YamlConfig.kt
```diff
@@ -43,15 +55,29 @@ public class YamlConfig(private val yaml: YamlMap) : ApplicationConfig {
         val parts = path.split('.')
         val yaml = parts.dropLast(1).fold(yaml) { yaml, part -> yaml[part] as? YamlMap ?: return null }
         val value = yaml[parts.last()] ?: return null
-        return ConfigValue(value, path)
+        return when (value) {
+            is YamlLiteral -> resolveValue(value.content, root)?.let { LiteralConfigValue(key = path, value = it) }
+
+            is YamlList -> {
+                val values = value.content.map { element ->
+                    element.asLiteralOrNull()?.content?.let { resolveValue(it, root) }
```
marychatte, line 63: Can't it be a nested list of lists?

rsinukov: we support only lists of strings and lists of objects, which are accessed through `configList(path)` method

PR #3488, opened 2023-04-17
KTOR-5813 Fix creating GMTDate from Calendar with a different timezone

e5l:
Fix [https://youtrack.jetbrains.com/issue/KTOR-5813/GMTDate-timestamp-doesnt-reflect-timezone-when-created-using-Calendar.toDate-method]


marychatte (approved):
lgtm, only check codestyle tests, please

PR #3554, opened 2023-04-19
KTOR-912 Fix adding Set-Cookie on every api request

marychatte:
[KTOR-912](https://youtrack.jetbrains.com/issue/KTOR-912) Sessions: Set-Cookie is added on every api request.


rsinukov (commented):
This can be a breaking change. Should we add a config property to opt-in for this behaviour?

rsinukov (approved):

PR #3558, opened 2023-04-20
KTOR-4275 API to use java.nio.Path as resources

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-4275/API-to-use-java.nio.Path-as-resources


e5l (changes requested):
Please change target branch to 3.0.0-eap

PR #3564, opened 2023-04-21
KTOR-5829 WebSockets: requests to a non-existing route cause server to lock up

rsinukov:
Not sure how to add a test for this, since the bug is observed only as a side effect of a lot of requests


ratiofu:
@rsinukov, perhaps the reproduction I had provided on the [original issue](https://youtrack.jetbrains.com/issue/KTOR-5829) could be converted to a (unit) test?

e5l (approved):
please add tests

PR #3575, opened 2023-04-26
Wasm

igoriakovlev:
Wasm target support

Not intended to merge into the main


e5l:
Please check the codestyle build and rebase the branch on `main`

e5l (changes requested):
please also drop all build hacks like `build_snapshot_train=true` and etc

e5l (changes requested):
Please check API dump and codestyle

igoriakovlev:
> Please check API dump and codestyle

Fixed

build.gradle.kts
```diff
@@ -92,18 +94,14 @@ apply(from = "gradle/compatibility.gradle")
 plugins {
     id("org.jetbrains.dokka") version "1.9.10" apply false
     id("org.jetbrains.kotlinx.binary-compatibility-validator") version "0.13.2"
-    id("kotlinx-atomicfu") version "0.22.0" apply false
+    id("kotlinx-atomicfu") version "0.22.0-wasm2" apply false
```
e5l, line 97: let's wait for the version without `wasm2` suffix

buildSrc/src/main/kotlin/WasmConfig.kt
```diff
@@ -0,0 +1,63 @@
...
+                        }
+                    }
+                )
+            }
+        }
+    }
+}
+
+private fun Project.configureWasmTestTasks() {
+    val shouldRunWasmBrowserTest = !hasProperty("teamcity") || hasProperty("enable-wasm-tests")
```
e5l, line 56: Please keep the same property name as for js. It looks like we don't need a separate.

ktor-http/wasmJs/src/io/ktor/http/URLBuilderWasm.kt
```diff
@@ -0,0 +1,30 @@
+/*
+* Copyright 2014-2023 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+*/
+
+package io.ktor.http
+
+import io.ktor.util.*
+
+private fun locationOrigin(): String = js(
```
e5l, line 9: It looks like we have 2 copies of the same code, so it will be needed to copy bug fixes multiple times as well. Could we make an umbrella source set to share code between js and wasm and expect actual for the diference?

e5l (changes requested):

ktor-http/wasmJs/src/io/ktor/http/URLBuilderWasm.kt
igoriakovlev: Fixed

buildSrc/src/main/kotlin/JsConfig.kt
```diff
@@ -14,7 +14,7 @@ fun Project.configureJs() {
         sourceSets {
             val jsTest by getting {
                 dependencies {
-                    implementation(npm("puppeteer", "*"))
+                    implementation(npm("puppeteer", "21.5.0"))
```
e5l, line 17: please extract to the config file

ktor-client/ktor-client-core/wasmJs/src/io/ktor/client/engine/js/JsClientEngine.kt
```diff
@@ -0,0 +1,172 @@
...
+
+@Suppress("UNUSED_PARAMETER")
+private fun createWebSocketNodeJs(
+    socketCtor: JsAny,
+    urlString_capturingHack: String,
+    headers_capturingHack: JsAny
+): WebSocket =
+    js("new socketCtor(urlString_capturingHack, { headers: headers_capturingHack })")
+
+internal class JsClientEngine(
```
e5l, line 36: WasmJs

e5l: or make it as expect-actual for shared source set

e5l:
@igoriakovlev, it looks like FileCacheTest is failing because of the coroutines update. We need to wait for @qwwdfsad to investigate

e5l (approved):

e5l:
Hey @igoriakovlev, thanks for the PR. LGTM, merged

PR #3576, opened 2023-04-26
KTOR-5855 Make System Property to Set outgoingToBeProcessed Size for WebSockets

rsinukov:


ktor-shared/ktor-websockets/common/src/io/ktor/websocket/DefaultWebSocketSession.kt
```diff
@@ -331,3 +331,5 @@ internal class DefaultWebSocketSessionImpl(
         private val EmptyPong = Frame.Pong(ByteArray(0), NonDisposableHandle)
     }
 }
+
+internal expect val outgoingChannelCapacity: Int
```
e5l, line 335: OUTGOING_CHANNEL_CAPACITY

e5l (approved):
lgtm

PR #3584, opened 2023-04-27
KTOR-5861 Allow passing custom agent and other options

wkornewald:
**Subsystem**
Client

**Motivation**
We have to configure the supported TLS versions and cipher suites and do other customizations by passing a custom `Agent` to node-fetch.

**Solution**
In order to allow flexibly passing arbitrary options to node-fetch this PR just exposes a `nodeOptions` config var which can be set arbitrarily (esp. `agent` which itself takes tons of possible options).


wkornewald:
Is anything blocking this PR?

ktor-client/ktor-client-core/js/src/io/ktor/client/engine/js/Js.kt
```diff
@@ -20,13 +20,18 @@ import io.ktor.client.engine.*
...
-    override fun create(block: HttpClientEngineConfig.() -> Unit): HttpClientEngine =
-        JsClientEngine(HttpClientEngineConfig().apply(block))
+public object Js : HttpClientEngineFactory<JsClientEngineConfig> {
+    override fun create(block: JsClientEngineConfig.() -> Unit): HttpClientEngine =
+        JsClientEngine(JsClientEngineConfig().apply(block))
+}
+
+
+public open class JsClientEngineConfig : HttpClientEngineConfig() {
+    public var nodeOptions: dynamic = js("Object").create(null)
```
rsinukov, line 30: Please add an API doc for this property.

rsinukov (changes requested):
Hi and sorry for such long delay. I think this change is good to go. But can you add a test, please?

ktor-client/ktor-client-core/js/src/io/ktor/client/engine/js/Js.kt
wkornewald: Done

rsinukov:
Do you mind checking the JS compilation error?

ktor-client/ktor-client-tests/common/src/io/ktor/client/tests/utils/CommonClientTestUtils.kt
```diff
@@ -109,8 +109,8 @@ private suspend fun concurrency(level: Int, block: suspend (Int) -> Unit) {
     }
 }
 
-class TestClientBuilder<T : HttpClientEngineConfig>(
-    var config: HttpClientConfig<T>.() -> Unit = {},
+class TestClientBuilder<out T : HttpClientEngineConfig>(
+    var config: HttpClientConfig<@UnsafeVariance T>.() -> Unit = {},
```
wkornewald, line 113: The alternative would be to change `ClientLoader.clientTests` to use `TestClientBuilder<out HttpClientEngineConfig>` but that looks more like a breaking API change. This solution is more general.

Should I even switch `HttpClientConfig` to use `out T` instead? There are a few places in the code that use explicit `HttpClientConfig<out T>` which would become redundant, but also a few more places that would require `@UnsafeVariance`. Not much, but maybe there was a reason why HttpClientConfig wasn't `out T` in the first place? I don't know the project too well. :)

rsinukov (approved):
LGTM

PR #3592, opened 2023-05-03
KTOR-5421 The Logging plugin doesn't log full deserialization errors

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-5421

### DIFF
--- a/ktor-shared/ktor-serialization/ktor-serialization-kotlinx/ktor-serialization-kotlinx-xml/jvm/test/XmlClientKotlinxSerializationTest.kt
+++ b/ktor-shared/ktor-serialization/ktor-serialization-kotlinx/ktor-serialization-kotlinx-xml/jvm/test/XmlClientKotlinxSerializationTest.kt
@@ -39,4 +39,9 @@ class XmlClientKotlinxSerializationTest : AbstractClientContentNegotiationTest()
     @Ignore
     override fun testSerializeNull() {
     }
+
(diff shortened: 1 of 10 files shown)


ktor-shared/ktor-serialization/ktor-serialization-kotlinx/ktor-serialization-kotlinx-xml/jvm/test/XmlClientKotlinxSerializationTest.kt
```diff
@@ -39,4 +39,9 @@ class XmlClientKotlinxSerializationTest : AbstractClientContentNegotiationTest()
     @Ignore
     override fun testSerializeNull() {
     }
+
+    @Test
+    @Ignore
+    override fun testSerializeFailureHasOriginalCauseMessage() {
```
marychatte, line 45: Is it okay that these tests are ignored? Some of them also have error messages, but they have another format

rsinukov: Gson doesn't throw exceptions because it has no handling for non nullable properties
Protobuf throws not helpful exception, like `Expected wire type 2, but found 0`
XML throws an exception before, on type mismatch

PR #3593, opened 2023-05-03
KTOR-5872 RateLimiters for every request key live in memory forever

rsinukov:


ktor-server/ktor-server-plugins/ktor-server-rate-limit/jvmAndNix/src/io/ktor/server/plugins/ratelimit/RateLimitInterceptors.kt
```diff
@@ -46,17 +49,30 @@ private fun PluginBuilder<RateLimitInterceptorsConfig>.rateLimiterPluginBuilder(
...
+                is RateLimiter.State.Exhausted -> {
+                    LOGGER.trace("Declining ${call.request.uri} because of too many requests")
+                    call.respond(HttpStatusCode.TooManyRequests)
+                }
+
+                is RateLimiter.State.Available -> {
+                    clearOnRefillJobs[providerKey]?.cancel()
+                    clearOnRefillJobs[providerKey] = application.launch {
+                        delay(state.refillAtTimeMillis - getTimeMillis())
+                        // there is a race here, where we can remove limiter that was just consumed,
```
marychatte, line 69: Is it a rare situation when such race can occur?

rsinukov: One coroutine should retrieve the rate limiter, while another resumed after the `delay` above. The effect of it will be that request will not be counted in a limit, allowing more requests to be served than specified. 

I think, comparing these two options, it's better to have the race than locking server, even if there is a lock per key.

marychatte (approved):

PR #3594, opened 2023-05-03
KTOR-5876 Allow access to RateLimiters related to call

rsinukov:


ktor-server/ktor-server-plugins/ktor-server-rate-limit/jvmAndNix/src/io/ktor/server/plugins/ratelimit/RateLimitInterceptors.kt
```diff
@@ -53,6 +53,10 @@ private fun PluginBuilder<RateLimitInterceptorsConfig>.rateLimiterPluginBuilder(
             val rateLimiterForCall = registry.computeIfAbsent(providerKey) {
                 provider.rateLimiter(call, key)
             }
+            call.attributes.put(
+                RateLimitersForCallKey,
+                call.attributes.getOrNull(RateLimitersForCallKey).orEmpty() + Pair(provider.name, rateLimiterForCall)
```
marychatte, line 58: Wouldn't it be better to create a `list` before `forEach`, add items here and put to `call.attributes` after `forEach`, because `+` creates a new list each time?

rsinukov: In practice, it should not be much of a difference because there is a small number of rate limiters per route. Usually, it will be not more than two.

marychatte (approved):
lgtm

PR #3595, opened 2023-05-03
[QUIC] Basic Streams

Mr3zee:
**Subsystem**
ktor-network

**Solution**
Added basic implementations of stream management for QUIC.

### DIFF
--- a/ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/connections/QUICConnection_v1.kt
+++ b/ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/connections/QUICConnection_v1.kt
@@ -159,7 +159,7 @@ internal class QUICConnection_v1(
     }
 
     private fun handleError(raisedError: QUICTransportError, frameType: FrameType_v1? = null) {
-        error("Error occurred: ${raisedError.toDebugString()}${frameType?.let { "in frame $it" } ?: ""}")
+        error("Error occurred: ${raisedError.toDebugString()}${frameType?.let { " in frame $it" } ?: ""}")
     }
 
     private fun onTransportParametersKnown() {
@@ -242,7 +242,8 @@ internal class QUICConnection_v1(
      */
     private inner class ProtocolCommunicationProviderImpl : ProtocolCommunicationProvider {
         private val logger = logger()
-        private var offset = 0L
+        private var handshakeOffset = 0L
+        private val buffer = BytePacketBuilder()
 
         override val messageChannel: Channel<TLSMessage> = Channel(Channel.UNLIMITED)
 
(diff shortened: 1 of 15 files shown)


ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/streams/QUICStreamChannel.kt
```diff
@@ -7,23 +7,14 @@ package io.ktor.network.quic.streams
...
-    public val outgoing: SendChannel<QUICStream>
-
-    /**
-     * Send QUIC stream.
+     * Create new QUIC stream.
      */
-    public suspend fun send(stream: QUICStream) {
-        outgoing.send(stream)
-    }
+    public suspend fun createStream(): QUICStream
```
whyoleg, line 17: there should be possibility to open both unidirectional and bidirectional streams

ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/streams/QUICStream.kt
```diff
@@ -0,0 +1,109 @@
...
+package io.ktor.network.quic.streams
+
+import io.ktor.network.quic.bytes.*
+import io.ktor.utils.io.bits.*
+import io.ktor.utils.io.core.*
+
+public interface QUICStream {
+    public val streamId: Long
+
+    public val output: Output
```
whyoleg, line 14: may be it makes more sense to use Byte[Read/Write]Channel for I/O ?

ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/connections/QUICConnection_v1.kt
```diff
@@ -726,6 +762,89 @@ internal class QUICConnection_v1(
         private val logger = logger()
     }
 
+    private inner class StreamManager {
+        private val receiveStates = hashMapOf<Long, ReceiveStreamState>()
```
whyoleg, line 766: note: Long will be boxed (not really an issue, just pointing)

also other note about synchronisation of this states, I mean, thread-safety
I don't see directly in code, if those can be used from multiple threads, so just pointing for additional attention.

ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/streams/QUICStreamChannel.kt
Mr3zee: Yes, for now it is only bidirectional to make PoC implementation (in current state in works for basic cases from the start to the end). I plan to follow the spec more precisely in next iterations, including support for both uni- and bi- directional streams

ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/connections/QUICConnection_v1.kt
Mr3zee: Yep, for now the goal was to make it work, it is definitely not the best approach. 

Multiple threads should not work with the same connection (at least for now)

ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/streams/QUICStream.kt
Mr3zee: Hm, need to check

ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/connections/QUICConnection_v1.kt
```diff
@@ -726,6 +762,89 @@ internal class QUICConnection_v1(
...
+
+        private val streams = hashMapOf<Long, QUICStreamImpl>()
+
+        private val sendChannel: Channel<StreamFrame> = Channel(Channel.UNLIMITED)
+
+        init {
+            // todo close
+            CoroutineScope(Dispatchers.Default).launch {
+                while (isActive) {
```
e5l, line 775: Could you tell me why it's needed?

ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/connections/QUICConnection_v1.kt
```diff
@@ -726,6 +762,89 @@ internal class QUICConnection_v1(
...
+                )
+
+                receiveStates[streamId] = state
+            }
+
+            receiveStates[streamId]!!.receive(streamData, offset, fin)
+        }
+
+        private suspend fun onDataReceived(streamId: Long, dataChunk: ByteReadPacket) {
+            if (!streams.containsKey(streamId)) {
```
e5l, line 820: invert and early return

ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/sockets/QUICSocketBase.kt
```diff
@@ -27,7 +27,7 @@ internal abstract class QUICSocketBase(
 
     init {
         CoroutineScope(Dispatchers.Default).launch {
-            while (true) {
+            while (isActive) {
                 try {
                     receiveAndProcessDatagram()
                 } catch (e: Exception) {
```
e5l, line 33: cause

e5l (changes requested):
Please explain why `isActive` is needed. It also would be great to see some tests at this stage

ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/streams/QUICStream.kt
Mr3zee: I didn't find a quick way to use these interfaces here, and I need to finish my thesis by Monday, so I'll keep as it is for now, but I've added a TODO to revisit this place when I continue the work. Thank you the point though!

ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/connections/QUICConnection_v1.kt
Mr3zee: From the `isActive` doc:

Check this property in long-running computation loops to support cancellation:
```kotlin
while (isActive) {
     // do some computation
}
```

Mr3zee: I know that I don't close it though :) But is does no harm here, and may prevent future me from forgetting to add it when I support cancelation.

ktor-network/ktor-network-quic/jvmAndNix/src/io/ktor/network/quic/connections/QUICConnection_v1.kt
Mr3zee: There is a statement after `if` that should be executed, so I don't think it early return would improve the code here

PR #3602, opened 2023-05-09
KTOR-5899 Darwin engine does not support streaming of request body

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-5899


marychatte (approved):
lgtm

PR #3603, opened 2023-05-09
KTOR-5867 Ktor Client Unable to Stream Responses in Javascript

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-5867


marychatte (approved):
lgtm

PR #3605, opened 2023-05-11
KTOR-5409 Response doesn't contain error message

marychatte:
[KTOR-5409](https://youtrack.jetbrains.com/issue/KTOR-5409)


rsinukov:
I think this change is too invasive. Maybe change it to logging an error, then returning it?

ktor-server/ktor-server-host-common/jvmAndNix/src/io/ktor/server/engine/DefaultEnginePipeline.kt
```diff
@@ -54,7 +54,6 @@ public suspend fun handleFailure(call: ApplicationCall, error: Throwable) {
     tryRespondError(call, defaultExceptionStatusCode(error) ?: HttpStatusCode.InternalServerError)
 }
 
-@OptIn(InternalAPI::class)
 public suspend fun logError(call: ApplicationCall, error: Throwable) {
```
e5l, line 57: please add KDoc

e5l (approved):
lgtm

ktor-server/ktor-server-host-common/jvmAndNix/src/io/ktor/server/engine/DefaultEnginePipeline.kt
marychatte: Can you check KDoc, please?

PR #3606, opened 2023-05-11
KTOR-5879 Bearer auth token refresh hangs after prior refresh threw an exception

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-5879

### DIFF
--- a/ktor-client/ktor-client-plugins/ktor-client-auth/common/test/io/ktor/client/plugins/auth/AuthTest.kt
+++ b/ktor-client/ktor-client-plugins/ktor-client-auth/common/test/io/ktor/client/plugins/auth/AuthTest.kt
@@ -16,6 +16,7 @@ import io.ktor.http.*
 import io.ktor.http.auth.*
 import io.ktor.test.dispatcher.*
 import io.ktor.util.*
+import io.ktor.utils.io.errors.*
 import kotlinx.coroutines.*
 import kotlin.test.*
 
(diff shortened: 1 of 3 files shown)


ktor-client/ktor-client-plugins/ktor-client-auth/common/test/io/ktor/client/plugins/auth/AuthTest.kt
```diff
@@ -600,6 +601,38 @@ class AuthTest : ClientLoader() {
...
+                        }
+                        val token = client.get("$TEST_SERVER/auth/bearer/token/second?delay=500").bodyAsText()
+                        BearerTokens(token, token)
+                    }
+                }
+            }
+        }
+        test { client ->
+            firstCall = true
+            client.get("$TEST_SERVER/auth/bearer/first").bodyAsText()
```
marychatte, line 625: Maybe we want to check it as `assertEquals("OK", client.get("$TEST_SERVER/auth/bearer/first").bodyAsText())`

ktor-client/ktor-client-plugins/ktor-client-auth/common/test/io/ktor/client/plugins/auth/AuthTest.kt
```diff
@@ -600,6 +601,38 @@ class AuthTest : ClientLoader() {
...
+
+            assertFails {
```
marychatte, line 627: And also, maybe we want to check the message of an error to be sure that it is the expected error

marychatte (approved):
lgtm, only small comments about the test

PR #3611, opened 2023-05-15
Improve consistency of MergedApplicationConfig.kt

BierDav:
**Subsystem**
Server module
- Config

**Motivation**
A bug that was fixed my Rustam introduced a new property, that doesn't fit the pattern and makes the code a bit less readable and unefficient.

**Solution**
By also caching the keys of the second ApplicationConfig this consistency is restored


rsinukov (approved):
LGTM

PR #3619, opened 2023-05-24
KTOR-505 Add Server-sent events plugin for client and support for OkHttp engine

marychatte:
[KTOR-505](https://youtrack.jetbrains.com/issue/KTOR-505)

[KTOR-5938 Design](https://youtrack.jetbrains.com/issue/KTOR-5938)
[Specification](https://html.spec.whatwg.org/multipage/server-sent-events.html#server-sent-events)


ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/sse/ServerSentEvents.kt
```diff
@@ -0,0 +1,97 @@
...
+     * [ServerSentEvents] configuration.
+     */
+    @KtorDsl
+    public class Config {
+        /**
+         * The reconnection time. If the connection to the server is lost,
+         * the client will wait for the specified time before attempting to reconnect.
+         * Note that this parameter is not supported for some engines.
+         */
+        public var reconnectionTimeMillis: Long = 1000
```
whyoleg, line 42: may be also add support for kotlin.time.Duration?

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/sse/ServerSentEvents.kt
```diff
@@ -0,0 +1,97 @@
...
+public object ServerSentEventsCapability : HttpClientEngineCapability<Unit> {
+    override fun toString(): String = "ServerSentEventsCapability"
+}
+
+/**
+ * Client Server-sent events plugin.
+ *
+ * @property reconnectionTimeMillis - time for client to wait before attempting to reconnect.
+ */
+public class ServerSentEvents(
```
whyoleg, line 29: Why not to use new Client Plugin API?

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/sse/ServerSentEvents.kt
```diff
@@ -0,0 +1,97 @@
...
+}
+
+/**
+ *  Server-sent event receiving from server.
+ *
+ *  @property event - a string identifying the type of event described
+ *  @property id - the event ID.
+ *  @property data - the data field for the message.
+ */
+public class ServerSentEvent(public val event: String? = null, public val id: String? = null, public val data: String) {
```
whyoleg, line 87: Does it make sense to introduce this declaration in some `ktor-shared-sse` module to be able to share it with server implementation and adapt it based on it? Or somehow else align ktor server and client SSE support

ktor-client/ktor-client-okhttp/jvm/src/io/ktor/client/engine/okhttp/OkHttpServerSentEventsSession.kt
```diff
@@ -0,0 +1,54 @@
...
+        get() = _incoming
+
+    override fun onOpen(eventSource: EventSource, response: Response) {
+        super.onOpen(eventSource, response)
+        originResponse.complete(response)
+    }
+
+    override fun onEvent(eventSource: EventSource, id: String?, type: String?, data: String) {
+        super.onEvent(eventSource, id, type, data)
+        _incoming.trySendBlocking(ServerSentEvent(type, id, data))
```
whyoleg, line 35: I think, that it's better to use `Channel(Channel.UNLIMITED)` and use just `trySend` here instead of blocking variant

ktor-client/ktor-client-okhttp/jvm/src/io/ktor/client/engine/okhttp/OkHttpServerSentEventsSession.kt
```diff
@@ -0,0 +1,54 @@
...
+
+    internal val originResponse: CompletableDeferred<Response> = CompletableDeferred()
+
+    private val _incoming = Channel<ServerSentEvent>()
+
+    override val incoming: ReceiveChannel<ServerSentEvent>
+        get() = _incoming
+
+    override fun onOpen(eventSource: EventSource, response: Response) {
+        super.onOpen(eventSource, response)
```
whyoleg, line 29: `super` calls are redundant

whyoleg (commented):
Some minor (or not) comments :)

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/sse/ServerSentEvents.kt
marychatte:
```kotlin
public var reconnectionTime: Duration
        get() = reconnectionTimeMillis.milliseconds
        set(value) {
            reconnectionTimeMillis = value.inWholeMilliseconds
        }
```

Do you mean to add also something like this?

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/sse/ServerSentEvents.kt
marychatte: Yes, you are right. Will move to `ktor-shared-sse` module

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/sse/ServerSentEvents.kt
marychatte: Thanks for comments, I will rewrite it with new Client Plugin API

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/sse/ServerSentEvents.kt
whyoleg: yes, or even leave only type safe way via `Duration`. It depends on what is the overall strategy in ktor regarding `millis` properties. There are a lot of them in different plugins. Somewhere there are extensions for `Duration`, somewhere not.
I believe that 3.0 is upcoming release and `kotlin.time.Duration` is already Stable, may be it's a good idea to check all plugins (server and client) and may be other places, there `millis` properties are used and add `Duration` counterpart.

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/sse/ClientServerSentEventsSession.kt
```diff
@@ -0,0 +1,19 @@
...
+package io.ktor.client.plugins.sse
+
+import io.ktor.sse.*
+import kotlinx.coroutines.*
+import kotlinx.coroutines.channels.*
+
+/**
+ * A client Server-sent events session.
+ */
+public interface ClientServerSentEventsSession : CoroutineScope {
```
e5l, line 14: Could we name it `ClientSSESession`? `SSE` is a well-known keyword; it can be searched. Is it possible to have an SSE session for the server in the future?

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/sse/ServerSentEventsConfig.kt
```diff
@@ -0,0 +1,20 @@
...
+
+package io.ktor.client.plugins.sse
+
+import kotlin.time.*
+import kotlin.time.Duration.Companion.milliseconds
+
+/**
+ * A config for the [ServerSentEvents] plugin.
+ */
+public class ServerSentEventsConfig {
```
e5l, line 13: same question for naming

e5l (changes requested):
LGTM, except for inconsistent naming

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/sse/ClientServerSentEventsSession.kt
marychatte: Thanks, I will rename it and related things. Yes, it is possible to have an SSE session for the server in the future

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/sse/builders.kt
```diff
@@ -0,0 +1,177 @@
...
+public fun HttpClientConfig<*>.SSE(config: SSEConfig.() -> Unit) {
+    install(SSE) {
+        config()
+    }
+}
+
+/**
+ * Opens a [ClientSSESession].
+ */
+public suspend fun HttpClient.serverSentEventsSession(
```
e5l, line 26: should we also rename it in api to `sseSession`?

e5l (approved):
lgtm, please check the comment

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/sse/builders.kt
marychatte: There is also a similar function with the name `sseSession `. I thought to leave both variants like in WebSocket plugin, where it is `ws` and `websocket`

PR #3623, opened 2023-05-27
Fix explicitApi issues

Tapchicoma:
After the fix in KGP, explicitApi now works in MPP projects leading to project compilation error.


woainikk:
https://youtrack.jetbrains.com/issue/KT-58682/Explicit-api-mode-does-not-apply-in-MPP-projects

PR #3640, opened 2023-05-31
KTOR-793 Make LDAP return generic Principal instead of UserIdPrincipal

marychatte:
[KTOR-793](https://youtrack.jetbrains.com/issue/KTOR-793)


e5l (approved):
lgtm

PR #3655, opened 2023-06-13
KTOR-6001 Cache returns null when vary header has more fields in the cached response

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-6001

### DIFF
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/cache/storage/UnlimitedCacheStorage.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/cache/storage/UnlimitedCacheStorage.kt
@@ -42,7 +42,9 @@ internal class UnlimitedStorage : CacheStorage {
 
     override suspend fun find(url: Url, varyKeys: Map<String, String>): CachedResponseData? {
         val data = store.computeIfAbsent(url) { ConcurrentSet() }
-        return data.find { it.varyKeys == varyKeys }
+        return data.find {
+            varyKeys.all { (key, value) -> it.varyKeys[key] == value }
+        }
     }
 
     override suspend fun findAll(url: Url): Set<CachedResponseData> = store[url] ?: emptySet()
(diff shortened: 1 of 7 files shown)


ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/cache/storage/UnlimitedCacheStorage.kt
```diff
@@ -42,7 +42,9 @@ internal class UnlimitedStorage : CacheStorage {
 
     override suspend fun find(url: Url, varyKeys: Map<String, String>): CachedResponseData? {
         val data = store.computeIfAbsent(url) { ConcurrentSet() }
-        return data.find { it.varyKeys == varyKeys }
+        return data.find {
```
marychatte, line 45: There is one more `find` function on line 23, should we also change it like here?

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/cache/storage/UnlimitedCacheStorage.kt
```diff
@@ -42,7 +42,9 @@ internal class UnlimitedStorage : CacheStorage {
...
+            varyKeys.all { (key, value) -> it.varyKeys[key] == value }
```
marychatte, line 46: Shouldn't we also check the equal size of maps?

marychatte (commented):
Sorry, accidentally missed this pr :) I have some comments

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/cache/storage/UnlimitedCacheStorage.kt
rsinukov: No, it's ok if cached version has more vary properties, than the 304 response

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/cache/storage/UnlimitedCacheStorage.kt
rsinukov: This is the legacy storage, but I changed it as well. Thanks!

marychatte (approved):
lgtm

PR #3657, opened 2023-06-19
Update ktlint version

e5l:


ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/websocket/builders.kt
```diff
@@ -42,8 +42,9 @@ public suspend fun HttpClient.webSocketSession(
                 val sessionCompleted = CompletableDeferred<Unit>()
                 sessionDeferred.complete(session)
                 session.outgoing.invokeOnClose {
-                    if (it != null) sessionCompleted.completeExceptionally(it)
-                    else sessionCompleted.complete(Unit)
+                    if (it != null) {
```
rsinukov, line 45: i really don't like this formatting. maybe replace such changes with `when`?

rsinukov (approved):

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/websocket/builders.kt
e5l: yep, we can

Stexxe: Or just wrap the else expression with curly brackets

Stexxe (approved):

PR #3659, opened 2023-06-21
KTOR-6036 Add deprecations for old IO

e5l:


ktor-io/common/src/io/ktor/utils/io/Deprecation.kt
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright 2014-2023 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.utils.io
+
+internal const val IO_DEPRECATION_MESSAGE = """
+    We're migrating to the new kotlinx-io library.
+    This declaration is deprecated and will be removed in Ktor 4.0.0
```
rsinukov, line 9: Should we explain why there is no replacement yet and what people should do if it's in our public API?

rsinukov (changes requested):
Can we avoid suppressing all deprecation in the file?

ktor-io/common/src/io/ktor/utils/io/Deprecation.kt
e5l: This API will be dropped without replacement

PR #3660, opened 2023-06-22
KTOR-5688 Make NSURLSessionDelegate constructor public

e5l:


marychatte (approved):
lgtm

rsinukov (changes requested):
I think it's ok to make constructor public and delete factory functions. But I would prefer to keep everything else internal or private

ktor-client/ktor-client-darwin-legacy/darwin/src/io/ktor/client/engine/darwin/KtorLegacyNSURLSessionDelegate.kt
```diff
@@ -35,11 +34,12 @@ public fun KtorLegacyNSURLSessionDelegate(): KtorLegacyNSURLSessionDelegate {
  *   * URLSession:task:willPerformHTTPRedirection:newRequest:completionHandler:
  */
 @OptIn(UnsafeNumber::class)
-public class KtorLegacyNSURLSessionDelegate internal constructor(
-    private val challengeHandler: ChallengeHandler?
+public class KtorLegacyNSURLSessionDelegate(
```
rsinukov, line 37: factory function can be removed, if you add default parameter

ktor-client/ktor-client-darwin-legacy/darwin/src/io/ktor/client/engine/darwin/internal/legacy/DarwinLegacyTaskHandler.kt
```diff
@@ -17,11 +16,11 @@ import platform.Foundation.*
 import kotlin.coroutines.*
 
 @OptIn(DelicateCoroutinesApi::class)
-internal class DarwinLegacyTaskHandler(
+public class DarwinLegacyTaskHandler(
```
rsinukov, line 19: why do other classes need to be public?

rsinukov (approved):
LGTM, but minot comments

ktor-client/ktor-client-darwin-legacy/darwin/src/io/ktor/client/engine/darwin/internal/legacy/DarwinLegacyTaskHandler.kt
e5l: thanks, fixed

PR #3664, opened 2023-06-27
KTOR-6013 Fix deeply nested keys in `MapApplicationConfig::toMap`

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-6013


e5l (approved):
lgtm

PR #3669, opened 2023-07-05
Fix resource leak in CIO ConnectionFactory

jooohn:
**Subsystem**
Client, CIO

**Motivation**
There's a resource leak bug in CIO ConnectionFactory. This bug causes a reduction in the maximum number of connections we can have over time.

**Solution**
Release `limit` semaphore when it fails to acquire per-address semaphore.

### DIFF
--- a/ktor-client/ktor-client-cio/jvmAndNix/src/io/ktor/client/engine/cio/ConnectionFactory.kt
+++ b/ktor-client/ktor-client-cio/jvmAndNix/src/io/ktor/client/engine/cio/ConnectionFactory.kt
@@ -22,14 +22,18 @@ internal class ConnectionFactory(
         configuration: SocketOptions.TCPClientSocketOptions.() -> Unit = {}
     ): Socket {
         limit.acquire()
-        val addressSemaphore = addressLimit.computeIfAbsent(address) { Semaphore(addressConnectionsLimit) }
-        addressSemaphore.acquire()
-
         return try {
-            aSocket(selector).tcpNoDelay().tcp().connect(address, configuration)
+            val addressSemaphore = addressLimit.computeIfAbsent(address) { Semaphore(addressConnectionsLimit) }
+            addressSemaphore.acquire()
+
+            try {
+                aSocket(selector).tcpNoDelay().tcp().connect(address, configuration)
+            } catch (cause: Throwable) {
+                // a failure or cancellation
+                addressSemaphore.release()
+                throw cause
+            }
         } catch (cause: Throwable) {
-            // a failure or cancellation
-            addressSemaphore.release()
             limit.release()
             throw cause
         }
--- a/ktor-client/ktor-client-cio/jvmAndNix/test/ConnectionFactoryTest.kt
+++ b/ktor-client/ktor-client-cio/jvmAndNix/test/ConnectionFactoryTest.kt
@@ -1,0 +1,99 @@
+/*
+ * Copyright 2014-2023 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+import io.ktor.client.engine.cio.*
+import io.ktor.network.selector.*
+import io.ktor.network.sockets.*
+import io.ktor.utils.io.core.*
+import kotlinx.coroutines.*
+import kotlin.test.*
+
+class ConnectionFactoryTest {
+
+    private val selectorManager = SelectorManager()
+
+    @Test
+    fun testLimitSemaphore() = runBlocking {
+        val connectionFactory = ConnectionFactory(
+            selectorManager,
+            connectionsLimit = 2,
+            addressConnectionsLimit = 1,
+        )
+        withServerSocket(0) { socket0 ->
+            withServerSocket(1) { socket1 ->
+                withServerSocket(2) { socket2 ->
+                    connectionFactory.connect(socket0.localAddress as InetSocketAddress)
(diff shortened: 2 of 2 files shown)


ktor-client/ktor-client-cio/jvmAndNix/src/io/ktor/client/engine/cio/ConnectionFactory.kt
```diff
@@ -22,14 +22,18 @@ internal class ConnectionFactory(
...
         limit.acquire()
-        val addressSemaphore = addressLimit.computeIfAbsent(address) { Semaphore(addressConnectionsLimit) }
-        addressSemaphore.acquire()
-
         return try {
-            aSocket(selector).tcpNoDelay().tcp().connect(address, configuration)
+            val addressSemaphore = addressLimit.computeIfAbsent(address) { Semaphore(addressConnectionsLimit) }
+            addressSemaphore.acquire()
+
+            try {
```
rsinukov, line 29: Why nested `try`/`catch` block is necessary here?

rsinukov (commented):
Hi and thanks for PR. Can you explain steps in which this bug can be reproduced?

ktor-client/ktor-client-cio/jvmAndNix/src/io/ktor/client/engine/cio/ConnectionFactory.kt
jooohn: Hi @rsinukov, thanks for the review.

> Why nested try/catch block is necessary here?

It's to make sure that the `limit` semaphore is released when failure.

In the original code, there's no try-catch between the `limit` acquisition and `addressSemaphore` acquisition, so if it succeeds to acquire `limit` semaphore but it fails to acquire the `addressSemaphore` (e.g. by timeout), then it didn't release the `limit` semaphore - this is the resource leak.

> Can you explain steps in which this bug can be reproduced?

With the original code, this new test I added will be blocked forever.

https://github.com/ktorio/ktor/blob/7c08c9cb359d8196f6adb218cf6d1a8282e00fa5/ktor-client/ktor-client-cio/jvmAndNix/test/ConnectionFactoryTest.kt#L60

Because [this failure](https://github.com/ktorio/ktor/blob/7c08c9cb359d8196f6adb218cf6d1a8282e00fa5/ktor-client/ktor-client-cio/jvmAndNix/test/ConnectionFactoryTest.kt#L72) doesn't release the `limit` semaphore, this attempt cannot acquire the `limit` semaphore.

rsinukov: Thanks for the explanation. But isn't it enough just to move `addressSemaphore.acquire()` into the outer `try/catch` block?

jooohn: Do you mean, something like this?

```kotlin
limit.acquire()
val addressSemaphore = try {
  addressLimit.computeIfAbsent(address) { Semaphore(addressConnectionsLimit) }
} catch (cause: Throwable) {
  limit.release()
  throw cause
}

try {
  aSocket(selector).tcpNoDelay().tcp().connect(address, configuration)
} catch (cause: Throwable) {
  // a failure or cancellation
  addressSemaphore.release()

  limit.release() // <= we need limit.release() here as well

  throw cause
}
```

We can, but I believe nested try-catch is more straightforward. Once `limit` is acquired, we need to release it whenever error happens in this method.

rsinukov: No, i mean just moving `addressSemaphore.acquire()` inside the `try/catch` and leave everything else as is. Since it's the only suspendible call, only it can throw an exception. And there is no point in having one extra nested block
```
        limit.acquire()
        val addressSemaphore = addressLimit.computeIfAbsent(address) { Semaphore(addressConnectionsLimit) }
        return try {
            addressSemaphore.acquire()
            aSocket(selector).tcpNoDelay().tcp().connect(address, configuration)
        } catch (cause: Throwable) {
            limit.release()
            throw cause
        }
```

jooohn: Hmm, it looks it will release `addressSemaphore` when it fails to `addressSemaphore.acquire` (by cancellation while waiting for the semaphore acquisition for example) - which means `addressSemaphore` will accidentally release more semaphore than it should 🤔

jooohn:
@rsinukov Thanks for approving! Added new commit since [it failed some tests](https://ktor.teamcity.com/buildConfiguration/Ktor_KtorMatrixCoreLinuxJava8/224522?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandPull+Request+Details=true&expandBuildTestsSection=true&expandBuildChangesSection=true), but it seems pushing new changes didn't trigger the tests automatically. Could you trigger the tests again?

rsinukov:
@jooohn thanks for fixing it. I see that native tests are still failing. Can you take a look at it? 
You may need to set `ktor.ide.jvmAndCommonOnly=false` inside `gradle.properties` to make IJ index native targets.

jooohn:
@rsinukov thanks for the advice! I hope the last commit fixes the failing tests 👀

PR #3679, opened 2023-07-07
KTOR-6080 Decompressed response/request should unset Content-Encoding header

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-6080


e5l (approved):
Could you check if Jakarta test is related?

PR #3681, opened 2023-07-08
KTOR-6073 Webjars plugin should include caching headers and ETag by default

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-6073

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-webjars/jvm/src/io/ktor/server/webjars/Webjars.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-webjars/jvm/src/io/ktor/server/webjars/Webjars.kt
@@ -1,18 +1,27 @@
 package io.ktor.server.webjars
 
 import io.ktor.http.*
+import io.ktor.http.content.*
 import io.ktor.server.application.*
 import io.ktor.server.request.*
 import io.ktor.server.response.*
 import io.ktor.util.*
 import io.ktor.util.date.*
 import org.webjars.*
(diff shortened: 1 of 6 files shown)


ktor-server/ktor-server-plugins/ktor-server-webjars/jvm/src/io/ktor/server/webjars/Webjars.kt
```diff
@@ -28,6 +37,39 @@ public class WebjarsConfig {
...
+    public fun etag(block: (WebJarInfo) -> String?) {
+        etagExtractor = block
+    }
+
+    /**
+     * Specifies a value for [HttpHeaders.CacheControl] to be used in the response.
+     * By default, it is 90 days.
+     * Return `null` from this block to omit the header.
+     *
+     * Note: for this property to work, you need to install the [CachingHeaders] plugin.
```
marychatte, line 68: Can we install `CachingHeaders` and `ConditionalHeaders` with the `Webjars` plugin by default? Maybe we can add an option in the `Webjars` config to omit them then? I'm not sure it is convenient to install another plugin to be able to work with property from our plugin

rsinukov: They are in different modules, so no, it's not easy. Maybe we can check dependency in runtime, but it won't help much, since the users will need to add the dependency anywat.

PR #3682, opened 2023-07-08
KTOR-6007 renderSetCookieHeader shouldn't ignore maxAge = 0

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-6007


e5l (changes requested):
please check the CI

e5l (approved):

Issue #3557, opened 2023-04-20
ReaderJob & WriterJob leak in TLS socket and prevent context from completion


olehs:
https://github.com/ktorio/ktor/blob/0081f943b434bdd0afd82424b389629e89a89461/ktor-network/ktor-network-tls/jvm/src/io/ktor/network/tls/TLSClientSessionJvm.kt#L40-L48

Jobs created by these two methods should be cancelled when socket is closed

rmmcosta:
Does this issue still makes sense after the #3690 fix?

olehs:
I didn't track latest changes as we used a workaround (see below), so can't say for sure. I'll check it and if any issues appear, report them in a new topic.

```kotlin
    private class TLSSocket(private val tlsSocket: Socket) : Socket by tlsSocket {
        private var readerJob: ReaderJob? = null
        private var writerJob: WriterJob? = null

        override fun attachForReading(channel: ByteChannel): WriterJob {
            return tlsSocket.attachForReading(channel).also {
                writerJob = it
            }
        }

        override fun attachForWriting(channel: ByteChannel): ReaderJob {
            return tlsSocket.attachForWriting(channel).also {
                readerJob = it
            }
        }

        override fun close() {
            readerJob?.cancel()
            writerJob?.cancel()
        }
    }
```

Thank you

PR #3690, opened 2023-07-11
Avoid crashing when TLS TCP socket is closed

t-bast:
**Subsystem**
Client, TLS over TCP

**Motivation**
When using a plain TCP socket with TLS, two worker jobs are spawned:

- cio-tls-input-loop
- cio-tls-output-loop

These worker jobs live in their own scope, and exceptions they throw can't easily be caught by the caller: the only option is to install a `CoroutineExceptionHandler` for the whole TLS socket, which is quite hacky and may hide bugs.

When the TCP socket is closed, this is only caught when trying to write on the output channel, which throws a `ClosedSendChannelException`.

See the following tickets for that issue:

- https://youtrack.jetbrains.com/issue/KTOR-5178/TLSSocket-cannot-catch-the-exception-thrown-by-appDataOutputLoop
- https://youtrack.jetbrains.com/issue/KTOR-4360/Android-Impossible-to-catch-the-ClosedSendChannelException-when-TLS-connection-socket-is-closed
- https://github.com/ktorio/ktor/issues/3557

**Solution**
We now catch that exception which cleanly stops the background job.

The unit test added fails when we remove the `catch`. It is unfortunately a slightly roundabout way of testing our behavior, because it's the last check (`channel.isClosedForWrite`) that will fail if you comment the `catch`. You'll also see the `ClosedSendChannelException` in the test output if you comment the `catch` block, but it's not sufficient to make the test fail, because the test runner allows exceptions in background threads (but it will crash an android app for example). I'm not sure if there are better ways to test this?

### DIFF
--- a/ktor-network/ktor-network-tls/jvm/src/io/ktor/network/tls/TLSClientSessionJvm.kt
+++ b/ktor-network/ktor-network-tls/jvm/src/io/ktor/network/tls/TLSClientSessionJvm.kt
@@ -47,7 +47,6 @@ private class TLSSocket(
             appDataOutputLoop(this.channel)
         }
 
-    @OptIn(ExperimentalCoroutinesApi::class)
     private suspend fun appDataInputLoop(pipe: ByteWriteChannel) {
         try {
             input.consumeEach { record ->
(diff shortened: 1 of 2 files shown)


e5l (approved):

t-bast:
The test seems to be flaky on some platforms, mostly because it's not actually really testing what we'd like.
Any idea on how I could improve that? Is that test actually necessary?

t-bast:
@e5l any chance this could be included or reviewed?

e5l:
Thanks! Will take a look

t-bast:
@e5l is there anything I can do to help integrate this fix?

e5l:
lgtm, merged

PR #3693, opened 2023-07-13
KTOR-6098 Netty: Allow listening only for HTTP/1.1 protocol SSL connections

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-6098


e5l (approved):
Is requestQueueLimit unused?

e5l:
Should we fix it instead of dropping it?

rsinukov:
> Is requestQueueLimit unused?

Yes, it is unused for a long time already. I think even before 2.0.0

PR #3696, opened 2023-07-15
KTOR-6108 KtorServlet does not support yaml configuration

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-6108


marychatte (approved):
Looks great

PR #3699, opened 2023-07-18
KTOR-6120 staticFiles responds twice if both index and defaultPath are set

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-6120


marychatte (approved):
lgtm

PR #3700, opened 2023-07-19
KTOR-6125 ContentNegotiation with protobuf fails on empty body

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-6125

### DIFF
--- a/ktor-shared/ktor-serialization/ktor-serialization-kotlinx/common/src/io/ktor/serialization/kotlinx/KotlinxSerializationConverter.kt
+++ b/ktor-shared/ktor-serialization/ktor-serialization-kotlinx/common/src/io/ktor/serialization/kotlinx/KotlinxSerializationConverter.kt
@@ -72,7 +72,7 @@ public class KotlinxSerializationConverter(
         val fromExtension = extensions.asFlow()
             .map { it.deserialize(charset, typeInfo, content) }
             .firstOrNull { it != null || content.isClosedForRead }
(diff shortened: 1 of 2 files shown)


ktor-shared/ktor-serialization/ktor-serialization-kotlinx/common/src/io/ktor/serialization/kotlinx/KotlinxSerializationConverter.kt
```diff
@@ -72,7 +72,7 @@ public class KotlinxSerializationConverter(
         val fromExtension = extensions.asFlow()
             .map { it.deserialize(charset, typeInfo, content) }
             .firstOrNull { it != null || content.isClosedForRead }
-        if (fromExtension != null || content.isClosedForRead) return fromExtension
+        if (extensions.isNotEmpty() && (fromExtension != null || content.isClosedForRead)) return fromExtension
```
marychatte, line 75: Should we also add this check in `serializeNullable(..) { ... if (fromExtension != null) return fromExtension ... }`?

rsinukov: no, because if `fromExtension != null` then `extensions` are always not empty

marychatte (approved):
lgtm

PR #3701, opened 2023-07-20
KTOR-5979 Fix tail messages decompression

e5l:


rsinukov (approved):
LGTM, but I think you used the wrong ticket number in the title

e5l:
Thanks, will fix

PR #3702, opened 2023-07-20
KTOR-6119 Persistant Cookie Storage: Make Cookie.matches and Cookie.fillDefaults methods public

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-6119


marychatte (approved):
lgtm

PR #3711, opened 2023-08-01
KTOR-5881 Remove blocking API from multipart file handling

rsinukov:


ktor-client/ktor-client-tests/common/src/io/ktor/client/tests/utils/Generators.kt
```diff
@@ -7,33 +7,38 @@ package io.ktor.client.tests.utils
...
         val content = when (it) {
             is PartData.FileItem -> filenameContentTypeAndContentString(it.provider, it.headers)
             is PartData.FormItem -> it.value
-            is PartData.BinaryItem -> filenameContentTypeAndContentString(it.provider, it.headers)
+            is PartData.BinaryItem -> filenameContentTypeAndContentString(
+                { ByteReadChannel(it.provider().readBytes()) },
+                it.headers
+            )
+
             is PartData.BinaryChannelItem -> error("Not implemented")
```
whyoleg, line 30: BTW: this could be supported now

ktor-http/common/src/io/ktor/http/content/Multipart.kt
```diff
@@ -27,7 +27,7 @@ public sealed class PartData(public val dispose: () -> Unit, public val headers:
      * @property provider of content bytes
      */
     public class FileItem(
```
whyoleg, line 29: may be it could be just merged with BinaryChannelItem?

whyoleg (commented):
minor questions :)

ktor-http/common/src/io/ktor/http/content/Multipart.kt
whyoleg: But does it make sense to have specific implementation for it? This property could be an extension on `BinaryChannelItem`
May be, even for some reason someone will want to have `BinaryItem` with file content.

I don't know all specifics but looks a little odd, that we have `BinaryItem`, `FileItem` and `BinaryChannelItem` with almost the same semantics but different.

rsinukov: `BinaryItem` uses `Input` which is blocking and already deprecated in the `main` branch. I don't see a problem in having separate entity for files in multipart, even if it's almost the same. Having `originalFileName` property on `BinaryChannelItem` can be confusing.

PR #3714, opened 2023-08-03
KTOR-3266 Inconsistent behavior for different engines when exception is thrown in the writer of WriteChannelContent

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-3266

### DIFF
--- a/ktor-server/ktor-server-cio/jvmAndNix/src/io/ktor/server/cio/backend/ServerPipeline.kt
+++ b/ktor-server/ktor-server-cio/jvmAndNix/src/io/ktor/server/cio/backend/ServerPipeline.kt
@@ -148,7 +148,7 @@ public fun CoroutineScope.startServerConnectionPipeline(
                     actorChannel.close()
                     connection.input.copyAndClose(requestBody as ByteChannel)
                     break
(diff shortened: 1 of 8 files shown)


e5l:
It looks like it also fixes https://youtrack.jetbrains.com/issue/KTOR-4862/Ktor-hangs-if-exception-occurs-during-write-response-body

ktor-server/ktor-server-cio/jvmAndNix/src/io/ktor/server/cio/backend/ServerPipeline.kt
```diff
@@ -148,7 +148,7 @@ public fun CoroutineScope.startServerConnectionPipeline(
                     actorChannel.close()
                     connection.input.copyAndClose(requestBody as ByteChannel)
                     break
-                } else if (!expectedHttpBody && requestBody is ByteChannel) { // not upgraded, for example 404
+                } else if (requestBody is ByteChannel) { // not upgraded, for example 404
```
e5l, line 151: It looks like it will add obsolete flush

e5l (changes requested):
please check the CI

e5l (approved):

PR #3718, opened 2023-08-07
KTOR-6122 Add type to AttributeKey equals

e5l:
Fix [KTOR-6122](https://youtrack.jetbrains.com/issue/KTOR-6122/AttributeKey-equality-comparison-breaks-type-safety)


rsinukov (changes requested):
Please check JS build

rsinukov (approved):

PR #3720, opened 2023-08-08
KTOR-6147 Darwin: App hangs when streaming request without network connection

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-6147

Unfortunately, can't add test for this, because it's reproducible only without network connection

### DIFF
--- a/ktor-client/ktor-client-darwin/darwin/src/io/ktor/client/engine/darwin/DarwinUtils.kt
+++ b/ktor-client/ktor-client-darwin/darwin/src/io/ktor/client/engine/darwin/DarwinUtils.kt
@@ -9,13 +9,10 @@ import io.ktor.client.engine.*
 import io.ktor.http.content.*
 import io.ktor.util.*
 import io.ktor.utils.io.*
-import io.ktor.utils.io.core.*
 import io.ktor.utils.io.errors.*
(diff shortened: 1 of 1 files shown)


ktor-client/ktor-client-darwin/darwin/src/io/ktor/client/engine/darwin/DarwinUtils.kt
```diff
@@ -53,10 +50,19 @@ internal suspend fun OutgoingContent.toDataOrStream(): Any? {
                     var offset = 0
                     val read = channel.readAvailable(buffer, 0, 4096)
                     while (offset < read) {
+                        while (!outputStream.hasSpaceAvailable) {
+                            yield()
```
marychatte, line 54: How does it solve the problem?

rsinukov: `outputStream.write()` will block if there is no space available and since it's not suspendable,  it will not throw on coroutine context cancellation. `yield()` adds a suspendable call, to make sure that we don't block

marychatte (approved):
lgtm

PR #3722, opened 2023-08-10
KTOR-4225 Fix receiving body in the response validator

e5l:
Fix [KTOR-4225](https://youtrack.jetbrains.com/issue/KTOR-4225/HttpResponseValidator-consumes-HTTP-response-body)


ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/DoubleReceivePlugin.kt
```diff
@@ -0,0 +1,113 @@
...
+ * ```
+ */
+@OptIn(InternalAPI::class)
+public val SaveBodyPlugin: ClientPlugin<SaveBodyPluginConfig> = createClientPlugin(
+    "DoubleReceivePlugin",
+    ::SaveBodyPluginConfig
+) {
+
+    client.receivePipeline.intercept(HttpReceivePipeline.Before) { response ->
+        if (this@createClientPlugin.pluginConfig.disabled) return@intercept
```
rsinukov, line 65: it's better to save config properties in local values, because they are mutable

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/internal/ByteChannelReplay.kt
```diff
@@ -0,0 +1,46 @@
...
+        val value = content
+        if (value != null) {
+            return ByteReadChannel(value)
+        }
+
+        if (origin.closedCause != null) {
+            throw origin.closedCause!!
+        }
+
+        return GlobalScope.writer(Dispatchers.Unconfined) {
```
rsinukov, line 25: should there be some sort of mutex to avoid concurrency issues?

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/internal/ByteChannelReplay.kt
```diff
@@ -0,0 +1,46 @@
...
+            val body = BytePacketBuilder()
+            try {
+                while (!origin.isClosedForRead) {
+                    if (origin.availableForRead == 0) origin.awaitContent()
+                    val packet = origin.readPacket(origin.availableForRead)
+
+                    body.writePacket(packet.copy())
```
rsinukov, line 32: is it necessary to  copy?

rsinukov (changes requested):
Can you check JS build?

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/DoubleReceivePlugin.kt
e5l: done

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/internal/ByteChannelReplay.kt
e5l: It will not copy actual bytes, so yep

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/DoubleReceivePlugin.kt
```diff
@@ -0,0 +1,115 @@
...
+    client.receivePipeline.intercept(HttpReceivePipeline.Before) { response ->
+        if (disabled) return@intercept
+
+        val attributes = response.call.attributes
+        if (attributes.contains(SKIP_SAVE_BODY)) return@intercept
+
+        val bodyReplay = ByteChannelReplay(response.content)
+
+        val result = object : HttpResponse() {
+            override val call: HttpClientCall = response.call
```
rsinukov, line 74: Please use `HttpClientCall.wrapWithContent(content)` instead. Otherwise `response != response.call.response` and it may lead to bugs.

rsinukov (changes requested):

rsinukov (approved):
LGTM, but please revert empty catches

PR #3725, opened 2023-08-11
KTOR-6182 Throw UnknownServiceException if cleartext traffic is not permitted

Goooler:
https://youtrack.jetbrains.com/issue/KTOR-6182

This part had been implemented in Okhttp & HttpURLConnection, we just need to patch it for CIO.

Staled some code from:
https://github.com/square/okhttp/blob/735ed1a6e5a15042f6d7e8f100fb3c8376f6aa27/okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/RealRoutePlanner.kt#L205-L208
https://github.com/square/okhttp/blob/550cc11989e5117ed3d8afaf20c7b6c419b5351f/okhttp/src/main/kotlin/okhttp3/internal/platform/AndroidPlatform.kt#L105-L148

### DIFF
--- a/ktor-client/ktor-client-core/jvm/src/io/ktor/client/plugins/CleartextTraffic.jvm.kt
+++ b/ktor-client/ktor-client-core/jvm/src/io/ktor/client/plugins/CleartextTraffic.jvm.kt
@@ -1,0 +1,61 @@
+/*
+ * Copyright 2014-2025 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.client.plugins
+
+import java.lang.reflect.InvocationTargetException
+
+/**
+ * This explicit check avoids activating in Android Studio with Android specific classes available when running plugins inside the IDE.
+ */
+private val isAndroid: Boolean = System.getProperty("java.vm.name") == "Dalvik"
+
+internal actual fun isCleartextTrafficPermitted(hostname: String): Boolean {
+    if (!isAndroid) return true
+
+    // Stolen from https://github.com/square/okhttp/blob/65624f5f5b3dd3084b88e4608eaa55cd69d02e07/okhttp/src/main/kotlin/okhttp3/internal/platform/AndroidPlatform.kt#L105-L122
+    return try {
+        val networkPolicyClass = Class.forName("android.security.NetworkSecurityPolicy")
+        val getInstanceMethod = networkPolicyClass.getMethod("getInstance").apply { isAccessible = true }
+        val networkSecurityPolicy = getInstanceMethod(null)
+        api24IsCleartextTrafficPermitted(hostname, networkPolicyClass, networkSecurityPolicy)
+    } catch (_: ClassNotFoundException) {
+        true
+    } catch (_: NoSuchMethodException) {
+        true
+    } catch (e: IllegalAccessException) {
+        throw AssertionError("unable to determine cleartext support", e)
+    } catch (e: IllegalArgumentException) {
+        throw AssertionError("unable to determine cleartext support", e)
+    } catch (e: InvocationTargetException) {
+        throw AssertionError("unable to determine cleartext support", e)
+    }
+}
+
+@Throws(InvocationTargetException::class, IllegalAccessException::class)
+private fun api24IsCleartextTrafficPermitted(
+    hostname: String,
+    networkPolicyClass: Class<*>,
+    networkSecurityPolicy: Any
+): Boolean = try {
+    val isCleartextTrafficPermittedMethod = networkPolicyClass
+        .getMethod("isCleartextTrafficPermitted", String::class.java)
+        .apply { isAccessible = true }
+    isCleartextTrafficPermittedMethod(networkSecurityPolicy, hostname) as Boolean
+} catch (_: NoSuchMethodException) {
+    api23IsCleartextTrafficPermitted(networkPolicyClass, networkSecurityPolicy)
+}
+
+@Throws(InvocationTargetException::class, IllegalAccessException::class)
+private fun api23IsCleartextTrafficPermitted(
+    networkPolicyClass: Class<*>,
+    networkSecurityPolicy: Any
+): Boolean = try {
+    val isCleartextTrafficPermittedMethod = networkPolicyClass
+        .getMethod("isCleartextTrafficPermitted")
+        .apply { isAccessible = true }
+    isCleartextTrafficPermittedMethod(networkSecurityPolicy) as Boolean
+} catch (_: NoSuchMethodException) {
+    true
+}
--- a/ktor-client/ktor-client-core/api/ktor-client-core.api
+++ b/ktor-client/ktor-client-core/api/ktor-client-core.api
@@ -240,6 +240,10 @@ public final class io/ktor/client/plugins/BodyProgressKt {
 	public static final fun onUpload (Lio/ktor/client/request/HttpRequestBuilder;Lio/ktor/client/content/ProgressListener;)V
 }
 
+public final class io/ktor/client/plugins/CleartextTrafficKt {
+	public static final fun getCleartextTraffic ()Lio/ktor/client/plugins/api/ClientPlugin;
+}
+
 public final class io/ktor/client/plugins/ClientRequestException : io/ktor/client/plugins/ResponseException {
 	public fun <init> (Lio/ktor/client/statement/HttpResponse;Ljava/lang/String;)V
 	public fun getMessage ()Ljava/lang/String;
--- a/ktor-client/ktor-client-core/api/ktor-client-core.klib.api
+++ b/ktor-client/ktor-client-core/api/ktor-client-core.klib.api
@@ -1400,6 +1400,8 @@ final val io.ktor.client.plugins.websocket/pingInterval // io.ktor.client.plugin
     final inline fun (io.ktor.client.plugins.websocket/WebSockets).<get-pingInterval>(): kotlin.time/Duration? // io.ktor.client.plugins.websocket/pingInterval.<get-pingInterval>|<get-pingInterval>@io.ktor.client.plugins.websocket.WebSockets(){}[0]
 final val io.ktor.client.plugins/BodyProgress // io.ktor.client.plugins/BodyProgress|{}BodyProgress[0]
     final fun <get-BodyProgress>(): io.ktor.client.plugins.api/ClientPlugin<kotlin/Unit> // io.ktor.client.plugins/BodyProgress.<get-BodyProgress>|<get-BodyProgress>(){}[0]
+final val io.ktor.client.plugins/CleartextTraffic // io.ktor.client.plugins/CleartextTraffic|{}CleartextTraffic[0]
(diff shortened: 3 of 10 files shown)


rsinukov (commented):
Thanks for the PR.

Can you explain what is the current behaviour and what is the benefit of this change, please?

Goooler:
If we are using `CIO` to request an HTTP URL on Android 9+, there is no exception thrown, but thrown with the use of `OkHttp` or `Android` engine, cause the `isCleartextTrafficPermitted` check had been implemented by them, like [this](https://github.com/square/okhttp/blob/735ed1a6e5a15042f6d7e8f100fb3c8376f6aa27/okhttp/src/jvmMain/kotlin/okhttp3/internal/connection/RealRoutePlanner.kt#L205-L208), this is [a suggested behavior on Android](https://developer.android.com/reference/android/security/NetworkSecurityPolicy#isCleartextTrafficPermitted()):
> This flag is honored on a best effort basis because it's impossible to prevent all cleartext traffic from Android applications given the level of access provided to them.

We should catch up what Okhttp did on Android, to provide the same experience.

ktor-client/ktor-client-cio/jvmAndNix/src/io/ktor/client/engine/cio/Endpoint.kt
```diff
@@ -199,6 +200,12 @@ internal class Endpoint(
         connections.incrementAndGet()
 
         try {
+            if (!Platform.isCleartextTrafficPermitted(host)) {
+                throw UnknownServiceException(
+                    "CLEARTEXT communication to $host not permitted by network security policy"
+                )
+            }
```
Goooler, line 207: Not very sure if we should put this check here.

rsinukov (changes requested):

ktor-client/ktor-client-cio/jvm/src/io/ktor/client/engine/cio/Platform.kt
```diff
@@ -0,0 +1,61 @@
...
+ * Copyright 2014-2023 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.client.engine.cio
+
+import java.lang.reflect.InvocationTargetException
+
+internal actual object Platform {
+    /**
+     * his explicit check avoids activating in Android Studio with Android specific classes available when running plugins inside the IDE.
```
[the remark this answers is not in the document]
Goooler: ced616f2241c1a5e03adef9f3613849d83c25ab7

ktor-client/ktor-client-cio/jvmAndNix/src/io/ktor/client/engine/cio/Endpoint.kt
Goooler: @yschimke Can you take a look?

ktor-client/ktor-client-cio/jvm/src/io/ktor/client/engine/cio/Platform.kt
```diff
@@ -0,0 +1,61 @@
...
+
+import java.lang.reflect.InvocationTargetException
+
+internal actual object Platform {
+    /**
+     * This explicit check avoids activating in Android Studio with Android specific classes available when running plugins inside the IDE.
+     */
+    private val isAndroid: Boolean = System.getProperty("java.vm.name") == "Dalvik"
+
+    actual fun isCleartextTrafficPermitted(hostname: String): Boolean {
```
yschimke, line 15: These aren't the full story for OkHttp.

We compile against robolectric android-all in order to call these APIs directly.

https://github.com/square/okhttp/blob/okhttp_4x/okhttp/src/main/kotlin/okhttp3/internal/platform/AndroidPlatform.kt

It's way simpler than the reflection.

ktor-client/ktor-client-cio/jvmAndNix/src/io/ktor/client/engine/cio/Endpoint.kt
yschimke: I don't know enough about CIO to comment, is this HTTP only here? Or could it be HTTPS also?

Probably worth adding tests to validate the behaviour for both cases.

Some extra context for OkHttp.  We position it 

- After application interceptors have a chance to modify the request, in case they have an AlwaysSecureInterceptor configured.
- Before we connect via various options - proxies, or IPv4/IPv6 fallback, which might involve connecting to an IP instead of a host. So we avoid working around the platform restrictions.

ktor-client/ktor-client-cio/jvm/src/io/ktor/client/engine/cio/Platform.kt
Goooler: Yeah, I borrowed these code from the old implementation, cause I see there is no Android Runtime applied in CIO module.

yschimke: Same for OkHttp. It's a compile only dependency

e5l (changes requested):
Hey @Goooler, thank you for the PR.

It looks like this code is Android specific, it also could be helpful not only for CIO.
Could you please extract this check to the separate client plugin?

Goooler:
What do you mean? You mean I should move `Platform` into a new module under `ktor-client/ktor-client-plugins`?

e5l:
Yep! We also can install this plugin by default in case the VM is Dalvik

ktor-client/ktor-client-plugins/ktor-client-platform/common/src/Platforming.kt
```diff
@@ -0,0 +1,21 @@
...
+ */
+
+package io.ktor.client.plugins.platform
+
+import io.ktor.client.plugins.api.ClientPlugin
+import io.ktor.client.plugins.api.createClientPlugin
+import io.ktor.client.request.host
+import io.ktor.utils.io.errors.UnknownServiceException
+
+public val Platforming: ClientPlugin<Any> = createClientPlugin("Platforming", ::Any) {
```
Goooler, line 12: How can I install this plugin for CIO by default?

ktor-client/ktor-client-plugins/ktor-client-platform/nonJvm/src/Platform.kt
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright 2014-2024 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.client.plugins.platform
+
+internal actual object Platform {
```
Goooler, line 7: Seems `nonJvm` source set hasn't been recognized, how can I implement this actual object for non-jvm platforms?

image

ktor-client/ktor-client-plugins/ktor-client-platform/common/src/Platforming.kt
e5l: let's make this in the separate PR, it looks like it require service loader or some similar mechanism

ktor-client/ktor-client-plugins/ktor-client-platform/nonJvm/src/Platform.kt
Goooler: 9ff34673ab4ffe5a2889fe55f1600ac93ef4129a

Goooler:
Seems the style check failure is not related to my change.

ktor-client/ktor-client-plugins/ktor-client-platform/common/src/Platforming.kt
Goooler: We need

- https://youtrack.jetbrains.com/issue/KT-25892
- https://github.com/whyoleg/sweet-spi

wkornewald:
We really need this to be configurable for certain clients because OSCP checks very often require HTTP cleartext requests. If they're disallowed at the Android level it's impossible to enable OCSP.

yschimke:
@wkornewald are there android HTTP clients that will build on top of ktor, that actually do these checks themselves? I thought those checks are at the TLS/Conscrypt layer, well below Ktor?

https://github.com/google/conscrypt/blob/5bd23689463a69a80563178eb64b878587b723b3/common/src/main/java/org/conscrypt/TrustManagerImpl.java#L754

wkornewald:
When using Bouncy Castle instead of Conscrypt the check happens using the normal Java APIs instead of C/C++ and that fails on Android when cleartext traffic is disallowed. In our project we can't use Conscrypt because it ignores some security configs and is hard-coded to allow certain ciphers which we MUST disable due to government/certification requirements. Globally allowing cleartext traffic is something we want to still avoid, so we'd like to use a separate HTTP-allowed client just for OCSP checks while the rest of the app stays under the global cleartext-disallowed rule.

yschimke:
is this blocked on something, or just needs a rebase?

Goooler:
Seems it's still blocked by https://youtrack.jetbrains.com/issue/KT-25892.

Goooler:
I have an idea: install the built-in `CleartextTraffic` plugin for all `HttpClientEngine`s, and throw `UnknownServiceException`s before reaching OkHttp, HttpURLConnection, or any other engine-specific implementation. WDYT?

ktor-client/ktor-client-core/jvm/src/io/ktor/client/plugins/CleartextTraffic.jvm.kt
```diff
@@ -0,0 +1,66 @@
...
+    networkPolicyClass: Class<*>,
+    networkSecurityPolicy: Any
+): Boolean = try {
+    val isCleartextTrafficPermittedMethod = networkPolicyClass
+        .getMethod("isCleartextTrafficPermitted")
+        .apply { isAccessible = true }
+    isCleartextTrafficPermittedMethod(networkSecurityPolicy) as Boolean
+} catch (_: NoSuchMethodException) {
+    true
+}
```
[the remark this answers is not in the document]
Goooler: `()` is the invocation of `invoke`.

ktor-client/ktor-client-core/jvm/src/io/ktor/client/plugins/CleartextTraffic.jvm.kt
```diff
@@ -0,0 +1,66 @@
...
+    networkPolicyClass: Class<*>,
+    networkSecurityPolicy: Any
+): Boolean = try {
+    val isCleartextTrafficPermittedMethod = networkPolicyClass
+        .getMethod("isCleartextTrafficPermitted", String::class.java)
+        .apply { isAccessible = true }
+    isCleartextTrafficPermittedMethod(networkSecurityPolicy, hostname) as Boolean
+} catch (_: NoSuchMethodException) {
+    api23IsCleartextTrafficPermitted(networkPolicyClass, networkSecurityPolicy)
+}
```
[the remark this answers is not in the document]
Goooler: `()` is the invocation of `invoke`.

PR #3726, opened 2023-08-11
[QUIC] Packets retransmission 1

Mr3zee:
**Subsystem**
ktor-network

**Motivation**
QUIC packet retransmission requires caching individual frames, which in current solution is hard

**Solution**
Refactor frames and packets writing operations to allow to introduce recovery module


e5l (approved):
We should use Array or drop List<Long> in the future

PR #3727, opened 2023-08-14
KTOR-6172 Add Server-sent events (SSE) plugin for server

marychatte:
[KTOR-6172](https://youtrack.jetbrains.com/issue/KTOR-6172)

[Design](https://youtrack.jetbrains.com/issue/KTOR-6172/Add-Server-sent-events-SSE-plugin-for-server#focus=Comments-27-7997979.0-0)


ktor-server/ktor-server-plugins/ktor-server-sse/jvmAndNix/src/io/ktor/server/sse/DefaultServerSSESession.kt
```diff
@@ -0,0 +1,38 @@
...
+import io.ktor.utils.io.*
+import kotlinx.coroutines.sync.*
+import kotlin.coroutines.*
+
+internal class DefaultServerSSESession(
+    private val output: ByteWriteChannel,
+    override val call: ApplicationCall,
+    override val coroutineContext: CoroutineContext
+) : ServerSSESession {
+    private val mutex = Mutex()
```
rsinukov, line 19: why is it needed?

rsinukov (commented):
Can you list a new API and it's usage here in the ticket, please?

ktor-server/ktor-server-plugins/ktor-server-sse/jvmAndNix/src/io/ktor/server/sse/DefaultServerSSESession.kt
marychatte: Because concurrent writing is not supported for `ByteWriteChannel` and we need to order writing events

e5l (approved):

rsinukov (approved):

mcxinyu:
But it seems to be blocking when used, and the client appears to receive the data after the server has sent all the data.

mcxinyu:
Did I misunderstand? Isn't the SSE function that the server calls `send(event: ServerSendEvent)` once and the client receives data once?
So, the `EventSource.addEventListener ('message ', function (e) {}` in JavaScript is no longer effective, so it's better to directly initiate an HTTP GET request.

marychatte:
> But it seems to be blocking when used, and the client appears to receive the data after the server has sent all the data.

Hi @mcxinyu! It shouldn't be like this. Can you please provide a code with an example?

mcxinyu:
> > But it seems to be blocking when used, and the client appears to receive the data after the server has sent all the data.
> 
> Hi @mcxinyu! It shouldn't be like this. Can you please provide a code with an example?

Hi, @marychatte

I tried to find the problem and found out that the Compression plugin was used, which has implementation `ConditionsHolderBuilder.minimumSize(minSize: Long)` for compressing logic when `OutgoingContent.contentLength` is null, which is worth discussing.

Because the `OutgoingContent.contentLength` of `send (event: ServerSendEvent)` is null, this condition will be triggered.

https://github.com/ktorio/ktor/blob/ed3392de048c141ce0600aa13dee2b60ec9a910d/ktor-server/ktor-server-plugins/ktor-server-compression/jvm/src/io/ktor/server/plugins/compression/Config.kt#L170

You can find the example I wrote below to reproduce the problem.

https://github.com/mcxinyu/ktor/blob/cac427761a6ac8ec4d6286fce541339072314bbe/ktor-server/ktor-server-plugins/ktor-server-sse-test/jvmAndNix/src/io/ktor/server/sse/test/Application.kt

marychatte:
@mcxinyu Thank you! I created an issue: [KTOR-6327](https://youtrack.jetbrains.com/issue/KTOR-6327) Disable compression for SSE requests

PR #3735, opened 2023-08-24
Release buildPacket on success

JordanLongstaff:
**Subsystem**
`io.ktor.utils.io.core`

**Motivation**
The function `buildPacket` is useful, but it has a try-catch to catch exceptions in building a packet, and releases the resources of the `BytePacketBuilder` _only_ if an exception is caught. If the packet builds successfully, the resources do not get released.

**Solution**
Replace the `catch` with a `finally`, which will run and release resources regardless of whether the packet is built successfully.


e5l:
Hey @JordanLongstaff, in case of success, resources are transferred to the result

PR #3736, opened 2023-08-25
KTOR-6217 SSE plugin support in client engines

marychatte:
[KTOR-6217](https://youtrack.jetbrains.com/issue/KTOR-6217)


e5l (approved):
lgtm

PR #3737, opened 2023-08-27
Support  Websocket Subprotocols For JS Clients (KTOR-4001)

undermark5:
**Subsystem**
JS Engines Websockets
Java Engine Websockets

**Motivation**
Fixes KTOR-4001

There needs to be a way to set the subprotocols the websocket supports/requires (See [KTOR-4001](https://youtrack.jetbrains.com/issue/KTOR-4001/Server-sent-a-subprotocol-but-none-was-requested-when-using-Node-WebSockets#focus=Comments-27-5866649.0-0)) because certain server implementations will reject connections that do not specify the correct subprotocols, or [being important for handler resolution](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/WebSocket#parameters). Currently, [apollographql/apollo-kotlin](https://github.com/apollographql/apollo-kotlin) (a downstream consumer of KTOR) has implemented a work around because the `createWebSocket` method provided by KTOR does not allow for passing the protocols directly, nor does it support consuming from the [standard protocol upgrade header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Protocol_upgrade_mechanism#sec-websocket-protocol) `sec-websocket-protocol`.

**Solution**
This implementation is based off the the [downstream workaround in apollo-kotlin](https://github.com/apollographql/apollo-kotlin/blob/3c429280d290ec885512497cf34f57181b667a7e/libraries/apollo-runtime/src/jsMain/kotlin/com/apollographql/apollo3/network/ws/JsWebSocketEngine.kt#L84).

Another option is that KTOR could expose a `protocols` parameter to `createWebSocket` rather than relying solely on the headers, though one would likely want also define/document how any conflicts would be resolved if that parameter were to be exposed.

------------------------------
[Initial Discovery of issue](https://github.com/apollographql/apollo-kotlin/pull/3913#issuecomment-1065745594)
[Another case of server rejecting WS connection](https://github.com/apollographql/apollo-kotlin/pull/4445#issue-1400572611)

### DIFF
--- a/ktor-client/ktor-client-core/js/src/io/ktor/client/engine/js/JsClientEngine.kt
+++ b/ktor-client/ktor-client-core/js/src/io/ktor/client/engine/js/JsClientEngine.kt
@@ -66,8 +66,8 @@ internal class JsClientEngine(
         urlString_capturingHack: String,
         headers: Headers
     ): WebSocket {
-        val (protocolHeaderNames, otherHeaderNames) = headers.names().partition {
-            it.equals("sec-websocket-protocol", true)
+        val protocolHeaderNames = headers.names().filter { headerName ->
+            headerName.equals("sec-websocket-protocol", true)
         }
         val protocols = protocolHeaderNames.mapNotNull { headers.getAll(it) }.flatten().toTypedArray()
         return when (PlatformUtils.platform) {
--- a/ktor-client/ktor-client-core/js/src/io/ktor/client/engine/js/JsClientEngine.kt
+++ b/ktor-client/ktor-client-core/js/src/io/ktor/client/engine/js/JsClientEngine.kt
@@ -65,23 +65,21 @@ internal class JsClientEngine(
     private fun createWebSocket(
         urlString_capturingHack: String,
         headers: Headers
-    ): WebSocket = when (PlatformUtils.platform) {
-        Platform.Browser -> js("new WebSocket(urlString_capturingHack)")
-        else -> {
-            val ws_capturingHack = js("eval('require')('ws')")
-            val headers_capturingHack: dynamic = object {}
-            val subProtocolHeaderValues = mutableListOf<String>()
-            headers.forEach { name, values ->
-                // Capture any and all websocket subprotocol headers to pass to the WebSocket constructor
-                if (name.equals("sec-websocket-protocol", true)) {
-                    subProtocolHeaderValues.addAll(values)
+    ): WebSocket {
+        val (protocolHeaderNames, otherHeaderNames) = headers.names().partition { it.equals("sec-websocket-protocol", true) }
+        val protocols = protocolHeaderNames.mapNotNull { headers.getAll(it) }.flatten().toTypedArray()
+        return when (PlatformUtils.platform) {
+            Platform.Browser -> js("new WebSocket(urlString_capturingHack, protocols)")
+            else -> {
+                val ws_capturingHack = js("eval('require')('ws')")
+                val headers_capturingHack: dynamic = object {}
+                headers.forEach { name, values ->
+                    headers_capturingHack[name] = values.joinToString(",")
                 }
-                headers_capturingHack[name] = values.joinToString(",")
+                js("new ws_capturingHack(urlString_capturingHack, protocols, { headers: headers_capturingHack })")
             }
-            val subprotocols_capturingHack = subProtocolHeaderValues.toTypedArray()
-            js("new ws_capturingHack(urlString_capturingHack, subprotocols_capturingHack, { headers: headers_capturingHack })")
         }
-        }
+    }
 
     private suspend fun executeWebSocketRequest(
         request: HttpRequestData,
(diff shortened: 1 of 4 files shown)


ktor-client/ktor-client-core/js/src/io/ktor/client/engine/js/JsClientEngine.kt
```diff
@@ -65,15 +65,21 @@ internal class JsClientEngine(
...
         headers: Headers
-    ): WebSocket = when (PlatformUtils.platform) {
-        Platform.Browser -> js("new WebSocket(urlString_capturingHack)")
-        else -> {
-            val ws_capturingHack = js("eval('require')('ws')")
-            val headers_capturingHack: dynamic = object {}
-            headers.forEach { name, values ->
-                headers_capturingHack[name] = values.joinToString(",")
+    ): WebSocket {
+        val (protocolHeaderNames, otherHeaderNames) = headers.names().partition {
```
rsinukov, line 69: why `partition` and not just `filter`?

rsinukov (changes requested):
Hi and thanks for the PR. Can you please add a test for the change?

ktor-client/ktor-client-core/js/src/io/ktor/client/engine/js/JsClientEngine.kt
undermark5: That may be an artifact of the apollographql specific work around (there is an instance where the other headers are examined in their code, I'll double check this here to see if it still makes sense, but filter would probably be good enough)

undermark5:
@rsinukov got busy with other things at the moment and forgot to post here. I'm unfamiliar with the testing structure that is setup in the project, and couldn't easily find documentation around adding additional tests or where the existing tests live, so some guidance around this would be helpful.

rsinukov:
Sorry for the delay in the reply. 
You can add a new test route for the server in `buildSrc/src/main/kotlin/test/server/tests/WebSockets.kt`. After this you can add a new test in `ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/WebSocketTest.kt` based on other tests in this file. This test will be used for all available engines. 
You may need to set `ktor.ide.jvmAndCommonOnly=false` inside `gradle.properties` to run JS tests locally.

rsinukov (approved):
LGTM

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/WebSocketTest.kt
```diff
@@ -299,7 +299,7 @@ class WebSocketTest : ClientLoader() {
         }
 
         test { client ->
-            assertFailsWith<WebSocketException> {
+            assertFailsWith<Exception> {
```
undermark5, line 302: It seems like the different engines can throw different exceptions and I'm unable to figure out a mechanism to handle these exceptions gracefully, the upgrade requests are correctly failing when missing the sub protocol.

undermark5:
So, it would appear that the Java client engine also is failing to respect the subprotocols from the headers and there doesn't appear to be a way to set them directly. If it did, we wouldn't see `io.ktor.client.tests.WebSocketTest.testWebsocketRequiringSubProtocolWithSubProtocol[Java][jvm]` [fail](https://ktor.teamcity.com/buildConfiguration/Ktor_KtorMatrixCoreLinuxJava11/231427?buildTab=tests&status=failed&expandedTest=build%3A%28id%3A231427%29%2Cid%3A2000002183)

undermark5:
Something like [this](https://github.com/Undermark5-Software/ktor/commit/696ef8ca301bb798d418b539eec83acf795a728a) should fix that Should I also include that here? 4001 was specifically for JS client engines, but if there are other client engines with similar problems it may make sense to also include them here.

rsinukov:
Thanks for the investigation. We will appreciate if you can fix it in the java engine as well.

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/WebSocketTest.kt
```diff
@@ -271,4 +272,38 @@ class WebSocketTest : ClientLoader() {
...
+                    header(HttpHeaders.SecWebSocketProtocol, "test-protocol")
+                }
+            ) {
+                send(Frame.Text("test"))
+                val result = incoming.receive() as Frame.Text
+                assertEquals("test", result.readText())
+            }
+        }
+    }
+    @Test
```
rsinukov, line 295: codestyle will fail here, because it requires a new line in between two methods

rsinukov (changes requested):
Almost LGTM. I re-run the CI.

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/WebSocketTest.kt
undermark5: This should be fixed now.

rsinukov (approved):

PR #3738, opened 2023-08-27
KTOR-6221: Fix reduced concurrent reqs in Apache5

Sebmaster:
**Subsystem**
Client, Apache5 engine

**Motivation**
When using the Apache5 engine, total concurrent requests to a single route were limited to 5 requests. This causes an unexpected bottleneck when switching engines.

Issue link is https://youtrack.jetbrains.com/issue/KTOR-6221/Apache5-engine-limits-concurrent-requests-to-individual-route-to-5.

**Solution**
This just fixes the type/missing call to `setMaxConnPerRoute` which was intended all along.


rsinukov (approved):
lgtm, thanks!

PR #3662, opened 2023-06-27
Update Gradle version to 8

e5l:

PR #3743, opened 2023-09-01
Update Dokka to 1.9.0 and add a workaround for build errors

IgnatBeresnev:
**Subsystem**
Infrastructure

**Motivation**
Dokka stopped working after #3662 was merged into `main`. If you run Dokka's tasks now, it will lead to build failures related to the resolution of platform dependencies, likely caused by the combination of Gradle 8 + `kotlin.mpp.enableCInteropCommonization=true`

**Solution**
Update Dokka to 1.9.0 and use the proposed workaround flag, which changes the underlying dependency resolution algorithm.

For more details, see https://github.com/Kotlin/dokka/issues/3153.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -91,7 +91,7 @@ val disabledExplicitApiModeProjects = listOf(
 apply(from = "gradle/compatibility.gradle")
 
 plugins {
-    id("org.jetbrains.dokka") version "1.7.20" apply false
+    id("org.jetbrains.dokka") version "1.9.0" apply false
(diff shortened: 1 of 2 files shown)


IgnatBeresnev:
Note: because ktor has many nested modules, you need to run Dokka's multimodule task with a semicolon (i.e only on the root project): 
```
./gradlew :dokkaHtmlMultiModule
```

I've sent a fix for the publishing script and explained it in more detail there, see https://github.com/ktorio/api.ktor.io/pull/11

rsinukov (approved):

PR #3748, opened 2023-09-18
KTOR-6292 Make client use Dispatchers.IO by default

e5l:
Fix [KTOR-6292](https://youtrack.jetbrains.com/issue/KTOR-6292)


gradle.properties
```diff
@@ -4,7 +4,7 @@
 
 # ktor
 # Makes Ktor import and resolve
-ktor.ide.jvmAndCommonOnly=true
+ktor.ide.jvmAndCommonOnly=false
```
rsinukov, line 7: is it intended?

rsinukov (approved):

gradle.properties
e5l: nope, will revert

PR #3751, opened 2023-09-18
Optimizing imports for ktor-utils

mjovanc:
**Motivation**
Removing unused imports and optimizing imports that can be improved using IntelliJ IDEA "Optimize imports".

### DIFF
--- a/ktor-utils/jvm/src/io/ktor/util/date/DateJvm.kt
+++ b/ktor-utils/jvm/src/io/ktor/util/date/DateJvm.kt
@@ -4,7 +4,6 @@
 
 package io.ktor.util.date
 
-import io.ktor.util.*
 import java.util.*
 
 private val GMT_TIMEZONE = TimeZone.getTimeZone("GMT")


mjovanc:
Is the build failure known? I see that commits on `main` branch has failures as well.

rsinukov:
Hi! Do you mind running `./gradlew formatKotlin` to fix the checkstyle and commit it here?

rsinukov:
There are build error on native targets. Can you check it please? You may need to set `jvmAndCommonOnly=false` inside `gradle.properties`

mjovanc:
> There are build error on native and jvm targets. Can you check it please? You may need to set `jvmAndCommonOnly=false` inside `gradle.properties`

Sure. Will check it out right away.

mjovanc:
@rsinukov Rebased and updated. Now it works to run the task assemble without build errors.

mjovanc:
@rsinukov I think it passed now. I changed to use only ktor-utils. I think it's best to split it up for each top level library.

rsinukov:
Hey! Again sorry for the long reply. OOM in stress tests are known issue. Same goes for win and macOS builds.

rsinukov (approved):

ktor-utils/common/src/io/ktor/util/internal/LockFreeLinkedList.kt
```diff
@@ -11,8 +11,6 @@ package io.ktor.util.internal
  */
 
 import kotlinx.atomicfu.*
-import kotlin.jvm.*
```
rsinukov, line 14: Can you revert it please? The build fails on native without it.

rsinukov (changes requested):

ktor-utils/common/src/io/ktor/util/internal/LockFreeLinkedList.kt
mjovanc: Solved.

rsinukov:
Compilation is still failing on win

mjovanc:
> Compilation is still failing on windows

Sorry for the spamming, but I do not have a Windows machine to test it locally. I made some changes now again to see if it passes.

mjovanc:
Added back some uncertain imports to see if the build passes. I have no way of testing it locally on Windows so sorry for the spamming.

PR #3763, opened 2023-09-28
Compression conditions should not be triggered when `OutgoingContent.contentLength` is null

mcxinyu:
Compression conditions should not be triggered when `OutgoingContent.contentLength` is null.


rsinukov:
Hi @mcxinyu,
Can you explain why do you need this change, please?

mcxinyu:
> Hi @mcxinyu, Can you explain why do you need this change, please?

The reason can be found here. #3727 [issuecomment](https://github.com/ktorio/ktor/pull/3727#issuecomment-1738297061)

mcxinyu:
> Hi @mcxinyu, Can you explain why do you need this change, please?

I think that when `contentLength` is null, the compression logic should not be called because the data may be empty.

rsinukov:
But not necessary, the response may have chunked encoding. Are there any specific problems that you try to address here?

mcxinyu:
I didn't know there was such a situation.

PR #3769, opened 2023-09-29
KTOR-6327 Disable compression for SSE on server side

marychatte:
[KTOR-6327](https://youtrack.jetbrains.com/issue/KTOR-6327)


e5l (approved):

ktor-server/ktor-server-tests/jvm/test/io/ktor/server/plugins/CompressionTest.kt
```diff
@@ -741,6 +742,28 @@ class CompressionTest {
         assertEquals("unknown", responseUnknown.bodyAsText())
     }
 
+    @Test
+    fun testSkipCompressionForSSEResponse(): Unit = testApplication {
+        install(Compression) {
+            deflate {
+                minimumSize(1024)
```
rsinukov, line 749: should it be less than response to enable compression?

PR #3772, opened 2023-10-03
Optimizing imports in ktor-client library

mjovanc:
**Motivation**
Removing unused imports and optimizing imports that can be improved using IntelliJ IDEA "Optimize imports" for the `ktor-client` and its sub-libraries.


ktor-client/ktor-client-core/common/test/ClientPluginsTest.kt
```diff
@@ -5,18 +5,17 @@
...
 import io.ktor.client.request.*
 import io.ktor.client.statement.*
 import io.ktor.http.*
 import io.ktor.http.content.*
 import io.ktor.test.dispatcher.*
-import io.ktor.util.*
-import io.ktor.utils.io.*
-import io.ktor.utils.io.core.*
-import kotlinx.coroutines.*
+import kotlin.Boolean
```
rsinukov, line 14: Why do you need to import `kotlin.` packages?

rsinukov:
Please check native builds on CI

mjovanc:
Adding back the native libraries, since that was not a purpose to do so.

rsinukov (approved):

PR #3781, opened 2023-10-12
KTOR-6356 Make DefaultHeaders plugin Kotlin native compatible

marychatte:
[KTOR-6356](https://youtrack.jetbrains.com/issue/KTOR-6356)

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-default-headers/nix/src/io/ktor/server/plugins/defaultheaders/DefaultHeaders.nix.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-default-headers/nix/src/io/ktor/server/plugins/defaultheaders/DefaultHeaders.nix.kt
@@ -4,26 +4,11 @@
 
 package io.ktor.server.plugins.defaultheaders
 
+import io.ktor.server.application.*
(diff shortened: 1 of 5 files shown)


ktor-server/ktor-server-plugins/ktor-server-default-headers/nix/src/io/ktor/server/plugins/defaultheaders/DefaultHeaders.nix.kt
```diff
@@ -0,0 +1,29 @@
...
+
+package io.ktor.server.plugins.defaultheaders
+
+import kotlinx.cinterop.*
+import platform.posix.*
+
+@OptIn(ExperimentalForeignApi::class)
+internal actual fun readVersion(): String {
+    memScoped {
+        val versionFilePath = "/nix/resources/ktor-version.txt"
```
e5l, line 13: it looks like it will not work if the user doesn't have "ktor-version.txt" file in the application resources

e5l (changes requested):

e5l (approved):
Awesome, good job! LGTM

PR #3782, opened 2023-10-16
KTOR-6281 Darwin: EOFException when sending multipart data using Ktor

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-6281


e5l (approved):
LGTM

e5l:
Please check code style before merging. Could you tell me if we can move test to common?

PR #3786, opened 2023-10-18
KTOR-2682 Add RequestBodyLimit plugin

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-2682


ktor-server/ktor-server-plugins/ktor-server-body-limit/jvmAndNix/src/io/ktor/server/plugins/bodylimit/RequestBodyLimit.kt
```diff
@@ -0,0 +1,74 @@
...
+public val RequestBodyLimit: RouteScopedPlugin<RequestBodyLimitConfig> = createRouteScopedPlugin(
+    "RequestBodyLimit",
+    ::RequestBodyLimitConfig
+) {
+
+    val bodyLimit = pluginConfig.bodyLimit
+
+    on(BeforeReceive) { call, content ->
+        val limit = bodyLimit(call)
+        if (limit == Long.MAX_VALUE) return@on null
```
e5l, line 41: Please check ContentLength in advance

ktor-server/ktor-server-plugins/ktor-server-body-limit/jvmAndNix/src/io/ktor/server/plugins/bodylimit/RequestBodyLimit.kt
```diff
@@ -0,0 +1,74 @@
...
+import io.ktor.server.plugins.*
+import io.ktor.server.request.*
+import io.ktor.utils.io.*
+import io.ktor.utils.io.pool.*
+import kotlinx.coroutines.*
+
+/**
+ * A configuration for [RequestBodyLimit] plugin.
+ */
+public class RequestBodyLimitConfig {
```
e5l, line 17: Could you tell me if we can have a method on `call.request` to adjust requirements for a specific call in routing?

e5l (changes requested):
Could we also introduce API to

e5l (approved):

PR #3789, opened 2023-10-19
Optimizing imports in ktor-shared library

mjovanc:
**Motivation**
Removing unused imports and optimizing imports that can be improved using IntelliJ IDEA "Optimize imports" for the `ktor-shared` and its sub-libraries.


rsinukov (approved):
lgtm

PR #3793, opened 2023-10-23
KTOR-6390 Fix error message for OkHttpSSESession

marychatte:
[KTOR-6390](https://youtrack.jetbrains.com/issue/KTOR-6390)

### DIFF
--- a/ktor-client/ktor-client-okhttp/jvm/src/io/ktor/client/engine/okhttp/OkHttpSSESession.kt
+++ b/ktor-client/ktor-client-okhttp/jvm/src/io/ktor/client/engine/okhttp/OkHttpSSESession.kt
@@ -5,6 +5,7 @@
 package io.ktor.client.engine.okhttp
 
 import io.ktor.client.plugins.sse.*
+import io.ktor.http.*
 import io.ktor.sse.*
 import kotlinx.coroutines.*
 import kotlinx.coroutines.channels.*
@@ -36,8 +37,7 @@ internal class OkHttpSSESession(
     }
 
     override fun onFailure(eventSource: EventSource, t: Throwable?, response: Response?) {
-        val error = t?.let { SSEException(it) }
-            ?: SSEException("Unexpected error occurred")
+        val error = t?.let { SSEException(it) } ?: mapException(response)
         originResponse.completeExceptionally(error)
         _incoming.close()
         serverSentEventsSource.cancel()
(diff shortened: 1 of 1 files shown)


rsinukov (approved):

ktor-client/ktor-client-core/common/src/io/ktor/client/request/HttpRequest.kt
```diff
@@ -309,3 +309,9 @@ public fun HttpRequestData.isUpgradeRequest(): Boolean {
 public fun HttpRequestData.isSseRequest(): Boolean {
     return body is SSEClientContent
 }
+
+@InternalAPI
+@Suppress("KDocMissingDocumentation")
+public fun needToProcessSSE(data: HttpRequestData, status: HttpStatusCode): Boolean {
+    return data.isSseRequest() && status != HttpStatusCode.Unauthorized
```
rsinukov, line 316: Why only `Unauthorized`? What if it's redirect, for example?

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/sse/SSE.kt
```diff
@@ -58,14 +60,28 @@ public val SSE: ClientPlugin<SSEConfig> = createClientPlugin(
...
     }
 
-    transformResponseBody { response, content, requestedType ->
-        if (content !is ClientSSESession) {
-            LOGGER.trace("Skipping non SSE response from ${response.request.url}: $content")
-            content
-        } else {
-            LOGGER.trace("Receive SSE session from ${response.request.url}: $content")
-            HttpResponseContainer(requestedType, content)
+    client.responsePipeline.intercept(HttpResponsePipeline.Transform) { (info, session) ->
```
rsinukov, line 63: why not use `transformResponseBody` here?

ktor-client/ktor-client-core/common/src/io/ktor/client/request/HttpRequest.kt
```diff
@@ -309,3 +309,11 @@ public fun HttpRequestData.isUpgradeRequest(): Boolean {
 public fun HttpRequestData.isSseRequest(): Boolean {
     return body is SSEClientContent
 }
+
+@InternalAPI
+@Suppress("KDocMissingDocumentation")
+public fun needToProcessSSE(data: HttpRequestData, status: HttpStatusCode, headers: Headers): Boolean {
+    return data.isSseRequest() &&
+        status == HttpStatusCode.OK &&
```
rsinukov, line 317: not sure if it's needed here, but can we check that status is successful instead of OK?

rsinukov (approved):

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/sse/SSE.kt
marychatte: Because in `transformResponseBody` we have the next check: `if (content !is ByteReadChannel) return@intercept` and `ClientSSESession` isn't a `ByteReadChannel`

ktor-client/ktor-client-core/common/src/io/ktor/client/request/HttpRequest.kt
marychatte: Can we leave `OK` because in spec it says next: `if res's status is not 200, or if res's Content-Type is not text/event-stream, then fail the connection`?

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/sse/SSE.kt
rsinukov: ok

PR #3806, opened 2023-11-03
KTOR-6420 ContentNegotiation: Adding charset to content type breaks request matching

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-6420


e5l (approved):
lgtm

e5l:
Could you check the CI?

PR #3814, opened 2023-11-07
KTOR-6318 Calculate session encryption mac over encrypted value

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-6318


e5l (approved):
oh, it was hard

lgtm

PR #3824, opened 2023-11-10
Fix ReadTimeoutException on Netty

marychatte:
Before in the case of `ReadTimeoutException`:
1. The client gets an empty response
2. Only `Netty` knows about exception 
3. The request is still executed

Now in the case of `ReadTimeoutException`:
1. The client gets a `408 Request Timeout` response
2. Ktor server will get a `CancellationException` and response will contain status `500`
3. Connection with a client is closed

### DIFF
--- a/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationCallHandler.kt
+++ b/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationCallHandler.kt
@@ -5,16 +5,18 @@
 package io.ktor.server.netty
 
 import io.ktor.http.*
+import io.ktor.http.HttpHeaders
 import io.ktor.server.application.*
 import io.ktor.server.engine.*
 import io.ktor.server.netty.http1.*
-import io.ktor.server.response.*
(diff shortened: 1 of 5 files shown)


ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationCallHandler.kt
```diff
@@ -47,6 +54,19 @@ internal class NettyApplicationCallHandler(
         }
     }
 
+    override fun exceptionCaught(ctx: ChannelHandlerContext, cause: Throwable) {
+        when (cause) {
+            is ReadTimeoutException -> {
+                runBlocking {
```
e5l, line 60: I see 2 problems here:
- runBlocking can possibly block netty thread from handling requests if `call.respond` contains blocking logic
- `call.respond` could be executed independently in other thread and `double response` can be thrown. 

Could we try cancelling the coroutine handling call instead, so it respond in `finally` block?

e5l (changes requested):
lgtm, except runBlocking

e5l (approved):
Awesome! Lgtm

PR #3829, opened 2023-11-14
KTOR-6400 Add attribute for extending isStaticContent() functionality in Webjars

bjhham:
**Subsystem**
StaticContent, Webjars

**Motivation**
[KTOR-6400](https://youtrack.jetbrains.com/issue/KTOR-6400) Marking files provided by Webjars plugin as static files

**Solution**
Introduced a public attribute for any static file response producer to add the path to the call.  This allows plugins like `Webjars` (or any other) to return true for the `ApplicationCall.isStaticContent()` extension function.  Previously, this function only checked if a path parameter was present from the routing.

### DIFF
--- a/ktor-server/ktor-server-core/jvm/src/io/ktor/server/http/content/StaticContent.kt
+++ b/ktor-server/ktor-server-core/jvm/src/io/ktor/server/http/content/StaticContent.kt
@@ -16,6 +16,13 @@ import java.net.*
 import java.nio.file.*
 import kotlin.io.path.*
 
+/**
+ * Attribute to assign the path of a static file served in the response.  The main use of this attribute is to indicate
+ * to subsequent interceptors that a static file was served via the `ApplicationCall.isStaticContent()` extension
+ * function.
+ */
+public val StaticFileLocationProperty: AttributeKey<String> = AttributeKey("StaticFileLocation")
+
 private const val pathParameterName = "static-content-path-parameter"
 
 private val staticRootFolderKey = AttributeKey<File>("BaseFolder")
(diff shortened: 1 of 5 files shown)


ktor-server/ktor-server-plugins/ktor-server-webjars/jvm/test/io/ktor/server/webjars/WebjarsTest.kt
```diff
@@ -130,6 +135,24 @@ class WebjarsTest {
         }
     }
 
+    @Test
+    fun classifiedAsStatic() {
+        var isStatic = false
+        var location = ""
+        withTestApplication {
```
rsinukov, line 142: `withTestApplication { ... }` is deprecated and we try to use new API `testApplication { ... }`

ktor-server/ktor-server-core/jvm/src/io/ktor/server/http/content/StaticContent.kt
```diff
@@ -16,6 +16,13 @@ import java.net.*
 import java.nio.file.*
 import kotlin.io.path.*
 
+/**
+ * Attribute to assign the path of a static file served in the response.  The main use of this attribute is to indicate
+ * to subsequent interceptors that a static file was served via the `ApplicationCall.isStaticContent()` extension
+ * function.
+ */
+public val StaticFileLocationProperty: AttributeKey<String> = AttributeKey("StaticFileLocation")
```
rsinukov, line 24: do we actually need the path? maybe just add a unit as a marker?

rsinukov (changes requested):
Some minor comments.

Also, whenever you add/change/delete public API, you need to run `./gradlew apiDump` and commit changes to keep track of binary compatibility

ktor-server/ktor-server-plugins/ktor-server-webjars/jvm/test/io/ktor/server/webjars/WebjarsTest.kt
bjhham: Oh right I meant to ask - in this case where the whole file is the old way would it be better to retrofit to the new method or just leave what's there?

rsinukov (approved):
LGTM, but please check the codestyle. You can run `./gradlew formatKotlin` locally to auto fix it

marychatte (approved):
Great work! Also, thanks for test refactoring!

PR #3830, opened 2023-11-15
KTOR-2121 Update trailcard logic for handling trailing slashes

bjhham:
**Subsystem**
Routing

**Motivation**
[KTOR-2121](https://youtrack.jetbrains.com/issue/KTOR-2121) {...} (tailcard) does not match URLs ending with '/'

**Solution**
Made it so the tailcard selector treats empty segments the same as others, and allowed failures for when there is no segment.

This is a pretty significant change to the routing logic, and could break some people if they rely on routes like `/foo/{...}` to match on `/foo`.  I was thinking maybe we ought to include some flag to opt in or out of this behaviour, wdyt?

There were also some intentional tests for the opposite behaviour than described in the bug, although the logic there is inconsistent:

1. Routes like `/foo/{...}/` should be considered invalid because the wildcard here is described as "the rest of the URL path" in the documentation.
2. There's no reason why `/foo` should match on `/foo/{...}`, it ought to be matched as `/foo{...}`.

Edit: change (2) was reverted, we rely too much on this behaviour and it would be too disruptive

### DIFF
--- a/ktor-server/ktor-server-tests/jvmAndNix/test/io/ktor/tests/server/routing/RoutingResolveTest.kt
+++ b/ktor-server/ktor-server-tests/jvmAndNix/test/io/ktor/tests/server/routing/RoutingResolveTest.kt
@@ -21,7 +21,8 @@ fun resolve(
     routing: RouteNode,
     path: String,
     parameters: Parameters = Parameters.Empty,
-    headers: Headers = Headers.Empty
+    headers: Headers = Headers.Empty,
+    tracers: List<(RoutingResolveTrace) -> Unit> = emptyList(),
 ): RoutingResolveResult {
     return withTestApplication {
         RoutingResolveContext(
@@ -36,7 +37,7 @@ fun resolve(
                 }
                 headers.flattenForEach { name, value -> request.addHeader(name, value) }
             },
-            emptyList()
+            tracers
         ).resolve()
     }
 }
@@ -409,8 +410,8 @@ class RoutingResolveTest {
             }
         }
 
-        on("resolving /foo") {
-            val result = resolve(root, "/foo")
+        on("resolving /foo/") {
+            val result = resolve(root, "/foo/")
 
             it("should successfully resolve") {
                 assertTrue(result is RoutingResolveResult.Success)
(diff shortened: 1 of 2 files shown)


rsinukov:
Please check failing tests

bjhham:
So after reviewing the failed tests, they're all caused from correcting the false positive match on the `/foo` route, and now I'm thinking plenty of people might rely on this behaviour, so I've decided we should probably let this sleeping dog lie for now.  For this fix, I'll just ensure matching works on the trailing `/`.

ktor-server/ktor-server-tests/jvmAndNix/test/io/ktor/tests/server/routing/RoutingResolveTest.kt
```diff
@@ -452,8 +453,8 @@ class RoutingResolveTest {
             }
         }
 
-        on("resolving /foo") {
-            val result = resolve(root, "/foo")
+        on("resolving /foo/") {
```
e5l, line 456: Could you tell me why the old test has been changed? we need to make sure that we don't break any old cases or at least document the breaking change

e5l (changes requested):

ktor-server/ktor-server-tests/jvmAndNix/test/io/ktor/tests/server/routing/RoutingResolveTest.kt
bjhham: I forgot to roll back this change in the test actually, the behaviour in this case will remain the same.  I had originally changed it but decided it would be too big of a breaking change to carry it forward.  It was originally a fix the first case in the ticket:

Expected:

    /api/subtree - do not match (according to [KTOR-372](https://youtrack.jetbrains.com/issue/KTOR-372/Ktor-should-resolve-trailing-slash-as-a-different-URL) logic)

Current behavior:

    /api/subtree - match
    
But as it turns out we rely quite a bit on this behaviour, especially for serving static files and the like.  I think if we want to change this particular one it will need to be a separate ticket with some configuration item.

rsinukov (approved):

e5l (approved):

PR #3849, opened 2023-11-26
Attributes: Compute if absent, suspend variant

nathanfallet:
**Subsystem**
Ktor utils

**Motivation**
`Attributes` provides a `computeIfAbsent` method, but this method cannot be called with a suspend block. For example, you want to get something from your database on a `call` when handling a request on ktor server and cache it in attributes, you need `computeIfAbsent` to be able to take a suspend block. 

**Solution**
I added a variant to call it with a suspend block.

### DIFF
--- a/ktor-utils/common/src/io/ktor/util/Attributes.kt
+++ b/ktor-utils/common/src/io/ktor/util/Attributes.kt
@@ -113,6 +113,11 @@ public interface Attributes {
     public fun <T : Any> computeIfAbsent(key: AttributeKey<T>, block: () -> T): T
 
     /**
+     * Gets a value of the attribute for the specified [key], or calls supplied suspend [block] to compute its value
+     */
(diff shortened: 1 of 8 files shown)


e5l:
Hey, @nathanfallet! Thanks for the PR.

It looks like this implementation has a race condition, and the object can be created multiple times because of the suspension. Could you tell me why you need the suspend version of this method?

nathanfallet:
My motivation was to cache the user for the request. I get the user from the database (corresponding to the access token of the request) only the first time, then I reuse the cached result since for this request it's the same user.

e5l:
Could you try storing lazy coroutine instead?
Something like:
```
val value = async(start = CoroutineStart.LAZY) { fetchUser() }
```

And await when you need it

e5l:
I will close this PR. Please reopen if you have any questions

PR #3850, opened 2023-11-27
Add resource URL cache

bjhham:
After playing around with the allocation benchmarks a bit, I found that the `ClassLoader.getResources()` call was the most expensive operation and can be avoided most of the time - resulting in 20-30% improvement by removing ~90 strings allocated from looking through the classpath on each request (which could potentially be much larger on big projects).

The caveat here is that introducing a cache for the resource URLs could create problems for projects that rely on swapping jars at runtime, so I'm thinking it might make sense to introduce this in 3.0.0 with an opt-out option.

I also have a few improvements in `ktor-benchmarks` but I lack the ability to push there atm 🥺

### DIFF
--- a/ktor-server/ktor-server-core/jvm/src/io/ktor/server/http/content/StaticContentResolution.kt
+++ b/ktor-server/ktor-server-core/jvm/src/io/ktor/server/http/content/StaticContentResolution.kt
@@ -42,6 +42,8 @@ public fun ApplicationCall.resolveResource(
     return null
 }
 
+private val resourceCache by lazy { mutableMapOf<String, Pair<URL, OutgoingContent.ReadChannelContent>>() }
+
 @OptIn(InternalAPI::class)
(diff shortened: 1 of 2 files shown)


ktor-server/ktor-server-core/jvm/src/io/ktor/server/http/content/StaticContentResolution.kt
```diff
@@ -42,6 +42,8 @@ public fun ApplicationCall.resolveResource(
     return null
 }
 
+private val resourceCache by lazy { mutableMapOf<String, Pair<URL, OutgoingContent.ReadChannelContent>>() }
```
e5l, line 45: Please add test for fecthing a single resource multiple times

e5l (approved):

e5l:
please also check the CI

PR #3851, opened 2023-11-27
KTOR-6494 Add https to default allowed schemas for CORS

bjhham:
**Subsystem**
CORS (Server)

**Motivation**
[KTOR-6494](https://youtrack.jetbrains.com/issue/KTOR-6494) CORS: `allowHost` without the second argument doesn't allow the secure host

^ This seems like a pretty valid concern, only allowing http by default is quite strange and will inevitably lead to friction when deploying to production.

**Solution**
Simply add https to the default schemas.


rsinukov (approved):

rsinukov:
please check codestyle and api dump

PR #3852, opened 2023-11-28
Release 2.3.7

marychatte:


Stexxe (approved):
Why this PR contains commits to fix the tests and the build?

e5l (approved):

marychatte:
> Why this PR contains commits to fix the tests and the build?

@Stexxe, fixed the build because one PR used the Kotlin version from a 3.0.0 branch, which is different from 2.3.7
And fixed tests because they were falling

PR #3853, opened 2023-11-28
KTOR-4724 Allow double encoding by default for String.encodeURLPath()

bjhham:
**Subsystem**
ktor-http

**Motivation**
[KTOR-4724](https://youtrack.jetbrains.com/issue/KTOR-4724) encodeURLPath skips encoding of % if the input contains %XX sequence

**Solution**
Introduced an optional flag for avoiding double-encoding behaviour, so that the function will encode all % characters as %25 by default.


ktor-server/ktor-server-host-common/api/ktor-server-host-common.api
```diff
@@ -56,6 +56,10 @@ public final class io/ktor/server/engine/ApplicationEngineEnvironmentBuilder {
 	public final fun setWatchPaths (Ljava/util/List;)V
 }
 
+public final class io/ktor/server/engine/ApplicationEngineEnvironmentKt {
+	public static final fun applicationEngineEnvironment (Lkotlin/jvm/functions/Function1;)Lio/ktor/server/engine/ApplicationEngineEnvironment;
+}
+
```
bjhham, line 62: No idea why this has changed...

bjhham: It seems it's still failing in TC, something funky is going on with this module.

e5l (approved):
lgtm, please resolve API check before merging

ktor-server/ktor-server-host-common/api/ktor-server-host-common.api
bjhham: Good news: that gives me what's expected on TC 🥳 
Bad news: there's still a bunch of mysterious changes in that API that have nothing to do with my changes 😕 
Good news: the API dump is the same in main 🎉 
Bad news: something was merged with a big API update that wasn't picked up in the CI 😱

e5l: that's fine, this module was dropped

PR #3854, opened 2023-11-29
Remove assumption "JKS" is always available.

alsutton:
On some platforms (e.g. Android) Java KeyStore (JKS) support is not available. Using the default platform keystore ensures that keys can still be stored on those platforms.

**Subsystem**
Server and Network

**Motivation**
Attempting to use the keystore related code results in an error on Android, switching the assumption that "JKS" is always available to using the platform provided default resolves this.

**Solution**
Change hard coded "JKS" to use of `getDefaultType()` which provides the keystore format available on the system. On Android this is "BKS" (ref: https://cs.android.com/android/platform/superproject/main/+/main:libcore/luni/src/main/java/java/security/security.properties;l=50)


alexstyl (approved):
Seems straightforward :+1:

e5l (approved):

PR #3855, opened 2023-12-04
KTOR-6505 Fix NumberFormatException in HttpCache

marychatte:
[KTOR-6505](https://youtrack.jetbrains.com/issue/KTOR-6505)

### DIFF
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/cache/HttpCacheEntry.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/cache/HttpCacheEntry.kt
@@ -71,8 +71,11 @@ internal fun HttpResponse.cacheExpires(isShared: Boolean, fallback: () -> GMTDat
     val maxAge = cacheControl.firstOrNull { it.value.startsWith(maxAgeKey) }
         ?.value?.split("=")
         ?.get(1)?.let { value ->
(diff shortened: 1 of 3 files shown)


ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/cache/HttpCacheEntry.kt
```diff
@@ -70,7 +70,13 @@ internal fun HttpResponse.cacheExpires(isShared: Boolean, fallback: () -> GMTDat
 
     val maxAge = cacheControl.firstOrNull { it.value.startsWith(maxAgeKey) }
         ?.value?.split("=")
-        ?.get(1)?.toInt()
+        ?.get(1)?.let { value ->
+            if (value.all { it.isDigit() }) {
+                value.toIntOrNull() ?: Int.MAX_VALUE
```
bjhham, line 75: Seems like it would make sense to use Long here instead for one less cast and allow for larger numbers.  Though it's not like anyone would ever notice 😄

bjhham (approved):

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/cache/HttpCacheEntry.kt
marychatte: Yes, thank you! It was my first thought, but I abandoned it by accident :)

PR #3856, opened 2023-12-05
KTOR-6525 Migrate to JUnit5

bjhham:
**Subsystem**
Test Suite

**Motivation**
KTOR-6525 Migrate to JUnit5

**Solution**
- Parameterized tests now require arguments in the method.
- Test execution requires Java 11 now, so `javaLauncher` is set for `jvmTest` task
- CoroutinesTimeout is now an annotation that can be applied at the class or function level
- Used `kotlin.test` where possible
- Many other little quirks


bjhham:
I'll still need to mess around with the timeouts a bit, all the tests work locally though.

ktor-client/ktor-client-cio/jvm/test/io/ktor/client/engine/cio/CIORequestTest.kt
```diff
@@ -55,6 +51,7 @@ class CIORequestTest : TestWithKtor() {
         }
     }
 
+    @CoroutinesTimeout(10_000)
```
e5l, line 54: Could we use class-level annotation instead?
https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-debug/kotlinx.coroutines.debug.junit5/-coroutines-timeout/

ktor-server/ktor-server-test-suites/jvm/src/io/ktor/server/testing/suites/ConfigTestSuite.kt
```diff
@@ -57,6 +56,8 @@ abstract class ConfigTestSuite(val engine: ApplicationEngineFactory<*, *>) {
             server.stop(0, 100, TimeUnit.SECONDS)
         }
 
-        assertTrue("Stop time is $time", time < 100)
+        assert(time < 100) {
```
e5l, line 59: Could you tell me why raw assert instead of assertTrue is used?

ktor-server/ktor-server-test-host/jvm/src/io/ktor/server/testing/BaseTestJvm.kt
```diff
@@ -5,31 +5,42 @@
...
+import kotlinx.coroutines.debug.junit5.*
+import org.junit.jupiter.api.*
+import java.lang.reflect.*
+import java.util.*
 import kotlin.time.*
 import kotlin.time.Duration.Companion.seconds
 
+@CoroutinesTimeout(5 * 60 * 1000)
+@ErrorCollectorTest
 actual abstract class BaseTest actual constructor() {
```
e5l, line 19: Should we keep `ktor-server-test-host` using junit4 and make a separate module for junit5? Let's discuss

ktor-server/ktor-server-jetty-jakarta/build.gradle.kts
```diff
@@ -4,6 +4,7 @@ kotlin {
     sourceSets {
         jvmMain {
             dependencies {
+                api(kotlin("test"))
```
e5l, line 7: could you tell me why do we need test dependency in the main module?

ktor-server/ktor-server-plugins/ktor-server-metrics/jvm/test/io/ktor/server/metrics/dropwizard/DropwizardMetricsTests.kt
```diff
@@ -96,7 +95,9 @@ class DropwizardMetricsTests {
 
         handleRequest { uri = "/uri" }
 
-        assertThat(registry.names, everyItem(startsWith(prefix)))
+        assertThat(registry.names).each {
```
e5l, line 98: we could use `assertAll` from junit 5 or make our own extension

e5l (changes requested):
Thanks for the PR, it's mostly good. Please check the dependencies and some minor stuff.

ktor-shared/ktor-junit/build.gradle.kts
```diff
@@ -0,0 +1,14 @@
+/*
+ * Copyright 2014-2023 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+description = "Common extensions for JUnit 5 testing framework"
```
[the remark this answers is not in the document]
bjhham: Shouldn't be, I just missed the configuration for avoiding the publish.

How is that done, is it here in `build.gradle.kts`?

```kotlin
extra["skipPublish"] = mutableListOf<String>()
```

ktor-server/ktor-server-test-suites/jvm/src/io/ktor/server/testing/suites/ConfigTestSuite.kt
bjhham: Oh I meant to replace with `kotlin.test` which just has the arguments swapped, I'll fix.

ktor-server/ktor-server-jetty-jakarta/build.gradle.kts
bjhham: Good catch, I must have got my build scripts mixed up.

ktor-server/ktor-server-test-host/jvm/src/io/ktor/server/testing/BaseTestJvm.kt
bjhham: The problem here is that it's used in other projects, so we'd have a swath of tests using junit 4.  The only big difference here is that we're going from 60 seconds timeout for each test to 5 minutes total because of how the annotations work.  You can override the annotation on the child classes and methods if another timeout is needed, though it doesn't seem to be a problem atm...

ktor-client/ktor-client-cio/jvm/test/io/ktor/client/engine/cio/CIORequestTest.kt
bjhham: Yeah I originally had moved it to the class annotation but it was timing out because it applies for the execution of the whole collection of tests.  I can just sum them up and use the one annotation.

ktor-server/ktor-server-test-host/jvm/src/io/ktor/server/testing/BaseTestJvm.kt
e5l: lgtm

ktor-shared/ktor-junit/build.gradle.kts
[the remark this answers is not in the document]
bjhham: We've got some other internal test projects that aren't included there...  Is the idea that external developers can use those projects to test new engines etc?

e5l (approved):

e5l:
Please wait for CI builds before merging

ktor-shared/ktor-junit/build.gradle.kts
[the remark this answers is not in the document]
e5l: Exactly

rsinukov (approved):

PR #3860, opened 2023-12-16
JS: Fix origin retrieval on platforms that have window, but no location

baconz:
**Subsystem**
Client/HTTP (JS)

**Motivation**
On React Native platforms there is a window object, no node environment, and no location, so URLBuilder was crashing hard.

**Solution**
This improves the safety of the code that retrieves the origin.

https://youtrack.jetbrains.com/issue/KTOR-6576/URLBuilder-crashes-on-React-Native-platforms

### DIFF
--- a/ktor-http/js/src/io/ktor/http/URLBuilderJs.kt
+++ b/ktor-http/js/src/io/ktor/http/URLBuilderJs.kt
@@ -10,17 +10,23 @@ import io.ktor.util.*
  * Hostname of current origin.
  *
  * It uses "localhost" for all platforms except js.
+ *
+ * Note that not all platforms have location set. React Native platofrms expose window without a location.
  */
 public actual val URLBuilder.Companion.origin: String
     get() = when (PlatformUtils.platform) {
         Platform.Browser -> {
             js(
                 """
-                var origin = ""
+                var tmpLocation = null
                 if (typeof window !== 'undefined') {
-                  origin = window.location.origin 
+                  tmpLocation = window.location
                 } else {
-                  origin = self.location.origin 
+                  tmpLocation = self.location
+                }
+                var origin = ""
+                if (tmpLocation) {
+                  origin = location.origin
                 }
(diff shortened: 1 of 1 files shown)


ktor-http/js/src/io/ktor/http/URLBuilderJs.kt
```diff
@@ -10,17 +10,23 @@ import io.ktor.util.*
...
                 if (typeof window !== 'undefined') {
-                  origin = window.location.origin 
+                  tmpLocation = window.location
                 } else {
-                  origin = self.location.origin 
+                  tmpLocation = self.location
+                }
+                var origin = ""
+                if (tmpLocation) {
+                  origin = location.origin
```
rsinukov, line 29: Should it be `tmpLocation` instead of `location` here?  If yes, why call it `tmpLocation`?

rsinukov (changes requested):
Thanks for the PR. I have one question though.

ktor-http/js/src/io/ktor/http/URLBuilderJs.kt
rsinukov: Do you also mind checking that `self` is defined above? This will fix https://youtrack.jetbrains.com/issue/KTOR-5978

baconz: > Should it be tmpLocation instead of location here?

Yes, indeed it should. `location` worked in my testing because `self` and `window` both represent global scope, so you can reference their members directly.

> If yes, why call it `tmpLocation`?

This tripped me up! At first I tried with code like:

```
                if (typeof window !== 'undefined') {
                  location = window.location
                } else {
                  location = self.location
                }
```

But in the browser, when you set `location = window.location` it is cute little shortcut to reload the page!!!

> Do you also mind checking that self is defined above? 

Done!

rsinukov (approved):
LGTM

PR #3861, opened 2023-12-18
KTOR-6175 Change default session serializer to the one that uses kotlinx-serialization

rsinukov:
https://youtrack.jetbrains.com/issue/KTOR-6175


bjhham (approved):

bjhham:
Looks like CI is not so happy... Classes not marked with `@Serializable`.

PR #3864, opened 2023-12-21
KTOR-5181 `RoutingContext` should not be `CoroutineScope`

rsinukov:


e5l (changes requested):
Could you please check the CI?

e5l:
Hey @bjhham, could you please take a look? I will take this to fix while @rsinukov is on vacation

ktor-client/ktor-client-plugins/ktor-client-call-id/jvmAndNix/test/io/ktor/client/plugins/callid/CallIdTest.kt
```diff
@@ -12,6 +12,7 @@ import io.ktor.server.plugins.callid.*
 import io.ktor.server.response.*
 import io.ktor.server.routing.*
 import io.ktor.server.testing.*
+import kotlin.coroutines.*
```
bjhham, line 15: Looks like an unused import here.

ktor-server/ktor-server-tests/jvmAndNix/test/io/ktor/tests/server/plugins/CallIdTest.kt
```diff
@@ -15,6 +15,7 @@ import io.ktor.server.response.*
 import io.ktor.server.routing.*
 import io.ktor.server.testing.*
 import io.ktor.util.pipeline.*
+import kotlin.coroutines.*
```
bjhham, line 18: Possibly here too.

ktor-client/ktor-client-plugins/ktor-client-call-id/jvmAndNix/test/io/ktor/client/plugins/callid/CallIdTest.kt
e5l: it use for accessing `coroutineContext` of the suspend function

bjhham: Ah I see it was a compile error in CI before 💡

bjhham (approved):

bjhham:
Looks like there's a problem with how my retry test magic works, I'll look into it 😞 

And it seems like the allocation tests fail sometimes, might need more warmup or something... 🤔

e5l (approved):

PR #3868, opened 2024-01-01
Add warning to docs to specify all optional parameters on some engines

Mats-Hjalmar:
**Subsystem**
Client, WebSockets

**Motivation**
I was stuck on making WebSockets work for iOS using `Darwin` engine only to find out these optional fields were required.

**Solution**
Add a warning to the docs to help developers not get stuck where i did.


e5l:
@Mats-Hjalmar, could you log an issue? I think it should be fixed as a bug

PR #3870, opened 2024-01-02
Kotlin K2 Migration: Override abstract members explicitly on expect s…

tbogdanova:
…ide KT-59739 KT-59887

**Subsystem**
None.

**Motivation**
The Kotlin team checks ktor with the new versions of the K2 compiler. Found problems are investigated and if the problem is not planned to be fixed, migration is added. Branch: `kotlin-community/k2/dev`. To make maintaining the branch easier, we suggest moving migrations to the main branch. 

[KT-59739 ](https://youtrack.jetbrains.com/issue/KT-59739)

**Solution**
In K2, it's required to override abstract members explicitly on expect side.


e5l (approved):
lgtm

PR #3875, opened 2024-01-04
Increase timeout for testEmbeddedServerCancellation and update failure

bjhham:
**Subsystem**
Test suite

**Motivation**
Fix for a test that fails sometimes due to a timeout on waiting for server to shut down.

**Solution**
Simply increase the timeout - since the server startup time can take quite long on the TC server due to competing threads, this ought to resolve the periodic timeouts.


e5l (approved):

e5l:
Thanks for the PR! LGTM

PR #3882, opened 2024-01-08
KTOR-6562 Add plugin manifest for CSRF plugin

bjhham:
**Subsystem**
Server, CSRF plugin

**Motivation**
Plugin is not available in project generator.

**Solution**
This is an example of how to include your plugin meta data for the registry builder project.


e5l:
Hey @bjhham, thanks for the PR!

Could you tell me if we can store plugin manifests in the plugin registry repo? It may be a problem to make PRs for all plugin repos with a breaking change or Ktor update.

e5l:
Could you tell me if it's still required if we move this manifest to the plugin registry?

bjhham:
Nope, we can resolve everything from the registry repo.  I'll close this.

PR #3884, opened 2024-01-09
Add retry extension for junit tests to fix flaky tests

bjhham:
Added junit extension to retry tests a single time when they hit an exception, hopefully this will improve the random failures 🤞


e5l (approved):
Hey @bjhham, good job! lgtm

PR #3895, opened 2024-01-12
Retrying my retry logic to handle flaky connections in tests

bjhham:
**Subsystem**
Testing

**Motivation**
My previous attempt for invoking the tests a second time had some bad logic.

**Solution**
Now it properly uses JUnit5 extensions to re-run the test multiple times as required, and you can see the results of each attempt.  There's also another annotation for more granular control (using @RetryableTest instead of @Test).


e5l (approved):

e5l:
LGTM

PR #3911, opened 2024-01-15
KTOR-6574 Fix duplicate headers in ServletApplicationEngine with Tomcat

marychatte:
**Subsystem**
Server: ktor-server-servlet, ktor-server-servlet-jakarta

**Motivation**
[KTOR-6574](https://youtrack.jetbrains.com/issue/KTOR-6574/Double-Transfer-Encoding-header-while-using-respondOutputStream-on-Tomcat)
If Ktor project is running as a [WAR](https://ktor.io/docs/war.html) application under the Tomcat application server, a user has duplicate headers `Transfer-Encoding` and `Connection` in response.

**Solution**
The problem is on the `Tomcat` side or `Gretty` + `Tomcat` side because the user doesn't use `Tomcat` from Ktor, only `ServletApplicationEngine` is used. The solution is to configure the `managedByEngineHeaders` property in `ServletApplicationEngine`
Also, I gave in to write test since it's reproducible with gretty and it was hard to run test as WAR under the Tomcat but I tested it locally by publish to maven local and run in side project using gretty plugin.

### DIFF
--- a/ktor-server/ktor-server-servlet-jakarta/jvm/src/io/ktor/server/servlet/jakarta/ServletApplicationEngine.kt
+++ b/ktor-server/ktor-server-servlet-jakarta/jvm/src/io/ktor/server/servlet/jakarta/ServletApplicationEngine.kt
@@ -5,6 +5,7 @@
 package io.ktor.server.servlet.jakarta
 
 import io.ktor.events.*
+import io.ktor.http.*
 import io.ktor.server.application.*
 import io.ktor.server.config.*
 import io.ktor.server.config.ConfigLoader.Companion.load
@@ -19,6 +20,20 @@ import kotlin.coroutines.*
  */
 @MultipartConfig
 public open class ServletApplicationEngine : KtorServlet() {
+
+    override val managedByEngineHeaders: Set<String>
+        get() {
+            servletContext.getAttribute(ApplicationAttributeKey)?.let {
+                return emptySet()
+            }
+
+            val servletContext = servletContext
+            return if ("tomcat" in (servletContext.serverInfo?.toLowerCasePreservingASCIIRules() ?: "")) {
+                setOf(HttpHeaders.TransferEncoding, HttpHeaders.Connection)
+            } else {
+                emptySet()
+            }
+        }
 
     private val embeddedServer: EmbeddedServer<ApplicationEngine, ApplicationEngine.Configuration>? by lazy {
         servletContext.getAttribute(ApplicationAttributeKey)?.let {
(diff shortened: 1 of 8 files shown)


ktor-server/ktor-server-servlet-jakarta/jvm/src/io/ktor/server/servlet/jakarta/ServletApplicationEngine.kt
```diff
@@ -20,6 +21,20 @@ import kotlin.coroutines.*
...
+                return emptySet()
+            }
+
+            val servletContext = servletContext
+            return if ("tomcat" in (servletContext.serverInfo?.toLowerCasePreservingASCIIRules() ?: "")) {
+                setOf(HttpHeaders.TransferEncoding, HttpHeaders.Connection)
+            } else {
+                emptySet()
+            }
+        }
```
bjhham, line 36: How about something like...

```kotlin
// Some shared location
fun ServletContext.isTomcat() =
    getAttribute(ApplicationAttributeKey) == null && serverInfo.contains("tomcat", ignoreCase = true)

if (servletContext.isTomcat())
    setOf(HttpHeaders.TransferEncoding, HttpHeaders.Connection)
else emptySet()
```

ktor-server/ktor-server-servlet-jakarta/jvm/test/io/ktor/tests/servlet/jakarta/ConfigTest.kt
```diff
@@ -36,6 +36,7 @@ class ConfigTest {
             every { contextPath } returns "/"
             every { getAttribute(ApplicationAttributeKey) } returns null
             every { getAttribute(EnvironmentAttributeKey) } returns null
+            every { serverInfo } returns ""
```
bjhham, line 39: Looks like this line is not required

ktor-server/ktor-server-servlet-jakarta/jvm/src/io/ktor/server/servlet/jakarta/ServletApplicationEngine.kt
marychatte: Thank you! It looks much easier :) I'm not sure It can be a shared location because `ktor-server-servlet` and `ktor-server-servlet-jakarta` have different `ServletContext` (from packages `javax.servlet` and `jakarta.servlet`)

bjhham: Ah unfortunate, copy pasta it is 😞

ktor-server/ktor-server-servlet-jakarta/jvm/test/io/ktor/tests/servlet/jakarta/ConfigTest.kt
bjhham: Ah cool I understand 👍

bjhham (approved):

PR #3929, opened 2024-01-17
KTOR-6657 Misleading assertion message in the testEmbeddedServerCance…

bjhham:
…llation test

Also added retry so it should fail much less often.

### DIFF
--- a/ktor-server/ktor-server-test-suites/jvm/src/io/ktor/server/testing/suites/SustainabilityTestSuite.kt
+++ b/ktor-server/ktor-server-test-suites/jvm/src/io/ktor/server/testing/suites/SustainabilityTestSuite.kt
@@ -254,6 +254,7 @@ abstract class SustainabilityTestSuite<TEngine : ApplicationEngine, TConfigurati
         assertTrue(job!!.isCancelled)
     }
 
+    @RetryableTest(2)
     @OptIn(ExperimentalCoroutinesApi::class)
     @Test
     fun testEmbeddedServerCancellation() {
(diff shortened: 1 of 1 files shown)


ktor-server/ktor-server-test-suites/jvm/src/io/ktor/server/testing/suites/SustainabilityTestSuite.kt
```diff
@@ -277,7 +278,7 @@ abstract class SustainabilityTestSuite<TEngine : ApplicationEngine, TConfigurati
                 }
             } catch (cause: TimeoutCancellationException) {
                 DebugProbes.printJob(parent)
-                fail("Server did shut down in time after cancelling parent!")
+                fail("Server did shut down in time!")
```
Stexxe, line 281: So the test passes when the server shutdown is timed out?

bjhham: No it will continue to fail when it doesn't shut down in time.  I just updated the message and added a retry so we won't get random failures on TC so often.

Stexxe: So should the failed message reflect that, e.g. ` fail("Server did NOT shut down in time!")`?

bjhham: Oh yes, I missed that 😆

Stexxe (approved):

PR #3930, opened 2024-01-17
KTOR-3290 Add SSE support for test engine

marychatte:
**Subsystem**
Server sse

**Motivation**
We have an old ticket [KTOR-3290](https://youtrack.jetbrains.com/issue/KTOR-3290) with [PR](https://github.com/ktorio/ktor/pull/2659). However, it became outdated because we have SSE support now, so we can just add a support for test to engine to have simple way to test


bjhham (approved):
Nice, that client code is so simple ❤️

PR #3931, opened 2024-01-17
KTOR-6660 Add call property to ClientSSESession

marychatte:
**Subsystem**
Client, sse

**Motivation**
[KTOR-6660](https://youtrack.jetbrains.com/issue/KTOR-6660)

**Solution**
By analogy with `Websockets`


bjhham (approved):
Makes sense 👍

PR #3946, opened 2024-01-19
Fix test-host gradle problem + add API validation

bjhham:
**Subsystem**
Server, Test Application

**Motivation**
[KTOR-6668](https://youtrack.jetbrains.com/issue/KTOR-6668) Ktor test application library references internal module

There's actually a few problems:
1. I added an internal lib, `ktor-junit`, as an api dependency in `ktor-server-test-host`, making it impossible to import externally
2. There are some internal classes being exposed in the `ktor-server-test-host`, like `BaseTest`.
3. We currently don't validate the API compatibility on this module, despite it being public, so we could accidentally introduce breaking changes.

**Solution**
1. Moved the `BaseTest` and its friends over to a new internal module `ktor-server-test-base` which is now referenced only from our test classpaths.
2. Removed `ktor-server-test-host` from the API test exemptions, so we'll have to keep it consistent.


settings.gradle.kts
```diff
@@ -16,6 +16,9 @@ pluginManagement {
         }
     }
 }
+plugins {
+    id("org.gradle.toolchains.foojay-resolver-convention") version "0.5.0"
+}
```
bjhham, line 21: Intellij trying to be helpful, will remove.

e5l (approved):
Awesome! LGTM

PR #3948, opened 2024-01-19
add buildUrl

MarcusDunn:
like `buildString` but for URL's

**Subsystem**
http

**Motivation**
I dislike calling 
```kotlin
UrlBuilder().apply {
    // ...
}.build()
```
and yet seem to do it every-time I construct a URL.

**Solution**
```kotlin
public fun buildUrl(block: URLBuilder.() -> Unit): Url = URLBuilder().apply(block).build()
```

**Some examples**

with no `url`
```kotlin
fun issuer(region: String, userPoolId: String) = buildUrl {
    protocol = URLProtocol.HTTPS
    host = "cognito-idp.$region.amazonaws.com"
    appendPathSegments(userPoolId)
}
```

with `url`
```kotlin 
fun domain(region: String, userPoolId: String): Url = buildUrl {
    takeFrom(issuer(region, userPoolId))
    appendPathSegments(".well-known", ".jwks.json")
}
```

### DIFF
--- a/ktor-http/api/ktor-http.api
+++ b/ktor-http/api/ktor-http.api
@@ -992,6 +992,7 @@ public final class io/ktor/http/URLUtilsKt {
 	public static final fun Url (Lio/ktor/http/URLBuilder;)Lio/ktor/http/Url;
 	public static final fun Url (Ljava/lang/String;)Lio/ktor/http/Url;
 	public static final fun appendUrlFullPath (Ljava/lang/Appendable;Ljava/lang/String;Lio/ktor/http/ParametersBuilder;Z)V
+	public static final fun buildUrl (Lkotlin/jvm/functions/Function1;)Lio/ktor/http/Url;
 	public static final fun getFullPath (Lio/ktor/http/Url;)Ljava/lang/String;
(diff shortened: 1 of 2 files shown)


e5l (approved):
Hey @MarcusDunn, thanks for the PR. Could you please run `./gradlew apiDump` and commit changes?

morki:
@e5l @MarcusDunn Isn't this just duplication of `url` function builder already present in Ktor?

```kotlin
package io.ktor.server.util

/**
 * Construct a URL
 */
public fun url(block: URLBuilder.() -> Unit): String = URLBuilder().apply(block).buildString()

/**
 * Creates an url using current call's schema, path and parameters as initial
 * and then invokes [block] function on the url builder so amend parameters
 */
public inline fun ApplicationCall.url(block: URLBuilder.() -> Unit = {}): String =
    URLBuilder.createFromCall(this).apply(block).buildString()

```

MarcusDunn:
I've have it in my project because I perfer the `Url` type `String` and the `Url(url {})` call is also awkward - there's absolutly a case to be made `buildUrl` deserves to remain as free function in my project instead of included directly in ktor though.

e5l:
Hey @MarcusDunn, thanks for the fast response! Merged

PR #3951, opened 2024-01-23
KTOR-6659 Send body in SSE request

marychatte:
**Subsystem**
Client, sse

**Motivation**
[KTOR-6659](https://youtrack.jetbrains.com/issue/KTOR-6659/SSE-request-drops-request-body)

**Solution**
Allow to send request body in SSE requests


ktor-client/ktor-client-android/jvm/src/io/ktor/client/engine/android/AndroidClientEngine.kt
```diff
@@ -62,7 +62,7 @@ public class AndroidClientEngine(override val config: AndroidEngineConfig) : Htt
             config.requestConfig(this)
 
             if (data.method in METHODS_WITHOUT_BODY) {
-                if (outgoingContent is OutgoingContent.NoContent) {
+                if (outgoingContent is OutgoingContent.NoContent || data.isSseRequestWithoutBody()) {
```
bjhham, line 65: Seems like an abstraction reversal here with the SSE checks... It might make sense to introduce a `OutgoingContent.ContentWrapper` type that just delegates to an `OutgoingContent` property, then add an `isEmpty()` function to `OutgoingContent` that returns false for everything except `NoContent` and the new wrapper type which delegates.  Then this would just be `if (outgoingContent.isEmpty())`.

ktor-client/ktor-client-android/jvm/src/io/ktor/client/engine/android/AndroidClientEngine.kt
```diff
@@ -95,7 +95,7 @@ public class AndroidClientEngine(override val config: AndroidEngineConfig) : Htt
 
             val responseBody: Any = if (needToProcessSSE(data, statusCode, responseHeaders)) {
                 DefaultClientSSESession(
-                    data.body as SSEClientContent,
+                    outgoingContent as SSEClientContent,
                     content,
                     callContext
                 )
```
bjhham, line 101: This `needsToProcessSSE` could also be abstracted to an interface like...

```kotlin
interface ResponseAdapter {
    fun adapt(status: HttpStatusCode, headers: Headers, body: Any): Any?
}
```

Then do something like `responseBody: Any = data.attributes.getOrNull(ResponseAdapter)?.adapt(statusCode, headers, body) ?: content`.

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/plugins/ServerSentEventsTest.kt
```diff
@@ -309,4 +310,24 @@ class ServerSentEventsTest : ClientLoader(timeoutSeconds = 120) {
...
+            client.sse({
+                url("$TEST_SERVER/sse/echo")
+                setBody(body)
+                contentType(contentType)
+            }) {
+                assertEquals(contentType, call.request.contentType()?.withoutParameters())
+                assertEquals(body, incoming.single().data)
+            }
+        }
+    }
```
bjhham, line 332: Would be nice to include a test to trigger the protocol upgrade error and maybe one for some streaming content.

bjhham (approved):
Optional comments if you're in the mood for refactoring.

ktor-client/ktor-client-android/jvm/src/io/ktor/client/engine/android/AndroidClientEngine.kt
marychatte: Thanks a lot for the refactoring suggestions! I think it makes sense to introduce `OutgoingContent.ContentWrapper`. I wasn't really sure about `ResponseAdapter` in the comment lower because it looks like `SseResponseAdapter` is the only usage of it, but please correct me if I am wrong

bjhham: Oh yeah, it's more about providing an extension point for future similar plugins - seems to be a common enough pattern anyway.

bjhham (approved):
👍 👍

e5l (changes requested):
please check the CI. it looks like js tests are failing with sse

ktor-client/ktor-client-mock/common/src/io/ktor/client/engine/mock/MockUtils.kt
```diff
@@ -27,8 +28,7 @@ public suspend fun OutgoingContent.toByteArray(): ByteArray = when (this) {
         }
         channel.toByteArray()
     }
-
-    else -> ByteArray(0)
+    is OutgoingContent.ProtocolUpgrade, is OutgoingContent.NoContent -> ByteArray(0)
```
e5l, line 31: please consider extracting a constant

e5l (approved):
good job, please check small comments

PR #3952, opened 2024-01-23
Update TLS.kt with docs about coroutineContext parameter

rocketraman:
**Subsystem**
Client/Server, network-tls

**Motivation**
Update docs based on discussion in Slack: https://kotlinlang.slack.com/archives/C0A974TJ9/p1705535602239209.

**Solution**
Add docs explaining that the `coroutineContext` parameter to `tls()` should generally not be a child of the enclosing scope.


e5l (approved):
Hey @rocketraman! 🚀 Thanks for the PR. LGTM

PR #3963, opened 2024-01-25
Drop jvmOnly flag and enable configuration caching

e5l:
At the moment, `commonAndJvmOnly` flag breaks build caching between CLI and IDE, which leads to inconsistent builds.

I dropped this flag and fix several issues preventing configuration caching, so it should be better during IDE import


bjhham (approved):
All checks green, beautiful 🥲

PR #3966, opened 2024-01-26
KTOR-6687 Empty CSRF onFailure option is misleading

bjhham:
**Subsystem**
Server, CSRF

**Motivation**
- KTOR-6687 onFailure without an explicit response would result in a 200 empty body

**Solution**
- Added the default error response when there is no response committed in the onFailure block

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-csrf/jvmAndNix/src/io/ktor/server/plugins/csrf/CSRFConfig.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-csrf/jvmAndNix/src/io/ktor/server/plugins/csrf/CSRFConfig.kt
@@ -17,9 +17,7 @@ public class CSRFConfig {
     internal var originMatchesHost = false
     internal val allowedOrigins = mutableListOf<Url>()
     internal val headerChecks = mutableMapOf<String, HeaderPredicate>()
-    internal var handleFailure: suspend ApplicationCall.(String) -> Unit = { message ->
-        respond(HttpStatusCode.BadRequest, "Cross-site request validation failed; $message")
-    }
+    internal var handleFailure = respondBadRequestIfNotCommitted
 
     /**
      * All incoming requests must have an "Origin" header matching one of the hosts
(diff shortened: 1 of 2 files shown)


ktor-server/ktor-server-plugins/ktor-server-csrf/jvmAndNix/src/io/ktor/server/plugins/csrf/CSRF.kt
```diff
@@ -38,6 +38,11 @@ public val CSRF: RouteScopedPlugin<CSRFConfig> = createRouteScopedPlugin("CSRF",
     val headerChecks = pluginConfig.headerChecks
     val onFailure = pluginConfig.handleFailure
 
+    if (!checkHost && allowedOrigins.isEmpty() && headerChecks.isEmpty()) {
```
Stexxe, line 41: Can you tell me how this code block is related to the onFailure problem?

bjhham: This is the second bullet in the description, to address that an empty configuration doesn't provide any functionality.  There's no real valid option for the default behaviour in this plugin so I'm proposing that we could log a warning when there's no validation defined.

Stexxe: Every commit should have a prefix with a YouTrack issue ID. I don't see the ID for the second commit.

Stexxe: Also, having a separate PR for an issue is a good practice.

bjhham: Sure, I'll move it to another 👍

e5l (approved):

PR #3968, opened 2024-01-29
KTOR-6695 Add configuration warning when nothing is provided for CSRF

bjhham:
**Subsystem**
Server, CSRF

**Motivation**
- [KTOR-6695](https://youtrack.jetbrains.com/issue/KTOR-6695) CSRF: The allowOrigin method enables the Origin Header validation

**Solution**
After thinking about how to provide some default behaviour for the CSRF plugin, I am of the opinion that it would be unwise to select one of the mitigations automatically for the user.  Instead, we ought to report an error for when it is an empty configuration.  Personally, I'd rather throw an exception, but logging a warning should be enough.

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-csrf/jvmAndNix/src/io/ktor/server/plugins/csrf/CSRF.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-csrf/jvmAndNix/src/io/ktor/server/plugins/csrf/CSRF.kt
@@ -38,6 +38,11 @@ public val CSRF: RouteScopedPlugin<CSRFConfig> = createRouteScopedPlugin("CSRF",
     val headerChecks = pluginConfig.headerChecks
     val onFailure = pluginConfig.handleFailure
 
+    if (!checkHost && allowedOrigins.isEmpty() && headerChecks.isEmpty()) {
+        application.log.warn("No validation options provided for CSRF plugin - requests will not be verified!")
+        return@createRouteScopedPlugin
+    }
+
     onCall { call ->
 
         if (call.request.httpMethod in setOf(HttpMethod.Get, HttpMethod.Head, HttpMethod.Options)) {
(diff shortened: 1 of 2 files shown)


Stexxe (commented):
Can you explain how the fix is related to the https://youtrack.jetbrains.com/issue/KTOR-6695?

bjhham:
Yes, my assertion is that the default behaviour not enabling any validation is important, but we ought to inform the developer that it's not a valid configuration for CSRF.

As for the phrasing of "allow", I don't have any ideas atm.  The current choice was just to maintain some similarity with CORS, but I understand the behaviour is different.

Stexxe:
> Yes, my assertion is that the default behaviour not enabling any validation is important, but we ought to inform the developer that it's not a valid configuration for CSRF.
> 
> As for the phrasing of "allow", I don't have any ideas atm. The current choice was just to maintain some similarity with CORS, but I understand the behaviour is different.

It might be important, but I don't see how the current changes fix the problem from the linked YouTrack issue.

bjhham:
From what I understood, this issue is twofold: 1) nothing happens with the empty config, and 2) `allow` phrasing implies that something does happen, i.e. there is an implicit empty whitelist before calling this function.  This is to address the first item.  I'm open for suggestions on item 2.

bjhham:
After discussing with @e5l, I've updated the default behaviour for this to block any non-local origins - so when the configuration is empty, you'll only be able to connect to localhost, 127.0.0.1, and 0.0.0.0.  There's an informative log message when this happens.

ktor-server/ktor-server-plugins/ktor-server-csrf/jvmAndNix/src/io/ktor/server/plugins/csrf/CSRF.kt
```diff
@@ -38,6 +38,19 @@ public val CSRF: RouteScopedPlugin<CSRFConfig> = createRouteScopedPlugin("CSRF",
     val headerChecks = pluginConfig.headerChecks
     val onFailure = pluginConfig.handleFailure
 
+    if (!checkHost && allowedOrigins.isEmpty() && headerChecks.isEmpty()) {
```
Stexxe, line 41: Just to clarify, are the local host origins always allowed?

Stexxe (approved):

ktor-server/ktor-server-plugins/ktor-server-csrf/jvmAndNix/src/io/ktor/server/plugins/csrf/CSRF.kt
bjhham: No, when you provide some configuration it's exactly what you provide.  I don't think allowing them all the time would expose any vulnerabilities, unless the attacker manages to get a server running on the victim's local machine or something, though it's probably safer to leave it like this.

PR #3971, opened 2024-01-29
KTOR-6666 Fix Ktor build problem; cannot compile native targets

bjhham:
**Subsystem**
Server, build

**Motivation**
- [KTOR-6666](https://youtrack.jetbrains.com/issue/KTOR-6666)

**Solution**
Looks like the serialization lib got updated again, this will fix the maven publishing / allocation tests.  I'll try to find a solution that allows us to upgrade this.


bjhham:
Looks like 1.6.2 serialization supports 1.9.21, and they must have dropped support for earlier versions.  I'm not sure exactly what we'd need to update our kotlin version to, but it would involve a kotlin version update.

e5l:
Thanks for noticing! lgtm

PR #3972, opened 2024-01-29
Release 2.3.8

e5l:


ktor-client/ktor-client-cio/jvm/test/io/ktor/client/engine/cio/CIOSpecificHttpsTest.kt
```diff
@@ -143,4 +146,33 @@ class CIOSpecificHttpsTest : TestWithKtor() {
...
+                                return emptyArray()
+                            }
+                        }
+                    }
+                }
+            }
+            test { client ->
+                assertEquals(
+                    HttpStatusCode.MethodNotAllowed,
+                    client.get("https://pt.api.sandbox.npay.eu/token/Grant").status
```
Stexxe, line 173: Should we have an external URL in our tests?

Stexxe (approved):

gradle/libs.versions.toml
```diff
@@ -1,52 +1,52 @@
 [versions]
 kotlin-version = "1.8.22"
-kotlinx-html-version = "0.9.1"
+kotlinx-html-version = "0.11.0"
 coroutines-version = "1.7.1"
 atomicfu-version = "0.19.0"
-serialization-version = "1.5.1"
+serialization-version = "1.6.2"
```
bjhham, line 6: I do believe this will cause problems for publishing the artifacts due to a weird native compilation problem.

https://github.com/ktorio/ktor/pull/3971

bjhham:
Some weirdness going on with the CI... looks like a Java 8/11 problem with logback specifically for the OpenAPITest...? 

```
java.lang.UnsupportedClassVersionError: ch/qos/logback/classic/spi/LogbackServiceProvider has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
```

Then also some JDK problem in NettyConfigurationTest `kotlin.reflect.jvm.internal.KotlinReflectionInternalError: Unresolved class: class java.util.concurrent.TimeUnit$3`

gradle/libs.versions.toml
```diff
@@ -1,52 +1,52 @@
 [versions]
 kotlin-version = "1.8.22"
-kotlinx-html-version = "0.9.1"
+kotlinx-html-version = "0.11.0"
```
bjhham, line 3: Seems this may also be causing some problems with building common artifacts...

```
java.lang.IllegalStateException: e: Could not find "/mnt/agent/work/a5ccb8c76251a2e/ktor/ktor-server/ktor-server-plugins/ktor-server-html-builder/build/kotlinTransformedMetadataLibraries/jvmAndNixMain/org.jetbrains.kotlinx-kotlinx-html-0.11.0-commonMain-W4uI5A.klib" in [/home/teamcity/.local/share/kotlin/daemon]
```

e5l: yep

e5l:
> Some weirdness going on with the CI... looks like a Java 8/11 problem with logback specifically for the OpenAPITest...?
> 
> ```
> java.lang.UnsupportedClassVersionError: ch/qos/logback/classic/spi/LogbackServiceProvider has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
> ```
> 
> Then also some JDK problem in NettyConfigurationTest `kotlin.reflect.jvm.internal.KotlinReflectionInternalError: Unresolved class: class java.util.concurrent.TimeUnit$3`

it looks like we need to revert logback update

ktor-client/ktor-client-cio/jvm/test/io/ktor/client/engine/cio/CIOSpecificHttpsTest.kt
jkauk-n: Hello, I work at the company responsible for this endpoint. I noticed you added this link in your tests and wondered about the same thing. For now, this URL is stable, but I can't promise it will be so in the future.

e5l: Hey, @Stexxe, @jkauk-n, thanks for the report. We will drop the URL asap

jkauk-n: All right, thanks. This is a sandbox URL, so there's no harm in using it for now or blocking your release because of it.

bjhham (approved):
The allocation tests are failing to compile correctly now! 🙃

PR #3973, opened 2024-01-29
KTOR-6614 Fix SSE client breaking non-SSE requests

bjhham:
**Subsystem**
Client, SSE

**Motivation**
- [KTOR-6614](https://youtrack.jetbrains.com/issue/KTOR-6614) Installing SSE breaks GET/POST functionality

**Solution**
Added an attribute flag to enable SSE transform; previously it would apply to all requests.

### DIFF
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/sse/SSE.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/sse/SSE.kt
@@ -49,6 +49,10 @@ public val SSE: ClientPlugin<SSEConfig> = createClientPlugin(
         LOGGER.trace("Sending SSE request ${request.url}")
         request.setCapability(SSECapability, Unit)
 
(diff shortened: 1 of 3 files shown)


ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/sse/SSE.kt
```diff
@@ -49,6 +49,10 @@ public val SSE: ClientPlugin<SSEConfig> = createClientPlugin(
         LOGGER.trace("Sending SSE request ${request.url}")
         request.setCapability(SSECapability, Unit)
 
+        if (getAttributeValue(request, sseRequestAttr) != true) {
```
marychatte, line 52: Can we add this check before `LOGGER.trace("Sending SSE request ${request.url}")`?

marychatte (approved):
LGTM

PR #3979, opened 2024-02-04
Add better description of `TLSConfigBuilder.addCertificateChain()`

aueam:
**Subsystem**
Client (tls, two-way authentication)

**Motivation**
It is not clear when the client will use the certificate. This description could explain to others why their client does not send the certificate to the server.

**Solution**
Just add better description of it.


e5l (approved):
Hey @aueam, thanks for the description. LGTM

PR #3981, opened 2024-02-07
Supports static linking for curl engine

dtretyakov:
**Subsystem**
Client, curl engine

**Motivation**
We need a way to support consistent set of features across all native platforms. Fixes KTOR-6754

**Solution**
We're using scripts to update cinterop for the following platforms:
* linuxX64/linuxArm64
* macosX64/macosArm64
* mingwX64


dtretyakov:
The following builds are failing:
* [Linux](https://ktor.teamcity.com/buildConfiguration/Ktor_KtorMatrixNativeLinuxX64/246500)
* [Windows](https://ktor.teamcity.com/buildConfiguration/Ktor_KtorMatrixNativeWindowsX64/246501)

e5l (changes requested):
> ld.lld: error: undefined symbol: explicit_bzero
@dtretyakov, could you check the build?

whyoleg:
@e5l probably could be closed in favor of #4445

PR #3986, opened 2024-02-21
KTOR-6642 Fix overriding MDC context map in CallLogging

marychatte:
**Subsystem**
Server, CallLogging

**Motivation**
[KTOR-6642](https://youtrack.jetbrains.com/issue/KTOR-6642)

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-call-logging/jvm/test/io/ktor/server/plugins/callloging/CallLoggingTest.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-call-logging/jvm/test/io/ktor/server/plugins/callloging/CallLoggingTest.kt
@@ -20,6 +20,7 @@ import io.ktor.server.routing.*
 import io.ktor.server.testing.*
 import io.ktor.util.logging.Logger
 import kotlinx.coroutines.*
+import kotlinx.coroutines.slf4j.*
 import org.fusesource.jansi.*
 import org.slf4j.*
 import org.slf4j.event.*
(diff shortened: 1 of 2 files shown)


ktor-server/ktor-server-plugins/ktor-server-call-logging/jvm/test/io/ktor/server/plugins/callloging/CallLoggingTest.kt
```diff
@@ -531,6 +532,33 @@ class CallLoggingTest {
...
+                MDC.put("name", "<name>")
+
+                withContext(MDCContext()) {
+                    call.respondText { "OK" }
+                }
+            }
+        }
+
+        client.get("/")
+        assertContains(messages, "INFO: / [hardcoded=1, mdc-uri=/, name=<name>]")
```
bjhham, line 559: I think we could use another assertion to ensure that the MDC is the reverted to the original state after the context of the call.  I don't know if this is how it works now, but it's how I imagine it _should_ work 🤔 

So something like:
```kotlin
client.get("/")
logger.info("call from outside ktor")

assertContains(messages, "INFO: / [hardcoded=1, mdc-uri=/, name=<name>]")
assertTrue { messages.any { "name" in it && "mdc-uri" !in it } }
```

bjhham (approved):

ktor-server/ktor-server-plugins/ktor-server-call-logging/jvm/test/io/ktor/server/plugins/callloging/CallLoggingTest.kt
marychatte: Yes, thank you a lot, I will change

bjhham (approved):

PR #3993, opened 2024-03-04
Add a header to SSE response to address the issue of nginx buffering …

mcxinyu:
Add a header to SSE response to address the issue of nginx buffering causing events to not be timely sending.

[link](https://stackoverflow.com/questions/48506093/nginx-buffering-flask-event-stream-in-docker-image)


e5l (approved):
Hey @mcxinyu, thanks for the PR. LGTM

PR #4003, opened 2024-03-22
Specific HTTP status codes have been added

lavaFrai:
Subsystem: Client/Server, HTTP Status Codes

Motivation:
This PR solves the problem of the lack of support for some specific HTTP status codes that can be used to improve communication between the client and the server. Adding these codes is important to provide a more accurate representation of the state of the system and its responses.

Solution:
The following HTTP status codes have been added to this PR:
- 418 I'm a Teapot: A code indicating that the server refuses to brew coffee because it is, constantly, a kettle. This is a reference to the Hyper Text Coffee Pot Control Protocol, defined in April Fools' Day jokes of 1998 and 2014.
- 267 Doubtful But Okay: The HTTP status of Oleg Tinkov's name is 267 Doubtful But Okay. This response status of the web server or application means that the parameters (URL or data) of the request have questionable semantics (possibly conflicting data was transmitted by the client), but, nevertheless, the request was accepted and processed. A little more about this code: https://github.com/maximal/http-267

Both codes were added in order to expand the server's capabilities for more accurate and informative customer feedback.


e5l:
Hey @lavaFrai, thanks for the PR!

I don't know if we can accept the PR. These names are not part of the standard and might be reused for different status codes :)

PR #4007, opened 2024-03-27
KTOR-6878 Fix stack overflow when SuspendFunctionGun trying to access…

Stexxe:
https://youtrack.jetbrains.com/issue/KTOR-6878


e5l (approved):
lgtm

PR #4008, opened 2024-03-28
KTOR-6851 Parsing Content-Length to Long instead of Int to prevent Nu…

Stexxe:
https://youtrack.jetbrains.com/issue/KTOR-6851


e5l (approved):
lgtm

PR #4009, opened 2024-03-28
KTOR-1219 Make Android logger log to Logcat on Android if the SLF4J b…

Stexxe:
https://youtrack.jetbrains.com/issue/KTOR-1219

### DIFF
--- a/ktor-client/ktor-client-plugins/ktor-client-logging/jvm/src/io/ktor/client/plugins/logging/LoggerJvm.kt
+++ b/ktor-client/ktor-client-plugins/ktor-client-logging/jvm/src/io/ktor/client/plugins/logging/LoggerJvm.kt
@@ -6,6 +6,7 @@ package io.ktor.client.plugins.logging
 
 import io.ktor.client.*
 import org.slf4j.*
+import org.slf4j.helpers.NOPLoggerFactory
 
 public actual val Logger.Companion.DEFAULT: Logger
     get() = object : Logger {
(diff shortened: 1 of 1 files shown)


ktor-client/ktor-client-plugins/ktor-client-logging/jvm/src/io/ktor/client/plugins/logging/LoggerJvm.kt
```diff
@@ -16,11 +17,37 @@ public actual val Logger.Companion.DEFAULT: Logger
...
+        }
+
+        return MessageLengthLimitingLogger(delegate = LogcatLogger(logClass))
+    }
+
+private class LogcatLogger(private val logClass: Class<*>) : Logger {
+    private val tag = "Ktor Client"
+    override fun log(message: String) {
+        runCatching {
+            val method = logClass.getDeclaredMethod("i", String::class.java, String::class.java)
```
e5l, line 46: please consider extracting this to the class level scope. I would also add fallback

ktor-client/ktor-client-plugins/ktor-client-logging/jvm/src/io/ktor/client/plugins/logging/LoggerJvm.kt
```diff
@@ -16,11 +17,37 @@ public actual val Logger.Companion.DEFAULT: Logger
...
- * length of 4068 characters
+ * Android [Logger]: Logs to the Logcat on Android if the SLF4J provider isn't found.
+ * Otherwise, uses the [Logger.Companion.DEFAULT].
+ * Breaks up long log messages that would be truncated by Android's max log
+ * length of 4068 characters.
  */
 public val Logger.Companion.ANDROID: Logger
-    get() = MessageLengthLimitingLogger()
+    get() {
+        val logger = Logger.DEFAULT
```
e5l, line 27: please consider making an internal object to avoid multiple initializations

e5l (changes requested):

ktor-client/ktor-client-plugins/ktor-client-logging/jvm/src/io/ktor/client/plugins/logging/LoggerJvm.kt
Stexxe: What do you mean? In the body of `get` the default logger is created only once.

e5l: It will be new every get call

e5l (approved):

PR #4012, opened 2024-03-29
KTOR-6883 Add callbacks before waiting for connection establishment t…

Stexxe:
https://youtrack.jetbrains.com/issue/KTOR-6883


e5l (approved):
lgtm

PR #4013, opened 2024-03-31
[KTOR-6886] Android Http Client using Android's HttpEngine on API 34+

yschimke:
**Subsystem**
Client

**Motivation**
On Android UrlConnection by default doesn't support HTTP/2, and is an old stack.

HttpEngine is now part of the Android 14 APIs (API 34+), and supports HTTP/2, HTTP/3 and much more advanced features.

Apps can install this as the process singleton URL handler, but this could have other impacts.

**Solution**

Rather than switching to the very different API HttpEngine, this change minimises the differences by using `HttpEngine.openConnection(URL)` and otherwise working the same as the existing UrlConnection code.

TODO

- [x] Cleanup
- [ ] Work out how to run the included but non functional instrumentation test


yschimke:
@e5l @osipxd let me know if I should rebase my change, or if you want to take over the PR?

yschimke:
@osipxd @bjhham I'm assuming there will be some directional changes requested, so any chance at a review?

Any opinions on instrumentation tests?

ktor-client/ktor-client-android/jvm/src/io/ktor/client/engine/android/Android14URLConnectionFactory.kt
```diff
@@ -0,0 +1,26 @@
...
+            "AndroidEngineConfig.context must be set when using HttpEngine; prefer applicationContext."
+        }.applicationContext
+        return HttpEngine.Builder(ctx)
+            .apply(config.httpEngineConfig)
+            .build()
+    }
+
+    override operator fun invoke(urlString: String): HttpURLConnection {
+        return engine.openConnection(URI.create(urlString).toURL()) as HttpURLConnection
+    }
```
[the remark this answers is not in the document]
yschimke: Ignore, new behaviour is just as valid. And avoids deprecated APIs.

yschimke:
@coderabbitai generate docstrings

ktor-client/ktor-client-android/jvm/src/io/ktor/client/engine/android/Android14URLConnectionFactory.kt
```diff
@@ -0,0 +1,27 @@
...
+import android.net.http.HttpEngine
+import java.net.HttpURLConnection
+import java.net.URI
+
+// @RequiresExtension(extension = Build.VERSION_CODES.S, version = 7)
+internal class AndroidNetHttpEngineFactory(private val config: AndroidEngineConfig) : URLConnectionFactory {
+    private val engine by lazy { buildEngine() }
+
+    private fun buildEngine(): HttpEngine {
+        val ctx = requireNotNull(config.context) {
```
Goooler:
```suggestion
        val ctx = checkNotNull(config.context) {
```

`requireNotNull` throws `IllegalArgumentException` instead of `IllegalStateException`.

MohammedKHC:
@e5l @osipxd @yschimke  Any update??

Mr3zee:
We are looking at it, sorry for the delay
Did you figure out your issue with the instrumentation tests?

yschimke:
I'll take a look again this weekend.

ktor-client/ktor-client-android/jvm/src/io/ktor/client/engine/android/AndroidEngineConfig.kt
```diff
@@ -45,4 +46,26 @@ public class AndroidEngineConfig : HttpClientEngineConfig() {
...
+
+    /**
+     * Android Context required to initialize the Android HttpEngine path (API 34 or S Extensions ≥ 7).
+     * If this is `null`, or when a proxy is configured, the engine falls back to the legacy
+     * `HttpURLConnection` path.
+     */
+    public var context: android.content.Context? = null
+        set(value) {
+            field = value
+        }
```
osipxd, line 70: I think we could improve DX here by providing a single function instead of two fields. Something like this:

```kotlin
public fun useHttpEngine(context: android.content.Context, configure: HttpEngine.Builder.() -> Unit = {})
```

ktor-client/ktor-client-android/jvm/src/io/ktor/client/engine/android/Android14URLConnectionFactory.kt
```diff
@@ -0,0 +1,27 @@
...
+
+package io.ktor.client.engine.android
+
+import android.net.http.HttpEngine
+import java.net.HttpURLConnection
+import java.net.URI
+
+// @RequiresExtension(extension = Build.VERSION_CODES.S, version = 7)
+internal class AndroidNetHttpEngineFactory(private val config: AndroidEngineConfig) : URLConnectionFactory {
+    private val engine by lazy { buildEngine() }
```
osipxd, line 13: Probably we should shutdown the engine from `AndroidClientEngine.close`

ktor-client/ktor-client-android/jvm/src/io/ktor/client/engine/android/AndroidClientEngine.kt
```diff
@@ -91,7 +102,7 @@ public class AndroidClientEngine(override val config: AndroidEngineConfig) : Htt
                 .mapKeys { it.key?.lowercase(Locale.getDefault()) ?: "" }
                 .filter { it.key.isNotBlank() }
 
-            val version: HttpProtocolVersion = HttpProtocolVersion.HTTP_1_1
+            val version: HttpProtocolVersion = urlFactory.protocolFromRequest(connection)
```
osipxd, line 105: **Note:** I've created an issue to fix this [KTOR-9728](https://youtrack.jetbrains.com/issue/KTOR-9728)

ktor-client/ktor-client-android/jvm/src/io/ktor/client/engine/android/AndroidClientEngine.kt
```diff
@@ -34,6 +34,16 @@ public class AndroidClientEngine(override val config: AndroidEngineConfig) : Htt
 
     override val supportedCapabilities: Set<HttpClientEngineCapability<*>> = setOf(HttpTimeoutCapability, SSECapability)
 
+    private val urlFactory = if (config.httpEngineDisabled ||
+        !isHttpEngineAvailable() ||
+        config.proxy != null ||
```
osipxd, line 39: It seems they've recently added [`setProxyOptions`](https://developer.android.com/reference/android/net/http/HttpEngine.Builder#setProxyOptions(android.net.http.ProxyOptions)) available on SDK 37+ and S extension 22+

Probably it makes sense to move this logic into `isHttpEngineAvailable` and pass the `config` as a parameter.

osipxd:
> Any opinions on instrumentation tests?

Do they fail to run with Robolectric? We don’t currently have infrastructure for running instrumented tests, but there is demand (for example in the WebRTC client), and this PR might be another good reason to spin up an emulator on CI.

PR #4015, opened 2024-04-01
KTOR-6850 Fix the query parameters object for each engine but Netty t…

Stexxe:
https://youtrack.jetbrains.com/issue/KTOR-6850

### DIFF
--- a/ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/request/PipelineRequest.kt
+++ b/ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/request/PipelineRequest.kt
@@ -97,3 +97,23 @@ public fun ApplicationRequest.encodeParameters(parameters: Parameters): Paramete
         }
     }.build()
 }
+
+/**
+ * Converts parameters to query parameters by fixing the [Parameters.get] method
+ * to make it return an empty string for the query parameter without value
+ */
+public fun Parameters.toQueryParameters(): Parameters {
+    val parameters = this
+    return object : Parameters {
+        override fun get(name: String): String? {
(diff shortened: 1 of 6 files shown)


ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/request/PipelineRequest.kt
```diff
@@ -97,3 +97,23 @@ public fun ApplicationRequest.encodeParameters(parameters: Parameters): Paramete
         }
     }.build()
 }
+
+/**
+ * Converts parameters to query parameters by fixing the [Parameters.get] method
+ * to make it return an empty string for the query parameter without value
+ */
+public fun Parameters.toQueryParameters(): Parameters {
```
e5l, line 105: Could we make it internal?

e5l (changes requested):

ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/request/PipelineRequest.kt
Stexxe: Do you mean with the internal annotation?

e5l: nope, visibility. I would rather copy the method if needed to keep the fix for the patch release

Stexxe: I don't think it's a good idea because it would require copying to 4 different modules.

e5l (approved):

PR #4017, opened 2024-04-04
Release 2.3.10

e5l:


Stexxe (approved):

e5l:
it looks like version bump cause problems, let me revert it back

e5l:
@bjhham, looks better now. Please check out

marychatte (commented):
It seems that our JS tests didn't run at all because of the absent `cleanWasmJsTest` task, that is only present from `3.0.0`. Also `Allocation test` doesn't compile

bjhham (approved):

PR #4018, opened 2024-04-04
Apply kotlinx-atomicfu compiler plugin

mvicsokolova:
**Motivation**
`kotlinx-atomicfu` is on the way to stabilization, as the next step I'm going to enable compiler transformations by default with the following getting rid of legacy transformations. For that it would be great to enable compiler plugin transformations in ktor. This PR enables JVM and Kotlin Native IR transformations performed by atomicfu-compiler-plugin. 

I've disabled JS IR transformations, for now there is a discussion regarding removal of the JS part of the compiler plugin.

**Solution**
Minor changes were made in the code: 
- replaced `minusAssign`/`plusAssign` operators with `addAndGet` call, because the operators are not supported by the compiler plugin.
- introduced atomic extension function to avoid passing atomic reference as a parameter (`AtomicRef<*>.setContinuation(..)`).
- atomic properties are recommended to be private, their values may be exposed via delegated properties.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -24,8 +24,8 @@ kotlin.incremental.multiplatform=true
 kotlin.mpp.applyDefaultHierarchyTemplate=false
 kotlin.native.ignoreIncorrectDependencies=true
 kotlin.native.binary.memoryModel=experimental
-#kotlinx.atomicfu.enableJvmIrTransformation=true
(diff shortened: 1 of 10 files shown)


mvicsokolova:
@e5l Could you please take a look? 🙏🏻

e5l (approved):

PR #4021, opened 2024-04-11
KTOR-6909 Add SocketTimeout for Test Engine

marychatte:
**Subsystem**
Test Client

**Motivation**
[KTOR-6909](https://youtrack.jetbrains.com/issue/KTOR-6909)


ktor-server/ktor-server-test-host/jvmAndNix/src/io/ktor/server/testing/TestApplicationEngine.kt
```diff
@@ -271,11 +275,13 @@ public class TestApplicationEngine(
     public fun createCall(
         readResponse: Boolean = false,
         closeRequest: Boolean = true,
+        timeoutAttributes: HttpTimeoutConfig? = null,
```
e5l, line 278: could we try making it without API change?

e5l (approved):

PR #4037, opened 2024-05-15
KTOR-7016 Lazy initialize iconv charsets

e5l:
Fix [KTOR-7016](https://youtrack.jetbrains.com/issue/KTOR-7016) Embedded Linux device without iso-8859-1 and UTF-16 cannot use ktor-network


marychatte (approved):
LGTM

PR #4042, opened 2024-05-17
Add fast compilation and jvmTest check for PRs

e5l:
Add fast check for compilation and jvmTests

### DIFF
--- a/.github/workflows/check-build.yml
+++ b/.github/workflows/check-build.yml
@@ -4,46 +4,24 @@ on:
   pull_request:
     branches:
       - main
+permissions:
+  contents: read
 
 jobs:
   compile:
-
     runs-on: ubuntu-latest
-
     steps:
-    - name: Checkout repository
-      uses: actions/checkout@v2
-
+    - uses: actions/checkout@v4.1.4
     - name: Set up JDK 11
-      uses: actions/setup-java@v2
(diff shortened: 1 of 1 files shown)


Mr3zee:
@e5l I'm not an expert in GH Actions, but maybe the error will go away when using this https://github.com/gradle/gradle-build-action ?

e5l:
I'm not sure, cause it's already started gradle but can't find dependencies:
```
e: file:///home/runner/work/ktor/ktor/ktor-io/common/src/io/ktor/utils/io/temporary.kt:4:50 Unresolved reference: coroutines
```

hfhbd:
It is the correct action, but using the argument syntax is deprecated. Anyway, this is unrelated to the actual compiling failure, which was caused by a wrong commit: I used the main branch and it did work: https://github.com/hfhbd/ktor/pull/91
BTW you can also remove the wrapper validation file and just enable: `validate-wrappers: true`. Alternatively, I can create the PR.

e5l:
@hfhbd, thanks. It helped.

@Mr3zee, fixed. Please check.

Mr3zee (dismissed):

hfhbd:
But keep in mind, the build doesn't fail if a jvm test failed.

e5l:
Yep, let me check if we can bring test results to the UI

PR #4043, opened 2024-05-21
Fix flaky testErrorForProtocolUpgradeRequestBody 

marychatte:
**Subsystem**
Client, CIO, SSE

**Solution**
Extract check of needing processing content from `launch`. Because otherwise, we have the following situation:
``` kotlin
fun makeDedicatedRequest(...) {
       try {
                writeRequest(...)
                readResponse(...)
        } catch (cause: Throwable) {
            // error must be caught here
            throw cause.mapToKtor(request)
        }
}

fun writeRequest(...) = withContext(callContext) {
    writeBody(...)
}

fun writeBody(...) {
    val scope = CoroutineScope(callContext + CoroutineName("Request body writer"))
    scope.launch {
        try {
            processOutgoingContent(...) // error is throwing here
        }
    }
}
```
So where is a race and a case when because of `launch` in `writeBody`, `readResponse` starts and finishes earlier than the error is throwing in `launch`, and the error isn't caught 

Proposed solution: Extract check before `launch`, so because it's in `withContext` `readResponse` will starts after it

### DIFF
--- a/ktor-client/ktor-client-cio/jvmAndNix/src/io/ktor/client/engine/cio/utils.kt
+++ b/ktor-client/ktor-client-cio/jvmAndNix/src/io/ktor/client/engine/cio/utils.kt
@@ -111,6 +111,9 @@ internal suspend fun writeBody(
         if (closeChannel) output.close()
         return
     }
+    if (request.body is OutgoingContent.ProtocolUpgrade) {
+        throw UnsupportedContentTypeException(request.body)
+    }
 
     val contentLength = request.headers[HttpHeaders.ContentLength] ?: request.body.contentLength?.toString()
     val contentEncoding = request.headers[HttpHeaders.TransferEncoding]
@@ -123,7 +126,7 @@ internal suspend fun writeBody(
     val scope = CoroutineScope(callContext + CoroutineName("Request body writer"))
     scope.launch {
         try {
-            if (!processOutgoingContent(request, request.body, channel)) return@launch
+            processOutgoingContent(request, request.body, channel)
         } catch (cause: Throwable) {
             channel.close(cause)
             throw cause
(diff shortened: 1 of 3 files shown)


ktor-client/ktor-client-cio/jvmAndNix/src/io/ktor/client/engine/cio/utils.kt
```diff
@@ -142,20 +144,21 @@ internal suspend fun writeBody(
...
-    request: HttpRequestData,
-    body: OutgoingContent,
-    channel: ByteWriteChannel
-): Boolean {
+private fun needToProcessOutgoingContent(body: OutgoingContent): Boolean = when (body) {
+    is OutgoingContent.NoContent -> false
+    is OutgoingContent.ProtocolUpgrade -> throw UnsupportedContentTypeException(body)
+    is OutgoingContent.ContentWrapper -> needToProcessOutgoingContent(body.delegate())
+    else -> true
+}
```
bjhham, line 152: Might be better here to introduce an `isEmpty()` function in `OutgoingContent` then override in the implementations.  The default behaviour could be something like `contentLength?.let { it > 0 } == true`.

So then the check would be `if (request.body.isEmpty()) return`.

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/plugins/ServerSentEventsTest.kt
```diff
@@ -351,28 +351,30 @@ class ServerSentEventsTest : ClientLoader(timeoutSeconds = 120) {
     }
 
     @Test
-    fun testErrorForProtocolUpgradeRequestBody() = clientTests(listOf("OkHttp")) {
-        config {
-            install(SSE)
-        }
+    fun testErrorForProtocolUpgradeRequestBody() = repeat(10000) {
```
bjhham, line 354: 10k iterations seems a little excessive here 😄

ktor-client/ktor-client-cio/jvmAndNix/src/io/ktor/client/engine/cio/utils.kt
```diff
@@ -142,20 +145,15 @@ internal suspend fun writeBody(
...
+private suspend fun processOutgoingContent(request: HttpRequestData, body: OutgoingContent, channel: ByteWriteChannel) {
     when (body) {
-        is OutgoingContent.NoContent -> return false
         is OutgoingContent.ByteArrayContent -> channel.writeFully(body.bytes())
         is OutgoingContent.ReadChannelContent -> body.readFrom().copyAndClose(channel)
         is OutgoingContent.WriteChannelContent -> body.writeTo(channel)
-        is OutgoingContent.ProtocolUpgrade -> throw UnsupportedContentTypeException(body)
         is OutgoingContent.ContentWrapper -> processOutgoingContent(request, body.delegate(), channel)
+        is OutgoingContent.ProtocolUpgrade -> error("unreachable code")
+        is OutgoingContent.NoContent -> error("unreachable code")
```
bjhham, line 155: I wonder if you inlined this function if the compiler would be able to figure this out... maybe with the 2.0 compiler 🤔.  I'll check on it when I do the upgrade.

bjhham (approved):

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/plugins/ServerSentEventsTest.kt
marychatte: Heh, it's not finished yet 😄 will remove it after green CI tests to be sure it's not flaky anymore

PR #4044, opened 2024-05-22
Use KotlinHierarchyTemplate for structuring source sets

antohaby:
KotlinHierarchyTemplate allows writing KMP source set structures with less code and without explicit dependsOn relations.

There is one downside is that all source sets are created now if at least one of the target is declared. i.e. for example if JS target is declared but wasm is not. Then jsAndWasmSharedMain source set will be created anyway.

### DIFF
--- a/buildSrc/src/main/kotlin/TargetsConfig.kt
+++ b/buildSrc/src/main/kotlin/TargetsConfig.kt
@@ -5,6 +5,7 @@
 
 import org.gradle.api.*
 import org.gradle.kotlin.dsl.*
+import org.jetbrains.kotlin.gradle.*
 import org.jetbrains.kotlin.gradle.plugin.*
 import org.jetbrains.kotlin.gradle.targets.js.dsl.*
 import java.io.*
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/kotlin/TargetsConfig.kt
```diff
@@ -52,31 +53,47 @@ fun Project.configureTargets() {
...
-            if (hasJsAndWasmShared) {
-                val commonMain by getting {}
-                val jsAndWasmSharedMain by creating {
-                    dependsOn(commonMain)
-                }
-                val commonTest by getting {}
-                val jsAndWasmSharedTest by creating {
-                    dependsOn(commonTest)
+        @OptIn(ExperimentalKotlinGradlePluginApi::class)
+        val ktorTargetsHierarchy = KotlinHierarchyTemplate {
```
whyoleg, line 57: just curious: is there a reason to extract this into a separate variable instead of just calling `applyHierarchyTemplate {...}`?

buildSrc/src/main/kotlin/TargetsConfig.kt
```diff
@@ -52,31 +53,47 @@ fun Project.configureTargets() {
...
+            // extract nix group so it can be reused later as `group("nix")`
+            group("nix") {
```
whyoleg, line 59: just curious[2]: is there a reason to declare it top-level and not under `common`? Does it make any difference?

buildSrc/src/main/kotlin/TargetsConfig.kt
```diff
@@ -89,205 +106,13 @@ fun Project.configureTargets() {
...
-                        implementation(kotlin("test"))
-                    }
-                }
-
-                windowsTargets().forEach {
-                    getByName("${it}Main").dependsOn(windowsMain)
-                    getByName("${it}Test").dependsOn(windowsTest)
-                }
-            }
+            if (hasJvmAndNix) { jvm(); nixTargets(); }
```
whyoleg, line 109: just minor unrelated remark: those functions (nixTargets/posixTargets/etc) are extensions on `Project` - so it would be more clear to call them at least at `kotlin` extension level (because of `jvm` function call).
Even better to refactor them to be extensions on `KotlinMultiplatformExtension` - but of course it's out of scope here :)

whyoleg (commented):
just passing by...
it's just nice to see how this new `hierarchy` API simplifies complex sourceSet hierarchies

buildSrc/src/main/kotlin/TargetsConfig.kt
antohaby: there is no reason. so it can be merged into one call.

buildSrc/src/main/kotlin/TargetsConfig.kt
antohaby: it can be moved under common. however then it might look not entirely correct as I don't want nix to be dependsOn common, but only on posix and jvmAndNix (but apply target hierarchy algorithm is capable to solve this problem as in this case it will not add direct dependsOn edge).

again I wasn't doing that with much of a through thinking of where exactly to place this group.

buildSrc/src/main/kotlin/TargetsConfig.kt
antohaby: agree

buildSrc/src/main/kotlin/TargetsConfig.kt
antohaby: > but apply target hierarchy algorithm is capable to solve this problem as in this case it will not add direct dependsOn edge

or maybe not, I need to double check

antohaby: > or maybe not, I need to double check
checked. it is alright. here is the source set structure I get via this script:

```
tasks.register("printKMPSourceSetStructure") {
    doFirst {
        println("digraph {")
        kotlin.sourceSets.forEach { sourceSet ->
            sourceSet.dependsOn.forEach { dependsOn ->
                println("  \"${sourceSet.name}\" -> \"${dependsOn.name}\"")
            }
        }
        println("}")
    }
}
```

image

whyoleg: cool, thanks for explaining and checking :)

antohaby:
Ok I pushed changed based on Oleg's input.

There are few interesting pictures I'd like to show:

This is how SourceSet structure looked like in :ktor-network project. Impossible to read. 
image


But here is normalized version where all redundant dependsOn edges is removed. And it is much cleaner
image

But take a look for example at iosTest you can see that it doesn't connect neither to commonTest nor to jvmAndNixTest.
And other source sets are also dongle here and there. 

With applyTargetHierarchy it look now like this. Much better and comprehendable. And I don't have to normalize this version either. So it is already in a good shape.
image

PR #4045, opened 2024-05-22
Use KotlinHierarchyTemplate for structuring source sets

antohaby:
KotlinHierarchyTemplate allows writing KMP source set structures with less code and without explicit dependsOn relations.

There is one downside is that all source sets are created now if at least one of the target is declared. i.e. for example if JS target is declared but wasm is not. Then jsAndWasmSharedMain source set will be created anyway.

It also fixes the issue with Kotlin 2.0 
In :ktor-network jvmAndNixTest has an expect that in K1 was actualized in iosTest there was no direct or transitive dependsOn relation from iosTest to jvmAndNixTest and thus in K2 such actualization was prohibited. In this commit Source Set structure is sanitized and now dependsOn relations are in good order.

**Subsystem**
Gradle build

**Motivation**
Depends on relations between Kotlin Source Sets are messed in Ktor project. And Ktor is not to blame here, but Kotlin Multiplatform's dependsOn relations. They are error prone and very verbose. And since K2 Kotlin Compiler requires correctness in Source Set structure. i.e. in K1 due to its compilation model it was possible to violate symbols visibility and expect/actual. A bit more details you can find in related issue: https://youtrack.jetbrains.com/issue/KT-68212/K2-KMP-has-no-corresponding-expected-declaration-with-dependsOn-relationship

This is how SourceSet structure looked like in :ktor-network project. Impossible to read. 
image

Same Source Set structure but normalized i.e. redundant depends on edges are removed
image

But take a look for example at iosTest you can see that it doesn't connect neither to commonTest nor to jvmAndNixTest.
And other source sets are also dongle here and there.

**Solution**
Apply kotlin target hierarchy template.  there is some information on kotlinlang web site https://kotlinlang.org/docs/multiplatform-hierarchy.html#default-hierarchy-template 

After applying the target hierarchy source set structure look much cleaner for both main and test source sets. And it doesn't have to be normalized. It is already in a good shape.
image

PS: kotlin-community/dev branch also has related PR (!4044)

### DIFF
--- a/buildSrc/src/main/kotlin/TargetsConfig.kt
+++ b/buildSrc/src/main/kotlin/TargetsConfig.kt
@@ -5,7 +5,7 @@
 
 import org.gradle.api.*
 import org.gradle.kotlin.dsl.*
-import org.jetbrains.kotlin.gradle.plugin.*
+import org.jetbrains.kotlin.gradle.*
 import org.jetbrains.kotlin.gradle.targets.js.dsl.*
 import java.io.*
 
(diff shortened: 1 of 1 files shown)


e5l:
Hey @antohaby, thanks for the PR. I'll check it next week after the conference!

e5l (approved):

e5l:
Hey @antohaby, thanks for the PR. LGTM

PR #4049, opened 2024-05-27
Revert "fix(deps): update coroutines.version to v1.8.1 (#4028)"

bjhham:
This reverts commit d96f14697c9371f0efe07e1349b9b65f7d53e646.

All of our common compilation targets have been broken for the last week after this dependency update.  I just noticed today that the allocation tests target on all PRs are failing because of this.


bjhham:
The allocation tests still failed, but it DID compile 🎉

PR #4053, opened 2024-05-28
Fix NoTransformationFoundException when SSE with Logging plugin

marychatte:
**Subsystem**
Client, SSE, Logging

**Motivation**
[KTOR-6912](https://youtrack.jetbrains.com/issue/KTOR-6912)


bjhham (approved):
lgtm 👍

PR #4058, opened 2024-05-30
Fix missing charset for static js, css and svg resources

marychatte:
**Subsystem**
Server, Static

**Motivation**
[KTOR-6655](https://youtrack.jetbrains.com/issue/KTOR-6655)

1) `application/javascript` obsoleted in favor of `text/javascript` (https://www.iana.org/assignments/media-types/application/javascript)
2) Add charset for `svg`, it's optional parameter (https://www.w3.org/TR/SVG/mimereg.html)

### DIFF
--- a/ktor-http/common/src/io/ktor/http/FileContentType.kt
+++ b/ktor-http/common/src/io/ktor/http/FileContentType.kt
@@ -65,7 +65,14 @@ private val extensionsByContentType: Map<ContentType, List<String>> by lazy {
 internal fun List<ContentType>.selectDefault(): ContentType {
     val contentType = firstOrNull() ?: ContentType.Application.OctetStream
     return when {
-        contentType.contentType == "text" && contentType.charset() == null -> contentType.withCharset(Charsets.UTF_8)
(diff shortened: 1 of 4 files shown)


ktor-http/common/src/io/ktor/http/FileContentType.kt
```diff
@@ -65,7 +65,14 @@ private val extensionsByContentType: Map<ContentType, List<String>> by lazy {
...
     return when {
-        contentType.contentType == "text" && contentType.charset() == null -> contentType.withCharset(Charsets.UTF_8)
+        contentType.match(ContentType.Text.Any) && contentType.charset() == null -> contentType.withCharset(
+            Charsets.UTF_8
+        )
+
+        contentType.match(ContentType.Image.SVG) && contentType.charset() == null -> contentType.withCharset(
+            Charsets.UTF_8
+        )
+
```
bjhham, line 75: Seems like there's quite a few other text-based mime types hidden under "application/*", like xml, rss, soap.  Could add them in while you're in the neighbourhood.

bjhham (approved):
👍

PR #4061, opened 2024-05-31
Update MavenPublication's scm configuration

mr3y-the-programmer:
Update scm url and add connection & developer connection properties.

This change makes ktor searchable & indexable by KMP libraries portals like https://klibs.io/


mr3y-the-programmer:
Apparently, it was https://klibs.io/ fault which wasn't handling urls ending in `.git` properly, see https://x.com/IgnatBeresnev/status/1796638523384107374

That being said, I think I can close this PR.

PR #4065, opened 2024-06-03
KTOR-7043 Explicit testApplication configuration from file

bjhham:
**Subsystem**
Testing

**Motivation**
[KTOR-7043](https://youtrack.jetbrains.com/issue/KTOR-7043) Test Application explicit loading of modules

I noticed this when generating projects with the file configuration options (hocon or yaml), the tests would break due to duplicate modules being loaded.  This seems like a pretty significant design problem with the `testApplication` API, because it makes it so you cannot change your configuration without breaking all of your tests, and it hinders test isolation for checking particular modules.

**Solution**
Removed implicit `DefaultTestConfig` function call for JVM tests, and instead use an empty `MapConfiguration` for all platforms.

The current solution could perhaps be improved with some additional DSL.  If you want to load the file confuration now, you need something like this:

```
environment {
    config = ConfigLoader.load()
}
```

Note, this should only go into 3.0.0.

### DIFF
--- a/ktor-server/ktor-server-core/jvm/src/io/ktor/server/config/HoconApplicationConfig.kt
+++ b/ktor-server/ktor-server-core/jvm/src/io/ktor/server/config/HoconApplicationConfig.kt
@@ -5,7 +5,6 @@
 package io.ktor.server.config
 
 import com.typesafe.config.*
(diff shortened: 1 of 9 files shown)


Stexxe (approved):

marychatte (approved):
LGTM

PR #4066, opened 2024-06-04
Add EventGroupSelectorManager instead of ActorSelectorManager

marychatte:


ktor-network/jvm/src/io/ktor/network/selector/eventgroup/Eventloop.kt
```diff
@@ -0,0 +1,104 @@
...
+
+import io.ktor.network.selector.*
+import kotlinx.coroutines.*
+import java.nio.channels.*
+
+internal class Eventloop {
+    val scope = newThreadContext(nThreads = 1).wrapInScope()
+
+    fun run(): Job {
+        return scope.launch { runLoop() }
```
andrsuh, line 15: What if coroutine complete unexpectedly? Can this lead to the situation where we loosing event loops one by one silently? Do we need some completion handler that analyse if the completion was controlled or unexpected, notify about this and resurrect coroutine to maintain the constant number of event loops?

e5l:
@marychatte, could you check the CI?

PR #4073, opened 2024-06-13
KTOR-7094 Update netty to 4.1.111.Final

bjhham:
**Subsystem**
Server, Netty

**Motivation**
[KTOR-7094](https://youtrack.jetbrains.com/issue/KTOR-7094) Update netty to 4.1.111.Final

**Solution**
Update version number.


e5l:
btw we just have a user PR is open :)

bjhham:
haha I only considered that after opening this and it was too late 😄

PR #4074, opened 2024-06-13
KTOR-360: add parseUrl function

anton-erofeev:
**Subsystem**
ktor-http

**Motivation**
issue https://github.com/ktorio/ktor/issues/1903. 
Currently URL parsing returns a valid URL when an invalid URL is passed in (for "bogus" it returns "http://localhost/bogus"). Also there is no method that would return null instead of an exception if url parsing fails.

**Solution**
added `parseUrl` function to the `URLUtils.kt`, it tries to parse the URL using the passed string and, if unsuccessful, returns null


e5l (approved):
Hey @anton-erofeev, thanks for the PR!

LGTM, will merge after CI

PR #4094, opened 2024-06-20
Make MalformedInputException actually inherit from exception

DRSchlaubi:
**Subsystem**
io utils

**Motivation**
Since the class is called Exception, it should be an ✨ Exception ✨
**Solution**
I decided to make it an `IOException` as the JVM version also extends from IOException


e5l (approved):

e5l:
Hey @DRSchlaubi, thanks for the PR. LGTM

PR #4095, opened 2024-06-21
KTOR-4273 Fix typo in CallLogging plugin package

bjhham:
**Subsystem**
Server

**Motivation**
- [KTOR-4273](https://youtrack.jetbrains.com/issue/KTOR-4273) CallLogging: package is misspelled

**Solution**
Add g.


bjhham:
@e5l let me know if you think I should introduce a type alias here to avoid breaking changes.  It's a bit cleaner to just change the package name, but also it's best to avoid breaking changes when possible 🤷

e5l (approved):
Good job! I'm fine with renaming, but we need to document the breaking changes. @vnikolova could you help?

vnikolova:
@e5l Sure, we can document this in the migration guide for `3.0.0`.

PR #4104, opened 2024-06-27
Update kotlin to 2.0.0 and fix native compilation

e5l:


bjhham:
Looks like TC's hitting some compile problems with the IO stuff.  This may be why I had opted to wait for the kotlinx-io merge originally.

e5l:
It worked locally for me (except Charsets fix). Let me try to reproduce.

e5l:
@bjhham, you're right. We can't suppress new warnings

PR #4105, opened 2024-06-27
KTOR-6417 Allow changing dispatcher in the client

e5l:
Fix [KTOR-6417](https://youtrack.jetbrains.com/issue/KTOR-6417) Ktor Client MockEngine dispatcher handle removed
[KTOR-6462](https://youtrack.jetbrains.com/issue/KTOR-6462) Ktor clients and servers should use Dispatchers.IO.limitedParallelism(...) wherever possible

We need the possibility to set custom dispatchers to enable users to reuse dispatchers and configure test-specific dispatchers in tests.


marychatte (approved):
LGTM, only please check JPMS check

e5l:
It looks like the JPMS check is broken in the main. I will check it separatelly

PR #4107, opened 2024-06-28
Remove lookahead session from CIO writer

bjhham:
This seems to fix the memory problems currently in the CIO engine for kotlinx-io.  There is still an issue with the Netty pipeline after attempting to make the change there.


e5l (approved):
Looks nice!

PR #4110, opened 2024-07-07
KTOR-7190: Add support for encoded non-text queries

demitsuri:
**Subsystem**
The changes only affect URLBuilder internals and some other internal APIs. No public API was affected.

**Motivation**
It fixes https://youtrack.jetbrains.com/issue/KTOR-7190/URLBuilder.encodedParameters-cant-handle-encoded-non-text-query-data.

**Solution**
Former solution assumed that any query parameter is able to be decoded as an UTF-8 string. And any encoded query parameter was decoded during the building an URL and after that encoded back to create a textual representation of an URL. Unfortunately it crashes an app if it tries to add something like bittorrent info_hash parameter (also see https://github.com/ktorio/ktor/issues/1351#issuecomment-561324452) because there are a lot of byte sequences which can't be decoded as an UTF-8 string.

My solution reverses the data flow. I added one special ParametersBuilder for undecodable query parameters where they are stored in encoded form. And URLBuilder.encodedParameters became just a wrapper which stores decodable parameters in URLBuilder.parameters and undecodable ones in the private URLBuilder.rawEncodedParameters.


demitsuri:
@bjhham Please take a look.

bjhham (changes requested):
So there's a few things I don't like about this approach of using two maps for the parameters:

1. The order of the parameters coming out is not the same as it is going in.
2. It further complicates an already-complicated set of types.
3. It creates problems for any code interacting with `Url.parameters` where it's expecting a complete set.

I'd propose removing the `isDecodable` function and instead introducing a new line of functions like `String.tryDecode` that instead returned a sealed type `UrlString { Decoded, Encoded }`, then the `Parameters` and `ParametersBuilder` types can just use these types internally instead of decoding everything then encoding again later.

I can help with updating the `Parameter` types if you'd like; I've been working on a pretty large refactor of our URL types to try to resolve some of our problems with default requests etc.

demitsuri:
@bjhham Thank you for your review. I'll try to update the code according to your proposal shortly.

Mr3zee:
No activity for a while, closing the PR. If you want to reopen it, please address the issues above.

PR #4111, opened 2024-07-07
KTOR-7193 RoutingRoot: remove preemptive tracing check

SIMULATAN:
**Subsystem**
Server Core

**Motivation**
The check prevented users from changing the log level of the `LOGGER` at a later point as the interceptor responsible isn't added.
See [KTOR-7193](https://youtrack.jetbrains.com/issue/KTOR-7193/Tracing-allow-changing-log-level-at-runtime) for an extensive rationale.

**Solution**
Remove the check. The SLF4j implementation will do it prior to every log anyway. The performance impact should be absolutely neglectable.


e5l (approved):
Hey @SIMULATAN, thanks for the PR. LGTM

PR #4115, opened 2024-07-11
Fix race condition on rate limiter plugin

elfogre:
**Subsystem**
Server, rate limiter plugin

**Motivation**
There is a comment about a race condition that can be easily solved. Checking reference of element to delete avoid to delete incorrect registry entries.

**Solution**
When removing from a concurrentMap we can use remove(key, entry) to atomically check if entry is the same element we want to delete from map. This way no race condition could happened. Also I add a check to not add jobs to remove Unlimited rateLimiters, as it would never be removed.

To test it (race condition scenarios usually involved difficult tests) I need to change visibility of an internal class. I accept suggestions here :sweat_smile:


ktor-server/ktor-server-plugins/ktor-server-rate-limit/jvmAndNix/src/io/ktor/server/plugins/ratelimit/RateLimit.kt
```diff
@@ -14,7 +14,7 @@ internal val LOGGER = KtorSimpleLogger("io.ktor.server.plugins.ratelimit.RateLim
 internal val LIMITER_NAME_GLOBAL = RateLimitName("KTOR_GLOBAL_RATE_LIMITER")
 internal val LIMITER_NAME_EMPTY = RateLimitName("KTOR_NO_NAME_RATE_LIMITER")
 
-internal data class ProviderKey(
+public data class ProviderKey(
```
e5l, line 17: Could we move tests in ktor-server-rate-limit with the `testApplication` and make it internal?

e5l (approved):
Hey, @elfogre! Looks good. Could you adjust a single thing before merging?

e5l (approved):

PR #4116, opened 2024-07-11
KTOR-7201 Fix data truncated in receiveParameters and receiveMultipart

e5l:
Fix [KTOR-7201](https://youtrack.jetbrains.com/issue/KTOR-7201) Data truncated in receiveParameters and receiveMultipart

### DIFF
--- a/ktor-http/ktor-http-cio/jvm/src/io/ktor/http/cio/CIOMultipartDataBase.kt
+++ b/ktor-http/ktor-http-cio/jvm/src/io/ktor/http/cio/CIOMultipartDataBase.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2019 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ * Copyright 2014-2024 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
  */
 
 package io.ktor.http.cio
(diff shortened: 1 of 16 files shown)


ktor-http/ktor-http-cio/jvm/src/io/ktor/http/cio/CIOMultipartDataBase.kt
```diff
@@ -67,8 +68,18 @@ public class CIOMultipartDataBase(
...
+            val packet = body.readRemaining(formFieldLimit.toLong())
+            if (!body.isClosedForRead) {
+                body.awaitContent()
+                if (!body.isClosedForRead) {
+                    packet.release()
+                    val cause = IllegalStateException("Form field size limit exceeded: $formFieldLimit")
+                    body.cancel(cause)
+                    throw cause
+                }
+            }
```
bjhham, line 82: Could introduce an extension method here like `body.assertClosedForRead { packet.release(); "Form field size limit exceeded: $formFieldLimit" }` to make it a little more declarative and reuseable.

bjhham (approved):

ktor-http/ktor-http-cio/jvm/src/io/ktor/http/cio/CIOMultipartDataBase.kt
e5l: I'll make `suspend fun ByteReadChannel.exhausted(): Boolean` instead

PR #4117, opened 2024-07-11
KTOR-2914 Try to fix crash due to network errors

Thomas-Vos:
[KTOR-2914](https://youtrack.jetbrains.com/issue/KTOR-2914) "java.lang.IllegalArgumentException: Failed requirement." in SelectorManagerSupport

### DIFF
--- a/ktor-network/jvm/src/io/ktor/network/sockets/DatagramSocketImpl.kt
+++ b/ktor-network/jvm/src/io/ktor/network/sockets/DatagramSocketImpl.kt
@@ -7,6 +7,7 @@ package io.ktor.network.sockets
 import io.ktor.network.selector.*
 import io.ktor.network.util.*
 import io.ktor.utils.io.core.*
+import io.ktor.utils.io.errors.*
 import kotlinx.coroutines.*
 import kotlinx.coroutines.channels.*
 import java.nio.*
@@ -51,6 +52,7 @@ internal class DatagramSocketImpl(
                 channel.send(receiveImpl())
             }
         } catch (_: ClosedChannelException) {
+        } catch (cause: IOException) {
         }
     }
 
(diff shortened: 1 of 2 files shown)


ktor-network/jvm/src/io/ktor/network/sockets/DatagramSocketImpl.kt
```diff
@@ -51,6 +52,7 @@ internal class DatagramSocketImpl(
                 channel.send(receiveImpl())
             }
         } catch (_: ClosedChannelException) {
+        } catch (cause: IOException) {
```
Thomas-Vos, line 55: This can be thrown by `channel.receive(buffer)` in the same file, so the exception must be caught here to prevent the app from crashing. This was already done for native: https://github.com/ktorio/ktor/blob/main/ktor-network/nix/src/io/ktor/network/sockets/DatagramSocketNative.kt#L50

ktor-network/nix/src/io/ktor/network/sockets/DatagramSocketNative.kt
```diff
@@ -48,6 +48,7 @@ internal class DatagramSocketNative(
             }
         } catch (_: ClosedSendChannelException) {
         } catch (cause: IOException) {
+        } catch (cause: PosixException) {
```
Thomas-Vos, line 51: `PosixException` is throw further down in this file, so the exception must be caught here to prevent the app from crashing.

Thomas-Vos:
Hi @e5l, could you please look into this as you investigated the issue in the past?

e5l:
Sure, let me check!

e5l (approved):
Hey @Thomas-Vos, thank you! LGTM

e5l:
I'm not sure if the all the problems are fixed, could you tell me if we can also have a test for it?

Thomas-Vos:
I'll see what I can do, but last time I failed reproducing it in a test, but I can try again. First I want to use a Ktor eap build in my next app update and see if all is resolved now.

PR #4120, opened 2024-07-13
KTOR-7209: fix Text.chomp values separator

anton-erofeev:
**Subsystem**
ktor-utils

**Motivation**
issue: [KTOR-7209](https://youtrack.jetbrains.com/issue/KTOR-7209/io.ktor.util.TextKt.chomp-doesnt-work-on-strings-with-more-than-one-character)
`chomp` assumes the input string is exactly one character, however it is expected to work with any non-empty separator

**Solution**
Fixed offset in substring from `1` to `separator.length`


e5l (approved):
lgtm

PR #4140, opened 2024-07-21
KTOR-5289 Change TCP connect on native to non-blocking and cancellable

Thomas-Vos:
[KTOR-5289](https://youtrack.jetbrains.com/issue/KTOR-5289) withTimeout doesn't cancel socket connection on native

This PR changes the connect call to not block. That means connect is now cancellable and works with timeouts. Furthermore, I noticed that in some conditions connect can block for very long (on iOS), which is now also resolved as it can be cancelled.

On JVM the connect call is already non-blocking, so no changes needed there.

My second commit changes the exception type from `IllegalStateException` to `IOException`, as I think this is better . You can revert it if needed as it is a separate commit.


Thomas-Vos:
I didn't like that all types of `Throwable` are caught, so changed it to catch only `PosixException`. This should be the only exception that can be thrown, right? Now also set as `cause` so it is easier to debug errors. Also in a separate commit.

e5l (approved):

e5l (approved):
lgtm

PR #4157, opened 2024-07-24
KTOR-6946 headers in the default request are overriding new custom headers

igloo12:
**Subsystem**
Client

**Motivation**
I need to override the default headers with custom headers 

**Solution**
The issue is with this code block. 
```
val defaultRequest = DefaultRequestBuilder().apply {
    headers.appendAll(this@intercept.context.headers)
    plugin.block(this)
}
```

The plugin call is erasing the headers. I reversed the order and added a new method to the builder to replace the ones set by the default block 
```
 val defaultRequest = DefaultRequestBuilder().apply {
    plugin.block(this)
    headers.replaceAll(this@intercept.context.headers)
}
```

### DIFF
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/DefaultRequest.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/DefaultRequest.kt
@@ -70,8 +70,12 @@ public class DefaultRequest private constructor(private val block: DefaultReques
             scope.requestPipeline.intercept(HttpRequestPipeline.Before) {
                 val originalUrlString = context.url.toString()
                 val defaultRequest = DefaultRequestBuilder().apply {
-                    headers.appendAll(this@intercept.context.headers)
                     plugin.block(this)
+                    val contextHeaders: HeadersBuilder = this@intercept.context.headers
+                    contextHeaders.entries().forEach { (name, values) ->
+                        headers.remove(name)
+                        headers.appendAll(name, values)
+                    }
                 }
                 val defaultUrl = defaultRequest.url.build()
                 mergeUrls(defaultUrl, context.url)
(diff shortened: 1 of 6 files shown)


marychatte (commented):
@igloo12 Thank you for the PR! I noticed that these changes break some tests in `DefaultRequestTest.kt`, and I'm not sure if it's better to change the plugin's behavior. Isn't the next workaround applicable to your case?
```kotlin
defaultRequest {
    headers.appendIfNameAbsent(HttpHeaders.ContentType, ContentType.Application.Json.toString())
}
```

igloo12:
Hi @marychatte ,

Thank you for the response. Yes, I can do that, or I can call the set method which replaces it. I think the issue is the terminology for the headers is confusing. I assumed I was appending to the headers list. Instead, it is appending to a value list for a single header.  I will close this PR.

igloo12:
Hi @marychatte, 

I messed up. I zoomed in too much. This is still a bug because you can  only have one content type, so it should always replace

igloo12:
@marychatte, I changed the PR to override only the content type. My logic is the content type set via `contextType()`, which hides the underlying header setting/appending logic. When a user sets contextType, the default should be overridden because there can only be one. 

What do you think?

e5l:
@marychatte, could you check?

marychatte:
As described in the PR, the plugin `DefaultRequest` can override headers that the user sets. However, the  user's headers should have higher priority:
```kotlin
defaultRequest {
    contentType(ContentType.Application.Json) // defaultRequest header
}

val response = client.get("/") {
    contentType(ContentType.Application.Xml) // users' header
}
```
The final `ContentType` should be `ContentType.Application.Xml`

So, we need to set the user's headers, and also because `defaultRequest` logic could depend on them, for example:
```kotlin
defaultRequest {
   headers.appendIfNameAbsent("header", "value)
 }
```

But there is a small problem that some `Headers` functions override value completely, like with function `contentType(...)`, because they are written through `headers.set(...)`

**Proposed solution:** 
1. Set and save headers after user
2. Apply `DefaultRequest` plugin
3. Check that the user's headers weren't overridden: we just append firstly user header values and then `DefaultRequest` header values
_Note:_ Different check for the `HttpHeaders.Cookie` because if it has several values, they are presented as one ,and they are combined automatically in the function `cookie(...)`

e5l (approved):
lgtm

PR #4163, opened 2024-07-28
KTOR-5936: add support for CORS origin with trailing slash

anton-erofeev:
**Subsystem**
ktor-server-plugin/ktor.server.plugins.cors

**Motivation**
[KTOR-5936](https://youtrack.jetbrains.com/issue/KTOR-5936): every origin with a trailing slash at the end is considered valid and the request passes successfully, even if this origin was not added to the CORS config

**Solution**
Corrected origin validation and origin normalization

### DIFF
--- a/ktor-server/ktor-server-tests/jvmAndNix/test/io/ktor/tests/server/plugins/CORSTest.kt
+++ b/ktor-server/ktor-server-tests/jvmAndNix/test/io/ktor/tests/server/plugins/CORSTest.kt
@@ -1008,6 +1008,30 @@ class CORSTest {
     }
 
     @Test
+    fun testCorsValidationWithTrailingSlashOrigin () = testApplication {
+        application {
+            install(CORS){}
+            routing {
+                get {
+                    call.respond(HttpStatusCode.OK)
+                }
+            }
+        }
+
+        client.get("/") {
+             header(HttpHeaders.Origin, "http://localhost:3000/")
+        }.let {
+            assertEquals(HttpStatusCode.Forbidden, it.status)
+        }
+
+        client.get("/") {
+            header(HttpHeaders.Origin, "http://localhost:3000")
+        }.let {
+            assertEquals(HttpStatusCode.Forbidden, it.status)
+        }
+    }
+
+    @Test
     fun originValidation() = testApplication {
         install(CORS) {
             allowSameOrigin = false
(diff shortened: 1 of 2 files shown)


ktor-server/ktor-server-tests/jvmAndNix/test/io/ktor/tests/server/plugins/CORSTest.kt
```diff
@@ -1007,6 +1007,30 @@ class CORSTest {
...
+             header(HttpHeaders.Origin, "http://localhost:3000/")
+        }.let {
+            assertEquals(HttpStatusCode.Forbidden, it.status)
+        }
+
+        client.get("/") {
+            header(HttpHeaders.Origin, "http://localhost:3000")
+        }.let {
+            assertEquals(HttpStatusCode.Forbidden, it.status)
+        }
```
bjhham, line 1031: Shouldn't this be...

    client.get("/") {
        header(HttpHeaders.Origin, "http://localhost:3000/")
    }.let {
        assertEquals(HttpStatusCode.OK, it.status)
    }

    client.get("/") {
        header(HttpHeaders.Origin, "http://localhost:3000")
    }.let {
        assertEquals(HttpStatusCode.OK, it.status)
    }

...?

bjhham (changes requested):

ktor-server/ktor-server-tests/jvmAndNix/test/io/ktor/tests/server/plugins/CORSTest.kt
anton-erofeev: in this test CORS does not contain any settings(`install(CORS){}`) and the client host is different from the server host([allowSameOrigin](https://github.com/ktorio/ktor/blob/main/ktor-server/ktor-server-plugins/ktor-server-cors/jvmAndNix/src/io/ktor/server/plugins/cors/CORS.kt#L170) does not apply here)
[this check](https://github.com/ktorio/ktor/blob/main/ktor-server/ktor-server-plugins/ktor-server-cors/jvmAndNix/src/io/ktor/server/plugins/cors/CORS.kt#L171) will not pass because this host is not contained in CORS config hosts and it should return [403 forbidden](https://github.com/ktorio/ktor/blob/main/ktor-server/ktor-server-plugins/ktor-server-cors/jvmAndNix/src/io/ktor/server/plugins/cors/CORSUtils.kt#L77)
Am I missing something?

bjhham: Apologies, I misread the issue and assumed the test provided in the comments was the intended behaviour.

bjhham (approved):

PR #4172, opened 2024-07-31
KTOR-7264 Revert migration of monitor property to fix backwards compatibility

bjhham:
**Subsystem**
Server

**Motivation**
[KTOR-7264](https://youtrack.jetbrains.com/issue/KTOR-7264) Backwards compatibility for monitor

This change breaks the Koin plugin, and anything else that would have relied on the monitor property. 

**Solution**
Moved monitor back to `ApplicationEnvironment` with a deprecation warning for direct access.


bjhham:
I'm thinking of adding a new test pipeline for any API file changes to verify external plugins as well.  Now that we have them registered, it should be much easier to avoid breaking them.

e5l (approved):

e5l:
Good catch, lgtm!

PR #4178, opened 2024-08-01
Use ByteArray pool for sending UDP datagrams on native

Thomas-Vos:
This was already done for JVM but not for native. Code is similar to JVM code.


e5l:
Hey @Thomas-Vos, could you try using direct read instead of pooling? So we will use the ByteArray directly from Buffer segment without making a copy.

For instance it can be done with `ByteReadChannel.read { array, startIndex, endIndex -> }`

Thomas-Vos:
@e5l Nice idea! Implemented something, what do you think? It does use internal IO API, not sure if that is alright. I could try to implement the same for JVM.

Thomas-Vos:
Implemented the same on JVM for sending.

Cannot figure out how to do it on the JVM for receiving as well, because `UnsafeBufferOperations.writeToTail` has a limit of `Segment.SIZE` (8192) which is too low for an entire UDP packet. Unless there is a `ByteBuffer` that wraps multiple `ByteArray`s?

e5l:
@Thomas-Vos, you need to do it in a multiple steps

Thomas-Vos:
@e5l if this is in reply to my JVM receive question: that won't work, as DatagramChannel.send expects a single ByteBuffer and cannot be done with multiple calls.

e5l:
Yep, and we need to make copy in this case

e5l:
@Thomas-Vos, could you tell me if it's ready for the review?

Thomas-Vos:
No, still running into an issue, currently working on it.

Thomas-Vos:
@e5l done.

Pools must be used for sending when datagram size is above 8192. For receiving always. Send/receive must be done in a single call for UDP so this is the only way I know this can be solved.

I added a test to test the new code when datagram size is above 8192. I noticed that this was already broken in existing Ktor code for both JVM and native, even though e.g. pools were used on JVM. Not sure if this was always the case or a recent issue. But now this is fixed.

Please let me know if you have any questions.

ktor-network/posix/src/io/ktor/network/sockets/DatagramSocketNative.kt
```diff
@@ -90,41 +91,32 @@ internal class DatagramSocketNative(
         val clientAddressLength: UIntVarOf<UInt> = alloc()
         clientAddressLength.value = sizeOf<sockaddr_storage>().convert()
 
-        val buffer = BytePacketBuilder()
-
-        try {
-            val count = buffer.write { memory, startIndex, endIndex ->
```
Thomas-Vos, line 96: buffer.write was limited to 8192 bytes, which is now fixed.

ktor-network/jvm/src/io/ktor/network/sockets/DatagramSendChannel.kt
```diff
@@ -146,7 +188,9 @@ private fun failInvokeOnClose(handler: ((cause: Throwable?) -> Unit)?) {
     throw IllegalStateException(message)
 }
 
-private fun Datagram.writeMessageTo(buffer: ByteBuffer) {
-    packet.readAvailable(buffer)
```
Thomas-Vos, line 150: readAvailable is limited to 8192 bytes, which is now fixed. A `readTo` function for ByteBuffer is missing, so implemented an alternative solution instead.

ktor-network/jvm/src/io/ktor/network/sockets/DatagramSendChannel.kt
```diff
@@ -48,41 +50,81 @@ internal class DatagramSendChannel(
...
+                        element.packet.discard()
+                    }
+                }
             }
         } finally {
             lock.unlock()
         }
 
-        if (result) {
-            element.packet.close()
```
Thomas-Vos, line 67: close actually does not recycle the head. Now fixed with `buffer.position(buffer.limit())` or `element.packet.discard()`.

ktor-network/jvm/src/io/ktor/network/sockets/DatagramSendChannel.kt
Thomas-Vos: Found the `readFully` function which does what is needed. (readFully also only read up to 8192 bytes which is fixed now)

Thomas-Vos:
The tests pass now on all platforms.

For some reason using repeat(10) instead of 4 makes the test time out on linux.

ktor-io/jvm/src/io/ktor/utils/io/core/ByteReadPacketExtensions.jvm.kt
```diff
@@ -18,7 +18,9 @@ public fun Source.readAvailable(buffer: ByteBuffer): Int {
 }
 
 public fun Source.readFully(buffer: ByteBuffer) {
-    readAtMostTo(buffer)
+    while (!exhausted()) {
+        readAtMostTo(buffer)
```
e5l, line 22: Please check the case when `buffer.remaining() == 0`, having a unit test would be great

e5l (changes requested):

ktor-io/jvm/src/io/ktor/utils/io/core/ByteReadPacketExtensions.jvm.kt
Thomas-Vos: @e5l thanks, fixed.

e5l (approved):

PR #4151, opened 2024-07-23
KTOR-4080 Support ktor-network on Windows

Thomas-Vos:
[KTOR-4080](https://youtrack.jetbrains.com/issue/KTOR-4080) Add support for mingw to ktor-network in order to make server and client work on windows

PR #4187, opened 2024-08-02
Add support for mingw(Windows) to all modules including client/server CIO engines

whyoleg:
After https://github.com/ktorio/ktor/pull/4151 it's rather easy to add support for `mingw` target to all other modules which depends on `ktor-network`, specifically server core and client/server CIO engines.

Unfortunately I don't have Windows PC at my side, so I can test only that it compiles.


ktor-server/ktor-server-core/mingwX64/test/BuildApplicationConfigNixTest.kt
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+import io.ktor.server.engine.*
+import platform.posix.*
+import kotlin.test.*
+
+class BuildApplicationConfigNixTest {
+
+//    @Test
+//    fun testPropertyConfig() {
```
whyoleg, line 12: The only thing that is missing it this test. There is no `getenv`/`setenv` for mingw out-of-the-box.
And without Windows PC I can't test it properly

ktor-server/ktor-server-core/nix/src/ServerEngineUtils.nix.kt
```diff
@@ -0,0 +1,15 @@
...
+
+package io.ktor.server.engine
+
+import kotlinx.cinterop.*
+import platform.posix.*
+
+@OptIn(UnsafeNumber::class, ExperimentalForeignApi::class)
+internal actual val WORKING_DIRECTORY_PATH: String get() = memScoped {
+    val result = allocArray<ByteVar>(512)
+    getcwd(result, 512.convert())
```
whyoleg, line 13: it's splitted per nix and mingwX64 because `getcwd` second argument is Int in place and UInt in other and commonizer don't support it

Thomas-Vos:
It seems Windows does not like connecting to `0.0.0.0`.

ktor-server/ktor-server-core/mingwX64/test/BuildApplicationConfigNixTest.kt
Thomas-Vos: So what about this? https://github.com/ktorio/ktor/blob/20cb0455457e86ea5cf01fb90ccaa945932317b2/ktor-server/ktor-server-config-yaml/posix/src/io/ktor/server/config/yaml/YamlConfigNix.kt#L75

whyoleg:
Applied the same fix for mingw as was done for jvm in https://github.com/ktorio/ktor/pull/3157
May be it makes sense to do this additionally on `ktor-network` side - I don't know

ktor-server/ktor-server-core/mingwX64/test/BuildApplicationConfigNixTest.kt
Thomas-Vos: Can you use `putenv`?

whyoleg: I found online that `putenv` doesn't copy pointer which is passed but just stores it.
On current moment it accepts `String`, this means K/N cinterop creates temporary pointer for the duration of a bridge call.
So I can try to use it, but can't test it, so may be someone else could do it :)

whyoleg: > So what about this?

Oh, sorry, there is no `setenv`/`unsetenv`...

Thomas-Vos: That sounds pretty bad, but I am not too familiar with this. Maybe https://learn.microsoft.com/en-us/windows/win32/api/processenv/nf-processenv-setenvironmentvariablew can be used.

whyoleg:
somehow I missed that #3157 is only about the logged message :)
So I've decided to apply the fix for `ktor-network` module.
The alternatives are:
1. fix tests to use `127.0.0.1` instead of `0.0.0.0`
2. fix ALL code to use `127.0.0.1` instead of `0.0.0.0` including default values used, like in `embeddedServer` - looks like this is the most correct way as per comment - https://youtrack.jetbrains.com/issue/KTOR-4834/Default-host-address-0.0.0.0-isnt-reachable-on-Windows#focus=Comments-27-6427178.0-0

I'm ready to adapt to any variant if needed

ktor-server/ktor-server-core/mingwX64/test/BuildApplicationConfigNixTest.kt
whyoleg: Thanks!
used `SetEnvironmentVariable` - let's see what CI will say (I absolutely love CI-driven development)

Thomas-Vos:
But only for connect calls, right? Because bind still needs to bind to 0.0.0.0, even on Windows.

whyoleg:
> Because bind still needs to bind to 0.0.0.0, even on Windows.

But, I see, that in your PR https://github.com/ktorio/ktor/pull/4151/files#diff-30b8c07ebfc0e269ea90abeed583bf7b6959388ee31c6f5205b7652a3d97a278 you have changed it to 127.0.0.1 everywhere, so I thought it apples to both :)

So, AFAIU, both 0.0.0.0 and 127.0.0.1 should work for bind `bind` and do the same thing, but for `connect` only 127.0.0.1 works, so we should fix 0.0.0.0 to mean 127.0.0.1
Or may be you do have other/better ideas? Also, if you have access to Windows PC it would be nice if you could checkout locally and check what will be the easiest/correct way - this will be very appreciated! Tank you for review BTW!

Thomas-Vos:
Ah, I see the confusion. I had trouble with `localhost` not working in the tests so I changed it so `127.0.0.1`. (it had something to do with localhost resolving to ipv6 which caused issues, unsure why, but that fixed it)

> So, AFAIU, both 0.0.0.0 and 127.0.0.1 should work for bind `bind` and do the same thing, but for `connect` only 127.0.0.1 works, so we should fix 0.0.0.0 to mean 127.0.0.1

Yes, so for connect, change 0.0.0.0 to 127.0.0.1. That seems right.

But for bind, both are valid, and there is a difference. Binding a TCP/UDP server to 0.0.0.0 will accept connections from anywhere (any host). But binding it to 127.0.0.1 will only accept connections from the same host.

On Windows, binding a TCP/UDP server to 0.0.0.0 works fine.

So it is important to not change the address for bind calls, but only for connect.


> Or may be you do have other/better ideas? Also, if you have access to Windows PC it would be nice if you could checkout locally and check what will be the easiest/correct way - this will be very appreciated! Tank you for review BTW!

I do have a Windows PC, but at the moment really busy with my own project, that's why I ported ktor-network to Windows :). Maybe I can look into it in the future. But maybe someone else with more networking experience on Windows can look into this?

whyoleg:
Thank you @Thomas-Vos!
This was really helpful! Looks like I need to rewind my knowledge about TCP/IP stack...
So, in the end I've stick with the solution to change only 2 IPs used in tests. The reasons are:
1. AFAIU after more research, we should not replace `0.0.0.0` with `127.0.0.1`, because this is just incorrect for Windows - there is no rerouting there like on other platforms. So overall, for localhost dev/tests `bind` and `connect` should use just `127.0.0.1` and everything should work fine. It's also possible to use `0.0.0.0` for `bind`, but then, client should still use `127.0.0.1` - not very convenient, but fine.
2. Keeping in mind this, I've adapted 2 tests, which `bind` socket to `0.0.0.0` and then use it `localAddress` property as address to `connect` to it; `localAddress` returns `0.0.0.0` in those cases. Now those tests just use `127.0.0.1` and so no problems there.

I do believe this should be fine, and tests will pass. If you have any other suggestions here, I will be ready to adapt!

e5l:
@whyoleg awesome job!
@Thomas-Vos, thank you for the deep check!

I will review in a moment :)

ktor-server/ktor-server-core/nix/src/ServerEngineUtils.nix.kt
e5l: we could make expect-actual for this, but it can be done as housekeeping

e5l (approved):
LGTM

ktor-server/ktor-server-core/nix/src/ServerEngineUtils.nix.kt
whyoleg: In PR with commonization of server I replaced it with `kotlinx.io.files` API in `common` code - [`SystemFileSystem.resolve(Path(".")).toString()`](https://github.com/ktorio/ktor/blob/5bd92a11beb8e775fc0de32f27d0e213c714f8ec/ktor-server/ktor-server-core/common/src/io/ktor/server/engine/ServerEngineUtils.kt)
Feel free to comment there on this, as I'm not fully sure, that it's safe to use `kotlinx.io.files` API in `ktor`

PR #4188, opened 2024-08-02
KTOR-3205 Switch from Application to Server

bjhham:
**Subsystem**
Server

**Motivation**
[KTOR-3205](https://youtrack.jetbrains.com/issue/KTOR-3205) Application / Server naming in Ktor, and server entities design

**Solution**
Replaced `Application` from practically everywhere, inserting type aliases for all public types to maintain backwards compatibility.


ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/engine/EmbeddedServer.kt
```diff
@@ -17,18 +17,18 @@ import kotlin.coroutines.*
...
 ) {
     /**
      * Provides events on Application lifecycle
      */
     public val monitor: Events
 
-    public val environment: ApplicationEnvironment
-    public val application: Application
+    public val environment: ServerEnvironment
+    public val server: Server
```
whyoleg, line 31: It feels a bit strange that `EmbeddedServer` has `server` property.
Also, what is difference then between `Server` and `EmbeddedServer` - it's not really clear from the documentation even

ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/application/PipelineCall.kt
```diff
@@ -15,31 +15,34 @@ import io.ktor.utils.io.*
...
      * An [PipelineResponse] that is a server response.
      */
-    public val response: ApplicationResponse
+    public val response: ServerResponse
 
     /**
      * An application being called.
      */
-    public val application: Application
+    public val server: Server
```
whyoleg, line 45: would be nice to have deprecated with `ReplaceWith` for renamed properties/functions, like this:
```suggestion
    public val server: Server
    @Deprecated("Renamed to server", ReplaceWith("server"))
    public val application: Server
```

whyoleg (commented):
I do have several questions regarding renaming:
1. client modules in ktor in most cases use `*HttpClient*` terminology, but after this, server modules will use `*Server*` without mentioning `Http` - probably it's fine and may be more words will be just too much, though it feels a bit inconsistent
2. there is still `io.ktor.server.application` package - do you have any plans to change it?
3. minor: some files were not renamed and still mention `Application` instead of `Server`

ktor-server/ktor-server-core/jvmAndNix/src/io/ktor/server/engine/EmbeddedServer.kt
bjhham: Yes, it was strange when it had an `application` property, and is now _very_ strange looking.  I think we could get rid of the field entirely, although this would involve some refactoring.

bjhham:
> client modules in ktor in most cases use *HttpClient* terminology, but after this, server modules will use *Server* without mentioning Http - probably it's fine and may be more words will be just too much, though it feels a bit inconsistent

This does seem a bit odd, I'm not sure what to do about it atm...

> there is still io.ktor.server.application package - do you have any plans to change it?

Yes, before proceeding with this I'll move the new types into the new package and leave the deprecated typealiases in the old package.

> minor: some files were not renamed and still mention Application instead of Server

I'll need to make a more thorough pass for this and the properties.  Right now I'm still trying to judge if this change is really worth it.  It's better semantically, but there's so many references it might be better to let this sleeping dog lie.

e5l:
@bjhham, we can choose `HttpServer` for consistency

e5l:
ATM it's hard to review, let's agree to have API dump and review source-breaking changes all together

PR #4190, opened 2024-08-04
Drop `assertFailsWith` hack as it's not used and not needed anymore

whyoleg:
* `assertFailsWith` is not used anymore in `common` code, but rather default `kotlin.test` implementation is used there, so looks like it's not needed anymore
* `assertFailsAndContainsCause` is not used at all


e5l (approved):

e5l:
@whyoleg, could you please rebase?

PR #4192, opened 2024-08-04
Commonize server modules and support JS and WasmJS targets

whyoleg:
**Note: this PR is based on https://github.com/ktorio/ktor/pull/4187 (support for windows in K/N server) and requires it to be merged first**

PR consists of 3 "parts":
1. first commits contain some general fixes to be compatible with JS/WasmJS in different shared modules
2. then goes commonization of `ktor-server-core` and `ktor-server-test-host`
3. then goes commonization of all server plugins which were compatible with K/Jvm + K/Native (every plugin/module commonization is in separate commit)

There is no yet server engine implementation for JS/WasmJS (except for that in `test-host`), but I would like to take a look if it's possible to somehow wrap `nodejs` HTTP server and run tests.

Current PR doesn't contain anything for WasmWasi target, as it needs more work (because we can't share what we have in JS) and need dependencies updates ([kotlinx.datetime is not supported yet](https://github.com/Kotlin/kotlinx-datetime/pull/366), [kotlinx.coroutines support is in RC](https://github.com/Kotlin/kotlinx.coroutines/releases)). I will probably take a look later.

There are some minor changes in public API/ABI (updated ABI dump is in last commit). `Platform` class changes could be changed to be backward compatible if needed.

I've tried to refactor as minimal code as possible to ease the review, but in several places there were small changes to the implementation which were required.

Reviewing PR probably will be easier per commit, but in most cases `commonization` can be described as:
* move sources/tests from `jvmAndPosix` to `common`
* rewrite tests which use old `withTestApplication` to `testApplication`

Additionally, because some tests were rewritten a lot (because of migration), GitHub doesn't always shows them as diff, but as a new file (because those were moved from `jvmAndPosix` to `common`).

There are some failures in JVM tests, but I don't really understand if those are flaky or there is something broken. No broken tests locally for Native, Js and Wasm. Could you take a look please?

**Notes/Questions regarding test future improvements:**
1: deprecated `withTestApplication` API will be not supported by JS/WasmJS server - only `testApplication` DSL is supported. WDYT about deprecating old API with `ERROR` in Ktor 3.0? There are some amount of tests, which still uses old API, but as seen in this PR migration is rather straightforward but manual.

2: in this PR I've tried to use `runTest` from `kotlinx.coroutines` as much as possible to be able to make tests more convenient. Still, `testSuspend` is stills used a lot in the wild. I've also introduced `runTestWithRealTime` to overcome issues that some tests don't work properly with delay skipping. WDYT about deprecating `testSuspend` in Ktor 3.0 and gradually migrating it to `runTest` or `runTestWithRealTime` depending on the use-case? In most cases `runTest` will be enough.

3: `testApplication` now uses `runTestWithRealTime` because some of the tests use `delay` or `withTimeout` which require real time. Though, I do think, that maybe it would be better to use just `runTest` there and adapt tests which need real time? WDYT?

4: there are some tests which require for the test server to stop, but because for JS/WasmJS we need single call in function (because of returned `Promise`), I've introduced a separate function `runTestApplication` which is `suspend` and is used under the hood in `testApplication` (which now just a combination of `runTest` + `runTestApplication`). But, I don't really like the naming here comparing to `runTest`.
```kotlin
// just basic suspend test
fun tests() = runTest {}
// how to write test for server
fun test() = testApplication {} 
// using `runTestApplication`
fun test() = runTest {
  runTestApplication {}
}
```
As you see, it's looks a bid odd and I would propose to may be introduce `runTestApplication` as an alias to `testApplication` deprecating the latter and find better name for current `runTestApplication`, WDYT?

5: Also, as `ktor-server-test-base` is now used almost exclusively with `ktor-server-test-suites`, WDYT about merging them together?

There are not a lot of `not-common` modules left (those which doesn't support all Kotlin targets supported by ktor), they are out of scope here:
* ktor-network and ktor-network-tls - needs sockets for JS/WasmJS
* ktor-server-cio and ktor-client-cio - needs `ktor-network`
* ktor-server-config-yaml - needs `yamlkt` dependency


ktor-client/ktor-client-plugins/ktor-client-call-id/build.gradle.kts
```diff
@@ -11,11 +11,15 @@ kotlin {
...
             dependencies {
                 api(project(":ktor-server:ktor-server-test-host"))
                 api(project(":ktor-server:ktor-server-plugins:ktor-server-call-id"))
             }
         }
     }
 }
+
+// tests need server, so can't be run in browser
+tasks.named("jsBrowserTest") { enabled = false }
```
e5l, line 24: Consider using `onlyIf { false }`, cause it can be reenabled with `enabled`

ktor-io/wasmJs/src/io/ktor/utils/io/charsets/Encoding.wasmJs.kt
```diff
@@ -14,7 +14,7 @@ public actual abstract class Charset(internal val _name: String) {
 
     actual override fun equals(other: Any?): Boolean {
         if (this === other) return true
-        return other is Charset && (_name != other._name)
+        return other is Charset && (_name == other._name)
```
e5l, line 17: good catch

ktor-server/ktor-server-core/common/src/io/ktor/server/engine/CommandLine.kt
```diff
@@ -104,7 +104,7 @@ internal fun buildApplicationConfig(args: List<Pair<String, String>>): Applicati
     val configPaths = args.filter { it.first == "-config" }.map { it.second }
 
     val commandLineConfig = MapApplicationConfig(commandLineProperties)
-    val environmentConfig = getConfigFromEnvironment()
+    val environmentConfig = MapApplicationConfig(getKtorEnvironmentProperties())
```
e5l, line 107: We could have an mpp library to work with Env variables :)

ktor-server/ktor-server-core/common/test/io/ktor/server/application/ApplicationPluginTest.kt
```diff
@@ -0,0 +1,331 @@
...
+import io.ktor.server.testing.*
+import io.ktor.util.*
+import io.ktor.utils.io.*
+import io.ktor.utils.io.core.*
+import kotlinx.coroutines.test.*
+import kotlin.test.*
+
+class ApplicationPluginTest {
+    @Test
+    fun test_empty_plugin_does_not_break_pipeline() = testApplication {
```
e5l, line 21: Could you tell me the reason for `_` instead of camel case?

ktor-server/ktor-server-core/jvm/src/io/ktor/server/engine/EnvironmentUtilsJvm.kt
```diff
@@ -73,7 +73,24 @@ internal actual fun ApplicationEnvironmentBuilder.configurePlatformProperties(ar
         ?: ApplicationEnvironment::class.java.classLoader
 }
 
-internal actual fun getConfigFromEnvironment(): ApplicationConfig = System.getProperties()
-    .toMap()
-    .filterKeys { (it as String).startsWith("ktor.") }
-    .let { env -> MapApplicationConfig(env.map { it.key as String to it.value as String }) }
+internal actual fun getKtorEnvironmentProperties(): List<Pair<String, String>> = buildList {
```
e5l, line 76: Environment Properties :D

e5l: Could you add a note in KDoc for the common declaration of the values used on what platforms? :)

e5l (approved):
Hey @whyoleg! That was a huge pr, thanks for migrating test API as well!

 Could you check small comments before we can merge?

ktor-server/ktor-server-core/common/src/io/ktor/server/engine/CommandLine.kt
whyoleg: Yeah, I was thinking about extracting it into `ktor-environment`.
Similar libraries:
* https://github.com/ScottPierce/kotlin-env-var
* https://github.com/luca992/getenv-kt
Though, they support less, than already available here.
Looks good as an issue in YT may be

ktor-server/ktor-server-core/common/test/io/ktor/server/application/ApplicationPluginTest.kt
whyoleg: ATM it was easier to migrate from ` `(space) to `-`(dash) automatically - if you have any idea on how to automate the migration to camel case, I would be glad to hear

ktor-server/ktor-server-core/jvm/src/io/ktor/server/engine/EnvironmentUtilsJvm.kt
whyoleg: It's already there: https://github.com/ktorio/ktor/blob/ca0de5a707d578bdc6be4d3d927ac1a95e4d521d/ktor-server/ktor-server-core/common/src/io/ktor/server/engine/CommandLine.kt#L129-L131

whyoleg:
@e5l FYI, I've created separate YT issues for all points mentioned above:
* https://youtrack.jetbrains.com/issue/KTOR-7284/Deprecate-withTestApplication-APIs-with-ERROR-and-migrate-to-testApplication
* https://youtrack.jetbrains.com/issue/KTOR-7285/Deprecate-testSuspend-in-favour-of-kotlinx.coroutines.test.runTest
* https://youtrack.jetbrains.com/issue/KTOR-7286/Decide-on-how-testApplication-should-behave-regarding-virtual-time
* https://youtrack.jetbrains.com/issue/KTOR-7287/Decide-on-naming-of-server-testing-API
* https://youtrack.jetbrains.com/issue/KTOR-7288/Introduce-shared-ktor-environment-ktor-config-module
* https://youtrack.jetbrains.com/issue/KTOR-7289/Support-androidNative-targets
* https://youtrack.jetbrains.com/issue/KTOR-7290/Support-wasmWasi-target

As for node server, the issue already exists: https://youtrack.jetbrains.com/issue/KTOR-865/Add-node.js-server-engine

e5l:
thank you!

lppedd:
@whyoleg KTOR-865 is private btw, can't access it.

whyoleg:
Oops, missed this, but I think it's fine also to make it public, probably it's by accident, @e5l, WDYT?
Also, I've found another public issue: https://youtrack.jetbrains.com/issue/KTOR-450/Nodejs-express-server-engine-implementation - it has stars/likes, so probably it's better to track it there and just close KTOR-865

lppedd:
@whyoleg I'd just avoid tying Node to Express. Better keep the issue where no references to a framework are present.

whyoleg:
We can rename it to be framework agnostic probably. I just don't think that we need to have 2 different issues for the same idea as most likely there is no need to have both of them. Node APIs should be enough to create an engine and the engine is just an implementation detail and minimally exposed to ktor users. Also most of the comments in the later issue is just about having possibility to run server on JS, not express specifically.

PR #4201, opened 2024-08-08
KTOR-7243 Support reloading of zip files in StaticContent

marychatte:
**Subsystem**
Server, StaticContent

**Motivation**
[KTOR-7243](https://youtrack.jetbrains.com/issue/KTOR-7243)

**Solution**
Add a flag `reloadOnRequest`, and if it is set to true, then reload the zip every time a request comes

I was also thinking about [`WatchService`](https://docs.oracle.com/javase%2F8%2Fdocs%2Fapi%2F%2F/java/nio/file/WatchService.html) for `FileSystem`, but it's not really obvious for me, because on request we can try to get and go through incoming `WatchEvents`, we can keep map of  modified/deleted/created `WatchEvents` and reload `file` or entire `zip` if it was changed. But maybe a variant with reloading zip on request is okay to go. It's much simpler and more obvious

### DIFF
--- a/ktor-server/ktor-server-core/jvm/src/io/ktor/server/http/content/StaticContent.kt
+++ b/ktor-server/ktor-server-core/jvm/src/io/ktor/server/http/content/StaticContent.kt
@@ -8,6 +8,7 @@ import io.ktor.http.*
 import io.ktor.http.content.*
 import io.ktor.server.application.*
 import io.ktor.server.application.hooks.*
+import io.ktor.server.http.content.FileSystemPaths.Companion.paths
 import io.ktor.server.response.*
 import io.ktor.server.routing.*
 import io.ktor.util.*
@@ -254,7 +255,7 @@ public fun Route.staticZip(
     remotePath = remotePath,
     basePath = basePath,
     index = index,
-    fileSystem = FileSystems.newFileSystem(zip, environment.classLoader),
+    fileSystem = FileSystems.newFileSystem(zip, environment.classLoader).paths(),
     block = block
 )
 
@@ -272,7 +273,7 @@ public fun Route.staticFileSystem(
     remotePath: String,
     basePath: String?,
     index: String? = "index.html",
-    fileSystem: FileSystem = FileSystems.getDefault(),
+    fileSystem: FileSystemPaths = FileSystems.getDefault().paths(),
     block: StaticContentConfig<Path>.() -> Unit = {}
 ): Route {
     val staticRoute = StaticContentConfig<Path>().apply(block)
@@ -559,7 +560,7 @@ private suspend fun ApplicationCall.respondStaticFile(
 }
 
 private suspend fun ApplicationCall.respondStaticPath(
-    fileSystem: FileSystem,
+    fileSystem: FileSystemPaths,
     index: String?,
     basePath: String?,
     compressedTypes: List<CompressedFileType>?,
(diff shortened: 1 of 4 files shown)


bjhham (changes requested):
So I think approach might not be so great for performance and maintainability.  It ties us to auditing every call to a complex type `FileSystem` in the downstream code, because the return values could be stale or it might be non-performant with the reload.

I'd stick with the original plan of using the `WatchService` to refresh on changes, but from what I can tell, we only use `getPath()` in the downstream code, so it'd be better to introduce an interface for a more specific delegation.

So you can replace all the FileSystem references in the downstream code to something like this:

```kotlin
interface FileSystemPaths {
    companion object {
        fun FileSystem.paths() = object : FileSystemPaths {
            override fun getPath(filePath: String): Path = this@paths.getPath(filePath)
        }
    }

    fun getPath(filePath: String): Path
}
```

Then kick off a coroutine for handling updates like so in the delegator:

```kotlin
class ReloadingZipFileSystem(
    zip: Path,
    classLoader: ClassLoader,
    private var delegate: FileSystemPaths
) : FileSystemPaths by delegate, CoroutineScope {
    init {
        val watchService = FileSystems.getDefault().newWatchService()
        zip.parent.register(
            watchService,
            StandardWatchEventKinds.ENTRY_CREATE,
            StandardWatchEventKinds.ENTRY_DELETE,
            StandardWatchEventKinds.ENTRY_MODIFY
        )

        launch(Dispatchers.IO) {
            var key: WatchKey
            while ((watchService.take().also { key = it }) != null) {
                if (key.pollEvents().isNotEmpty())
                    delegate = FileSystems.newFileSystem(zip, classLoader).paths()
                key.reset()
            }
        }
    }
}
```

ktor-server/ktor-server-tests/jvm/test/io/ktor/server/plugins/StaticContentTest.kt
```diff
@@ -1273,6 +1275,56 @@ class StaticContentTest {
...
+
+        val firstResponse = client.get("static/$firstFileName")
+        assertEquals(HttpStatusCode.OK, firstResponse.status)
+        assertEquals(firstContent, firstResponse.bodyAsText())
+
+        firstZipFile.delete()
+        val secondZipFile = createZipFile(path, secondFileName, secondContent)
+
+        // there is a delay in the file system watcher about 2 seconds
+        delay(3000)
```
bjhham, line 1305: Maybe to avoid flakiness here, you could introduce another file system watcher in the test with a mutex to block on, so when the test FS watcher triggers we can assume that the zip watcher has triggered.

ktor-server/ktor-server-core/jvm/src/io/ktor/server/http/content/StaticContent.kt
```diff
@@ -249,14 +252,61 @@ public fun Route.staticZip(
     basePath: String?,
     zip: Path,
     index: String? = "index.html",
+    reloadOnRequest: Boolean = false,
```
bjhham, line 255: Maybe we can avoid introducing this parameter and instead always use the auto-reloading?  I don't think it will have any significant drawbacks and should make things a lot easier.

ktor-server/ktor-server-tests/jvm/test/io/ktor/server/plugins/StaticContentTest.kt
marychatte: Good idea, thanks

bjhham: Ah, I was hoping the delay could be removed entirely, I didn't realize it was 2 second polling behind the scenes.  Would be nice if we could avoid introducing long tests like this, but I don't see any way around it here.

bjhham (approved):

PR #4202, opened 2024-08-08
KTOR-7299 Fix "Fail to select descriptor…" error on descriptor close

Thomas-Vos:
[KTOR-7299](https://youtrack.jetbrains.com/issue/KTOR-7299) IOException: Fail to select descriptor 134 for ACCEPT

This PR resolves the issue in test `testDescriptorErrorDoesNotFailOtherSockets()`.

When one of the descriptors is closed, ALL descriptors managed by the selector would throw the "Fail to select descriptor" error. That happened because `errorSet` is not cleared by the select call on failure, the docs say the sets are "undefined".

This is quite a bit issue, because if one socket would fail, all (unrelated) sockets of the selector manager would fail.

The other test `testDescriptorError()` already passed, but added it to verify the behaviour.


Thomas-Vos:
The tests are in nix instead of posix as the Windows selector code does not seem to wake up on selector close (I checked).

ktor-network/nix/src/io/ktor/network/selector/SelectUtilsNix.kt
```diff
@@ -102,7 +102,15 @@ internal actual class SelectorHelper {
...
-                    // Thrown if the descriptor was closed.
+                    // Thrown if any of the descriptors was closed.
+                    // This means the sets are undefined so do not rely on their contents.
+                    watchSet.forEach { event ->
+                        if (!isDescriptorValid(event.descriptor)) {
+                            event.fail(IOException("Bad descriptor ${event.descriptor} for ${event.interest}"))
+                            watchSet.remove(event)
+                        }
+                    }
+                    continue
```
Thomas-Vos, line 113: This is the important part, make sure processSelectedEvents is not called on failure.

e5l (approved):

PR #4203, opened 2024-08-09
Fix IndexOutOfBoundsException in mingwX64 selector code

Thomas-Vos:
Found an issue in the mingw selector code.

When a single descriptor would suspend for both reading and writing, it would cause an `IndexOutOfBoundsException`.

Not really sure how to make a test that suspends on both reading and writing.

### DIFF
--- a/ktor-network/windows/src/io/ktor/network/selector/SelectUtilsWindows.kt
+++ b/ktor-network/windows/src/io/ktor/network/selector/SelectUtilsWindows.kt
@@ -7,6 +7,7 @@ package io.ktor.network.selector
 import io.ktor.network.util.*
 import io.ktor.util.collections.*
 import io.ktor.utils.io.*
+import io.ktor.utils.io.errors.*
 import kotlinx.cinterop.*
 import kotlinx.coroutines.*
 import platform.posix.*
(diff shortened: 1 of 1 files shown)


ktor-network/windows/src/io/ktor/network/selector/SelectUtilsWindows.kt
```diff
@@ -138,19 +138,20 @@ internal actual class SelectorHelper {
...
         }
 
-        watchSet.forEachIndexed { index, event ->
+        watchSet.forEach { event ->
             if (event.descriptor in closeSet) {
                 completed.add(event)
-                return@forEachIndexed
+                return@forEach
             }
-            val wsaEvent = wsaEvents[index]
```
Thomas-Vos, line 153: This caused the `IndexOutOfBoundsException` as the size of `wsaEvents` does not match the `watchSet` size when a single descriptor is added multiple times (when suspended on read and write). Changed it to a Map with the descriptor as key. Apologies for the issue!

Thomas-Vos:
Also added some result checks for `WSACreateEvent` and `WSACreateEvent`.

e5l (approved):
Good job, thank you! Merged

PR #4207, opened 2024-08-09
KTOR-7303 Remove read/write channels for UDP sockets

Thomas-Vos:
[KTOR-7303](https://youtrack.jetbrains.com/issue/KTOR-7303) Remove read/write channels for UDP sockets

I didn't remove the `attachForReading`/`attachForWriting` functions from the internal NIOSocketImpl. Is that alright or should they be moved to `SocketImpl`?


e5l (approved):
lgtm as discussed

PR #4209, opened 2024-08-12
KTOR-7294: allow for exp. retry delays of below 2 seconds

easimon:
**Subsystem**
Http Client, retries with exponential backoff

**Motivation**
The exponential backoff has two issues

- The "base delay" multiplier is fixed to 1 second
- The initial delay is already `base^1 * 1 second`, resulting in 2 seconds for the first delay with the default base of 2.0

While you could set a non-default base of e.g. 1.1, this flattens mainly the exponential increase, and still does not allow a initial delay < 1 second. Setting the base to `0 < delay < 1` would even lead to decreasing delays. This is at least counter-intutive. 

**Solution**

- Allow the delay multiplier to be set to arbitrary millisecond values
- Fix the initial delay to not be `base ^ retry`, but `base ^(retry - 1)`

Examples: 

`base = 2, baseDelay = 1000` produces delays of 1000, 2000, 4000, 8000 ... ms
`base = 2, baseDelay = 100` produces delays of 100, 200, 400, 800 ... ms
`base = 1.1, baseDelay = 100` produces delays of 100, 110, 121, 133.1 ... ms


marychatte (approved):
Thanks a lot for the PR! 
I just run `apiDump` check and will merge the PR after CI checks

PR #4210, opened 2024-08-12
Refactor CORS number regex

anton-erofeev:
**Subsystem**
ktor-server-cors

**Motivation**
numberRegex is passed in several methods in CORS.kt, however it is only used in CORSutils.kt and the value does not change. Therefore it can be moved to CORSutils.kt

**Solution**
removed numberRegex from method arguments and moved it to CORSutils.kt


bjhham (approved):
Good call, thanks 👍

PR #4211, opened 2024-08-12
Prioritize explicit headers

gimlet2:
**Subsystem**
Client common engine

**Motivation**
I've had an issue sending HTTP request with form data. A server was rejecting the request based on content-type. It was expecting `application/x-www-form-urlencoded` but [FormDataContent](https://github.com/ktorio/ktor/blob/80b68afb5a5133e1909299fc4d1bf81645dd3bd4/ktor-client/ktor-client-core/common/src/io/ktor/client/request/forms/FormDataContent.kt#L30) adds also an encoding. 

**Solution**
Explicitly provided headers should have higher priority then calculated.

### DIFF
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/engine/Utils.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/engine/Utils.kt
@@ -36,8 +36,8 @@ public fun mergeHeaders(
     block: (key: String, value: String) -> Unit
 ) {
     buildHeaders {
-        appendAll(requestHeaders)
         appendAll(content.headers)
+        appendAll(requestHeaders) 
     }.forEach { key, values ->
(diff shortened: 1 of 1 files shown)


e5l:
@bjhham, could you check?

bjhham (changes requested):
Yeah, this makes a lot of sense, thanks!

Could you please add a test in `ktor-client-core` somewhere to enforce this logic?

gimlet2:
@bjhham yes, I would like to. Can you please give me some guidance on what would be the best place to add it? I was trying to find some good spot for it but wasn't able to find any existing test for that functionality.

bjhham:
@gimlet2 sorry, I meant to say `ktor-client-tests`, the best location would be `io.ktor.client.tests.ClientHeadersTest` under the common test source path.

gimlet2:
That changes doesn't solve the original issue

PR #4213, opened 2024-08-14
KTOR-7265 Fix for ContentEncoding deadlock

bjhham:
**Subsystem**
Client, ContentEncoding, SaveBody

**Motivation**
- [KTOR-7265](https://youtrack.jetbrains.com/issue/KTOR-7265) ContentEncoding: request hangs when using 3.0.0-beta-2

**Solution**
The problem would arise when the `SaveBody` plugin's `ByteChannelReplay` would be called by two consumers and the first wouldn't read the full response.  This would cause the 
deferred result to never complete, which would make subsequent calls hang indefinitely.

The simplest solution here is to just cancel the writer on the first caller when the second tries to read.  This allows the stream to finish copying without relying on being read by the first caller.

This solution has the caveat that you can run into problems when you try accessing the response with multiple threads.  I attempted to do a much more extensive change to always do the copying up-front so we can rely on some thread safety, but this causes some other problems where streaming is expected, so I'll defer that work for a later time.

### DIFF
--- a/buildSrc/src/main/kotlin/test/server/tests/Encoding.kt
+++ b/buildSrc/src/main/kotlin/test/server/tests/Encoding.kt
@@ -63,6 +63,16 @@ internal fun Application.encodingTestServer() {
                 }
                 setCompressionEndpoints()
             }
+            route("/big-plain-text") {
+                get {
+                    call.respondText {
+                        buildString {
+                            for (i in 1..10_000)
+                                appendLine("I will not introduce deadlocks.")
+                        }
+                    }
+                }
+            }
         }
     }
 }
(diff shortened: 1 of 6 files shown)


buildSrc/src/main/kotlin/test/server/tests/Encoding.kt
```diff
@@ -63,6 +63,16 @@ internal fun Application.encodingTestServer() {
                 }
                 setCompressionEndpoints()
             }
+            route("/big-plain-text") {
+                get {
+                    call.respondText {
+                        buildString {
+                            for (i in 1..10_000)
+                                appendLine("I will not introduce deadlocks.")
```
e5l, line 71: :D

e5l (approved):

PR #4217, opened 2024-08-15
KTOR-6412: add AcceptEncoding constants

anton-erofeev:
**Subsystem**
http

**Motivation**
https://youtrack.jetbrains.com/issue/KTOR-6412

**Solution**
Added Accept Encoding constants


e5l:
@marychatte, could you please check?

marychatte (changes requested):
Thanks for PR!
It's a good idea to add constants! Maybe we can use `HeaderValueWithParameters` as it's done already with [`ContentType`](https://github.com/ktorio/ktor/blob/1eeb7c11c93ad1cce550c71df1a189d22c74148f/ktor-http/common/src/io/ktor/http/ContentTypes.kt#L14). Because it is also can be a parameter [;q=](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Encoding#q) (qvalues weighting). And please add Kdoc and some tests

anton-erofeev:
> Thanks for PR! It's a good idea to add constants! Maybe we can use `HeaderValueWithParameters` as it's done already with [`ContentType`](https://github.com/ktorio/ktor/blob/1eeb7c11c93ad1cce550c71df1a189d22c74148f/ktor-http/common/src/io/ktor/http/ContentTypes.kt#L14). Because it is also can be a parameter [;q=](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Encoding#q) (qvalues weighting). And please add Kdoc and some tests

Hi @marychatte!
I updated the PR with review points fixes, could you please take a look?

marychatte (commented):
@anton-erofeev Good work on the PR! Thank you a lot
I added a few comments, but in general I think everything is good!

PR #4221, opened 2024-08-16
Fix formatting

bjhham:
**Subsystem**
Client + Server

**Motivation**
The formatting check has been failing for a while and I need a win today 😭 

**Solution**
I ran the `formatKotlin` gradle target


marychatte (approved):
win🏆

PR #4225, opened 2024-08-16
KTOR-5085: Add multiplatform support for XML serialization

osipxd:
**Subsystem**
`ktor-serialization-kotlinx-xml`

**Motivation**
https://youtrack.jetbrains.com/issue/KTOR-5085/Add-multiplatform-support-for-XML

**Solution**
Moved sources and some tests from `jvm` to `common` source set.

Docs change: https://github.com/ktorio/ktor-documentation/pull/513

### DIFF
--- a/ktor-shared/ktor-serialization/ktor-serialization-kotlinx/ktor-serialization-kotlinx-xml/build.gradle.kts
+++ b/ktor-shared/ktor-serialization/ktor-serialization-kotlinx/ktor-serialization-kotlinx-xml/build.gradle.kts
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2021 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ * Copyright 2014-2024 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
  */
 
 plugins {
@@ -7,20 +7,20 @@ plugins {
 }
 
 kotlin.sourceSets {
-    jvmMain {
+    commonMain {
         dependencies {
             api(project(":ktor-shared:ktor-serialization:ktor-serialization-kotlinx"))
             api(libs.xmlutil.serialization)
         }
     }
+    commonTest {
+        dependencies {
+            implementation(project(":ktor-shared:ktor-serialization:ktor-serialization-kotlinx:ktor-serialization-kotlinx-tests")) // ktlint-disable max-line-length
+        }
+    }
     jvmTest {
         dependencies {
-            api(project(":ktor-client:ktor-client-tests"))
-            api(project(":ktor-server:ktor-server-test-host"))
-            api(project(":ktor-client:ktor-client-plugins:ktor-client-content-negotiation:ktor-client-content-negotiation-tests")) // ktlint-disable max-line-length
-            api(project(":ktor-shared:ktor-serialization:ktor-serialization-kotlinx:ktor-serialization-kotlinx-tests"))
-
-            api(libs.logback.classic)
+            implementation(project(":ktor-client:ktor-client-plugins:ktor-client-content-negotiation:ktor-client-content-negotiation-tests")) // ktlint-disable max-line-length
         }
     }
 }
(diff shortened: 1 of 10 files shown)


ktor-shared/ktor-serialization/ktor-serialization-kotlinx/ktor-serialization-kotlinx-xml/build.gradle.kts
```diff
@@ -1,26 +1,26 @@
...
+    }
     jvmTest {
         dependencies {
-            api(project(":ktor-client:ktor-client-tests"))
-            api(project(":ktor-server:ktor-server-test-host"))
-            api(project(":ktor-client:ktor-client-plugins:ktor-client-content-negotiation:ktor-client-content-negotiation-tests")) // ktlint-disable max-line-length
-            api(project(":ktor-shared:ktor-serialization:ktor-serialization-kotlinx:ktor-serialization-kotlinx-tests"))
-
-            api(libs.logback.classic)
+            implementation(project(":ktor-client:ktor-client-plugins:ktor-client-content-negotiation:ktor-client-content-negotiation-tests")) // ktlint-disable max-line-length
```
osipxd, line 23: Seems like configuration `implementation` fits better than `api` here, so I replaced it.

ktor-shared/ktor-serialization/ktor-serialization-kotlinx/ktor-serialization-kotlinx-xml/common/src/io/ktor/serialization/kotlinx/xml/XmlSupport.kt
```diff
@@ -21,13 +21,16 @@ import nl.adaptivity.xmlutil.serialization.*
...
     repairNamespaces = true
+    // TODO: Remove Suppress and OptIn after update to xmlutil 0.90.2+ or Kotlin 2.0.20+
+    //  https://github.com/pdvrieze/xmlutil/issues/219
+    //  https://youtrack.jetbrains.com/issue/KT-69182/K2-OptIn-on-enum-companion-blocks-enum-constants
+    @Suppress("UnnecessaryOptInAnnotation")
+    @OptIn(XmlUtilInternal::class)
     xmlDeclMode = XmlDeclMode.None
     indentString = ""
     autoPolymorphic = false
-    this.xmlDeclMode
```
osipxd, line 30: Wasn't able to track down the reason why this line is here. If there is the reason, please let me know.

e5l:
@marychatte, could you check?

ktor-shared/ktor-serialization/ktor-serialization-kotlinx/ktor-serialization-kotlinx-xml/common/src/io/ktor/serialization/kotlinx/xml/XmlSupport.kt
marychatte: From what I see it's a redundant line, and it does nothing

marychatte (changes requested):
Thanks a lot for the PR!
I just noticed some important things:
1. `gradle build` fails with the next:
```
Execution failed for task ':ktor-shared:ktor-serialization:ktor-serialization-kotlinx:ktor-serialization-kotlinx-xml:transformCommonMainDependenciesMetadata'. 
...
No matching variant of io.github.pdvrieze.xmlutil:serialization:0.90.1 was found. The consumer was configured to find a library for use during 'kotlin-metadata', preferably optimized for non-jvm, as well as attribute 'org.jetbrains.kotlin.native.target' with value 'watchos_device_arm64', attribute 'org.jetbrains.kotlin.platform.type' with value 'native' but:
....
- Incompatible because this component declares a component, as well as attribute 'org.jetbrains.kotlin.native.target' with value 'watchos_arm64' and the consumer needed a component, as well as attribute 'org.jetbrains.kotlin.native.target' with value 'watchos_device_arm64'
So I can see from the output that it found watchos_x64, watchos_simulator_arm64  but not watchos_device_arm64
```
So I can see from the output that it found `watchos_x64`, `watchos_simulator_arm64`  but not `watchos_device_arm64`
2. There is also falling test `XmlSerializationTest` for `ios`, `macos` e.t.c.
Because it's a one more whitespace: `Expected <<User id="1" login="vasya"/>>, actual <<User id="1" login="vasya" />>`

osipxd:
@marychatte, thank you for your review!

> 1. `gradle build` fails

Unfortunately, I have only Linux x64 machine at the moment and all apple targets are ignored with message "", so I cannot reproduce this problem.

> 2. There is also falling test `XmlSerializationTest`

Reproduced and fixed. Also, found out that `[js, node]` tests were failing because of [this issue](https://github.com/pdvrieze/xmlutil/issues/83) so I disabled them for now. 

[Here is the correct diff link](https://github.com/ktorio/ktor/compare/c87db9979..b551c29a6) to review the changes.

marychatte (approved):
Great work!

PR #4231, opened 2024-08-20
KTOR-7324 Fix ByteChannel busy loop

bjhham:
**Subsystem**
Client + Server, Core I/O

**Motivation**
[KTOR-7324](https://youtrack.jetbrains.com/issue/KTOR-7324) Infinite creation of CancellableContinuationImpl on ByteChannel.flush

Any time we have two threads trying to read from an empty ByteChannel, we'll encounter a busy loop where the two threads will create unlimited CancellableContinuationImpl instances.

It happens in this loop:

while (flushBufferSize >= CHANNEL_MAX_SIZE) {
    suspendCancellableCoroutine {
        trySuspendSlot(it) { flushBufferSize >= CHANNEL_MAX_SIZE }
    }
}

The trySuspendSlot will resume the other thread and vice versa, effectively putting them both into a busy wait until something is written to the channel.

**Solution**
Throw an error when we have two sleep invocations of the same type (read or write).  This will prevent the busy loop scenario and instead throw an error, which should at least help us flush out where these problems are coming from.


ktor-io/common/src/io/ktor/utils/io/ByteChannel.kt
```diff
@@ -133,65 +135,99 @@ public class ByteChannel(public val autoFlush: Boolean = false) : ByteReadChanne
...
 
     @OptIn(InternalAPI::class)
     private suspend fun sleepForWrite() {
-        while (flushBufferSize >= CHANNEL_MAX_SIZE) {
+        while (flushBufferSize >= CHANNEL_MAX_SIZE && _closedCause.value == null) {
             suspendCancellableCoroutine {
-//                writeContinuation.init(it.intercepted())
-                trySuspendSlot(it) { flushBufferSize >= CHANNEL_MAX_SIZE }
-//                writeContinuation.getOrThrow()
+                trySuspendSlot(Slot.Write(it)) { flushBufferSize >= CHANNEL_MAX_SIZE && _closedCause.value == null }
```
e5l, line 150: if slots show up in the profile we can reuse them within the call

ktor-io/common/src/io/ktor/utils/io/ByteChannel.kt
```diff
@@ -133,65 +135,99 @@ public class ByteChannel(public val autoFlush: Boolean = false) : ByteReadChanne
...
+            continuation.resume()
+        }
+    }
+
+    private inline fun <reified TaskType : Slot> trySuspendSlot(
+        slot: TaskType,
+        crossinline sleepCondition: () -> Boolean
+    ) {
+        // When the slot is full, try resuming the current task
+        if (!suspensionSlot.compareAndSet(Slot.Empty, slot)) {
```
e5l, line 191: I like the idea of replacing null with a symbol

ktor-io/common/src/io/ktor/utils/io/ByteChannel.kt
```diff
@@ -133,65 +135,99 @@ public class ByteChannel(public val autoFlush: Boolean = false) : ByteReadChanne
...
+        data object Empty : Slot()
+        data class Closed(val cause: Throwable?) : Slot()
+        abstract class TaskSlot : Slot() {
+            abstract val continuation: Continuation<Unit>
+            override fun resume() =
+                continuation.resume(Unit)
+            override fun resumeWithException(throwable: Throwable) =
+                continuation.resumeWithException(throwable)
         }
+        data class Read(override val continuation: Continuation<Unit>) : TaskSlot()
```
e5l, line 230: Could you tell me why it should be a data class and if we need inheritance?

e5l (changes requested):
Good job on this! Could you check minor comments before merging?

e5l:
@marychatte, could you also check?

ktor-io/common/src/io/ktor/utils/io/ByteChannel.kt
bjhham: Oh, there's no reason.  I originally had duplication then refactored and didn't drop the data.  I'll drop the data.

ktor-io/common/src/io/ktor/utils/io/ByteChannel.kt
bjhham: There's just a small increase in continuation allocations when I run the benchmark test.  There should be less... I'll see if I can bring it down...

ktor-io/common/src/io/ktor/utils/io/ByteChannel.kt
```diff
@@ -4,32 +4,35 @@
...
 import kotlinx.coroutines.*
 import kotlinx.io.*
-import kotlin.concurrent.*
+import kotlin.concurrent.Volatile
 import kotlin.coroutines.*
+import kotlin.jvm.*
 
 @InternalAPI
-public val CHANNEL_MAX_SIZE: Int = 4096
+public const val CHANNEL_MAX_SIZE: Int = 1024 * 1024
```
bjhham, line 16: So wrapping the continuation with the local type increases the memory allocations when waiting for I/O a little bit, although we can avoid a lot of suspension just by increasing our buffer size here - this also seems to increase performance a bit too.  From what I can tell we don't read our files fully into memory anywhere, so moving the flush to the end would be problematic with large files.

Since the underlying buffer segments are recycled, I think increasing our general buffer size will generally (and counterintuitively) reduce memory overhead.

e5l: lgtm

e5l (commented):
LGTM! It looks like CI fail is related to change, let's check the reason before merging

e5l (approved):

PR #4235, opened 2024-08-21
KTOR-7323 Drop marker interface requirements in Ktor auth

bjhham:
**Subsystem**
Server, Authentication

**Motivation**
[KTOR-7323](https://youtrack.jetbrains.com/issue/KTOR-7323) Drop marker interface requirements in Ktor auth

These interfaces don't do anything and they prevent users from using decoupled domain objects for their principal and credential types.

**Solution**
Removed the type boundaries and deprecated the interfaces.


morki:
This KDoc now does not look right:

```
* Retrieves an authenticated [Any] for `this` call.
```

e5l (approved):
lgtm

PR #4236, opened 2024-08-21
KTOR-7289: Initial support for android native targets

whyoleg:
Fixes https://youtrack.jetbrains.com/issue/KTOR-7289/Support-androidNative-targets

TODO:
* needs `kotlinx-html` release with android native targets for `ktor-server-html-builder`
* `ISO_8859_1` charset is not supported - only UTF-8, because iconv is available only from Android API 28 (need hacks in cinterop)
* there is a lot of copy-paste from nix mainly, because:
  * commonizer doesn't work well in some cases and so it's not possible to use some declarations from common code (f.e in `ktor-network`
  * it's not possible with current auto-module configuration via `TargetsConfig` class to setup, that in some modules `androidNative` targets should be like nix, while in others it should be on it's own

Otherwise, it compiles and links :)
Probably it could be tested via an approach shown in https://github.com/Kotlin/kotlinx-datetime/pull/344
> To test, run an Android emulator, and then, in the command line, ./gradlew androidNativeArm64TestBinaries && adb push core/build/bin/androidNativeArm64/debugTest/test.kexe /data/local/tmp/ && adb shell /data/local/tmp/test.kexe

Not sure still about the quality of those changes...
Feel free to suggest how to make it better.

But after this, it will be much simpler to depend on ktor in libraries which have all Kotlin/Native targets

Next thing on my list of "small contributions" is to check if it's possible to improve ktor build to be compatible with Configuration Cache.


vbarthel-fr:
I'm not knowledgeable enough to give any constructive feedback on those changes, but since I successfully managed to start a ktor server from an android native binary from [whyoleg/android-native](https://github.com/whyoleg/ktor/tree/android-native) thanks to this work, I'm writing a little "thanks a lot" message :)

I just had to modify the publication script so that I can publish androidNativeXXX ktor artifacts locally to make them accessible from my test project. 

```diff
diff --git a/buildSrc/src/main/kotlin/Publication.kt b/buildSrc/src/main/kotlin/Publication.kt
index 1b62c68b0..6d02ae1f9 100644
--- a/buildSrc/src/main/kotlin/Publication.kt
+++ b/buildSrc/src/main/kotlin/Publication.kt
@@ -47,6 +47,15 @@ fun isAvailableForPublication(publication: Publication): Boolean {
         "macosArm64"
     )
 
+    val androidPublication = setOf(
+        "androidNativeArm32",
+        "androidNativeArm64",
+        "androidNativeX64",
+        "androidNativeX86"
+    )
+
+    result = result || name in androidPublication
+
     result = result || (HOST_NAME == "macos" && name in macPublications)
 
     return result
```

Screenshot 2024-09-11 at 21 44 24

Thanks again 🙏

whyoleg:
Note: if this PR will be merged, TC configuration for publishing should be updated to include new targets

e5l:
@osipxd, could you check when you have time?

buildSrc/src/main/kotlin/NativeUtils.kt
```diff
@@ -1,13 +1,30 @@
 /*
- * Copyright 2014-2021 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ * Copyright 2014-2024 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
  */
 
 import org.gradle.api.*
 
-fun Project.posixTargets(): List<String> = nixTargets() + windowsTargets()
+fun Project.posixTargets(): List<String> = nixTargets() + windowsTargets() + androidNativeTargets()
```
osipxd, line 7: Should android native be a part of `nixTargets` as it is Unix-like OS? If so, we could drop almost all duplicating code.

osipxd (commented):
Thank you for the contribution! 

I would try to reuse `nix` sources for `androidNative` targets. I've already tried to do it, but faced with the problem you described in the PR description when some posix functions used in `SocketUtilsNix` cannot be commonized.
I'm trying to fix this problem by splitting `nix` to `nix32` and `nix64`. I'll let you know about results of this experiment.

buildSrc/src/main/kotlin/NativeUtils.kt
whyoleg: In theory - it should be, but on practice because of the issues with commonizer there will be a lot of unresolved declarations in `nix` source-set :(

whyoleg: Probably the hierarchy should be similar to `atomicfu`: https://github.com/Kotlin/kotlinx-atomicfu/blob/92ff33b4319970b8ab164c61c8ebfb6a69702955/atomicfu/build.gradle#L154-L172

So:
* posix (= native)
  * nix
    * unix
      * linux
      * apple
    * androidNative
      * androidNative32
      * androidNative64
  * windows

osipxd: Here is PR needed for simple hierarchy changes: https://github.com/ktorio/ktor/pull/4349

osipxd:
@whyoleg, I've managed to get it work without splitting into `nix` and `unix` source-sets.
Be careful, I've force-pushed into your branch.

buildSrc/src/main/kotlin/NativeUtils.kt
```diff
@@ -1,12 +1,23 @@
...
  */
 
 import org.gradle.api.*
 
 fun Project.posixTargets(): List<String> = nixTargets() + windowsTargets()
 
-fun Project.nixTargets(): List<String> = darwinTargets() + linuxTargets()
+fun Project.nixTargets(): List<String> = darwinTargets() + linuxTargets() + androidNativeTargets()
+
+fun Project.supportsAndroidNative(): Boolean = project.targetIsEnabled("androidNative")
```
whyoleg, line 11: Q: probably the same should be done for `watchOS` above? In future PR (can do after this is merged)

buildSrc/src/main/kotlin/TargetsConfig.kt
```diff
@@ -117,7 +135,14 @@ private val hierarchyTemplate = KotlinHierarchyTemplate {
  * - `target.js.nodeJs`
  * - `target.js.browser`
  * - `target.wasmJs.browser`
+ * - `target.androidNative`
  */
 internal fun Project.targetIsEnabled(target: String): Boolean {
     return findProperty("target.$target") != "false"
 }
+
+/** Original `withAndroidNativeArm32` has bug and matches to `X86` actually. */
```
whyoleg, line 144:
```suggestion
/**
 * Original `withAndroidNativeArm32` has bug and matches to `X86` actually. 
 * Remove after https://youtrack.jetbrains.com/issue/KT-71866/KotlinHierarchyBuilder.withAndroidNativeArm32-selects-the-wrong-KonanTarget is fixed
 */
```

ktor-server/ktor-server-plugins/ktor-server-html-builder/gradle.properties
```diff
@@ -0,0 +1,6 @@
+#
+# Copyright 2014-2024 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+#
+
+# Needs support in kotlinx-html: https://github.com/Kotlin/kotlinx.html/pull/288
```
whyoleg, line 5: the PR is already merged and new versions is need to be released
cc @e5l

ktor-shared/ktor-serialization/ktor-serialization-kotlinx/ktor-serialization-kotlinx-xml/gradle.properties
```diff
@@ -5,3 +5,5 @@
 # xmlutil uses DOM API under the hood so it doesn't support Node.js.
 # Issue: https://github.com/pdvrieze/xmlutil/issues/83
 target.js.nodeJs=false
+# xmlutil doesn't provide android native target
```
whyoleg, line 8: Let's create an issue for the `xmlutil` and link here
I don't think it should be hard to support there

whyoleg (commented):
Nice!
I do like how better it is now!
Thanks!

P.S. it's a bit strange to review my own PR :)

ktor-shared/ktor-serialization/ktor-serialization-kotlinx/ktor-serialization-kotlinx-xml/gradle.properties
osipxd: Created: https://github.com/pdvrieze/xmlutil/issues/240

buildSrc/src/main/kotlin/NativeUtils.kt
osipxd: Yes, it is a good idea, feel free to open a PR

osipxd:
Not everything is so smooth 🥲 
```
Task ':ktor-utils:androidNativeArm32MetadataElements' uses this output of task ':ktor-utils:commonizeCInterop' without declaring an explicit or implicit dependency.
```
I can't reproduce this problem locally (because of https://github.com/gradle/gradle/issues/27576 I suppose).

whyoleg:
> I can't reproduce this problem locally (because of https://github.com/gradle/gradle/issues/27576 I suppose).

Same at my side, can't reproduce :(

Interesting, that the issue happens with the same target for which hierarchy API is wrong :)

osipxd:
> target for which hierarchy API is wrong

Do you mean `withAndroidNative()` and `withApple()` declared in groups containing other groups?

The same task [is still failing](https://ge.jetbrains.com/s/76guum4tholw2) :(

ktor-utils/build.gradle.kts
```diff
@@ -7,7 +7,7 @@ plugins {
 }
 
 kotlin {
-    createCInterop("threadUtils", nixTargets()) {
+    createCInterop("threadUtils", nixTargets() - androidNativeTargets()) {
```
osipxd, line 10: The problem exists only for this module where I exclude android native targets from C interop. Maybe it is related

whyoleg:
> Do you mean withAndroidNative() and withApple() declared in groups containing other groups?

Nope, I'm talking about the issue with `withAndroidNativeArm32` - the same target task is failing on CI `androidNativeArm32MetadataElements` - not sure if it's related though

ktor-utils/build.gradle.kts
whyoleg: we could try to split this interop per related targets.
E.g create `threadUtilsLinux` and `threadUtilsDarwin` with distinct sets of targets (or even different `def` files) - may be there is some bug in commonizer logic

whyoleg: Hey @osipxd!
I was able to overcome the issue here: https://github.com/ktorio/ktor/pull/4236/commits/bbd3151262f4c533d53995c78f19e89bc3fcff7f
Previously failed build is now [green](https://ktor.teamcity.com/buildConfiguration/Ktor_AllocationTests/282160) :)

osipxd (approved):
Great job, thank you!
Created a task to update CI configuration: [KTOR-7535](https://youtrack.jetbrains.com/issue/KTOR-7535)

PR #4237, opened 2024-08-22
KTOR-7334 Fix for AttributeKey byte code incompatibility

bjhham:
**Subsystem**
Client/Server, Utils

**Motivation**
[KTOR-7334](https://youtrack.jetbrains.com/issue/KTOR-7334) Bytecode incompatibility in AttributeKey

I encountered this when testing with Koin and the latest 3.0 EAP.  Turns out replacing the original constructor with a function of the same name doesn't produce the correct bytecode for backwards compatibility, which will break any external plugins using the `AttributeKey` class.

**Solution**
1. Added some crafty annotations to maintain backwards compatibility.
2. Updated the class to allow support for generics.

There is the caveat that external plugins using `AttributeKey` won't be able to take advantage of the type differentiation until the upgrade to 3.0.  BUT they won't blow up from a `NoSuchMethodFoundException` 😄

### DIFF
--- a/ktor-utils/common/src/io/ktor/util/Attributes.kt
+++ b/ktor-utils/common/src/io/ktor/util/Attributes.kt
@@ -4,6 +4,7 @@
 
 package io.ktor.util
 
+import kotlin.jvm.*
 import kotlin.reflect.*
 
 /**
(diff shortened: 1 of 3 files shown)


ktor-utils/common/src/io/ktor/util/Attributes.kt
```diff
@@ -4,24 +4,26 @@
...
 /**
  * Specifies a key for an attribute in [Attributes]
  * @param T is a type of the value stored in the attribute
  * @param name is a name of the attribute for diagnostic purposes. Can't be blank
  */
-public class AttributeKey<T : Any> @PublishedApi internal constructor(
+public data class AttributeKey<T : Any> @JvmOverloads constructor(
     public val name: String,
-    private val type: String
+    private val type: KType = typeOf<Any>(),
```
bjhham, line 26: I updated the class to keep the type information here.  I think it could be useful for the dependency injection story, and maybe for better forensics when the types don't align.

The constructor also needed to be exposed for when the KType can't be inferred, specifically for the example above with ConditionalHeaders (the suspend keyword breaks the type inference).

ktor-utils/common/src/io/ktor/util/Attributes.kt
```diff
@@ -4,24 +4,27 @@
...
 
 /**
  * Specifies a key for an attribute in [Attributes]
  * @param T is a type of the value stored in the attribute
  * @param name is a name of the attribute for diagnostic purposes. Can't be blank
  */
+@JvmSynthetic
 public inline fun <reified T : Any> AttributeKey(name: String): AttributeKey<T> =
-    AttributeKey(name, T::class.toString())
+    AttributeKey(name, typeOf<T>())
```
e5l, line 17: please check if it works on wasm and js

e5l (approved):

ktor-utils/common/src/io/ktor/util/Attributes.kt
bjhham: The tests pass ✅ 

One of the few platforms that actually pass... 😞

PR #4240, opened 2024-08-24
Fix UDP/TCP receive on Windows

Thomas-Vos:
**Subsystem**
ktor-network on mingwX64

**Motivation**
The recvfrom call can fail with `-1L`, which means there is an error. On Windows, the `WSAGetLastError()` function is called to retrieve the error, but for some reason this can return `0` (which means NO error). This does not happen that often, but when receiving UDP messages continually, this does occur about once per minute for me.

**Solution**
Solution is to just ignore the error as `0` means no error. Why this happens, no idea. UDP receive still works after this strange error and packets can be received. Not sure how to reproduce this in a test.

### DIFF
--- a/ktor-network/posix/src/io/ktor/network/sockets/CIOReader.kt
+++ b/ktor-network/posix/src/io/ktor/network/sockets/CIOReader.kt
@@ -31,8 +31,10 @@ internal fun CoroutineScope.attachForReadingImpl(
                     when (bytesRead) {
                         0 -> close = true
                         -1 -> {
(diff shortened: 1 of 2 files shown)


Thomas-Vos:
After some testing, it seems TCP has the same issue. Applied the workaround there as well, and now after a `0` error it keeps working fine.

Thomas-Vos:
@e5l could you review this, please?

e5l (approved):
lgtm! It would be nice to have a test

PR #4257, opened 2024-08-29
Websockets remove unnecessary memory allocations when trace logging level disabled.

berezovik-rs:
Websockets remove unnecessary memory allocations when trace logging level disabled.

**Subsystem**
Server websockets

**Motivation**
Unnecessary memory allocations if log level > TRACE

**Solution**
Add inline function with isTraceEnabled flag check


ktor-shared/ktor-websockets/common/src/io/ktor/websocket/DefaultWebSocketSession.kt
```diff
@@ -125,10 +125,12 @@ internal class DefaultWebSocketSessionImpl(
...
-        LOGGER.trace(
-            "Starting default WebSocketSession($this) " +
-                "with negotiated extensions: ${negotiatedExtensions.joinToString()}"
-        )
+        if (LOGGER.isTraceEnabled) {
+            LOGGER.trace(
+                "Starting default WebSocketSession($this) " +
+                    "with negotiated extensions: ${negotiatedExtensions.joinToString()}"
+            )
+        }
```
bjhham, line 133: Instead of duplicating the check here, could you please introduce an inline function in `ktor-utils/common/src/io/ktor/util/logging/Logger.kt`?

```kotlin
/**
 * Check `isTraceEnabled` flag before logging to save some memory allocations.
 */
public inline fun Logger.trace(message: () -> String) {
    if (isTraceEnabled) trace(message())
}
```

Then we can start replacing our other trace calls with this to save on memory allocations.

bjhham (changes requested):

bjhham (approved):
Could you

bjhham:
Could you please run:

```
./gradlew apiDump
```

And commit the resulting API file changes for ktor-utils prior to merging?

berezovik-rs:
> Could you please run:
> 
> ```
> ./gradlew apiDump
> ```
> 
> And commit the resulting API file changes for ktor-utils prior to merging?

Done

PR #4258, opened 2024-08-29
KTOR-7315 Deprecate old Jetty and Tomcat engines

bjhham:
**Subsystem**
Server, Jetty + Tomcat engines

**Motivation**
[KTOR-7315](https://youtrack.jetbrains.com/issue/KTOR-7315) Deprecate old jetty / tomcat / servlet engines

**Solution**
Add deprecation annotations to produce warnings.

### DIFF
--- a/ktor-server/ktor-server-jetty/jvm/src/io/ktor/server/jetty/EngineMain.kt
+++ b/ktor-server/ktor-server-jetty/jvm/src/io/ktor/server/jetty/EngineMain.kt
@@ -10,6 +10,10 @@ import io.ktor.server.engine.*
 /**
  * Jetty engine
  */
+@Deprecated(
+    "The ktor-server-jetty module is deprecated and will be removed in the next major release as it " +
+        "references an outdated version of Jetty. Please use the ktor-server-jetty-jakarta module instead."
+)
 public object EngineMain {
     /**
      * Main function for starting EngineMain with Jetty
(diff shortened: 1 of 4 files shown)


ktor-server/ktor-server-jetty/jvm/src/io/ktor/server/jetty/EngineMain.kt
```diff
@@ -10,6 +10,7 @@ import io.ktor.server.engine.*
 /**
  * Jetty engine
  */
+@Deprecated("This references an old version of Jetty. Use ktor-server-jetty-jakarta module instead")
```
e5l, line 13: "Please consider changing dependency from 'ktor-server-jetty' to 'ktor-server-jetty-jakarta'"

e5l (approved):
@vnikolova, could you also check?

ktor-server/ktor-server-jetty/jvm/src/io/ktor/server/jetty/EngineMain.kt
vnikolova: I suggest the following:
"The `ktor-server-jetty` module is deprecated and will be removed in the next major release as it references an outdated version of Jetty. Please use the `ktor-server-jetty-jakarta` module instead."

e5l: We can replace next major release with `4.0.0` explicitly

PR #4259, opened 2024-08-30
Fix gradle memory issue for running tests

bjhham:
**Subsystem**
Tests / CI

**Motivation**
Gradle daemon is exhausting it's memory in TC.

**Solution**
Disable the double receive test.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -11,9 +11,8 @@ version=3.0.0-rc-1-SNAPSHOT
 # gradle
 org.gradle.daemon=true
 org.gradle.caching=true
-org.gradle.vfs.watch=false
-org.gradle.configureondemand=true
-org.gradle.jvmargs=-Xmx8g -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=HEAP -Dfile.encoding=UTF-8 -XX:+UseParallelGC
+org.gradle.vfs.watch=true
+org.gradle.jvmargs=-Xmx6g -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8 -XX:+UseParallelGC
 
 # kotlin
 kotlin.native.ignoreDisabledTargets=true
--- a/gradle.properties
+++ b/gradle.properties
@@ -11,8 +11,9 @@ version=3.0.0-rc-1-SNAPSHOT
 # gradle
 org.gradle.daemon=true
 org.gradle.caching=true
-org.gradle.vfs.watch=true
-org.gradle.jvmargs=-Xmx8g -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8 -XX:+UseParallelGC
+org.gradle.vfs.watch=false
+org.gradle.configureondemand=true
+org.gradle.jvmargs=-Xmx8g -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=HEAP -Dfile.encoding=UTF-8 -XX:+UseParallelGC
 
 # kotlin
 kotlin.native.ignoreDisabledTargets=true
--- a/gradle.properties
+++ b/gradle.properties
@@ -12,7 +12,7 @@ version=3.0.0-rc-1-SNAPSHOT
 org.gradle.daemon=true
 org.gradle.caching=true
 org.gradle.vfs.watch=true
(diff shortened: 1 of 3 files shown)


bjhham:
Turns out I was trying to adjust the heap size greater than the build agent's memory, but the true culprit seems to be the rather cavalier allocations happening in the `DoubleRecieveTest`, which amounts to 500MB under the profiler.  I'll try to see if this is caused by some problematic allocations in our code, but for now I'll just reduce the big byte array size and use base64 encoding instead of making a big list of bytes.

gradle.properties
```diff
@@ -12,7 +12,8 @@ version=3.0.0-rc-1-SNAPSHOT
 org.gradle.daemon=true
 org.gradle.caching=true
 org.gradle.vfs.watch=true
-org.gradle.jvmargs=-Xmx6g -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8 -XX:+UseParallelGC
+org.gradle.configureondemand=true
+org.gradle.jvmargs=-Xmx6g -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=HEAP -Dfile.encoding=UTF-8 -XX:+UseParallelGC
```
marychatte, line 16: Why do we need `-XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=HEAP`?

ktor-server/ktor-server-plugins/ktor-server-double-receive/common/test/DoubleReceiveTest.kt
```diff
@@ -11,18 +11,21 @@ import kotlin.test.*
...
             ByteReadChannel(content),
             EmptyCoroutineContext
         )
 
         repeat(3) {
-            val channel = cache.read()
-            val received = channel.readRemaining().readByteArray()
-            assertEquals(content.toList(), received.toList())
+            val received = cache.read().readRemaining().readByteArray()
+            assertEquals(content.size, received.size, "Received content size should match")
```
marychatte, line 25: In another [commit](https://github.com/ktorio/ktor/pull/4259/commits/59f2ab9bdb2cad12efa4639d5f43e94dac79ba62), `base64` was used, and you were writing in a comment about it. Is it suddenly dropped?

bjhham: I initially replaced it with the base64 so that we'd save a bit of memory and get a decent output on the comparison, but then I discovered that the byte array is actually large enough that the output is pointless to look at, so I replaced it with a simple check.

gradle.properties
```diff
@@ -12,7 +12,8 @@ version=3.0.0-rc-1-SNAPSHOT
 org.gradle.daemon=true
 org.gradle.caching=true
 org.gradle.vfs.watch=true
-org.gradle.jvmargs=-Xmx6g -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8 -XX:+UseParallelGC
+org.gradle.configureondemand=true
```
e5l, line 15: we need to watch this carefully, it can cause side effects

ktor-server/ktor-server-plugins/ktor-server-double-receive/jvm/test/DoubleReceiveTestJvm.kt
```diff
@@ -13,7 +13,7 @@ class DoubleReceiveTestJvm {
 
     @Test
     fun testFileCache() = runBlocking {
-        val content = ByteArray(16 * 1024 * 1024) { it.toByte() }
```
e5l, line 16: Could you tell me why 16 mb is large?

e5l (approved):

ktor-server/ktor-server-plugins/ktor-server-double-receive/jvm/test/DoubleReceiveTestJvm.kt
bjhham: It was burning 500mb in the profiler, which was partly caused by the assert copying all the bytes into objects in the list, but after improving that it was still hitting the cap on TC.  I'll continue with the investigation here, but for now I just want to get the CI working 😅

gradle.properties
bjhham: I can revert the gradle changes here, I was just trying stuff to see if anything would stick before noticing it was always failing at the same test.

PR #4260, opened 2024-09-02
KTOR-7270 Rename ApplicationProperties to avoid some confusion

bjhham:
**Subsystem**
Server

**Motivation**
[KTOR-7270](https://youtrack.jetbrains.com/issue/KTOR-7270) Review naming for ApplicationProperties

I found this naming very confusing, since "properties" is very commonly used for key-value pairs of configuration items.

**Solution**
Renamed to `ServerConfig`:
 - because it's definitely a "configuration" object
 - it is the central-most configuration required for running any server i.e. "root"
 
It _is_ a bit of an obscure name, but that is fitting for the obscurity of its function.  This object is generally just used internally for passing around modules, paths, and environment.  It would make more sense if the paths and the _development mode_ flag belonged to environment and the logging and modules belonged to the application itself, but shifting those things around would be a major effort.


ktor-server/ktor-server-test-host/common/src/io/ktor/server/testing/TestApplication.kt
```diff
@@ -207,11 +207,11 @@ public open class TestApplicationBuilder {
     }
 
     /**
-     * Adds a configuration block for the [ApplicationProperties].
+     * Adds a configuration block for the [ApplicationRootConfig].
      * @see [testApplication]
      */
     @KtorDsl
-    public fun testApplicationProperties(block: ApplicationPropertiesBuilder.() -> Unit) {
+    public fun testRootConfig(block: ApplicationRootConfigBuilder.() -> Unit) {
```
e5l, line 214: Please consider having `@Deprecated` for the old name

ktor-client/ktor-client-android/jvm/test/io/ktor/client/engine/android/AndroidSpecificHttpsTest.kt
```diff
@@ -27,7 +27,7 @@ import kotlin.test.Test
 class AndroidSpecificHttpsTest : TestWithKtor() {
     override val server: EmbeddedServer<*, *> = embeddedServer(
         Netty,
-        applicationProperties {
+        rootConfig {
```
e5l, line 30: If we're going to rename `application` to `server` at some point, `serverConfig` also looks like a good name.
@bjhham, @marychatte, what do you think?

e5l (approved):
lgtm in general, please consider adding deprecation cycle

ktor-client/ktor-client-android/jvm/test/io/ktor/client/engine/android/AndroidSpecificHttpsTest.kt
marychatte: I thought about `serverConfig` before I saw your comment, so sound good for me

ktor-server/ktor-server-core/common/src/io/ktor/server/application/Application.kt
```diff
@@ -48,33 +48,34 @@ public class ApplicationPropertiesBuilder(
...
 
-    internal fun build(): ApplicationProperties =
-        ApplicationProperties(environment, modules, watchPaths, rootPath, developmentMode, parentCoroutineContext)
+    internal fun build(): ApplicationRootConfig =
+        ApplicationRootConfig(environment, modules, watchPaths, rootPath, developmentMode, parentCoroutineContext)
 }
 
 /**
- * An application config with which the application is running.
+ * Core configuration for a running server.
```
whyoleg, line 56: as per this line, it looks like `ServerConfig`/`HttpServerConfig` could be a good name here, WDYT?
Though, it's only if other entities are renamed...

ktor-server/ktor-server-plugins/ktor-server-default-headers/nix/interop/utils.def
```diff
@@ -0,0 +1,5 @@
+---
+
+char* read_ktor_version() {
```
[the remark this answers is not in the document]
bjhham: Looks like dumpApi generating some erroneous files 👎  I'll remove

ktor-client/ktor-client-android/jvm/test/io/ktor/client/engine/android/AndroidSpecificHttpsTest.kt
bjhham: If we're going to rename from `Application -> Server` at some point, then naming this `ServerConfig` would hit us with the dilemma of renaming the current `ApplicationConfig`.  I suppose we could go full circle and rename the current `ApplicationConfig` to `ServerProperties` at this point since it holds property references.

ktor-server/ktor-server-test-host/common/src/io/ktor/server/testing/TestApplication.kt
bjhham: This object / functions were introduced last year in the 3.0 stream, which is why I didn't bother with deprecation / typealias.

ktor-server/ktor-server-core/common/src/io/ktor/server/application/Application.kt
bjhham: If we do the Application -> Server rename it'll have to be in 4.0, in which case I'd rather remove this object because it doesn't really fit with the abstraction.  There's also the problem with there already being an `ApplicationConfig` which refers to the loaded properties.

ktor-client/ktor-client-android/jvm/test/io/ktor/client/engine/android/AndroidSpecificHttpsTest.kt
bjhham: Maybe I'll just go with ServerConfig for now since it seems most intuitive in the current scheme, and if there's future breaking changes then it doesn't matter so much since this is rarely used.

PR #4261, opened 2024-09-02
KTOR-4433 Fix Netty error on CORS trying to set header on committed requests

bjhham:
**Subsystem**
Server, Netty

**Motivation**
- [KTOR-4433](https://youtrack.jetbrains.com/issue/KTOR-4433) Netty: UnsupportedOperationException is thrown when responding in CallSetup and CORS plugin is installed

**Solution**
Ignore headers after the response is committed.  This aligns the behaviour with the other engines.

Ideally we'd have some mechanism for skipping other interceptors / phases in the application pipeline when you want to respond during the `CallSetup` phase and the like, though from what I can tell there is nothing in the platform for achieving this.

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-cors/common/src/io/ktor/server/plugins/cors/CORS.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-cors/common/src/io/ktor/server/plugins/cors/CORS.kt
@@ -81,6 +81,8 @@ internal fun PluginBuilder<CORSConfig>.buildPlugin() {
      * a plugin installation.
      */
     onCall { call ->
+        if (call.response.isCommitted) return@onCall
+
         if (!allowsAnyHost || allowCredentials) {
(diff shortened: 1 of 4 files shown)


e5l (approved):
LGTM

e5l:
Please check the code style before merging

e5l:
btw this oom is more frequent now

bjhham:
Looks like maybe the retries are just slowing down tests so they time out after 90 minutes.

bjhham:
I've added a little more time for the timeout for now 🙈

bjhham:
I think the failures on Windows are related... Perhaps there's a lag on clearing the sockets.

bjhham:
Oh I see, there is also a memory problem sometimes 😓

bjhham:
My changes in https://github.com/ktorio/ktor/pull/4231 should fix the memory problem.

PR #4271, opened 2024-09-05
KTOR-7398 Migrate ktor-server-metrics tests to testApplication DSL

osipxd:
**Subsystem**
Server, `ktor-server-metrics-*` modules

**Motivation**
[KTOR-7398](https://youtrack.jetbrains.com/issue/KTOR-7398/) Remove usage of deprecated withTestApplication DSL.

**Solution**
Migrated to testApplication DSL.

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-metrics-micrometer/jvm/test/io/ktor/server/metrics/micrometer/MicrometerMetricsTests.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-metrics-micrometer/jvm/test/io/ktor/server/metrics/micrometer/MicrometerMetricsTests.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2021 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ * Copyright 2014-2024 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
  */
 
 package io.ktor.server.metrics.micrometer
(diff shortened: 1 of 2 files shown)


ktor-server/ktor-server-plugins/ktor-server-metrics-micrometer/jvm/test/io/ktor/server/metrics/micrometer/MicrometerMetricsTests.kt
```diff
@@ -267,29 +253,31 @@ class MicrometerMetricsTests {
...
-                if (call.response.status() == HttpStatusCode.NotFound) {
-                    noHandlerHandledRequest = true
-                }
-            } catch (t: Throwable) {
-                throwableCaughtInEngine = t
-                if (call.response.status() == null) {
-                    call.respond(HttpStatusCode.InternalServerError)
+    private fun ApplicationTestBuilder.installDefaultBehaviour() {
+        application {
+            (engine as TestApplicationEngine).callInterceptor = {
```
osipxd, line 258: Is there a better way to specify `callInterceptor` without casting?

marychatte: You can use the fuinction `intercept`:
```kotlin
        application {
            val beforeSetup = PipelinePhase("BeforeSetup")
            insertPhaseBefore(ApplicationCallPipeline.Setup, beforeSetup)
            intercept(beforeSetup) {
                try {
                    proceed()
                    if (call.response.status() == HttpStatusCode.NotFound) {
                        noHandlerHandledRequest = true
                    }
                } catch (t: Throwable) {
                    throwableCaughtInEngine = t
                    if (call.response.status() == null) {
                        call.respond(HttpStatusCode.InternalServerError)
                    }
                }
            }
        }
```
It's important to add phase `beforeSetup` because `catch` here should work after `catch` in hook `CallFailed` (which is used in `MicrometerMetrics`), and `CallFailed` is set before `Setup` phase

marychatte (approved):
Good work!

PR #4272, opened 2024-09-05
KTOR-7397 Migrate ktor-server-auth tests to testApplication DSL

osipxd:
**Subsystem**
Server, `ktor-server-auth-*`

**Motivation**
[KTOR-7397](https://youtrack.jetbrains.com/issue/KTOR-7397/) Remove usage of deprecated `withTestApplication` DSL.

**Solution**
Migrated to `testApplication` DSL


ktor-server/ktor-server-plugins/ktor-server-auth-ldap/jvm/test/io/ktor/tests/auth/ldap/LdapAuthTest.kt
```diff
@@ -27,148 +28,144 @@ import kotlin.test.*
 class LdapAuthTest {
```
osipxd, line 28: This test fails for me both before refactoring and after it, so I cannot check the correctness of the refactoring.

e5l (approved):
looks good!

PR #4276, opened 2024-09-06
KTOR-6826 Include relocation notes for old ktor modules

bjhham:
**Subsystem**
Build System

**Motivation**
- [KTOR-6826](https://youtrack.jetbrains.com/issue/KTOR-6826) Include relocation notes for old Ktor modules
- [KTOR-7393](https://youtrack.jetbrains.com/issue/KTOR-7393) Redirect dependencies from ktor-server-test-base

**Solution**
I added a map for projects that have been moved, so that they can automatically be resolved by maven, and should include a note in the central listing.

I also noticed we have several artifacts that shouldn't have been included in the publications, so I added a couple to the `skipPublish` extension property, and included a clause in the publish condition to skip parent projects (projects without sub-projects, like `ktor-server`).

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -74,7 +74,25 @@ apply(from = "gradle/verifier.gradle")
 
 extra["skipPublish"] = mutableListOf(
     "ktor-server-test-base",
-    "ktor-junit"
+    "ktor-server-test-suites",
+    "ktor-server-tests",
+    "ktor-junit",
+)
+extra["relocatedArtifacts"] = mutableMapOf(
+    "ktor-server-auth" to "ktor-auth",
+    "ktor-server-auth-jwt" to "ktor-auth-jwt",
+    "ktor-server-auth-ldap" to "ktor-auth-ldap",
+    "ktor-server-freemarker" to "ktor-freemarker",
+    "ktor-server-metrics" to "ktor-metrics",
+    "ktor-server-metrics-micrometer" to "ktor-metrics-micrometer",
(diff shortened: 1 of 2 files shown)


build.gradle.kts
```diff
@@ -74,7 +74,25 @@ apply(from = "gradle/verifier.gradle")
 
 extra["skipPublish"] = mutableListOf(
     "ktor-server-test-base",
-    "ktor-junit"
+    "ktor-server-test-suites",
+    "ktor-server-tests",
+    "ktor-junit",
+)
+extra["relocatedArtifacts"] = mutableMapOf(
+    "ktor-server-auth" to "ktor-auth",
```
osipxd, line 82: **Nit:** `"ktor-server-auth" to "ktor-auth"` is a bit confusing. It looks like `ktor-server-auth` was relocated to `ktor-auth`. But honestly, I don't have any proposition here. Maybe another name for the extra could make it better, but I can't come up with anything better than `"newToOldArtifactMapping"` which looks cumbersome

osipxd (approved):
Looks good to me! Only one thought I left in a comment

build.gradle.kts
bjhham: I was thinking the same, actually.  I'll replace it with `mapOf(...).invert()` so that the declaration is more readable and the index is still correct.

osipxd (approved):

osipxd:
Oh, there are conflicts because of my PR just merged

PR #4277, opened 2024-09-06
osipxd/reconfigure-renovate

osipxd:
**Subsystem**
Renovate and dependency resolution

**Motivation**
Current Renovate configuration is very noisy

**Solution**
- Disabled update checks for packages we don't want to update.
- Fixed Ktor EAP versioning strategy
- Disabled semantic commits as they doesn't look consistent in our project
- Enabled adding of `dependencies` label to distinct such PRs visually from others
- Enabled automerge for minor and patch versions


renovate.json
```diff
@@ -1,5 +1,60 @@
...
+    {
+      "description": "Do not update Jetty 9.x used in ktor-server-jetty",
+      "matchPackageNames": [
+        "org.eclipse.jetty*"
+      ],
+      "matchCurrentVersion": "[,10.0.0)",
+      "matchUpdateTypes": [
+        "major"
+      ],
+      "enabled": false
```
bjhham, line 46: Will we still get patch updates with this rule?  There do still support some critical security updates for the 9.x releases...

bjhham (approved):

renovate.json
osipxd: Yes, we disable only major updates here

gradle/libs.versions.toml
```diff
@@ -1,189 +1,189 @@
 [versions]
-kotlin-version = "2.0.20"
-kotlinx-html-version = "0.11.0"
+kotlin = "2.0.20"
+kotlinx-html = "0.11.0"
```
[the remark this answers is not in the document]
osipxd: The only problem of the suffix is that when we reference such version from build script we have to write "version" twice: `libs.versions.ktlint.version.get()` vs `libs.versions.ktlint.get()`

bjhham:
The removal of the `-version` suffixes might cause some trouble with our maintenance releases.  Adding these changes into earlier version will create some conflicts.  I guess that's just a one-time problem though assuming we include these changes.

Issue #423, opened 2018-05-29
[client] Masking for WS should be enabled by default


Prototik:
According [RFC 6455 section 5.1](https://tools.ietf.org/html/rfc6455#section-5.1):

>  a client MUST mask all frames that it
>  sends to the server (see Section 5.3 for further details).  (Note
>  that masking is done whether or not the WebSocket Protocol is running
>  over TLS.)  The server MUST close the connection upon receiving a
>  frame that is not masked. 

I think client should conform RFC and enable masking by default unless user requested to disable masking. Currently it disabled by default.

e5l:
Hi, @Prototik. Thanks for the report, would be fixed soon

e5l:
Fixed in `master`

PR #4295, opened 2024-09-10
Migrate ktor-server-websockets tests to testApplication

osipxd:
**Subsystem**
Server, `ktor-server-websockets` tests

**Motivation**
[KTOR-7405](https://youtrack.jetbrains.com/issue/KTOR-7405) Remove usage of deprecated `withTestApplication` DSL.

**Solution**
Migrate to `testApplication` DSL.

It wasn't as simple as for other modules. Actually, I had to rewrite most of the tests to make them work.

It is better to review test-by-test in an IDE.

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-websockets/jvm/test/io/ktor/tests/websocket/WebSocketTest.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-websockets/jvm/test/io/ktor/tests/websocket/WebSocketTest.kt
@@ -1,34 +1,33 @@
 /*
-* Copyright 2014-2021 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
-*/
+ * Copyright 2014-2024 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
 
 package io.ktor.tests.websocket
 
+import io.ktor.client.*
 import io.ktor.client.plugins.websocket.*
+import io.ktor.client.plugins.websocket.cio.*
 import io.ktor.serialization.*
 import io.ktor.server.application.*
-import io.ktor.server.routing.*
 import io.ktor.server.testing.*
 import io.ktor.server.websocket.*
 import io.ktor.server.websocket.WebSockets
 import io.ktor.util.*
 import io.ktor.util.reflect.*
-import io.ktor.utils.io.*
+import io.ktor.utils.io.charsets.*
 import io.ktor.utils.io.core.*
 import io.ktor.websocket.*
 import kotlinx.coroutines.*
 import kotlinx.coroutines.channels.*
 import kotlinx.coroutines.debug.junit5.*
 import kotlinx.io.*
-import java.nio.*
-import java.nio.charset.*
-import java.time.*
-import java.util.*
-import java.util.concurrent.CancellationException
+import kotlin.random.*
 import kotlin.test.*
+import kotlin.time.Duration.Companion.seconds
 
 @CoroutinesTimeout(30_000)
 class WebSocketTest {
+
     class Data(val string: String)
 
     private val customContentConverter = object : WebsocketContentConverter {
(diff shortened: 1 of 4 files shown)


ktor-server/ktor-server-plugins/ktor-server-websockets/jvm/test/io/ktor/tests/websocket/WebSocketTest.kt
```diff
@@ -118,593 +106,402 @@ class WebSocketTest {
             webSocket("/echo") {
                 val data = receiveDeserialized<Data>(typeInfo<Data>())
                 sendSerialized(data, typeInfo<Data>())
-                outgoing.send(Frame.Close())
```
osipxd, line 121: I've removed explicit sending of `Frame.Close` almost everywhere as it is sent automatically at the end anyway

ktor-server/ktor-server-plugins/ktor-server-websockets/jvm/test/io/ktor/tests/websocket/WebSocketTest.kt
```diff
@@ -118,593 +106,402 @@ class WebSocketTest {
...
+        val executed = Job()
+        routing {
+            webSocket("/echo") {
+                assertFailsWith<WebsocketConverterNotFoundException>("No converter was found for websocket") {
+                    receiveDeserialized<Data>()
+                }
+                assertFailsWith<WebsocketConverterNotFoundException>("No converter was found for websocket") {
+                    sendSerialized(Data("hello"))
                 }
+                executed.complete()
```
osipxd, line 133: > "What happens on a server stays on the server" (c)

Assertion failures on a server are caught and logged by the server, so they don't affect the test result. Is there a recommended way to make assertions on a server? My current solution produces a timeout if one of the assertions fails, because `executed.complete()` will never be called, but I don't like that we miss actual assertion errors.

ktor-server/ktor-server-plugins/ktor-server-websockets/jvm/test/io/ktor/tests/websocket/WebSocketTest.kt
```diff
@@ -118,593 +106,402 @@ class WebSocketTest {
...
         }
-    }
-
-    @Test
-    fun testMasking() {
-        withTestApplication {
-            application.install(WebSockets)
-            application.routing {
-                webSocketRaw("/echo") {
-                    masking = true
```
osipxd, line 260: I've removed this test. It seems like masking [was enabled by default](https://github.com/ktorio/ktor/issues/423), and I have no idea how to test it without access to the actual byte stream.

ktor-server/ktor-server-plugins/ktor-server-websockets/jvm/test/io/ktor/tests/websocket/WebSocketTest.kt
```diff
@@ -118,593 +106,402 @@ class WebSocketTest {
...
-            handleWebSocket("/echo") {
-                setBody(hex("""0x88 0x02 0xe8 0x03""".trimHex()))
-            }.let { call ->
-                call.response.awaitWebSocket(Duration.ofSeconds(10))
-                assertEquals("0x88 0x02 0xe8 0x03".trimHex(), hex(call.response.byteContent!!))
-            }
+        var receivedClose = false
+        createWebSocketsClient().webSocketRaw(path = "/") {
+            send(Frame.Close())
+            receivedClose = incoming.receive() is Frame.Close
```
osipxd, line 206: I'm not sure about this test. It seems it should check that a client is able to send and receive `Frame.Close`, am I right?

ktor-server/ktor-server-plugins/ktor-server-websockets/jvm/test/io/ktor/tests/websocket/WebSocketTest.kt
```diff
@@ -118,593 +106,402 @@ class WebSocketTest {
...
                 }
             }
         }
 
-        handleWebSocketConversation("/close/me") { incoming, _ ->
+        var callClosedProperly = false
+        createWebSocketsClient().webSocketRaw(path = "/close/me") {
             assertTrue(incoming.receive() is Frame.Close)
             assertNull(incoming.receiveCatching().getOrNull())
+            callClosedProperly = true
```
osipxd, line 485: Speaking of missed assertion failures. `webSocketRaw` swallows all exceptions that occur in the given `block`. That's why I have to add a flag `callClosedProperly` to check later if it was changed to `true`.

It differs from `webSocket` which has no such problem, so I wonder if it is correct behavior.

bjhham: Oh, that's interesting... Maybe the block isn't connected with the parent coroutine.  Could you log a YT bug to investigate?

bjhham (approved):

ktor-server/ktor-server-plugins/ktor-server-websockets/jvm/test/io/ktor/tests/websocket/WebSocketTest.kt
bjhham: It'd be better to catch and send the errors from the server I would think.  Normally, you could add exception handling with https://ktor.io/docs/server-status-pages.html but I guess for websockets you'd want to send a custom message or something.

ktor-server/ktor-server-plugins/ktor-server-websockets/jvm/test/io/ktor/tests/websocket/WebSocketTest.kt
osipxd: Created [KTOR-7432](https://youtrack.jetbrains.com/issue/KTOR-7432/Inconsistent-behavior-when-an-exception-is-thrown-inside-webSocket-and-webSocketRaw)

ktor-server/ktor-server-plugins/ktor-server-websockets/jvm/test/io/ktor/tests/websocket/WebSocketTest.kt
osipxd: Status pages plugin is not added as a dependency here, so I replaced `Job` with `Deferred<Result<Unit>>`. Could you [take a look](https://github.com/ktorio/ktor/pull/4295/commits/6af7ca5bce5c12060462c89317c569ccb768dd12)?

PR #4296, opened 2024-09-10
Update Gradle Build Cache config to permit read-only access

adam-enko:
- Always configure build cache, even if the credentials aren't present.
- When on CI, enable pushing to remove build cache, but disable local cache (to force CI to update the remote cache).
- Always set the credentials. Even if they're missing, then Gradle should permit read-only access.

This will improve build performance for all users.

Further configuration might be necessary to permit read-only access to the Gradle Build Cache instance, because I can't access https://ktor-gradle-cache.teamcity.com/cache

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -12,18 +12,23 @@ pluginManagement {
 
 rootProject.name = "ktor"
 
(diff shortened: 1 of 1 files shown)


bjhham:
Nice 🎉

PR #4298, opened 2024-09-10
Update Ktor monorepo in buildSrc for test server

bjhham:
Resolves build error found in https://github.com/ktorio/ktor/pull/4286


osipxd (approved):

bjhham:
It would appear that this is somehow breaking all the HTTP timeout tests now...  Hmm.

bjhham:
This seems to be a legitimate bug.  Renovate and our test suite have finally accomplished something other than annoying me. 👏 🥇

buildSrc/src/main/kotlin/test/server/tests/Timeout.kt
```diff
@@ -35,7 +35,7 @@ internal fun Application.timeoutTest() {
                         override val contentType = ContentType.Application.OctetStream
                         override suspend fun writeTo(channel: ByteWriteChannel) {
                             for (offset in response.indices) {
-                                channel.writeFully(response, offset, 1)
+                                channel.writeFully(response, offset, offset + 1)
```
bjhham, line 38: Turns out the API changed from this argument being the length to the offset.  Because the test server kinda swallows exceptions, it took me a while to figure out what was going on there 😅

osipxd: It would be nice to implement something to automatically propagate exceptions thrown on the test server. What do you think?

bjhham: Adding exception handling has uncovered a test that's been silently failing for a while. 😅

bjhham: Oh, there is an ancient bug that prevents me from fixing this. https://youtrack.jetbrains.com/issue/KTOR-633/How-to-change-io.ktor.http.cio.CIOMultipartDataBase.formFieldLimit

bjhham: Oh nm, I got burned by the already-running test server.

osipxd: "20 minutes adventure" :D

osipxd: I've just discovered `ErrorCollector` + `EngineTestBase.createServer`, which is pretty close to what I meant under "automatically propagate exceptions thrown on the test server". But I thought about a standard mechanism for all users of `testApplication` DSL.

In other words, make this test fail by default:
```kotlin
@Test
fun `failure on a server`() = testApplication {
    routing {
        get("/") { fail() }
    }

    client.get("/")
}
```

osipxd: It is not related to the test server, so I will move this conversation into Slack

PR #4303, opened 2024-09-11
KTOR-7012 Deprecate `InputStream.toByteReadChannel` with fallback to ktor-io module

osipxd:
**Subsystem**
Server, `ktor-core`

**Motivation**
[KTOR-7012](https://youtrack.jetbrains.com/issue/KTOR-7012/) `InputStream.toByteReadChannel` creates an unattached job

**Solution**
Deprecate this extension in favor of the variant from `ktor-io` which doesn't suffer from this problem.


bjhham (approved):
Nice 👍

PR #4305, opened 2024-09-11
KTOR-7355 Check in usePreconfiguredSession that delegate is set

osipxd:
**Subsystem**
Client, `ktor-client-darwin`

**Motivation**
[KTOR-7355](https://youtrack.jetbrains.com/issue/KTOR-7355/) Preconfigured NSURLSession causes requests to hang

**Solution**
It is a valid behavior because the session was created without the delegate.
Nevertheless, it is quite unobvious behavior, so I added check that delegate was specified and example of usage to KDoc.

### DIFF
--- a/ktor-client/ktor-client-darwin/darwin/src/io/ktor/client/engine/darwin/DarwinClientEngineConfig.kt
+++ b/ktor-client/ktor-client-darwin/darwin/src/io/ktor/client/engine/darwin/DarwinClientEngineConfig.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2019 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ * Copyright 2014-2024 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
  */
 
 package io.ktor.client.engine.darwin
@@ -108,9 +108,9 @@ public class DarwinClientEngineConfig : HttpClientEngineConfig() {
      * ```
      * val delegate = KtorNSURLSessionDelegate()
      * val session = NSURLSession.sessionWithConfiguration(
-     *     NSURLSessionConfiguration.defaultSessionConfiguration,
+     *     NSURLSessionConfiguration.defaultSessionConfiguration(),
      *     delegate,
-     *     delegateQueue = NSOperationQueue()
+     *     delegateQueue = null
      * )
      *
      * usePreconfiguredSession(session, delegate)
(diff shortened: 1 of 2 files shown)


ktor-client/ktor-client-darwin/darwin/src/io/ktor/client/engine/darwin/DarwinClientEngineConfig.kt
```diff
@@ -104,9 +104,22 @@ public class DarwinClientEngineConfig : HttpClientEngineConfig() {
...
+     *     delegateQueue = NSOperationQueue()
+     * )
+     *
+     * usePreconfiguredSession(session, delegate)
+     * ```
+     *
      * @see [KtorNSURLSessionDelegate] for details.
      */
     public fun usePreconfiguredSession(session: NSURLSession, delegate: KtorNSURLSessionDelegate) {
+        requireNotNull(session.delegate) { "The session must be created with KtorNSURLSessionDelegate as a delegate" }
```
osipxd, line 122: Need help to come up with a more "actionable" error message

bjhham: Kinda a weird API for this.  Maybe the error could be "Invalid session: delegate field is null".  I dunno if that really helps 🤷

bjhham (approved):

ktor-client/ktor-client-darwin/darwin/src/io/ktor/client/engine/darwin/DarwinClientEngineConfig.kt
osipxd: > Kinda a weird API for this

Yes, this is an API supposed to be used only by those who actually need to specify custom delegate. For simple cases when you need to provide custom configuration, there is `configureSession` function.

I thought about removing `delegate` parameter as soon as we can get it using `session.delegate`, but it is not a solution. `KtorNSURLSessionDelegate` could be wrapped with another delegate, so all we can do is fast-fail on the obviously wrong configuration when the delegate is not set at all.

What do you think about this error message?
```
Invalid session: delegate field is null
Possible solutions:

1. Consider using `configureSession` instead of `usePreconfiguredSession` if the purpose is only to change session configuration.

2. For more complex cases, make sure you've set a delegate while creating the `session`.
```

osipxd: @vnikolova, could you also take a look? 🙏

vnikolova: I think your suggestion is good @osipxd. I would reword it a bit for more clarity and switch the places of the solutions as the second one seems to be more directly related to the error.
```
Invalid session configuration: delegate field is null
Possible solutions:

1. Ensure that you set a valid delegate when creating the `session`. For more details, see `KtorNSURLSessionDelegate`.

2. If you're only modifying session configuration, consider using `configureSession` instead of `usePreconfiguredSession`.

```

PR #4308, opened 2024-09-12
KTOR-7356 Increase multi-part formdata limit

bjhham:
**Subsystem**
Server, I/O

**Motivation**
[KTOR-7356](https://youtrack.jetbrains.com/issue/KTOR-7356) Multipart/form-data: Form limit applied for binary and file items

Users having difficulty with the low limit getting triggered during file upload posting in slack recently when trying out the latest build.  The limit here is generally used to avoid DDoS attacks with endless streams, so most other frameworks set this in the megabytes.

**Solution**
Bumped the default limit to 50MB and added some actionable details to the thrown exception.

### DIFF
--- a/ktor-http/ktor-http-cio/jvm/src/io/ktor/http/cio/CIOMultipartDataBase.kt
+++ b/ktor-http/ktor-http-cio/jvm/src/io/ktor/http/cio/CIOMultipartDataBase.kt
@@ -7,7 +7,6 @@ package io.ktor.http.cio
 import io.ktor.http.*
 import io.ktor.http.content.*
 import io.ktor.utils.io.*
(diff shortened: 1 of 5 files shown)


morki:
Should there be any limit at all? Doesn't existing `RequestBodyLimit` plugin solve it?

bjhham:
It appears that we can remove this mechanism for limiting form body size in favor of the dedicated plugin, but it would be a pretty disruptive change at this point in the release cycle, so I created an issue to track for the next major release https://youtrack.jetbrains.com/issue/KTOR-7443/Multi-part-form-field-limit-is-redundant

Stexxe (approved):

PR #4310, opened 2024-09-12
Fix flaky test WebSocketTest.testFragmentationMaxSize

osipxd:
**Subsystem**
Tests

**Motivation**
[Test history](https://ktor.teamcity.com/test/8431731462014885548?currentProjectId=Ktor_ProjectKtorCore)

**Solution**
Await for server finish before assertions


bjhham (approved):
One of these days we'll get these tests back into a functional state 🙌

PR #4313, opened 2024-09-13
WA: Disable JS legacy transformation manually until it is completely removed from the atomicfu plugin.

mvicsokolova:
This change fixes this [train failure](https://buildserver.labs.intellij.net/buildConfiguration/Kotlin_KotlinDev_KotlinxTrainKtorK2/565822720). 

```
e: KLIB resolver: The same 'unique_name=io.ktor:ktor-io' found in more than one library: /mnt/agent/work/1edc2f021dd239f7/project.ktor/ktor-io/build/classes/atomicfu/js/main, /mnt/agent/work/1edc2f021dd239f7/project.ktor/ktor-io/build/classes/kotlin/js/main
Please file an issue to https://kotl.in/issue and meanwhile use CLI flag `-Xklib-duplicated-unique-name-strategy` with one of the following values:
allow-all-with-warning: Use all KLIB dependencies, even when they have same `unique_name` property.
allow-first-with-warning: Use the first KLIB dependency with clashing `unique_name` property. No order guarantees are given though.
deny: Fail a compilation with the error.
```

Through an oversight `atomicfu-gradle-plugin` kept tranformJs option (corresponding to the legacy JS transformation) enabled by default, and with no *.js files to transform it just copied the compileJs output directory.

Here is the similar fix merged into kotlinx.coroutines: https://github.com/Kotlin/kotlinx.coroutines/pull/4223

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -102,7 +102,6 @@ apply(from = "gradle/compatibility.gradle")
 plugins {
     id("org.jetbrains.dokka") version "1.9.20" apply false
     id("org.jetbrains.kotlinx.binary-compatibility-validator") version "0.16.3"
-    //id("org.jetbrains.kotlinx.atomicfu") version "0.25.0" apply false
     id("com.osacky.doctor") version "0.10.0"
 }
 
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -102,7 +102,7 @@ apply(from = "gradle/compatibility.gradle")
 plugins {
     id("org.jetbrains.dokka") version "1.9.20" apply false
     id("org.jetbrains.kotlinx.binary-compatibility-validator") version "0.16.3"
-    id("org.jetbrains.kotlinx.atomicfu") version "0.25.0" apply false
+    //id("org.jetbrains.kotlinx.atomicfu") version "0.25.0" apply false
     id("com.osacky.doctor") version "0.10.0"
 }
 
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -128,7 +128,7 @@ allprojects {
     if (nonDefaultProjectStructure.contains(project.name)) return@allprojects
 
     apply(plugin = "kotlin-multiplatform")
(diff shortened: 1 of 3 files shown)


bjhham:
Could you check the CI build error?

Error:
```
org.gradle.api.InvalidUserCodeException: Querying the mapped value of map(flatmap(provider(task 'generatePrecompiledScriptPluginAccessors', class org.gradle.kotlin.dsl.provider.plugins.precompiled.tasks.GeneratePrecompiledScriptPluginAccessors))) before task ':buildSrc:generatePrecompiledScriptPluginAccessors' has completed is not supported
```

Scan:
https://scans.gradle.com/s/sabm6fewg2hke

mvicsokolova:
Seems that now, my fix is not the reason of the build failure anymore. @bjhham could please have a look? 🙏

Also, I've changed the base branch from `kotlin-community` to `main`. I'm not sure, what is the process of providing the fixes for the `kotlin-community` in ktor, but at least in other kotlinx libraries, the fix should appear in the dev branch first.

build.gradle.kts
```diff
@@ -102,7 +102,7 @@ apply(from = "gradle/compatibility.gradle")
 plugins {
     id("org.jetbrains.dokka") version "1.9.20" apply false
     id("org.jetbrains.kotlinx.binary-compatibility-validator") version "0.16.3"
-    id("org.jetbrains.kotlinx.atomicfu") version "0.25.0" apply false
+    //id("org.jetbrains.kotlinx.atomicfu") version "0.25.0" apply false
```
mvicsokolova, line 105: What was the original purpose of disabling the plugin here? 👀

osipxd: I think the purpose is to add the plugin to the build classpath but apply it later

e5l, line 105:
```suggestion
```

e5l (approved):

e5l:
LGTM, will be merged after CI check

PR #4315, opened 2024-09-13
Replace java.time with kotlin.time in websockets configuration

osipxd:
**Subsystem**
Client/Server, `ktor-server-websockets`, `ktor-client-core`, `ktor-shared:ktor-websockets`

**Motivation**
[KTOR-7446](https://youtrack.jetbrains.com/issue/KTOR-7446/) Replace `java.time` options in Websockets plugin

**Solution**
Replaced `java.time.Duration` with `kotlin.time.Duration`. Moved `Duration` extensions to the common source set. Added `Duration` support to a websocket client.


ktor-server/ktor-server-plugins/ktor-server-websockets/jvm/src/io/ktor/server/websocket/DurationsJvm.kt
```diff
@@ -0,0 +1,28 @@
...
+
+package io.ktor.server.websocket
+
+import java.time.*
+
+/**
+ * Ping interval or `null` to disable pinger. Please note that pongs will be handled despite this setting.
+ */
+@Deprecated("Use variant with kotlin.time.Duration", level = DeprecationLevel.ERROR)
+public inline var DefaultWebSocketServerSession.pingInterval: Duration?
```
osipxd, line 14: Now I'm not sure that it worth to keep these extensions with deprecation as they could confuse users 🤔

ktor-shared/ktor-websockets/common/src/io/ktor/websocket/DefaultWebSocketSession.kt
```diff
@@ -46,18 +50,26 @@ public interface DefaultWebSocketSession : WebSocketSession {
      */
     @InternalAPI
     public fun start(negotiatedExtensions: List<WebSocketExtension<*>> = emptyList())
+
+    public companion object {
+        /**
+         * Ping interval meaning pinger is disabled.
+         * @see pingIntervalMillis
+         */
+        public const val PINGER_DISABLED: Long = 0
```
osipxd, line 59: To be discussed: With this constant code seems a bit clearer, but I'm not sure if I chose the right place for it.

bjhham: I think it'd be best under the `WebSockets` companion since that's the only type users are generally exposed to.

bjhham (approved):

ktor-shared/ktor-websockets/common/src/io/ktor/websocket/DefaultWebSocketSession.kt
osipxd: The only thing I don't like is that it will be two constants, one for a client plugin and another for a server plugin.
I thought about making these constants pointing to the same constant in the shared module, but it will result in imports ambiguity.

```kotlin
public const val PINGER_DISABLED: Long = DefaultWebSocketSession.PINGER_DISABLED
```

osipxd: @bjhham, what do you think?

bjhham: Oh I see, maybe in that case it can be a top-level declaration.

PR #4316, opened 2024-09-15
KTOR-7035: Register MeterFilters before the very first Meter is registered.

sergejsha:
**Subsystem**
Server, plugins, ktor-server-metrics-micrometer

**Motivation**
The PR fixes annoying warning as described in https://youtrack.jetbrains.com/issue/KTOR-7035/

**Solution**
Registration of the MeterFilter is moved up, so that the filter is registered before the gauge meter. Warning message disappears.


bjhham (approved):
Thanks for the fix!

jonatan-ivanov:
I'm not familiar with the support terms and branching strategy of KTOR and neither the [issue](https://youtrack.jetbrains.com/issue/KTOR-7035) nor this PR have any milestones/labels set right now that would indicate the release version.
Are you planning to back-port this to 2.3.x?

PR #4318, opened 2024-09-16
KTOR-7449 Fix for uninitialized request

bjhham:
**Subsystem**
Server, Routing

**Motivation**
[KTOR-7449](https://youtrack.jetbrains.com/issue/KTOR-7449) UninitializedPropertyAccessException for `request` property when using MonitoringEvent(RoutingCallStarted)

**Solution**
Use lazy delegate over lateinit var.  The previous approach seemed a bit crazy to me.

### DIFF
--- a/ktor-server/ktor-server-core/common/test/io/ktor/server/routing/RouteTest.kt
+++ b/ktor-server/ktor-server-core/common/test/io/ktor/server/routing/RouteTest.kt
@@ -4,10 +4,16 @@
 
 package io.ktor.tests.routing
 
+import io.ktor.client.plugins.api.*
+import io.ktor.client.request.*
 import io.ktor.http.*
+import io.ktor.server.application.*
+import io.ktor.server.request.*
 import io.ktor.server.response.*
 import io.ktor.server.routing.*
+import io.ktor.server.routing.RoutingRoot.Plugin.RoutingCallStarted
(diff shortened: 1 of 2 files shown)


ktor-server/ktor-server-core/common/test/io/ktor/server/routing/RouteTest.kt
```diff
@@ -122,4 +128,19 @@ class RouteTest {
             assertEquals(expected, endpoints.map { it.toString() }.toSet())
         }
     }
+
+    // KTOR-7449 Ensure request is properly initialized in the routing call
```
Stexxe, line 132: It's redundant to write the issue ID here because it can be determined from the commit message

ktor-server/ktor-server-core/common/test/io/ktor/server/routing/RouteTest.kt
```diff
@@ -122,4 +128,19 @@ class RouteTest {
...
+    @Test
+    fun routingInitialization() = testApplication {
+        application {
+            monitor.subscribe(RoutingCallStarted) { call ->
+                println(call.request.httpMethod)
```
Stexxe, line 137: Maybe add assert instead of println?

Stexxe (approved):

PR #4327, opened 2024-09-20
KTOR-7427 Re-introduce readFully byte channel operation

bjhham:
**Subsystem**
Server, I/O

**Motivation**
[KTOR-7427](https://youtrack.jetbrains.com/issue/KTOR-7427) `ByteReadChannel.readFully(dst: ByteArray, offset: Int, length: Int)` is missing since 3.0.0-beta-2

**Solution**
Re-introduce the extension function.

### DIFF
--- a/ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
+++ b/ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
@@ -446,16 +446,20 @@ public val ByteReadChannel.availableForRead: Int
  * Reads all [length] bytes to [dst] buffer or fails if channel has been closed.
  * Suspends if not enough bytes available.
  */
-@OptIn(InternalAPI::class)
 public suspend fun ByteReadChannel.readFully(out: ByteArray) {
+    readFully(out, 0, out.size)
+}
+
+@OptIn(InternalAPI::class)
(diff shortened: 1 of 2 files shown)


ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
```diff
@@ -442,20 +442,33 @@ public suspend inline fun ByteReadChannel.read(crossinline block: suspend (ByteA
 public val ByteReadChannel.availableForRead: Int
     get() = readBuffer.buffer.size.toInt()
 
+
 /**
- * Reads all [length] bytes to [dst] buffer or fails if channel has been closed.
- * Suspends if not enough bytes available.
+ * Reads bytes available bytes into the provided [out] buffer, or fails
```
osipxd, line 447: Nit: typo
```suggestion
 * Reads all available bytes into the provided [out] buffer, or fails
```

osipxd (approved):

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
bjhham: I just dropped the `readFully` with sans args in favor of default args for simplicity.

PR #4328, opened 2024-09-20
KTOR-7448 Register shutdown hook earlier

osipxd:
**Subsystem**
Server, `ktor-server-core` + all engines

**Motivation**
[KTOR-7448](https://youtrack.jetbrains.com/issue/KTOR-7448) The ApplicationStopping event isn't raised when application is terminated with SIGINT

**Solution**
Register shutdown hook at the very start. Extract some common logic to tho common source set to align behavior between platforms.


ktor-server/ktor-server-core/api/ktor-server-core.api
```diff
@@ -707,8 +707,7 @@ public final class io/ktor/server/engine/ShutDownUrl$EnginePlugin : io/ktor/serv
 	public synthetic fun install (Lio/ktor/util/pipeline/Pipeline;Lkotlin/jvm/functions/Function1;)Ljava/lang/Object;
 }
 
-public final class io/ktor/server/engine/ShutdownHookJvmKt {
-	public static final fun addShutdownHook (Lio/ktor/server/engine/ApplicationEngine;Lio/ktor/events/Events;Lkotlin/jvm/functions/Function0;)V
+public final class io/ktor/server/engine/ShutdownHookKt {
 	public static final fun addShutdownHook (Lio/ktor/server/engine/EmbeddedServer;Lkotlin/jvm/functions/Function0;)V
```
osipxd, line 711: This function was introduced in v3.0.0-beta1

ktor-server/ktor-server-core/jsAndWasmShared/src/io/ktor/server/engine/ShutdownHook.jsAndWasmShared.kt
```diff
@@ -7,16 +7,7 @@ package io.ktor.server.engine
...
- * Adds automatic application shutdown hooks management. Should be used **before** starting the engine.
- * Once application termination noticed, [stop] block will be executed.
- * Please note that a shutdown hook only registered when the application is running. If the application
- * is already stopped then there will be no hook and no [stop] function invocation possible.
- * So [stop] block will be called once or never.
- */
-public actual fun ApplicationEngine.addShutdownHook(monitor: Events, stop: () -> Unit) {
-    addProcessShutdownHook(stop)
-}
+internal actual val SHUTDOWN_HOOK_ENABLED = true
```
osipxd, line 10: Are there alternatives to `System.getProperty` for native and JS platforms?

osipxd (commented):
I don't know how to test it. Do you have any ideas?

ktor-server/ktor-server-core/jsAndWasmShared/src/io/ktor/server/engine/ShutdownHook.jsAndWasmShared.kt
bjhham: Not that I know of...

bjhham (approved):

bjhham:
Can't think of a decent way to test this either... Would be easier if the server could be injected with something like a `HostRuntime` that included the needed platform calls so you could mock them out, though I fear that would be too much change for this bug fix.

PR #4329, opened 2024-09-20
KTOR-7442 Sessions: custom SessionSerializer is no longer supported

bjhham:
**Subsystem**

**Motivation**
[KTOR-7442](https://youtrack.jetbrains.com/issue/KTOR-7442) Sessions: custom SessionSerializer is no longer supported

**Solution**
Removed the implicit call to initialize the kotlinx serializer when constructing the cookie session builder.

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-sessions/common/src/io/ktor/server/sessions/SessionsBuilder.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-sessions/common/src/io/ktor/server/sessions/SessionsBuilder.kt
@@ -72,10 +72,7 @@ public fun <S : Any> SessionsConfig.cookie(
     val sessionType = typeInfo.type as KClass<S>
 
     val builder = CookieIdSessionBuilder(sessionType, typeInfo.kotlinType!!)
-        .apply {
-            block()
-            init()
-        }
+        .apply(block)
     cookie(name, builder, sessionType, storage)
 }
 
@@ -214,10 +211,7 @@ public fun <S : Any> SessionsConfig.cookie(
     val sessionType = typeInfo.type as KClass<S>
 
     val builder = CookieSessionBuilder(sessionType, typeInfo.kotlinType!!)
-        .apply {
-            block()
-            init()
-        }
+        .apply(block)
     cookie(name, sessionType, builder)
 }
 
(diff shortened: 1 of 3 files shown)


ktor-server/ktor-server-plugins/ktor-server-sessions/common/src/io/ktor/server/sessions/SessionsBuilder.kt
```diff
@@ -316,11 +316,14 @@ internal constructor(
...
 ) {
+    private var _serializer: SessionSerializer<S>? = null
 
     /**
      * Specifies a serializer used to serialize session data.
      */
-    public var serializer: SessionSerializer<S> = defaultSessionSerializer(typeInfo)
+    public var serializer: SessionSerializer<S>
+        set(value) { _serializer = value }
+        get() = _serializer ?: defaultSessionSerializer(typeInfo)
```
osipxd, line 326: Won't it be a problem that default session serialized could be created several times?

bjhham: Yeah, that's a good point.  I assumed from the name "Builder" that it would only reference the fields a single time for populating the the static instance, but it appears to not be the case.

ktor-server/ktor-server-plugins/ktor-server-sessions/common/src/io/ktor/server/sessions/SessionsBuilder.kt
```diff
@@ -343,6 +347,15 @@ internal constructor(
...
      */
     public val cookie: CookieConfiguration = CookieConfiguration()
+
+    /**
+     * Ensures fields are initialized properly.
+     */
+    internal fun init() {
+        if (!this::serializer.isInitialized) {
+            serializer = defaultSessionSerializer(typeInfo)
+        }
```
osipxd, line 357: Could we use lazy property instead?
```kotlin
private var _serializer: SessionSerializer<S>? = null
private val defaultSerializer: SessionSerializer<S> by lazy { defaultSessionSerializer(typeInfo }

public var serializer: SessionSerializer<S>
    set(value) { _serializer = value }
    get() = _serializer ?: defaultSerializer
```

bjhham: Yeah, good call, fixes the temporal coupling 👍

osipxd (approved):

PR #4337, opened 2024-09-24
KTOR-7483 Avoid caching requests with Authorization header

e5l:
Fix [KTOR-7483](https://youtrack.jetbrains.com/issue/KTOR-7483) HttpCache: Cache collision in the plugin when Authorization header is used [ZD-6809850]


ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/cache/HttpCache.kt
```diff
@@ -151,6 +151,7 @@ public class HttpCache private constructor(
             scope.sendPipeline.intercept(CachePhase) { content ->
                 if (content !is OutgoingContent.NoContent) return@intercept
                 if (context.method != HttpMethod.Get || !context.url.protocol.canStore()) return@intercept
+                if (context.headers.contains(HttpHeaders.Authorization)) return@intercept
```
bjhham, line 154: Could this instead be configurable this behaviour as the default?  Assuming the authentication is still enforced for cached responses, it would be useful to cache these responses as well.

bjhham (approved):

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/cache/HttpCache.kt
e5l: yep, let me add a config property

NilsBarlaug: Hi, @e5l! 😄  A bit late to the game, but the spec only forbids storing of responses to requests with `Authorization` header in _shared_ caches. If we store them in private caches, it is configurable through `isSharedClient` - and then we don't need the `cacheRequestWithAuth` property. I think these kinda things are why there is the distinction between private and shared caches.

e5l: Hey @NilsBarlaug, thanks for the notice. Let me check this

e5l: https://github.com/ktorio/ktor/pull/4368

PR #4345, opened 2024-09-25
Preparation for NodeJS CIO server

whyoleg:
Note: this is a part of ongoing implementation of client/server support for CIO on JS/WasmJS (https://github.com/ktorio/ktor/compare/main...whyoleg:ktor:common-cio)

NodeJS APIs are asynchronous by default but server-related functions in `ktor` are non-suspend.
This PR has breaking changes which will be needed to allow to implement CIO server via NodeJS for JS/WasmJS in 3.1.

### DIFF
--- a/ktor-client/ktor-client-cio/jvmAndPosix/src/io/ktor/client/engine/cio/ConnectionFactory.kt
+++ b/ktor-client/ktor-client-cio/jvmAndPosix/src/io/ktor/client/engine/cio/ConnectionFactory.kt
@@ -1,6 +1,6 @@
 /*
-* Copyright 2014-2021 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
-*/
+ * Copyright 2014-2024 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
 
 package io.ktor.client.engine.cio
 
@@ -27,7 +27,7 @@ internal class ConnectionFactory(
             addressSemaphore.acquire()
 
             try {
-                aSocket(selector).tcpNoDelay().tcp().connect(address, configuration)
+                aSocket(selector).tcp().connect(address, configuration)
             } catch (cause: Throwable) {
                 // a failure or cancellation
                 addressSemaphore.release()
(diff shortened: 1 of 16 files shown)


whyoleg:
FYI @e5l

ktor-client/ktor-client-cio/jvmAndPosix/src/io/ktor/client/engine/cio/ConnectionFactory.kt
```diff
@@ -27,7 +27,7 @@ internal class ConnectionFactory(
             addressSemaphore.acquire()
 
             try {
-                aSocket(selector).tcpNoDelay().tcp().connect(address, configuration)
+                aSocket(selector).tcp().connect(address, configuration)
```
e5l, line 30: Could you tell me why we dropped tcpNoDelay here?

e5l (changes requested):

ktor-client/ktor-client-cio/jvmAndPosix/src/io/ktor/client/engine/cio/ConnectionFactory.kt
whyoleg: yeah, sure:
1. because `noDelay` is true by default: [source](https://github.com/ktorio/ktor/blob/6ab2a63747a5c0d58c882155b905a3601f5927cd/ktor-network/jvmAndPosix/src/io/ktor/network/sockets/SocketOptions.kt#L161-L164)
2. because `tcpNoDelay` function does nothing here as it should be called on builder with `options: SocketOptions.TCPClientSocketOptions` to do anything: [source](https://github.com/ktorio/ktor/blob/6ab2a63747a5c0d58c882155b905a3601f5927cd/ktor-network/jvmAndPosix/src/io/ktor/network/sockets/Builders.kt#L35-L65) - so to work it should be called on result of `tcp()`, and has no effect if called on just `aSocket()`

So in this PR I've additionally deprecated `tcpNoDelay` because it's not needed anymore and it's behaviour is confusing

e5l (approved):

e5l:
Thanks for noticing, lgtm

PR #4346, opened 2024-09-25
KTOR-7484 Fix for performance issue in multi-part data retrieval

bjhham:
**Subsystem**
Server, I/O

**Motivation**
[KTOR-7484](https://youtrack.jetbrains.com/issue/KTOR-7484) multipart/form-data: Request timeout when receiving FileItem with 3.0.0-rc-1

**Solution**
Replaced the problematic `readUntilDelimiter` function with a common utility function `readUntil` that scans for the string in O(n) time and O(d) space where d is the delimiter.


e5l (approved):
LGTM, good catch! The `copyUntilBoundary` was here for the backward compatibility reasons, but we can drop it if it works badly

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
```diff
@@ -490,3 +491,173 @@ public fun ByteWriteChannel.rethrowCloseCauseIfNeeded() {
...
+
+/**
+ * Attempts to skip over a sequence of bytes in the channel that matches the provided [byteString].
+ *
+ * @param byteString The byte sequence to be matched and skipped over.
+ * @return [SkipResult.Ok] if the sequence matches and is successfully skipped,
+ *         [SkipResult.EndOfFile] if the end of the channel is reached before completing the match,
+ *         [SkipResult.Invalid] if the sequence does not match the channel's bytes.
+ */
+public suspend fun ByteReadChannel.trySkip(byteString: ByteString): SkipResult {
```
Stexxe, line 594: Would the not skipped bytes be returned to the channel (if the result is `SkipResult.Invalid`)?

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
```diff
@@ -490,3 +491,173 @@ public fun ByteWriteChannel.rethrowCloseCauseIfNeeded() {
...
+ * @param ignoreMissing Whether to ignore the missing byteString and return the count of read bytes upon reaching the end of input.
+ * @return The number of bytes read, not including the search string.
+ * @throws IOException If the limit is exceeded or the byteString is not found and ignoreMissing is false.
+ */
+public suspend fun ByteReadChannel.readUntil(
+    matchString: ByteString,
+    writeChannel: ByteWriteChannel,
+    limit: Long = Long.MAX_VALUE,
+    ignoreMissing: Boolean = false,
+): Long {
```
bjhham, line 513: This function will read directly to the output channel when looking for the matching bytestring, so we can rely on the coroutine suspension to keep the memory overhead in line.  I figured it would be a useful addition to the utility functions here, along with some skip functions, for when parsing directly from the bytechannel stream.

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
bjhham: Actually, I ended up using `skipIfFound` everywhere for this reason, which only consumes the bytes if the string is discovered.  I'll just remove the unused skip functions.

ktor-io/common/src/io/ktor/utils/io/ByteChannel.kt
```diff
@@ -70,7 +70,7 @@ public class ByteChannel(public val autoFlush: Boolean = false) : ByteReadChanne
         }
 
         if (_readBuffer.size < CHANNEL_MAX_SIZE) moveFlushToReadBuffer()
-        return _closedCause.value == null
+        return _readBuffer.size >= min
```
bjhham, line 73: @e5l I noticed this small issue in `ByteChannel` when testing the `skipIfFound` function - after the writer closes the channel, then the reader would always return false even if the required content is available.

PR #4347, opened 2024-09-26
KTOR-7492 Fix read System properties from jvm

e5l:
[KTOR-7492](https://youtrack.jetbrains.com/issue/KTOR-7492) Ktor doesn't read System properties from JVM


marychatte (approved):
lgtm

PR #4348, opened 2024-09-26
KTOR-7500 Fix multi-part read lock issue when part is skipped

bjhham:
**Subsystem**
Server, I/O

**Motivation**
[KTOR-7500](https://youtrack.jetbrains.com/issue/KTOR-7500) MultiPartData.readPart does not return null when stream ends

**Solution**
Introduce a reference to the previous part to automatically dispose.  This is necessary because of a change in 3.0 to use `ByteReadChannel` for file data, which is required for streaming the body properly.

Note: this will make it so you can't read the parts out of order, though I don't think that makes sense or was officially supported in any way before.

### DIFF
--- a/ktor-io/common/src/io/ktor/utils/io/SourceByteReadChannel.kt
+++ b/ktor-io/common/src/io/ktor/utils/io/SourceByteReadChannel.kt
@@ -4,6 +4,7 @@
 
 package io.ktor.utils.io
 
+import io.ktor.utils.io.core.*
 import kotlinx.io.*
 import kotlin.concurrent.*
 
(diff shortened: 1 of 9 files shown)


ktor-io/common/src/io/ktor/utils/io/SourceByteReadChannel.kt
```diff
@@ -26,7 +27,7 @@ internal class SourceByteReadChannel(private val source: Source) : ByteReadChann
 
     override suspend fun awaitContent(min: Int): Boolean {
         closedCause?.let { throw it }
-        return false
+        return source.remaining >= min
```
bjhham, line 30: This isn't so much related to the change, but I noticed the logic here was wrong.

e5l (approved):
lgtm

e5l:
Please check CI before merging

PR #4349, opened 2024-09-26
Use targets hierarchy API

osipxd:
**Subsystem**
Infrastructure

**Motivation**
[KTOR-7501](https://youtrack.jetbrains.com/issue/KTOR-7501/Use-target-hierarchy-DSL) Simplify targets hierarchy configuration 
More context here: https://github.com/ktorio/ktor/pull/4045

**Solution**
Configured target hierarchy using new [targets hierarchy DSL](https://github.com/JetBrains/kotlin/blob/v2.0.20/libraries/tools/kotlin-gradle-plugin-api/src/common/kotlin/org/jetbrains/kotlin/gradle/dsl/KotlinHierarchyDsl.kt)


whyoleg:
FYI: there was an attempt previously to migrate to target hierarchy. https://github.com/ktorio/ktor/pull/4044 and https://github.com/ktorio/ktor/pull/4045
But it was reverted (AFAIK by @e5l) because of some issues with long paths on Windows

buildSrc/src/main/kotlin/JsConfig.kt
```diff
@@ -1,33 +1,40 @@
...
             }
         }
     }
 
+    configureJsTasks(nodeJsEnabled)
     configureJsTestTasks()
 }
 
-private fun Project.configureJsTasks() {
+private fun Project.configureJsTasks(nodeJsEnabled: Boolean) {
```
whyoleg, line 34: minor: probably this could be inclined into DSL above to simplify the buildscript

buildSrc/src/main/kotlin/JvmConfig.kt
```diff
@@ -15,41 +16,43 @@ fun Project.configureJvm() {
...
-                    api("org.jetbrains.kotlin:kotlin-stdlib-jdk8:${Versions.kotlin}")
-                    api("org.jetbrains.kotlinx:kotlinx-coroutines-jdk8:${Versions.coroutines}") {
-                        exclude(module = "kotlin-stdlib")
-                        exclude(module = "kotlin-stdlib-jvm")
-                        exclude(module = "kotlin-stdlib-jdk8")
-                        exclude(module = "kotlin-stdlib-jdk7")
+        sourceSets {
+            jvmMain {
+                dependencies {
+                    if (jdk > 6) {
```
whyoleg, line 25: FYI: there is no need anymore to provide different versions of stdlib as everything now is available in just stdlib - which is applied by default.
the same for coroutines dependency.

So it's possible to simplify everything and just not add stdlib/coroutines in platform sourceSets, as `coroutines` are already added in `CommonConfig`

buildSrc/src/main/kotlin/TargetsConfig.kt
```diff
@@ -22,329 +22,91 @@ val Project.hasJsAndWasmShared: Boolean get() = files.any { it.name == "jsAndWas
...
-                    findByName("nixMain")?.let { dependsOn(it) }
-                }
+@OptIn(ExperimentalKotlinGradlePluginApi::class)
+private val hierarchyTemplate = KotlinHierarchyTemplate {
+    withSourceSetTree(KotlinSourceSetTree.main, KotlinSourceSetTree.test)
 
-                val linuxTest by getting {
-                    findByName("nixTest")?.let { dependsOn(it) }
+    common {
+        withCompilations { true }
```
whyoleg, line 72: AFAIK this line is not needed

buildSrc/src/main/kotlin/WasmConfig.kt
```diff
@@ -2,30 +2,36 @@
...
+            wasmJsTest {
                 dependencies {
-                    implementation(npm("puppeteer", Versions.puppeteer))
+                    implementation(npm("puppeteer", libs.versions.puppeteer.get()))
                 }
             }
         }
     }
 
+    configureWasmTasks()
```
whyoleg, line 34: the same here as with JS

osipxd:
> FYI: there was an attempt previously to migrate to target hierarchy. #4044 and #4045
> But it was reverted (AFAIK by @e5l) because of some issues with long paths on Windows

@e5l do you remember what problem was there? The only thing I found is that it was reverted as part of the migration to kotlinx-io (https://github.com/ktorio/ktor/pull/4032)

e5l:
@osipxd, import, and build failed on the Windows machine because of the long cache path generated by the Kotlin Gradle plugin. We need to check Windows project import in the IDE

osipxd:
> We need to check Windows project import in the IDE

Okay, I have a windows machine, so I will check

osipxd:
Import and build work fine on my Windows PC. Nevertheless, I've enabled [Windows build on CI](https://ktor.teamcity.com/buildConfiguration/Ktor_KtorMatrixCoreWindowsJava11/280849?hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandPull+Request+Details=true) to check it also on our CI.

bjhham (approved):

osipxd:
Let's give it a try 🤞

PR #4352, opened 2024-09-27
KTOR-7504 Kotlinx-io response extension and experimental JSON response

bjhham:
**Subsystem**
Server, I/O

**Motivation**
[KTOR-7504](https://youtrack.jetbrains.com/issue/KTOR-7504) Kotlinx-io response extension

I had just noticed that we don't have any easy way to respond with a `Source` type, and I figured we could also leverage the `kotlinx-io` API in our JSON serialization.

It could be a nice way for demoing the `kotlinx-io` support in 3.0.

**Solution**
Introduce a `respondSource` extension function that behaves the same way as `respondBytes` or `respondText`.


e5l (approved):
Good idea, please check code style and API dump

PR #4358, opened 2024-10-01
KTOR-7518 Add ReadableByteChannel to ByteReadChannel interop util

bjhham:
**Subsystem**
Shared, I/O

**Motivation**
[KTOR-7518](https://youtrack.jetbrains.com/issue/KTOR-7518) ReadableByteChannel to ByteReadChannel JVM interop is missing

I found this was missing during migration work for Space as a large sample project.  There are places where `ReadableByteChannel` is used for Ktor inputs.

**Solution**
Introduced a utility extension function for converting `ReadableByteChannel` to a `RawSource` and `ByteReadChannel` for async I/O.


e5l (approved):
I think we will need to deprecate this API in the future: it's intended to be used externally

PR #4368, opened 2024-10-04
KTOR-7483 Allow auth header when client is not shared

e5l:
Fix [KTOR-7483](https://youtrack.jetbrains.com/issue/KTOR-7483) HttpCache: Cache collision in the plugin when Authorization header is used [ZD-6809850]


osipxd (approved):
LGTM! Please, check code-style before merge

PR #4369, opened 2024-10-04
KTOR-7479 Quick fix for compatibility with R8 full mode

osipxd:
**Subsystem**
Client/Server

**Motivation**
TBD

**Solution**
TBD


ktor-utils/common/src/io/ktor/util/reflect/Type.kt
```diff
@@ -15,15 +15,35 @@ public expect val KType.platformType: Type
...
  */
-public data class TypeInfo(
+public class TypeInfo(
     public val type: KClass<*>,
     public val reifiedType: Type,
     public val kotlinType: KType? = null
-)
+) {
+
+    public constructor(type: KClass<*>, kotlinType: KType) : this(
```
osipxd, line 28: I would convert this second constructor into a function. And rename `typeInfo` to `TypeInfo`. What do you think?

ktor-utils/jvm/src/io/ktor/util/reflect/TypeInfoJvm.kt
```diff
@@ -10,11 +10,18 @@ public actual typealias Type = java.lang.reflect.Type
...
-    val kType = typeOf<T>()
-    val reifiedType = kType.javaType
-    return typeInfoImpl(reifiedType, T::class, kType)
+    val kType = typeOfOrNull<T>()
+    val reifiedType = try {
+        // kType != null always returns false, so we try to get javaType anyway
+        kType!!.javaType
+    } catch (_: NullPointerException) {
+        // Fallback to a type without generics if we can't get KType for the given T
+        T::class.java
```
osipxd, line 19: I will create an issue in the Kotlin project and mention its number here

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/api/CreatePluginUtils.kt
```diff
@@ -48,20 +48,26 @@ public fun <PluginConfigT : Any> createClientPlugin(
     name: String,
     createConfiguration: () -> PluginConfigT,
     body: ClientPluginBuilder<PluginConfigT>.() -> Unit
-): ClientPlugin<PluginConfigT> =
-    object : ClientPlugin<PluginConfigT> {
-        override val key: AttributeKey<ClientPluginInstance<PluginConfigT>> = AttributeKey(name)
```
osipxd, line 53: One of the problems was here because. AttributeKey call inside a non-inline function produces something like this:
```java
KTypeParameter var10006 = Reflection.typeParameter(new FunctionReferenceImpl(3, CreatePluginUtilsKt.class, "createClientPlugin", "createClientPlugin(Ljava/lang/String;Lkotlin/jvm/functions/Function0;Lkotlin/jvm/functions/Function1;)Lio/ktor/client/plugins/api/ClientPlugin;", 1), "PluginConfigT", KVariance.INVARIANT, false);
```
And it doesn't work after obfuscation.

bjhham (approved):
lgtm, nice sleuthing 👍

PR #4371, opened 2024-10-07
KTOR-6589 Replace MD5 with SHA-3

e5l:
Fix KTOR-6589

### DIFF
--- a/ktor-client/ktor-client-core/jvm/src/io/ktor/client/plugins/cache/storage/FileCacheStorage.kt
+++ b/ktor-client/ktor-client-core/jvm/src/io/ktor/client/plugins/cache/storage/FileCacheStorage.kt
@@ -84,7 +84,7 @@ private class FileCacheStorage(
         }
     }
 
-    private fun key(url: Url) = hex(MessageDigest.getInstance("MD5").digest(url.toString().encodeToByteArray()))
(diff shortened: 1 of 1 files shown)


whyoleg:
FYI: SHA3 is supported only on JDK 17+ and not supported on Android when I last checked (https://github.com/whyoleg/cryptography-kotlin/blob/fa49703c4e9dd40c4aae741eb28a5a932f07e49c/cryptography-providers-tests-api/src/commonMain/kotlin/support.kt#L29-L31)
Probably it's better to use `SHA-256`.

e5l:
@olme04, thanks, will replace

ktor-client/ktor-client-core/jvm/src/io/ktor/client/plugins/cache/storage/FileCacheStorage.kt
```diff
@@ -84,7 +84,7 @@ private class FileCacheStorage(
         }
     }
 
-    private fun key(url: Url) = hex(MessageDigest.getInstance("MD5").digest(url.toString().encodeToByteArray()))
+    private fun key(url: Url) = hex(MessageDigest.getInstance("SHA-3").digest(url.toString().encodeToByteArray()))
```
e5l, line 87:
```suggestion
    private fun key(url: Url) = hex(MessageDigest.getInstance("SHA-256").digest(url.toString().encodeToByteArray()))
```

bjhham (approved):

PR #4374, opened 2024-10-07
KTOR-6963 Fix darwin maxFrameSize option

e5l:
Fix [KTOR-6963](https://youtrack.jetbrains.com/issue/KTOR-6963) Darwin: The `maxFrameSize` option has no effect

### DIFF
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/websocket/WebSockets.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/websocket/WebSockets.kt
@@ -186,7 +186,7 @@ public class WebSockets internal constructor(
 
                 LOGGER.trace("Receive websocket session from ${context.request.url}: $session")
 
-                if (plugin.maxFrameSize != Long.MAX_VALUE) {
+                if (plugin.maxFrameSize != Int.MAX_VALUE.toLong()) {
                     session.maxFrameSize = plugin.maxFrameSize
                 }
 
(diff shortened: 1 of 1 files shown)


ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/websocket/WebSockets.kt
```diff
@@ -185,6 +185,11 @@ public class WebSockets internal constructor(
                 }
 
                 LOGGER.trace("Receive websocket session from ${context.request.url}: $session")
+
+                if (plugin.maxFrameSize != Long.MAX_VALUE) {
```
marychatte, line 189: Why compare with `Long.MAX_VALUE` when the default value is `Int.MAX_VALUE.toLong()`? https://github.com/ktorio/ktor/blob/3cd270690ca092fd3c01f82e2655821ec587009f/ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/websocket/WebSockets.kt#L117

marychatte:
Tests are falling for OkHttp engine :(

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/websocket/WebSockets.kt
e5l: Good catch, will fix
```suggestion
                if (plugin.maxFrameSize != Int.MAX_VALUE.toLong()) {
```

e5l:
@marychatte, fixed with the fix of default check

marychatte (approved):
lgtm

PR #4375, opened 2024-10-07
KTOR-7540 Turn TypeInfo.reifiedType into an extension-property

osipxd:
**Subsystem**
`ktor-utils`

**Motivation**
[KTOR-7540](https://youtrack.jetbrains.com/issue/KTOR-7540/Remove-reflection-utils-used-only-on-JVM-target-from-common-source-set) - JVM-specific field `reifiedType` has leaked into the common source set

**Solution**
Remove the field from `TypeInfo` and provide an extension property only for the JVM target.

### DIFF
--- a/ktor-utils/jvm/src/io/ktor/util/reflect/TypeInfoJvm.kt
+++ b/ktor-utils/jvm/src/io/ktor/util/reflect/TypeInfoJvm.kt
@@ -6,30 +6,27 @@ package io.ktor.util.reflect
 
 import kotlin.reflect.*
 
+@Deprecated("Not used anymore in common code as it was needed only for JVM target.")
 public actual typealias Type = java.lang.reflect.Type
 
+/** Type with substituted generics. */
 @OptIn(ExperimentalStdlibApi::class)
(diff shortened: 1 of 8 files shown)


ktor-utils/jvm/src/io/ktor/util/reflect/TypeInfoJvm.kt
```diff
@@ -6,30 +6,27 @@ package io.ktor.util.reflect
...
-    } catch (_: NullPointerException) {
-        // Fallback to a type without generics if we can't get KType for the given T
-        T::class.java
-    }
-    return TypeInfo(T::class, reifiedType, kType)
-}
+public val TypeInfo.reifiedType: java.lang.reflect.Type
+    // Fallback to a type without generics if we couldn't get KType.
+    // For example, when class signature was stripped by ProGuard/R8
+    get() = kotlinType?.javaType ?: type.java
```
osipxd, line 17: Now the hack looks much better. For some reason, the problem with null check doesn't exist for a class field.

bjhham (approved):
👍 👍

PR #4377, opened 2024-10-08
KTOR-7543 Fix a couple backwards compatibility issues in 3.0

bjhham:
**Subsystem**
Server, Routing

**Motivation**
[KTOR-7543](https://youtrack.jetbrains.com/issue/KTOR-7543) Common backwards compatibility problems for 3.0

These are a couple of items that are tough to solve with structural replacement, and I've found are very common in projects.

**Solution**
- Added coroutine scope to RoutingContext so that `launch` references don't break.
- Added property alias through an extension for `call`, which is generally always used for interceptors - this only applies to pipeline contexts where the context extends `ApplicationCall`


e5l (approved):

e5l:
@bjhham, please check API dump before merging

PR #4384, opened 2024-10-09
Fix configuration for kotlinx.train

osipxd:
**Subsystem**
Infrastructure

**Motivation**
Isolate kotlinx.train configuration better and fix versions substitution

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -88,6 +88,11 @@ subprojects {
 
         configureSourceSets()
         setupJvmToolchain()
+
+        compilerOptions {
+            languageVersion = getKotlinLanguageVersion()
+            apiVersion = getKotlinApiVersion()
+        }
     }
 
     val skipPublish: List<String> by rootProject.extra
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -2,55 +2,10 @@
  * Copyright 2014-2024 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
  */
 
-import org.gradle.api.Project
 import org.jetbrains.dokka.gradle.*
 import org.jetbrains.kotlin.gradle.dsl.*
 import org.jetbrains.kotlin.gradle.tasks.*
 import org.jetbrains.kotlin.konan.target.*
-
-buildscript {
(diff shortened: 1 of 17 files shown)


build.gradle.kts
```diff
@@ -142,7 +97,7 @@ subprojects {
     configureCodestyle()
 }
 
-println("Using Kotlin compiler version: ${org.jetbrains.kotlin.config.KotlinCompilerVersion.VERSION}")
+println("Using Kotlin compiler version: ${libs.versions.kotlin.get()}")
```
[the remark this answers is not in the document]
osipxd: Yes, as we override the kotlin version alias with the value from `kotlin_version` in `settings.gradle.kts`

buildSrc/src/main/kotlin/Train.kt
```diff
@@ -46,35 +47,26 @@ fun Project.filterSnapshotTests() {
...
-    val skipSnapshotChecks = rootProject.properties["skip_snapshot_checks"] != null
-    if (!skipSnapshotChecks) {
-        check(version, atomicfu_version, "atomicfu")
-        check(version, coroutines_version, "coroutines")
-        check(version, serialization_version, "serialization")
-    }
-    repositories {
-        mavenLocal()
-        maven(url = "https://oss.sonatype.org/content/repositories/snapshots")
+    if (!rootProject.hasProperty("skip_snapshot_checks")) {
```
tbogdanova, line 58: What is this property and where will it be used?

tbogdanova: Sorry, I see we use it in our config when building libraries

tbogdanova: But do I understand correctly that this parameter only makes sense when building ktor?

tbogdanova: for atomicfu, coroutines, serialization in project branches I also see logic with this parameter. 
I will add it to ktor build for train.

osipxd: This property might be useful for running train locally with different versions, but I don't think it should ever be used on CI

PR #4390, opened 2024-10-09
Add changelog for Ktor 3.0.0

Stexxe:


e5l (commented):
Could you check if all changes from RCs and betas are also included?

Stexxe:
> Could you check if all changes from RCs and betas are also included?

There would be too many issues then in a single release changelog.

bjhham:
I forgot to mark this little improvement as fixed [KTOR-7548](https://youtrack.jetbrains.com/issue/KTOR-7548) Introduce Flow API for multi-part

PR #4393, opened 2024-10-10
KTOR-6803 Use last resolved address to release connection

osipxd:
**Subsystem**

**Motivation**
[KTOR-6803](https://youtrack.jetbrains.com/issue/KTOR-6803) CIO: Requests face connection timeouts when executed on the Android main dispatcher

**Solution**
The problem was caused by a second address resolution performed in `releaseConnection()`. If something happens during this resolution (an exception is thrown or the resolved address differs from the first one), we never release a connection.

### DIFF
--- a/ktor-client/ktor-client-cio/jvm/test/io/ktor/client/engine/cio/CIORequestTest.kt
+++ b/ktor-client/ktor-client-cio/jvm/test/io/ktor/client/engine/cio/CIORequestTest.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2019 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ * Copyright 2014-2024 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
  */
 
 package io.ktor.client.engine.cio
@@ -16,12 +16,12 @@ import io.ktor.server.engine.*
 import io.ktor.server.netty.*
 import io.ktor.server.response.*
 import io.ktor.server.routing.*
+import io.mockk.*
 import kotlinx.coroutines.*
 import kotlinx.coroutines.debug.junit5.*
+import java.net.*
 import java.nio.channels.*
 import kotlin.test.*
-import kotlin.test.Ignore
-import kotlin.test.Test
 
 @CoroutinesTimeout(60_000)
 class CIORequestTest : TestWithKtor() {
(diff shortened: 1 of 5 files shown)


ktor-client/ktor-client-cio/jvmAndPosix/src/io/ktor/client/engine/cio/Endpoint.kt
```diff
@@ -282,7 +288,7 @@ internal class Endpoint(
     }
 
     private fun releaseConnection() {
-        val address = InetSocketAddress(host, port)
+        val address = lastResolvedAddress ?: return
```
osipxd, line 291: Not sure if `lastResolvedAddress` can be `null`. If not, we probably should throw an exception here

ktor-client/ktor-client-cio/jvm/test/io/ktor/client/engine/cio/CIORequestTest.kt
```diff
@@ -159,4 +162,14 @@ class CIORequestTest : TestWithKtor() {
             }
         }
     }
+
+    @Test
+    fun testInetAddressResolvedOncePerRequest() = testWithEngine(CIO) {
+        test { client ->
+            mockkStatic(InetAddress::getByName) {
+                client.prepareGet { url(path = "/echo", port = serverPort) }.execute()
+                verify(exactly = 1) { InetAddress.getByName(any()) }
```
osipxd, line 171: Do you have any ideas on how to test it better than this?

ktor-client/ktor-client-core/common/src/io/ktor/client/engine/HttpClientEngine.kt
```diff
@@ -91,14 +90,13 @@ public interface HttpClientEngine : CoroutineScope, Closeable {
...
 
-        val context = callContext + KtorCallContextElement(callContext)
-        return async(context) {
+        return withContext(callContext + KtorCallContextElement(callContext)) {
             if (closed) {
                 throw ClientEngineClosedException()
             }
 
             execute(requestData)
-        }.await()
```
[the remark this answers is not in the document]
e5l: It will be problematic, let's discuss

ktor-client/ktor-client-cio/jvmAndPosix/src/io/ktor/client/engine/cio/Endpoint.kt
```diff
@@ -204,7 +211,7 @@ internal class Endpoint(
                 val connect: suspend CoroutineScope.() -> Socket = {
                     connectionFactory.connect(address) {
                         this.socketTimeout = socketTimeout
-                    }
+                    }.also { lastResolvedAddress = address }
```
e5l, line 214: I would name it `connectionAddress`, cause it's the address used for connection

e5l (approved):

PR #4394, opened 2024-10-13
KTOR-7536: add mime types

anton-erofeev:
**Subsystem**
Core

**Motivation**
https://youtrack.jetbrains.com/issue/KTOR-7536

**Solution**
Added new MIME types


osipxd:
Thank you for the pull request!

I see that `loadMimes()` fails on CI with an exception `IllegalArgumentException: Failed to parse gcode`. The reason is that media type `gcode` doesn't follow the [format specified by RFC 9110](https://datatracker.ietf.org/doc/html/rfc9110#section-8.3.1).

I think the better source for media types would be the [IANA's Media Type registry](https://www.iana.org/assignments/media-types/media-types.xhtml). Could you update your PR with media types from this registry? It would be useful to add a link to this registry as a comment to `rawMimes`

osipxd (changes requested):
Thank you for the pull request!

I see that `loadMimes()` fails on CI with an exception `IllegalArgumentException: Failed to parse gcode`. The reason is that media type `gcode` doesn't follow the [format specified by RFC 9110](https://datatracker.ietf.org/doc/html/rfc9110#section-8.3.1).

I think the better source for media types would be the [IANA's Media Type registry](https://www.iana.org/assignments/media-types/media-types.xhtml). Could you update your PR with media types from this registry? It would be useful to add a link to this registry as a comment to `rawMimes`

osipxd:
> I see 2 options here - remove incompatible types, or leave these types

It is okay to have common non-standard types in the list, and I would keep these types for backward compatibility. However, I see that there are a lot of such types, and also not all media types from the IANA registry should be added to this list. So let me discuss this with the team. 

> The [source](https://www.iana.org/assignments/media-types/media-types.xhtml), as far as I can see, doesn’t contain extensions for types

Yes, it turns out to be more problematic than I thought. File extensions aren't mandatory for a media type to be added to the registry, so some of the media types specify them it in the "Additional information" section (see [audio/aac](https://www.iana.org/assignments/media-types/audio/aac)), and some don't (see [video/H261](https://www.iana.org/assignments/media-types/video/H261)).
I will come back with an action plan a bit later.

osipxd:
I've discussed this with the team and created two new tasks:
- [KTOR-7586](https://youtrack.jetbrains.com/issue/KTOR-7586) Improve parsing of supported media types (MIME types)
- [KTOR-7587](https://youtrack.jetbrains.com/issue/KTOR-7587) Update media types (MIME types) list

Let’s stick to adding only the media types requested in the original issue not to block the issue from being resolved. Updating the entire list is not so easy, so it's better to do it in a separate issue and PR (KTOR-7587).
Are you interested in implementing any of the new tasks yourself?

anton-erofeev:
> Let’s stick to adding only the media types requested in the original issue not to block the issue from being resolved.

looks good, I updated the merge request, please have a look 
>  Are you interested in implementing any of the new tasks yourself?

I can proceed with https://youtrack.jetbrains.com/issue/KTOR-7586 if you don't mind

osipxd (approved):
LGTM! Thank you for your contribution 🎉 

> I can proceed with https://youtrack.jetbrains.com/issue/KTOR-7586 if you don't mind
Sure, feel free to open a PR

PR #4396, opened 2024-10-14
Increase gradle heap to 10g

e5l:


e5l:
Didn't check, let's go with 10g and try to reduce it back in some time

PR #4397, opened 2024-10-14
KTOR-7571 byte read channel cpu lock

ascheja:
**Subsystem**
ktor-io

**Motivation**
See [KTOR-7571](https://youtrack.jetbrains.com/issue/KTOR-7571)

**Solution**
This attempts to solve a cpu-bound endless loop by both passing the number of expected bytes to `awaitContent()` and (in case `awaitContent()` doesn't actually suspend like [`SourceByteReadChannel`](https://github.com/ktorio/ktor/blob/c99356f5ea8a202bfaf022a57ab8de0643354e66/ktor-io/common/src/io/ktor/utils/io/SourceByteReadChannel.kt#L28)) to `yield()` to other coroutines between iterations.


e5l (approved):
Hey @ascheja, thank you for the PR
lgtm

PR #4399, opened 2024-10-15
KTOR-4318, KTOR-7561: Fix digest auth parameters

osipxd:
**Subsystem**
Client/Server, 

**Motivation**
- [KTOR-4318](https://youtrack.jetbrains.com/issue/KTOR-4318) Digest Auth does not implement nc parameter correctly according to RFC 7616
- [KTOR-7561](https://youtrack.jetbrains.com/issue/KTOR-7561) Digest auth: `username` and `cnonce` parameters aren't surrounded with quotes

**Solution**
Update parameters’ implementation to follow RFC 7616 guidelines.

This PR fixes only these two bugs, while we have more implementation inconsistencies with the RFC. I've created a separate issue to update Digest Auth implementation according to the RFC - [KTOR-7578](https://youtrack.jetbrains.com/issue/KTOR-7578/Update-Digest-authentication-implementation-according-to-RFC-7616)

It's better to review commit-by-commit

### DIFF
--- a/ktor-client/ktor-client-plugins/ktor-client-auth/common/src/io/ktor/client/plugins/auth/providers/DigestAuthProvider.kt
+++ b/ktor-client/ktor-client-plugins/ktor-client-auth/common/src/io/ktor/client/plugins/auth/providers/DigestAuthProvider.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2019 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ * Copyright 2014-2024 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
  */
 
 package io.ktor.client.plugins.auth.providers
@@ -179,7 +179,7 @@ public class DigestAuthProvider(
                 this["response"] = hex(token)
                 this["uri"] = url.fullPath
                 actualQop?.let { this["qop"] = it }
-                this["nc"] = nonceCount.toString()
+                this["nc"] = nonceCount.toString(radix = 16).padStart(length = 8, padChar = '0')
                 @Suppress("DEPRECATION_ERROR")
                 this["algorithm"] = algorithmName
             }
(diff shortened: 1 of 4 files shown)


ktor-http/common/src/io/ktor/http/auth/HeaderValueEncoding.kt
```diff
@@ -25,3 +25,7 @@ public enum class HeaderValueEncoding {
      */
     URI_ENCODE
 }
+
+public fun interface HeaderValueEncodingStrategy {
+    public fun chooseEncoding(name: String): HeaderValueEncoding
+}
```
osipxd, line 31: If this solution is OK, I will add documentation for this interface.

ktor-client/ktor-client-plugins/ktor-client-auth/api/ktor-client-auth.api
```diff
@@ -113,6 +113,7 @@ public final class io/ktor/client/plugins/auth/providers/DigestAuthCredentials {
 }
 
 public final class io/ktor/client/plugins/auth/providers/DigestAuthProvider : io/ktor/client/plugins/auth/AuthProvider {
+	public static final field Companion Lio/ktor/client/plugins/auth/providers/DigestAuthProvider$Companion;
```
e5l, line 116: Should we make the companion internal?

ktor-http/common/src/io/ktor/http/auth/HeaderValueEncoding.kt
e5l: It looks like the target branch should be at least 3.1.0-eap, please check if it is possible to fix without new public api

e5l (approved):

e5l:
lgtm, please check if it can be fixed in patch

ktor-http/common/src/io/ktor/http/auth/HeaderValueEncoding.kt
e5l: It also might be good to encode directly, so strategy will be an edge case

e5l (changes requested):
lgtm, left some comments

ktor-http/common/src/io/ktor/http/auth/HeaderValueEncoding.kt
osipxd: Do you mean adding some `HeaderValueEncoder` which will be responsible for encoding header values?

e5l: Yep

osipxd: I've managed to fix these bugs without introducing new APIs

e5l (approved):
Nice, lgtm! 
Please add issue numbers in the issue title :)

PR #4408, opened 2024-10-17
KTOR-6759 Darwin: Throw specific exception for SSL Pinning failure

osipxd:
**Subsystem**
Client, Darwin

**Motivation**
[KTOR-6759](https://youtrack.jetbrains.com/issue/KTOR-6759) Darwin: Ambiguous DarwinHttpRequestException for SSL Pinning failure

The problem is that we call `completionHandler` with `NSURLSessionAuthChallengePerformDefaultHandling` on SSL pinning failure and then third-party API throws an exception. There are no mechanisms to attach additional data to this exception (or at least I don't know any).
The place where the actual error occurs and the place where `DarwinHttpRequestException` is constructed are two completely different places.

**Solution**
Save actual failure thrown during challenge handling to a`DarwinTaskHandler` and throw it later instead of creating a generic one. Throw specific exception from `CertificatePinner` to make it possible to catch pinning failures specifically:

```kotlin
try {
    client.get("/some-api-with-pinning")
} catch(cause: TLSPeerUnverifiedException) {
    println("SSL pinning failure detected.")
}
```

`TLSPeerUnverifiedException` is a multiplatform exception, and on JVM it is mapped to `SSLPeerUnverifiedException`, so it should be possible to catch pinning failure in common code.

### DIFF
--- a/ktor-client/ktor-client-darwin/darwin/src/io/ktor/client/engine/darwin/certificates/CertificatePinner.kt
+++ b/ktor-client/ktor-client-darwin/darwin/src/io/ktor/client/engine/darwin/certificates/CertificatePinner.kt
@@ -6,8 +6,13 @@ package io.ktor.client.engine.darwin.certificates
 
 import io.ktor.client.engine.darwin.*
 import kotlinx.cinterop.*
-import platform.CoreCrypto.*
-import platform.CoreFoundation.*
+import platform.CoreCrypto.CC_SHA1
+import platform.CoreCrypto.CC_SHA1_DIGEST_LENGTH
+import platform.CoreCrypto.CC_SHA256
+import platform.CoreCrypto.CC_SHA256_DIGEST_LENGTH
+import platform.CoreFoundation.CFDictionaryGetValue
+import platform.CoreFoundation.CFStringCreateWithCString
+import platform.CoreFoundation.kCFStringEncodingUTF8
 import platform.Foundation.*
 import platform.Security.*
 
@@ -248,15 +253,8 @@ public data class CertificatePinner(
      * Returns list of matching certificates' pins for the hostname. Returns an empty list if the
      * hostname does not have pinned certificates.
      */
-    internal fun findMatchingPins(hostname: String): List<PinnedCertificate> {
-        var result: List<PinnedCertificate> = emptyList()
-        for (pin in pinnedCertificates) {
-            if (pin.matches(hostname)) {
-                if (result.isEmpty()) result = mutableListOf()
-                (result as MutableList<PinnedCertificate>).add(pin)
-            }
-        }
-        return result
+    private fun findMatchingPins(hostname: String): List<PinnedCertificate> {
+        return pinnedCertificates.filter { it.matches(hostname) }
     }
 
     /**
@@ -403,7 +401,6 @@ public data class CertificatePinner(
     ) {
         /**
          * Pins certificates for `pattern`.
-         *
          *
          * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.client.engine.darwin.certificates.CertificatePinner.Builder.add)
          *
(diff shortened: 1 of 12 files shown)


ktor-client/ktor-client-darwin/darwin/src/io/ktor/client/engine/darwin/certificates/CertificatePinner.kt
```diff
@@ -436,3 +430,5 @@ public data class CertificatePinner(
         )
     }
 }
+
+public class CertificatePinnerException(message: String) : IOException(message)
```
osipxd, line 434: I have concerns regarding this exception:
- Previously `DarwinHttpRequestException` was thrown for such cases, so it is potentially breaking change
- This exception is used for any error thrown by the pinner. Should we give a possibility to distinct between these errors?
- Should we provide a multiplatform exception to be able to handle SSL pinning problems in a unified manner?

e5l (approved):

ktor-client/ktor-client-darwin/darwin/src/io/ktor/client/engine/darwin/certificates/CertificatePinner.kt
theextremeprogrammer: @osipxd Thanks for this PR! I have very little context on the Ktor code, so I don't have much feedback to share. I can see that the exception contains the output message which is exactly what I was hoping to be able to access. I also appreciate how you included some tests. Thanks for your help and @e5l thank you for your review as well! 🙏

theextremeprogrammer (approved):

ktor-client/ktor-client-darwin/darwin/src/io/ktor/client/engine/darwin/certificates/CertificatePinner.kt
```diff
@@ -351,7 +343,7 @@ public data class CertificatePinner(
         val keyTypeRSA = CFBridgingRelease(kSecAttrKeyTypeRSA) as NSString
         val keyTypeECSECPrimeRandom = CFBridgingRelease(kSecAttrKeyTypeECSECPrimeRandom) as NSString
 
-        val size: Int = publicKeySize.intValue.toInt()
+        val size: Int = publicKeySize.intValue
```
e5l, line 346: Please keep toInt or use convert()

e5l (approved):
lgtm, please check comments

ktor-network/ktor-network-tls/jvm/src/io/ktor/network/tls/CipherSuitesJvm.kt
```diff
@@ -12,3 +12,6 @@ internal actual fun CipherSuite.isSupported(): Boolean = when (platformVersion.m
     "1.6.0" -> platformVersion.minor >= 181 || keyStrength <= 128
     else -> true
 }
+
+public actual typealias TLSException = javax.net.ssl.SSLException
```
osipxd, line 16: This is a source-compatible, but binary-incompatible change for JVM. `TLSException` has been made public in 3.0.0-beta2, so it shouldn't be extensively used yet, but I can make this change binary compatible by introducing a new class `TlsException` and deprecating the old one:

```kotlin
public actual typealias = TlsException = javax.net.ssl.SSLException

@Deprecated("Use TlsException instead")
class TLSException(message: String, cause: Throwable?) : TlsException(message, cause)
```

ktor-client/ktor-client-darwin/darwin/src/io/ktor/client/engine/darwin/certificates/CertificatePinner.kt
osipxd: Fixed. It seems IDEA removed these calls on cleanup 🤔

ktor-network/ktor-network-tls/jvm/src/io/ktor/network/tls/CipherSuitesJvm.kt
e5l: Let's make it binary compatible

PR #4411, opened 2024-10-18
KTOR-6004 Support TCP and Unix socket for wasm-js and js via NodeJS

whyoleg:
**Subsystem**
ktor-network

**Motivation**
Fixes https://youtrack.jetbrains.com/issue/KTOR-6004/Support-NodeJs-target-for-ktor-network

This is the first PR in a series to support ktor client/server for all targets available. Following PRs will contain CIO server and client engines support for NodeJS. The whole scope of changes can be found [here](https://github.com/ktorio/ktor/compare/main...whyoleg:ktor:commonize-cio).

Fully commonizing ktor-network and other modules, will also significantly simplify (from the build-logic perspective) support for wasm-wasi in ktor-network and through the whole stack of ktor

**Solution**
Use [node:net](https://nodejs.org/api/net.html) for sockets. `SelectorManager` is no-op for nodejs, as there is no need for it. Still it's better to commonize it to simplify the configuration. It will probably be also needed for wasm-wasi support.
In future, more high-level API could be introduced for sockets and `SelectorManager` could be made internal.


e5l:
Please check API and code style build. I would also check publishToMavenLocal.

whyoleg:
`apiCheck`/`lintKotlin`/`publishToMavenLocal` works fine locally

osipxd (approved):
Thank you for the PR! Looks good to me, just a couple of comments

ktor-network/jsAndWasmShared/src/io/ktor/network/sockets/nodejs/node.net.kt
```diff
@@ -0,0 +1,165 @@
...
+    fun address(): ServerLocalAddressInfo?
+    fun listen(options: ServerListenOptions)
+
+    // stop accepting new connections
+    fun close()
+
+    fun on(event: String /* "close", "listening" */, listener: () -> Unit)
+    fun on(event: String /* "connection" */, listener: (socket: Socket) -> Unit)
+    fun on(event: String /* "error" */, listener: (error: JsError) -> Unit)
+    fun on(event: String /* "drop" */, listener: (drop: ServerConnectionDrop) -> Unit)
```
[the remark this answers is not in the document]
whyoleg: It's possible to do something like [this](https://github.com/ktorio/ktor/blob/40042dc845e38dcdc2ba2fee0224fbfd2b13d172/ktor-client/ktor-client-core/wasmJs/src/io/ktor/client/engine/js/node/NodeFetch.kt#L21-L25).
Or I could just write extensions like `onError`/`onConnection` just below, and use them outside.

ktor-network/jsAndWasmShared/src/io/ktor/network/sockets/nodejs/node.net.kt
```diff
@@ -0,0 +1,165 @@
+/*
+ * Copyright 2014-2024 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.network.sockets.nodejs
+
+import io.ktor.network.sockets.*
+
+internal external interface JsError // js.Error
```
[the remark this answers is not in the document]
whyoleg: > Should we move JsError to ktor-utils?

There is `JsError` in `ktor-client-core`, but it's a bit different. Not sure, that it makes sense to extract those.

> Kotlin compiler map js.Error to Throwable

And they are - [in JS target](https://github.com/ktorio/ktor/blob/bef964ddcb685a83291a08513b7499301af2ba30/ktor-network/js/src/io/ktor/network/sockets/nodejs/node.net.js.kt#L30-L31), but not for wasm.
In wasmJs, `Throwable` is wasm exception, while `js.Error` is JS exception. And so far, I don't think that there function to map from one to another out-of-the-box.
AFAIK in Kotlin/Wasm exception `thrown` from JS part will be represented as [JsException](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.js/-js-exception/). More info on [kotlinlang](https://kotlinlang.org/docs/wasm-js-interop.html#kotlin-wasm-and-kotlin-js-interoperability-differences).
But here, the exception is not thrown, but passed via callback. So probably to convert it we could write something like:
```
fun convert(error: JsError): Throwable {
  try {
    throw(error)
  } catch(cause: Throwable) {
    return cause
  }
}

private fun throw(error: JsError): Nothing = js("throw error")
```
I will try to experiment with it. The alternative will be to just have something like:
```
fun convert(error: JsError): Throwable = CustomException("Exception from js: $error")
```

may be @bashor nows some other way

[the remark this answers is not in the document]
osipxd: Thank you for the explanation!

ktor-network/js/src/io/ktor/network/sockets/nodejs/node.net.js.kt
```diff
@@ -0,0 +1,45 @@
+/*
+ * Copyright 2014-2024 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.network.sockets.nodejs
+
+import io.ktor.network.sockets.*
+import org.khronos.webgl.*
+
+internal actual fun nodeNet(): NodeNet? = js("eval('require')('node:net')").unsafeCast<NodeNet?>()
```
osipxd, line 10: Won't it lead to [KTOR-6158](https://youtrack.jetbrains.com/issue/KTOR-6158)?

osipxd: I'm okay with the first option if we create an issue to track compatibility of network module with ES and add this context there. @e5l what do you think?

e5l: yep, the first option looks good

osipxd: @whyoleg, could you [create an issue](https://youtrack.jetbrains.com/newIssue?project=KTOR) and rebase this branch one more time? :)

whyoleg: created https://youtrack.jetbrains.com/issue/KTOR-7659/Make-ktor-network-compatible-with-ES-modules-for-nodejs
and rebased

osipxd:
Thank you for your contribution! :tada:

PR #4412, opened 2024-10-18
Remove space from default user agent

finalchild:
**Subsystem**
Client

**Motivation**
The product identifier in a user agent should not contain spaces.

https://httpwg.org/specs/rfc9110.html#field.user-agent

**Solution**
Make it kebab-case.

### DIFF
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/engine/Utils.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/engine/Utils.kt
@@ -16,7 +16,7 @@ import kotlin.coroutines.*
  * Default user agent to use in ktor client.
  */
 @InternalAPI
-public val KTOR_DEFAULT_USER_AGENT: String = "Ktor client"
+public val KTOR_DEFAULT_USER_AGENT: String = "ktor-client"
 
 private val DATE_HEADERS = setOf(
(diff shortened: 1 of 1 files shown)


e5l (approved):

e5l:
Hey @finalchild, I see in the spec that User Agent allows using spaces, and many agents are using them:
For instance:
```
Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/117.0.5938.132 Safari/537.36
```

Unless there is no reason to change it, I would consider keeping user agent consistent

simnalamburt:
That is because only "Mozilla" (which has no whitespaces) is a product identifier. The following parts are allowed to have whitespaces.

In Ktor's case, there is no slash or parentheses between "Ktor" and "client", which makes "Ktor client" as a product identifier, that is not allowed have whitespaces

finalchild:
Yes. "Ktor client" is parsed as two product identifiers, which semantically means there is a software called "Ktor" and there is another software called "client" which is less significant or less specific in terms of identifying the user agent.

Of course browser user agents are a mess and they almost have no rules whatsoever, but http libraries often stick to the official rules.

e5l (approved):

PR #4413, opened 2024-10-18
KTOR-7596 Make multipart Content-Type check case-insensitive

osipxd:
**Subsystem**
Server, `ktor-http-cio`

**Motivation**
[KTOR-7596](https://youtrack.jetbrains.com/issue/KTOR-7596) receiveMultipart fails with "IOException: Failed to parse multipart" when content-type is capitalized

**Solution**
Use case-insensitive check for multipart content-type

### DIFF
--- a/ktor-server/ktor-server-jetty/jvm/src/io/ktor/server/jetty/JettyKtorHandler.kt
+++ b/ktor-server/ktor-server-jetty/jvm/src/io/ktor/server/jetty/JettyKtorHandler.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2019 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ * Copyright 2014-2024 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
  */
 
 package io.ktor.server.jetty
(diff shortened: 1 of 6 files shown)


ktor-server/ktor-server-jetty/jvm/src/io/ktor/server/jetty/JettyKtorHandler.kt
```diff
@@ -72,7 +71,7 @@ internal class JettyKtorHandler(
     ) {
         try {
             val contentType = request.contentType
-            if (contentType != null && contentType.startsWith("multipart/")) {
+            if (contentType != null && contentType.startsWith("multipart/", ignoreCase = true)) {
```
e5l, line 74: I would extract constant here

e5l (approved):
lgtm

ktor-server/ktor-server-jetty/jvm/src/io/ktor/server/jetty/JettyKtorHandler.kt
osipxd: I could add a constant `ContentType.Application.TYPE`. I also thought about adding an operator function `contains`:

```kotlin
public object ContentType {
    public object Application {
        // ...

        public operator fun contains(contentType: CharSequence): Boolean =
            contentType.startsWith("$APPLICATION/", ignoreCase = true)
    }
}
```
And then check the content type using `in`: 
```kotlin
if (contentType in ContentType.Application) { ... }
```
What do you think?

osipxd: I'll make it in a separate PR

e5l: @bjhham, what do you think?

PR #4414, opened 2024-10-18
KTOR-7606 Make Closable implement AutoCloseable

osipxd:
**Subsystem**
Core, `ktor-ui`

**Motivation**
Improve UX - make it possible to call `AutoCloseable` extensions on our `Closeable` type
[KTOR-7606](https://youtrack.jetbrains.com/issue/KTOR-7606) Migrate to kotlin.AutoCloseable

**Solution**
Make `Closable` implement `AutoCloseable`. Use `kotlin.use` extension function instead of our implementation.


ktor-io/common/src/io/ktor/utils/io/core/Closeable.kt
```diff
@@ -4,25 +4,9 @@
...
-            cause.addSuppressed(closeException)
-        }
-        throw cause
-    } finally {
-        if (!closed) {
-            this?.close()
-        }
-    }
-}
+public expect interface Closeable : AutoCloseable
```
osipxd, line 9: We could deprecate our `Closable` interface and remove it in a major update if it is unnecessary.

e5l (approved):

PR #4420, opened 2024-10-22
KTOR-7644 Make re-auth status codes configurable

wkornewald:
Some services use 403 instead of 401. Changing them might be impossible. With this change Ktor can flexibly work with any broken service.

[KTOR-7644](https://youtrack.jetbrains.com/issue/KTOR-7644) Auth: Make re-auth/refresh status codes configurable

### DIFF
--- a/ktor-client/ktor-client-plugins/ktor-client-auth/common/src/io/ktor/client/plugins/auth/Auth.kt
+++ b/ktor-client/ktor-client-plugins/ktor-client-auth/common/src/io/ktor/client/plugins/auth/Auth.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2019 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ * Copyright 2014-2024 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
  */
 
 package io.ktor.client.plugins.auth
@@ -40,6 +40,7 @@ public class AuthConfig {
      *
      * You can set this value via [reAuthorizeOnResponse].
      */
+    @InternalAPI
     public var isUnauthorizedResponse: suspend (HttpResponse) -> Boolean = { it.status == HttpStatusCode.Unauthorized }
         private set
 
@@ -49,6 +50,7 @@ public class AuthConfig {
      * Use this to change the value of [isUnauthorizedResponse].
      */
     public fun reAuthorizeOnResponse(block: suspend (HttpResponse) -> Boolean) {
+        @OptIn(InternalAPI::class)
         isUnauthorizedResponse = block
     }
 }
@@ -66,6 +68,7 @@ public val AuthCircuitBreaker: AttributeKey<Unit> = AttributeKey("auth-request")
  *
  * @see [AuthConfig] for configuration options.
  */
+@OptIn(InternalAPI::class)
 public val Auth: ClientPlugin<AuthConfig> = createClientPlugin("Auth", ::AuthConfig) {
     val providers = pluginConfig.providers.toList()
 
@@ -75,7 +78,6 @@ public val Auth: ClientPlugin<AuthConfig> = createClientPlugin("Auth", ::AuthCon
     val tokenVersionsAttributeKey =
         AttributeKey<MutableMap<AuthProvider, Int>>("ProviderVersionAttributeKey")
 
-    @OptIn(InternalAPI::class)
     fun findProvider(
         call: HttpClientCall,
         candidateProviders: Set<AuthProvider>
@@ -124,7 +126,6 @@ public val Auth: ClientPlugin<AuthConfig> = createClientPlugin("Auth", ::AuthCon
         return true
     }
 
-    @OptIn(InternalAPI::class)
     suspend fun Send.Sender.executeWithNewToken(
         call: HttpClientCall,
         provider: AuthProvider,


ktor-client/ktor-client-plugins/ktor-client-auth/common/src/io/ktor/client/plugins/auth/Auth.kt
```diff
@@ -128,14 +130,14 @@ public val Auth: ClientPlugin<AuthConfig> = createClientPlugin("Auth", ::AuthCon
...
+        if (origin.response.status !in pluginConfig.reAuthStatusCodes) return@on origin
         if (origin.request.attributes.contains(AuthCircuitBreaker)) return@on origin
 
         var call = origin
 
         val candidateProviders = HashSet(providers)
 
-        while (call.response.status == HttpStatusCode.Unauthorized) {
+        while (call.response.status in pluginConfig.reAuthStatusCodes) {
             LOGGER.trace("Received 401 for ${call.request.url}")
```
osipxd, line 141: Could you also change log messages to reflect the changes and replace `.trace(...)` call with `.trace { ... }` in this file?

osipxd (commented):
Thank you for the pull request!
Let's discuss an alternative approach before we merge this.

ktor-client/ktor-client-plugins/ktor-client-auth/common/src/io/ktor/client/plugins/auth/Auth.kt
```diff
@@ -128,14 +130,14 @@ public val Auth: ClientPlugin<AuthConfig> = createClientPlugin("Auth", ::AuthCon
...
-        if (origin.response.status != HttpStatusCode.Unauthorized) return@on origin
+        if (origin.response.status !in pluginConfig.reAuthStatusCodes) return@on origin
         if (origin.request.attributes.contains(AuthCircuitBreaker)) return@on origin
 
         var call = origin
 
         val candidateProviders = HashSet(providers)
 
-        while (call.response.status == HttpStatusCode.Unauthorized) {
+        while (call.response.status in pluginConfig.reAuthStatusCodes) {
```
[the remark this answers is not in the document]
wkornewald: This sounds much better than the list. I've pushed that change.

ktor-client/ktor-client-plugins/ktor-client-auth/common/src/io/ktor/client/plugins/auth/Auth.kt
wkornewald: Sure. Done in the isUnauthorized commit.

osipxd (approved):
LGTM. Thank you!
@bjhham, @marychatte, could you also take a look?

ktor-client/ktor-client-plugins/ktor-client-auth/common/src/io/ktor/client/plugins/auth/Auth.kt
```diff
@@ -40,6 +42,7 @@ public val AuthCircuitBreaker: AttributeKey<Unit> = AttributeKey("auth-request")
  * You can learn more from [Authentication and authorization](https://ktor.io/docs/auth.html).
  *
  * [providers] - list of auth providers to use.
+ * [isUnauthorized] - lambda function to control whether a response is unauthorized and should trigger a re-auth.
```
wkornewald, line 45: Something else: this revers to variables that are actually defined on AuthConfig. Shouldn't the docstring be defined there instead or at least refer to the full path `[AuthConfig.providers]`? I just copied the existing code, but maybe this should be fixed, too.

ktor-client/ktor-client-plugins/ktor-client-auth/common/src/io/ktor/client/plugins/auth/Auth.kt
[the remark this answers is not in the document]
marychatte: @osipxd 
I like the idea with lambda!

> returning 200 OK with actual status code in a body

Do you mean it could be used like this?
 `isUnauthorized = { it.bodyAsText() == "401" }`
 If yes, then `isUnauthorized` function should be suspend

marychatte:
Thanks for the PR! Can we also please add tests for it?

wkornewald:
> Thanks for the PR! Can we also please add tests for it?

Sure, added one and switched to a suspend lambda.

ktor-client/ktor-client-plugins/ktor-client-auth/common/src/io/ktor/client/plugins/auth/Auth.kt
```diff
@@ -23,9 +24,20 @@ private class AtomicCounter {
...
 public class AuthConfig {
+    /**
+     * [AuthProvider] list to use.
+     */
     public val providers: MutableList<AuthProvider> = mutableListOf()
+
+    /**
+     * A lambda function to control whether a response is unauthorized and should trigger a refresh / re-auth.
+     */
+    public var isUnauthorized: suspend (HttpResponse) -> Boolean = { it.status == HttpStatusCode.Unauthorized }
```
osipxd, line 40: Let's refine the API a bit to make it more consistent to the existing ones.

```suggestion
    internal var isUnauthorized: suspend (HttpResponse) -> Boolean = { it.status == HttpStatusCode.Unauthorized }

    /** ... */
    public fun isUnauthorized(block: suspend (HttpResponse) -> Boolean) {
        isUnauthorized = block
    }
```

Though, I'm not sure about naming. Probably, `isUnauthorizedResponse` would be better. @marychatte, what do you think?

buildSrc/src/main/kotlin/test/server/tests/Auth.kt
```diff
@@ -129,7 +129,8 @@ internal fun Application.authTestServer() {
                     val token = call.request.headers["Authorization"]
                     if (token.isNullOrEmpty() || token.contains("invalid")) {
                         call.response.header(HttpHeaders.WWWAuthenticate, "Bearer realm=\"TestServer\"")
-                        call.respond(HttpStatusCode.Unauthorized)
+                        val status = call.request.queryParameters["status"]?.toIntOrNull() ?: 401
```
osipxd, line 132: Nit:
```suggestion
                        val status = call.request.queryParameters["status"]?.toIntOrNull() ?: HttpStatusCode.Unauthorized.value
```

osipxd (changes requested):

ktor-client/ktor-client-plugins/ktor-client-auth/common/src/io/ktor/client/plugins/auth/Auth.kt
wkornewald: But then you couldn't build on top of existing values:

```kotlin
val original = isUnauthorized
isUnauthorized = { original(it) || ... }
```

osipxd: We could make only `set()` private then, giving the possibility to get the configured `isAuthorized`:
```kotlin
public var isUnauthorized: suspend (HttpResponse) -> Boolean = { it.status == HttpStatusCode.Unauthorized }
    private set()
```
The point is to specify this field without an assignment operator.

buildSrc/src/main/kotlin/test/server/tests/Auth.kt
osipxd: I thought that the value here should be equal to the value used in the default `isUnauthorized` implementation.
It is a very minor comment, you can ignore it :)

ktor-client/ktor-client-plugins/ktor-client-auth/common/src/io/ktor/client/plugins/auth/Auth.kt
wkornewald: Why is a getter + a differently named setter function better than an assignment operator unifying those two?

wkornewald: Please take a look at the latest commit. I've changed it. Though, this solution seems less direct and understandable, especially when there's a public getter (which probably is missing in many of the other APIs?). The `reAuthorizeOnStatuses` extra function can also still be provided with a var, too, so I don't see that argument.

osipxd (approved):
Thank you for keeping this branch up-to-date!

Could you [allow editing by maintainers](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests)? I want to mark `isUnauthorizedResponse` as `@InternalAPI`.

wkornewald:
Oh right I forgot the InternalAPI markers. You should have access now.

ktor-client/ktor-client-plugins/ktor-client-auth/common/src/io/ktor/client/plugins/auth/Auth.kt
```diff
@@ -23,9 +23,36 @@ private class AtomicCounter {
...
+    /**
+     * The currently set function to control whether a response is unauthorized and should trigger a refresh / re-auth.
+     *
+     * By default checks against HTTP status 401.
+     *
+     * You can set this value via [reAuthorizeOnResponse].
+     */
+    @InternalAPI
+    public var isUnauthorizedResponse: suspend (HttpResponse) -> Boolean = { it.status == HttpStatusCode.Unauthorized }
+        private set
```
wkornewald, line 45: Why is this an internal API? My intention was to allow access to the current value, so you can extend an existing rule.

osipxd: We want to preserve the right to introduce breaking changes to this field. Keeping it public for those who’re ready to further breaking changes.

PR #4421, opened 2024-10-22
KTOR-7620 Make Url class @Serializable and JVM Serializable

wkornewald:
In our project we had to define our own UrlSerializer. It would be much nicer to have this in the Ktor library itself, so it works out of the box (similar to how Cookie was recently extended).

Also, types like Url and Cookie should be java.io.Serializable. Otherwise Android crashes when using those types as e.g. screen arguments. This happens very quickly when Url is used indirectly as part of a data class where we wanted type safety.

### DIFF
--- a/ktor-http/common/src/io/ktor/http/Url.kt
+++ b/ktor-http/common/src/io/ktor/http/Url.kt
@@ -2,6 +2,8 @@
  * Copyright 2014-2021 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
  */
 
+@file:OptIn(InternalAPI::class)
+
 package io.ktor.http
 
 import io.ktor.utils.io.*
@@ -263,8 +265,9 @@ internal val Url.encodedUserAndPassword: String
         appendUserAndPassword(encodedUser, encodedPassword)
     }
 
-public class UrlSerializer : KSerializer<Url> {
-    override val descriptor: SerialDescriptor = PrimitiveSerialDescriptor("Url", PrimitiveKind.STRING)
+public object UrlSerializer : KSerializer<Url> {
+    override val descriptor: SerialDescriptor =
+        PrimitiveSerialDescriptor("io.ktor.http.Url", PrimitiveKind.STRING)
 
     override fun deserialize(decoder: Decoder): Url =
         Url(decoder.decodeString())
--- a/ktor-http/common/src/io/ktor/http/Url.kt
+++ b/ktor-http/common/src/io/ktor/http/Url.kt
@@ -4,6 +4,11 @@
 
 package io.ktor.http
 
+import io.ktor.utils.io.*
+import kotlinx.serialization.*
+import kotlinx.serialization.descriptors.*
+import kotlinx.serialization.encoding.*
+
 /**
  * Represents an immutable URL
  *
@@ -18,6 +23,7 @@ package io.ktor.http
  * @property password password part of URL
  * @property trailingQuery keep trailing question character even if there are no query parameters
  */
+@Serializable(with = UrlSerializer::class)
 public class Url internal constructor(
     protocol: URLProtocol?,
     public val host: String,
@@ -29,7 +35,7 @@ public class Url internal constructor(
     public val password: String?,
     public val trailingQuery: Boolean,
     private val urlString: String
-) {
+) : JvmSerializable {
     init {
         require(specifiedPort in 0..65535) {
             "Port must be between 0 and 65535, or $DEFAULT_PORT if not set. Provided: $specifiedPort"
@@ -222,6 +228,8 @@ public class Url internal constructor(
         return urlString.hashCode()
     }
 
+    private fun writeReplace(): Any = JvmSerializerReplacement(UrlJvmSerializer, this)
+
     public companion object
 }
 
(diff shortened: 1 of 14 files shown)


ktor-io/jvm/src/io/ktor/utils/io/JvmSerializable.jvm.kt
```diff
@@ -0,0 +1,39 @@
...
+
+@Suppress("UNCHECKED_CAST")
+public actual fun <T : Any> JvmSerializerReplacement(serializer: JvmSerializer<T>, value: T): Any =
+    DefaultJvmSerializerReplacement(serializer, value)
+
+@PublishedApi // IMPORTANT: changing the class name would result in serialization incompatibility
+internal class DefaultJvmSerializerReplacement<T : Any>(
+    private var serializer: JvmSerializer<T>?,
+    private var value: T?
+) : Externalizable {
```
wkornewald, line 19: Should such an API better be integrated as part of the Kotlin stdlib itself or at least in a separate library for reuse in other projects? I'll add the same code to my open-source lib, too (https://github.com/ensody/ReactiveState-Kotlin).

ktor-http/common/src/io/ktor/http/Url.kt
```diff
@@ -254,3 +262,22 @@ internal val Url.encodedUserAndPassword: String
...
+
+    override fun deserialize(decoder: Decoder): Url =
+        Url(decoder.decodeString())
+
+    override fun serialize(encoder: Encoder, value: Url) {
+        encoder.encodeString(value.toString())
+    }
+}
+
+internal object UrlJvmSerializer : JvmSerializer<Url> {
```
wkornewald, line 277: Just an idea: To make this slightly easier to implement one could also add a reusable serializer which utilizes kotlinx-serialization (e.g. using the JSON internally).

e5l (changes requested):
Could you tell me why can't we mark it with @Serializable from kotlinx.serialization?

wkornewald:
> Could you tell me why can't we mark it with @serializable from kotlinx.serialization?

This PR is doing that, but that's not enough.

On Android many APIs work with anything that's Parcelable/Serializable. Imagine you have some random data class which contains a `Url`. You pass that object into an Android API and that tries to parcel/serialize it. There is no kotlinx.serialization involved in this case.

It's very error-prone and inconvenient to force everyone to check all the attributes of every data class and manually use kotlinx.serialization in many different places to ensure that the data is compatible with the Android API. The better solution is to make all widely used types compatible with `java.io.Serializable` or at least Parcelable. It's the same issue with kotlinx-datetime where people also run into crashes because `Instant` is incompatible with Android and nobody wants to manually write serializers and remember to use JSON serialization in thousands of places. Everything should just work without crashes and without thinking.

For this to be possible, the types that are very commonly used in the UI like `Url`, `Instant`, etc. need to be compatible with `java.io.Serializable`. But we can also chat or have a quick Huddle on the JetBrains Slack. I'm available there under "Waldemar Kornewald".

e5l:
That sounds good; let's mark parent interfaces with the `@InternalApi` annotation.

@osipxd, could you also check?

e5l (approved):

ktor-http/common/src/io/ktor/http/Url.kt
```diff
@@ -254,3 +262,22 @@ internal val Url.encodedUserAndPassword: String
     get() = buildString {
         appendUserAndPassword(encodedUser, encodedPassword)
     }
+
+public class UrlSerializer : KSerializer<Url> {
+    override val descriptor: SerialDescriptor = PrimitiveSerialDescriptor("Url", PrimitiveKind.STRING)
```
osipxd, line 267: Let's use a qualified name
```suggestion
    override val descriptor: SerialDescriptor = PrimitiveSerialDescriptor("io.ktor.http.Url", PrimitiveKind.STRING)
```

ktor-http/common/src/io/ktor/http/Url.kt
osipxd: I like this idea! Would you like to implement it in this PR or in a separate one?

ktor-io/jvm/src/io/ktor/utils/io/JvmSerializable.jvm.kt
osipxd: Regarding `JvmSreializable` interface, you can vote on this issue:
- [KT-48243](https://youtrack.jetbrains.com/issue/KT-48243) Publish `kotlin.io.Serializable`

I haven't found any issue proposing multiplatform `Externalizable` implementation, so feel free to create one!

ktor-shared/ktor-junit/jvm/src/io/ktor/junit/Assertions.kt
```diff
@@ -29,3 +31,10 @@ fun <T> assertAll(collection: Iterable<T>, assertion: (T) -> Unit) {
         }
     )
 }
+
+inline fun <reified T : Any> assertSerializable(obj: T): T {
+    val encoded = ByteArrayOutputStream().also {
+        ObjectOutputStream(it).writeObject(obj)
+    }.toByteArray()
+    return ObjectInputStream(encoded.inputStream()).readObject() as T
```
osipxd, line 39: Should the `assertEquals` call be right here?
```suggestion
    val decoded = ObjectInputStream(encoded.inputStream()).readObject() as T
    assertEquals(obj, decoded)
    return decoded
```

wkornewald: It would have to be optional (default true) because not all objects might be equal after deserialization, though in most cases they would be. But yes, that's actually how we do in our implementation. I was just afraid of the code looking 100% the same...

ktor-http/common/src/io/ktor/http/Url.kt
wkornewald: A separate one. This should rather happen in kotlinx.serialization, so it's generally reusable.

PR #4422, opened 2024-10-22
KTOR-6158 JS. Remove redundant `require` calls

turansky:
**Subsystem**
Client, JS

**Motivation**
https://youtrack.jetbrains.com/issue/KTOR-6158

**Solution**
Remove redundant `require` calls


e5l (approved):
please replace class with interface to avoid breaking changes

turansky:
> please replace class with interface to avoid breaking changes

Done

whyoleg:
Probably the same changes could be applied for wasmJs target

Vampire:
Can we also get rid of the remaining two `require` calls? (for `ws` and `crypto`)
I guess the same problems will be hit with those if their code paths are hit?

turansky:
> Can we also get rid of the remaining two require calls? (for `ws` and `crypto`)

2 additional separated PR-s (1 for `ws` and 1 for `crypto`) - preferable option

turansky:
@osipxd Thank you for assistance :)

PR #4424, opened 2024-10-23
Fix libcurl for arm64

89jd:
**Subsystem**
Client

**Motivation**
I couldn't build / link libcurl on arm64 platform. The reason is the one in the repo used a newer version of GCC to compile curl

**Solution**
I have compiled it with 8.3.3 to match the gcc used in kotlin multiplatform


e5l (changes requested):
Could you check if we can replace gnutls version with non GPL alternative? We don't link the library dynamically at the moment (the LGPL library must remain separable)

89jd:
It does dynamically link to gnu i think. I need the shared libraries and the linker flags to compile my app which uses this Ktor version

```
linkerOpts(
                    "-lbrotlidec",
                    "-lcom_err",
                    "-lgnutls",
                    "-lgssapi_krb5",
                    "-lidn2",
                    "-lk5crypto",
                    "-lkrb5",
                    "-llber",
                    "-lldap",
                    "-lnettle",
                    "-lnghttp2",
                    "-lpsl",
                    "-lrtmp",
                    "-lssh2",
                    "-lz",
```

I can just build from source, but wasn't sure if that would be alright. Want me to do that again and push it?

e5l:
let me try to check if I can do this locally

e5l:
but if you can build standalone non-gnu binary, it would be great

89jd:
Do you have any preferences to these features:

```

curl version:     8.10.1
  SSL:              enabled (OpenSSL)
  SSH:              no      (--with-{libssh,libssh2})
  zlib:             no      (--with-zlib)
  brotli:           no      (--with-brotli)
  zstd:             no      (--with-zstd)
  GSS-API:          no      (--with-gssapi)
  GSASL:            no      (libgsasl not found)
  TLS-SRP:          enabled
  resolver:         POSIX threaded
  IPv6:             enabled
  Unix sockets:     enabled
  IDN:              no      (--with-{libidn2,winidn})
  Build docs:       enabled (--disable-docs)
  Build libcurl:    Shared=no, Static=yes
  Built-in manual:  enabled
  --libcurl option: enabled (--disable-libcurl-option)
  Verbose errors:   enabled (--disable-verbose)
  Code coverage:    disabled
  SSPI:             no      (--enable-sspi)
  ca cert bundle:   /etc/ssl/certs/ca-certificates.crt
  ca cert path:     /etc/ssl/certs
  ca fallback:      no
  LDAP:             no      (--enable-ldap / --with-ldap-lib / --with-lber-lib)
  LDAPS:            no      (--enable-ldaps)
  RTSP:             enabled
  RTMP:             no      (--with-librtmp)
  PSL:              enabled
  Alt-svc:          enabled (--disable-alt-svc)
  Headers API:      enabled (--disable-headers-api)
  HSTS:             enabled (--disable-hsts)
  HTTP1:            enabled (internal)
  HTTP2:            no      (--with-nghttp2)
  HTTP3:            no      (--with-ngtcp2 --with-nghttp3, --with-quiche, --with-openssl-quic, --with-msh3)
  ECH:              no      (--enable-ech)
  WebSockets:       enabled
  Protocols:        dict file ftp ftps gopher gophers http https imap imaps ipfs ipns mqtt pop3 pop3s rtsp smb smbs smtp smtps telnet tftp ws wss
  Features:         alt-svc AsynchDNS HSTS HTTPS-proxy IPv6 Largefile NTLM PSL SSL threadsafe TLS-SRP UnixSockets

``

e5l:
http2 and http3 would be nice to have

whyoleg:
JFYI: accidentally, yesterday I was decided to check if it possible to build libcurl via Conan :) 
I was able to re-configure my openssl-builds for building libcurl:

https://github.com/whyoleg/openssl-builds/actions/runs/11641066156

While the build is green, there are some failures, mostly related to non-desktop targets are built and can be downloaded from artifacts there, or from the link below:

https://github.com/whyoleg/openssl-builds/actions/runs/11641066156/artifacts/2136057610

I haven't configured any flags so far, defaults are used. For now, libcurl built there should depend only on openssl and zlib

But, I haven't tested, that everything works fine when adding them to the project :)

BTW, what Kotlin targets do you think will benefit from static libraries? Or we could use static for all targets? Or we could split curl engine in two artefacts as I've done in [cryptography-kotlin](https://whyoleg.github.io/cryptography-kotlin/modules/cryptography-provider-openssl3/)?

e5l:
Hey @89jd, sorry for the delay. It looks like a more general and reproducible approach exists as #4445. Could you check if it helps to solve your issue?

e5l:
Obsolete by #4445

PR #4433, opened 2024-10-29
KTOR-8145 Add contains operator to ContentType objects

osipxd:
**Subsystem**
Core

**Motivation**
[KTOR-8145](https://youtrack.jetbrains.com/issue/KTOR-8145) Add operator contains to ContentType objects
Add a simple way to check if a content type has one of the well-known types. This check should be case-insensitive, that is why it is error-prone to check content type using `contentType.startsWith(...)`

**Solution**
Add an operator `contains` to ContentType objects checking that the content type prefix is equal to the expected type name ignoring its casing.


bjhham (commented):
I dig it 😎

bjhham (approved):

PR #4440, opened 2024-10-31
KTOR-7469 Fix IllegalArgumentException in HttpCookies when server returns a raw cookie with not allowed characters

marychatte:
**Subsystem**
Client, HttpCookies plugin

**Motivation**
[KTOR-7469](https://youtrack.jetbrains.com/issue/KTOR-7469)

**Solution**
The exception is thrown if a cookie with Raw encoding contains an invalid character such as:
```kotlin
isWhitespace() || this < ' ' || this in setOf(';', ',', '"')
```
Some servers send cookies to clients that contain these characters, however, the documentation does not support it. The solution is to remove this check because it doesn't break [Security Considerations](https://datatracker.ietf.org/doc/html/rfc6265#section-8)

### DIFF
--- a/ktor-http/common/src/io/ktor/http/Cookie.kt
+++ b/ktor-http/common/src/io/ktor/http/Cookie.kt
@@ -168,14 +168,7 @@ public fun renderSetCookieHeader(
  * Encode cookie value using the specified [encoding]
  */
 public fun encodeCookieValue(value: String, encoding: CookieEncoding): String = when (encoding) {
-    CookieEncoding.RAW -> when {
-        value.any { it.shouldEscapeInCookies() } ->
-            throw IllegalArgumentException(
-                "The cookie value contains characters that cannot be encoded in RAW format. " +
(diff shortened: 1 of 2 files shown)


ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/plugins/CookiesIntegrationTests.kt
```diff
@@ -241,5 +241,23 @@ class CookiesIntegrationTests : ClientLoader() {
         }
     }
 
+    @Test
+    fun testInvalidCharactersInCookie() = clientTests(listOf("Js")) {
```
osipxd, line 245: We already have a test checking this behavior: `RenderSetCookieTest.renderCookieThrowsOnNotEncodedExtensions()[jvm]`
Could you change it instead of adding a new one?

osipxd (approved):
LGTM

PR #4446, opened 2024-11-03
use encoded nonceCount to create response token

sproctor:
**Subsystem**
Client Auth Plugin

**Motivation**
Fixes https://youtrack.jetbrains.com/issue/KTOR-7681

**Solution**
The "response" token was being generated with "nc" encoded as a decimal string, when in Ktor 3.0.1, it was changes to be an 8 digit hex string. Passing the properly encoded "nc" into the token fixes the breakage.


osipxd (approved):
LGTM. Thank you for the quick fix!

PR #4450, opened 2024-11-05
Fix names of test packages according to location

anton-erofeev:
**Subsystem**
server

**Motivation**
Package directive does not match the file location

**Solution**
Fixed names of test packages according to location


osipxd (approved):
Thank you for the fix!

PR #4454, opened 2024-11-05
KTOR-7698 Make ServiceLoader.load call optimizable by R8

osipxd:
**Subsystem**
Client/Server

**Motivation**
[KTOR-7698](https://youtrack.jetbrains.com/issue/KTOR-7698) ServiceLoader.load call is slow on Android

**Solution**
Upstreamed @whyoleg's optimization done by him in #4441 and applied it to all `ServiceLoader.load` calls.


ktor-utils/jvm/src/io/ktor/util/reflect/ServiceLoader.kt
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * Loads implementations of service [T] using [ServiceLoader.load].
+ *
+ * NOTE: ServiceLoader should use specific call convention to be optimized by R8 on Android:
+ * `ServiceLoader.load(X.class, X.class.getClassLoader()).iterator()`
+ * See: https://r8.googlesource.com/r8/+/refs/heads/main/src/main/java/com/android/tools/r8/ir/optimize/ServiceLoaderRewriter.java
+ */
+@InternalAPI
+public inline fun <reified T> loadService(): Iterable<T> = Iterable {
```
whyoleg, line 18: note: it's better to recheck, that R8 will optimize this pattern with `inline` and `reified` usage. While from the code perspective it should be the same, I'm not confident, that the byte code will be the same.

osipxd: Yes, I'm not sure either 😅 I'm going to test it

osipxd: Yes, the optimization works!

During the testing I've found a more serious startup performance issue not related to `ServiceLoader`. We call `typeOf` during initialization of each plugin, and it is extremely slow when user have `kotlin-reflect` in dependencies.
Unoptimized service loader call takes ~20ms while initialization of HttpClient with five plugins installed takes more than a second! 😱 
I'll create a new issue for this problem.

osipxd: New issue:
- [KTOR-7715](https://youtrack.jetbrains.com/issue/KTOR-7715) Plugins’ initialization is slow on Android if kotlin-reflect is present

ktor-utils/jvm/src/io/ktor/util/reflect/ServiceLoader.kt
```diff
@@ -0,0 +1,23 @@
...
+    ServiceLoader.load(
+        T::class.java,
+        T::class.java.classLoader,
+    ).iterator()
+}
```
osipxd, line 23: I've tried to replace it with the following code, to retrieve iterator only once:
```kotlin
public inline fun <reified T> loadService(): Iterable<T>
    val iterator = ServiceLoader.load(
        T::class.java,
        T::class.java.classLoader,
    ).iterator()
    
    return Iterable { iterator }
}
```

But for some reason this change [breaks serialization tests](https://ktor.teamcity.com/buildConfiguration/Ktor_KtorMatrixCoreLinuxJava21/291227?expandBuildDeploymentsSection=false&hideTestsFromDependencies=false&expandBuildTestsSection=true&hideProblemsFromDependencies=false&expandPull+Request+Details=true&expandBuildChangesSection=true) 🤔

whyoleg: iterators are single use, so they can not be used like this and looks like in serialization tests,  result of `loadService()` is called multiple times and so when it's called second time, it will immediately start returning `hasNext=false`. (e.g https://pl.kotl.in/JkZSEveSX)

So looks like it's better to use `Iterable { ... }.toList()` as it was before

osipxd: Ah, it explains everything. I'll add two separate functions then: one to load all services and another to load only the first.

osipxd: Fixed 👉  https://github.com/ktorio/ktor/pull/4454/commits/1eeba0739dab72ce5b3d0f23ebec351656acfbd3

whyoleg: LGTM ^_^

bjhham (approved):
🤯

PR #4458, opened 2024-11-06
KTOR-6632 Support receiving multipart data with Ktor client

e5l:
Fix [KTOR-6632](https://youtrack.jetbrains.com/issue/KTOR-6632) Support receiving multipart data with Ktor client


ktor-client/ktor-client-core/build.gradle.kts
```diff
@@ -8,6 +8,7 @@ kotlin.sourceSets {
     commonMain {
         dependencies {
             api(project(":ktor-http"))
+            api(project(":ktor-http:ktor-http-cio"))
```
whyoleg, line 11: this dependency is a bit suspicious :)

May be it will be better to introduce this support in `ktor-http` (or in `ktor-client-core` directly) and just deprecate API in `CIO`?

e5l: The `ktor-http-cio` stands for ktor HTTP coroutines io. It is reused both in server and client. Could you tell me what problems are with this dependency?

whyoleg: Mostly, there are no problems, except for naming confusion between CIO engines and half-related to them `ktor-http-cio`. Currently `ktor-http-cio` contains mostly default implementation for features which are not available in some engines (e.g like this multi-part support). And after this PR it will become a dependency of both `ktor-client-core` and `ktor-server-core`, so it will makes no sense for me to even have `ktor-http-cio` and it will be better to just introduce this functionality/API in `ktor-http`.

ktor-http/ktor-http-cio/jsAndWasmShared/src/io/ktor/http/cio/MultipartJsAndWasm.kt
```diff
@@ -0,0 +1,14 @@
+package io.ktor.http.cio
+
+import io.ktor.utils.io.ByteReadChannel
+import io.ktor.utils.io.cancel
+import io.ktor.utils.io.discard
+import io.ktor.utils.io.exhausted
```
osipxd, line 6: Hmm, `ij_kotlin_name_count_to_use_star_import` option from `.editorconfig` is not respected. Are you using an IDE different from IDEA?

```suggestion
import io.ktor.utils.io.*
```

ktor-http/ktor-http-cio/jsAndWasmShared/src/io/ktor/http/cio/MultipartJsAndWasm.kt
```diff
@@ -0,0 +1,14 @@
...
+
+/*
+ * Copyright 2014-2024 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
```
osipxd, line 10: Minor. Usually we place the copyright note on the first line in a file.

ktor-http/ktor-http-cio/common/src/io/ktor/http/cio/MultipartCommon.kt
```diff
@@ -4,16 +4,32 @@
```
osipxd, line 4: Can we keep the name `Multipart.kt` for binary compatibility?

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/DefaultTransform.kt
```diff
@@ -111,6 +112,28 @@ public fun HttpClient.defaultTransformers() {
...
             }
 
+            MultiPartData::class -> {
+                val contentType =
+                    context.response.contentType() ?: throw IllegalStateException("No content type in response")
+                if (!contentType.match(ContentType.MultiPart.FormData)) {
+                    throw IllegalStateException(
+                        "Expected multipart/form-data, got $contentType"
+                    )
+                }
```
osipxd, line 122: Minor
```suggestion
                val contentType = checkNotNull(context.response.contentType()) { "No content type in response" }
                check(contentType.match(ContentType.MultiPart.FormData)) { "Expected multipart/form-data, got $contentType" }
```

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/DefaultTransform.kt
```diff
@@ -111,6 +112,28 @@ public fun HttpClient.defaultTransformers() {
...
+
+                val rawContentType = context.response.headers[HttpHeaders.ContentType]
+                    ?: error("No content type provided for multipart")
```
osipxd, line 125: Feels redundant as we already have `contentType`

osipxd (changes requested):
Good job!
Please check the comments below

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/DefaultTransform.kt
e5l: but it's string, I want to avoid serialization

ktor-http/ktor-http-cio/common/src/io/ktor/http/cio/MultipartCommon.kt
e5l: 👀

ktor-http/ktor-http-cio/jsAndWasmShared/src/io/ktor/http/cio/MultipartJsAndWasm.kt
e5l: 👀

osipxd (approved):
LGTM! Please, check code style before merge

PR #4462, opened 2024-11-07
KTOR-7722 content negotiation client accept header control

rocketraman:
**Subsystem**
Client, ContentNegotiation

**Motivation**
See https://youtrack.jetbrains.com/issue/KTOR-7722/ContentNegotiation-client-plugin-no-way-to-opt-out-of-Accept-on-a-per-request-basis.

**Solution**
Implement the two recommendations outlined in the previous issue.

1. Allow configuring the default q-value for registered content types. By default, for backward compatibility, the behavior is the same as today: no q-value is set. However, this config parameter allows users to set configure the default accept with a lower q setting. With this configuration, user's can easily override the preferred Accept on a per-request basis.
2. If the registered content type is already set on the request with additional parameters, the plugin does not add the accept header again.


marychatte (changes requested):
Thank you for the idea! I think it's better to introduce a new method `exclude(contentType)` instead. So it will look like:
```kotlin
install(ContentNegotiation) {
    register(ContentType.Application.Json, TestContentConverter())
}

client.get("https://test.com/") {
    accept(ContentType.Application.Pdf)
    exclude(ContentType.Application.Json)
}
```

rocketraman:
> Thank you for the idea! I think it's better to introduce a new method `exclude(contentType)` instead.

@marychatte  I can do that. However, it seems like that change could be made in addition to the changes I have already made rather than instead of. It would still seem useful to be able to set the default q-value, as well as the change to not add a duplicate accept type in the presence of an existing one with different parameters. Thoughts?

osipxd:
I'd say adding the same content type, but with parameters should simply replace the default one.
At least I can't come up with any use case when we would keep the default content type **and** the same content type with parameters. Is there any?

rocketraman:
> I'd say adding the same content type with parameters should simply replace the default one.
> At least I can't come up with any use case when we would keep the default content type **and** the same content type with parameters. Is there any?

Correct. Keeping the content type and adding the same content type with parameters is what the existing code does, and one of the items fixed by my PR. So therefore it seems like we agree this change should be made. 

The other change that is in the PR so far is the ability to set a default q value for the registered content types. Which I think is also useful.

rocketraman:
@marychatte I've added the `exclude` functionality which was requested. As per earlier comments, I've left the other functionality added as well. Please review.

osipxd:
> The other change is the ability to set a default q value for the registered content types.

Couldn't this be solved by adding a "q" value to the default content type?

```kotlin
install(ContentNegotiation) {
    register(ContentType.Application.Json.withParameter("q", "0.8"), TestContentConverter())
}
```

rocketraman:
> > The other change is the ability to set a default q value for the registered content types.
> 
> Couldn't this be solved by adding a "q" value to the default content type?
> 
> ```kotlin
> install(ContentNegotiation) {
>     register(ContentType.Application.Json.withParameter("q", "0.8"), TestContentConverter())
> }
> ```
> 
> However I agree it might look weird.

I thought about that, but there were a couple of reasons I didn't go with that approach.

First, the value that is passed to `register` is used in quite a few places. I saw that registering a content type with parameters would violate some assumptions the existing code was making (for example, [this line of code](https://github.com/Kotlin/ktor/blob/40c0095346c4dcf37982e4bee1bc46bd488de565/ktor-client/ktor-client-plugins/ktor-client-content-negotiation/common/src/io/ktor/client/plugins/contentnegotiation/ContentNegotiation.kt#L70)).

Second, most people don't call the `register` function directly, but instead use type-specific registration functions, [like this one](https://github.com/Kotlin/ktor/blob/babe511ac1561efe2e202e7e9dc7a36c19efc170/ktor-shared/ktor-serialization/ktor-serialization-kotlinx/ktor-serialization-kotlinx-json/common/src/io/ktor/serialization/kotlinx/json/JsonSupport.kt#L54-L59). I didn't want to complicate or change this API.

ktor-client/ktor-client-plugins/ktor-client-content-negotiation/common/src/io/ktor/client/plugins/contentnegotiation/ContentNegotiation.kt
```diff
@@ -29,6 +30,11 @@ internal val DefaultCommonIgnoredTypes: Set<KClass<*>> = setOf(
 
 internal expect val DefaultIgnoredTypes: Set<KClass<*>>
 
+/**
+ * Shows that request should skip auth and refresh token procedure.
```
marychatte, line 34: KDoc doesn't match. Can we fix it, please?

ktor-client/ktor-client-plugins/ktor-client-content-negotiation/common/src/io/ktor/client/plugins/contentnegotiation/ContentNegotiation.kt
```diff
@@ -29,6 +30,11 @@ internal val DefaultCommonIgnoredTypes: Set<KClass<*>> = setOf(
...
+ */
+public val ExcludeContentTypes: AttributeKey<List<ContentType>> = AttributeKey("exclude-content-types")
```
marychatte, line 36: minor: I would rename it to `ExcludedContentTypes` and it's better to use name `AttributeKey("ExcludedContentTypesAttr")` for consistency

ktor-client/ktor-client-plugins/ktor-client-content-negotiation/common/src/io/ktor/client/plugins/contentnegotiation/ContentNegotiation.kt
```diff
@@ -46,6 +52,12 @@ public class ContentNegotiationConfig : Configuration {
 
     internal val registrations = mutableListOf<ConverterRegistration>()
 
+    /**
+     * By default, `Accept` headers for registered content types will have no q value (implicit 1.0). Set this to
+     * change that behavior. This is useful to override the preferred `Accept` content types on a per-request basis.
+     */
+    public var defaultAcceptHeaderQValue: String? = null
```
marychatte, line 59: Can we make it `Double`?

ktor-client/ktor-client-plugins/ktor-client-content-negotiation/common/src/io/ktor/client/plugins/contentnegotiation/ContentNegotiation.kt
```diff
@@ -140,11 +152,25 @@ public val ContentNegotiation: ClientPlugin<ContentNegotiationConfig> = createCl
...
+        } else {
+            registrations
+        }
 
-            if (request.headers.contains(HttpHeaders.Accept, it.contentTypeToSend.toString())) return@forEach
-            request.accept(it.contentTypeToSend)
+        requestRegistrations.forEach {
+            val acceptHeaders = request.headers.getAll(HttpHeaders.Accept).orEmpty()
+            if (acceptHeaders.none { h -> ContentType.parse(h).match(it.contentTypeToSend) }) {
+                LOGGER.trace("Adding Accept=${it.contentTypeToSend.contentType} header with q=0.8 for ${request.url}")
```
marychatte, line 165: `header with q=0.8` -> `header with q=$qValue`

ktor-client/ktor-client-plugins/ktor-client-content-negotiation/common/src/io/ktor/client/plugins/contentnegotiation/ContentNegotiation.kt
```diff
@@ -140,11 +152,25 @@ public val ContentNegotiation: ClientPlugin<ContentNegotiationConfig> = createCl
...
+            val excluded = request.attributes[ExcludeContentTypes]
+            registrations.filter { registration -> excluded.none { registration.contentTypeToSend.match(it) } }
+        } else {
+            registrations
+        }
 
-            if (request.headers.contains(HttpHeaders.Accept, it.contentTypeToSend.toString())) return@forEach
-            request.accept(it.contentTypeToSend)
+        requestRegistrations.forEach {
+            val acceptHeaders = request.headers.getAll(HttpHeaders.Accept).orEmpty()
```
marychatte, line 163: We can move `val acceptHeaders =` before `requestRegistrations.forEach {`

marychatte (changes requested):
@rocketraman Thanks for the changes! I wrote some comments, but the general idea is good
 @osipxd What do you think about the name `defaultAcceptHeaderQValue `?

marychatte:
> First, the value that is passed to register is used in quite a few places. I saw that registering a content type with parameters would violate some assumptions the existing code was making (for example, [this line of code](https://github.com/Kotlin/ktor/blob/40c0095346c4dcf37982e4bee1bc46bd488de565/ktor-client/ktor-client-plugins/ktor-client-content-negotiation/common/src/io/ktor/client/plugins/contentnegotiation/ContentNegotiation.kt#L70)).

I think it's a bug in the code, and it should be:
```kotlin
val matcher = when  {
    contentType.match(ContentType.Application.Json) -> JsonContentTypeMatcher
    else -> defaultMatcher(contentType)
}
```
Can we fix it in this PR as well?

rocketraman:
@marychatte Thanks for the review. I've pushed 3 additional commits.

marychatte (approved):
LGTM

PR #4479, opened 2024-11-13
Fix Unix Sockets on Windows

brian-mcnamara:
**Subsystem**
Network

**Motivation**
I am working with Unix domain sockets (client), and using ktor to communicate between processes via the socket. When I switched to testing on Windows, I noticed that when I tried to connect to the socket, it would instantly close. Upon further digging, the difference appears to be in SocketImpl, upon calling channel.connect, Windows (at least my machine) returns false, which brings it into a blocking condition until the connection is connected. However, as part of that block, the selfConnect would return true since both local and remote addresses are not of type InetSocketAddress, which results in the return condition returning true, and thus closing the connection. leading to a connection closed exception

**Solution**
Since there was a comment which indicated that selfConnect was for TCP over the network, I simply added a check to validate the local and remote are both not InetSocketAddress, which would be the case for Unix sockets. 

Note: 
TCPSocketTest.testEchoOverUnixSockets was failing without this change on my machine (which I had to dig around to figure out how to switch the runtime JVM version so the test would actually run. Note for anyone in the future, the jvm version is set in JvmConfig.kt, see Test.configureJavaLauncher extension function)


e5l (changes requested):
Hey @brian-mcnamara, thanks for the PR!
Could you please check the compilation error for the task `linkDebugTestMingwX64`?

brian-mcnamara:
Hey @e5l , thanks for taking a look. Can we retry rerunning CI? I'm seeing no reason this change would cause a compile error

brian-mcnamara:
Failed again, I will work on getting mingw working locally

e5l (approved):

e5l:
Hey @brian-mcnamara, thank you for the PR! LGTM

PR #4481, opened 2024-11-14
KTOR-7359 Implement a suspending version of EmbeddedServer.start and EmbeddedServer.stop

whyoleg:
**Subsystem**
Server

**Motivation**
* fixes https://youtrack.jetbrains.com/issue/KTOR-7459
* alternative to https://youtrack.jetbrains.com/issue/KTOR-2939
* prerequisite for https://github.com/ktorio/ktor/pull/4466

**Solution**

Only `startSuspend` and `stopSuspend` functions were added with minimal bridge for JVM and Native implementations.
That was done like this, because in most cases users will anyway call `start` and `stop`, and it's not clear, which engines could support `suspend` start/stop (e.g CIO can, for sure, but for others it's not that clear).
That's why minimally invasive changes were implemented to support wasm-js and js case, where `runBlocking` is not available.

PR also contains changes to `EngineTestBase` for js and wasm-js as this is the main consumer for those new APIs.

Note: probably some commonization is possible for both `EmbeddedServer` and `EngineTestBase`, it should be investigated.

### DIFF
--- a/ktor-server/ktor-server-test-base/jsAndWasmShared/src/io/ktor/server/test/base/EngineTestBase.jsAndWasmShared.kt
+++ b/ktor-server/ktor-server-test-base/jsAndWasmShared/src/io/ktor/server/test/base/EngineTestBase.jsAndWasmShared.kt
@@ -53,7 +53,7 @@ actual constructor(
     @OptIn(DelicateCoroutinesApi::class)
     @AfterTest
     fun tearDownBase() {
-        GlobalScope.launch { server?.stopSuspend(0, 500) }
+        GlobalScope.launch { server?.stopSuspend(gracePeriodMillis = 0, timeoutMillis = 500) }
     }
 
     protected actual suspend fun createAndStartServer(
--- a/ktor-server/ktor-server-test-base/jsAndWasmShared/src/io/ktor/server/test/base/EngineTestBase.jsAndWasmShared.kt
+++ b/ktor-server/ktor-server-test-base/jsAndWasmShared/src/io/ktor/server/test/base/EngineTestBase.jsAndWasmShared.kt
@@ -16,6 +16,7 @@ import io.ktor.server.testing.*
 import io.ktor.util.logging.*
 import kotlinx.coroutines.*
 import kotlin.coroutines.*
+import kotlin.test.*
 import kotlin.time.Duration.Companion.seconds
 
 actual abstract class EngineTestBase<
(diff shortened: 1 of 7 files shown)


ktor-server/ktor-server-test-base/jsAndWasmShared/src/io/ktor/server/test/base/EngineTestBase.jsAndWasmShared.kt
```diff
@@ -32,14 +33,35 @@ actual constructor(
...
     protected actual annotation class Http2Only actual constructor()
 
-    protected actual var port: Int = 0
+    /**
+     * It's not possible to find a free port during test setup,
+     * as on JS (Node.js) all APIs are non-blocking (suspend).
+     * That's why we assign port after the server is started in [startServer]
+     * Note: this means, that [port] can be used only after calling [createAndStartServer] or [startServer].
+     */
+    private var _port: Int = 0
```
e5l, line 42: Please consider extracting a constant for non initialized value and use a different value (other than default port)

ktor-server/ktor-server-test-base/jsAndWasmShared/src/io/ktor/server/test/base/EngineTestBase.jsAndWasmShared.kt
```diff
@@ -101,7 +125,8 @@ actual constructor(
         // we start it on the global scope because we don't want it to fail the whole test
         // as far as we have retry loop on call side
         val starting = GlobalScope.async {
-            server.start(wait = false)
+            server.startSuspend(wait = false)
```
e5l, line 128: Should we make an extension function here like `server.listeningPorts()`

e5l (approved):
could you please check the CI?

e5l:
Please check the code style problems

ktor-server/ktor-server-test-base/jsAndWasmShared/src/io/ktor/server/test/base/EngineTestBase.jsAndWasmShared.kt
whyoleg: done, used -1

ktor-server/ktor-server-test-base/jsAndWasmShared/src/io/ktor/server/test/base/EngineTestBase.jsAndWasmShared.kt
whyoleg: I guess that the commend was for this line:
```
_port = server.engine.resolvedConnectors().first().port
```

I don't think, that we do really need this, as this is rather specific and mostly needed in tests where we need to launch both server and client (e.g ktor, or in projects like rsocket-kotlin)

osipxd (approved):
LGTM. Could you rebase the PR one more time? 🙏

PR #4482, opened 2024-11-15
KTOR-7777 Release proper InetSocketAddress

divinae:
**Subsystem**
Client

**Motivation**
We encountered a lot of crashes after updating to the Ktor v3 with the root causes:
```Caused by: java.lang.IllegalStateException: The number of released permits cannot be greater than 100```
 
The problem is that we request 4 calls to the same endpoint asynchronously. However, InetSocketAddress(host, port) could have different results (IP address could change) - the current solution stores only the last address.

**Solution**
Store InetSocketAddress separately per connect request.

### DIFF
--- a/ktor-client/ktor-client-cio/jvmAndPosix/src/io/ktor/client/engine/cio/Endpoint.kt
+++ b/ktor-client/ktor-client-cio/jvmAndPosix/src/io/ktor/client/engine/cio/Endpoint.kt
@@ -38,8 +38,6 @@ internal class Endpoint(
     private val deliveryPoint: Channel<RequestTask> = Channel()
     private val maxEndpointIdleTime: Long = 2 * config.endpoint.connectTimeout
 
-    private var connectionAddress: InetSocketAddress? = null
-
     private val timeout = launch(coroutineContext + CoroutineName("Endpoint timeout($host:$port)")) {
         try {
             while (true) {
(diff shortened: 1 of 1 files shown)


osipxd (commented):
Thank you for the fix! I've created an issue not to miss this change from release notes:
- [KTOR-7777](https://youtrack.jetbrains.com/issue/KTOR-7777) Connections aren't released properly if there are multiple parallel connections to the same address

Do you think you can write tests covering this issue?

divinae:
> Thank you for the fix! I've created an issue not to miss this change from release notes:
> - [KTOR-7777](https://youtrack.jetbrains.com/issue/KTOR-7777) Connections aren't released properly if there are multiple parallel connections to the same address
> 
> Do you think you can write tests covering this issue?

I could try on Monday.

divinae:
> > Thank you for the fix! I've created an issue not to miss this change from release notes:
> > 
> > * [KTOR-7777](https://youtrack.jetbrains.com/issue/KTOR-7777) Connections aren't released properly if there are multiple parallel connections to the same address
> > 
> > Do you think you can write tests covering this issue?
> 
> I could try on Monday.

I'm sorry, but unfortunately, I am not competent to write these tests. I don't know how to mock socket interactions, etc.

e5l (approved):

e5l:
Hey @divinae, thanks for the PR. LGTM. We will add tests

divinae:
> Hey @divinae, thanks for the PR. LGTM. We will add tests

Thanks! When could I expect a new release with a fix?

osipxd:
This change is going to be included in the next patch release (v3.0.2). It will be released at the end of the month.

PR #4486, opened 2024-11-18
KTOR-7795 Use version catalogs everywhere

osipxd:
**Subsystem**
Build Infrastructure

**Motivation**
[KTOR-7795](https://youtrack.jetbrains.com/issue/KTOR-7795) Use version catalogs for all dependencies

**Solution**
Find and replace all places where dependencies were added using string notation.


bjhham (approved):
Beautiful 🥲

PR #4493, opened 2024-11-19
KTOR-4534 Add retry for flaky tests

e5l:
[KTOR-4534](https://youtrack.jetbrains.com/issue/KTOR-4534) Fix Flaky Tests


e5l:
It looks like they are flaky because of bad timeouts. Let's keep retry to make the build green

PR #4494, opened 2024-11-19
KTOR-7683 Fix for empty response with respondSource

bjhham:
**Subsystem**
Server, I/O

**Motivation**
[KTOR-7683](https://youtrack.jetbrains.com/issue/KTOR-7683) call.respondSource returns empty response but passes in tests

**Solution**
Turns out, the problem was using the `Source.remaining` extension value as the content length, which only checks the size of the buffer, which starts as 0 when wrapping an `InputStream` with the `asSource()` extension.

The fix here is adding an optional `contentLength` parameter on `respondSource`, so the responsibility is delegated to the user.  I also generalized these functions to use `RawSource` to reflect the `kotlinx-io` API, and allow for non-buffered sources.


ktor-server/ktor-server-core/common/src/io/ktor/server/response/ApplicationResponseFunctions.kt
```diff
@@ -157,13 +157,15 @@ public suspend fun ApplicationCall.respondBytes(
...
  */
 public suspend fun ApplicationCall.respondSource(
-    source: Source,
+    source: RawSource,
     contentType: ContentType? = null,
     status: HttpStatusCode? = null,
+    contentLength: Long? = null,
 ) {
-    respond(ChannelWriterContent({ writeBuffer(source) }, contentType, status, source.remaining))
+    respond(ChannelWriterContent({ writeBuffer(source) }, contentType, status, contentLength))
```
marychatte, line 168: Can it break something that we pass `contentLength` instead `source.remaining` now, when `contentLength` is `null`?

marychatte (approved):
lgtm

PR #4499, opened 2024-11-21
Add ability to retry runTest on backend

e5l:


bjhham (approved):
👍 👍

e5l:
it looks like server is not restarted between repeats, let me check if I can fix it

osipxd:
This failure seems caused by the latest Tomcat update: https://github.com/ktorio/ktor/pull/4470

e5l:
Thanks, will check what has been changed

ktor-server/ktor-server-test-base/common/src/io/ktor/server/test/base/BaseTest.kt
```diff
@@ -4,13 +4,33 @@
...
 expect abstract class BaseTest() {
     open val timeout: Duration
+
+    open fun beforeTest()
+    open fun afterTest()
+
     fun collectUnhandledException(error: Throwable) // TODO: better name?
-    fun runTest(block: suspend CoroutineScope.() -> Unit): TestResult
-    fun runTest(timeout: Duration, block: suspend CoroutineScope.() -> Unit): TestResult
+    fun runTest(timeout: Duration = 60.seconds, block: suspend CoroutineScope.() -> Unit): TestResult
```
osipxd, line 19: I'm not sure about such a big default timeout. Successful tests usually run faster.

osipxd (approved):

ktor-server/ktor-server-test-base/common/src/io/ktor/server/test/base/BaseTest.kt
e5l: Could you tell me if you have any value in mind?

osipxd: According to [our Develocity statistics](https://ge.jetbrains.com/scans/tests?search.rootProjectNames=ktor&tests.container=*.server.*&tests.sortField=MEAN_DURATION), 30 seconds should be enough for all tests.

PR #4500, opened 2024-11-22
KTOR-7831 Configure Gradle Daemon JDK

osipxd:
**Subsystem**
Infrastructure

**Motivation**
[KTOR-7831](https://youtrack.jetbrains.com/issue/KTOR-7831) Use Gradle JVM toolchain for Daemon
To make builds more reproducible and independent of the environment in which they’re run.

**Solution**
- Specify Daemon JVM in `gradle/gradle-daemon-jvm.properties`
- Migrate test server to `jetty-jakarta`

**Depends on:**

- https://github.com/ktorio/ktor/pull/4502
- https://github.com/ktorio/ktor-build/pull/104
- https://github.com/e5l/ktor-test-image/pull/3

The only problem is that Gradle Daemon toolchains aren’t properly supported in IDEA, so it is required to manually configure the IDEA to use JDK 21 in the "Project Structure" window:
- https://youtrack.jetbrains.com/issue/IDEA-358562/Support-Gradle-Daemon-Toolchains-in-UI

_It is better to review commit-by-commit_


osipxd:
JS builds should become green after https://github.com/e5l/ktor-test-image/pull/3 merged and published.

bjhham (approved):
Screenshot 2024-12-12 at 10 43 54

e5l (approved):
lgtm, btw we might try netty as well

PR #4503, opened 2024-11-25
KTOR-7845 Fix for threading issue in flushAndClose for reader job channels

bjhham:
**Subsystem**
Core, I/O

**Motivation**
[KTOR-7845](https://youtrack.jetbrains.com/issue/KTOR-7845) File is not commited after closing writeChannel() of the file

**Solution**
The problem here is that `flushAndClose()` doesn't wait for the `ReaderJob` to finish so it has no effect on the final destination of the data.  By adding a hook to wait for the job to complete, we can daisy chain these jobs with some trust that `flushAndClose()` will work properly from the perspective of the caller.

For reference, here is the `writeChannel` utility function:

```kotlin
@OptIn(DelicateCoroutinesApi::class)
public fun File.writeChannel(
    coroutineContext: CoroutineContext = Dispatchers.IO
): ByteWriteChannel = GlobalScope.reader(CoroutineName("file-writer") + coroutineContext, autoFlush = true) {
    @Suppress("BlockingMethodInNonBlockingContext")
    RandomAccessFile(this@writeChannel, "rw").use { file ->
        val copied = channel.copyTo(file.channel)
        file.setLength(copied) // truncate tail that could remain from the previously written data
    }
}.channel
```

### DIFF
--- a/ktor-io/api/ktor-io.api
+++ b/ktor-io/api/ktor-io.api
@@ -173,6 +173,10 @@ public abstract interface class io/ktor/utils/io/ChannelJob {
 	public abstract fun getJob ()Lkotlinx/coroutines/Job;
 }
 
+public final class io/ktor/utils/io/CloseHookByteWriteChannelKt {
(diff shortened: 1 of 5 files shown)


e5l (changes requested):
Please change the target branch to 3.1.0

e5l (approved):

PR #4505, opened 2024-11-25
KTOR-7828 Revert removed content-length check

e5l:
[KTOR-7828](https://youtrack.jetbrains.com/issue/KTOR-7828) CIO: Response body truncated because read amount of bytes isn't compared against Content-Length


bjhham (approved):

osipxd:
Let's update Kotlin in a separate PR:
- https://github.com/ktorio/ktor/pull/4512

e5l:
sure, I just can't merge this because of failing build. We need to wait for the new ktlint version

PR #4507, opened 2024-11-26
Fix KTOR-7665 Jackson should respond with 500 when Jackson converter fails with exception

marychatte:
**Subsystem**
Server, Content Negotiation

**Motivation**
[KTOR-7665](https://youtrack.jetbrains.com/issue/KTOR-7665)


marychatte:
After many attempts to understand why it fails, I understand that it works fine and as designed :)
More information in the [comments](https://youtrack.jetbrains.com/issue/KTOR-7665)

PR #4519, opened 2024-11-29
KTOR-7830 HttpCache may cache response before its transformation

osipxd:
**Subsystem**
Client

**Motivation**
[KTOR-7830](https://youtrack.jetbrains.com/issue/KTOR-7830) Installing HttpCache before ContentEncoding prevents response body to be decoded

**Solution**
The problem is that both `HttpCache` and `ContentEncoding` use phase `HttpReceivePipeline.State` to handle responses, so installation order matters here. I've changed `HttpCache` plugin to operate in an additional phase after `State` phase, when all response transformations should be applied.


bjhham (approved):
Nice fix!  The unexpected interactions between plugins is a never-ending source of stress.

PR #4526, opened 2024-12-03
KTOR-7876 Fix Sink.writeText corrupts long streams for Jetty server

marychatte:
**Subsystem**
Server
**Motivation**
[KTOR-7876 Jetty: respondBytesWriter with writeStringUtf8 corrupts long streams containing non-ascii chars](https://youtrack.jetbrains.com/issue/KTOR-7876)

### DIFF
--- a/ktor-io/common/src/io/ktor/utils/io/core/Strings.kt
+++ b/ktor-io/common/src/io/ktor/utils/io/core/Strings.kt
@@ -89,7 +89,7 @@ public fun Source.readTextExactCharacters(charactersCount: Int, charset: Charset
 }
 
 /**
- * Writes [text] characters in range \[[fromIndex] .. [toIndex]) with the specified [charset]
+ * Writes [text] characters in range [fromIndex] .. [toIndex]) with the specified [charset]
  */
 public fun Sink.writeText(
     text: CharSequence,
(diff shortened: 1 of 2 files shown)


ktor-io/common/src/io/ktor/utils/io/core/Strings.kt
```diff
@@ -89,18 +89,14 @@ public fun Source.readTextExactCharacters(charactersCount: Int, charset: Charset
 }
 
 /**
- * Writes [text] characters in range \[[fromIndex] .. [toIndex]) with the specified [charset]
+ * Writes [text] characters in range [fromIndex] .. [toIndex]) with the specified [charset]
```
e5l, line 92: Could you check the KDoc and mention explicitly included and excluded part of the ranges?

ktor-io/common/src/io/ktor/utils/io/core/Strings.kt
```diff
@@ -89,18 +89,14 @@ public fun Source.readTextExactCharacters(charactersCount: Int, charset: Charset
...
  */
 public fun Sink.writeText(
     text: CharSequence,
     fromIndex: Int = 0,
     toIndex: Int = text.length,
     charset: Charset = Charsets.UTF_8
 ) {
-    if (charset === Charsets.UTF_8) {
-        return writeString(text.toString(), fromIndex, toIndex)
```
e5l, line 101: Could you tell me why whiteStrings corrupts non ASCII? Should we log this to the kotlinx-io instead?

e5l (changes requested):

marychatte:
I rechecked the problem, it wasn't on the Ktor side, and this anomaly doesn't appear anymore

PR #4531, opened 2024-12-04
 KTOR-6501 Add feedback link to KDocs

e5l:
Fix https://youtrack.jetbrains.com/issue/KTOR-6501


ktor-client/ktor-client-android/jvm/src/io/ktor/client/engine/android/Android.kt
```diff
@@ -25,6 +25,7 @@ import io.ktor.client.engine.*
  * ```
  *
  * You can learn more about client engines from [Engines](https://ktor.io/docs/http-client-engines.html).
+ * [Report a problem](https://ktor.io/feedback?fqname=io.ktor.client.engine.android.Android)
```
osipxd, line 28: Should we add a blank line between "Report a problem" and a comment body? To make it displayed on a separate line.

e5l: sure, let me fix the generator

e5l: @osipxd, please check.

e5l: The generator can also be used on CI checks and it takes about 2 seconds to run on the whole code base

ktor-utils/common/src/io/ktor/util/Attributes.kt
```diff
@@ -12,6 +12,8 @@ import kotlin.reflect.*
  * Specifies a key for an attribute in [Attributes]
  * @param T is a type of the value stored in the attribute
  * @param name is a name of the attribute for diagnostic purposes. Can't be blank
+ *
+ * [Report a problem](https://ktor.io/feedback?fqname=io.ktor.util.AttributeKey)
```
osipxd, line 16: This is rendered as a link in a parameter description as it is placed under `@param` tags

image

e5l: @olme04, could you suggest is there any way to put it on the bottom line of the quick help?

whyoleg: I don't think it's possible. In this case `[Report a problem]` should be above tags, as `tags` could be only written in the end of the KDoc block.

May be this could be done inside ktor IDEA plugin?

ktor-utils/jvm/src/io/ktor/util/reflect/ServiceLoader.kt
```diff
@@ -13,6 +13,9 @@ import java.util.*
 
 /**
  * Loads all implementations of the service [T] using [ServiceLoader.load].
+ *
+ * [Report a problem](https://ktor.io/feedback?fqname=io.ktor.util.reflect.loadServices)
+ *
  * @see loadServiceOrNull
  */
 @InternalAPI
```
osipxd, line 21: Should we skip `@InternalApi` declarations from the generator?

ktor-utils/common/src/io/ktor/util/Attributes.kt
e5l: fixed within generator

ktor-utils/jvm/src/io/ktor/util/reflect/ServiceLoader.kt
e5l: I think it's ok if they have KDoc written. At the moment, I'm skipping only public without KDocs

ktor-client/ktor-client-core/posix/src/io/ktor/client/HttpClient.kt
```diff
@@ -12,6 +12,8 @@ import io.ktor.utils.io.*
  *
  * The [HttpClientEngine] is selected from the dependencies.
  * https://ktor.io/docs/http-client-engines.html
+ *
+ * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.client.HttpClient)
```
osipxd, line 16: Minor. Should we distinct "fqname" for a class `HttpClient` and a function `HttpClient`?

osipxd (commented):
👍 
The only question bothering me is why the "Report a problem" button is generated not for all class members.

bjhham:
It would be real cool if we could generate this through some KDoc extension instead of adding it everywhere in the code.

bjhham:
Looks like you've incurred the wrath of CodeStyle for touching all the files 😄

e5l:
@bjhham it is better to have it in the source code, cause the link will also be available through quick help of IDE

build-logic/src/main/kotlin/ktorbuild/CInterop.kt
```diff
@@ -41,6 +41,8 @@ import org.jetbrains.kotlin.gradle.plugin.mpp.KotlinNativeTarget
  *     )
  * }
  * ```
+ *
+ * [Report a problem](https://ktor.io/feedback/?fqname=ktorbuild.createCInterop)
```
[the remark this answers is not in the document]
e5l: Thanks, fixed

ktor-client/ktor-client-cio/common/src/io/ktor/client/engine/cio/CIOEngineConfig.kt
```diff
@@ -10,10 +10,14 @@ import io.ktor.network.tls.*
...
+ *
+ * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.client.engine.cio.CIOEngineConfig)
  */
 public class CIOEngineConfig : HttpClientEngineConfig() {
     /**
      * Provides access to [Endpoint] settings.
+     *
+     * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.client.engine.cio.CIOEngineConfig.endpoint)
      */
     public val endpoint: EndpointConfig = EndpointConfig()
```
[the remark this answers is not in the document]
e5l: Fixed

bjhham (approved):

ktor-client/ktor-client-core/posix/src/io/ktor/client/HttpClient.kt
e5l: We may include the line number, could you tell me what you think?

osipxd: I think we can solve this problem later, if necessary

e5l: Sure, I added the line numbers to the lib: https://github.com/e5l/kdoc-annotator

PR #4532, opened 2024-12-06
KTOR-7934 Fix iOS, JS, WASM, Curl fails with IllegalStateException: Content-Lengt…

e5l:
Fix [KTOR-7934 JS/WASM fails with "IllegalStateException: Content-Length mismatch" on requesting gzipped content](https://youtrack.jetbrains.com/issue/KTOR-7934)


ktor-client/ktor-client-plugins/ktor-client-encoding/common/test/ContentEncodingTest.kt
```diff
@@ -121,4 +122,12 @@ class ContentEncodingTest : ClientLoader() {
...
         }
     }
+
+    @Test
+    fun testGzipWithContentLengthWithoutPlugin() = clientTests {
+        test { client ->
+            client.get("$TEST_URL/gzip-with-content-length")
+                .bodyAsBytes()
+        }
+    }
```
bjhham, line 132: Looks like this needs an assert.

ktor-client/ktor-client-core/jsAndWasmShared/src/io/ktor/client/utils/HeadersUtils.kt
```diff
@@ -0,0 +1,20 @@
...
+ * The body is already decompressed for wasm and js engine,
+ * so we're dropping these headers
+ * to make sure all the plugins and checks work with correct content length and encoding.
+ */
+internal fun HeadersBuilder.dropCompressionHeaders() {
+    if (contains(HttpHeaders.ContentEncoding)) {
+        remove(HttpHeaders.ContentEncoding)
+        remove(HttpHeaders.ContentLength)
+    }
+}
```
bjhham, line 20: This could interfere with any other introspection on the response headers... Maybe it would be a better solution to introduce some flag to the engine interface to indicate implicit decompression for the client to check.

bjhham (approved):
Interesting issue 👍

ktor-client/ktor-client-core/jsAndWasmShared/src/io/ktor/client/utils/HeadersUtils.kt
e5l: The problem is with content-length header has wrong value after browser decoding. Let's check it as a separate issue

PaulWoitaschek:
> KTOR-7934 Fix JS/WASM

This Youtrack indicates that this is not just JS/WASM, so you might want to change the title to reflect this fixes an iOS bug too:
https://youtrack.jetbrains.com/issue/KTOR-7943

GazimSoliev:
this error happens cuz of HttpResponseValidator in my case

PR #4534, opened 2024-12-06
KTOR-7682 Fix TestApplication.stop() doesn't stop the application anymore

osipxd:
**Subsystem**
Test Infrastructure

**Motivation**
[KTOR-7682 TestApplication.stop() doesn't stop the application anymore](https://youtrack.jetbrains.com/issue/KTOR-7682)

**Solution**
The problem occurs because `TestApplication.client` uses an instance of an application different from the returned by `TestApplication` constructor-like function and calls `start()` on it. `TestApplication.start()` calls `TestApplicationEngine.start()` under the hood which rewrites `cancellationJob`, so coroutine waiting for cancellation never ends.

To fix this problem:
- make `TestApplication` return `application` from builder – the same as used in a client
- remove builder from `TestApplication` constructor to make builder the only way to get `TestApplication` and break circular dependency: `TestApplication` -> `ApplicationTestBuilder` -> `TestApplication`

_It is better to review commit-by-commit_


bjhham (approved):
Nice!

PR #4541, opened 2024-12-10
KTOR-7946 Fix FormFieldLimit overwritten by default arg

bjhham:
**Subsystem**
Server, Core

**Motivation**
[KTOR-7946](https://youtrack.jetbrains.com/issue/KTOR-7946) FormFieldLimit is overwritten by default arg

**Solution**
Change default parameter to use null and defer to the attribute.


ktor-server/ktor-server-core/common/src/io/ktor/server/request/ApplicationReceiveFunctions.kt
```diff
@@ -188,9 +188,11 @@ public var ApplicationCall.formFieldLimit: Long
  * @throws ContentTransformationException when content cannot be transformed to the [MultiPartData].
  */
 public suspend inline fun ApplicationCall.receiveMultipart(
-    formFieldLimit: Long = DEFAULT_FORM_FIELD_MAX_SIZE
```
e5l, line 191: it looks like we need old method with hidden for backward compatibility

bjhham: Isn't it the same bytecode to change an arg to nullable?  I think don't think it could break anything...

e5l: it changes L to Long, so yep, it's different

osipxd: We could use `-1` instead of `null` here to keep this function backward compatible

bjhham: Oh yeah that could work 👍

osipxd (approved):
LGTM! It would be great to have a test for this case

PR #4542, opened 2024-12-10
Update contributing guide

osipxd:
- During a test of our contributing guide on Ubuntu Noble (24.04LTS) I found that `libtinfo-dev` [doesn't exist](https://packages.ubuntu.com/noble/libtinfo-dev) for this Ubuntu version. Actually, we don't need to install it explicitly as it is already included in `libncurses-dev`
- After merging #4500 it is required to have JDK 21 installed to build the project.


osipxd:
@bsautner, could you take a look, please?

bjhham (approved):

PR #4548, opened 2024-12-12
Refactor ClientLoader.clientTests

osipxd:
**Subsystem**
Test Infrastructure

**Motivation**
[KTOR-7967](https://youtrack.jetbrains.com/issue/KTOR-7967) Make client tests independent

**Solution**
There were several issues with the `clientTests` function:
- On JS/WasmJS platform Mocha timeout was applied, so it was impossible to set a bigger timeout for tests individually
- The configured timeout was applied to the whole test, including different clients and retries. So after the timeout error, there was no sense in retrying,

Refactored client tests to apply timeout to each attempt individually.
The main difficulty of this refactor is JS/WasmJS support. `runTest` returns `TestResult`, which is mapped to `Promise<Unit>` on JS platform, so we can't just retry the test in a loop. For JS we should use `Promise.then { ... }` and `Promise.catch { ... }`.

**Depends on:**
- #4550


bjhham (approved):
Nice 🥇

PR #4553, opened 2024-12-16
KTOR-7584 HTMX extension for Ktor

bjhham:
**Subsystem**
Server, New plugin

**Motivation**
[KTOR-7584](https://youtrack.jetbrains.com/issue/KTOR-7584) First-class HTMX support

**Solution**
This extends the routing and HTML builder DSL with some handy HTMX helpers.  I tried to ere on the conservative side here by not including everything and adding an experimental flag so we can change things later after testing.  We don't have the HTML fragments extension yet in the HTML DSL, but you can still do everything you need using valid HTML and attributes.


ktor-shared/ktor-htmx/common/src/io/ktor/htmx/HxAttributeKeys.kt
```diff
@@ -0,0 +1,183 @@
+/*
+ * Copyright 2014-2024 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+@file:Suppress("ConstPropertyName")
+
+package io.ktor.htmx
+
+public object HxAttributeKeys {
+    /**
+     * issues a GET to the specified URL
```
osipxd, line 10: Here and in some other files KDocs are starting with a lowercase letter. Could you fix it?
I also think that we could add a link to the official docs to the KDoc of this class: https://htmx.org/reference/#attributes-additional

ktor-server/ktor-server-plugins/ktor-server-htmx/common/src/io/ktor/server/htmx/HxHeaders.kt
```diff
@@ -0,0 +1,60 @@
...
+    public val triggerName: String? get() = headers[HxRequestHeaders.TriggerName]
+}
+
+@ExperimentalHtmxApi
+public class HXResponseHeaders(private val headers: ResponseHeaders) : StringMap {
+
+    public var location: String? by HxResponseHeaders.Location
+    public var pushUrl: String? by HxResponseHeaders.PushUrl
+    public var redirect: String? by HxResponseHeaders.Redirect
+    public var refresh: String? by HxResponseHeaders.Refresh // TODO boolean
```
osipxd, line 49: Should we make it boolean now? Or `ExperimentalHtmxApi` allows breaking changes? 😅

ktor-server/ktor-server-plugins/ktor-server-htmx/common/src/io/ktor/server/htmx/HxAttributes.kt
```diff
@@ -0,0 +1,53 @@
...
+    public var target: String? by HxAttributeKeys.Target
+    public var trigger: String? by HxAttributeKeys.Trigger
+    public var vals: String? by HxAttributeKeys.Vals
+
+    public val on: On
+        get() = On(map)
+
+    public fun on(event: String, script: String) {
+        map["hx-on:$event"] = script
+    }
```
osipxd, line 39: I don't remember what we decided here. Keep both options of syntax?
```kotlin
on["event] = "script"
on("event", "script")
```

ktor-server/ktor-server-plugins/ktor-server-htmx/common/src/io/ktor/server/htmx/HxAttributes.kt
```diff
@@ -0,0 +1,53 @@
...
+
+@ExperimentalHtmxApi
+@HtmlTagMarker
+public class HxAttributes(override val map: DelegatingMap) : StringMapDelegate {
+    public var get: String? by HxAttributeKeys.Get
+    public var post: String? by HxAttributeKeys.Post
+    public var pushUrl: String? by HxAttributeKeys.PushUrl
+    public var select: String? by HxAttributeKeys.Select
+    public var selectOob: String? by HxAttributeKeys.SelectOob
+    public var swap: String? by HxAttributeKeys.Swap
```
osipxd, line 28: We've had type-safe implementations for some of these fields. Will it be part of some future PR?

Mr3zee (approved):

ktor-server/ktor-server-plugins/ktor-server-htmx/common/src/io/ktor/server/htmx/HxRouting.kt
```diff
@@ -0,0 +1,44 @@
...
+public val Route.hx: HXRoute get() = HXRoute(this)
+
+/**
+ * Scope child routes to apply when `HX-Request` header is supplied.
+ */
+@ExperimentalHtmxApi
+public fun Route.hx(configuration: HXRoute.() -> Unit): Route = with(HXRoute(this)) {
+    header(HxRequestHeaders.Request, "true") {
+        configuration()
+    }
```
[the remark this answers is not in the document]
bjhham: The compiler doesn't like this unfortunately.  I'll just make the constructor internal so the logic is enforced by only allowing access through the helper.

ktor-server/ktor-server-plugins/ktor-server-htmx/common/src/io/ktor/server/htmx/HxRouting.kt
```diff
@@ -0,0 +1,51 @@
...
+public value class HxRoute internal constructor(private val route: Route) : Route by route {
+    internal companion object {
+        internal fun wrap(route: Route) =
+            HxRoute(route.createChild(HttpHeaderRouteSelector(HxRequestHeaders.Request, "true")))
+    }
+
+    /**
+     * Sub-routes only apply to a specific HX-Target header.
+     */
+    public fun target(expectedTrigger: String, body: Route.() -> Unit): Route =
```
osipxd, line 43:
```suggestion
    public fun target(expectedTarget: String, body: Route.() -> Unit): Route =
```

ktor-utils/common/src/io/ktor/util/collections/MapDelegates.kt
```diff
@@ -0,0 +1,79 @@
...
+public interface StringMapDelegate : StringMap {
+    public val map: MutableMap<String, String>
+
+    override fun set(key: String, value: String): Unit = map.set(key, value)
+    override fun get(key: String): String? = map[key]
+    override fun remove(key: String): String? = map.remove(key)
+}
+
+/**
+ * Simplifies property access delegation for HxAttributes when setting attribute values from a string constant.
```
osipxd, line 27: Minor. Probably we could drop "HxAttributes" from the description?

osipxd (approved):

ulrikandersen:
This is amazing. Nice work! 🎊 

> We don't have the HTML fragments extension yet in the HTML DSL

@bjhham Do you know if that is being tracked anywhere? It would be really useful with the HTMX plugin where we often respond with just a `div` etc.

osipxd:
> Do you know if that is being tracked anywhere?

Here it is: [KTOR-8195](https://youtrack.jetbrains.com/issue/KTOR-8195) Partial HTML response

PR #4556, opened 2024-12-17
KTOR-7963 Handle ClosedWatchServiceException

bjhham:
**Subsystem**
Server

**Motivation**
[KTOR-7963](https://youtrack.jetbrains.com/issue/KTOR-7963) Uncaught ClosedWatchServiceException exception thrown by finalizer when closing the server

**Solution**
Catch exceptions from polling.  I was unable to reproduce this reliably in a unit test, it seems to be pretty random with timing.


osipxd (approved):
👍

PR #4566, opened 2024-12-19
KTOR-7723 Improve clientTests DX

osipxd:
**Subsystem**
Test Infrastructure

**Motivation**
1. [KTOR-7723](https://youtrack.jetbrains.com/issue/KTOR-7723) Engine exclusion from clientTests is confusing
2. Test execution time reported by Gradle is not true. On JVM it counts time spent on the whole test including all engines and retries, on JS it counts Promise creation time.

**Solution**
- Changed syntax to exclude engines from tests
- Added test status reporting for each try with test duration

image


bjhham (approved):
Beautiful 🥲

PR #4579, opened 2025-01-03
KTOR-8061 MicrometerMetricsConfig add UptimeMetrics to standard meterBinders

SIMULATAN:
**Subsystem**
Server - Plugin Micrometer

**Motivation**
Lots of standard dashboards include the JVM process uptime / start time. The default Ktor Micrometer Plugin doesn't include this metric, causing panels referencing it to be empty.

**Solution**
Add the incredibly simple [UptimeMetrics](https://github.com/micrometer-metrics/micrometer/blob/main/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/system/UptimeMetrics.java) MeterBinder. It adds the `process.uptime` and `process.start.time` metrics.


e5l (approved):
Hey @SIMULATAN, thanks for the PR! LGTM!

PR #4590, opened 2025-01-09
KTOR-865 Support CIO server for wasm-js and js

whyoleg:
**Subsystem**
Server CIO

**Motivation**
Probably fixes https://youtrack.jetbrains.com/issue/KTOR-865/Add-node.js-server-engine
More specific issues description/name should be filled mentioning wasmJs and the issue should be made public
_If needed I can create a separate issue._

**Solution**
Most of the code in `ktor-server-cio` is just copied with minor changes, additionally test modules are adapted to support js/wasmJs server tests with CIO


osipxd (approved):
Thank you for reopening the PR! LGTM 👍

PR #4592, opened 2025-01-10
KTOR-7806 Add logging with standard Okhttp-like format

Stexxe:
https://youtrack.jetbrains.com/issue/KTOR-7806

### DIFF
--- a/ktor-client/ktor-client-plugins/ktor-client-logging/common/src/io/ktor/client/plugins/logging/Logging.kt
+++ b/ktor-client/ktor-client-plugins/ktor-client-logging/common/src/io/ktor/client/plugins/logging/Logging.kt
@@ -30,6 +30,15 @@ import kotlinx.io.Buffer
 private val ClientCallLogger = AttributeKey<HttpClientCallLogger>("CallLogger")
 private val DisableLogging = AttributeKey<Unit>("DisableLogging")
 
+public enum class LoggingFormat {
+    Default,
+    /**
+     * [OkHttp logging format](https://github.com/square/okhttp/blob/master/okhttp-logging-interceptor/src/main/kotlin/okhttp3/logging/HttpLoggingInterceptor.kt#L56).
+     * Writes only application-level logs because the low-level HTTP communication is hidden within the engine implementations.
+     */
+    OkHttp
+}
+
 /**
  * A configuration for the [Logging] plugin.
  */
@@ -40,11 +49,7 @@ public class LoggingConfig {
 
     private var _logger: Logger? = null
 
-    /**
-     * If true, turns on the [OkHttp logging format](https://github.com/square/okhttp/blob/master/okhttp-logging-interceptor/src/main/kotlin/okhttp3/logging/HttpLoggingInterceptor.kt#L56).
-     * Writes only application-level logs because the low-level HTTP communication is hidden within the engine implementations.
-     */
-    public var okhttpFormat: Boolean = false
+    public var format: LoggingFormat = LoggingFormat.Default
 
     /**
      * Specifies a [Logger] instance.
@@ -90,7 +95,7 @@ public val Logging: ClientPlugin<LoggingConfig> = createClientPlugin("Logging", 
     val level: LogLevel = pluginConfig.level
     val filters: List<(HttpRequestBuilder) -> Boolean> = pluginConfig.filters
     val sanitizedHeaders: List<SanitizedHeader> = pluginConfig.sanitizedHeaders
-    val okHttpFormat = pluginConfig.okhttpFormat
+    val okHttpFormat = pluginConfig.format == LoggingFormat.OkHttp
 
     fun shouldBeLogged(request: HttpRequestBuilder): Boolean = filters.isEmpty() || filters.any { it(request) }
 
--- a/ktor-client/ktor-client-plugins/ktor-client-logging/common/src/io/ktor/client/plugins/logging/Logging.kt
+++ b/ktor-client/ktor-client-plugins/ktor-client-logging/common/src/io/ktor/client/plugins/logging/Logging.kt
@@ -106,7 +106,12 @@ public val Logging: ClientPlugin<LoggingConfig> = createClientPlugin("Logging", 
      * Long?: body size if calculated.
      * ByteReadChannel: body channel with the original data.
      */
-    suspend fun detectIfBinary(body: ByteReadChannel, contentLength: Long?, contentType: ContentType?, headers: Headers): Triple<Boolean, Long?, ByteReadChannel> {
+    suspend fun detectIfBinary(
+        body: ByteReadChannel,
+        contentLength: Long?,
+        contentType: ContentType?,
+        headers: Headers
+    ): Triple<Boolean, Long?, ByteReadChannel> {
         if (headers.contains(HttpHeaders.ContentEncoding)) {
             return Triple(true, contentLength, body)
         }
@@ -156,7 +161,13 @@ public val Logging: ClientPlugin<LoggingConfig> = createClientPlugin("Logging", 
         return Triple(isBinary, contentLength, body)
     }
 
-    suspend fun logRequestBody(content: OutgoingContent, contentLength: Long?, headers: Headers, method: HttpMethod, body: ByteReadChannel) {
+    suspend fun logRequestBody(
+        content: OutgoingContent,
+        contentLength: Long?,
+        headers: Headers,
+        method: HttpMethod,
+        body: ByteReadChannel
+    ) {
         val (isBinary, size, newBody) = detectIfBinary(body, contentLength, content.contentType, headers)
 
         if (!isBinary) {
@@ -183,8 +194,13 @@ public val Logging: ClientPlugin<LoggingConfig> = createClientPlugin("Logging", 
         }
     }
 
-    suspend fun logOutgoingContent(content: OutgoingContent, method: HttpMethod, headers: Headers, process: (ByteReadChannel) -> ByteReadChannel = { it }): OutgoingContent? {
-        return when(content) {
+    suspend fun logOutgoingContent(
+        content: OutgoingContent,
+        method: HttpMethod,
+        headers: Headers,
+        process: (ByteReadChannel) -> ByteReadChannel = { it }
+    ): OutgoingContent? {
+        return when (content) {
             is OutgoingContent.ByteArrayContent -> {
                 val bytes = content.bytes()
                 logRequestBody(content, bytes.size.toLong(), headers, method, ByteReadChannel(bytes))
@@ -227,7 +243,11 @@ public val Logging: ClientPlugin<LoggingConfig> = createClientPlugin("Logging", 
         val uri = URLBuilder().takeFrom(request.url).build().pathQuery()
         val body = request.body
         val headers = HeadersBuilder().apply {
-            if (body is OutgoingContent && request.method != HttpMethod.Get && request.method != HttpMethod.Head && body !is EmptyContent) {
+            if (body is OutgoingContent &&
+                request.method != HttpMethod.Get &&
+                request.method != HttpMethod.Head &&
+                body !is EmptyContent
+            ) {
                 body.contentType?.let {
                     appendIfNameAbsent(HttpHeaders.ContentType, it.toString())
                 }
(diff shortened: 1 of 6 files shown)


ktor-client/ktor-client-plugins/ktor-client-logging/common/src/io/ktor/client/plugins/logging/Logging.kt
```diff
@@ -31,6 +40,12 @@ public class LoggingConfig {
 
     private var _logger: Logger? = null
 
+    /**
+     * If true, turns on the [OkHttp logging format](https://github.com/square/okhttp/blob/master/okhttp-logging-interceptor/src/main/kotlin/okhttp3/logging/HttpLoggingInterceptor.kt#L56).
+     * Writes only application-level logs because the low-level HTTP communication is hidden within the engine implementations.
+     */
+    public var standardFormat: Boolean = false
```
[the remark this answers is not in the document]
Stexxe: I don't want to complicate things by introducing a formatter interface. If we decide to add another format, I suggest considering the new interface only then.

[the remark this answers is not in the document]
e5l: Should we consider `okhttpLoggingFormat` name?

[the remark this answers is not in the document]
e5l: I would introduce an enum for that cause it would also be nice to support a `curl` format.

ktor-client/ktor-client-plugins/ktor-client-logging/jvm/test/io/ktor/client/plugins/logging/NewFormatTest.kt
```diff
@@ -0,0 +1,1289 @@
...
+import org.junit.jupiter.api.BeforeEach
+import java.net.UnknownHostException
+import kotlin.coroutines.CoroutineContext
+import kotlin.test.Test
+import kotlin.test.assertContentEquals
+import kotlin.test.assertEquals
+import kotlin.test.assertFailsWith
+import kotlin.test.assertTrue
+
+class NewFormatTest {
```
e5l, line 39: `OkHttpFormatTest`

ktor-client/ktor-client-plugins/ktor-client-logging/common/src/io/ktor/client/plugins/logging/Logging.kt
```diff
@@ -239,6 +613,34 @@ public val Logging: ClientPlugin<LoggingConfig> = createClientPlugin("Logging",
...
+
+        if (encodedQuery.isEmpty()) return@buildString
+
+        append("?")
+        append(encodedQuery)
+    }
+}
+
+@OptIn(DelicateCoroutinesApi::class)
+private fun calcRequestBodySize(content: Any, headers: Headers): Long {
```
e5l, line 632: computeRequestBodySize

ktor-client/ktor-client-plugins/ktor-client-logging/common/src/io/ktor/client/plugins/logging/Logging.kt
```diff
@@ -239,6 +613,34 @@ public val Logging: ClientPlugin<LoggingConfig> = createClientPlugin("Logging",
...
+    check(content is OutgoingContent)
+
+    return when (content) {
+        is OutgoingContent.ByteArrayContent -> content.bytes().size.toLong()
+        is OutgoingContent.ContentWrapper -> calcRequestBodySize(content.delegate(), content.headers)
+        is OutgoingContent.NoContent -> 0
+        is OutgoingContent.ProtocolUpgrade -> 0
+        else -> error("Unable to calculate the size for type ${content::class.simpleName}")
```
e5l, line 640: Could you tell me if the exception is supposed to be handled or printed?

I would return `null` and handle this case explicitly

ktor-client/ktor-client-plugins/ktor-client-logging/common/src/io/ktor/client/plugins/logging/Logging.kt
```diff
@@ -239,6 +613,34 @@ public val Logging: ClientPlugin<LoggingConfig> = createClientPlugin("Logging",
     ResponseObserver.install(ResponseObserver.prepare { onResponse(observer) }, client)
 }
 
+private fun Url.pathQuery(): String {
+    return buildString {
```
e5l, line 617: Please consider adding:
-  a non-empty browser test with `window.location`
- test with default request plugin and relative path

e5l (changes requested):

ktor-client/ktor-client-plugins/ktor-client-logging/common/src/io/ktor/client/plugins/logging/Logging.kt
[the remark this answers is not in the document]
osipxd: Even if we don't introduce a public interface for formatter, I would move formatting logic to a separate class or file to make the logic of the plugin itself cleaner and more readable.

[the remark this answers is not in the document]
Stexxe: > Should we consider okhttpLoggingFormat name?

Sure, I suggest `okhttpFormat` for brevity.

> I would introduce an enum for that cause it would also be nice to support a curl format.

I wouldn't do that because we don't have the implementation for that. Also, I don't quite understand why we need more formats as nobody requested them.

> Even if we don't introduce a public interface for formatter, I would move formatting logic to a separate class or file to make the logic of the plugin itself cleaner and more readable.

It's quite readable at the moment. Can you point to the specific issues?

ktor-client/ktor-client-plugins/ktor-client-logging/common/src/io/ktor/client/plugins/logging/Logging.kt
Stexxe: It's more like a guard. If the exception is thrown then it's a bug.

ktor-client/ktor-client-plugins/ktor-client-logging/common/src/io/ktor/client/plugins/logging/Logging.kt
Stexxe: > test with default request plugin and relative path

Wouldn't I test the `DefaultRequest` plugin then? If not, can you explain the subject under test?

e5l: Set `window.location.href = "https://ktor.io"`

So this call will give you a different result: https://github.com/ktorio/ktor/blob/c9739da4023dbf3127d5b91bdb621811fc1ee556/ktor-http/jsAndWasmShared/src/io/ktor/http/URLBuilderJs.kt#L9

ktor-client/ktor-client-plugins/ktor-client-logging/common/src/io/ktor/client/plugins/logging/Logging.kt
```diff
@@ -31,6 +40,12 @@ public class LoggingConfig {
 
     private var _logger: Logger? = null
 
+    /**
+     * If true, turns on the [OkHttp logging format](https://github.com/square/okhttp/blob/master/okhttp-logging-interceptor/src/main/kotlin/okhttp3/logging/HttpLoggingInterceptor.kt#L56).
+     * Writes only application-level logs because the low-level HTTP communication is hidden within the engine implementations.
+     */
+    public var okhttpFormat: Boolean = false
```
e5l, line 47: I'm still not sure if it's a good idea to have a library-specific flag. Let's have an enum.

ktor-client/ktor-client-plugins/ktor-client-logging/common/src/io/ktor/client/plugins/logging/Logging.kt
e5l: Please add the test to make sure the exception is not lost

ktor-client/ktor-client-plugins/ktor-client-logging/common/src/io/ktor/client/plugins/logging/Logging.kt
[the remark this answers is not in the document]
e5l: @Stexxe, if we introduce a new or custom log format, we will have to drop this field: only one format can be enabled. So the possibility of breaking change in the future is high.

e5l (changes requested):

ktor-client/ktor-client-plugins/ktor-client-logging/common/src/io/ktor/client/plugins/logging/Logging.kt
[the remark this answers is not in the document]
Stexxe: I've added the enum.

ktor-client/ktor-client-plugins/ktor-client-logging/common/src/io/ktor/client/plugins/logging/Logging.kt
Stexxe: > Set `window.location.href = "https://ktor.io"`
> 
> So this call will give you a different result:
> 
> https://github.com/ktorio/ktor/blob/c9739da4023dbf3127d5b91bdb621811fc1ee556/ktor-http/jsAndWasmShared/src/io/ktor/http/URLBuilderJs.kt#L9

This method is used only with the `URLBuilder.Companion.origin` property, which I don't use at all.

ktor-client/ktor-client-plugins/ktor-client-logging/common/src/io/ktor/client/plugins/logging/Logging.kt
```diff
@@ -7,20 +7,38 @@ package io.ktor.client.plugins.logging
...
+import kotlinx.coroutines.launch
+import kotlinx.io.Buffer
 
 private val ClientCallLogger = AttributeKey<HttpClientCallLogger>("CallLogger")
 private val DisableLogging = AttributeKey<Unit>("DisableLogging")
 
+public enum class LoggingFormat {
+    Default,
+    /**
+     * [OkHttp logging format](https://github.com/square/okhttp/blob/master/okhttp-logging-interceptor/src/main/kotlin/okhttp3/logging/HttpLoggingInterceptor.kt#L56).
```
osipxd, line 36: Let's change the link to a permanent one
```suggestion
     * [OkHttp logging format](https://github.com/square/okhttp/blob/parent-4.12.0/okhttp-logging-interceptor/src/main/kotlin/okhttp3/logging/HttpLoggingInterceptor.kt#L48-L105).
```

ktor-client/ktor-client-plugins/ktor-client-logging/common/src/io/ktor/client/plugins/logging/Logging.kt
```diff
@@ -31,6 +49,8 @@ public class LoggingConfig {
 
     private var _logger: Logger? = null
 
+    public var format: LoggingFormat = LoggingFormat.Default
```
osipxd, line 52: Nit. A brief comment would be great

ktor-client/ktor-client-plugins/ktor-client-logging/common/src/io/ktor/client/plugins/logging/Logging.kt
```diff
@@ -69,15 +89,342 @@ public class LoggingConfig {
...
+
+            if (size != null) {
+                logger.log("<-- END HTTP (${duration}ms, $type $size-byte body omitted)")
+            } else {
+                logger.log("<-- END HTTP (${duration}ms, $type body omitted)")
+            }
+        }
+    }
+
+    suspend fun logResponseStdFormat(response: HttpResponse): HttpResponse {
```
osipxd, line 362: Probably, should be renamed
```suggestion
    suspend fun logResponseOkHttpFormat(response: HttpResponse): HttpResponse {
```

ktor-client/ktor-client-plugins/ktor-client-logging/common/src/io/ktor/client/plugins/logging/Logging.kt
```diff
@@ -69,15 +89,342 @@ public class LoggingConfig {
...
+                }
+
+                val (origChannel, newChannel) = channel.split(client)
+                logRequestBody(content, content.contentLength, headers, method, newChannel)
+                LoggedContent(content, origChannel)
+            }
+        }
+    }
+
+    suspend fun logRequestStdFormat(request: HttpRequestBuilder): OutgoingContent? {
```
osipxd, line 245: And this
```suggestion
    suspend fun logRequestOkHttpFormat(request: HttpRequestBuilder): OutgoingContent? {
```

osipxd (approved):

ktor-client/ktor-client-plugins/ktor-client-logging/common/src/io/ktor/client/plugins/logging/Logging.kt
```diff
@@ -69,15 +91,342 @@ public class LoggingConfig {
...
+    fun isBody(): Boolean = level == LogLevel.BODY || level == LogLevel.ALL
+
+    /**
+     * Detects if the body is a binary data
+     * @return
+     * Boolean: true if the body is a binary data.
+     * Long?: body size if calculated.
+     * ByteReadChannel: body channel with the original data.
+     */
+    suspend fun detectIfBinary(
```
e5l, line 116: can be top-level internal and covered by unit test

ktor-client/ktor-client-plugins/ktor-client-logging/common/src/io/ktor/client/plugins/logging/Logging.kt
```diff
@@ -165,6 +514,23 @@ public val Logging: ClientPlugin<LoggingConfig> = createClientPlugin("Logging",
             return@on
         }
 
+        if (okHttpFormat) {
```
e5l, line 517: it would be nice to have extensions for okHttp format grouped in a separate file

e5l (approved):
lgtm

PR #4597, opened 2025-01-13
KTOR-8045 Fix for NPE race condition in dispose

bjhham:
**Subsystem**
Server, DoubleReceive

**Motivation**
[KTOR-8045](https://youtrack.jetbrains.com/issue/KTOR-8045) DoubleReceive: NullPointerException caused by race condition

**Solution**
Moving the body's discard call outside the coroutine avoids the race condition when dispose is called twice.

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-double-receive/common/src/io/ktor/server/plugins/doublereceive/ByteArrayCache.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-double-receive/common/src/io/ktor/server/plugins/doublereceive/ByteArrayCache.kt
@@ -53,7 +53,7 @@ internal class MemoryCache(
     override fun dispose() {
         GlobalScope.launch {
             reader.discard()
-            fullBody?.discard()
         }
+        fullBody?.discard()
     }
 }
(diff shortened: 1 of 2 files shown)


ktor-server/ktor-server-plugins/ktor-server-double-receive/common/src/io/ktor/server/plugins/doublereceive/ByteArrayCache.kt
```diff
@@ -53,7 +53,7 @@ internal class MemoryCache(
     override fun dispose() {
         GlobalScope.launch {
             reader.discard()
```
osipxd, line 55: This `.discard()` call is not affected because of `isClosedForRead` check under the hood, right?

osipxd (approved):
LGTM! 🏆

ktor-server/ktor-server-plugins/ktor-server-double-receive/common/src/io/ktor/server/plugins/doublereceive/ByteArrayCache.kt
bjhham: Yep, I don't know if it was intended to be thread safe, but it seems to be 😄

PR #4598, opened 2025-01-13
KTOR-8057 Introduce base class for client engine tests

osipxd:
**Subsystem**
Client, Test Infrastructure

**Motivation**
[KTOR-8057](https://youtrack.jetbrains.com/issue/KTOR-8057) Introduce base class for client tests
[KTOR-8058](https://youtrack.jetbrains.com/issue/KTOR-8058) Don't publish internal test artifacts

**Solution**
- Move test infrastructure classes from `ktor-client-tests` to a new module `ktor-client-test-base`
- Add new class `ClientEngineTest` simplifying testing of client engines, enabling timeouts and retries.
- Migrate existing client engine tests to the new test infrastructure

It's simpler to review commit by commit


build.gradle.kts
```diff
@@ -26,9 +26,12 @@ val configuredVersion: String by extra
 
 apply(from = "gradle/verifier.gradle")
 
-extra["skipPublish"] = mutableListOf(
+val internalProjects = listOf(
+    "ktor-client-tests",
+    "ktor-server-test-base",
```
osipxd, line 32: @bjhham, is it okay that I exclude `ktor-server-test-base` from publishing?

ktor-client/ktor-client-test-base/common/src/io/ktor/client/test/base/ClientLoader.kt
```diff
@@ -88,7 +88,7 @@ abstract class ClientLoader(private val timeout: Duration = 1.minutes) {
 
     /** Excludes the specified [engines] from test execution. */
     fun except(engines: List<String>): EngineSelectionRule {
-        val skipPatterns = engines.map(EnginePattern::parse)
+        val skipPatterns = engines.map(EnginePattern.Companion::parse)
```
osipxd, line 91:
```suggestion
        val skipPatterns = engines.map(EnginePattern::parse)
```

ktor-client/ktor-client-test-base/common/src/io/ktor/client/test/base/ClientEngineTest.kt
```diff
@@ -0,0 +1,19 @@
...
+package io.ktor.client.test.base
+
+import io.ktor.client.engine.*
+import io.ktor.test.*
+import kotlin.time.Duration
+import kotlin.time.Duration.Companion.minutes
+
+abstract class ClientEngineTest<T : HttpClientEngineConfig>(private val factory: HttpClientEngineFactory<T>) {
+
+    fun testClient(
```
osipxd, line 14: Any ideas regarding naming are welcome

build.gradle.kts
bjhham: Yeah, it kinda looks like `ktor-test-base` was a typo here, or maybe it was renamed... Pretty sure `ktor-test-base` doesn't exist right?

osipxd: `ktor-test-base` is a former `ktor-junit`. I've renamed it to use as a multiplatform module sharing common test utils between client and server tests. My question was about `ktor-server-test-base` because it wasn't included in this list and it could be done for some purpose.

bjhham: Ah so that's it!  It's all coming back to me now - so `ktor-server-test-base` was being used previously in lieu of `ktor-server-test-host`, so we needed to publish it with the relocation link.  So now we can remove it from the publications and all should be good.

ktor-client/ktor-client-test-base/common/src/io/ktor/client/test/base/ClientEngineTest.kt
bjhham: I would've chosen the same 😄

bjhham (approved):
All looks good to me, nice job 👍

PR #4602, opened 2025-01-14
KTOR-7213 Allow access to the configuration options of a HttpRequestRetry client plugin

e5l:
Fix [KTOR-7213](https://youtrack.jetbrains.com/issue/KTOR-7213) Access to the configuration options of a HttpClient plugin to tweak or wrap them with additional logic


marychatte (approved):
lgtm, only changes in `ktor-io.api` are probably not needed

PR #4609, opened 2025-01-15
KTOR-7194 Deferred session fetching for public endpoints

bjhham:
**Subsystem**
Server, Sessions

**Motivation**
[KTOR-7194](https://youtrack.jetbrains.com/issue/KTOR-7194) SessionStorage.read() is called for non-authenticated routes and static assets

**Solution**
This change provides an option in the sessions configuration `deferred = true` that defers session retrieval until it is used in the request.

This works by introducing a `CurrentSession` implementation that uses lazy retrieval to avoid unnecessary calls to the backing storage.

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-sessions/common/src/io/ktor/server/sessions/SessionDeferral.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-sessions/common/src/io/ktor/server/sessions/SessionDeferral.kt
@@ -1,0 +1,12 @@
+/*
+ * Copyright 2014-2025 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.server.sessions
+
+import io.ktor.server.application.ApplicationCall
+
+/**
+ * Creates a lazy loading session from the given providers.
+ */
+internal expect fun createDeferredSession(call: ApplicationCall, providers: List<SessionProvider<*>>): StatefulSession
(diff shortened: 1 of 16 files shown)


ktor-server/ktor-server-plugins/ktor-server-sessions/common/src/io/ktor/server/sessions/SessionsConfig.kt
```diff
@@ -18,6 +18,13 @@ public class SessionsConfig {
      */
     public val providers: List<SessionProvider<*>> get() = registered.toList()
 
+    /**
+     * When set to true, sessions will be lazily retrieved from storage.
+     *
+     * Note: this is only available for JVM in Ktor 3.0
+     */
+    public var deferred: Boolean = false
```
e5l, line 26: I would mark it with OptIn and look for a better explaining name

e5l: Maybe we should have a system property (like `io.ktor.server.sessions.lazycreate`) for this instead

e5l (changes requested):

ktor-server/ktor-server-plugins/ktor-server-sessions/common/src/io/ktor/server/sessions/SessionsConfig.kt
bjhham: Good idea, this way we don't need to change the API 👍

ktor-server/ktor-server-plugins/ktor-server-sessions/common/src/io/ktor/server/sessions/SessionDeferral.kt
```diff
@@ -0,0 +1,16 @@
+/*
+ * Copyright 2014-2025 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.server.sessions
+
+import io.ktor.server.application.ApplicationCall
+
+internal const val SESSIONS_DEFERRED_FLAG = "io.ktor.server.sessions.deferred"
```
e5l, line 9: Please consider documenting this flag

e5l (approved):

PR #4610, opened 2025-01-16
KTOR-7139 Fix exception thrown in onCallRespond makes the client wait…

e5l:
Fix [KTOR-7139](https://youtrack.jetbrains.com/issue/KTOR-7139) Exception thrown in onCallRespond makes the client wait for response indefinitely


bjhham (approved):
LGMT 👍

PR #4613, opened 2025-01-17
KTOR-8008 Fix uncaught exceptions in websocket pinger

bjhham:
**Subsystem**
Core, I/O

**Motivation**
[KTOR-8008](https://youtrack.jetbrains.com/issue/KTOR-8008) Uncaught cannot write to a channel errors from ws-pinger

**Solution**
- Fixed the immediate problem by simply ignoring closed channel exceptions, since this is normal for the ping operation when it tries to write to a closed connection before cancellation.
- Introduced sub-types for closed byte channel exceptions over the current IOException typealias approach to allow for more granular exception handling
- Replaced the custom getter for `CloseToken.cause` with `wrapCause()` to be more explicit with the behavior.


osipxd (approved):
👍 

Btw, is it that "impactful change"?

bjhham:
@osipxd I would say it's not an impactful change because the new exception types are still under the IOException hierarchy and the pinger catch clause should just bring the behaviour back to what was expected in earlier releases

PR #4615, opened 2025-01-17
KTOR-7743 Remove redundant configurations for JVM target

osipxd:
**Subsystem**
Infrastructure

**Motivation**
Cleanup before moving JVM target configs to the `build-logic` in [KTOR-7743](https://youtrack.jetbrains.com/issue/KTOR-7743).

**Solution**
- Remove jetty-apln-boot as it is not necessary anymore
- Move test logback config to `ktor-server-test-base` and remove `testOutput` configuration

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -57,10 +57,6 @@ subprojects {
     configureTargets()
     if (CI) configureTestTasksOnCi()
 
-    configurations {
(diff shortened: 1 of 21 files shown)


osipxd:
> Gotta love seeing the red 🔥

But not in tests :D

PR #4616, opened 2025-01-17
KTOR-8083 Refactor automatic targets configuration

osipxd:
**Subsystem**
Infrastructure

**Motivation**
[KTOR-8083](https://youtrack.jetbrains.com/issue/KTOR-8083) Refactor automatic targets configuration

**Solution**
I've introduced `KtorTargets` to encapsulate targets configuring logic.
Instead of manually created boolean flags for each target type, we rely on the actual hierarchy. Unfortunately, it is not possible to access this hierarchy in KGP, so I had to introduce `KotlinHierarchyTracker` remembering the hierarchy configured in the `KotlinHierarchyTemplate`.

It's better to review commit by commit.


bjhham (approved):
🫶

PR #4618, opened 2025-01-20
KTOR-8091 Fix ktor-client-curl publishing

osipxd:
**Subsystem**
Infrastructure

**Motivation**
[KTOR-8091](https://youtrack.jetbrains.com/issue/KTOR-8091) ktor-client-curl artifacts aren’t published after EAP 1146

**Solution**
Aggregating tasks relay on the flags declared in `TargetsConfig`. Unfortunately, these flags were out of sync a bit (#4616 should fix this problem) and this caused `ktor-client-curl` not to be published.
```kotlin
private fun Project.configureAggregatingTasks() {
    // ...
    // These three were false if only "desktop" source-set is present.
    if (hasLinux) registerAggregatingTask("Linux", linuxTargets)
    if (hasWindows) registerAggregatingTask("Windows", windowsTargets)
    if (hasDarwin) registerAggregatingTask("Darwin", darwinTargets)
    // ...
}
```


bjhham (approved):
I think we need a Venn diagram for all these targets 😂

osipxd:
> I think we need a Venn diagram for all these targets 😂

After https://github.com/ktorio/ktor/pull/4616 it won't be needed anymore. I've implemented programmatic tracking of relations between targets and their groups.

PR #4619, opened 2025-01-20
KTOR-8051 Fix EOFException on retry body check

bjhham:
**Subsystem**
Client, HttpRequestRetry

**Motivation**
[KTOR-8051](https://youtrack.jetbrains.com/issue/KTOR-8051) HttpRequestRetry: race condition for isClosedForRead leads to EOFException: Channel is already closed

**Solution**
This was actually caused by some unnecessary checks in byte read operations.  When the read buffer is closed, then it should fail through the normal path, throwing a `ClosedReadChannelException`.

### DIFF
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/HttpRequestRetry.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/HttpRequestRetry.kt
@@ -235,7 +235,6 @@ public class HttpRequestRetryConfig {
  * }
  * ```
  */
-@OptIn(InternalAPI::class)
 @Suppress("NAME_SHADOWING")
 public val HttpRequestRetry: ClientPlugin<HttpRequestRetryConfig> = createClientPlugin(
     "RetryFeature",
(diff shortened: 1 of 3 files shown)


ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/HttpRequestRetry.kt
```diff
@@ -317,9 +317,7 @@ public val HttpRequestRetry: ClientPlugin<HttpRequestRetryConfig> = createClient
                 call = proceed(subRequest)
                 if (!shouldRetry(retryCount, maxRetries, shouldRetry, call)) {
                     // throws exception if body is corrupt
-                    if (call.response.isSaved && !call.response.rawContent.isClosedForRead) {
-                        call.response.readBytes(0)
-                    }
+                    call.response.throwOnInvalidResponseBody()
```
bjhham, line 320: I just cleaned this up a little, added some comments, and removed the unnecessary check.

bjhham:
I'll investigate why this is flaky and fix it here https://ktor.teamcity.com/buildConfiguration/Ktor_KtorMatrixCoreLinuxJava8/313230

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
```diff
@@ -460,12 +460,9 @@ public val ByteReadChannel.availableForRead: Int
  */
 @OptIn(InternalAPI::class)
 public suspend fun ByteReadChannel.readFully(out: ByteArray, start: Int = 0, end: Int = out.size) {
-    if (isClosedForRead) throw EOFException("Channel is already closed")
-
     var offset = start
     while (offset < end) {
         if (readBuffer.exhausted()) awaitContent()
-        if (isClosedForRead) throw EOFException("Channel is already closed")
```
osipxd, line 468: Was this check redundant after the previous line?

osipxd (approved):
👍

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
bjhham: It'll hit a `ClosedReadChannelException` on the next line if the channel is closed.

PR #4625, opened 2025-01-26
Fix library and include paths for ktor-client-curl interop

KitsuneAlex:
**Subsystem**
ktor-client-curl

**Motivation**
While trying to cross-compile an application for X64 on my arm64 Mac, i discovered that the linker wasn't looking for any binaries in the project-provided prebuilt binaries, nor was it using the right headers.
Instead it was looking for the libs and headers on the regular system paths, which can lead to irreproducible builds and ABI issues.

**Solution**
This PR adds explicit, project-relative paths to the interop configuration for libcurl.

### DIFF
--- a/ktor-client/ktor-client-curl/desktop/interop/libcurl.def
+++ b/ktor-client/ktor-client-curl/desktop/interop/libcurl.def
@@ -3,13 +3,20 @@ package = libcurl
 headers = curl/curl.h
 headerFilter = curl/*
 
+compilerOpts = -Idesktop/interop/include
+
 # there is no libz by default installed on windows, so we need to include it statically
 staticLibraries.mingw = libcurl.a libssl.a libcrypto.a libz.a
 compilerOpts.mingw = -DCURL_STATICLIB
+linkerOpts.mingw_x64 = -Ldesktop/interop/lib/mingwX64
 
 # there is no need to use openssl on macos, as curl is built using out-of-the-box security framework
 staticLibraries.osx = libcurl.a
 linkerOpts.osx = -framework Security -framework SystemConfiguration
+linkerOpts.osx_x64 = -Ldesktop/interop/lib/macosX64
+linkerOpts.osx_arm64 = -Ldesktop/interop/lib/macosArm64
 
 staticLibraries.linux = libcurl.a libssl.a libcrypto.a
 linkerOpts.linux = -lz
+linkerOpts.linux_x64 = -Ldesktop/interop/lib/linuxX64
+linkerOpts.linux_arm64 = -Ldesktop/interop/lib/linuxArm64


KitsuneAlex:
What's the build error here if i may ask? ^^
I built it on my Linux main machine, on my Mac  and a Windows  VM and that worked fine. Cross-compilation also worked fine with this change.

KitsuneAlex:
The CI build error seems to be a CIO related timeout, maybe a flakey test?

ktor-client/ktor-client-curl/desktop/interop/libcurl.def
```diff
@@ -3,13 +3,20 @@ package = libcurl
 headers = curl/curl.h
 headerFilter = curl/*
 
+compilerOpts = -Idesktop/interop/include
```
whyoleg, line 6: those are added in build script file: https://github.com/ktorio/ktor/blob/697582c4bf2850c75fd1b87d8a5723e215217afa/ktor-client/ktor-client-curl/build.gradle.kts#L14-L15

and yes, they could be moved inside `.def` file - there is no difference in behavior here

ktor-client/ktor-client-curl/desktop/interop/libcurl.def
```diff
@@ -3,13 +3,20 @@ package = libcurl
...
+
 # there is no libz by default installed on windows, so we need to include it statically
 staticLibraries.mingw = libcurl.a libssl.a libcrypto.a libz.a
 compilerOpts.mingw = -DCURL_STATICLIB
+linkerOpts.mingw_x64 = -Ldesktop/interop/lib/mingwX64
```
whyoleg, line 11: this will not work, as we need to provide another flag to search for static libraries to embed: [libraryPaths](https://kotlinlang.org/docs/native-definition-file.html#include-a-static-library)

whyoleg (changes requested):
Hey!
Have you received the error on latest [EAP](https://ktor.io/eap/) version (3.1.0-eap-1216) or on latest stable (3.0.3)?
curl client cinterop setup was recently reworked to use static libraries on all platforms: https://github.com/ktorio/ktor/pull/4445
For me, it looks like those changes are not needed now

ktor-client/ktor-client-curl/desktop/interop/libcurl.def
KitsuneAlex: Mistake on my end here, will fix.

ktor-client/ktor-client-curl/desktop/interop/libcurl.def
KitsuneAlex: Then only the linker paths are required. Will fix.

KitsuneAlex:
> Hey! Have you received the error on latest [EAP](https://ktor.io/eap/) version (3.1.0-eap-1216) or on latest stable (3.0.3)? curl client cinterop setup was recently reworked to use static libraries on all platforms: #4445 For me, it looks like those changes are not needed now

Latest stable 3.0.3, Kotlin 2.1.0, Gradle 8.12 on macOS 15.2.

whyoleg:
Could you recheck the latest EAP on your side?

KitsuneAlex:
> Could you recheck the latest EAP on your side?

Will do, thank you very much :)

KitsuneAlex:
The EAP seems to be missing the linuxArm64 target preventing my project to build in my CI. I'll disable the target for now  to test it.

osipxd:
> The EAP seems to be missing the linuxArm64 target preventing my project to build in my CI. I'll disable the target for now to test it.

As I can see `ktor-client-curl` for Arm64 is published ([3.1.0-eap-1216](https://maven.pkg.jetbrains.space/public/p/ktor/eap/io/ktor/ktor-client-curl-linuxarm64/3.1.0-eap-1216/)). Could you provide more information about the problem you face when trying to build your project with EAP?

osipxd:
> I was using 1212

Ah, yes. Publishing of `ktor-client-curl` was broken in that version

PR #4629, opened 2025-01-27
KTOR-7734 Do not abort fetch after its completion

osipxd:
**Subsystem**
Client (JS, WasmJS)

**Motivation**
[KTOR-7734](https://youtrack.jetbrains.com/issue/KTOR-7734) "AbortError: BodyStreamBuffer was aborted" error when canceling parent job

**Solution**
The problem was caused by calling `controller.abort()` after request fetching was already completed (together with its coroutine). It resulted in propagation of this exception to the global level without any possibility to catch it.

Unfortunately, I didn't manage to write a test for this case.


e5l (approved):
Nice, LGTM!

bjhham (approved):
Nice find

PR #4636, opened 2025-01-30
KTOR-8133 Fix concurrent flush and close

bjhham:
**Subsystem**
Core, I/O

**Motivation**
[KTOR-8133](https://youtrack.jetbrains.com/issue/KTOR-8133) Audit Ktor codebase for potential byte channel race conditions

**Solution**
This extra close can lead to two concurrent writes.  Normally, the consumer closes the channel, which finishes this job.

Note: This is the last sketchy looking spot I've found in the core code.  I'll keep combing through the rest of the project, but with a lower priority.


e5l (commented):
Could you tell me what will happen if the `channel` is not closed outside of the reader?

@osipxd, @marychatte let's check together

osipxd (approved):

PR #4637, opened 2025-01-31
KTOR-8135 Socket accept not throwing error on socket close on native

Thomas-Vos:
[KTOR-8135](https://youtrack.jetbrains.com/issue/KTOR-8135) Socket accept not throwing error on socket close on native

Test failed before on native macosArm64 but was successful on JVM, so native specific bug

### DIFF
--- a/ktor-network/common/test/io/ktor/network/sockets/tests/TCPSocketTest.kt
+++ b/ktor-network/common/test/io/ktor/network/sockets/tests/TCPSocketTest.kt
@@ -168,4 +168,20 @@ class TCPSocketTest {
         serverConnection.close()
         server.close()
     }
+
+    @Test
+    fun testAcceptErrorOnSocketClose() = testSockets { selector ->
+        val socket = aSocket(selector)
+            .tcp()
+            .bind(InetSocketAddress("127.0.0.1", 0))
+
+        launch {
+            assertFailsWith<IOException> {
+                socket.accept()
+            }
+        }
+        delay(100) // Make sure socket is awaiting connection using ACCEPT
+
+        socket.close()
+    }
 }
--- a/ktor-network/jsAndWasmShared/src/io/ktor/network/sockets/ServerSocketContext.kt
+++ b/ktor-network/jsAndWasmShared/src/io/ktor/network/sockets/ServerSocketContext.kt
@@ -55,7 +55,7 @@ internal class ServerSocketContext(
 private class ServerSocketImpl(
     override val localAddress: SocketAddress,
     override val socketContext: Job,
-    private val incoming: ReceiveChannel<Socket>,
+    private val incoming: Channel<Socket>,
     private val server: Server
 ) : ServerSocket {
     override suspend fun accept(): Socket = incoming.receive()
@@ -68,6 +68,7 @@ private class ServerSocketImpl(
     }
 
     override fun close() {
+        incoming.close(IOException("Server socket closed"))
         socketContext.cancel("Server socket closed")
     }
 }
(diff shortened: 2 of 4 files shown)


ktor-network/jsAndWasmShared/src/io/ktor/network/sockets/ServerSocketContext.kt
```diff
@@ -68,6 +68,7 @@ private class ServerSocketImpl(
     }
 
     override fun close() {
+        incoming.close(IOException("Server socket closed"))
```
whyoleg, line 71: I think it's better to use `cancel` instead of `close` here
this will also allow to revert `incoming` to use `ReceiveChannel` to respect the contract

bjhham: I think the close makes sense here to keep the exception in the same family as JVM (currently throws a `java.nio.channels.ClosedChannelException`) since cancel would throw a cancellation exception instead.  Maybe it ought to go on line 66 in lieu of the cancel call, or maybe we should just use cancellation exceptions everywhere instead of IOException, though it seems leaving it with IOException would be the least disruptive change for now.

ktor-network/common/test/io/ktor/network/sockets/tests/TCPSocketTest.kt
```diff
@@ -165,4 +165,20 @@ class TCPSocketTest {
...
+            .bind(InetSocketAddress("127.0.0.1", 0))
+
+        launch {
+            assertFailsWith<IOException> {
+                socket.accept()
+            }
+        }
+        delay(100) // Make sure socket is awaiting connection using ACCEPT
+
+        socket.close()
```
bjhham, line 182:
```suggestion
        val acceptJob = launch {
            assertFailsWith<NullPointerException> {
                socket.accept()
            }
        }
        yield()

        socket.close()
        acceptJob.join()
```

bjhham (changes requested):

ktor-network/common/test/io/ktor/network/sockets/tests/TCPSocketTest.kt
Thomas-Vos: I cannot reproduce the original issue with your test consistently (when replacing `NullPointerException` with `IOException`. Your test fails with a different IOException as the actual ACCEPT call is done after the socket was already closed (sometimes). A real delay is needed as the yield does not seem to work properly in this case, that is because the selection code is not react instantly. The wake up signal (sometimes) has a small delay.

ktor-network/jsAndWasmShared/src/io/ktor/network/sockets/ServerSocketContext.kt
Thomas-Vos: However, there are more inconsistencies. For example `attachForReading`/`attachForWriting` on sockets. On some platforms they are cancelled (with `CancellationException`), but on some platforms nothing happens. On iOS this even causes TCP sockets to hang when closed, which I would say is a bug, https://youtrack.jetbrains.com/issue/KTOR-8144. Trying to improve this in my other PR but having some issues with tests failing.

ktor-network/common/test/io/ktor/network/sockets/tests/TCPSocketTest.kt
bjhham: Ah, ok!  👍

bjhham (approved):

Thomas-Vos:
Can we get this merged so I can rebase my other PR?

PR #4639, opened 2025-01-31
KTOR-8136 Introduce ServerSocket.port to simplify port access for the bound server

e5l:
[KTOR-8136](https://youtrack.jetbrains.com/issue/KTOR-8136) Introduce ServerSocket.port to simplify port access for the bound server

### DIFF
--- a/ktor-network/common/src/io/ktor/network/sockets/SocketAddress.kt
+++ b/ktor-network/common/src/io/ktor/network/sockets/SocketAddress.kt
@@ -4,7 +4,29 @@
 
 package io.ktor.network.sockets
 
+/**
+ * Represents a socket address abstraction.
+ *
+ * This sealed class serves as the base type for different kinds of socket addresses,
+ * such as Internet-specific or other platform-dependent address types.
+ * Implementations of this class are expected to be platform-specific and provide
+ * details necessary to work with socket connections or bindings.
+ */
 public expect sealed class SocketAddress
+
+/**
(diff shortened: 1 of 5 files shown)


ktor-network/common/src/io/ktor/network/sockets/SocketAddress.kt
```diff
@@ -4,8 +4,30 @@
...
+/**
+ * Retrieves the port number associated with this socket address.
+ *
+ * If the `SocketAddress` instance is of type `InetSocketAddress`, the associated port is returned.
+ * Otherwise, an `UnsupportedOperationException` is thrown, as the provided address type does not support ports.
+ *
+ * @return the port number of the socket address if available.
+ * @throws UnsupportedOperationException if the socket address type does not support a port.
+ */
+public fun SocketAddress.port(): Int = when (this) {
```
osipxd, line 26: Maybe name this function `portOrNull()` or `portOrThrow()` to make it clear for a user that an unhappy path is possible here?

osipxd (approved):

ktor-network/common/src/io/ktor/network/sockets/SocketAddress.kt
e5l: I would keep it simple here:
- having a nullable version will lead to constant null checks everywhere
- it's method, not property. So the exception is assumed

PR #4641, opened 2025-01-31
KTOR-7734 JS: Catch error on channel reader closure

osipxd:
**Subsystem**
Client (JS/WasmJS)

**Motivation**
[KTOR-7734](https://youtrack.jetbrains.com/issue/KTOR-7734) "AbortError: BodyStreamBuffer was aborted" error when canceling parent job

**Solution**
Solution in #4629 wasn’t correct.
The actual root cause of the issue was an uncaught exception thrown inside a `Promise` returned by `reader.cancel()`. Added catching for this exception and reformatted code a bit.

It's better to review these two commits separately.

### DIFF
--- a/ktor-client/ktor-client-core/js/src/io/ktor/client/engine/js/browser/BrowserFetch.kt
+++ b/ktor-client/ktor-client-core/js/src/io/ktor/client/engine/js/browser/BrowserFetch.kt
@@ -8,11 +8,12 @@ import io.ktor.client.engine.js.*
 import io.ktor.client.fetch.*
 import io.ktor.utils.io.*
 import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.suspendCancellableCoroutine
+import kotlinx.coroutines.await
 import org.khronos.webgl.Uint8Array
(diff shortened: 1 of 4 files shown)


ktor-client/ktor-client-core/js/src/io/ktor/client/engine/js/browser/BrowserFetch.kt
```diff
@@ -31,13 +32,13 @@ internal fun CoroutineScope.channelFromStream(
...
     } catch (cause: Throwable) {
-        reader.cancel(cause)
+        reader.cancel(cause).catch { /* ignore */ }.await()
         throw cause
     }
 }.channel
 
 internal suspend fun ReadableStreamDefaultReader<Uint8Array>.readChunk(): Uint8Array? =
-    suspendCancellableCoroutine { continuation ->
+    suspendCoroutine { continuation ->
```
osipxd, line 41: Made this coroutine non-cancellable to ensure the exception thrown on `controller.abort()` is propagated to the upper level.

bjhham (approved):
Huzzah!

PR #4632, opened 2025-01-28
KTOR-8105 Fix concurrent flush attempts corrupting segment pool

bjhham:
**Subsystem**
Client, CIO

**Motivation**
Should solve all of our concurrency woes:
- [KTOR-8105](https://youtrack.jetbrains.com/issue/KTOR-8105) Race condition when writing to a buffer leads to NPE inside CIOReaderKt.readFrom
- [KTOR-8086](https://youtrack.jetbrains.com/issue/KTOR-8086) NPE in readBuffer
- [KTOR-8096](https://youtrack.jetbrains.com/issue/KTOR-8096) ArrayIndexOutOfBounds kotlinx-io

**Solution**
I found that the NIOSocketImpl is flushing the same channel as the TLS encoder coroutine, which can potentially lead to corruption due to the encoder writing the "close" message, then performing its own flushAndClose.

The stack traces in question:

```
[1318671573] write @ call-context#1025
  io.ktor.utils.io.ByteChannel.flush(ByteChannel.kt:91)
  io.ktor.utils.io.ByteChannel.flushAndClose(ByteChannel.kt:127)
  io.ktor.utils.io.CloseHookByteWriteChannel.flushAndClose(CloseHookByteWriteChannel.kt:21)
  io.ktor.utils.io.ByteWriteChannelKt$close$1.invoke(ByteWriteChannel.kt:39)
  io.ktor.utils.io.ByteWriteChannelKt$close$1.invoke(ByteWriteChannel.kt:39)
  kotlin.coroutines.intrinsics.IntrinsicsKt__IntrinsicsJvmKt$createCoroutineUnintercepted$$inlined$createCoroutineFromSuspendFunction$IntrinsicsKt__IntrinsicsJvmKt$1.invokeSuspend(IntrinsicsJvm.kt:223)
  kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
  kotlinx.coroutines.internal.DispatchedContinuationKt.resumeCancellableWith(DispatchedContinuation.kt:279)
  kotlinx.coroutines.intrinsics.CancellableKt.startCoroutineCancellable(Cancellable.kt:16)
  io.ktor.utils.io.ByteWriteChannelOperationsKt.fireAndForget(ByteWriteChannelOperations.kt:215)
  io.ktor.utils.io.ByteWriteChannelKt.close(ByteWriteChannel.kt:39)
[...]

PR #4643, opened 2025-02-02
KTOR-8144 Fix socket close with open read/write channels on Native

Thomas-Vos:
[KTOR-8144](https://youtrack.jetbrains.com/issue/KTOR-8144) Unable to close socket with open read/write channels on Native

### DIFF
--- a/ktor-network/common/test/io/ktor/network/sockets/tests/TCPSocketTest.kt
+++ b/ktor-network/common/test/io/ktor/network/sockets/tests/TCPSocketTest.kt
@@ -147,9 +147,14 @@ class TCPSocketTest {
         }
 
         val clientConnection = tcp.connect("127.0.0.1", 8003)
+
+        val clientInput = clientConnection.openReadChannel()
+        val clientOutput = clientConnection.openWriteChannel()
+
         val serverConnection = serverConnectionPromise.await()
 
         val serverInput = serverConnection.openReadChannel()
+        val serverOutput = serverConnection.openWriteChannel()
 
         // Need to make sure reading from server is done first, which will suspend because there is nothing to read.
         // Then close the connection from client side, which should cancel the reading because the socket disconnected.
@@ -161,8 +166,18 @@ class TCPSocketTest {
         assertFailsWith<EOFException> {
             serverInput.readByte()
         }
+        assertTrue(serverInput.isClosedForRead)
+
+        assertTrue(clientConnection.isClosed)
+        assertTrue(clientInput.isClosedForRead)
+        assertTrue(clientOutput.isClosedForWrite)
 
         serverConnection.close()
+        delay(100) // closing takes some time
+        assertTrue(serverConnection.isClosed)
+        assertTrue(serverOutput.isClosedForWrite)
+
         server.close()
+        assertTrue(server.isClosed)
     }
 }
--- a/ktor-network/posix/src/io/ktor/network/sockets/TCPServerSocketNative.kt
+++ b/ktor-network/posix/src/io/ktor/network/sockets/TCPServerSocketNative.kt
@@ -82,7 +82,7 @@ internal class TCPServerSocketNative(
     }
 
     override fun close() {
-        _socketContext.complete()
+        _socketContext.cancel("Server socket closed")
         _socketContext.invokeOnCompletion {
             ktor_shutdown(descriptor, ShutdownCommands.Both)
             // Descriptor is closed by the selector manager
--- a/ktor-network/jvmAndPosix/test/io/ktor/network/sockets/tests/UDPSocketTest.kt
+++ b/ktor-network/jvmAndPosix/test/io/ktor/network/sockets/tests/UDPSocketTest.kt
@@ -102,6 +102,7 @@ class UDPSocketTest {
         assertTrue(clientSocket.isClosed)
     }
 
+    @OptIn(DelicateCoroutinesApi::class)
     @Test
     fun testClose() = testSockets { selector ->
         val socket = aSocket(selector)
@@ -112,6 +113,8 @@ class UDPSocketTest {
 
         socket.socketContext.join()
         assertTrue(socket.isClosed)
+        assertTrue(socket.incoming.isClosedForReceive)
+        assertTrue(socket.outgoing.isClosedForSend)
     }
 
     @Test
--- a/ktor-network/posix/src/io/ktor/network/sockets/DatagramSocketNative.kt
+++ b/ktor-network/posix/src/io/ktor/network/sockets/DatagramSocketNative.kt
@@ -61,7 +61,7 @@ internal class DatagramSocketNative(
 
     override fun close() {
         receiver.cancel()
-        context.complete()
+        context.cancel("Socket closed")
         context.invokeOnCompletion {
             ktor_shutdown(descriptor, ShutdownCommands.Both)
             // Descriptor is closed by the selector manager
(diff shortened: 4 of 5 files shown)


Thomas-Vos:
Also made changes on JVM for consistency and found that `openWriteChannel` was not cancelled which is now cancelled as well. Please let me know what you think about this.

Thomas-Vos:
I'm working on moving some JVM code from `NioSocketImpl` to common code so it can be used on native.

ktor-network/common/src/io/ktor/network/selector/SocketBase.kt
```diff
@@ -0,0 +1,106 @@
...
+        get() = socketContext
+
+    override fun dispose() {
+        close()
+    }
+
+    override fun close() {
+        if (!closeFlag.compareAndSet(false, true)) return
+
+        readerJob.value?.channel?.close()
```
Thomas-Vos, line 36: This line was removed by https://github.com/ktorio/ktor/pull/4632/files on main branch, but this causes socket close to get stuck on JVM as well. For now I reverted it to get the test working, but please let me know what to do with this. Any ideas?

ktor-network/posix/src/io/ktor/network/sockets/TCPServerSocketNative.kt
```diff
@@ -65,28 +64,28 @@ internal class TCPServerSocketNative(
                 }
             }
         }
+        buildOrClose(clientDescriptor) {
```
Thomas-Vos, line 67: If accept descriptor setup fails the descriptor is now closed which was not done before

e5l:
Hey @Thomas-Vos, thanks for the PR. Let me check!

Thomas-Vos:
@e5l thanks, still working on it. Need to fix some issues. It is getting pretty big, let me know if you would prefer multiple smaller PRs with fixes.

ktor-network/common/src/io/ktor/network/selector/SocketBase.kt
```diff
@@ -0,0 +1,121 @@
+/*
+ * Copyright 2014-2025 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.network.selector
```
whyoleg, line 5: JIC: probably the package should be `io.ktor.network.sockets` :)

Thomas-Vos: Yes, thanks!

ktor-network/common/test/io/ktor/network/sockets/tests/TCPSocketTest.kt
```diff
@@ -169,19 +169,22 @@ class TCPSocketTest {
...
+            val serverConnection = serverConnectionPromise.await()
 
             client.cancel()
 
             assertFailsWith<CancellationException> {
                 readChannel.readByte()
             }
+
+            client.close()
+            serverConnection.close()
```
Thomas-Vos, line 187: Needed to add these close calls to prevent `jsNodeTest` task from getting stuck. Tests were all green but the gradle task just got stuck. Reproducible locally and on CI (https://ktor.teamcity.com/buildConfiguration/Ktor_KtorMatrixJavaScriptChromeNodeJs/316684). Is this a Kotlin bug?

ktor-network/common/src/io/ktor/network/sockets/SocketBase.kt
```diff
@@ -0,0 +1,121 @@
...
+        get() = socketContext
+
+    override fun dispose() {
+        close()
+    }
+
+    override fun close() {
+        if (!closeFlag.compareAndSet(false, true)) return
+
+        readerJob.value?.channel?.close()
```
Thomas-Vos, line 36: This line was removed by https://github.com/ktorio/ktor/pull/4632, but this causes socket close to get stuck on JVM as well on main branch. For now I reverted it to get the test working, but please let me know what to do with this. Any ideas? I don't think this should have been removed as it needed for the close to work properly (this is also exactly what was missing on Native)

ktor-network/common/test/io/ktor/network/sockets/tests/TCPSocketTest.kt
```diff
@@ -175,13 +254,56 @@ class TCPSocketTest {
...
+        val acceptJob = launch {
+            // The accept call should fail with IOException/PosixException because the socket was closed,
+            // but it must not be a bad descriptor error caused by closed descriptor in select call.
+            try {
                 socket.accept()
+            } catch (exception: IOException) {
+                assertFalse("Bad descriptor" in exception.message.orEmpty())
+            } catch (exception: Exception) {
+                assertTrue(exception.isPosixException())
             }
```
Thomas-Vos, line 266: I don't like how this is done, it would be better if `PosixException` extends `IOException`. That would make catching this exception in common code easier. Why is that not already the case?

Thomas-Vos:
@e5l PR is ready for review! Please check PR description for the fixes.

ktor-network/common/src/io/ktor/network/sockets/SocketBase.kt
Thomas-Vos: @bjhham any ideas?

e5l: I think it should be fixed by the recent remove flushAndClose call in the reader

ktor-network/common/src/io/ktor/network/sockets/SocketBase.kt
```diff
@@ -0,0 +1,121 @@
...
+    abstract fun attachForWritingImpl(channel: ByteChannel): ReaderJob
+
+    private inline fun <J : ChannelJob> attachFor(
+        name: String,
+        channel: ByteChannel,
+        ref: AtomicRef<J?>,
+        producer: () -> J
+    ): J {
+        if (closeFlag.value) {
+            val e = IOException("Socket closed")
```
e5l, line 63: cause

e5l: should we use cancellation exception as a cause?

e5l (approved):
Hey @Thomas-Vos, thank you for the huge investigation and good work!

Left minor comments, LGTM. Restarted CI

ktor-network/common/src/io/ktor/network/sockets/SocketBase.kt
Thomas-Vos: @e5l I moved this from JVM code to common code and changed it from `ClosedChannelException` to `IOException` so it works in common. I think `IOException` is better that `CancellationException` as reading/writing on a closed socket seems like an error. A `CancellationException` would just cause it to be ignored. What do you think is best?

ktor-network/common/src/io/ktor/network/sockets/SocketBase.kt
bjhham: Sorry I missed this - this line was actually quite important for the CIO client.  I'll investigate a new fix here.

Thomas-Vos: @bjhham but this is only in the socket `close` call. As long as you do not call close on a channel simultaneously or just before socket close from CIO I don't think it should be an issue. For example here: 

https://github.com/ktorio/ktor/blob/bd652e408f9e17c56fce061bb12d8adfc5c46fd9/ktor-client/ktor-client-cio/common/src/io/ktor/client/engine/cio/Endpoint.kt#L113-L114

If you replace the first line with `flushAndClose` it should not cause any issues. The close channel call launches a new coroutine, meaning you get concurrent flushes. There may be more places where this is done wrong, but I am not that familiar with the CIO code.

bjhham: In the case of KTOR-8105, it was caused by the TLS connector sending a close alert packet at the same time this flush/close was invoked.  In https://github.com/ktorio/ktor/pull/4670 I've made these calls run sequentially.

PR #4645, opened 2025-02-03
KTOR-7775 BasicAuth: Add clearToken method

osipxd:
**Subsystem**
Client, Basic Auth

**Motivation**
[KTOR-7775](https://youtrack.jetbrains.com/issue/KTOR-7775) Auth: BasicAuthProvider caches credentials until process death

**Solution**
Add `BasicAuthProvider.clearToken` similarly to `BearerAuthProvider`. So now the token can be cleared the following way:
```kotlin
import io.ktor.client.plugins.auth.authProvider

client.authProvider<BasicAuthProvider>()?.clearToken()
```

Probably, we should come up with a common solution for clearing authorization tokens, but unlikely in 3.1.0.

### DIFF
--- a/ktor-client/ktor-client-plugins/ktor-client-auth/common/src/io/ktor/client/plugins/auth/providers/BasicAuthProvider.kt
+++ b/ktor-client/ktor-client-plugins/ktor-client-auth/common/src/io/ktor/client/plugins/auth/providers/BasicAuthProvider.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2024 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ * Copyright 2014-2025 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
  */
 
 package io.ktor.client.plugins.auth.providers
@@ -161,6 +161,10 @@ public class BasicAuthProvider(
     override suspend fun refreshToken(response: HttpResponse): Boolean {
         tokensHolder.setToken(credentials)
         return true
+    }
+
+    public fun clearToken() {
+        tokensHolder.clearToken()
     }
 }
 
--- a/ktor-client/ktor-client-plugins/ktor-client-auth/common/src/io/ktor/client/plugins/auth/providers/BasicAuthProvider.kt
+++ b/ktor-client/ktor-client-plugins/ktor-client-auth/common/src/io/ktor/client/plugins/auth/providers/BasicAuthProvider.kt
@@ -163,6 +163,19 @@ public class BasicAuthProvider(
         return true
     }
 
+    /**
+     * Clears the currently stored authentication tokens from the cache.
+     *
+     * This method should be called in the following cases:
+     * - When the credentials have been updated and need to take effect
+     * - When you want to force re-authentication
+     * - When you want to clear sensitive authentication data
+     *
+     * Note: The result of [credentials] invocation is cached internally.
+     * Calling this method will force the next authentication attempt to fetch fresh credentials.
+     *
+     * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.client.plugins.auth.providers.BasicAuthProvider.clearToken)
+     */
     public fun clearToken() {
         tokensHolder.clearToken()
     }
--- a/ktor-client/ktor-client-plugins/ktor-client-auth/common/src/io/ktor/client/plugins/auth/providers/BasicAuthProvider.kt
+++ b/ktor-client/ktor-client-plugins/ktor-client-auth/common/src/io/ktor/client/plugins/auth/providers/BasicAuthProvider.kt
@@ -176,6 +176,7 @@ public class BasicAuthProvider(
      *
      * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.client.plugins.auth.providers.BasicAuthProvider.clearToken)
      */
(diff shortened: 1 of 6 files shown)


ktor-client/ktor-client-plugins/ktor-client-auth/common/src/io/ktor/client/plugins/auth/providers/BasicAuthProvider.kt
```diff
@@ -162,6 +162,23 @@ public class BasicAuthProvider(
...
+     * - When the credentials have been updated and need to take effect
+     * - When you want to force re-authentication
+     * - When you want to clear sensitive authentication data
+     *
+     * Note: The result of [credentials] invocation is cached internally.
+     * Calling this method will force the next authentication attempt to fetch fresh credentials.
+     *
+     * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.client.plugins.auth.providers.BasicAuthProvider.clearToken)
+     */
+    public fun clearToken() {
```
osipxd, line 179: We could make this method a part of `AuthProvider` interface, but it would be potentially breaking change. Only if not declare an empty implementation by default.

bjhham: You could introduce it with a default implementation of throwing an `UnsupportedOperationException`.  Then we can drop the default implementation for the next major release as a breaking change.  It's breaking the Liskov rule, but I think the pros outweigh the cons for having a standard function here.

osipxd: I'm in doubt. If we add this method, we assume that all implementations of `AuthProvider` should be caching, as this method doesn't make any sense for stateless implementations. Probably it would be better to introduce `CachingAuthProvider` containing `clearToken`, and providing a way to create custom auth providers using `AuthTokenHolder` under the hood?

```kotlin
/**
 * A base implementation for an authentication provider with authentication state caching.
 *
 * @param AuthState The type representing the authentication state, which contains necessary token or credential information.
 * @param fetchState A function for fetching the authentication state, typically used during first access
 * or during refresh (if [refreshState] is not overridden).
 */
public abstract class CachingAuthProvider<AuthState : Any>(
    private val fetchState: suspend () -> AuthState?,
) : AuthProvider {

    // If we go this way, AuthTokenHolder should be renamed to something else
    private val tokenHolder = AuthTokenHolder(fetchState)

    /** Returns the cached authentication state, or fetches it if it wasn't loaded before. */
    protected suspend fun loadState(): AuthState? = tokenHolder.loadToken()

    final override suspend fun addRequestHeaders(request: HttpRequestBuilder, authHeader: HttpAuthHeader?) {
        val tokens = loadState() ?: return
        addRequestHeaders(request, authHeader, tokens)
    }

    /**
     * Adds an authentication method headers and credentials.
     *
     * @param request The builder for an authenticated request
     * @param authHeader The value of `WWW-Authenticate` header from failed response, if exists
     * @param state The authentication state containing token or credential information for the request
     */
    protected abstract suspend fun addRequestHeaders(
        request: HttpRequestBuilder,
        authHeader: HttpAuthHeader?,
        state: AuthState,
    )

    final override suspend fun refreshToken(response: HttpResponse): Boolean {
        val newState = tokenHolder.setToken { refreshState(response) }
        return newState != null
    }

    /**
     * Refreshes the authentication state if required.
     *
     * @param response The HTTP response received after a request.
     * @return The updated authentication state or `null` if the state couldn't be refreshed.
     */
    protected open suspend fun refreshState(response: HttpResponse): AuthState? = fetchState()

    /** Clears the currently stored authentication state from the cache. */
    public fun clear() {
        tokenHolder.clearToken()
    }
}
```

osipxd: BTW, just noticed that we have a request to make `AuthTokenHolder` re-usable:
- [KTOR-8126](https://youtrack.jetbrains.com/issue/KTOR-8126) ktor-client-auth with custom providers

But we also have a request to make it possible to completely disable caching for auth providers:
- [KTOR-4759](https://youtrack.jetbrains.com/issue/KTOR-4759) Auth: BearerAuthProvider caches result of loadToken until process death

My proposed `CachingAuthProvider` implementation fixes the first issue, but doesn’t address the second one.

osipxd: So we have two different use cases:
1. A user uses some third-party token storage and wants to delegate responsibility for its lifecycle management to this storage (for example, Firebase SDK). In this case, `AuthProvider` shouldn't cache the token
2. A user uses their own backend to get a token and (likely) doesn't want extra requests to be made when the token is still valid. For this case caching works fine, but there is a danger of forgetting to clean it on logout.

Considering these two scenarios, I'd force users to explicitly choose whether they want caching enabled or not. For example, passing a flag to `loadToken`:
```kotlin
install(Auth) {
    bearer {
        loadTokens(cacheEnabled = false) { getFirebaseTokens() }

        // Deprecated. Specifies cacheEnabled = true implicitly.
        loadTokens { getFirebaseTokens() }
    }
}
```
Alternative solution would be to make it possible to override the implementation of token holder, but it seems a more complex though more flexible solution.

PR #4646, opened 2025-02-03
Remove use of Platform.isExperimentalMM

anton-erofeev:
**Subsystem**
ktor-utils

**Motivation**
since isExperimentalMM(Platform.kt) [was marked as deprecated](https://github.com/JetBrains/kotlin/commit/db74ab9e8e66c14c5d4b002c099c5f9c4033248d#diff-663457c28464a27bda27e339f1cc45ad7c5bc17c364f74288f5fb4d9274f7ea1R178) and now always returns true, it can be replaced with the value true

**Solution**
Removed use of Platform.isExperimentalMM


bjhham (changes requested):
Could you introduce a deprecation warning on this and also `PlatformUtils.IS_NEW_MM_ENABLED` so we can remove this in a future release?

bjhham (approved):

PR #4647, opened 2025-02-04
KTOR-8138 Fix idleTimeout for Jetty engine

bjhham:
**Subsystem**
Server, Jetty

**Motivation**
[KTOR-8138](https://youtrack.jetbrains.com/issue/KTOR-8138) Jetty idleTimeout not working

This issue was discovered by the Space team after a network interruption causing many connections to get stuck in the Packages feature.

**Solution**
This introduces a simple implementation for handling idle connection timeouts on reads/writes for the Jetty and (Jakarta) Jetty engines.  It makes use of the existing `idleTimeout` configuration property, which doesn't have any effect due to the [use of async servlets](https://javadoc.jetty.org/jetty-12/org/eclipse/jetty/server/AbstractConnector.html#setIdleTimeout(long)).


ktor-server/ktor-server-jetty-jakarta/jvm/test/io/ktor/tests/server/jetty/jakarta/JettyIdleTimeoutTest.kt
```diff
@@ -0,0 +1,131 @@
...
+
+            val response = inputStream.bufferedReader().readText()
+            assertEquals("HTTP/1.1 500 Server Error", response.lines().first())
+        }
+    }
+
+    @Test
+    fun idleTimeoutResponseWriter(): Unit = runTest {
+        val responseMessage = "Hello, world!".toByteArray()
+        var writeError: Exception? = null
```
osipxd, line 82: It's probably better to await the exception using deferred:
```kotlin
val writeError = CompletableDeferred<Exception>()

// Inside catch block
writeError.complete(cause)
        
// And later
assertIs<CloseByteChannelException>(writeError.await())
```

osipxd (approved):

ktor-server/ktor-server-jetty-jakarta/jvm/test/io/ktor/tests/server/jetty/jakarta/JettyIdleTimeoutTest.kt
bjhham: Good call 👍

PR #4648, opened 2025-02-05
KTOR-6185 Fix for OutOfMemoryError when uploading large files

bjhham:
**Subsystem**
Core, I/O

**Motivation**
[KTOR-6185](https://youtrack.jetbrains.com/issue/KTOR-6185) OutOfMemoryError when sending a large binary file through ByteReadChannel converted from InputStream

**Solution**
The problem was caused by directory transferring the entire contents of a `RawSource` to the `ByteChannel`'s `writeBuffer`, causing the application to run out of memory.  To avoid this, we instead transfer up to our maximum flush buffer size then suspend until we have room to write more.

### DIFF
--- a/ktor-io/common/test/ByteWriteChannelOperationsTest.kt
+++ b/ktor-io/common/test/ByteWriteChannelOperationsTest.kt
@@ -1,3 +1,7 @@
+/*
+ * Copyright 2014-2025 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
 import io.ktor.utils.io.discard
 import io.ktor.utils.io.reader
 import io.ktor.utils.io.writeBuffer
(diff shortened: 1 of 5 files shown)


ktor-io/common/test/ByteWriteChannelOperationsTest.kt
```diff
@@ -0,0 +1,40 @@
...
+import io.ktor.utils.io.writeBuffer
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.test.runTest
+import kotlinx.io.files.Path
+import kotlinx.io.files.SystemFileSystem
+import kotlin.test.Test
+
+/*
+ * Copyright 2014-2025 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
```
osipxd, line 12: Minor. Could you move this header to the top of the file?

ktor-io/common/test/ByteWriteChannelOperationsTest.kt
```diff
@@ -0,0 +1,40 @@
...
+private const val KB = 1024L
+private const val GB = KB * KB * KB
```
e5l, line 14: :D

e5l (approved):
lgtm!

ktor-io/common/test/ByteWriteChannelOperationsTest.kt
bjhham: I blame the IDE!

PR #4649, opened 2025-02-06
Fix environ for mingw

e5l:
Fix linking error for windows native build


ktor-server/ktor-server-core/mingwX64/src/io/ktor/server/engine/EnvironmentUtils.mingwX64.kt
```diff
@@ -16,3 +16,23 @@ internal actual fun setEnvironmentProperty(key: String, value: String): Unit = m
...
+        var current: CPointer<WCHARVar>? = rawEnv
+        while (current != null) {
+            val keyValue = current.toKString()
+            if (keyValue.isEmpty()) break
+            current += keyValue.length + 1
+
+            val (key, value) = keyValue.splitPair('=') ?: continue
+            if (key.startsWith("ktor.")) {
+                add(key to value)
+            }
```
osipxd, line 33: Just curious. Is it intentionally slightly different from the code in nix source-set?

```kotlin
val (key, value) = keyValue.splitPair('=') ?: continue
if (key.startsWith("ktor.")) {
    add(key to value)
}
```
vs
```kotlin
if (keyValue.startsWith("ktor.")) {
    val (key, value) = keyValue.splitPair('=') ?: continue
    add(key to value)
}
```

osipxd (approved):
👍

ktor-server/ktor-server-core/mingwX64/src/io/ktor/server/engine/EnvironmentUtils.mingwX64.kt
e5l: nope, didn't noticed it

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/HttpStatementTest.kt
```diff
@@ -40,14 +40,15 @@ class HttpStatementTest : ClientLoader() {
...
-    fun testGZipFromSavedResponse() = clientTests(except("native:CIO", "web:CIO")) {
+    fun testGZipFromSavedResponse() = clientTests(except("native:CIO", "web:CIO", "WinHttp")) {
         config {
             ContentEncoding {
                 gzip()
             }
         }
 
         test { client ->
+            println(client)
```
osipxd, line 51: Was added for debug?

e5l: yep, let me drop

PR #4651, opened 2025-02-06
KTOR-8163 WasmJS: Add callbacks before awaiting connection

osipxd:
**Subsystem**
Client (WasmJS)

**Motivation**
[KTOR-8163](https://youtrack.jetbrains.com/issue/KTOR-8163) WasmJS WebSocket client sometimes drops a frame received immediately after handshake

**Solution**
Cherry-pick fix made for JS engine: #4012


bjhham (approved):
Nice 👍

PR #4657, opened 2025-02-07
KTOR-8172 Fix issue in JVM byte channel read function

bjhham:
**Subsystem**
Core, I/O

**Motivation**
[KTOR-8172](https://youtrack.jetbrains.com/issue/KTOR-8172) ByteChannel read issue on min > 1

I found this problem when looking through the code for potential infinite loops.  There were no references to the function where min > 1 so it's not causing problems internally, but it's part of our public API so it seems important to fix.

**Solution**
When the min is supplied, we ought to use this as an argument for the `awaitContent()` function and throw EOF accordingly when said content is unavailable.

### DIFF
--- a/ktor-io/jvm/src/io/ktor/utils/io/ByteReadChannelOperations.jvm.kt
+++ b/ktor-io/jvm/src/io/ktor/utils/io/ByteReadChannelOperations.jvm.kt
@@ -96,9 +96,8 @@ public suspend fun ByteReadChannel.copyTo(channel: WritableByteChannel, limit: L
         }
     }
 
-    while (copied < limit) {
+    while (copied < limit && !isClosedForRead) {
         read(min = 0, consumer = copy)
-        if (isClosedForRead) break
     }
 
     closedCause?.let { throw it }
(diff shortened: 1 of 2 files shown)


ktor-io/jvm/src/io/ktor/utils/io/ByteReadChannelOperations.jvm.kt
```diff
@@ -184,19 +183,17 @@ public fun ByteReadChannel.readAvailable(block: (ByteBuffer) -> Int): Int {
  *
  * @param min amount of bytes available for read, should be positive or zero
  * @param consumer to be invoked when at least [min] bytes available for read
+ * @throws EOFException when there are less than [min] bytes available after the channel is closed
  */
 @OptIn(InternalAPI::class)
 public suspend inline fun ByteReadChannel.read(min: Int = 1, noinline consumer: (ByteBuffer) -> Unit) {
     require(min >= 0) { "min should be positive or zero" }
-    if (availableForRead > 0 && availableForRead >= min) {
```
bjhham, line 191: The `availableForRead` checks seem redundant when followed with `awaitContent` since the same check is made in the first line of the function.

e5l: Should we still try to make an optimistic read without `awaitContent`?

e5l (approved):

ktor-io/jvm/src/io/ktor/utils/io/ByteReadChannelOperations.jvm.kt
bjhham: The awaitContent will just return immediately when the buffer has the required content, so it's the same code 🤷

PR #4667, opened 2025-02-12
KTOR-6671 Fix Resources: a / route isn't resolved when there is a sib…

e5l:
Fix [KTOR-6671](https://youtrack.jetbrains.com/issue/KTOR-6671) Resources: a / route isn't resolved when there is a sibling `staticResources`

### DIFF
--- a/ktor-server/ktor-server-core/jvm/src/io/ktor/server/http/content/StaticContent.kt
+++ b/ktor-server/ktor-server-core/jvm/src/io/ktor/server/http/content/StaticContent.kt
@@ -607,16 +607,21 @@ private fun Route.staticContentRoute(
     remotePath: String,
     autoHead: Boolean,
     handler: suspend (ApplicationCall).() -> Unit
-) = route(remotePath) {
-    route("{$pathParameterName...}") {
-        get {
-            call.handler()
-        }
-        if (autoHead) {
-            method(HttpMethod.Head) {
-                install(StaticContentAutoHead)
-                handle {
-                    call.handler()
+) = createChild(object : RouteSelector() {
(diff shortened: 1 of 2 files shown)


ktor-server/ktor-server-core/jvm/src/io/ktor/server/http/content/StaticContent.kt
```diff
@@ -607,16 +607,21 @@ private fun Route.staticContentRoute(
...
-        }
-        if (autoHead) {
-            method(HttpMethod.Head) {
-                install(StaticContentAutoHead)
-                handle {
-                    call.handler()
+) = createChild(object : RouteSelector() {
+    override suspend fun evaluate(context: RoutingResolveContext, segmentIndex: Int): RouteSelectorEvaluation =
+        RouteSelectorEvaluation.Success(quality = RouteSelectorEvaluation.qualityTailcard)
+}).apply {
```
bjhham, line 613: Extracting the selector here could help with readability.  Something like:

```
createChild(TailcardSelector).apply { /* ... */ }

// Adds lower priority to the route so that it can be used as a fallback
object TailcardSelector : RouteSelector() {
    override suspend fun evaluate(context: RoutingResolveContext, segmentIndex: Int): RouteSelectorEvaluation =
        RouteSelectorEvaluation.Success(quality = RouteSelectorEvaluation.qualityTailcard)
}
```

e5l: yep, but I want to avoid creating new API with the fix

PR #4668, opened 2025-02-12
KTOR-8015 CIO: Do not accept CR as line delimiter in requests

osipxd:
**Subsystem**
Server (CIO)

**Motivation**
[KTOR-8015](https://youtrack.jetbrains.com/issue/KTOR-8015) Server accepts \r without a following \n as a valid line terminator in chunked transfer encoding

**Solution**
Add a parameter `lineEnding` to `readUTF8Line` to make it possible to ensure that a proper line ending is used.

### DIFF
--- a/ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
+++ b/ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
@@ -1,8 +1,6 @@
 /*
- * Copyright 2014-2024 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ * Copyright 2014-2025 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
  */
-
-@file:Suppress("DEPRECATION")
 
 package io.ktor.utils.io
 
@@ -250,7 +248,7 @@ public suspend fun ByteReadChannel.readRemaining(max: Long): Source {
 }
 
 /**
- * Reads all available bytes to [dst] buffer and returns immediately or suspends if no bytes available
+ * Reads all available bytes to [buffer] and returns immediately or suspends if no bytes available
  *
  * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.utils.io.readAvailable)
  *
@@ -399,7 +397,6 @@ private const val LF: Byte = '\n'.code.toByte()
  * Reads a line of UTF-8 characters to the specified [out] buffer.
  * It recognizes CR, LF and CRLF as a line delimiter.
  *
- *
  * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.utils.io.readUTF8LineTo)
  *
  * @param out the buffer to write the line to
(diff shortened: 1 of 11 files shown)


ktor-server/ktor-server-test-suites/jvm/src/io/ktor/server/testing/suites/SustainabilityTestSuite.kt
```diff
@@ -605,6 +607,49 @@ abstract class SustainabilityTestSuite<TEngine : ApplicationEngine, TConfigurati
...
+                    "47\r\n",
+                    "0\r\n\r\n",
+                    "GET /admin HTTP/1.1\r\n",
+                    "Host: localhost:8080\r\n",
+                    "Transfer-Encoding: chunked\r\n",
+                    "\r\n",
+                    "0\r\n\r\n",
+                )
+            }
+            delay(1.seconds)
```
osipxd, line 646: Any ideas on how to wait until the server settles down before going to assertions?

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
```diff
@@ -395,6 +396,32 @@ public suspend fun ByteReadChannel.discard(max: Long = Long.MAX_VALUE): Long {
 private const val CR: Byte = '\r'.code.toByte()
 private const val LF: Byte = '\n'.code.toByte()
 
+@JvmInline
+public value class LineEndingMode private constructor(private val mode: Int) {
```
osipxd, line 400: I could fix the problem only for chunked body, but it seems we should fix line endings in request/response parsing.

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
```diff
@@ -450,6 +489,12 @@ public suspend fun ByteReadChannel.readUTF8LineTo(out: Appendable, max: Int = In
     }
 }
 
+private fun LineEndingMode.assertIncludes(other: LineEndingMode) {
+    if (other !in this) {
+        throw EOFException("Unexpected line ending $other, while expected $this")
+    }
+}
```
bjhham, line 496: I don't think EOF is technically correct since it's malformed data but the input hasn't ended, maybe just an IOException would do.

Could there be scenarios here where you'd simply want to ignore the partial line ending?

osipxd: > Could there be scenarios here where you'd simply want to ignore the partial line ending?

That is how Netty works. It just ignores everything until the LF line ending. It makes a server more tolerant to malformed requests. The question is, do we want to make it tolerant?

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
```diff
@@ -408,8 +435,16 @@ private const val LF: Byte = '\n'.code.toByte()
  * @return `true` if a new line separator was found or max bytes appended. `false` if no new line separator and no bytes read.
  * @throws TooLongLineException if max is reached before encountering a newline or end of input
  */
-@OptIn(InternalAPI::class, InternalIoApi::class)
 public suspend fun ByteReadChannel.readUTF8LineTo(out: Appendable, max: Int = Int.MAX_VALUE): Boolean {
+    return readUTF8LineTo(out, max, lineEnding = LineEndingMode.Any)
+}
+
+@OptIn(InternalAPI::class, InternalIoApi::class)
+public suspend fun ByteReadChannel.readUTF8LineTo(
```
osipxd, line 443: I will add KDoc comments if this solution with the new API is acceptable

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
marychatte: I think it's better to throw an exception as Jetty and Tomcat do it. Also, a user says that the behavior of Google Load Balancer (in Netty it's the same: parse input as one request with two chunks) is incorrect for them

bjhham (approved):

PR #4670, opened 2025-02-13
KTOR-8105 Fix for concurrent flush attempts breaking CIO client (again)

bjhham:
**Subsystem**
Client, CIO

**Motivation**
Once again fixing these issues:
- [KTOR-8105](https://youtrack.jetbrains.com/issue/KTOR-8105) Race condition when writing to a buffer leads to NPE inside CIOReaderKt.readFrom
- [KTOR-8086](https://youtrack.jetbrains.com/issue/KTOR-8086) NPE in readBuffer
- [KTOR-8096](https://youtrack.jetbrains.com/issue/KTOR-8096) ArrayIndexOutOfBounds kotlinx-io

**Solution**
Previously, I removed a redundant close from the base socket implementation, but it would seem there is some other parts of the code base that rely on this behaviour.  So, I instead defer the close call to after the encoder job is finished closing.  Bonus fix: I noticed we're not encrypting our close records after the handshake is complete.  This is incorrect behaviour, so I fixed that as well.

### DIFF
--- a/ktor-network/common/src/io/ktor/network/sockets/SocketBase.kt
+++ b/ktor-network/common/src/io/ktor/network/sockets/SocketBase.kt
@@ -33,6 +33,7 @@ internal abstract class SocketBase(
     override fun close() {
         if (!closeFlag.compareAndSet(false, true)) return
 
+        // TODO this can be dangerous if there is another thread writing to this
         readerJob.value?.channel?.close()
         writerJob.value?.cancel()
         checkChannels()
--- a/ktor-network/ktor-network-tls/jvm/src/io/ktor/network/tls/TLSClientHandshake.kt
+++ b/ktor-network/ktor-network-tls/jvm/src/io/ktor/network/tls/TLSClientHandshake.kt
@@ -114,22 +114,26 @@ internal class TLSClientHandshake(
                 rawOutput.writeRecord(record)
             } catch (cause: Throwable) {
                 channel.close(cause)
+                break
             }
         }
     }.apply {
         invokeOnClose {
             launch(CoroutineName("cio-tls-closer")) {
-                val closeRecord = TLSRecord(
-                    TLSRecordType.Alert,
-                    packet = buildPacket {
-                        writeByte(TLSAlertLevel.WARNING.code.toByte())
-                        writeByte(TLSAlertType.CloseNotify.code.toByte())
-                    }
-                )
-                val record = if (useCipher) cipher.encrypt(closeRecord) else closeRecord
-                rawOutput.writeRecord(record)
-                rawOutput.flushAndClose()
-                closeTask.complete()
+                try {
+                    val closeRecord = TLSRecord(
+                        TLSRecordType.Alert,
+                        packet = buildPacket {
+                            writeByte(TLSAlertLevel.WARNING.code.toByte())
+                            writeByte(TLSAlertType.CloseNotify.code.toByte())
+                        }
+                    )
+                    val record = if (useCipher) cipher.encrypt(closeRecord) else closeRecord
+                    rawOutput.writeRecord(record)
+                    rawOutput.flushAndClose()
+                } finally {
+                    closeTask.complete()
+                }
             }
         }
     }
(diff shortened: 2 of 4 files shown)


Thomas-Vos:
https://github.com/ktorio/ktor/pull/4643#discussion_r1954369981

ktor-network/ktor-network-tls/jvm/src/io/ktor/network/tls/TLSClientSessionJvm.kt
```diff
@@ -49,15 +60,15 @@ private class TLSSocket(
...
+            base.input.consumeEach { record ->
                 val packet = record.packet
                 val length = packet.remaining
                 when (record.type) {
                     TLSRecordType.ApplicationData -> {
                         pipe.writePacket(record.packet)
                         pipe.flush()
                     }
-                    else -> throw TLSException("Unexpected record ${record.type} ($length bytes)")
+                    else -> throw TlsException("Unexpected record ${record.type} ($length bytes)")
```
osipxd, line 71: We have to throw `TLSException` for compatibility with code which hasn't been updated to use the new type in catch 😢

ktor-network/common/src/io/ktor/network/sockets/SocketBase.kt
```diff
@@ -33,6 +33,7 @@ internal abstract class SocketBase(
     override fun close() {
         if (!closeFlag.compareAndSet(false, true)) return
 
+        // TODO this can be dangerous if there is another thread writing to this
```
e5l, line 36: could we detect this and log?

ktor-network/ktor-network-tls/jvm/src/io/ktor/network/tls/TLSClientHandshake.kt
```diff
@@ -101,36 +102,44 @@ internal class TLSClientHandshake(
...
-                    channel.close(cause)
-                }
+        for (rawRecord in channel) {
+            try {
+                val record = if (useCipher) cipher.encrypt(rawRecord) else rawRecord
+                if (rawRecord.type == TLSRecordType.ChangeCipherSpec) useCipher = true
+
+                rawOutput.writeRecord(record)
+            } catch (cause: Throwable) {
+                channel.close(cause)
```
e5l, line 116: should we break?

e5l (commented):
left some comments

ktor-network/common/src/io/ktor/network/sockets/SocketBase.kt
bjhham: I'll log a ticket to investigate.  I don't think it makes sense to change the code here, but we could easily find out what parts of our code base rely on this close operation.  I'm thinking it should wait for the readerJob to finish and only close after some timeout since the ramifications of threading issues with kotlinx-io can be quite troublesome.

Thomas-Vos: Or awaitClosed could be used, I think it should already work. 
Would it make sense to replace
```
readerJob.value?.channel?.close()
```
with
```
launch { readerJob.value?.channel?.flushAndClose() }
```

bjhham: I'll explore some options in [the other ticket](https://youtrack.jetbrains.com/issue/KTOR-8201/Fix-socket-channel-close-handling).  Atm I don't think we really have a safe mechanism for flushing and closing a ByteChannel from outside the coroutine writing to it.

ktor-network/ktor-network-tls/jvm/src/io/ktor/network/tls/TLSClientSessionJvm.kt
```diff
@@ -24,15 +26,26 @@ internal actual suspend fun openTLSSession(
     val handshake = TLSClientHandshake(input, output, config, context)
     try {
         handshake.negotiate()
-    } catch (cause: ClosedSendChannelException) {
-        throw TLSException("Negotiation failed due to EOS", cause)
+    } catch (cause: Exception) {
```
e5l, line 29: Throwable

e5l (changes requested):

e5l (approved):

PR #4671, opened 2025-02-13
KTOR-8194 Fix SelectorManager dispatcher ignored on Native

Thomas-Vos:
[KTOR-8194](https://youtrack.jetbrains.com/issue/KTOR-8194) SelectorManager creates own thread instead of provided dispatcher on Native

Now Native now longer creates a separate thread per SelectorManager, but can use IO pool when IO dispatcher is given, just like JVM.

Also removed Job to make it consistent with JVM.

## Summary by CodeRabbit

- **Refactor**
  - Enhanced asynchronous network processing and resource lifecycle management for improved responsiveness and efficiency.
  - Simplified coroutine context management in the `WorkerSelectorManager` for better flexibility.
- **Tests**
  - Streamlined network socket closure verification to ensure more stable and accurate handling of resources.


Thomas-Vos:
Still have to check why the testEchoByteArray[macosX64] failed. Not sure whether this is caused by the changes in this PR or if the issue was already present. Cannot reproduce locally, currently unsure how this failure is even possible.

Thomas-Vos:
Test on CI is no longer failing after two runs, so #4680 fixed it or the test is flaky.

Thomas-Vos:
@e5l PR is ready for review

ktor-network/nix/src/io/ktor/network/selector/SelectUtilsNix.kt
```diff
@@ -48,7 +48,7 @@ internal actual class SelectorHelper {
     }
 
     actual fun start(scope: CoroutineScope): Job {
-        return scope.launch(CoroutineName("selector")) {
+        return scope.launch {
```
e5l, line 51: Could you tell me why we don't need the name? I thought it can help in debug

e5l (approved):
lgtm

ktor-network/nix/src/io/ktor/network/selector/SelectUtilsNix.kt
Thomas-Vos: I moved it to https://github.com/ktorio/ktor/pull/4671/files#diff-675619f9bef0e1f4942b7d0716aa05e1f3c09dad8aa1c0c79004d931e6dfd5a5R12 to keep it a bit more consistent to where it is defined on JVM. So it is not removed.

Thomas-Vos: It does indeed help with naming on supported platforms so it is important to keep.

PR #4678, opened 2025-02-14
KTOR-8190 Fix OOM issue in ByteChannel.copyTo

bjhham:
**Subsystem**
Core, I/O

**Motivation**
[KTOR-8190](https://youtrack.jetbrains.com/issue/KTOR-8190) OOM on SourceByteReadChannel for large input

**Solution**
When wrapping a `Source` with a `SourceByteReadChannel`, there were two spots where we can blow up our memory using `transferTo` because this operation transfers the entire contents of the underlying `RawSource`.  The fix is to always use `write()` or `readTo()` with the buffer size, so then we'll simply transfer our current big blob of memory then wait for a flush before proceeding.


osipxd (approved):

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
```diff
@@ -174,9 +160,11 @@ public suspend fun ByteReadChannel.copyTo(channel: ByteWriteChannel): Long {
     var result = 0L
     try {
         while (!isClosedForRead) {
-            result += readBuffer.transferTo(channel.writeBuffer)
+            if (readBuffer.exhausted()) awaitContent()
```
e5l, line 163: Can this change lead to the source cancellation ignoring?

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
```diff
@@ -174,9 +160,11 @@ public suspend fun ByteReadChannel.copyTo(channel: ByteWriteChannel): Long {
...
+            val count = readBuffer.remaining
+            readBuffer.readTo(channel.writeBuffer, count)
```
e5l, line 165: This operation looks strange to me: instead of changing a whole buffer, we copy only part of it. These bytes are already in memory, and moving the full buffer should be cheap and not cause any memory consumption. What we really need to figure out is how to prevent reading new bytes to the source (backpressure). 

 It also may cause severe performance impact. Please add a benchmark.

e5l (changes requested):

ktor-io/jvm/test/ByteWriteLargeTransfersTest.kt
```diff
@@ -0,0 +1,62 @@
...
+import kotlinx.coroutines.test.runTest
+import kotlinx.io.buffered
+import kotlinx.io.files.Path
+import kotlinx.io.files.SystemFileSystem
+import kotlin.test.Test
+
+private const val KB = 1024L
+private const val GB = KB * KB * KB
+
+class ByteWriteLargeTransfersTest {
```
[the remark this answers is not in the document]
bjhham: Oh yeah, looks like it.  I keep forgetting file I/O is multi-platform now 😄

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
bjhham: > These bytes are already in memory, and moving the full buffer should be cheap and not cause any memory consumption.

The problem occurs when using `SourceByteReadChannel`, which exposes the `Source` instance as its `readBuffer`, but the `Source` instance could be infinitely large.  So `readTo(remaining)` should be functionally the same as `source.buffer.transferTo()`.  I can look into changing the `SourceByteReadChannel` implementation so that its `readBuffer` reflects the `buffer` property of the `Source`, then use `require()` in its `awaitContent()`.

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
bjhham: I see what you mean, channel is closed during wait then next access to `readBuffer` throws ⚠️

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
bjhham: Yeah, simply updating the `SourceByteReadChannel` to use the source's buffer as its `readBuffer` seems to do the trick, with no changes required for the byte channel operations 🎉

ktor-io/jvm/test/ByteWriteLargeTransfersTest.kt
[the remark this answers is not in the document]
bjhham: Oh right, that's why I didn't move it to common 😅 

```
UnsupportedOperationException: Module os could not be loaded
  at kotlin.captureStackTrace(/mnt/agent/temp/buildTmp/_karma_webpack_735482/commons.js:135)
  at io.ktor:ktor-io_test.kotlin.captureStackTrace__externalAdapter(/mnt/agent/temp/buildTmp/_karma_webpack_735482/bfba7d90d6de907e641d.wasm)
  at io.ktor:ktor-io_test.kotlin.Throwable.<init>(/mnt/agent/temp/buildTmp/_karma_webpack_735482/bfba7d90d6de907e641d.wasm)
  at io.ktor:ktor-io_test.kotlin.Throwable.<init>(/mnt/agent/temp/buildTmp/_karma_webpack_735482/bfba7d90d6de907e641d.wasm)
  at io.ktor:ktor-io_test.kotlin.Exception.<init>(/mnt/agent/temp/buildTmp/_karma_webpack_735482/bfba7d90d6de907e641d.wasm)
  at io.ktor:ktor-io_test.kotlin.RuntimeException.<init>(/mnt/agent/temp/buildTmp/_karma_webpack_735482/bfba7d90d6de907e641d.wasm)
  at io.ktor:ktor-io_test.kotlin.UnsupportedOperationException.<init>(/mnt/agent/temp/buildTmp/_karma_webpack_735482/bfba7d90d6de907e641d.wasm)
  at io.ktor:ktor-io_test.kotlinx.io.node.loadModule(/mnt/agent/temp/buildTmp/_karma_webpack_735482/bfba7d90d6de907e641d.wasm)
  at lambda.invoke(/mnt/agent/temp/buildTmp/_karma_webpack_735482/bfba7d90d6de907e641d.wasm)
```

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
e5l: Nice

e5l (approved):
lgtm

PR #4680, opened 2025-02-14
Fix double actualClose call

Thomas-Vos:
The actualClose call could be called twice due to race condition when closing. Seems highly unlikely, but this could cause a descriptor to be closed twice, causing errors.

This piece of code was moved from JVM, where a double actualClose call would not have caused an issue (because closing jvm channel twice is ok), but on Native closing a descriptor twice is a serious issue.

Was able to reproduce this issue by adding `runBlocking { delay(1000) }` on line 85 just before `checkChannels()` call


e5l (approved):

e5l:
Hey @Thomas-Vos, thanks for the PR! Merged

PR #4683, opened 2025-02-18
KTOR-8201 Add safe close for readerJob

bjhham:
**Subsystem**
Core, I/O

**Motivation**
[KTOR-8201](https://youtrack.jetbrains.com/issue/KTOR-8201) Fix socket channel close handling

**Solution**
This introduces a suspending close for reader jobs, which will cancel the nested job, wait for it to join, then perform the flush and close.  This should prevent two coroutines from writing / flushing simultaneously.

### DIFF
--- a/ktor-io/api/ktor-io.api
+++ b/ktor-io/api/ktor-io.api
@@ -295,6 +295,7 @@ public abstract interface annotation class io/ktor/utils/io/PublicAPICandidate :
 }
 
 public final class io/ktor/utils/io/ReaderJob : io/ktor/utils/io/ChannelJob {
+	public final fun flushAndClose (Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
 	public final fun getChannel ()Lio/ktor/utils/io/ByteWriteChannel;
 	public fun getJob ()Lkotlinx/coroutines/Job;
 }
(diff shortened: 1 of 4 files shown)


ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
```diff
@@ -303,7 +303,17 @@ public class ReaderScope(
...
     public override val job: Job
-) : ChannelJob
+) : ChannelJob {
+    /**
+     * To avoid the risk of concurrent write operations, we cancel the nested job before
+     * performing `flushAndClose` on the [ByteWriteChannel].
+     */
+    public suspend fun flushAndClose() {
+        job.cancelChildren()
+        job.children.toList().joinAll()
```
osipxd, line 313: Minor. We could avoid an extra list creation.
```suggestion
        job.children.forEach { it.join() }
```

osipxd (approved):

ktor-io/api/ktor-io.api
```diff
@@ -295,6 +295,7 @@ public abstract interface annotation class io/ktor/utils/io/PublicAPICandidate :
 }
 
 public final class io/ktor/utils/io/ReaderJob : io/ktor/utils/io/ChannelJob {
+	public final fun flushAndClose (Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
```
e5l, line 298: please change the target branch to 3.2.0 or at least mark it with an @InternalAPI

e5l (approved):

PR #4686, opened 2025-02-19
KTOR-8210 Fix copy() and Source multipart processing

bjhham:
**Subsystem**
Core, I/O

**Motivation**
[KTOR-8210](https://youtrack.jetbrains.com/issue/KTOR-8210) formData: implementation of copying Source is broken

The issue in the code was identified here https://github.com/Kotlin/kotlinx-io/issues/433 by @JakeWharton 

**Solution**
Use `Source.peek()` for copying; avoid use of `remaining` in places where `exhausted()` should be used.


e5l:
please refer to the origin issue in kotlinx-io

ktor-io/common/src/io/ktor/utils/io/core/ByteReadPacket.kt
```diff
@@ -49,8 +49,12 @@ public fun Source.readAvailable(out: kotlinx.io.Buffer): Int {
     return result.toInt()
 }
 
+@Deprecated(
+    "Use peek() instead",
+    ReplaceWith("peek()", "kotlinx.io.Source")
+)
 @OptIn(InternalIoApi::class)
-public fun Source.copy(): Source = buffer.copy()
+public fun Source.copy(): Source = buffer.peek()
```
e5l, line 57: should we just use peek directly?

e5l (approved):

e5l (changes requested):
please check the CI

e5l (approved):
looks nice!

bjhham:
Looks like one of the tests I added recently is flaky and it reproduces an issue in the CIO client that I'm also looking into 👀

bjhham:
Actually, it seems to be the expected behaviour when the timeout is triggered in the client.  I just need to increase the timeout because it runs a bit slow on TC.

ktor-io/common/src/io/ktor/utils/io/core/ByteReadPacket.kt
```diff
@@ -49,8 +49,15 @@ public fun Source.readAvailable(out: kotlinx.io.Buffer): Int {
...
+/**
+ * Returns a copy of the current buffer attached to this Source.
+ */
+@Deprecated(
+    "Use peek() or buffer.copy() instead, depending on your use case.",
+    ReplaceWith("peek()", "kotlinx.io.Source")
+)
 @OptIn(InternalIoApi::class)
-public fun Source.copy(): Source = buffer.copy()
+public fun Source.copy(): Source = peek()
```
Thomas-Vos, line 60: @bjhham This change caused some issues in my project. After a `Source.copy()` call, then closing the original Source, the copied source is now no longer valid. It was valid before. I don't think it makes sense to call a function and its description 'copy' if it does not actually copy. This is very unclear.

e5l: Maybe we need deprecate this with warnings with message to add more clarity

JakeWharton: The original implementation did not copy either, though, which was the whole motivation for the change.

A true copy would need to be implemented like:
```kotlin
return Buffer().write(peek().readByteString())
```
This fully buffers the underlying `Source` into its own `Buffer`, makes a `ByteString` which may share the backing segments, creates a new `Buffer`, and then may move those shared segments into the new `Buffer`.

Of course, the "may"s there might also result in massive byte array allocations and actual data copies, but they also may not.

Thomas-Vos: Thanks for the clarification, but still it is a behaviour change. In the past this Ktor copy function worked fine for me (it copied), before the kotlinx-io migration, but even after the migration it kept working fine in my code. Now that it was internally changed to `peek`, code broke throughout my project. Of course I just replaced the usages in my project, but still not that great for a `public` Ktor function that suddenly breaks with such exception.

```kt
import io.ktor.utils.io.core.buildPacket
import io.ktor.utils.io.core.discard
import kotlinx.io.InternalIoApi
import kotlinx.io.readByteArray
import kotlinx.io.writeString
import kotlin.test.Test

@OptIn(InternalIoApi::class)
class CopyTest {

    @Test
    fun test() {
        val packet = buildPacket { writeString("123") }
        val copy = packet.buffer.copy() // Or Ktor function: packet.copy()
        packet.discard()
        copy.readByteArray()
    }

    @Test
    fun test2() {
        val packet = buildPacket { writeString("123") }
        val copy = packet.peek() // Or Ktor function: packet.copy()
        packet.discard()
        copy.readByteArray()
    }
}
```


Screenshot 2025-02-27 at 20 36 27

PR #4687, opened 2025-02-19
KTOR-8206 Apply socket and connect timeouts to WS and SSE

osipxd:
**Subsystem**
Client

**Motivation**
[KTOR-8206](https://youtrack.jetbrains.com/issue/KTOR-8206) WebSocket and SSE don't respect connection timeout set in the HttpTimeout plugin

**Solution**
Previously all timeout settings were ignored for WS and SSE requests. I've changed this behavior to ignore only request timeout.


marychatte (approved):
lgtm

PR #4688, opened 2025-02-19
KTOR-8208 Java: Make webSocket field nullable

osipxd:
**Subsystem**
Client

**Motivation**
[KTOR-8208](https://youtrack.jetbrains.com/issue/KTOR-8208) Websockets: lateinit property websocket is not initialized when connection timeouts

**Solution**
Make `webSocket` nullable instead of `lateinit`. Also add names to all coroutines launched in the `JavaHttpWebSocket`


Mr3zee:
@osipxd ping

PR #4691, opened 2025-02-21
KTOR-8183 Add a simplified route string

adriandieter:
**Subsystem**
Server, server-core & metrics-micrometer

**Motivation**
[KTOR-8183](https://youtrack.jetbrains.com/issue/KTOR-8183) Add a simplified route string

**Solution**
Add `path` extension property to RoutingNode. It returns a String representation of the path matched by the route. RoutingNodes not related to the path, i.e. from the authorization plugin or the http method are ommited here.

Make the route label string of micrometer metrics configurable with the `transformRoute` function in `MicrometerMetricsConfig` and use `RoutingNode.path` as default.

This fixes the issue of unwanted path elements, like `/(authenticate "default")/uri`, in the metrics route label.
The original behaviour can be configured by setting
```kotlin
transformRoute { 
    parent.toString() 
}
```
in the `MicrometerMetricsConfig`.

### DIFF
--- a/ktor-server/ktor-server-core/common/src/io/ktor/server/routing/RoutingNode.kt
+++ b/ktor-server/ktor-server-core/common/src/io/ktor/server/routing/RoutingNode.kt
@@ -369,7 +369,6 @@ private fun RoutingNode.path(): String {
 }
 
 private fun RouteSelector.toPathElement(): String = when (this) {
-
     is PathSegmentConstantRouteSelector,
     is PathSegmentParameterRouteSelector,
     is PathSegmentOptionalParameterRouteSelector,
--- a/ktor-server/ktor-server-core/common/src/io/ktor/server/routing/RoutingNode.kt
+++ b/ktor-server/ktor-server-core/common/src/io/ktor/server/routing/RoutingNode.kt
@@ -352,7 +352,7 @@ private fun RoutingNode.getAllRoutes(endpoints: MutableList<RoutingNode>) {
 /**
  * String representation of the path matched by this route.
  *
- * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.server.routing.path)
+ * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.server.routing.RoutingNode.path)
  */
 public val RoutingNode.path: String
     get() = path()
--- a/ktor-server/ktor-server-core/common/src/io/ktor/server/routing/RoutingNode.kt
+++ b/ktor-server/ktor-server-core/common/src/io/ktor/server/routing/RoutingNode.kt
@@ -12,7 +12,6 @@ import io.ktor.util.*
 import io.ktor.util.pipeline.*
 import io.ktor.util.reflect.*
 import io.ktor.utils.io.*
-import kotlinx.coroutines.*
 import kotlin.coroutines.*
 
 /**
@@ -348,6 +347,39 @@ private fun RoutingNode.getAllRoutes(endpoints: MutableList<RoutingNode>) {
         endpoints.add(this)
     }
     children.forEach { it.getAllRoutes(endpoints) }
+}
+
+/**
+ * String representation of the path matched by this route.
+ *
+ * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.server.routing.path)
+ */
+public val RoutingNode.path: String
+    get() = path()
+
+private fun RoutingNode.path(): String {
+    val parentPath = parent?.path()
+    val selectorElement = selector.toPathElement()
+    return when {
+        parentPath == null -> selectorElement
+        selectorElement.isEmpty() -> parentPath
+        parentPath.endsWith('/') || selectorElement.startsWith('/') -> "$parentPath$selectorElement"
+        else -> "$parentPath/$selectorElement"
+    }
+}
+
+private fun RouteSelector.toPathElement(): String = when (this) {
+
+    is PathSegmentConstantRouteSelector,
+    is PathSegmentParameterRouteSelector,
+    is PathSegmentOptionalParameterRouteSelector,
+    is PathSegmentTailcardRouteSelector,
+    is PathSegmentWildcardRouteSelector,
+    is PathSegmentRegexRouteSelector -> toString()
+
+    is TrailingSlashRouteSelector -> "/"
+
+    else -> ""
 }
 
 @Deprecated("Please use route scoped plugins instead")
--- a/ktor-server/ktor-server-core/common/test/io/ktor/server/routing/RouteTest.kt
+++ b/ktor-server/ktor-server-core/common/test/io/ktor/server/routing/RouteTest.kt
@@ -155,7 +155,7 @@ class RouteTest {
                 // Routing nodes not related to path
                 route("omitted") {
                     contentType(ContentType.Text.CSV) {
-                        post("contentType"){}
+                        post("contentType") {}
                     }
                     param("order", "asc") {
                         post("param") {}
--- a/ktor-server/ktor-server-core/common/test/io/ktor/server/routing/RouteTest.kt
+++ b/ktor-server/ktor-server-core/common/test/io/ktor/server/routing/RouteTest.kt
@@ -138,4 +138,36 @@ class RouteTest {
 
         assertEquals(HttpStatusCode.OK, client.get("/").status)
     }
+
+    @Test
+    fun testPathProperty() = testApplication {
+        application {
+            val root = routing {
+                get {}
+                get("/") {}
+                get("/trailing/slash/") {}
+                get("/parameter/{mandatory}/{optional?}") {}
+                get("/wildcard/*") {}
+                get("/tailcard/{...}") {}
+                get("/parameter/tailcard/{path...}") {}
+                get(Regex("/.+regex")) {}
+            }
+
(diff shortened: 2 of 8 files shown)


ktor-server/ktor-server-core/common/src/io/ktor/server/routing/RoutingNode.kt
```diff
@@ -350,6 +349,39 @@ private fun RoutingNode.getAllRoutes(endpoints: MutableList<RoutingNode>) {
     children.forEach { it.getAllRoutes(endpoints) }
 }
 
+/**
+ * String representation of the path matched by this route.
+ *
+ * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.server.routing.path)
```
e5l, line 355: io.ktor.server.routing.RoutingNode.path

ktor-server/ktor-server-core/common/src/io/ktor/server/routing/RoutingNode.kt
```diff
@@ -350,6 +349,39 @@ private fun RoutingNode.getAllRoutes(endpoints: MutableList<RoutingNode>) {
...
+ */
+public val RoutingNode.path: String
+    get() = path()
```
e5l, line 358: The main motivation for this issue was the misleading formatting in the log. Should we go directly with the `toString` change?

e5l (changes requested):

e5l:
Hey @adriandieter, thanks you for the PR. Nice start! Some minor things need to be fixed, and maybe we can avoid making a new API so it can go to the patch release. Could you please take a look?

ktor-server/ktor-server-core/common/src/io/ktor/server/routing/RoutingNode.kt
adriandieter: I figured updating the toString could lead to a breaking change, for example in existing monitoring setups. I thought that was why [Vladislav Yaroshchuk](https://youtrack.jetbrains.com/users/vladislav.yaroshchuk) proposed a new property.

If that is not an issue I do think that putting this into main would also work out. That way, no additional configuration is required in the [micrometer plugin](https://github.com/ktorio/ktor/blob/747fef9366c02e0112c2130dcb95eaabb5fef7ee/ktor-server/ktor-server-plugins/ktor-server-metrics-micrometer/jvm/src/io/ktor/server/metrics/micrometer/MicrometerMetrics.kt#L213).

e5l: We can add system property to enable old behavior if this is a potential problem

adriandieter: Hey @e5l, thanks for the feedback!

On second thought, I think `RoutingNode.toString()` should at least also include the `HttpMethodRouteSelector`, which it currently does. 
The use case of having just the path of the route for metrics seems different. 
Having just the path in the `toString()`, would result in the RoutingNodes for 

- "/shop/order/shipment/(method:GET)",
- "/shop/order/shipment/(method:PUT)",
- "/shop/order/shipment/(method:POST)",

all having "/shop/order/shipment/" as value of toString().

However not having the method in the string used in metrics makes sense, since there the method is included in a separate field, like in the [example by ](https://youtrack.jetbrains.com/issue/KTOR-8183/Add-a-simplified-route-string#focus=Comments-27-11518622.0-0) @shchuko.

I think adding a separate path property or function makes sense here.
However, we could move it to the `ktor-server-plugins:ktor-server-metrics` module to keep it out of the `ktor-server-core` API.

e5l: > I think adding a separate path property or function makes sense here.

Yep, you're right! It's a good idea.

Please let me know when you need a review

adriandieter:
One point I was not sure about, is if we should set the new path property as default value for the metrics route label or keep the original behavior as default.

ktor-server/ktor-server-plugins/ktor-server-metrics-micrometer/jvm/src/io/ktor/server/metrics/micrometer/MicrometerMetrics.kt
```diff
@@ -194,7 +222,9 @@ public val MicrometerMetrics: ApplicationPlugin<MicrometerMetricsConfig> =
 
         application.monitor.subscribe(RoutingRoot.RoutingCallStarted) { call ->
             call.attributes.getOrNull(measureKey)?.let { measure ->
-                measure.route = call.route.parent.toString()
+                measure.route = call.route.let { route ->
+                    pluginConfig.transformRoute(route)
+                }
```
osipxd, line 227: Minor. Let's drop `.let { ... }` here
```suggestion
                measure.route = pluginConfig.transformRoute(call.route)
```

ktor-server/ktor-server-plugins/ktor-server-metrics-micrometer/jvm/src/io/ktor/server/metrics/micrometer/MicrometerMetrics.kt
```diff
@@ -119,6 +119,34 @@ public class MicrometerMetricsConfig {
...
+     * Remove a prefix from the path:
+     * ```kotlin
+     * transformRoute {
+     *     path.removePrefix("/path/prefix")
+     * }
+     * ```
+     *
+     * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.server.metrics.micrometer.MicrometerMetricsConfig.transformRoute)
+     */
+    public fun transformRoute(block : RoutingNode.() -> String) {
```
osipxd, line 147: I'd make `RoutingNode` a parameter as it is a common pattern for transforming functions.
```suggestion
    public fun transformRoute(block: (RoutingNode) -> String) {
```

ktor-server/ktor-server-plugins/ktor-server-metrics-micrometer/jvm/src/io/ktor/server/metrics/micrometer/MicrometerMetrics.kt
```diff
@@ -119,6 +119,34 @@ public class MicrometerMetricsConfig {
     public fun timers(block: Timer.Builder.(ApplicationCall, Throwable?) -> Unit) {
         timerBuilder = block
     }
+
+    internal var transformRoute: RoutingNode.() -> String = { path }
```
osipxd, line 123: Note: We should mark this issue as "impactful change" as it requires migration for those who relies on the old behavior

osipxd (commented):
@adriandieter, thank you for the PR! I have a couple of comments.
Please, cherry-pick the changes from this PR on top of `3.2.0-eap` branch. It is required as this PR introduces new APIs. After that please fix the [issues found by linter](https://ktor.teamcity.com/viewLog.html?buildId=323083&tab=Inspection&buildTypeId=Ktor_KtorCodeStyleVerifyKtLint) and update API dump using the following command:
```bash
./gradlew :ktor-server:ktor-server-core:apiDump :ktor-server:ktor-server-plugins:ktor-server-metrics-micrometer:apiDump
```

ktor-server/ktor-server-core/common/test/io/ktor/server/routing/RouteTest.kt
```diff
@@ -138,4 +138,36 @@ class RouteTest {
...
+        application {
+            val root = routing {
+                get {}
+                get("/") {}
+                get("/trailing/slash/") {}
+                get("/parameter/{mandatory}/{optional?}") {}
+                get("/wildcard/*") {}
+                get("/tailcard/{...}") {}
+                get("/parameter/tailcard/{path...}") {}
+                get(Regex("/.+regex")) {}
```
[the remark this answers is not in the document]
adriandieter: Added cases for contentType, param and header and HttpMethod is implicit in every test case.

adriandieter:
Thanks for the quick review @osipxd! Addressed all comments.

e5l (commented):
@osipxd, could you check?

osipxd (approved):
@adriandieter, thank you! Looks good to me 👍

ktor-server/ktor-server-core/common/src/io/ktor/server/routing/RoutingNode.kt
```diff
@@ -350,6 +349,38 @@ private fun RoutingNode.getAllRoutes(endpoints: MutableList<RoutingNode>) {
...
+    }
+}
+
+private fun RouteSelector.toPathElement(): String = when (this) {
+    is PathSegmentConstantRouteSelector,
+    is PathSegmentParameterRouteSelector,
+    is PathSegmentOptionalParameterRouteSelector,
+    is PathSegmentTailcardRouteSelector,
+    is PathSegmentWildcardRouteSelector,
+    is PathSegmentRegexRouteSelector -> toString()
```
osipxd, line 377: We could introduce a supertype `PathSegmentRouteSelector` for these selectors to make it possible to create custom route selectors appearing in path.
@bjhham, what do you think?

ChrisLahaye:
Could this be included as a patch in 3.1.3?

Issue #4692, opened 2025-02-21
[Bug] Cancel flow of SSESession.incoming will not disconnect connection in OkHttpSSESession implementation


hehua2008:
Look at these source code.
In the OkHttpSSESession implementation of SSESession, cancel incoming flow will **not** trigger serverSentEventsSource.cancel(), which will keep SSE still connecting !

```kotlin
internal class OkHttpSSESession(
    engine: OkHttpClient,
    engineRequest: Request,
    override val coroutineContext: CoroutineContext,
) : SSESession, EventSourceListener() {
    private val serverSentEventsSource = EventSources.createFactory(engine).newEventSource(engineRequest, this)

    ...

    private val _incoming = Channel<ServerSentEvent>(8)

    override val incoming: Flow<ServerSentEvent>
        get() = _incoming.receiveAsFlow()

    ...

    override fun onEvent(eventSource: EventSource, id: String?, type: String?, data: String) {
        _incoming.trySendBlocking(ServerSentEvent(data, type, id))
    }

    override fun onFailure(eventSource: EventSource, t: Throwable?, response: Response?) {
        ...

        _incoming.close()
        serverSentEventsSource.cancel()
    }

    override fun onClosed(eventSource: EventSource) {
        _incoming.close()
        serverSentEventsSource.cancel()
    }
}
```

But in the DefaultClientSSESession implementation of SSESession, cancel incoming flow will trigger coroutineContext.cancel() and input.cancel(), which will disconnect SSE connection !
```kotlin
@OptIn(InternalAPI::class)
@Deprecated("It should be marked with `@InternalAPI`, please use `ClientSSESession` instead")
public class DefaultClientSSESession(
    content: SSEClientContent,
    private var input: ByteReadChannel,
    override val coroutineContext: CoroutineContext
) : SSESession {
    ...

    public constructor(
        content: SSEClientContent,
        input: ByteReadChannel
    ) : this(content, input, content.callContext + Job() + CoroutineName("DefaultClientSSESession"))

    private var _incoming = flow {
        // inner while for parsing events of current input (=connection), and when the current input is closed,
        // we have an outer while to obtain new input
        while (this@DefaultClientSSESession.coroutineContext.isActive) {
            while (this@DefaultClientSSESession.coroutineContext.isActive) {
                val event = input.parseEvent() ?: break

                if (event.isCommentsEvent() && !showCommentEvents) continue
                if (event.isRetryEvent() && !showRetryEvents) continue

                emit(event)
[...]

PR #4694, opened 2025-02-21
Fixed an issue where canceling the incoming flow of OkHttpSSESession/DefaultClientSSESession failed to disconnect the SSE session.

hehua2008:
Fixed an issue where canceling the incoming flow of OkHttpSSESession failed to disconnect the SSE session.

**Subsystem**
Client
ktor-client/ktor-client-okhttp

**Motivation**
Refer to a bug/ticket #4692.
In the OkHttpSSESession implementation of SSESession, cancel incoming flow will NOT trigger serverSentEventsSource.cancel(), which will keep SSE still connecting !

**Solution**
- First, ```val incoming: Flow<ServerSentEvent>``` in  OkHttpSSESession should always return the same Flow instance, just as DefaultClientSSESession did.

- Second, cancel ```_incoming.receiveAsFlow()``` will just cancel the Flow but will **NOT** cancel the _incoming Channel, so we should use ```_incoming.consumeAsFlow()``` instead.

- Third, if ```incoming.trySendBlocking(ServerSentEvent(data, type, id))``` fails with a CancellationException, we should rethrow the CancellationException to let ```fun onFailure(eventSource: EventSource, t: Throwable?, response: Response?)``` handle it and close the SSE session.

### DIFF
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/sse/DefaultClientSSESession.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/sse/DefaultClientSSESession.kt
@@ -14,6 +14,7 @@ import kotlinx.coroutines.*
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.catch
 import kotlinx.coroutines.flow.flow
+import kotlinx.coroutines.flow.onCompletion
 import kotlin.coroutines.CoroutineContext
 
 @OptIn(InternalAPI::class)
@@ -61,7 +62,7 @@ public class DefaultClientSSESession(
     }.catch { cause ->
         when (cause) {
             is CancellationException -> {
-                close()
+                // CancellationException will be handled by onCompletion operator
             }
 
             else -> {
@@ -69,6 +70,12 @@ public class DefaultClientSSESession(
                 close()
                 throw cause
             }
+        }
+    }.onCompletion { cause ->
+        // Because catch operator only catch throwable occurs in upstream flow, so we use onCompletion operator instead
+        // to handle CancellationException occurs in either upstream flow or downstream flow.
+        if (cause is CancellationException) {
+            close()
         }
     }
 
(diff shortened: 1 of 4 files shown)


e5l:
Hey @hehua2008, lgtm. Could you add a test, so we can merge? :)

hehua2008:
For same bug in DefaultClientSSESession:
Because catch operator only catch throwable occurs in upstream flow, so we use onCompletion operator instead to handle CancellationException occurs in either upstream flow or downstream flow.

```kotlin
    }.catch { cause ->
        when (cause) {
            is CancellationException -> {
                // CancellationException will be handled by onCompletion operator
            }

            else -> {
                LOGGER.trace { "Error during SSE session processing: $cause" }
                close()
                throw cause
            }
        }
    }.onCompletion { cause ->
        // Because catch operator only catch throwable occurs in upstream flow, so we use onCompletion operator instead
        // to handle CancellationException occurs in either upstream flow or downstream flow.
        if (cause is CancellationException) {
            close()
        }
    }
```

hehua2008:
> Hey @hehua2008, lgtm. Could you add a test, so we can merge? :)

I have added testCancelSseSession in ServerSentEventsTest for ktor-client:
```kotlin
    @Test
    fun testCancelSseSession() = clientTests {
        config {
            install(SSE)
        }

        test { client ->
            coroutineScope {
                val job = launch {
                    val session = client.serverSentEventsSession("$TEST_SERVER/sse/hello?times=20&interval=100")
                    try {
                        session.incoming.collect()
                    } finally {
                        withContext(NonCancellable) {
                            delay(500)
                            assertFalse(session.isActive)
                        }
                    }
                }
                delay(500)
                job.cancelAndJoin()
            }
        }
    }
```

e5l (approved):

e5l (approved):
lgtm

osipxd:
Hey @hehua2008, thank you for the PR. Could you please fix import ordering in `DefaultClientSSESession`?

hehua2008:
> Hey @hehua2008, thank you for the PR! Could you please fix import ordering in `DefaultClientSSESession`?

Sorry for that, I have fixed it.

Savrov:
When we can expect it to be released?

Stexxe:
In the next patch release of Ktor 3.1.2.

Savrov:
I would like to know some est. date, cause it's a critical defect in our app and we can not release it without that fix.

Thank you

PR #4696, opened 2025-02-21
KTOR-8218 Write request and response logs as single messages

Stexxe:
https://youtrack.jetbrains.com/issue/KTOR-8218

### DIFF
--- a/ktor-client/ktor-client-plugins/ktor-client-logging/common/src/io/ktor/client/plugins/logging/Logging.kt
+++ b/ktor-client/ktor-client-plugins/ktor-client-logging/common/src/io/ktor/client/plugins/logging/Logging.kt
@@ -191,6 +191,7 @@ public val Logging: ClientPlugin<LoggingConfig> = createClientPlugin("Logging", 
         contentLength: Long?,
         headers: Headers,
         method: HttpMethod,
+        logLines: MutableList<String>,
         body: ByteReadChannel
     ) {
         val (isBinary, size, newBody) = detectIfBinary(body, contentLength, content.contentType, headers)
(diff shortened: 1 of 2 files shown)


ktor-client/ktor-client-plugins/ktor-client-logging/common/src/io/ktor/client/plugins/logging/Logging.kt
```diff
@@ -191,6 +191,7 @@ public val Logging: ClientPlugin<LoggingConfig> = createClientPlugin("Logging",
         contentLength: Long?,
         headers: Headers,
         method: HttpMethod,
+        logLines: MutableList<String>,
```
e5l, line 194: I would pass `log: (String) -> Unit` instead, cause you're not using any list functions except add

e5l: We also can make this function top-level internal and have some unit tests

e5l (approved):
lgtm, please check the comments

ktor-client/ktor-client-plugins/ktor-client-logging/common/src/io/ktor/client/plugins/logging/Logging.kt
Stexxe: > I would pass `log: (String) -> Unit` instead, cause you're not using any list functions except add

What's the point of that?

> We also can make this function top-level internal and have some unit tests

What this function would do? How would we test it?

e5l: > What's the point of that?
You don't allow future modification to drop the content of the list

> What this function would do? How would we test it?
All it is doing atm. I just thought we might replace some flaky integration tests with unit tests on this function

Stexxe: > You don't allow future modification to drop the content of the list

I don't see a problem in that if the decision is thoughtful.

> All it is doing atm. I just thought we might replace some flaky integration tests with unit tests on this function.

There would be a race condition If the client made multiple simultaneous requests while using the same buffer for log lines.

Stexxe: Also, my integration tests aren't flaky because I use the `testApplication`.

PR #4697, opened 2025-02-23
KTOR-7108: add more ContentType values

anton-erofeev:
**Subsystem**
ktor-http

**Motivation**
https://youtrack.jetbrains.com/issue/KTOR-7108

**Solution**
Added more common ContentType values from https://www.iana.org/assignments/media-types/media-types.xhtml


e5l (changes requested):
Hey @anton-erofeev, lgtm. Could you please run API dump and `formatKotlin` before we can merge? It also would be nice to set the target branch 3.2.0-eap and rebase.

PR #4715, opened 2025-02-28
KTOR-7298: Load modules dynamically only if development mode is enabled

osipxd:
**Subsystem**
Server

**Motivation**
[KTOR-7298](https://youtrack.jetbrains.com/issue/KTOR-7298) Android: "Array has more than one element" error when starting a server with release build

**Solution**
There were several problems with modules loading when code was optimized by R8 or ProGuard:
1. Extension `Function<*>.methodName()` could throw `KotlinReflectionInternalError` for obfuscated methods
2. `clazz.declaredConstructors.single()` in `executeModuleFunction` could throw `IllegalArgumentException` because the class could has either zero or more than one constructor (needs to be investigated further)
3. Dynamic loading of programmatic modules seems redundant for the case when development mode is disabled

Fixed all of these issues in separate commits so it's better to review commit-by-commit.

### DIFF
--- a/ktor-server/ktor-server-core/jvm/src/io/ktor/server/engine/EmbeddedServerJvm.kt
+++ b/ktor-server/ktor-server-core/jvm/src/io/ktor/server/engine/EmbeddedServerJvm.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2024 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ * Copyright 2014-2025 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
  */
 
 package io.ktor.server.engine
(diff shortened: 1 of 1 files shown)


ktor-server/ktor-server-core/jvm/src/io/ktor/server/engine/EmbeddedServerJvm.kt
```diff
@@ -356,23 +368,45 @@ actual constructor(
...
+        val name = runCatching { module.methodName() }
+            .onFailure { environment.log.debug("Module can't be loaded dynamically", it) }
+            .getOrElse { return null }
+
+        return { classLoader ->
+            val application = this
+            try {
+                launchModuleByName(name, classLoader, application)
+            } catch (cause: ReloadingException) {
+                environment.log.debug("Failed to load module '$name' dynamically, falling back to static loading", cause)
```
osipxd, line 401: Not sure about "static"/"dynamic" here. Is it clear what I mean?

bjhham: Perhaps "Failed to load module by classpath reference, falling back to currently loaded value"?

PR #4733, opened 2025-03-10
KTOR-8256 Ktor headers map matches only on hashCode

zibet27:
**Subsystem**
CIO Client/Server

**Motivation**
[KTOR-8256](https://youtrack.jetbrains.com/issue/KTOR-8256) Ktor headers map matches only on hashCode
[KTOR-7824](https://youtrack.jetbrains.com/issue/KTOR-7824) Ktor doesn't parse multiple headers

**Solution**
HttpHeadersMaps were implemented as a real hash multi map using Open addressing and Linear Probing.

- get method now returns all values with a given header name joined by commas.
- possibly "infinite" number of headers could be added.

HttpParser was changed to parse headers with multiple values separated by comma as described in [RFC](https://datatracker.ietf.org/doc/html/rfc7230#section-3.2.2).


ktor-utils/common/src/io/ktor/util/StringValues.kt
```diff
@@ -40,11 +40,11 @@ public interface StringValues {
     public val caseInsensitiveName: Boolean
 
     /**
-     * Gets first value from the list of values associated with a [name], or null if the name is not present
+     * Gets all values associated with a [name] of the header separated by comma, or null if the name is not present
      *
      * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.util.StringValues.get)
      */
-    public operator fun get(name: String): String? = getAll(name)?.firstOrNull()
```
bjhham, line 47: I don't think we'll want to update the default behavior on this interface - iirc it's used for query params and other similar maps.  You can just override in the `CIOHeaders` class.

ktor-http/ktor-http-cio/api/ktor-http-cio.api
```diff
@@ -72,7 +84,7 @@ public final class io/ktor/http/cio/HttpHeadersMap {
 
 public abstract class io/ktor/http/cio/HttpMessage : java/io/Closeable {
 	public fun close ()V
-	public final fun getHeaders ()Lio/ktor/http/cio/HttpHeadersMap;
```
bjhham, line 75: This one is a bit of a problem for backwards compatibility...  Looks like it might be best to go back to replacing `HttpHeadersMap` with the new implementation, then re-introducing the public functions with something like `nameAt(idx: Int) = nameAtOffset(offsets().toList()[idx])`.  It's unlikely anyone is relying on this class directly, so performance is less of a concern, but breaking the API in a patch would send us to library maintainer jail.

bjhham (changes requested):

ktor-http/ktor-http-cio/api/ktor-http-cio.api
zibet27: Ok, I will try it. But allocating 'offsets.toList()' on every call definitely would not help to pass allocation tests). Maybe it it makes sense to remember those offsets.

ktor-http/ktor-http-cio/api/ktor-http-cio.api
```diff
@@ -57,17 +57,33 @@ public final class io/ktor/http/cio/HttpBodyKt {
...
+	public final fun getAll (Ljava/lang/String;)Lkotlin/sequences/Sequence;
+	public final fun getValuesCount ()I
+	public final fun nameAtOffset (I)Ljava/lang/CharSequence;
+	public final fun offsets ()Lkotlin/sequences/Sequence;
+	public final fun put (IIII)V
+	public final fun release ()V
+	public fun toString ()Ljava/lang/String;
+	public final fun valueAtOffset (I)Ljava/lang/CharSequence;
+}
+
```
bjhham, line 71: Looks like this needs to be removed from the dump.

ktor-http/ktor-http-cio/common/src/io/ktor/http/cio/HttpHeadersMap.kt
```diff
@@ -6,36 +6,89 @@ package io.ktor.http.cio
...
+            }
+            nextIndex++
+            offset = (offset / HEADER_SIZE) % headerCapacity
+        }
+        return -1
+    }
+
+    public operator fun get(name: String): CharSequence? {
+        return getAll(name).firstOrNull()
+    }
```
bjhham, line 68: I think the sequence here may be the culprit in the allocations increase.  Try inline the logic for retrieving the first element here.

bjhham (approved):
One last item in the dump there + the allocations, other than that it lgtm 🎉

ktor-http/ktor-http-cio/common/src/io/ktor/http/cio/HttpHeadersMap.kt
```diff
@@ -6,36 +6,101 @@ package io.ktor.http.cio
...
+                headerIndex = nextHeaderIndex
+                continue
+            }
+            headerIndex = (headerIndex + 1) % headerCapacity
+        }
+    }
+
+    public fun offsets(): Sequence<Int> = headersData.headersStarts()
+
+    @Deprecated("Use put instead", ReplaceWith("put"))
```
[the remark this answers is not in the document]
zibet27: Sure. Now, it makes more sense.

PR #4734, opened 2025-03-11
KTOR-8222 Move publishing to build-logic

osipxd:
**Subsystem**
Infrastructure

**Motivation**
[KTOR-8222](https://youtrack.jetbrains.com/issue/KTOR-8222) Move publishing logic to precompiled script plugin

**Solution**
Finally, the last piece of build logic moved to `build-logic` module 🎉 

In order to test migration was smooth, I've added task `validatePublishedArtifacts` comparing list of published artifacts with the list stored in `gradle/artifacts.txt`. Running this task before and after migration gave me the same result.

Changes:
- Removed property `publishLocal` and added `mavenLocal()` to the list of publishing repositories.
- Removed property `relocatedArtifacts`. We could add it later if necessary
- Removed `verifier.gradle`. These checks will be added back as part of [KTOR-7739](https://youtrack.jetbrains.com/issue/KTOR-7739) "Validate a set of published artifacts during release"


osipxd:
@coderabbitai review

build-logic/src/main/kotlin/ktorbuild/internal/publish/ValidatePublishedArtifactsTask.kt
```diff
@@ -0,0 +1,134 @@
...
+ * consistency between the expected and the actual artifacts found in the [repositoryDirectory].
+ *
+ * By default, `artifactsDump` is located at `<rootProject>/gradle/artifacts.txt` and [TestRepository] location is used
+ * as a `repositoryDirectory`.
+ *
+ * Initially cherry-picked from kotlinx.coroutines build scripts
+ * See: https://github.com/Kotlin/kotlinx.serialization/blob/v1.8.0/buildSrc/src/main/kotlin/publishing-check-conventions.gradle.kts
+ */
+@Suppress("LeakingThis")
+internal abstract class ValidatePublishedArtifactsTask : DefaultTask() {
```
bjhham, line 36: Will this execute for the CI checks on PRs?  This seems important to catch new modules so that publication doesn't fail when we try to release.

osipxd: Unfortunately, not. But this will be done in [KTOR-7739](https://youtrack.jetbrains.com/issue/KTOR-7739) "Validate a set of published artifacts during release".

I can't include this task to release pipeline "as is" because it runs all publishing tasks at once while we publish different groups separately.

bjhham: Ah ok, thanks!

bjhham (approved):

PR #4736, opened 2025-03-14
KTOR-8304 File configuration for dependencies

bjhham:
**Subsystem**
Server, DI

**Motivation**
[KTOR-8304](https://youtrack.jetbrains.com/issue/KTOR-8304) File configuration for dependencies

**Solution**
This introduces reading of a string list configuration key `ktor.application.dependencies` for referencing classes or functions that will be injected into the DI container.

You can now include some configuration like:

```yaml
ktor:
  application:
    dependencies:
      - io.ktor.chat.PostgresKt.createDatabase
      - io.ktor.chat.MessagesRepository
```

Where the references can be either classes or function references.

The parameters for either class constructors or function references will be supplied from the DI resolver context, and the references will be constructed lazily when the dependencies plugin is first invoked.  This way, we can install the DI plugin at our leisure, and even mix providing dependencies from our file config and through the application module API.

See `DependencyInjectionJvmTest.kt` and `DependencyInjectTestModules.kt` for examples.


ktor-server/ktor-server-plugins/ktor-server-di/jvm/src/io/ktor/server/plugins/di/utils/ClasspathReference.jvm.kt
```diff
@@ -0,0 +1,76 @@
...
+        registry.set(DependencyKey(returnTypeInfo)) {
+            var lastError: Throwable? = null
+            for (function in functionCandidates) {
+                try {
+                    return@set function.callBy(
+                        reflection.mapParameters(function.parameters) { param ->
+                            when (param.type) {
+                                // special types, from application
+                                DependencyResolver::class.starProjectedType -> this@set
+                                ApplicationEnvironment::class.starProjectedType -> application.environment
```
bjhham, line 52: There was some interesting logic in `CallableUtils.kt` that I didn't replicate which fails when the classloader doesn't support classpath references.   I figured we can assume that the application will fail from module classpath references in that case.

bjhham:
@coderabbitai review

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyInjection.kt
```diff
@@ -20,13 +22,23 @@ import kotlin.reflect.KClass
  */
 public val DI: ApplicationPlugin<DependencyInjectionConfig> =
```
e5l, line 23: please update the KDoc showing examples with usages and configuration samples

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyProvider.kt
```diff
@@ -90,32 +94,20 @@ public class ImplicitCreateFunction(public val origin: ExplicitCreateFunction) :
  *
  * @throws AmbiguousDependencyException Always thrown when attempting to create a dependency
  * through the [create] method.
+ *
+ * TODO it would be nice to handle delegate properties somehow
```
e5l, line 98: please update KDoc with an issue

ktor-server/ktor-server-plugins/ktor-server-di/nonJvm/src/io/ktor/server/plugins/di/utils/ClasspathReferences.nonJvm.kt
```diff
@@ -0,0 +1,16 @@
...
+
+import io.ktor.server.application.Application
+import io.ktor.server.plugins.di.*
+
+internal actual fun installReference(
+    application: Application,
+    registry: DependencyRegistry,
+    reference: ClasspathReference
+) {
+    TODO("Reflection is unavailable for non-JVM targets for the time being")
```
e5l, line 15: I think we should use `error` and mention related issues for the Kotlin native, js, wasm

ktor-server/ktor-server-plugins/ktor-server-di/jvm/src/io/ktor/server/plugins/di/DependencyReflectionJvm.kt
```diff
@@ -39,14 +40,43 @@ public open class DependencyReflectionJvm : DependencyReflection {
...
+                    parameter.type.isMarkedNullable -> null // let value = null
+                    else -> throw cause
+                }
+            }
+        }.toMap()
+
+    /**
+     * Maps a parameter to a [DependencyKey].
+     *
+     * TODO annotation for named qualifier
```
e5l, line 78: same here

e5l (changes requested):
@vnikolova, could you check KDocs?

ktor-server/ktor-server-plugins/ktor-server-di/jvm/src/io/ktor/server/plugins/di/DependencyReflectionJvm.kt
bjhham: I'll just remove this line since it'll be part of the initial release https://youtrack.jetbrains.com/issue/KTOR-8321/Dependency-injection-named-argument-annotation

ktor-server/ktor-server-plugins/ktor-server-di/nonJvm/src/io/ktor/server/plugins/di/utils/ClasspathReferences.nonJvm.kt
bjhham: There's one for WASM 🤷  https://youtrack.jetbrains.com/issue/KT-63410/Support-full-reflection-on-Kotlin-Wasm

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyInjection.kt
```diff
@@ -17,16 +19,59 @@ import kotlin.reflect.KClass
  * 1. `provider`: the logic for registering new types
  * 2. `resolution`: the function for validating the provided types and creating the dependency map
  * 3. `reflection`: controls how objects are initialized from type references
+ *
+ * Dependencies can be declared via the dependencies DSL:
```
vnikolova, line 23:
```suggestion
 * You can declare dependencies using the dependencies DSL:
```

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyInjection.kt
```diff
@@ -165,3 +210,12 @@ public open class DependencyReflectionDisabledException() :
 public class DependencyAbstractTypeConstructionException(
     qualifiedName: String,
 ) : DependencyConstructionException("Cannot instantiate abstract type: $qualifiedName")
+
+/**
+ * Thrown when we are unable to resolve a static reference from the configuration file.
```
vnikolova, line 215:
```suggestion
 * Thrown when a static reference cannot be resolved from the configuration file.
```

ktor-server/ktor-server-plugins/ktor-server-di/jvm/src/io/ktor/server/plugins/di/DependencyReflectionJvm.kt
```diff
@@ -39,14 +40,41 @@ public open class DependencyReflectionJvm : DependencyReflection {
...
+    /**
+     * List constructors of a class in order of preference.
+     */
     public open fun <T : Any> findConstructors(kClass: KClass<T>): Collection<KFunction<T>> =
         kClass.constructors
 
+    /**
+     * Resolves the list of parameters from the provided resolve function.
+     *
+     * When optional or nullable, we ignore failures to retrieve the values.
```
vnikolova, line 57:
```suggestion
     * When optional or nullable, failures to retrieve the values are ignored.
```

vnikolova (approved):
KDocs look ok to me, just a few minor tweaks.

ktor-server/ktor-server-plugins/ktor-server-di/jvm/src/io/ktor/server/plugins/di/DependencyReflectionJvm.kt
```diff
@@ -39,14 +40,41 @@ public open class DependencyReflectionJvm : DependencyReflection {
...
 
+    /**
+     * Resolves the list of parameters from the provided resolve function.
+     *
+     * When optional or nullable, failures to retrieve the values are ignored.
+     */
+    public open fun mapParameters(
+        parameters: List<KParameter>,
+        resolve: (KParameter) -> Any?
+    ): Map<KParameter, Any?> =
```
e5l, line 62: Could you tell me when we can have a key without value and what it means?

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyProvider.kt
```diff
@@ -90,32 +94,24 @@ public class ImplicitCreateFunction(public val origin: ExplicitCreateFunction) :
  *
  * @throws AmbiguousDependencyException Always thrown when attempting to create a dependency
  * through the [create] method.
+ *
+ * TODO [KTOR-8322 Handle delegate pattern](https://youtrack.jetbrains.com/issue/KTOR-8322/Dependency-injection-handle-delegate-pattern)
```
e5l, line 98: Please write a note about the problem details instead of TODO

ktor-server/ktor-server-plugins/ktor-server-di/jvm/src/io/ktor/server/plugins/di/DependencyReflectionJvm.kt
bjhham: This is just support for nullable parameters - so if a function/constructor has a nullable type we set it to null when unable to resolve.

ktor-server/ktor-server-plugins/ktor-server-di/jvm/src/io/ktor/server/plugins/di/utils/ClasspathReference.jvm.kt
```diff
@@ -0,0 +1,76 @@
...
+import kotlin.reflect.full.starProjectedType
+import kotlin.reflect.jvm.jvmErasure
+import kotlin.reflect.jvm.kotlinFunction
+
+internal actual fun installReference(
+    application: Application,
+    registry: DependencyRegistry,
+    reference: ClasspathReference,
+) {
+    val reflection = registry.reflection as? DependencyReflectionJvm ?: DependencyReflectionJvm()
```
osipxd, line 19: So we don't support custom reflection implementation here? I think user could provide something optimized for theirs use case. Or `NoReflection` we discussed before.

ktor-server/ktor-server-plugins/ktor-server-di/jvm/src/io/ktor/server/plugins/di/utils/ClasspathReference.jvm.kt
```diff
@@ -0,0 +1,76 @@
...
+    val classLoader = Thread.currentThread().contextClassLoader
+    val clazz = classLoader.load(reference)
+
+    if (clazz.name == reference.value) {
```
osipxd, line 23: Minor. There might be a corner case with references like this:

```kotlin
@file:JvmName("Database")

package org.example

// org.example.Database.Database
fun Database(): Database = PostgressDatabase(...)
```
However, I think it's unlikely a common use case

ktor-server/ktor-server-plugins/ktor-server-di/jvm/test/io/ktor/server/plugins/di/DependencyInjectionJvmTest.kt
```diff
@@ -131,4 +133,138 @@ class DependencyInjectionJvmTest {
...
+            ::createBankService.qualifiedName,
+            ::createBankTellerWithLogging.qualifiedName,
+        ) {
+            val teller: BankTeller by dependencies
+            assertEquals(HELLO_CUSTOMER, teller.hello())
+        }
+    }
+
+    @Test
+    fun `install function ref with nullable args`() {
```
osipxd, line 242: Could you add a test when function reference points to a non-static or private function? Probably we should tweak exception message in these cases.

ktor-server/ktor-server-plugins/ktor-server-di/jvm/src/io/ktor/server/plugins/di/utils/ClasspathReference.jvm.kt
bjhham: Yeah, my thinking was that they can extend `DependencyReflectionJvm` for their custom implementation to override this behaviour.  It's not the best abstraction though... Part of the problem is that the interface lives in common, but most of the methods you'd want to override reference JVM-specific types.

ktor-server/ktor-server-plugins/ktor-server-di/jvm/src/io/ktor/server/plugins/di/utils/ClasspathReference.jvm.kt
bjhham: Ohh... Yeah, I think if people are using the `@file:JvmName()` annotation like that, then they can probably figure out why we're resolving the class name first.

ktor-server/ktor-server-plugins/ktor-server-di/jvm/test/io/ktor/server/plugins/di/DependencyInjectionJvmTest.kt
bjhham: Actually, I've inadvertently provided support for non-static functions, but it requires that an instance is registered to be called - this is because the function reference in kotlin includes an instance argument in the parameters that's treated normally.

As for the private function failure, you'll get an `java.lang.IllegalAccessException`, with the usual details.  I think that should be enough for debugging the failure.

I'll add the tests though, they're good scenarios to check.

I'd also like to include some extra details in the exceptions to reference back to the classpath reference in the config file... I'll add a task in YT to revisit this before the release.

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/utils/ClasspathReference.kt
```diff
@@ -8,24 +8,28 @@ import io.ktor.server.application.Application
...
         if (index == -1) return ""
         value.substring(0, index)
     }
 
     /**
      * The string value after the last delimiter (# or .) - generally the function name or class name.
      */
-    val name: String get() = value.lastIndexOfAny(delimiters).let { index ->
-        if (index == -1 || index == value.length - 1) return ""
+    val name: String get() = value.lastIndexOf('.').let { index ->
```
osipxd, line 31: Minor. If we check that `value` doesn't end with a dot (in init block or inside `invoke`), we will able to use `substringBeforeLast` and `substringAfterLast`

osipxd (approved):
LGTM!

ktor-server/ktor-server-plugins/ktor-server-di/jvm/test/io/ktor/server/plugins/di/DependencyInjectionJvmTest.kt
osipxd: > provided support for non-static functions, but it requires that an instance is registered to be called

Cool undocumented feature :D

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/utils/ClasspathReference.kt
bjhham: Oh right that's true 🤦

PR #4737, opened 2025-03-15
KTOR-8276 fix micrometer metrics oom trigger

hungrytech:
**Subsystem**
Micrometor metrics

**Motivation**
[KTOR-8276](https://youtrack.jetbrains.com/issue/KTOR-8276/MicrometerMetrics-different-path-404s-requests-can-be-abused-to-trigger-OOM)

**Solution**
Changed the default value of distinctNotRegisteredRoutes to false.

Since a test already exists for `no handler results in status 404, no route, and no exception if distinctNotRegisteredRoutes is false` no additional tests were written.


bjhham (approved):

hungrytech:
@bjhham 
Thank you for your review.

PR #4738, opened 2025-03-15
KTOR-8200 fix + encoding on appendPathSegments

hungrytech:
**Subsystem**
ktor-http

**Motivation**
[KTOR-8200](https://youtrack.jetbrains.com/issue/KTOR-8200/Inconsistent-behaviour-of-the-URLBuilder.appendPathSegments-and-URLBuilder.appendEncodedPathSegments)

**Solution**
Removed the '+' character from VALID_PATH_PART.


bjhham (changes requested):
Could you please look into fixing these tests?  It looks like it might just be a case of bad assertions, unless if you find some reason that they ought to expect the previous behaviour.

https://ktor.teamcity.com/buildConfiguration/Ktor_KtorMatrixCoreLinuxJava21/325088

hungrytech:
@bjhham 
Yes, I agree with what you mentioned. According to [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986), encoding may not be necessary. It seems that only the query part needs to be encoded.

bjhham:
Closing this, because + in the path is valid according to the RFC.

PR #4740, opened 2025-03-17
KTOR-7743 Make projects isolated

osipxd:
**Subsystem**
Build Infrastructure

**Motivation**
[KTOR-7743](https://youtrack.jetbrains.com/issue/KTOR-7743) Refactor Gradle configuration to use precompiled script plugins
[KTOR-8336](https://youtrack.jetbrains.com/issue/KTOR-8336) Remove empty artifacts from publication

**Solution**

**Removed cross-project configuration.** Plugins are applied in the project build script instead, so projects are self-content. This change removes coupling between projects and allows us to use [configuration on demand](https://docs.gradle.org/current/userguide/multi_project_configuration_and_execution.html) (now) and [isolated projects](https://docs.gradle.org/current/userguide/isolated_projects.html) (in the future).

**Changed dependency declaration syntax** to use `KotlinMultiplatformSourceSetConventions` introduced in Kotlin 1.9.20.
```kotlin
// Old style
kotlin.sourceSets {
    commonMain {
        dependencies {
            implementation(libs.kotlin.stdlib)
        }
    }
    commonTest {
        dependencies {
            implementation(libs.kotlin.test)
        }
    }
}

// New style
kotlin {
    sourceSets {
        commonMain.dependencies {
            implementation(libs.kotlin.stdlib)
        }
        commonTest.dependencies {
            implementation(libs.kotlin.test)
        }
    }
}
```
The new style is the same in terms of maximum nesting, but it has two advantages:
1. We can configure other kotlin-related settings (like C-interop) without need to add one more level of nesting
2. This style is more compact in vertical

**Refactored aggregating projects to use Lazy APIs.**
We have aggregating projects - projects using outputs of all other projects. As soon as our projects became independent, we should explicitly declare that evaluation of these projects depends on evaluation of all other projects. Moreover, evaluation should be done lazily, otherwise all projects might be evaluated without need. 

I've introduced `ProjectTagService` as a point of synchronization between "aggregating projects" and ordinary projects.
(Thanks to [kotlin-wrappers](https://github.com/JetBrains/kotlin-wrappers/blob/2025.3.15/buildSrc/src/main/kotlin/SubprojectService.kt) and @adam-enko for the idea).

_It's better review commit-by-commit._


ktor-client/build.gradle.kts
```diff
@@ -1,13 +1,15 @@
 /*
-* Copyright 2014-2021 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
-*/
+ * Copyright 2014-2025 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+plugins {
+    id("ktorbuild.project.library")
```
bjhham, line 6: It would be nice if we had some helpers in `buildSrc` to simplify this expression since it's everywhere.  Like `fun library() = plugins { id("ktorbuild.project.library") }`.  I'm guessing there's reasons that make this impossible or undesirable though...

bjhham (approved):
Is there validation for new libraries where the corresponding plugin is missing?

bjhham:
Can you merge main into 3.2.0-eap after this is merged?  I think it'll need a bit of resolving.

ktor-client/build.gradle.kts
osipxd: It think we could introduce accessors for plugins, like this:
```kotlin
plugins {
    ktorbuild.project.library
}
```
I don't want to hide `plugins` block behind shortcut because projects can have more than one plugin applied.

osipxd: Ah. I forgot that `plugins {}` block is very special. Gradle extracts it from build script and compiles separately, so it has some limitations. Extension-functions declared in `build-logic` aren't available in this block.

PR #4741, opened 2025-03-19
KTOR-7667: Enable Gradle configuration cache

osipxd:
**Subsystem**
Infrastructure

**Motivation**
[KTOR-7667](https://youtrack.jetbrains.com/issue/KTOR-7667) Enable Gradle configuration cache

**Solution**
Finally, I've found a hackaround for [KT-72933](https://youtrack.jetbrains.com/issue/KT-72933), and we can enable configuration cache 🎉 
It will work only for tasks not involving `MetadataDependencyTransformationTask`.

Some benchmark results for scenario "Rerun `:ktor-client:ktor-client-core:jvmTest` after a change in JVM source set."
image
_default_ - optimizations disabled
_COD_ - configuration on demand enabled
_CC_ - configuration cache enabled

Scenarios `default` and `COD` represent the case when there is no configuration cache that could be reused, let's call it "initial configuration".
Results show that configuration on demand significantly improves initial configuration configuration times and slightly improves storing and loading of configuration cache (because of smaller cache size)

**Next steps:**
- [KTOR-8332](https://youtrack.jetbrains.com/issue/KTOR-8332) Enable Gradle configuration cache for native tasks (as soon as KT-72933 is fixed)

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -43,6 +43,8 @@ doctor.enableTaskMonitoring=false
 org.gradle.daemon=true
 org.gradle.caching=true
 org.gradle.parallel=true
+org.gradle.configuration-cache=true
+org.gradle.configuration-cache.parallel=true
 org.gradle.kotlin.dsl.allWarningsAsErrors=true
 
 # kotlin
--- a/gradle.properties
+++ b/gradle.properties
@@ -56,6 +56,9 @@ kotlin.mpp.applyDefaultHierarchyTemplate=false
 kotlin.apple.xcodeCompatibility.nowarn=true
 kotlin.suppressGradlePluginWarnings=IncorrectCompileOnlyDependencyWarning
 kotlin.daemon.useFallbackStrategy=false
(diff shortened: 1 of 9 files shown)


gradle.properties
```diff
@@ -43,6 +43,8 @@ doctor.enableTaskMonitoring=false
 org.gradle.daemon=true
 org.gradle.caching=true
 org.gradle.parallel=true
+org.gradle.configuration-cache=true
+org.gradle.configuration-cache.parallel=true
```
osipxd, line 47: Probably we should also enable "configuration-on-demand" by default, but I'm not sure about it as this mode might silently break something. I would propose enabling this property locally and after couple of weeks we will add it here if everything goes okay.

```properties
org.gradle.configureondemand =true
```

bjhham (approved):

bjhham:
You might have broken some tests by making the build run twice as fast... 😅

PR #4742, opened 2025-03-19
KTOR-8306 Dependency injection tooling for tests

bjhham:
**Subsystem**
Server, DI

**Motivation**
[KTOR-8306](https://youtrack.jetbrains.com/issue/KTOR-8306) Dependency injection tooling for tests

This was requested in the design doc review as a possible improvement for test tooling.

**Solution**
Now, when the DI plugin is installed from the test engine, it will use a custom conflict policy that allows overriding implementations.

When writing tests, you can now do something like:

```kotlin
fun test() = testApplication {
    // 1. Install modules as they'd normally appear
    installProductionTypes()
    
    // 2. Replace specific implementations with mocks
    dependencies {
        provide<Database> { MockDatabaseInstance() }
    }
    
    // 3. Run tests
}
```

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyConflictPolicy.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyConflictPolicy.kt
@@ -68,3 +68,16 @@ public val DefaultConflictPolicy: DependencyConflictPolicy = DependencyConflictP
         is ExplicitCreateFunction -> current.ifImplicit { KeepPrevious } ?: Conflict
     }
 }
+
+/**
+ * During testing, we simply override previously declared values.
+ * This allows for replacing base implementations with mock values.
+ */
+public val LastEntryWinsPolicy: DependencyConflictPolicy = DependencyConflictPolicy { prev, current ->
+    require(current !is AmbiguousCreateFunction) { "Unexpected ambiguous function supplied" }
+    when (prev) {
+        is AmbiguousCreateFunction,
+        is ImplicitCreateFunction -> KeepNew
+        is ExplicitCreateFunction -> current.ifImplicit { KeepPrevious } ?: KeepNew
+    }
+}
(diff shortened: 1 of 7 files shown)


bjhham:
@coderabbitai review

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyInjection.kt
```diff
@@ -75,6 +82,9 @@ public val DI: ApplicationPlugin<DependencyInjectionConfig> =
         )
     }
 
+private fun PluginBuilder<*>.isTestEngine(): Boolean =
+    application.engine::class.simpleName == "TestApplicationEngine"
```
bjhham, line 86: Not sure if there's a better way to detect if you're inside the test engine, but this is what I came up with 🤷

bjhham, line 86: Not sure if there's a better way to detect if you're inside the test engine, but this is what I came up with 🤷

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyConflictPolicy.kt
```diff
@@ -68,3 +68,16 @@ public val DefaultConflictPolicy: DependencyConflictPolicy = DependencyConflictP
         is ExplicitCreateFunction -> current.ifImplicit { KeepPrevious } ?: Conflict
     }
 }
+
+/**
+ * During testing, we simply override previously declared values.
+ * This allows for replacing base implementations with mock values.
+ */
+public val LastEntryWinsPolicy: DependencyConflictPolicy = DependencyConflictPolicy { prev, current ->
```
e5l, line 76: would be nice to have trace logging for that

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyInjection.kt
e5l: lgtm

e5l:
btw `installProductionTypes` is not associated with modules for me. I would look for a better name for the docs.

e5l (approved):

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyConflictPolicy.kt
bjhham: Yeah, that's a good point, Koin's got some logging support, we ought to include it for this kind of thing.

PR #4744, opened 2025-03-20
KTOR-8291: replace cancel with cancelAndJoin()

tre3p:


ktor-server/ktor-server-core/common/src/io/ktor/server/application/Application.kt
```diff
@@ -127,7 +127,7 @@ public class Application internal constructor(
      */
     @Suppress("DEPRECATION_ERROR")
     public fun dispose() {
-        applicationJob.cancel()
+        applicationJob.cancelAndJoin()
```
bjhham, line 130: Doesn't look like this will work with the suspend call in a blocking context.  I'd suggest leaving this as is, then introducing a function for joining the application job and calling it from the relevant callers of dispose.

bjhham (changes requested):

PR #4745, opened 2025-03-21
KTOR-8344 Enable build cache for settings plugins

osipxd:
**Subsystem**
Infrastructure

**Motivation**
[KTOR-8344](https://youtrack.jetbrains.com/issue/KTOR-8344) Enable shared build cache for build-settings-logic

**Solution**
Enabled build cache for `build-settings-logic`.
This should reduce configuration times on CI by ~20 seconds (2 + 13 + 4) for all builds.

image

Also:
- removed redundant build scan customizations as they're [already done](https://github.com/gradle/common-custom-user-data-gradle-plugin/blob/v2.1/src/main/java/com/gradle/CustomBuildScanEnhancements.java) in `common-custom-user-data-gradle-plugin`.
- enabled obfuscation for usernames and hostnames


build-settings-logic/src/main/kotlin/ktorbuild.develocity.settings.gradle.kts
```diff
@@ -31,9 +34,9 @@ develocity {
             scanJournal.appendText("${Date()} — $buildScanUri — $startParameter\n")
         }
 
-        val skipBuildScans = settings.providers.gradleProperty("ktor.develocity.skipBuildScans")
-            .getOrElse("false")
-            .toBooleanStrict()
+        val skipBuildScans = providers.gradleProperty("ktor.develocity.skipBuildScans")
+            .orNull
+            .toBoolean()
```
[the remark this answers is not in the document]
osipxd: Thank you! It won't, as I'm using `String?.toBoolean` from Kotlin stdlib:

```kotlin
public actual inline fun String?.toBoolean(): Boolean = java.lang.Boolean.parseBoolean(this)
```

build-settings-logic/src/main/kotlin/customization.kt
```diff
@@ -1,64 +0,0 @@
-/*
- * Copyright 2014-2024 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
- */
-
-import com.gradle.develocity.agent.gradle.DevelocityConfiguration
-import org.gradle.api.initialization.Settings
-
-fun Settings.enrichTeamCityData() {
```
Mr3zee, line 8: Why did you remove all this?

osipxd: > as they're [already done](https://github.com/gradle/common-custom-user-data-gradle-plugin/blob/v2.1/src/main/java/com/gradle/CustomBuildScanEnhancements.java) in common-custom-user-data-gradle-plugin.

So now we have these values duplicated in build scans:
image

Mr3zee (approved):

PR #4747, opened 2025-03-21
Add the resource to client requests when using the `Resources` Plugin.

MarcusDunn:
**Subsystem**
Client, Resources

**Motivation**
This PR allows downstream plugins to introspect the resource used to create the request URL. 

The motivating use case for this is to allow spans to be named `GET api/user/{id}?query={query}` instead of `GET`, which is the best downstream telemetry can do right now.

This would result in a huge quality of life improvement around aggregating client requests to the same endpoint. Currently I'm writing regex against the full URL which is both slow, error prone and has to be done for each endpoint.

**Solution**
I've added a `RESOURCE` attribute to the request when the resources plugin is used.

~~I've made a seperate PR to opentelemetry-java-instrumentation to use this attribute to populate `url.template`  https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/13572~~

I will make a PR to opentelemetry-java-instrumentation once this is merged so I can test on that side. (as requested by their maintainers)

**Alternatives**

- Don't do anything. This is out of scope of KTOR. Application should be handling things like this.

### DIFF
--- a/ktor-client/ktor-client-plugins/ktor-client-resources/common/src/io/ktor/client/plugins/resources/builders.kt
+++ b/ktor-client/ktor-client-plugins/ktor-client-resources/common/src/io/ktor/client/plugins/resources/builders.kt
@@ -11,6 +11,7 @@ import io.ktor.client.statement.*
 import io.ktor.resources.*
 import io.ktor.resources.serialization.ResourcesFormat
 import io.ktor.util.AttributeKey
+import io.ktor.util.Attributes
 import kotlinx.serialization.*
 import io.ktor.client.request.delete as deleteBuilder
 import io.ktor.client.request.get as getBuilder
@@ -40,7 +41,7 @@ public suspend inline fun <reified T : Any> HttpClient.get(
 ): HttpResponse {
     val resources = resources()
     return getBuilder {
-        includeUrlTemplate(resources.resourcesFormat, serializer<T>())
+        attributes.putUrlTemplate(resources.resourcesFormat, serializer<T>())
         href(resources.resourcesFormat, resource, url)
         builder()
     }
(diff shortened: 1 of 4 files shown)


bjhham (approved):
Sorry, I missed this one!  Looks good 👍

MarcusDunn:
thanks @bjhham, anything you need from my end? I cant' see the TC results so not sure what to do on those failing checks.

bjhham:
Could you just perform a rebase to get the latest from main?  That should clear up any build problems I think.

MarcusDunn:
done.

ktor-client/ktor-client-plugins/ktor-client-resources/common/src/io/ktor/client/plugins/resources/builders.kt
```diff
@@ -277,11 +293,18 @@ public suspend inline fun <reified T : Any> HttpClient.prepareRequest(
...
         href(resources.resourcesFormat, resource, url)
         builder()
     }
 }
 
+/**
+ * The instance of the [Resource] annotated class used to for a request.
+ * Plugins may want to utilize this for monitoring.
+ */
+public val RESOURCE: AttributeKey<Any> = AttributeKey<Any>("RESOURCE")
```
bjhham: Could you actually change the string value to `Resource` to match the convention for attribute key casing?

bjhham (changes requested):
Looks like the new public constant needs an ABI dump.

Please run:
```
./gradlew dumpKotlinAbi
```
... and commit the snapshot result.

MarcusDunn:
sorry for the week delay @bjhham, but there's a fresh rebase and two commits for the ABI and the key change.

MarcusDunn:
@bjhham bumping this.

bjhham (approved):

bjhham:
Here's a YT ticket for tracking:
[KTOR-9709](https://youtrack.jetbrains.com/issue/KTOR-9709) Add resource attribute for client resources plugin

PR #4760, opened 2025-03-25
KTOR-8507: parents are include in query params when parent is an object.

MarcusDunn:
**Subsystem**
Resources

**Failing Test**
```kotlin
    @Resource("v1")
    private data object V1 {
        @Resource("api")
        data class Api(val parent: V1)
    }

    @Test
    fun testQueryParamsDoNotContainsObjectParent() {
        assertEquals(resourcesFormat.encodeToQueryParameters(serializer = serializer<V1.Api>()), emptySet())
    }
```

```
Expected :[]
Actual   :[Parameter(name=parent, isOptional=false)]
```

This is a bug, the parent is part of the path. not the query.

This PR fixes it.


bjhham (changes requested):
It doesn't really make sense to use objects as parameters here since they're statically scoped.  It would be better to rewrite your class here as `data object Api { val parent get() = V1 }`.

That being said, the added safety here is not unwelcome.  Could you review the one comment and also run `./gradlew formatKotlin` to pass the lint check?

MarcusDunn:
> `data object Api { val parent get() = V1 }`

This is much better. Never occurred to me.

> Could you review the one comment and also run ./gradlew formatKotlin to pass the lint check?

Seems to be running successfully without changes locally? I'm not able to see any CI runs to check why they are failin



Also, I ran into this when using some changes similar to https://github.com/ktorio/ktor/pull/4747 on my own project, would I be able to get a review on that? I would like to know if something like that would be considered to be merged, or if that's out of KTOR's scope and I should be working on a solution application-side...

bjhham (approved):
Looks good.  Sorry about the CI, we've been having some tests go flaky recently.  I can re-run them.

osipxd:
An issue to include this change in the changelog:
[KTOR-8507](https://youtrack.jetbrains.com/issue/KTOR-8507) Resources: Exclude a parent object from query params

PR #4761, opened 2025-03-26
KTOR-8291 Fix for coroutines not joined on shutdown

bjhham:
**Subsystem**
Server, Core

**Motivation**
[KTOR-8291](https://youtrack.jetbrains.com/issue/KTOR-8291) Application job is not joined during shutdown

**Solution**
Added a `disposeAndJoin()` internal API function on `Application` which waits for the application job to join after cancelling, then replaced all relevant references.  This is important for properly cleaning up resources when the server is closed or refreshed, because otherwise the process will stop before executing any `onCancelled` hooks are executed.

Note [KTOR-8338](https://youtrack.jetbrains.com/issue/KTOR-8338) will need to wait for 3.2.0, because it will require some more significant changes because only the CIO engine uses the application job as the parent for jobs spawned from the call context.


e5l (approved):
looks nice!

PR #4772, opened 2025-03-28
KTOR-6474 Fix compatibility of several plugins with SavedBodyPlugin

osipxd:
**Subsystem**
Client

**Motivation**
[KTOR-6474](https://youtrack.jetbrains.com/issue/KTOR-6474) SaveBodyPlugin: Logging plugin consumes response body
[KTOR-8302](https://youtrack.jetbrains.com/issue/KTOR-8302) Logging: "Content-Length mismatch" error when reading body in HttpResponseValidator

In addition to the issues above, `ContentEncoding` was also not compatible with `SaveBodyContent`.

**Solution**

The problem occurs when a plugin uses `call.wrapWithContent(content)` or `call.wrap(content, headers)`.
These calls neutralize the effect of `SaveBodyPlugin` making response body non-replayable. Plugins `ResponseObserver` (used for `Logging` implementation) and `ContentEncoding` are affected.

So when we combine any of the affected plugins with a plugin reading response body, there will be an issue.

Fixes:
- `ResponseObserver`: added logic to handle saved responses without `wrapWithContent` call.
- `ContentEncoding`: added a function `call.wrap(headers) { content }` with a lambda which will be called each time we read the content and used it instead

The next step is to deprecate the problematic functions in favor of versions with a lambda-parameter or, at least provide a warning about this behavior (to be done in [KTOR-8378](https://youtrack.jetbrains.com/issue/KTOR-8378)).


ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/observer/DelegatedCall.kt
```diff
@@ -40,6 +40,16 @@ public fun HttpClientCall.wrap(content: ByteReadChannel, headers: Headers): Http
     return DelegatedCall(client, content, this, headers)
 }
 
+/**
+ * Wrap existing [HttpClientCall] with new [headers] and content produced by the given [block].
+ *
+ * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.client.plugins.observer.wrap)
+ */
+@InternalAPI
+public fun HttpClientCall.wrap(headers: Headers, block: () -> ByteReadChannel): HttpClientCall {
```
e5l, line 49: could we make it internal for the patch release?

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/ResponseObserverTest.kt
```diff
@@ -92,4 +96,43 @@ class ResponseObserverTest : ClientLoader() {
...
+
+        test { client ->
+            // In the pipeline after ResponseObserver
+            client.receivePipeline.intercept(HttpReceivePipeline.After) { response ->
+                assertBodyCanBeReadMultipleTimes(response)
+                proceedWith(response)
+            }
+
+            val response = client.get("/")
+            assertTrue(wasSaved == true)
```
e5l, line 132: looks weird.
Let's split this assertNotNull and assertTrue, so when it fails we see where

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/observer/DelegatedCall.kt
```diff
@@ -40,6 +40,16 @@ public fun HttpClientCall.wrap(content: ByteReadChannel, headers: Headers): Http
     return DelegatedCall(client, content, this, headers)
 }
 
+/**
+ * Wrap existing [HttpClientCall] with new [headers] and content produced by the given [block].
```
e5l, line 44: please add an example of using it

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/observer/DelegatedCall.kt
```diff
@@ -40,6 +40,16 @@ public fun HttpClientCall.wrap(content: ByteReadChannel, headers: Headers): Http
...
+ *
+ * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.client.plugins.observer.wrap)
```
e5l, line 46: you don't have to add a link; we can run the link generator before 3.2.0

e5l (approved):
nicely done! please check CI before merging

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/observer/DelegatedCall.kt
osipxd: By the way, it seems like the `block` should take the current `rawContent` as an argument. What do you think?

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/observer/DelegatedCall.kt
```diff
@@ -40,6 +40,25 @@ public fun HttpClientCall.wrap(content: ByteReadChannel, headers: Headers): Http
...
+/**
+ * Wrap existing [HttpClientCall] with new [headers] and content produced by the given [block].
+ * The [block] will be called each time the response content is requested.
+ *
+ * ```
+ * // Create new response with modified headers and content
+ * val headers = headersOf("Custom-Header", "Custom-Value")
+ * val newCall = originalCall.wrap(headers) {
+ *     ByteReadChannel("Modified content")
+ * }
```
osipxd, line 52: @coderabbitai I've failed to come up with a useful example here. Do you have any suggestions?

osipxd: Thank you, fixed the example

bjhham (approved):

osipxd:
@coderabbitai pause

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/observer/DelegatedCall.kt
osipxd: Rebased onto EAP branch

osipxd:
@coderabbitai full review

PR #4775, opened 2025-03-31
Dependency injection annotations + config deserialization

bjhham:
**Subsystem**
Server, DI / Config

**Motivation**
- [KTOR-8321](https://youtrack.jetbrains.com/issue/KTOR-8321) Dependency injection named argument annotation
- [KTOR-8323](https://youtrack.jetbrains.com/issue/KTOR-8323) Dependency keys for configuration items
- [KTOR-7874](https://youtrack.jetbrains.com/issue/KTOR-7874) Configuration file deserialization

**Solution**
This introduces the `@Named` annotation for specifying named dependencies when using parameter injection, and `@Property` for referencing configuration values.

This also introduces the ability to deserialize properties from configuration, so when you have a few common properties it can be imported as a data class.

For example:
```
connection:
    domain: api.example.com
    path: /v1
    protocol: https
```

Can be imported like `application.property<Connection>("connection")` now instead of `application.environment.config.property("connection.domain").getString() x 3`.  This can also be used for parsing numbers and booleans.

Note that I ended up swapping the underlying implementation for the YAML config library to take advantage of some additional support for the interop between the raw nodes and serialization calls.


bjhham:
@coderabbitai review

ktor-server/ktor-server-config-yaml/jvmAndPosix/src/io/ktor/server/config/yaml/YamlConfig.kt
```diff
@@ -44,125 +46,161 @@ public expect fun YamlConfig(path: String?): YamlConfig?
...
+            )?.let { resolved ->
+                YamlScalar(resolved, path)
+            } ?: this
+            is YamlNull,
+            is YamlTaggedNode -> this
+        }
 
-        override fun getList(): List<String> = values
+    return replace() as YamlMap
+}
```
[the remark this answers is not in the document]
bjhham: I wouldn't worry about it this time, we're not expecting to blow the stack on some YAML config files.

ktor-server/ktor-server-config-yaml/build.gradle.kts
```diff
@@ -4,13 +4,14 @@
...
     id("ktorbuild.project.library")
+    id("kotlinx-serialization")
 }
 
 kotlin {
     sourceSets {
         commonMain.dependencies {
             api(project(":ktor-server:ktor-server-core"))
-            api(libs.yamlkt.serialization)
+            api(libs.kaml.serialization)
```
e5l, line 14: Changing API dependency looks like a breaking change. Could you check if we need this and if we can use implementation?

e5l: we still can use new

ktor-server/ktor-server-plugins/ktor-server-di/api/ktor-server-di.klib.api
```diff
@@ -15,12 +29,12 @@ abstract fun interface io.ktor.server.plugins.di/DependencyKeyCovariance { // io
...
+    abstract fun resolve(io.ktor.server.plugins.di/DependencyProvider, io.ktor.server.plugins.di/DependencyMap, io.ktor.server.plugins.di/DependencyReflection): io.ktor.server.plugins.di/DependencyResolver // io.ktor.server.plugins.di/DependencyResolution.resolve|resolve(io.ktor.server.plugins.di.DependencyProvider;io.ktor.server.plugins.di.DependencyMap;io.ktor.server.plugins.di.DependencyReflection){}[0]
 }
 
 abstract interface io.ktor.server.plugins.di/DependencyMap { // io.ktor.server.plugins.di/DependencyMap|null[0]
     abstract fun <#A1: kotlin/Any> get(io.ktor.server.plugins.di/DependencyKey): #A1 // io.ktor.server.plugins.di/DependencyMap.get|get(io.ktor.server.plugins.di.DependencyKey){0§<kotlin.Any>}[0]
-    abstract fun <#A1: kotlin/Any> getOrPut(io.ktor.server.plugins.di/DependencyKey, kotlin/Function0<#A1>): #A1 // io.ktor.server.plugins.di/DependencyMap.getOrPut|getOrPut(io.ktor.server.plugins.di.DependencyKey;kotlin.Function0<0:0>){0§<kotlin.Any>}[0]
+    abstract fun contains(io.ktor.server.plugins.di/DependencyKey): kotlin/Boolean // io.ktor.server.plugins.di/DependencyMap.contains|contains(io.ktor.server.plugins.di.DependencyKey){}[0]
```
e5l, line 37: please add old methods as deprecated to avoid breaking changes

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyInjection.kt
```diff
@@ -132,6 +132,18 @@ public class DependencyInjectionConfig {
...
+    internal var dependenciesMap: DependencyMap? = null
+
+    /**
+     * Include an additional source for dependencies.
+     *
+     * Sources added this way take precedence in the order of inclusion.
+     *
+     * Declared dependencies have precedence over all sources included this way.
+     */
+    public fun include(map: DependencyMap) {
```
e5l, line 144: could you add an example in KDocs

e5l (changes requested):
lgtm, please check KDocs and breaking changes

ktor-server/ktor-server-plugins/ktor-server-di/api/ktor-server-di.klib.api
bjhham: This is unreleased code for DI, so I think it should be ok.

ktor-server/ktor-server-config-yaml/build.gradle.kts
bjhham: Yeah, I was wondering what to do here...  Probably should've used implementation to begin with since it's not exposed anywhere in the API.

vnikolova (commented):
The KDocs look good to me! Added some minor suggestions below.

e5l (approved):
lgtm

PR #4777, opened 2025-04-01
KTOR-8236: support for 'Vary' header

tre3p:
**Subsystem**
ktor-http, ktor-server-core

**Motivation**
KTOR-8326. Added support for 'Vary' header in responses.

**Solution**
Append 'Accept-Encoding' header to 'Vary' if response was compressed based on Accept-Encoding values received from client call.


tre3p:
@bjhham Hey, can you check it out, please?

bjhham (changes requested):
Could you please include some unit tests to verify this behaviour?

Perhaps some assertions in `io.ktor.server.plugins.CompressionTest` or `io.ktor.server.plugins.StaticContentTest`

tre3p:
Done! Thanks for pointing out the files with tests, that was really helpful :)

tre3p:
Sorry, do you want me to do something with native tests? Honestly, I don't have an idea how my changes could affect this tests. Any tips?

PR #4787, opened 2025-04-08
KTOR-8318: fix apache entity producer handler

anton-erofeev:
**Subsystem**
Client, ktor-client-apache5

**Motivation**
https://youtrack.jetbrains.com/issue/KTOR-8318

**Solution**
Fixed entity producer handler

### DIFF
--- a/ktor-client/ktor-client-apache/jvm/src/io/ktor/client/engine/apache/ApacheRequestProducer.kt
+++ b/ktor-client/ktor-client-apache/jvm/src/io/ktor/client/engine/apache/ApacheRequestProducer.kt
@@ -129,8 +129,10 @@ internal class ApacheRequestProducer(
                 else -> builder.addHeader(key, value)
             }
         }
+        val isGetOrHeadOrOptions = method == HttpMethod.Get || method == HttpMethod.Head || method == HttpMethod.Options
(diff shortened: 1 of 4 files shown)


bjhham (commented):
Thanks for the PR!  It looks like there's an issue with `io.ktor.client.tests.plugins.ServerSentEventsTest` now.  Maybe it relies on GET for initiating SSE connections.  @marychatte could you advise?

anton-erofeev:
Thanks, in that case let's continue with the implementation via `!hasContent && isGetOrHeadOrOptions`. (btw, similar logic is already [used in ktor](https://github.com/ktorio/ktor/blob/c9c7cf79de1d80e1463b2d40e79782147967773d/ktor-http/ktor-http-cio/common/src/io/ktor/http/cio/HttpBody.kt#L58)) .  We need to leave the check for the http method because we have 3 cases

1) there is a body and we pass it for processing
2) there is no body and the method is `Get`/`Head`/`Options` - we do not pass the body (`dataProducer = null`)
3) there is no body and the method is not get/head/options - we pass an empty body for processing (we still create `ApacheRequestEntityProducer`)

please take a look at my latest commit

marychatte (approved):
Looks good to me

PR #4789, opened 2025-04-09
Add ',' to parameter quoting terminals in HttpHeaderValueParser

clemairej:
**Subsystem**
HTTP - header parsing

**Motivation**
I want to replace HTTP Header parsing within Intellij Platform (specifically collab tools subsystems) with Ktor parsing.
Sadly, we are currently blocked by this bug: parsing a header value with parameters surrounded by double-annotation marks, immediately followed by a comma fails. Ex: `k: "v", k2: "v2"` does not correctly parse after the first `"`.

**Solution**
Add the comma exception to the list of characters that can follow `"`

### DIFF
--- a/ktor-http/common/src/io/ktor/http/HttpHeaderValueParser.kt
+++ b/ktor-http/common/src/io/ktor/http/HttpHeaderValueParser.kt
@@ -213,7 +213,7 @@ private fun parseHeaderValueParameterValueQuoted(value: String, start: Int): Pai
         val currentChar = value[position]
 
         when {
-            currentChar == '"' && value.nextIsSemicolonOrEnd(position) -> {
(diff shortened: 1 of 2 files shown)


ktor-http/common/src/io/ktor/http/HttpHeaderValueParser.kt
```diff
@@ -238,5 +238,5 @@ private fun String.nextIsSemicolonOrEnd(start: Int): Boolean {
         position += 1
     }
 
-    return position == length || get(position) == ';'
+    return position == length || get(position) == ';' || get(position) == ','
```
bjhham, line 241: Could you please update the name of the function to reflect the new behavior?

bjhham (commented):
Could uou

bjhham (approved):

PR #4792, opened 2025-04-10
KTOR-8338 Fix for call context coroutines not being cancelled

bjhham:
**Subsystem**
Server, Netty

**Motivation**
[KTOR-8338](https://youtrack.jetbrains.com/issue/KTOR-8338) Coroutines launched from RoutingContext are not cancelled upon server shutdown

**Solution**
This avoids using the netty dispatcher for the user context within routing calls.  Previously, the underlying executor for the netty dispatcher would get closed, then `onComplete` would fail to trigger because of the missing executor.  Now, the netty dispatcher will only be used for handling the calls themselves, and not other coroutines launched from the call.

### DIFF
--- a/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationEngine.kt
+++ b/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationEngine.kt
@@ -23,7 +23,6 @@ import io.netty.channel.socket.nio.NioServerSocketChannel
 import io.netty.handler.codec.http.HttpObjectDecoder
 import io.netty.handler.codec.http.HttpServerCodec
 import kotlinx.coroutines.CompletableJob
-import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.asCoroutineDispatcher
 import java.net.BindException
 import java.util.concurrent.TimeUnit
--- a/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationEngine.kt
+++ b/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationEngine.kt
@@ -185,10 +185,6 @@ public class NettyApplicationEngine(
         }
     }
 
-    private val nettyDispatcher: CoroutineDispatcher by lazy {
-        NettyDispatcher
-    }
-
     private val workerDispatcher by lazy {
         workerEventGroup.asCoroutineDispatcher()
     }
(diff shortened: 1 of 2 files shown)


ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationEngine.kt
```diff
@@ -215,6 +206,10 @@ public class NettyApplicationEngine(
                 channel(getChannelClass().java)
             }
 
+            val userContext = applicationProvider().coroutineContext +
```
nomisRev, line 209: This is a significant change. I know a lot of `embeddedServer` users provide a dispatcher explicitly, so I think we should check `parentCoroutineContext != EmptyCoroutineContext` before setting a different dispatcher.

e5l: Please also check if we need to recreate context for every channel

e5l (approved):
lgtm, please check CI and comments before merging

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationEngine.kt
bjhham: > Please also check if we need to recreate context for every channel

This is only called on startup as far as I can tell.  I moved into this block because it's theoretically possible `applicationProvider()` could have a different result upon instantiation.

> This is a significant change. I know a lot of embeddedServer users provide a dispatcher explicitly, so I think we should check parentCoroutineContext != EmptyCoroutineContext before setting a different dispatcher.

@nomisRev could you elaborate a bit here?  From what I understand, the current behaviour is that the user-provided dispatcher for the application would be ignored in lieu of the Netty dispatcher - so this fix ought to make the behaviour consistent with other engines.

nomisRev: I just checked the code, and this is a big change. `application.coroutineContext = parentCoroutineContext + applicationJob`, so now we're connected the jobs whilst before we weren't. I checked git history, and @e5l you indeed make a fix for memory optimisation. [KTOR-3903](https://youtrack.jetbrains.com/issue/KTOR-3903).

So I think this would be better `if(application.parentCoroutineContext == EmptyCoroutineContext) Dispatchers.Default else application.parentCoroutineContext`.

bjhham: Ah, thanks for looking into it!  The problem with using the parent coroutine context here is that it won't actually fix the bug with routing context coroutines because we're relying on the cancellation cascade from the server which doesn't affect the parent context.

With regards to the memory optimisation, our allocation tests should pick up any changes there and they look good https://ktor.teamcity.com/buildConfiguration/Ktor_AllocationTests/33151 - could it be that coroutines have improved memory allocations since our optimisations?

Maybe I'll do a little more testing and discuss with @e5l next week.

nomisRev: Right, we're relying on the cascade I thought it was implemented differently now but that was just working through Netty's executor being shutdown, and nothing got cancelled. Since the Application Job is the root, using any other root would result in the same overhead. So KotlinX Coroutines should have fixed the memory footprint, or it'll need to, but we should find out from our AllocationTests/benchmarks.

Jetty has the same behavior as this PR, I checked for my own understanding.  Could you keep me in the loop on those if they run on the changes of this PR? 🙏

PR #4794, opened 2025-04-14
KTOR-8407 Performance fix for readUntil function

bjhham:
**Subsystem**
Core, I/O

**Motivation**
- [KTOR-8407](https://youtrack.jetbrains.com/issue/KTOR-8407) Receiving multipart without Content-Length is very slow
- [KTOR-8412](https://youtrack.jetbrains.com/issue/KTOR-8412) ByteChannel single-byte operations are slow

**Solution**
- There was some redundant checking in some of the simpler functions like `readByte()` that adds up, where I was able to shave off about a third of the time, but should improve performance a little bit everywhere in the project.
- The big fix was in the `readUntil` function which was reading and writing bytes in sequence - replacing this with an `indexOf` call for the first matching byte and using bulk transfers brought it from around 60 seconds down to 3 seconds for processing a gigabyte.


ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
```diff
@@ -568,53 +563,92 @@ public suspend fun ByteReadChannel.readUntil(
...
+        return false
+    }
 
-        if (rc > limit) {
-            throw IOException("Limit of $limit bytes exceeded while scanning for \"${matchString.decodeToString()}\"")
+    while (!input.exhausted() || awaitContent()) {
+        advanceToNextMatch()
+        writeChannel.flush()
+
+        if (readCompleteMatch()) {
```
Stexxe, line 639: What is the reason for extracting the code into a function and calling it only once? Such segregation harms readability.

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
```diff
@@ -568,53 +563,92 @@ public suspend fun ByteReadChannel.readUntil(
...
-            writeChannel.writeByte(byte)
-            rc++
+            partialMatchBuffer.writeByte(byte)
         }
+        return false
+    }
 
-        if (rc > limit) {
-            throw IOException("Limit of $limit bytes exceeded while scanning for \"${matchString.decodeToString()}\"")
+    while (!input.exhausted() || awaitContent()) {
```
Stexxe, line 635: Is it intended that the code in the `advanceToNextMatch` and `readCompleteMatch` functions have the same condition in the while loops?

ktor-io/common/test/ByteReadChannelOperationsTest.kt
```diff
@@ -232,6 +231,23 @@ class ByteReadChannelOperationsTest {
         }
     }
 
+    @Test
+    fun readUntilLargeRandom() = runTest {
```
Stexxe, line 235: What is purpose of this test?

ktor-io/common/src/io/ktor/utils/io/CountedByteReadChannel.kt
```diff
@@ -36,15 +36,27 @@ public class CountedByteReadChannel(public val delegate: ByteReadChannel) : Byte
...
         }
 
     @OptIn(InternalAPI::class)
     override suspend fun awaitContent(min: Int): Boolean {
-        return readBuffer.size >= min || delegate.awaitContent(min)
+        if (readBuffer.size >= min) {
+            return true
+        }
+        if (delegate.awaitContent(min)) {
+            transferFromDelegate()
```
Stexxe, line 49: Is it okay that we transfer bytes from the buffer of the source channel in the `awaitContent` method?

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
```diff
@@ -568,53 +563,92 @@ public suspend fun ByteReadChannel.readUntil(
...
 
-        if (matchIndex > 0 && byte != matchString[matchIndex]) {
-            appendPartialMatch()
-            resetPartialMatch(byte)
-        }
+            if (matchIndex > 0 && byte != matchString[matchIndex]) {
+                // update match index from our table
+                val oldMatchIndex = matchIndex
+                while (matchIndex > 0 && byte != matchString[matchIndex]) {
+                    matchIndex = partialMatchTable[matchIndex - 1]
```
Stexxe, line 609: Initially, the `matchIndex` is 0, and that's the only place I found where it's reassigned. So, how is that variable updated?

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
```diff
@@ -40,15 +40,11 @@ public suspend fun ByteReadChannel.toByteArray(): ByteArray {
 
 @OptIn(InternalAPI::class)
 public suspend fun ByteReadChannel.readByte(): Byte {
-    if (readBuffer.exhausted()) {
-        awaitContent()
-    }
-
-    if (readBuffer.exhausted()) {
+    val currentBuffer = readBuffer
```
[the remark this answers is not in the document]
bjhham: Yes, it actually looks like this:

```
@InternalAPI
override val readBuffer: Source
    get() {
        _closedCause.value?.throwOrNull(::ClosedReadChannelException)
        if (_readBuffer.exhausted()) moveFlushToReadBuffer()
        return _readBuffer
    }
```

Which is not so bad, but it adds up.  Especially when working with low-level, performance-sensitive code.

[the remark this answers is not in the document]
Stexxe: It's not apparent that we hide such an amount of work under the property access.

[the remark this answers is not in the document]
bjhham: Yeah, I don't like it either... Hoping to revisit it in 4.0.

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
bjhham: Actually, inlining seems problematic with the multiple return statements.  Maybe I'll extract a utility class here for legibility...

Stexxe: Please don't extract it to a separate class :)

bjhham: Too late, you've convinced me :D

bjhham: It's much nicer as a class now, you can use it for repeated searches to save some overhead too.

ktor-io/common/src/io/ktor/utils/io/ByteChannelScanner.kt
```diff
@@ -0,0 +1,173 @@
...
+     * @return The total number of bytes read up to the match or end of input
+     * @throws IOException if the match is not found and `ignoreMissing` is `false`.
+     */
+    internal suspend fun findNext(ignoreMissing: Boolean = false): Long {
+        bytesRead = 0L
+
+        while (!input.exhausted() || channel.awaitContent()) {
+            // Quick scan for first byte
+            advanceToNextPotentialMatch()
+            writeChannel.flush()
```
marychatte, line 56: Is it necessary to flush after every skipped segment, or can we do it less often?

marychatte (approved):

ktor-io/common/src/io/ktor/utils/io/ByteChannelScanner.kt
bjhham: I think I ought to replace with `flushIfNeeded()` and move inside the `advanceToNextPotentialMatch` method.

PR #4800, opened 2025-04-15
KTOR-8411 CountedByteWriteChannel: fix autoFlush not working

SerdaOzun:
autoFlush of the source channel doesn't make the channel auto flushing

**Subsystem**
ktor-io

**Motivation**
I reported the bug [KTOR-8411](https://youtrack.jetbrains.com/issue/KTOR-8411/CountedByteWriteChannel-autoFlush-of-the-source-channel-doesnt-make-the-channel-auto-flushing) yesterday and took a deeper look to get it fixed, since it looked simple.

**Solution**
I decided to move the `autoFlush` property higher up the hierarchy so that all implementations may actually have access to it and a cast to ByteChannel won't be necessary. The cast fails for some implementations due to the inheritance hierarchy, hence the bug.

### DIFF
--- a/ktor-io/common/src/io/ktor/utils/io/ByteWriteChannel.kt
+++ b/ktor-io/common/src/io/ktor/utils/io/ByteWriteChannel.kt
@@ -18,6 +18,9 @@ import kotlinx.io.*
  */
 public interface ByteWriteChannel {
 
+    public val autoFlush: Boolean
+        get() = false
+
     public val isClosedForWrite: Boolean
 
     public val closedCause: Throwable?
@@ -58,5 +61,5 @@ public fun ByteWriteChannel.cancel() {
 public suspend fun ByteWriteChannel.flushIfNeeded() {
     rethrowCloseCauseIfNeeded()
 
-    if ((this as? ByteChannel)?.autoFlush == true || writeBuffer.size >= CHANNEL_MAX_SIZE) flush()
+    if (this.autoFlush == true || writeBuffer.size >= CHANNEL_MAX_SIZE) flush()
 }
--- a/ktor-network/common/test/io/ktor/network/sockets/tests/TCPSocketTest.kt
+++ b/ktor-network/common/test/io/ktor/network/sockets/tests/TCPSocketTest.kt
@@ -306,4 +306,28 @@ class TCPSocketTest {
             socket.awaitClosed()
         }
     }
+
+    @Test
+    fun testAutoFlush() = testSockets { selector ->
+        val tcp = aSocket(selector).tcp()
+        val server: ServerSocket = tcp.bind("127.0.0.1", port = 0)
+
+        val serverConnectionPromise = async {
+            server.accept()
+        }
+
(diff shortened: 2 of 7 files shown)


ktor-network/common/test/io/ktor/network/sockets/tests/TCPSocketTest.kt
```diff
@@ -306,4 +306,32 @@ class TCPSocketTest {
...
+        val clientConnection = tcp.connect("127.0.0.1", port = server.port)
+        val serverConnection = serverConnectionPromise.await()
+        val serverInput = serverConnection.openReadChannel()
+
+        val writeChannel = clientConnection.openWriteChannel(autoFlush = true)
+        writeChannel.writeStringUtf8("Hello, world\n")
+        delay(100) //wait for read to be available
+        assertTrue(serverInput.availableForRead > 0)
+        val message = serverInput.readUTF8Line()
+        assertEquals("Hello, world", message)
```
bjhham, line 328: We can avoid delays here by reading the line and letting it suspend.

```suggestion
        val message = serverInput.readUTF8Line()
        assertEquals("Hello, world", message)
```

ktor-io/common/src/io/ktor/utils/io/ByteWriteChannel.kt
```diff
@@ -18,6 +18,8 @@ import kotlinx.io.*
  */
 public interface ByteWriteChannel {
 
+    public val autoFlush: Boolean?
```
bjhham, line 21: Might as well use:

```
public val autoFlush: Boolean get() = false
```

A default of false seems correct in this case.

bjhham (changes requested):
Thanks for the PR!

Could you change the target branch to `3.2.0-eap` for the pull request, and also run `./gradlew apiDump` and commit the API snapshot changes?

ktor-network/common/test/io/ktor/network/sockets/tests/TCPSocketTest.kt
SerdaOzun: I've changed it to your suggestion, but this would lead to tests running into a longer timeout should they fail in the future

SerdaOzun:
I messed up the rebase somewhere and to not waste further time I force pushed a single commit now. Hope that's ok

SerdaOzun:
> Could you rebase your fork as well? Looks like a couple of commits from main are in the diff.


I messed up the rebase somewhere and to not waste further time I force pushed a single commit now. Hope that's ok

bjhham (approved):

bjhham (changes requested):
Something has gone wonky with the history here...

SerdaOzun:
should be ok again

bjhham:
Sorry, I've reverted the force push of the base branch, it was our bad.  Can you revert back again 😓

bjhham (approved):

bjhham:
The JS/Wasm tests are hanging for some reason and only on this branch...  No idea why though.  I'll investigate.

PR #4803, opened 2025-04-17
#4801 Set HTTP method earlier in the request builder block

### LINKED ISSUE #4801: Request builder sees incorrect HttpMethod
The `HttpClient.post(builder: HttpRequestBuilder): HttpResponse` method (see [source code](https://github.com/ktorio/ktor/blob/14d10c0cdd270cf01b1580ac8edd0726f8a5d2d4/ktor-client/ktor-client-core/common/src/io/ktor/client/request/builders.kt#L216)) sets the HTTP method to `HttpMethod.Post` right before building the request. Other shortcut methods (e.g., put, delete, etc.) behave similarly. As a result, the following test fails:

```kotlin
@Test
fun reproducer() = testApplication {
    client.post("/") {
        assertEquals(expected = HttpMethod.Post, actual = method)
    }
}
```

Having access to the actual `HttpMethod` during request configuration can be useful — for example, when adding a signature parameter to the request.

As a workaround, I switched from using `HttpClient.post` to `HttpClient.request`, where explicitly setting the HTTP method makes the code more expressive and avoids ambiguity during request configuration.

Stexxe:
Fixes #4801

### DIFF
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/request/builders.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/request/builders.kt
@@ -383,6 +383,7 @@ public suspend inline fun HttpClient.put(block: HttpRequestBuilder.() -> Unit): 
     builder.method = HttpMethod.Put
     return request(builder.apply(block))
 }
+
 /**
  * Executes an [HttpClient]'s DELETE request with the parameters configured in [block].
  *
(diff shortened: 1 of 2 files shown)


bjhham (approved):
Interesting problem!

hfhbd:
@Stexxe This change is not part of the (auto-generated) changelog.

PR #4806, opened 2025-04-22
KTOR-8204 Fix for exception on empty chunked body response

bjhham:
**Subsystem**
Client, SSE, I/O

**Motivation**
[KTOR-8204](https://youtrack.jetbrains.com/issue/KTOR-8204) CIO: Exception Chunked stream has ended unexpectedly

**Solution**
It seems to be fairly common to experience connection hangups before events start writing which can lead to this chunked transfer encoding validation error.  Instead of throwing the exception now, we'll just return when there's no chunk size.

Also, I added a catch for closed channels to avoid throwing on normal disconnects when parsing, and I updated the SSE tests a little to shorten up the time they're taking to run.

### DIFF
--- a/ktor-http/ktor-http-cio/common/src/io/ktor/http/cio/ChunkedTransferEncoding.kt
+++ b/ktor-http/ktor-http-cio/common/src/io/ktor/http/cio/ChunkedTransferEncoding.kt
@@ -69,11 +69,8 @@ public suspend fun decodeChunked(input: ByteReadChannel, out: ByteWriteChannel) 
     var totalBytesCopied = 0L
 
     try {
-        while (true) {
-            chunkSizeBuffer.clear()
-            if (!input.readUTF8LineTo(chunkSizeBuffer, MAX_CHUNK_SIZE_LENGTH, httpLineEndings)) {
-                throw EOFException("Chunked stream has ended unexpectedly: no chunk size")
-            } else if (chunkSizeBuffer.isEmpty()) {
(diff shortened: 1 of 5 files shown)


ktor-http/ktor-http-cio/common/src/io/ktor/http/cio/ChunkedTransferEncoding.kt
```diff
@@ -69,11 +69,8 @@ public suspend fun decodeChunked(input: ByteReadChannel, out: ByteWriteChannel)
     var totalBytesCopied = 0L
 
     try {
-        while (true) {
-            chunkSizeBuffer.clear()
-            if (!input.readUTF8LineTo(chunkSizeBuffer, MAX_CHUNK_SIZE_LENGTH, httpLineEndings)) {
-                throw EOFException("Chunked stream has ended unexpectedly: no chunk size")
-            } else if (chunkSizeBuffer.isEmpty()) {
+        while (input.readUTF8LineTo(chunkSizeBuffer, MAX_CHUNK_SIZE_LENGTH, httpLineEndings)) {
```
marychatte, line 72: Should we call `chunkSizeBuffer.clear()` as before?

marychatte (approved):
lgtm

ktor-http/ktor-http-cio/common/src/io/ktor/http/cio/ChunkedTransferEncoding.kt
bjhham: Turns out it was redundant, because it's cleared at the end of each iteration, and the buffer pool is designed to always return cleared buffers ☺️

PR #4807, opened 2025-04-23
KTOR-8451: Fix toString for staticContentRoute

zhurs:
**Subsystem**
Server, core

**Motivation**
Route rendering is important in several places, for example in metrics collection.
Currently for static files we have metrics like
```
ktor_http_server_requests_seconds_sum{method="GET",route="/io.ktor.server.http.content.StaticContentKt$staticContentRoute$1@70830e08/images/{...}",status="404",
throwable="n/a"} 0.227668303
```

**Solution**
Provide correct toString implementation for anonymous Route subclass, created in `staticContentRoute` method.


bjhham (approved):
LGTM

osipxd:
Issue in YouTrack to include the fix in the changelog: https://youtrack.jetbrains.com/issue/KTOR-8451/Implement-toString-for-staticContentRoute

PR #4809, opened 2025-04-23
KTOR-8378: Deprecate wrap and wrapWithContent without lambda

osipxd:
**Subsystem**
Client

**Motivation**
[KTOR-8378](https://youtrack.jetbrains.com/issue/KTOR-8378) HttpClientCall: Deprecate `wrapWithContent` and `wrap` without a lambda

**Solution**
Deprecate old functions in favor of their versions with lambda.


osipxd:
@coderabbitai review

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/observer/DelegatedCall.kt
```diff
@@ -54,7 +63,6 @@ public fun HttpClientCall.wrap(content: ByteReadChannel, headers: Headers): Http
  *
  * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.client.plugins.observer.wrap)
  */
-@InternalAPI
 public fun HttpClientCall.wrap(headers: Headers, block: () -> ByteReadChannel): HttpClientCall {
```
osipxd, line 66: To be discussed. Should we pass current `rawContent` as a parameter to the `block`?
We could rename `wrap` into `transform` (or `transformContent`) and then it would be logical to have such a parameter.

bjhham: I feel like `withBody` might be more descriptive since you could put anything in the lambda here.

osipxd: Yes, it is better. But parameter `headers` is a bit off this naming 🤔

osipxd: What do you think about `transformResponse`?

```kotlin
public fun HttpClientCall.transformResponse(
    headers: Headers = response.headers,
    content: () -> ByteReadChannel,
): HttpClientCall
```

osipxd:
@coderabbitai review

osipxd:
I'm sorry for this "edit storm". I forgot that the request is not in WIP status 😅 
I hope it was the last edit, and now the request is ready for review.

@coderabbitai full review

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/observer/DelegatedCall.kt
```diff
@@ -1,106 +1,59 @@
...
- *
- * ```
- * // Example: Content decompression. See ContentEncoding implementation for full example
- * val originalContent = originalCall.response.content
- * val decodedCall = originalCall.wrap(headersOf("Content-Encoding", "identity")) {
- *     decodeGzip(originalContent) // returns a new ByteReadChannel with decoded content
- * }
- * ```
+ * **Warning:** The content of the returned call response is non-replayable, so it can be consumed only once.
+ * Consider using `wrapWithContent` with lambda instead.
```
osipxd, line 47: Oh no. I've noticed another typo here. It should be:
```suggestion
 * **Warning:** The content of the returned call response is non-replayable, so it can be consumed only once.
 * Consider using [withResponse].
```

osipxd:
@coderabbitai full review

ktor-client/ktor-client-core/common/src/io/ktor/client/call/DelegatedCall.kt
```diff
@@ -0,0 +1,78 @@
...
+ * val decodedCall = originalCall.withResponse(
+ *     headers = headersOf("Content-Encoding", "identity")
+ * ) { // this: HttpResponse ->
+ *     decodeGzip(rawContent) // returns a new ByteReadChannel with decoded content
+ * }
+ * ```
+ *
+ * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.client.call.withResponse)
+ */
+public fun HttpClientCall.withResponse(
```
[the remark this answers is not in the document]
osipxd: Yes, this option was on the table. I see two problems with this option:
1. We change not only the content but also headers,
2.  `map` implies transformation of \*something\*, and it is not true for the current signature.

I could change the signature to fix the second problem, but the first one would still remain:
```kotlin
public fun HttpClientCall.mapResponseContent(
    headers: Headers = response.headers,
    transform: (ByteReadChannel) -> ByteReadChannel,
): HttpClientCall
```

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/observer/DelegatedCall.kt
bjhham: Yeah, that works 👍

ktor-client/ktor-client-core/common/src/io/ktor/client/call/DelegatedCall.kt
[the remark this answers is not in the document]
bjhham: `replaceResponse()`??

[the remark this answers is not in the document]
e5l: lgtm!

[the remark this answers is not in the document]
osipxd: It is ready for review. There are some failed tests, but they're flaky.

e5l (approved):

PR #4810, opened 2025-04-24
KTOR-8367 New SaveBody plugin

osipxd:
**Subsystem**
Client

**Motivation**
[KTOR-8367](https://youtrack.jetbrains.com/issue/KTOR-8367) Deprecate SaveBodyPlugin in favor of HttpClientCall.save

**Solution**
- Deprecate SaveBodyPlugin and HttpRequestBuilder.skipSavingBody with a warning.
- Introduce a new internal plugin `SaveBody` calling `HttpClientCall.save` and releasing resources for non-streaming responses after `HttpSend` and before all other plugins.

### [Benchmark](https://github.com/ktorio/ktor-benchmarks/pull/55) results

_All results are shown for Apache engine because CIO has other changes affecting the number of allocations, and results weren’t clean. All engines have [shown similar results](https://ktor.teamcity.com/buildConfiguration/Ktor_AllocationTests/340988?buildTab=tests&name=Client)._

---

**"Hello, world!" response:**
image

---

**File response (~200 KB HTML file):**
image

---

**Gzip response (2 MB):**
image

---

**Streaming response (2 MB):**
image

---

All test cases (except the streaming response) show reduced footprint. The location `ByteChannelReplay` was removed, and the location `SavedCall` has doubled its size (because of added `save()` call in the `SaveBody` plugin). Streaming responses aren’t affected by the changes, as expected.
Benchmarks show that we save all responses in memory twice per request (previously it was the same). This is the point where we could save some memory.

The only case which will be affected by these changes badly is using `HttpRequestBuilder.skipSaveBody()` extension. Previously the responses for which this extension was called were saved to memory once, and now they will be saved twice. Previous behavior was misleading. The only correct way to skip saving the body is to work with a request in a streaming way: `client.prepareRequest(...).execute { ... }`


ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/SaveBody.kt
```diff
@@ -2,20 +2,48 @@
...
+        if (attributes.contains(SKIP_SAVE_BODY)) return@intercept
+
+        val savedResponse = try {
+            call.save().response
+        } finally {
+            response.cleanup()
+        }
+
+        attributes.put(RESPONSE_BODY_SAVED, Unit)
+        proceedWith(savedResponse)
```
osipxd, line 39: **To be discussed.** I'm waiting for cleanup before proceeding with the pipeline. Should we run cleanup in parallel instead?
```suggestion
        val cleanupJob: Job
        val savedResponse = try {
            call.save().response
        } finally {
            cleanupJob = launch { response.cleanup() }
        }

        attributes.put(RESPONSE_BODY_SAVED, Unit)
        proceedWith(savedResponse)
        cleanupJob.join()
```

osipxd:
@coderabbitai review

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/SaveBody.kt
```diff
@@ -2,20 +2,53 @@
...
+internal val SaveBody: ClientPlugin<Unit> = createClientPlugin("SaveBody") {
+    client.receivePipeline.intercept(HttpReceivePipeline.Before) { response ->
+        val call = response.call
+        val attributes = call.attributes
+        if (attributes.contains(SKIP_SAVE_BODY)) return@intercept
+
+        val savedResponse = try {
+            call.save().response
+        } finally {
+            runCatching { response.rawContent.cancel() }
```
osipxd, line 39: Can't call `response.cleanup()` here because it would cancel the coroutine job

ktor-client/ktor-client-core/common/src/io/ktor/client/statement/HttpStatement.kt
```diff
@@ -158,7 +158,10 @@ public class HttpStatement(
     @OptIn(InternalAPI::class)
     internal suspend fun fetchResponse(): HttpResponse = unwrapRequestTimeoutException {
         val builder = HttpRequestBuilder().takeFromWithExecutionContext(builder)
+
         val call = client.execute(builder)
+        // Save the body again to make sure that it is replayable after pipeline execution
+        // We need this because wrongly implemented plugins could make response body non-replayable
         val result = call.save().response
```
osipxd, line 165: This is the place I'm concerned about. Waiting for the allocations benchmark to check if it adds extra allocations

bjhham: Unfortunately our allocation benchmarks only test the server atm.  You might need to introduce a client test then run it against main and here.

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/SaveBody.kt
osipxd: @bjhham, am I right that cancelling the channel should release the connection?

ktor-client/ktor-client-core/common/src/io/ktor/client/statement/HttpStatement.kt
osipxd: Oh, I've completely missed this. let me check how it works

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/SaveBody.kt
```diff
@@ -96,6 +133,17 @@ public val HttpResponse.isSaved: Boolean
...
  */
+@Deprecated(
+    "Skipping of body saving for a specific request is not allowed anymore.\n" +
+        "$SHARE_USE_CASE_MESSAGE\n" +
+        API_REMOVAL_MESSAGE
+)
+@Suppress("UnusedReceiverParameter")
 public fun HttpRequestBuilder.skipSavingBody() {
-    attributes.put(SKIP_SAVE_BODY, Unit)
+    LOGGER.warn("Skipping of body saving for a specific request is not allowed anymore. $SHARE_USE_CASE_MESSAGE")
```
e5l, line 143: If I have a 4 GB request, why can't I skip saving it?

e5l (changes requested):

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/SaveBody.kt
osipxd: It won't be saved if you're working with it as with a streaming request.

In fact, this function was never doing what it was supposed to do. Even if you were calling it on a request, saving was skipped in `SaveBodyPlugin` but then the request anyways was saved in `HttpStatement.fetchResponse`

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/SaveBody.kt
bjhham: Yes, it should.  I don't know how well it's covered with tests though...

ktor-client/ktor-client-core/common/src/io/ktor/client/statement/HttpStatement.kt
osipxd: Attached benchmark results in the PR description

e5l (approved):

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/SaveBody.kt
```diff
@@ -0,0 +1,169 @@
...
+
+    const val PLUGIN_DEPRECATED_MESSAGE =
+        "This plugin is not needed anymore.\n" +
+            `api will be removed`
+
+    const val SKIP_SAVING_BODY_MESSAGE =
+        "Skipping of body saving for a specific request is not allowed anymore.\n" +
+            `use streaming syntax` + "\n\n" +
+            `api will be removed` + "\n" +
+            `share use case`
```
osipxd, line 168: @vnikolova, could you review these deprecation messages and KDocs? 🙏

vnikolova (commented):
Just some minor grammatical adjustments, otherwise looks good!

bjhham (approved):

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/SaveBody.kt
```diff
@@ -0,0 +1,169 @@
...
+ */
+@Deprecated(SKIP_SAVING_BODY_MESSAGE)
+@Suppress("UnusedReceiverParameter")
+public fun HttpRequestBuilder.skipSavingBody() {
+    LOGGER.warn(SKIP_SAVING_BODY_MESSAGE)
+}
+
+@Suppress("ConstPropertyName")
+private object Messages {
+    private const val `use streaming syntax` =
```
FooIbar, line 142: This will break Android compilation, space characters are only supported on API 30+ and should only be used in tests.
https://kotlinlang.org/docs/coding-conventions.html#names-for-test-methods
https://source.android.com/docs/core/runtime/dex-format#simplename

PR #4814, opened 2025-04-28
Improved DI for nullables, parameterized types

bjhham:
**Subsystem**
Server, DI

**Motivation**
- [KTOR-8305](https://youtrack.jetbrains.com/issue/KTOR-8305) Optional dependencies and nullable type support
- [KTOR-8434](https://youtrack.jetbrains.com/issue/KTOR-8434) Dependency injection - type parameter covariance supertypes

**Solution**
Now, you can optionally resolve dependencies using nullables (like `resolve<Foo?>()`).  This applies only when the dependency is not provided, and will continue to fail if some other exception was thrown when trying to resolve it.

I've also introduced support for resolving covariant type parameter arguments, so declaring `MutableList<String>` will work with resolving `List<CharSequence>` by checking the type bounds on the supertypes.

I discovered during testing that it was very easy to hit ambiguity exceptions from these kinds of inferences, so I introduced a ranking system based on distance to the declared type, so we'll only fail when you have two declarations that are the same distance (like declaring `List<Int>` and `Set<Int>` and attempting to resolve `Collection<Int>`).

There were also a few tweaks to the exception handling so things are easier to debug.

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyConflictPolicy.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyConflictPolicy.kt
@@ -62,10 +62,12 @@ public sealed interface DependencyConflictResult {
  */
 public val DefaultConflictPolicy: DependencyConflictPolicy = DependencyConflictPolicy { prev, current ->
     require(current !is AmbiguousCreateFunction) { "Unexpected ambiguous function supplied" }
-    when (prev) {
-        is AmbiguousCreateFunction,
-        is ImplicitCreateFunction -> current.ifImplicit { Ambiguous } ?: KeepNew
-        is ExplicitCreateFunction -> current.ifImplicit { KeepPrevious } ?: Conflict
+    val diff = current.distance() - prev.distance()
+    when {
+        diff < 0 -> KeepNew
+        diff > 0 || current == prev -> KeepPrevious
+        prev is ExplicitCreateFunction -> Conflict
+        else -> Ambiguous
     }
 }
 
@@ -80,4 +82,10 @@ public val LastEntryWinsPolicy: DependencyConflictPolicy = DependencyConflictPol
         is ImplicitCreateFunction -> KeepNew
         is ExplicitCreateFunction -> current.ifImplicit { KeepPrevious } ?: KeepNew
     }
+}
+
+private fun DependencyCreateFunction.distance(): Int = when (this) {
+    is ExplicitCreateFunction -> -1
+    is ImplicitCreateFunction -> distance
+    is AmbiguousCreateFunction -> functions.first().distance()
 }
--- a/ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyInjection.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyInjection.kt
@@ -184,7 +184,7 @@ public class DependencyInjectionConfig {
      *                      By default, this throws a `DuplicateDependencyException`.
      */
     public data class ProviderScope(
-        public var keyMapping: DependencyKeyCovariance = Supertypes,
+        public var keyMapping: DependencyKeyCovariance = Supertypes * Nullables,
         public var conflictPolicy: DependencyConflictPolicy = DefaultConflictPolicy,
         public var onConflict: (DependencyKey) -> Unit = { throw DuplicateDependencyException(it) }
     )
--- a/ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyInjection.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyInjection.kt
@@ -184,7 +184,7 @@ public class DependencyInjectionConfig {
      *                      By default, this throws a `DuplicateDependencyException`.
      */
     public data class ProviderScope(
-        public var keyMapping: DependencyKeyCovariance = Supertypes * Nullables,
+        public var keyMapping: DependencyKeyCovariance = DefaultKeyCovariance,
         public var conflictPolicy: DependencyConflictPolicy = DefaultConflictPolicy,
         public var onConflict: (DependencyKey) -> Unit = { throw DuplicateDependencyException(it) }
     )
@@ -201,21 +201,25 @@ public data class DependencyKey(
 
     override fun toString(): String = buildString {
         append(type.kotlinType ?: type.type)
-        if (name != null || qualifier != null) {
-            if (name != null) {
-                append("name = \"$name\"")
-            }
-            if (qualifier != null) {
-                append("qualifier = \"$qualifier\"")
-            }
+        if (name != null) {
+            append("(\"$name\")")
         }
     }
 }
 
 /**
+ * Determines if the type associated with a `DependencyKey` is nullable.
+ *
+ * This function checks whether the `kotlinType` property of the `type` in the `DependencyKey`
+ * is marked as nullable. If there is no `kotlinType`, it will return `false`.
+ */
+public fun DependencyKey.isNullable(): Boolean =
+    type.kotlinType?.isMarkedNullable == true
+
+/**
  * Common parent for dependency injection problems.
  */
-public open class DependencyInjectionException(message: String, cause: Throwable? = null) :
+public open class DependencyInjectionException(message: String? = null, cause: Throwable? = null) :
     RuntimeException(message, cause)
 
 /**
--- a/ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyKeyCovariance.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyKeyCovariance.kt
@@ -6,25 +6,34 @@ package io.ktor.server.plugins.di
 
 import io.ktor.server.plugins.di.utils.hierarchy
 import io.ktor.server.plugins.di.utils.toNullable
+import io.ktor.server.plugins.di.utils.typeParametersHierarchy
 import io.ktor.util.reflect.TypeInfo
 import io.ktor.utils.io.InternalAPI
 
 /**
  * Functional interface for mapping a dependency key to its covariant types.
  *
- * For example, `ArrayList<String> -> [List<String>, Collection<String>]`
+ * For example, `ArrayList<String> -> [ArrayList<String>, List<String>, Collection<String>]`
  */
 public fun interface DependencyKeyCovariance {
-    public fun map(key: DependencyKey): List<DependencyKey>
+    public fun map(key: DependencyKey, distance: Int): Sequence<KeyMatch>
 }
+
+internal typealias KeyMatch = Pair<DependencyKey, Int>
+
+internal val KeyMatch.key: DependencyKey get() = first
+internal val KeyMatch.distance: Int get() = second
 
 /**
  * Standard covariance mapping for matching dependency keys to supertypes and implemented interfaces.
  */
 @OptIn(InternalAPI::class)
 public val Supertypes: DependencyKeyCovariance =
-    DependencyKeyCovariance { key ->
-        key.type.hierarchy().map { DependencyKey(it, key.name) }
+    DependencyKeyCovariance { key, start ->
+        var distance = start
+        key.type.hierarchy().map {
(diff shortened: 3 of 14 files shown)


ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyConflictPolicy.kt
```diff
@@ -62,10 +62,15 @@ public sealed interface DependencyConflictResult {
  */
 public val DefaultConflictPolicy: DependencyConflictPolicy = DependencyConflictPolicy { prev, current ->
     require(current !is AmbiguousCreateFunction) { "Unexpected ambiguous function supplied" }
-    when (prev) {
-        is AmbiguousCreateFunction,
-        is ImplicitCreateFunction -> current.ifImplicit { Ambiguous } ?: KeepNew
-        is ExplicitCreateFunction -> current.ifImplicit { KeepPrevious } ?: Conflict
+    val diff = current.distance() - prev.distance()
+    if (diff < 0) {
+        KeepNew
```
e5l, line 67: I would add trace logging for the decision here, so it can be clear why exactly the dependency is selected (or conflict or ambiguous)

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyKeyCovariance.kt
```diff
@@ -33,8 +50,35 @@ public val Supertypes: DependencyKeyCovariance =
...
+@OptIn(InternalAPI::class)
+public val Nullables: DependencyKeyCovariance =
+    DependencyKeyCovariance { key, distance ->
+        val nullableVariant = key.type.toNullable()
+            ?: return@DependencyKeyCovariance emptySequence()
+        sequenceOf(key.copy(type = nullableVariant) to distance.inc())
+    }
+
+@OptIn(InternalAPI::class)
+public val OutTypeArgumentsSupertypes: DependencyKeyCovariance =
```
e5l, line 76: I would also document it

e5l (approved):
lgtm, we might want to check the reflection is not heavy on frequent calls

ktor-server/ktor-server-plugins/ktor-server-di/jvm/test/io/ktor/server/plugins/di/utils/TypesTest.kt
```diff
@@ -20,24 +20,40 @@ class TypesTest {
...
         assertContains(supertypes, typeInfo<HashMap<String, Cat>>())
         assertContains(supertypes, typeInfo<Map<String, Cat>>())
     }
+
+    @Test
+    fun `parameterized arg supertypes`() {
+        val supertypes = typeInfo<List<Cat>>().typeParametersHierarchy().toSet()
+        assertContains(supertypes, typeInfo<List<Pet>>())
+        assertContains(supertypes, typeInfo<List<Animal>>())
+        assertContains(supertypes, typeInfo<List<Companion>>())
```
osipxd, line 49: I think we should also test ordering as we rely on it in `OutTypeArgumentsSupertypes`. So it is better to use `assertContentEquals` to test both elements and their order.

```suggestion
        val actualSupertypes = typeInfo<List<Cat>>().typeParametersHierarchy()
        val expectedSupertypes = sequenceOf(
            typeInfo<List<Animal>>(),
            typeInfo<List<Pet>>(),
            typeInfo<List<Companion>>(),
        )

        assertContentEquals(expectedSupertypes, actualSupertypes)
```

ktor-server/ktor-server-plugins/ktor-server-di/jvm/test/io/ktor/server/plugins/di/DependencyInjectionJvmTest.kt
```diff
@@ -62,48 +60,42 @@ class DependencyInjectionJvmTest {
...
+    fun `circular references from create`() = runTestDI {
+        assertFailsWith<CircularDependencyException> {
+            dependencies {
+                provide<WorkExperience> { WorkExperience(this.resolve()) }
+                provide<PaidWork> { PaidWork(this.resolve()) }
+                provide<List<PaidWork>> { listOf(this.resolve()) }
             }
+            val workExperience: WorkExperience = dependencies.create()
+            fail("This should fail but returned $workExperience")
         }
```
osipxd, line 72: Minor. We could wrap only `dependencies.create` with `assertFailWith` to make it clear that the failure occurs only when creating the dependency. 
```suggestion
        dependencies {
            provide<WorkExperience> { WorkExperience(this.resolve()) }
            provide<PaidWork> { PaidWork(this.resolve()) }
            provide<List<PaidWork>> { listOf(this.resolve()) }
        }

        assertFailsWith<CircularDependencyException> {
            dependencies.create<WorkExperience>()
        }
```

osipxd (commented):
Great job!
I have some suggestions regarding tests and some comments on the code.

ktor-server/ktor-server-plugins/ktor-server-di/jvm/test/io/ktor/server/plugins/di/utils/TypesTest.kt
osipxd: Btw my changed code won't pass the test because it lacks `typeInfo<List<Cat>>()` at the fist position. Should it be included?

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyInjection.kt
```diff
@@ -202,20 +202,18 @@ public data class DependencyKey(
...
         if (name != null || qualifier != null) {
-            if (name != null) {
-                append("name = \"$name\"")
-            }
-            if (qualifier != null) {
-                append("qualifier = \"$qualifier\"")
-            }
+            append(
+                listOfNotNull(name, qualifier)
+                    .joinToString(prefix = "(", postfix = ")") { "\"$it\"" }
```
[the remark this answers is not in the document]
bjhham: Yeah, I don't think it's a cause for concern here.  Maybe I could format the qualifier differently or just omit it though.  It's mostly just used for exceptions.

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyProvider.kt
```diff
@@ -71,16 +71,20 @@ public class ExplicitCreateFunction(
...
 }
 
 /**
  * Represents an implicitly registered dependency creation function that delegates to its explicit parent.
  *
  * @property origin The instance of [ExplicitCreateFunction] that this class delegates creation logic to.
+ * @property distance The distance from the original key.
  */
-public class ImplicitCreateFunction(public val origin: ExplicitCreateFunction) : DependencyCreateFunction {
+public data class ImplicitCreateFunction(
```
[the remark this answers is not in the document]
bjhham: Oh, maybe I'll just implement equals instead.

ktor-server/ktor-server-plugins/ktor-server-di/jvm/test/io/ktor/server/plugins/di/utils/TypesTest.kt
bjhham: Oh yeah that makes sense.

bjhham: I think I'll drop the `mapInclude` function and ensure that all covariant mapping functions include the input key.  That should simplify things a little.

bjhham: Not that this would affect this function... I guess it makes more sense to only include supertypes.

bjhham: Changed my mind again, the type parameters supertypes always includes the input because it's iterating over all combinations, so I ought to do the same with the hierarchy call 😄

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyConflictPolicy.kt
bjhham: Oh, there already is one level higher 😄

bjhham:
@e5l 
> we might want to check the reflection is not heavy on frequent calls

This is only called on startup, so there shouldn't be any concern there 👍

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyKeyCovariance.kt
```diff
@@ -33,8 +44,46 @@ public val Supertypes: DependencyKeyCovariance =
...
+            key.type.toNullable()?.let { nullableVariant ->
+                yield(key.copy(type = nullableVariant) to distance.inc())
+            }
+        }
+    }
+
+/**
+ * For parameterized types, returns all combinations of supertypes for `out` type parameters.
+ *
+ * For example, `Pair<String, Int>` may return `Pair<CharSequence
```
vnikolova, line 78: Looks like the example is incomplete.

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyKeyCovariance.kt
```diff
@@ -33,8 +44,46 @@ public val Supertypes: DependencyKeyCovariance =
...
+        sequence {
+            yield(key to distance)
+            key.type.toNullable()?.let { nullableVariant ->
+                yield(key.copy(type = nullableVariant) to distance.inc())
+            }
+        }
+    }
+
+/**
+ * For parameterized types, returns all combinations of supertypes for `out` type parameters.
```
vnikolova, line 76: **Suggestion (for improved readability):** A [DependencyKeyCovariance] that generates supertypes for `out` type arguments of parameterized types.

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyProvider.kt
```diff
@@ -95,20 +117,41 @@ public class ImplicitCreateFunction(public val origin: ExplicitCreateFunction) :
...
+ * @throws AmbiguousException Always thrown when attempting to create a dependency
  * through the [create] method.
  */
 public data class AmbiguousCreateFunction(
     public override val key: DependencyKey,
     val functions: Set<DependencyCreateFunction>
 ) : DependencyCreateFunction {
+    public companion object {
+        /**
+         * Instantiate a new AmbiguousCreateFunction, if the provided functions are unique.
```
vnikolova, line 129:
```suggestion
         * Instantiate a new [AmbiguousCreateFunction] if the provided functions are unique.
```

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyProvider.kt
```diff
@@ -95,20 +117,41 @@ public class ImplicitCreateFunction(public val origin: ExplicitCreateFunction) :
...
+         *
+         * This also will flatten any provided `AmbiguousCreateFunction`s.
```
vnikolova, line 131:
```suggestion
         * Flattens any provided [AmbiguousCreateFunction]s.
```

vnikolova (commented):
Just a few minor comment regarding KDocs.

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyProvider.kt
```diff
@@ -95,20 +117,41 @@ public class ImplicitCreateFunction(public val origin: ExplicitCreateFunction) :
  * @property key The key for the dependency that caused the ambiguity.
  * @property keys A set of possible matching keys that caused the ambiguity.
  *
- * @throws AmbiguousDependencyException Always thrown when attempting to create a dependency
+ * @throws AmbiguousException Always thrown when attempting to create a dependency
  * through the [create] method.
  */
 public data class AmbiguousCreateFunction(
```
osipxd, line 123: Should we also refactor other data classes?

osipxd (approved):
LGTM!

PR #4673, opened 2025-02-13
KTOR-7104 Fix saving caches with different vary header

marychatte:
**Subsystem**
Client, Cache

**Motivation**
[KTOR-7104](https://youtrack.jetbrains.com/issue/KTOR-7104)
In the issue it's said:
> Cache isn't updated when Vary header for 304 response matches but is not equal to Vary for 200 response (however, it's answering with cached response)

But actually, it shouldn't update the cache as well as it shouldn't respond with cached data because in documentation we can find the next:

> The same Vary header value should be used on all responses for a given URL, including [304](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/304) Not Modified responses and the "default" response.

PR #4816, opened 2025-04-29
Ktor 8345: Make VaryHeader check lenient

StefMa:
**Subsystem**
Client, Caching

**Motivation** & **Solution**
This PR fixes https://youtrack.jetbrains.com/issue/KTOR-8345/HttpCache-InvalidCacheStateException-thrown-when-Vary-header-has-different-entries-is-overly-severe

Let me explain:
All starts with [this](https://youtrack.jetbrains.com/issue/KTOR-7104).
Someone requested to update cache in case the Vary header differs.
Actually this violates the Vary header spec.
See https://developer.mozilla.org/de/docs/Web/HTTP/Reference/Headers/Vary
> The same Vary header value should be used on all responses for a given URL, including [304](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/304) Not Modified responses and the "default" response.

To make Ktor more compliant with the specs, [this PR](https://github.com/ktorio/ktor/pull/4673) was opened.

The change leads to throwing of an `InvalidCacheStateException` in case the Vary header differs from the original response (200) to the second response (304).
E.g.
GET 200 -> Vary: Accept-Encoding, Language
GET 304 -> Vary: Accept-Encoding --> `InvalidCacheStateException`

**The problem with that approach?**
We (as client developers) can't fix this on ourselves.
We just end up with yelling at our backend team (or even worse, third-party servers) to fix their Vary header.
That is all we can do, sadly.

This PR makes this check more lenient (again).
How it was before❗ 
But instead of running blindly into issues like [KTOR-7104](https://youtrack.jetbrains.com/issue/KTOR-7104), we now log the problem with a clear description that we return "the best matching cache" (which doesn't mean it is the correct one!).
(I'm not sure if this Logging is also propergated to users of the lib, or if its just internal. Let me know if this is wrong and if I have to adjust it.)

---

Honestly, I'm not sure if this is the best-fitting solution.
But I also think that we can't serve both problems - relying on the spec & make client developers happy.
Since the later is IMHO more important, I thought we can relax the ktor client here a bit.

What do you think?

### DIFF
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/cache/HttpCache.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/cache/HttpCache.kt
@@ -344,7 +344,7 @@ public class HttpCache private constructor(
     ): CachedResponseData? = when {
         varyKeys.isNotEmpty() -> {
             val cache = storage.find(url, varyKeys)
-            if (varyKeys.size != cache.varyKeys.size) {
+            if (cache != null && cache.varyKeys.size != varyKeys.size) {
                 LOGGER.info { "Vary header size differs! This is an issue according to the Vary header specs (https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Vary). " +
                     "However, we will return the best matching cache as it is not your fault. Report the server you communicate with to fix it." }
             }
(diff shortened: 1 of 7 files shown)


StefMa:
@JohannesTrussellRasch yep, that might be a solution. But also a bit overkill I think 🤔.
Another solution would also be to disable either caching at all, or at least for the affected endpoint...

But still I think ktor should be more lenient here.

osipxd (approved):
I have [the same question](https://github.com/ktorio/ktor/pull/4816#discussion_r2112146090) as the rabbit about "Falling back to missing cache logic". The rest looks good to me

PR #4819, opened 2025-04-30
KTOR-8455: Introduces the ability to preserve context with HttpClient.

sid-r-322:
**Subsystem**
HTTP client and config

**Motivation**
Today, when using ktor's HTTP client, there is no provision to preserve CoroutineContext. Following the completion of the actual HTTP call made in an `async` block, any context that we might want to preserve is lost. For instance, MDC context, which is vital to logging in applications, cannot be preserved after the HTTPClient completes.

**Solution**
The solution provides the ability to supply a coroutine context provider lambda when configuring the HTTPClient, via its configuration. This will be executed per request executed by the HTTP client, which would then inject the supplied CoroutineContext during the async call.


bjhham (changes requested):
I think would fit better as an extension to the client plugin interface for manipulating the call coroutine context.

Perhaps a custom client hook, so it can be called like this from a client plugin:

```kotlin
on(CreateCallContext) { context: CoroutineContext, requestData: HttpRequestData ->
    context + myCustomDebuggingContext(requestData)
}
```

It would be nice to have a test in here as well with some real-world example usage.

adrianliz:
I am experiencing the same issue, it has been resolved? @bjhham

Mr3zee:
> The solution provides the ability to supply a coroutine context provider lambda when configuring the HTTPClient, via its configuration. This will be executed per request executed by the HTTP client, which would then inject the supplied CoroutineContext during the async call.

Hi! This violates structured concurrency, and the fix should instead address the root cause. I will close the PR for now. I also updated the original issue with a reproducer and reopened it.

PR #4822, opened 2025-05-02
KTOR-8454 Expose the i18n to ApplicationCall

hfhbd:
**Subsystem**
Server, i18n jvm plugin

**Motivation**
https://youtrack.jetbrains.com/issue/KTOR-8454

**Solution**
Added `ApplicationCall.i18n(key: String): String`


ktor-server/ktor-server-plugins/ktor-server-i18n/jvm/test/io/ktor/tests/i18n/TranslationTest.kt
```diff
@@ -67,6 +68,29 @@ class TranslationTest {
         assertEquals("English Key", body)
     }
 
+    @Test
```
Stexxe, line 71: I think it's worth adding a test with `StatusPages` plugin

ktor-server/ktor-server-plugins/ktor-server-i18n/jvm/test/io/ktor/tests/i18n/TranslationTest.kt
```diff
@@ -87,6 +111,26 @@ class TranslationTest {
...
+    fun canUseApplicationCall() = testApplication {
+        install(I18n) {
+            availableLanguages = testAvailableLanguages
+        }
+
+        routing {
+            get("/") {
+                val value = (call as ApplicationCall).i18n("some_key")
```
bjhham, line 122: Appears to be an unnecessary cast here.

ktor-server/ktor-server-plugins/ktor-server-i18n/jvm/src/io/ktor/i18n/I18n.kt
```diff
@@ -62,3 +72,33 @@ public val I18n: ApplicationPlugin<I18nConfiguration> = createApplicationPlugin(
...
+ * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.i18n.locale)
+ */
+public val RoutingContext.locale: Locale get() = call.locale
+
+/**
+ * Get the current accepted [locale](java.util.Locale) specified in the HTTP request
+ *
+ * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.i18n.locale)
+ */
+public val ApplicationCall.locale: Locale get() = attributes[LOCALE]
```
bjhham, line 88: It's a little confusing having two properties on the routing context and call here.  Would probably be better to just have it on the call.

ktor-server/ktor-server-plugins/ktor-server-i18n/jvm/src/io/ktor/i18n/I18n.kt
```diff
@@ -49,10 +53,16 @@ public val I18n: ApplicationPlugin<I18nConfiguration> = createApplicationPlugin(
         val acceptedLanguages = it.request.acceptLanguageItems()
 
         val bestMatchLanguage = acceptedLanguages.firstOrNull {
-            availableLanguages.contains(it.value)
+            it.value in availableLanguages
         }?.value ?: defaultLanguage
 
         it.attributes.put(REQUIRED_RESPONSE_LANGUAGE, bestMatchLanguage)
```
bjhham, line 59: Minor: might as well use the operator here as well.

ktor-utils/common/src/io/ktor/util/Attributes.kt
```diff
@@ -97,6 +97,15 @@ public interface Attributes {
...
     public fun <T : Any> put(key: AttributeKey<T>, value: T)
 
+    /**
+     * Creates or changes an attribute with the specified [key] using [value]
+     *
+     * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.util.Attributes.set)
+     */
+    public operator fun <T: Any> set(key: AttributeKey<T>, value: T) {
+        put(key, value)
+    }
```
bjhham, line 107: I'm not sure why we didn't use the set operator here before, maybe the type argument wasn't supported in earlier Kotlin versions.

bjhham (changes requested):
Thanks for the PR!  Could you please target this to the `3.2.0-eap` branch and rebase your branch accordingly?

ktor-server/ktor-server-plugins/ktor-server-i18n/jvm/src/io/ktor/i18n/I18n.kt
hfhbd: Sure. The use-case to expose the locale is server side rendering, eg thymeleaf, but the ThymeLeafContent uses `locale = Locale.current` as default, so you need to overwrite it to user the client language/locale.

ktor-server/ktor-server-plugins/ktor-server-i18n/jvm/test/io/ktor/tests/i18n/TranslationTest.kt
Stexxe: It requires to test the new method

ktor-server/ktor-server-plugins/ktor-server-i18n/jvm/test/io/ktor/tests/i18n/TranslationTest.kt
hfhbd: Added

ktor-server/ktor-server-plugins/ktor-server-i18n/jvm/test/io/ktor/tests/i18n/TranslationTest.kt
hfhbd: The cast is not needed anymore after adding a test using StatusPages. I was too lazy...

ktor-utils/jsAndWasmShared/src/io/ktor/util/logging/LoggerJs.kt
```diff
@@ -28,5 +28,5 @@ public actual interface Logger {
     public actual fun trace(message: String, cause: Throwable)
 }
 
-public actual val Logger.isTraceEnabled: Boolean
-    get() = true
```
hfhbd, line 32: No idea why this implementation didn't check the log level but always returned `true`.

ktor-server/ktor-server-plugins/ktor-server-i18n/jvm/src/io/ktor/i18n/I18n.kt
```diff
@@ -62,3 +71,26 @@ public val I18n: ApplicationPlugin<I18nConfiguration> = createApplicationPlugin(
...
+public fun ApplicationCall.i18n(@PropertyKey(resourceBundle = BUNDLE_KEY) key: String): String {
+    val value = attributes[RESOURCE_BUNDLE].getString(key)
+
+    I18N_LOGGER.debug {
+        "translating to $locale: $key=$value"
+    }
+    return value
+}
+
+internal const val BUNDLE_KEY = "messages/messages"
```
hfhbd, line 96: This value is very uncommon and not recognized by IntelliJ, because Java prefers a class name and not a slash: https://docs.oracle.com/javase/7/docs/api/java/util/ResourceBundle.html#getBundle(java.lang.String,%20java.util.Locale,%20java.lang.ClassLoader,%20java.util.ResourceBundle.Control)

IMHO we should change the key to messages only. While we could make it configurable, IDE support requires a compile time constant.

bjhham (approved):
lgtm 👍

ktor-server/ktor-server-plugins/ktor-server-i18n/jvm/src/io/ktor/i18n/I18n.kt
hfhbd: I agree we need to keep the path but did you test it too? I always get errors regarding unresolved property keys 🤔

hfhbd: I did some more tests: Turns out, the "new" syntax does not require any files changes.

ktor-server/ktor-server-plugins/ktor-server-i18n/build.gradle.kts
```diff
@@ -7,3 +7,13 @@ description = "I18n support to Ktor"
...
     id("ktorbuild.project.server-plugin")
 }
+
+kotlin {
+    sourceSets {
+        jvmTest {
+            dependencies {
+                implementation(project(":ktor-server:ktor-server-plugins:ktor-server-status-pages"))
+            }
+        }
```
osipxd, line 17: @coderabbitai, remember that syntax `jvmTest.dependencies { ... }` is preferred over `jvmTest { dependencies { ... } }` as it gives the level of nesting lower by one.

PR #4824, opened 2025-05-03
Build: Store multiple Gradle CC entries

hfhbd:
**Subsystem**
Build, Gradle

**Motivation**
Gradle only stores the last CC entry be default. It should store multiple entries to faster switch between tasks, like `test` and `apiDump`.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -45,6 +45,7 @@ org.gradle.caching=true
 org.gradle.parallel=true
 org.gradle.configuration-cache=true
 org.gradle.configuration-cache.parallel=true
+org.gradle.configuration-cache.entries-per-key=100
 org.gradle.kotlin.dsl.allWarningsAsErrors=true
 
 # kotlin


osipxd (commented):
Thank you for the PR! Nice undocumented option :)
It's the first time I’ve seen this option, so I have some questions.

Could you explain why you chose 100 entries? I haven't found any documentation, but noticed that Gradle itself [uses only 2 entries per key](https://github.com/gradle/gradle/blob/c14190295deab5fe44892f7a21bb2add68464409/gradle.properties#L6). Won't 100 entries per key affect the size of the stored Configuration Cache? Am I understood it correctly that the use-case when this property should help is when we switch from one branch to another and there?

hfhbd:
@osipxd Good question. It is still undocumented because the impact is not fully tested: https://github.com/gradle/gradle/issues/28022#issuecomment-2697531280 (https://github.com/gradle/gradle/pull/31086). I use `100` for middle-size JVM projects, maybe Ktor needs a lower value (because it is a massive MPP project).

osipxd:
Thank you for the additional context! Let's start with `3` and check how it affects CC miss rate.
Could you also enable `org.gradle.configureondemand` in this branch? It should make CC entries smaller and more reusable.

osipxd (approved):

PR #4826, opened 2025-05-06
KTOR-6569 Add option for Bearer auth to not cache client token

e5l:
Fix [KTOR-6569](https://youtrack.jetbrains.com/issue/KTOR-6569) Bearer auth: Don't cache client bearer token (option)


osipxd:
We have the same problem in all other `AuthProvider` implementations using `AuthTokenHolder`. Could you take a look at KTOR-8180 as a possible solution?

e5l:
yep, will rebase

Mr3zee (changes requested):
Looks LLMish to me 🤔 Please ask your agents to consider comments

e5l:
Closed as obsolete

PR #4827, opened 2025-05-06
KTOR-5708 Add data url support

e5l:
Fix [KTOR-5708](https://youtrack.jetbrains.com/issue/KTOR-5708) Url class mangles data URLs

### DIFF
--- a/ktor-http/common/src/io/ktor/http/URLBuilder.kt
+++ b/ktor-http/common/src/io/ktor/http/URLBuilder.kt
@@ -161,12 +161,17 @@ private fun <A : Appendable> URLBuilder.appendTo(out: A): A {
         }
 
         "about" -> {
-            out.appendAbout(host)
+            out.appendPayload(host)
             return out
         }
 
         "tel" -> {
-            out.appendTel(host)
+            out.appendPayload(host)
+            return out
+        }
+
+        "data" -> {
+            out.appendPayload(host)
             return out
         }
     }
@@ -199,12 +204,7 @@ private fun Appendable.appendFile(host: String, encodedPath: String) {
     append(encodedPath)
 }
 
-private fun Appendable.appendAbout(host: String) {
-    append(":")
-    append(host)
-}
-
-private fun Appendable.appendTel(host: String) {
+private fun Appendable.appendPayload(host: String) {
     append(":")
     append(host)
 }
--- a/ktor-http/common/src/io/ktor/http/URLParser.kt
+++ b/ktor-http/common/src/io/ktor/http/URLParser.kt
@@ -46,6 +46,12 @@ internal fun URLBuilder.takeFromUnsafe(urlString: String): URLBuilder {
         startIndex += schemeLength + 1
     }
 
+    // Special handling for data URLs
+    if (protocol.name == "data") {
+        host = urlString.substring(startIndex, endIndex)
+        return this
+    }
+
     // Auth & Host
     val slashCount = count(urlString, startIndex, endIndex, '/')
     startIndex += slashCount
--- a/ktor-http/common/test/io/ktor/tests/http/UrlTest.kt
+++ b/ktor-http/common/test/io/ktor/tests/http/UrlTest.kt
@@ -1,3 +1,4 @@
+@file:Suppress("ktlint:standard:max-line-length")
 /*
  * Copyright 2014-2024 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
  */
@@ -5,7 +6,7 @@
 package io.ktor.tests.http
 
 import io.ktor.http.*
-import kotlin.random.*
+import kotlin.random.Random
 import kotlin.test.*
 
 class UrlTest {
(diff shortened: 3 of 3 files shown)


ktor-http/common/src/io/ktor/http/URLProtocol.kt
```diff
@@ -58,12 +58,19 @@ public data class URLProtocol(val name: String, val defaultPort: Int) : JvmSeria
          */
         public val SOCKS: URLProtocol = URLProtocol("socks", 1080)
 
+        /**
+         * Data URL protocol for embedded content.
+         *
+         * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.http.URLProtocol.Companion.DATA)
+         */
+        public val DATA: URLProtocol = URLProtocol("data", DEFAULT_PORT)
```
osipxd, line 66: This seems redundant

ktor-http/common/src/io/ktor/http/URLBuilder.kt
```diff
@@ -169,6 +169,11 @@ private fun <A : Appendable> URLBuilder.appendTo(out: A): A {
             out.appendTel(host)
             return out
         }
+
+        "data" -> {
+            out.appendAbout(host) // We use the same append format as about URLs
```
osipxd, line 174: I think we could rename `appendAbout` into something more general and remove `appendTel`

ktor-http/common/test/io/ktor/tests/http/DataUrlTest.kt
```diff
@@ -0,0 +1,46 @@
+/*
+ * Copyright 2014-2024 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.tests.http
+
+import io.ktor.http.*
+import kotlin.test.*
+
+class DataUrlTest {
```
osipxd, line 10: Probably we should add tests into `UrlTest` instead. And add at least one forth-and-back test verifying URL rendering works as expected.

osipxd (commented):
I know this is still a draft, but a few things caught my eye

ktor-http/common/src/io/ktor/http/URLBuilder.kt
e5l: This part is called payload in the spec, so I will rename it to appendPayload

ktor-http/common/test/io/ktor/tests/http/DataUrlTest.kt
e5l: thanks fixed

ktor-http/common/src/io/ktor/http/URLProtocol.kt
e5l: let's keep it for consistency

ktor-http/common/test/io/ktor/tests/http/DataUrlTest.kt
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright 2014-2024 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.tests.http
+
+import io.ktor.http.*
+import kotlin.test.*
+
+class DataUrlTest
```
marychatte, line 10: Should we move the new tests from `UrlTest.kt` to this new class `DataUrlTest.kt`?

.gitignore
```diff
@@ -13,7 +13,7 @@ build
 !.idea/icon.svg
 out
 *.iml
-.vscode
+.vscode`
```
marychatte, line 16: Is it a mistake?

marychatte (approved):
lgtm
one question: do we need to have CLAUDE.md?

ktor-http/common/src/io/ktor/http/URLProtocol.kt
osipxd: As I can see, `URLProtocol` lists only protocols, not any schema, so schemas like `tel`, `about` are not included here. Are you sure `data` should be?

ktor-http/common/src/io/ktor/http/URLBuilder.kt
```diff
@@ -161,12 +161,17 @@ private fun <A : Appendable> URLBuilder.appendTo(out: A): A {
...
         }
 
         "tel" -> {
-            out.appendTel(host)
+            out.appendPayload(host)
+            return out
+        }
+
+        "data" -> {
+            out.appendPayload(host) // We use the same append format as about URLs
```
osipxd, line 174: Minor. The comment seems redundant now.
```suggestion
            out.appendPayload(host)
```

.gitignore
osipxd: It looks like this PR needs to be rebased onto the latest main

e5l, line 16:
```suggestion
.vscode
```

ktor-http/common/src/io/ktor/http/URLProtocol.kt
e5l: Sorry, didn't spot. Let's drop it

osipxd (approved):
lgtm

PR #4830, opened 2025-05-06
KTOR-8446 Refactoring to improve IDE UX

bjhham:
**Subsystem**
Server, DI

**Motivation**
[KTOR-8446](https://youtrack.jetbrains.com/issue/KTOR-8446) Dependency injection - reduce extension reliance for the sake of UX

**Solution**
Converted the `DependencyRegistry` from an interface to a final class and moved the various extension functions into it.  Since this was using only a single implementation `DependencyRegistryImpl` and it was a façade of other easy-to-implement interfaces meant for extension, we won't be losing any flexibility here.

Now, practically all DI functionality can be resolved with a single automatic import `io.ktor.server.plugins.di.dependencies`.  Moving the invoke operator to a function with `@KtorDSL` also fixes the DSL highlighting when working with the API.


ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyRegistry.kt
```diff
@@ -76,44 +52,119 @@ public class DependencyRegistryImpl(
...
+    public fun validate() {
         for (key in requiredKeys) {
             resolver.value.get<Any>(key)
         }
     }
+
+    /**
+     * Get the dependency from the map for the key represented by the type (and optionally, with the given name).
+     */
+    public inline fun <reified T> resolve(key: String? = null): T =
```
osipxd, line 78: What if we keep both, an extension function and a member function (calling the first one from the second one)? It would keep `resolve<T>()` and `provide<T>()` usable with `DependencyResolver` without casting it to `DependencyRegistry`. Does it make sense?

osipxd: Btw, I'm not in the context of the issue, but I though that only `provideDelegate` is problematic.

nomisRev: the invoke operator on dependencies was worse than provide delegate imo. Invoke is not even suggested to import by IDE, luckily provideDelegate. In the current nightly invoke is also suggested.

@e5l mentioned during another call that the reason for using extension functions everywhere was to reduce the scope of the binary to avoid binary incompat. Tagging him here so he’s aware.

bjhham: Good call, I forgot how handy the compiler is with extension functions 👍

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyRegistry.kt
```diff
@@ -76,44 +52,98 @@ public class DependencyRegistryImpl(
...
+    public inline operator fun <reified T> provideDelegate(
+        thisRef: Any?,
+        prop: KProperty<*>
+    ): ReadOnlyProperty<Any?, T> {
+        val key = DependencyKey(typeInfo<T>())
+            .also(::require)
+        return ReadOnlyProperty { _, _ ->
+            get(key)
+        }
+    }
```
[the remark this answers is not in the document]
bjhham: No thanks.

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyProvider.kt
```diff
@@ -44,6 +44,14 @@ public interface DependencyProvider {
     public val declarations: Map<DependencyKey, DependencyCreateFunction>
 }
 
+/**
+ * Basic call for providing a dependency, like `provide<Service> { ServiceImpl() }`.
+ */
+public inline fun <reified T> DependencyProvider.provide(
+    name: String? = null,
+    noinline provide: DependencyResolver.() -> T
+): Unit = set(DependencyKey(typeInfo<T>(), name), provide)
```
e5l, line 53: please avoid using `=` for functions with Unit return type

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyRegistry.kt
```diff
@@ -9,53 +9,29 @@ import io.ktor.server.plugins.di.MutableDependencyMap.Companion.asResolver
...
-        attributes.put(DependencyRegistryKey, value)
-    }
-
-/**
- * Basic implementation of [DependencyRegistry], which is a façade of [DependencyProvider] and a late-initialized
- * [DependencyResolver].  The resolver is available after the first call to a get function is made, which triggers the
- * [DependencyResolution] process to populate the instances.
- */
-public class DependencyRegistryImpl(
+public class DependencyRegistry(
```
e5l, line 29: 👍

e5l (approved):

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyProvider.kt
osipxd: @coderabbitai, remember this rule

PR #4839, opened 2025-05-07
Merge 3.2.0-eap into main

bjhham:
Since we're targeting 3.2.0 as our next release, we can now merge the EAP branch into main 🎊


osipxd:
The only thing I'm worried about is hanging JS tests. I think we should either fix them before merging the PR or exclude them from the required checks.

e5l (approved):

osipxd (approved):

PR #4841, opened 2025-05-08
KTOR-8476 Fix for startup sequence for DI validation

bjhham:
**Subsystem**
Server, DI

**Motivation**
[KTOR-8476](https://youtrack.jetbrains.com/issue/KTOR-8476) DI validation occurs after "application started" message

**Solution**
Added a new event which is fired after modules are loaded and before `ApplicationStarted`.  The problem was that the logging for "Application started..." happens first on this event, so interrupting the startup sequence doesn't make sense here.


ktor-server/ktor-server-core/common/src/io/ktor/server/application/DefaultApplicationEvents.kt
```diff
@@ -19,6 +18,15 @@ import kotlin.native.concurrent.*
...
 public val ApplicationStarting: EventDefinition<Application> = EventDefinition()
 
+/**
+ * Event definition that is fired after `ApplicationStarting` hooks, and before `ApplicationStarted`.
+ *
+ * Hooks registered for this event can interrupt the application startup if needed by throwing exceptions.
+ *
+ * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.server.application.ApplicationModulesLoaded)
+ */
+public val ApplicationModulesLoaded: EventDefinition<Application> = EventDefinition()
```
e5l, line 28: We could also have granular events for every module. @nomisRev, what do you think?

e5l (approved):

ktor-server/ktor-server-core/common/src/io/ktor/server/application/DefaultApplicationEvents.kt
bjhham: For this particular case, we only care when _all_ the modules are loaded.  Hooks for specific modules could be interesting, though they're a little awkward to reference externally - I guess the API would be something like `monitor.subscribe(ModuleLoaded) { module -> if (module == Application::routing) }`.

PR #4842, opened 2025-05-08
KTOR-8477 Fix for provided nullables and flexible Java API types

bjhham:
**Subsystem**
Server, DI

**Motivation**
[KTOR-8477](https://youtrack.jetbrains.com/issue/KTOR-8477) DI provide lambda type fails for java function returns

**Solution**
Adding a ? to the lambda return ensures the reified type is the non-null / non-flexible type.


nomisRev (approved):
LGTM! 🙌

PR #4846, opened 2025-05-10
Better configurability for MockEngine

jonapoul:
**Subsystem**
Client testing

**Motivation**
I recently started using Ktor as a replacement for OkHttp for various reasons, and I came across MockEngine as an apparent replacement for [MockWebServer](https://github.com/square/okhttp/tree/master/mockwebserver). One thing that I did quite a lot with MockWebServer was to:
1. initialise the client in a test's `@Before` block
2. pass it into the rest of the system under test
3. specify HTTP behaviour on a per-test basis
4. make the call
6. verify the behaviour
7. maybe repeat steps 3-6 several times, depending on test complexity

But with MockEngine as it is I need to pass in an initial MockRequestHandler, otherwise an exception is thrown.

**Solution**
This change adds the option to intialise MockEngine without a MockRequestHandler using the `MockEngine.empty()` function - allowing us to enqueue one or more handlers later in the test lifecycle. If no handler exists at the time of a call being made, it still throws `error()` as it did before.

Obviously this isn't going to cover all user's needs so I left the default constructor behaving as it always has, but personally I think this will help people to structure their tests a little better.

CI seems to be failing, but I'm fairly sure(???) that's not the fault of this branch.

### DIFF
--- a/ktor-client/ktor-client-mock/api/ktor-client-mock.api
+++ b/ktor-client/ktor-client-mock/api/ktor-client-mock.api
@@ -1,22 +1,28 @@
-public final class io/ktor/client/engine/mock/MockEngine : io/ktor/client/engine/HttpClientEngineBase {
+public class io/ktor/client/engine/mock/MockEngine : io/ktor/client/engine/HttpClientEngineBase {
 	public static final field Companion Lio/ktor/client/engine/mock/MockEngine$Companion;
 	public fun <init> (Lio/ktor/client/engine/mock/MockEngineConfig;)V
-	public fun <init> (Lio/ktor/client/engine/mock/MockEngineConfig;Z)V
 	public fun close ()V
-	public final fun enqueue (Lkotlin/jvm/functions/Function3;)Z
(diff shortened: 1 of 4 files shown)


bjhham (commented):
Nice idea!  I think the introduction of the queue functions on the `MockEngine` does create a little dissonance in the abstraction since the engine is externally immutable.  I think it might make sense here to introduce something like `class MockEngine.Queue: MockEngine()` to expose the queue methods.

Then the test would look something like:

```kotlin
val engine = MockEngine.Queue()
assertTrue(engine.isEmpty())

val client = HttpClient(engine)

engine += { respondOk("hello") }
val response1 = client.get { url("http://127.0.0.1/normal-request") }
assertEquals("hello", response1.body<String>())

engine += { respondError(HttpStatusCode.BadRequest) }
val response2 = client.get { url("http://127.0.0.1/failed-request") }
assertEquals(HttpStatusCode.BadRequest, response2.status)

assertFailsWith<IllegalStateException> {
    client.get { url("http://127.0.0.1/no-more-handlers-registered") }
}
```

jonapoul:
Thats fair (I'll take a look a bit later today). Out of interest - is there any reason why MockEngine should have an immutable interface?

PR #4853, opened 2025-05-13
KTOR-8447 DI cleanup support

bjhham:
**Subsystem**
Server, DI

**Motivation**
[KTOR-8447](https://youtrack.jetbrains.com/issue/KTOR-8447) Dependency injection - cleanup support

**Solution**
- Added `onShutdown` property to `DependencyInjectionConfig` that closes any `AutoCloseable` declarations, or can be replaced to provide other general functionality.
- Added `cleanup` function to the registry for handling individual hooks for executing when the server shuts down.  See the additions in `DependencyInjectionTest.kt` for an example of how it looks in the DSL.

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyInjection.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyInjection.kt
@@ -67,6 +67,7 @@ public val DI: ApplicationPlugin<DependencyInjectionConfig> =
 
         val configMap = ConfigurationDependencyMap(application.environment.config)
         val dependenciesMap = pluginConfig.dependenciesMap?.let { it + configMap } ?: configMap
+        val onShutdown = pluginConfig.onShutdown
 
         var registry = DependencyRegistry(
             provider,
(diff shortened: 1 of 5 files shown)


ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyInjection.kt
```diff
@@ -159,13 +173,19 @@ public object NoReflection : DependencyReflection {
...
     public var reflection: DependencyReflection = DefaultReflection
+
     public var provider: DependencyProvider = MapDependencyProvider()
         set(value) {
             field = value
             providerChanged = true
         }
+
     public var resolution: DependencyResolution = DefaultDependencyResolution
+    public var onShutdown: (DependencyKey, Any?) -> Unit = { _, instance ->
```
nomisRev, line 186: We'll add `suspend` support at the same time we add it to `Events`?

nomisRev (approved):
Looking good to me!

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyInjection.kt
bjhham: Yeah, that's what I'm thinking.  Wish I could do it now, but it's a little too tricky 🥲

PR #4855, opened 2025-05-14
KTOR-8494 DI async support

bjhham:
**Subsystem**
Server, DI

**Motivation**
[KTOR-8494](https://youtrack.jetbrains.com/issue/KTOR-8494) DI: Support for async dependencies

**Solution**
This introduces the functions `provideAsync` and `resolveAwait` to support asynchronous resolution dependencies that rely on network connections or other suspend operations.  Since we're using a linked hash map behind the scenes, we always limit the parallelism in the dispatcher to 1 so the async calls will populate dependencies in sequence.  The registry's context is a child of the application's to ensure structured concurrency.


ktor-server/ktor-server-plugins/ktor-server-di/api/ktor-server-di.klib.api
```diff
@@ -328,15 +331,17 @@ open class io.ktor.server.plugins.di/DependencyInjectionException : kotlin/Runti
...
+open class io.ktor.server.plugins.di/MapDependencyProvider : io.ktor.server.plugins.di/DependencyProvider, kotlinx.coroutines/CoroutineScope { // io.ktor.server.plugins.di/MapDependencyProvider|null[0]
+    constructor <init>(io.ktor.server.plugins.di/DependencyKeyCovariance = ..., io.ktor.server.plugins.di/DependencyConflictPolicy = ..., kotlin/Function1<io.ktor.server.plugins.di/DependencyKey, kotlin/Unit> = ..., kotlin.coroutines/CoroutineContext = ..., io.ktor.util.logging/Logger = ...) // io.ktor.server.plugins.di/MapDependencyProvider.<init>|<init>(io.ktor.server.plugins.di.DependencyKeyCovariance;io.ktor.server.plugins.di.DependencyConflictPolicy;kotlin.Function1<io.ktor.server.plugins.di.DependencyKey,kotlin.Unit>;kotlin.coroutines.CoroutineContext;io.ktor.util.logging.Logger){}[0]
```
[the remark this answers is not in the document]
bjhham: This code hasn't been released yet, the breaking change is OK this time.

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyRegistry.kt
```diff
@@ -30,7 +34,8 @@ public class DependencyRegistry(
     private val external: DependencyMap,
     private val resolution: DependencyResolution,
     public override val reflection: DependencyReflection,
-) : DependencyProvider by provider, DependencyResolver {
+    public override val coroutineContext: CoroutineContext,
+) : DependencyProvider by provider, DependencyResolver, CoroutineScope {
```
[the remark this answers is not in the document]
bjhham: Added a cancel in the shutdown hook.

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyInjection.kt
```diff
@@ -117,6 +120,7 @@ public val DI: ApplicationPlugin<DependencyInjectionConfig> =
                         environment.log.error("Exception during cleanup for $key; continuing", e)
                     }
                 }
+                registry.cancel()
```
nomisRev, line 123: `cancelAndJoin`? 🤔

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyInjection.kt
```diff
@@ -74,6 +76,7 @@ public val DI: ApplicationPlugin<DependencyInjectionConfig> =
             dependenciesMap,
             pluginConfig.resolution,
             pluginConfig.reflection,
+            application.coroutineContext + Dispatchers.Default.limitedParallelism(1),
```
nomisRev, line 79: I'm a bit surprised by this change. This would allow dependencies to initialise without blocking Ktor start-up, but would not allow them to initialise in parallel by default?

What is the reason behind this?

ktor-server/ktor-server-plugins/ktor-server-di/jvm/src/io/ktor/server/plugins/di/utils/Types.jvm.kt
```diff
@@ -147,6 +149,9 @@ private val ignoredSupertypes = setOf(
     java.lang.Comparable::class,
     Enum::class,
     java.lang.Enum::class,
+    Job::class,
+    CoroutineContext::class,
```
nomisRev, line 153: This means I cannot `provide` a `CoroutineContext`? I should instead inject specific types?

I feel the same about named dependencies, I avoid them and create new types instead.

nomisRev (approved):
Looks good to me!

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyInjection.kt
bjhham: It's not in a suspend context here 😞 and it'll get joined with with the application anyway.

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyInjection.kt
bjhham: It's to ensure that dependencies are all loaded a single time and in the proper order.  If you've got a chain of async dependencies things can quickly get out of hand.  When we get a concurrent hashmap implementation in common Kotlin we can improve on this, but atm the only way to implement thread-safety would be with mutexes which would require everything to use suspend functions.

ktor-server/ktor-server-plugins/ktor-server-di/jvm/src/io/ktor/server/plugins/di/utils/Types.jvm.kt
bjhham: You can still do `provide<CoroutineContext> { CoroutineName("blah") }` for example, but this will prevent registering the implicit keys for these common interfaces, so if you have `provide { CoroutineName("blah") }` then you would get the free resolution for `resolve<CoroutineContext>`.  Note too that this reflective inference is only available on JVM.

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyRegistry.kt
```diff
@@ -104,13 +109,39 @@ public class DependencyRegistry(
...
+     * Shorthand for `provide<Deferred<T>> { async { suspendFunction() } }`
+     */
+    public inline fun <reified T> provideAsync(
+        name: String? = null,
+        noinline provide: suspend DependencyResolver.() -> T?
+    ): KeyContext<Deferred<T>> =
+        provide<Deferred<T>>(name) {
+            async {
+                provide() ?: throw DependencyInjectionException(
+                    "Null result for async declaration: ${DependencyKey<T>(name)}"
```
osipxd, line 140: So it can't be optional/nullable?

osipxd (approved):
👍

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyRegistry.kt
```diff
@@ -104,13 +109,39 @@ public class DependencyRegistry(
...
+     * Provide a deferred dependency for the given type using the provided suspend function.
+     *
+     * Shorthand for `provide<Deferred<T>> { async { suspendFunction() } }`
+     */
+    public inline fun <reified T> provideAsync(
+        name: String? = null,
+        noinline provide: suspend DependencyResolver.() -> T?
+    ): KeyContext<Deferred<T>> =
+        provide<Deferred<T>>(name) {
+            async {
```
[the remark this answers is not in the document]
bjhham: I think it'd be a bit dangerous to allow overriding the context here, since it could lead to threading problems.

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyRegistry.kt
bjhham: Nope, there's a bit of a catch 22 here where it's either impossible to resolve "flexible" return types from Java libraries, or you can allow nullable return types.  The correct way to provide a null for now is to just skip the provide call and `resolve<Thing?>`.

PR #4860, opened 2025-05-15
KTOR-8479 DI lambda overloads

bjhham:
**Subsystem**
Server, DI

**Motivation**
- [KTOR-8479](https://youtrack.jetbrains.com/issue/KTOR-8479) DI lambda overloads
- [KTOR-8322](https://youtrack.jetbrains.com/issue/KTOR-8322) Dependency injection - handle delegate pattern

**Solution**
- Added overloads for lambdas so you can use constructor and function references easily - akin to the [support provided in Koin](https://github.com/InsertKoinIO/koin/blob/eb9303c5cf1b4bfad8e81a760e5fac9af2cf5621/projects/core/koin-core/src/commonMain/kotlin/org/koin/core/module/dsl/SingleOf.kt#L25).
- Added logic to resolve ambiguity while evaluating providers to allow for delegate types.  This works by using any single previously registered implementation, while the ambiguous exception is preserved if called from some other source.
- Also updated the testing support to use the first declaration instead, because this makes more sense especially when one of the production modules activates the validation.  I plan to provide some more support for unit testing in another PR.


ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyLambdaSupport.kt
```diff
@@ -0,0 +1,150 @@
+/*
+ * Copyright 2014-2025 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.server.plugins.di
+
+import io.ktor.utils.io.KtorDsl
+
+/**
```
bjhham, line 9: Truly my finest work 😅

ktor-server/ktor-server-plugins/ktor-server-di/api/ktor-server-di.api
```diff
@@ -18,7 +21,8 @@ public final class io/ktor/server/plugins/di/AmbiguousCreateFunction$Companion {
 }
 
 public final class io/ktor/server/plugins/di/AmbiguousDependencyException : io/ktor/server/plugins/di/DependencyInjectionException {
-	public fun <init> (Lio/ktor/server/plugins/di/DependencyKey;Ljava/util/Collection;)V
+	public fun <init> (Lio/ktor/server/plugins/di/AmbiguousCreateFunction;)V
+	public final fun getFunction ()Lio/ktor/server/plugins/di/AmbiguousCreateFunction;
 }
```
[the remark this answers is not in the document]
osipxd: This version is not released yet, so it is okay to make binary incompatible changes.

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyConflictPolicy.kt
```diff
@@ -72,15 +72,13 @@ public val DefaultConflictPolicy: DependencyConflictPolicy = DependencyConflictP
 }
 
 /**
- * During testing, we simply override previously declared values.
+ * During testing, we ignore conflicts.
  * This allows for replacing base implementations with mock values.
```
osipxd, line 76: > This allows for replacing base implementations with mock values.

Is this still true?

bjhham: Yes... Sorta... but the wording isn't great... The idea is that now you declare your dependencies, then call your modules.  The problem with adding them at the end is that you could have a module reading in a dependency prematurely.  I'm going to revisit this in another ticket that I reopened [KTOR-8306](https://youtrack.jetbrains.com/issue/KTOR-8306) to try to get the best UX for testing here.

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyLambdaSupport.kt
osipxd: A masterpiece 👍

osipxd (approved):
👍

PR #4866, opened 2025-05-19
KTOR-8519 Fix malformed message during startup

bjhham:
**Subsystem**
Server, Logging

**Motivation**
[KTOR-8519](https://youtrack.jetbrains.com/issue/KTOR-8519) Bad log message on server startup

**Solution**
Fix the small mistake in the code which lead to the weird log message.


ktor-server/ktor-server-core/jvm/src/io/ktor/server/engine/EmbeddedServerJvm.kt
```diff
@@ -337,7 +337,7 @@ actual constructor(
             engine.resolvedConnectors().forEach {
                 val host = escapeHostname(it.host)
                 environment.log.info(
-                    "Responding at ${it.type.name.lowercase()}:$it"
+                    "Responding at ${it.type.name.lowercase()}://$host:${it.port}"
```
osipxd, line 340: I think it was made to print the correct message for unix-socket connections. Should we properly implement toString instead?

PR #4874, opened 2025-05-21
Fix some ProGuard warnings

osipxd:
**Subsystem**
Client/Server

**Motivation**
[KTOR-8525](https://youtrack.jetbrains.com/issue/KTOR-8525) ktor-network produces ProGuard warning

**Solution**
The warning was caused by a lambda created inside the inline function. I've moved the lambda to the class field.
Also fixed another warning on the recently changed `combine(String, Any)` function

### DIFF
--- a/ktor-io/common/src/io/ktor/utils/io/ByteWriteChannelOperations.kt
+++ b/ktor-io/common/src/io/ktor/utils/io/ByteWriteChannelOperations.kt
@@ -143,6 +143,10 @@ public val ChannelJob.isCancelled: Boolean get() = job.isCancelled
 @OptIn(InternalCoroutinesApi::class)
 public fun ChannelJob.getCancellationException(): CancellationException = job.getCancellationException()
 
+public fun ChannelJob.invokeOnCompletion(block: (cause: Throwable?) -> Unit): DisposableHandle =
+    job.invokeOnCompletion(block)
+
+@Deprecated("Maintained for binary compatibility", level = DeprecationLevel.HIDDEN)
 public fun ChannelJob.invokeOnCompletion(block: () -> Unit) {
     job.invokeOnCompletion { block() }
 }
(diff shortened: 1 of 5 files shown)


ktor-server/ktor-server-core/common/src/io/ktor/server/config/MapApplicationConfig.kt
```diff
@@ -196,8 +195,8 @@ internal class MapApplicationConfigValue(
     }
 }
 
-private fun combine(root: String, relative: Any): String =
-    if (root.isEmpty()) relative.toString() else "$root.$relative"
+private fun combine(root: String, relative: Int): String = combine(root, relative.toString())
+private fun combine(root: String, relative: String): String = if (root.isEmpty()) relative else "$root.$relative"
```
osipxd, line 199: Investigating why changing the signature from `combine(String, String)` to `combine(String, Any)` caused ProGuard warning...

ktor-io/common/src/io/ktor/utils/io/ByteWriteChannelOperations.kt
```diff
@@ -143,6 +143,10 @@ public val ChannelJob.isCancelled: Boolean get() = job.isCancelled
 @OptIn(InternalCoroutinesApi::class)
 public fun ChannelJob.getCancellationException(): CancellationException = job.getCancellationException()
 
+public fun ChannelJob.invokeOnCompletion(block: (cause: Throwable?) -> Unit): DisposableHandle =
+    job.invokeOnCompletion(block)
+
+@Deprecated("Maintained for binary compatibility", level = DeprecationLevel.HIDDEN)
 public fun ChannelJob.invokeOnCompletion(block: () -> Unit) {
     job.invokeOnCompletion { block() }
 }
```
osipxd, line 152: Changed this to avoid one extra lambda creation.

bjhham (approved):
👍

PR #4875, opened 2025-05-21
KTOR-8162: Generate version catalogs

osipxd:
**Subsystem**
Infrastructure

**Motivation**
[KTOR-8162](https://youtrack.jetbrains.com/issue/KTOR-8162) Support Version Catalog

**Solution**
Added a new project `ktor-version-catalog`, which generates a catalog containing all the published projects.

There are two parts I don't like:
1. Gradle doesn't provide a lazy API for defining a version catalog, so we have to use hacks.
2. I didn't come up with a good way of normalizing of library alias. Currently, it is normalized using a map and it is error-prone if we add a new project requiring alias normalization.

The TOML file is provided only for review. I will delete if before merging.


ktor-version-catalog/build/version-catalog/libs.versions.toml
```diff
@@ -0,0 +1,125 @@
...
+client-curl = {group = "io.ktor", name = "ktor-client-curl", version.ref = "ktor" }
+client-darwin = {group = "io.ktor", name = "ktor-client-darwin", version.ref = "ktor" }
+client-darwin-legacy = {group = "io.ktor", name = "ktor-client-darwin-legacy", version.ref = "ktor" }
+client-encoding = {group = "io.ktor", name = "ktor-client-encoding", version.ref = "ktor" }
+client-gson = {group = "io.ktor", name = "ktor-client-gson", version.ref = "ktor" }
+client-ios = {group = "io.ktor", name = "ktor-client-ios", version.ref = "ktor" }
+client-jackson = {group = "io.ktor", name = "ktor-client-jackson", version.ref = "ktor" }
+client-java = {group = "io.ktor", name = "ktor-client-java", version.ref = "ktor" }
+client-jetty = {group = "io.ktor", name = "ktor-client-jetty", version.ref = "ktor" }
+client-jetty-jakarta = {group = "io.ktor", name = "ktor-client-jetty-jakarta", version.ref = "ktor" }
```
osipxd, line 32: Should we assign alias `client-jetty` to `ktor-client-jetty-jakarta` and create something like `client-jetty-legacy` for `ktor-client-jetty`?

ktor-version-catalog/build/version-catalog/libs.versions.toml
```diff
@@ -0,0 +1,125 @@
...
+server-tomcat = {group = "io.ktor", name = "ktor-server-tomcat", version.ref = "ktor" }
+server-tomcat-jakarta = {group = "io.ktor", name = "ktor-server-tomcat-jakarta", version.ref = "ktor" }
+server-velocity = {group = "io.ktor", name = "ktor-server-velocity", version.ref = "ktor" }
+server-webjars = {group = "io.ktor", name = "ktor-server-webjars", version.ref = "ktor" }
+server-websockets = {group = "io.ktor", name = "ktor-server-websockets", version.ref = "ktor" }
+sse = {group = "io.ktor", name = "ktor-sse", version.ref = "ktor" }
+testDispatcher = {group = "io.ktor", name = "ktor-test-dispatcher", version.ref = "ktor" }
+utils = {group = "io.ktor", name = "ktor-utils", version.ref = "ktor" }
+websocket-serialization = {group = "io.ktor", name = "ktor-websocket-serialization", version.ref = "ktor" }
+websockets = {group = "io.ktor", name = "ktor-websockets", version.ref = "ktor" }
```
osipxd, line 121: We can also create an alias `websockets-serialization` pointing to `ktor-websocket-serialization`. (Probably this project was named without `s` at the end by a mistake)

ktor-version-catalog/build.gradle.kts
```diff
@@ -0,0 +1,84 @@
...
+        "server-http-redirect" to "server-httpRedirect",
+        "server-method-override" to "server-methodOverride",
+        "server-partial-content" to "server-partialContent",
+        "server-rate-limit" to "server-rateLimit",
+        "server-request-validation" to "server-requestValidation",
+        "server-status-pages" to "server-statusPages",
+        "server-test-host" to "server-testHost",
+        "test-dispatcher" to "testDispatcher",
+    )
+}
```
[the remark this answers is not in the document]
osipxd: I'd also like to use an algorithmic approach, but it there is one problem. Not all aliases should be translated to camel case. For example `server-caching-headers` should be, because `cachingHeaders` is a plugin name. But `server-metrics-micrometer` shouldn't because we also have `server-metrics` and hierarchy makes sense here.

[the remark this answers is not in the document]
osipxd: I've managed to fix this in recent commits

ktor-version-catalog/build/version-catalog/libs.versions.toml
bjhham: Yeah I'd support that.  The shortened artifact name being the deprecated one is confusing enough for folks, this ought to make the choice a little easier.  Would also need to be applied for the server engine.

bjhham (approved):
Nice!  Do you think we ought to advise using the version catalog for new projects and/or update the project generator?

osipxd:
> Do you think we ought to advise using the version catalog for new projects and/or update the project generator?

Yes. It was the next step in my plan — to update the generator.

PR #4884, opened 2025-05-26
KTOR-8005 Async module support

bjhham:
**Subsystem**
Server, Core

**Motivation**
[KTOR-8005](https://youtrack.jetbrains.com/issue/KTOR-8005) Allow suspend Ktor modules

**Solution**
Added suspend to the modules and function arguments.

This unfortunately breaks bytecode compatibility, however adding overloads breaks compilation due to an ambiguity error.  I could move the new `embeddedServer` functions to a different package, but maybe this will be fine since it's unlikely we'll have libraries wrapping these functions.


ktor-server/ktor-server-core/jsAndWasmShared/src/io/ktor/server/engine/EmbeddedServer.jsAndWasmShared.kt
```diff
@@ -71,7 +73,7 @@ actual constructor(
 
     public actual fun start(wait: Boolean): EmbeddedServer<TEngine, TConfiguration> {
         addShutdownHook { stop() }
-        prepareToStart()
+        serverScope.launch { prepareToStart() }
         engine.start(wait)
         return this
```
bjhham, line 78: This is a bit of a problem, because we can't block in JS-land the engine can start before the modules are loaded.

ktor-server/ktor-server-test-host/api/ktor-server-test-host.api
```diff
@@ -35,7 +35,7 @@ public final class io/ktor/server/testing/TestApplication : io/ktor/server/testi
 
 public class io/ktor/server/testing/TestApplicationBuilder {
 	public fun <init> ()V
-	public final fun application (Lkotlin/jvm/functions/Function1;)V
+	public final fun application (Lkotlin/jvm/functions/Function2;)V
```
e5l, line 38: should we add deprecated hidden here as well?

ktor-server/ktor-server-test-host/common/src/io/ktor/server/testing/TestApplication.kt
```diff
@@ -284,7 +284,7 @@ public open class TestApplicationBuilder {
      * @see [testApplication]
      */
     @KtorDsl
-    public fun application(block: Application.() -> Unit) {
+    public fun application(block: suspend Application.() -> Unit) {
```
e5l, line 287: here

ktor-server/ktor-server-test-host/api/ktor-server-test-host.api
bjhham: I was thinking we don't need to worry about bytecode incompatibility for the test host since it's only used as a direct import for testing.  But now that you mention it, it doesn't hurt to include it 😄

PR #4886, opened 2025-05-27
KTOR-8259 Make CurlMultiApiHandler.perform non-blocking

osipxd:
**Subsystem**
`ktor-client-curl`

**Motivation**
[KTOR-8259](https://youtrack.jetbrains.com/issue/KTOR-8259) Linux curl engine doesn't work for simultaneous websocket and http request

**Solution**
The root cause was in `CurlMultiApiHandler.perform` implementation. It was waiting for `transfersRunning == 0` to exit and for websocket connections it happend only after they're closed. So `drainRequestQueue` was never called again until websocket connection is closed.

Another problem was in `cancelRequest`. It was calling `curl_multi_remove_handle` instead of `handleCompleted` and handle could be removed twice from a `multiHandle`. I removed this call and added `handleCompleted` call in `close` to ensure all completed and cancelled requests are handled.

Also enabled some flaky tests to check if they're still flaky: [KTOR-7847](https://youtrack.jetbrains.com/issue/KTOR-7847), [KTOR-7885](https://youtrack.jetbrains.com/issue/KTOR-7885)

It's better to review commit by commit.

### DIFF
--- a/ktor-client/ktor-client-curl/desktop/src/io/ktor/client/engine/curl/internal/CurlMultiApiHandler.kt
+++ b/ktor-client/ktor-client-curl/desktop/src/io/ktor/client/engine/curl/internal/CurlMultiApiHandler.kt
@@ -9,8 +9,8 @@ import io.ktor.utils.io.*
 import io.ktor.utils.io.core.*
 import io.ktor.utils.io.locks.*
 import kotlinx.cinterop.*
-import kotlinx.coroutines.*
-import kotlinx.io.*
+import kotlinx.coroutines.CompletableDeferred
+import kotlinx.io.readByteArray
 import libcurl.*
 
 private class RequestHolder @OptIn(ExperimentalForeignApi::class) constructor(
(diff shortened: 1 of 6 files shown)


ktor-client/ktor-client-curl/desktop/src/io/ktor/client/engine/curl/internal/CurlMultiApiHandler.kt
```diff
@@ -130,27 +130,22 @@ internal class CurlMultiApiHandler : Closeable {
...
+        synchronized(easyHandlesToUnpauseLock) {
+            var handle = easyHandlesToUnpause.removeFirstOrNull()
+            while (handle != null) {
+                curl_easy_pause(handle, CURLPAUSE_CONT)
+                handle = easyHandlesToUnpause.removeFirstOrNull()
+            }
+        }
+        curl_multi_perform(multiHandle, transfersRunning.ptr).verify()
+        if (transfersRunning.value != 0) {
+            curl_multi_poll(multiHandle, null, 0.toUInt(), 100, null).verify()
```
osipxd, line 145: Is this timeout value okay? It directly affects how fast HTTP request will be sent when we have established websocket connection so I've reduced it from 10 to 0.1 sec.

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/HttpTimeoutTest.kt
```diff
@@ -213,11 +208,8 @@ class HttpTimeoutTest : ClientLoader() {
 
     @Test
     fun testGetRequestTimeoutWithSeparateReceive() = clientTests(except("Js"), retries = 5) {
-//      https://youtrack.jetbrains.com/issue/KTOR-7847/Investigate-Flaky-timeout-tests-on-linuxX64
-        if (PlatformUtils.IS_NATIVE) return@clientTests
```
osipxd, line 217: Tried to rerun tests several times on CI and they always were successful. So removed this `if`

bjhham (approved):
Nice work 🎉

ktor-client/ktor-client-curl/desktop/src/io/ktor/client/engine/curl/internal/CurlMultiApiHandler.kt
e5l: I would extract this to env variable, just in case :)

PR #4893, opened 2025-05-29
KTOR-8215 Expose Application instance from testApplication

osipxd:
**Subsystem**
Test infrastructure

**Motivation**
[KTOR-8215](https://youtrack.jetbrains.com/issue/KTOR-8215) Application instance access in testApplication
[KTOR-7607](https://youtrack.jetbrains.com/issue/KTOR-7607) Add a way to create an `ApplicationCall` for testing

**Solution**
Added a public field `application` to `TestApplication` and `ApplicationTestBuilder`.


bjhham (approved):
It might cause a bit of confusion when you have something like:

```kotlin
testApplication {
    fun Application.module() {
        attributes[key] = "foo"
    }
    application { module() }  // add the module
    assertNotNull(attributes[key]) // fails because module is not applied until server starts
}
```

I've got no ideas on how to resolve this conundrum though.  The lack of access to the application instance is a bit of an annoyance, so even without some strategy to mitigate the confusion here, it's a good improvement.

PR #4895, opened 2025-05-29
KTOR-6508: Skip Content-Length header for GET, HEAD and OPTIONS

osipxd:
**Subsystem**
Client

**Motivation**
[KTOR-6508](https://youtrack.jetbrains.com/issue/KTOR-6508/The-Content-Length-0-header-is-added-for-GET-requests-sent-to-some-servers) The "Content-Length: 0" header is added for GET requests sent to some servers

**Solution**
The root cause is that we add "Content-Length" header in `mergeHeaders` function if `contentLength != null` without checking HTTP method. The problem is we don't have an access to HTTP method there, so we can solve this in two ways:
1. Introduce a new method in place of `mergeHeaders` that will be aware of HTTP method and use it.
2. Handle this corner case in every placy we use `mergeHeaders`.

For now I've implemented the second option, while I'm thinking about good API for the first option.


ktor-client/ktor-client-core/common/src/io/ktor/client/request/HttpRequest.kt
```diff
@@ -249,6 +248,25 @@ public class HttpRequestData @InternalAPI constructor(
...
+ * and calls [block] for each header.
+ *
+ * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.client.request.forEachHeader)
+ *
+ * @see mergeHeaders
+ */
+@InternalAPI
+public inline fun HttpRequestData.forEachHeader(
+    crossinline block: (key: String, value: String) -> Unit
+) {
```
osipxd, line 262: New API here

ktor-http/common/src/io/ktor/http/HttpMethod.kt
```diff
@@ -52,3 +54,16 @@ public data class HttpMethod(val value: String) {
...
+    HttpMethod.Head,
+    HttpMethod.Options,
+)
+
+/**
+ * Returns `true` if this request method can have a request body.
+ */
+@InternalAPI
+public val HttpMethod.supportsRequestBody: Boolean
+    get() = this !in REQUESTS_WITHOUT_BODY
```
osipxd, line 69: And here

bjhham (approved):
Nice 👍

PR #4897, opened 2025-05-30
Webrtc client for Android, JS and WasmJs targets

zibet27:
**Subsystem**
Client

**Motivation**

**Solution**
The solution contains a common KMP interface that allows 
- Create WebRTC Clients
- Establish peer connections 
- Send audio and video tracks (tracks are tightly bound to the protocol, so they were also included)
- Listen for ICE candidates, remote tracks, and connection status changes
- Obtain continuous connection statistics, etc

The API is "up to Ktor standards" and utilises flows and coroutines rather than callbacks. It is unified across those platforms (configs are custom for all engines), but still allows access to native objects if some custom logic is needed.

There are integration tests that 
- Use Karma (which is already run on the CI) for JS and Wasm srcsets (establish a real connection, but uses mock tracks)
- Require emulator (androidInstrumentedTests) for Android and use emulated camera and audio, so it could be problematic to run it on the CI. To make it unit tests, we can mock the Android API here, but the getstream-webrtc build, which is used under the hood, has only native libraries compiled for Android arm.

TODO:
- Add [datachannels](https://developer.mozilla.org/en-US/docs/Web/API/WebRTC_API/Using_data_channels) to the API, which could be useful for some applications.
- Add docs and more comments.

Example project that uses AndroidWebRTCEngine: [ktor-chat](https://github.com/ktorio/ktor-chat/pull/6)

### DIFF
--- a/kotlin-js-store/yarn.lock
+++ b/kotlin-js-store/yarn.lock
@@ -82,6 +82,11 @@
     "@jridgewell/resolve-uri" "^3.1.0"
     "@jridgewell/sourcemap-codec" "^1.4.14"
 
+"@js-joda/core@3.2.0":
+  version "3.2.0"
+  resolved "https://registry.yarnpkg.com/@js-joda/core/-/core-3.2.0.tgz#3e61e21b7b2b8a6be746df1335cf91d70db2a273"
+  integrity sha512-PMqgJ0sw5B7FKb2d5bWYIoxjri+QlW/Pys7+Rw82jSH0QN3rB05jZ/VrrsUdh1w4+i2kw9JOejXGq/KhDOX7Kg==
+
 "@puppeteer/browsers@1.8.0":
   version "1.8.0"
   resolved "https://registry.yarnpkg.com/@puppeteer/browsers/-/browsers-1.8.0.tgz#fb6ee61de15e7f0e67737aea9f9bab1512dbd7d8"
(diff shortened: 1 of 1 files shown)


build-settings-logic/settings.gradle.kts
```diff
@@ -2,6 +2,14 @@
  * Copyright 2014-2024 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
  */
 
+pluginManagement {
+    repositories {
+        google()
+        mavenCentral()
+        gradlePluginPortal()
+    }
+}
```
osipxd, line 11: I think this `pluginManagement` block can be removed as we don't apply any plugins from custom repositories to the `build-settings-logic` project

gradle.properties
```diff
@@ -67,6 +67,7 @@ kotlin.daemon.useFallbackStrategy=false
 # Enable new project model to be prepared for enabling isolated projects
 # TODO: Remove when we enable isolated projects in Gradle
 kotlin.kmp.isolated-projects.support=enable
+android.useAndroidX=true
```
osipxd, line 70: Minor. Let's separate it from Kotlin properties with blank line and comment
```suggestion

# Android
android.useAndroidX=true
```

ktor-client/ktor-client-webrtc/build.gradle.kts
```diff
@@ -0,0 +1,90 @@
...
+    id("com.android.library")
+    id("kotlinx-serialization")
+    id("ktorbuild.project.library")
+}
+
+kotlin {
+    jvmToolchain(17)
+
+    androidTarget {
+        publishLibraryVariants("release")
```
osipxd, line 21: This could be moved to the place where we declare `androidTarget` the first time, as it will be needed for any android module.

osipxd (commented):
For now I left only comments for infrastructure tings. But I should check it again after migration to KMP Android plugin

zibet27:
> oooook 😅 before I even start to review this, may I ask you to split this PR into at least 4 others: Infra changes, Android support, JS, WasmJs. If there is an overlap between those 3 targets, then also Core (so 5 PRs). Then in each of them, please make clean git history, so that the changes can be reviewed commit by commit. The last point is not very strict, so don't bother yourself too much, but nonetheless keep in mind that commit history should easily readable.

Sure, I agree it was too much code inside one PR!
I partitioned it into common, js, wasmJs and android parts, so I will close this PR. I also added some documentation.
- [Common API + tests](https://github.com/ktorio/ktor/pull/4902)
- [JS Engine](https://github.com/ktorio/ktor/pull/4903)
- [WasmJS Engine](https://github.com/ktorio/ktor/pull/4904)
- [Android Engine](https://github.com/ktorio/ktor/pull/4905)

zibet27:
This PR was split into 4 smaller ones.

PR #4901, opened 2025-06-03
KTOR-8528 Fix race condition when Netty removes headers for some resp…

Stexxe:
To avoid the race condition, we will wait for Netty to finish discarding certain headers like `Content-Length` for informational and 204 responses.
https://youtrack.jetbrains.com/issue/KTOR-8528


ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationResponse.kt
```diff
@@ -60,12 +64,26 @@ public abstract class NettyApplicationResponse(
...
 
     override suspend fun responseChannel(): ByteWriteChannel {
         val channel = ByteChannel()
         val chunked = headers[HttpHeaders.TransferEncoding] == "chunked"
         sendResponse(chunked, content = channel)
+
+        if (isInfoOrNoContentStatus()) {
+            sendCompleted.await()
+        }
+
```
bjhham: Looks like you could avoid the extra deferred here and just use the call's response job:

```
(call as NettyApplicationCall).responseWriteJob.join()
```

Maybe change the constructor arg to `val engineCall: NettyApplicationCall` to avoid the cast.

Is the logic here necessary for the "info / no content"?  I think we could drop the if clause to avoid any unknown scenarios where the netty engine might manipulate headers on it's own, and it would seem best practice to wait for netty to say the response is ready before we continue onto callbacks.

Stexxe: I've removed the extra deferred object. However, I would like to keep the logic of checking for info/no content because such synchronization could decrease the performance of ordinary responses.

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationResponse.kt
```diff
@@ -60,12 +61,30 @@ public abstract class NettyApplicationResponse(
...
         val channel = ByteChannel()
         val chunked = headers[HttpHeaders.TransferEncoding] == "chunked"
         sendResponse(chunked, content = channel)
+
+        if (isInfoOrNoContentStatus()) {
+            if (call is NettyApplicationCall) {
+                (call as NettyApplicationCall).responseWriteJob.join()
+            }
+        }
+
```
bjhham: Could you extract a function for this block and add KDoc to explain why this is necessary?  This is pretty non-local control so it's hard to understand the intent at face value.

bjhham (changes requested):

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationResponse.kt
Stexxe: I can add a comment to both places. Is it enough?

e5l (commented):
Could you explain in the PR body why the response headers were removed?

bjhham (approved):

PR #4902, opened 2025-06-03
ktor-client-webrtc module common API + tests

zibet27:
**Subsystem**
Client

**Motivation**

**Solution**
The solution contains a common KMP interface that allows

- Create WebRTC Clients
- Establish peer connections
- Send audio and video tracks (tracks are tightly bound to the protocol, so they were also included)
- Listen for ICE candidates, remote tracks, and connection status changes
- Obtain continuous connection statistics, etc

The API is "up to Ktor standards" and utilises flows and coroutines rather than callbacks. It is unified across those platforms (configs are custom for all engines), but still allows access to native objects if some custom logic is needed. There are integration tests. A mock JS target has been added so that Gradle can include the module in the build. Code documentation contains links to MDN.

TODO:
Add [datachannels](https://developer.mozilla.org/en-US/docs/Web/API/WebRTC_API/Using_data_channels) to the API.

This is a part of a previously created [pull request](https://github.com/ktorio/ktor/pull/4897).


ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRTC.kt
```diff
@@ -0,0 +1,269 @@
...
+
+import kotlinx.serialization.Serializable
+
+/**
+ * An object containing WebRTC protocol entities and abstractions.
+ * Provides the core types and interfaces needed for WebRTC peer-to-peer communication.
+ *
+ * @see <a href="https://developer.mozilla.org/en-US/docs/Web/API/WebRTC_API">MDN WebRTC API</a>
+ */
+public object WebRTC {
```
Mr3zee: `WebRtc` by Kotlin style guide

ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRTC.kt
```diff
@@ -0,0 +1,269 @@
...
+     * @property dtmf The DTMF sender associated with this RTP sender.
+     * @property track The media track being sent.
+     *
+     * @see <a href="https://developer.mozilla.org/en-US/docs/Web/API/RTCRtpSender">MDN RTCRtpSender</a>
+     */
+    public interface RtpSender {
+        public val dtmf: DtmfSender?
+        public val track: WebRTCMedia.Track?
+
+        public fun getNative(): Any
```
Mr3zee: What is this method?

ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRTCEngine.kt
```diff
@@ -0,0 +1,90 @@
...
+    public val config: WebRTCConfig
+    public val dispatcher: CoroutineDispatcher
+
+    /**
+     * Creates a new peer connection with the configured settings.
+     */
+    public suspend fun createPeerConnection(): WebRtcPeerConnection
+}
+
+internal expect fun ioDispatcher(): CoroutineDispatcher
```
Mr3zee: Don't we have anything like already ready in Ktor?

ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRTCEngine.kt
```diff
@@ -0,0 +1,90 @@
...
+
+/**
+ * Core engine interface for WebRTC functionality.
+ *
+ * Provides the ability to create peer connections and media tracks.
+ * Implementations of this interface handle the platform-specific WebRTC operations.
+ */
+public interface WebRTCEngine : CoroutineScope, Closeable, MediaTrackFactory {
+    public val config: WebRTCConfig
+    public val dispatcher: CoroutineDispatcher
```
Mr3zee: Why the need for both dispatcher and scope?

ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRTCMedia.kt
```diff
@@ -0,0 +1,157 @@
...
+
+        /**
+         * Enables or disables the track.
+         */
+        public fun enable(enabled: Boolean)
+
+        /**
+         * Gets the native platform-specific object representing this track.
+         */
+        public fun getNative(): Any
```
Mr3zee: Should where be expect/actual declaration?

ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRTCPeerConnection.kt
```diff
@@ -0,0 +1,173 @@
...
+ *
+ * @see <a href="https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection">MDN RTCPeerConnection</a>
+ */
+public abstract class WebRtcPeerConnection(iceCandidatesReplay: Int, remoteTracksReplay: Int) : Closeable {
+    protected val currentStats: MutableStateFlow<List<WebRTC.Stats>> = MutableStateFlow(listOf())
+
+    /**
+     * Flow of connection statistics.
+     */
+    public val statsFlow: StateFlow<List<WebRTC.Stats>> = currentStats.asStateFlow()
```
Mr3zee: I'd switch the names and put `Flow` suffix into internal properties.

Mr3zee (approved):
I only have minor issues, as I'm not closely familiar with the protocol itself

ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRTCEngine.kt
zibet27: Yes, we have, but I am not sure if we want to depend on the other module just for one function.

ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRTC.kt
zibet27: Returns the native platform-specific object representing RtpSender. I should add a comment here.

ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRTCMedia.kt
zibet27: A common target doesn't declare any implementations. They are declared in the specific targets. Also, there is no implicit default engine selection.

ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRTCEngine.kt
zibet27: Actually, we don't need a dispatcher field, it is used just to initialise a coroutine context.

ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRTC.kt
e5l: please also add kdocs for all other public API

ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRTCEngine.kt
```diff
@@ -0,0 +1,90 @@
...
+
+    public var iceServers: List<WebRTC.IceServer> = emptyList()
+    public var turnServers: List<WebRTC.IceServer> = emptyList()
+
+    /**
+     * Refresh rate of [WebRTC.Stats] in ms.
+     * Value less than or equal to zero means there will be no statistics collected.
+     * Defaults to -1.
+     * */
+    public var statsRefreshRate: Long = -1
```
e5l: please extract a constant for -1

ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRTCEngine.kt
e5l: we can introduce this in the kotlinx-io module under internal API and unify usages

ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRTCPeerConnection.kt
```diff
@@ -0,0 +1,173 @@
...
+     * The remote session description for this connection.
+     */
+    public abstract val remoteDescription: WebRTC.SessionDescription?
+
+    protected var negotiationNeededCallback: () -> Unit = {}
+
+    /**
+     * Returns platform-specific peer connection used under the hood. Use with caution.
+     */
+    public abstract fun getNativeConnection(): Any
```
e5l: Could we have expect-actual types instead of any? or at least generic parameter to have some type safety

ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRTCPeerConnection.kt
```diff
@@ -0,0 +1,173 @@
...
+     * The local session description for this connection.
+     */
+    public abstract val localDescription: WebRTC.SessionDescription?
+
+    /**
+     * The remote session description for this connection.
+     */
+    public abstract val remoteDescription: WebRTC.SessionDescription?
+
+    protected var negotiationNeededCallback: () -> Unit = {}
```
e5l: I would avoid mutable callbacks, it looks like leaking implementation detail

ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRTCPeerConnection.kt
```diff
@@ -0,0 +1,173 @@
...
+
+/**
+ * Base class for operations performed on media tracks.
+ */
+public sealed class Operation<T>(public val item: T)
+
+/**
+ * Represents an add operation for a media track.
+ */
+public class Add<T>(item: T) : Operation<T>(item)
```
e5l: Could you tell me if we have any other operation types? I would prefer using an enum instead of sealed class here

ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRTCPeerConnection.kt
zibet27: Returned types are platform-specific. We could add generics for all getNative() methods.
Is it better to return `T?` or `T` and throw an exception if the cast can't be performed?

ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRTCPeerConnection.kt
zibet27: Individual track could be also muted, unmuted, ended (not the same as removed) in the browser, but we can't subscribe for such events in the Android implementation. Should we add these events to common interface?
PS: I think it worth renaming Operation to TrackEvent, also making it an enum.

ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRTCPeerConnection.kt
zibet27: Do you think we should pass it to the constructor then?

e5l: Could you tell me if we need to change them during runtime?

zibet27: Depends on the implementation; usually, it doesn't need to be changed.

ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRTCEngine.kt
zibet27: Do you mean the ktor-io module or the kotlinx-io library?

e5l: ktor-io or even ktor-utils

ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRTCPeerConnection.kt
zibet27: Sorry, I forgot about another use case. It could be useful to initialise it later, in this case we can make it `lateinit var`. For example, in tests:

`client.createPeerConnection().use { pc1 ->`
`            client.createPeerConnection().use { pc2 ->`
`                val negotiationNeededCnt = atomic(0)`
`                pc1.onNegotiationNeeded { negotiationNeededCnt.incrementAndGet() }`
`               pc2.onNegotiationNeeded { negotiationNeededCnt.incrementAndGet() }`

Or in the user code:

`val connection = client.createPeerConnection()`
`manager= PeerManager(connection, .....)` 
`....`
`manager.setupListeners(scope)`  <- adds onNegotionNeeded callback, listens for stats  and other flows

ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRtc.kt
```diff
@@ -0,0 +1,277 @@
...
+        ALL,
+        RELAY
+    }
+
+    /**
+     * Represents the RTCP mux policy for the connection.
+     *
+     * @see <a href="https://developer.mozilla.org/en-US/docs/Web/API/RTCConfiguration/rtcpMuxPolicy">MDN rtcpMuxPolicy</a>
+     */
+    public enum class RTCPMuxPolicy {
```
bjhham:
```suggestion
    public enum class RtcpMuxPolicy {
```

ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRtc.kt
```diff
@@ -0,0 +1,277 @@
...
+     * Interface for sending DTMF (Dual-Tone Multi-Frequency) tones.
+     *
+     * @property toneBuffer The tone buffer containing the tones to be played.
+     * @property canInsertDTMF Whether DTMF tones can be inserted.
+     *
+     * @see <a href="https://developer.mozilla.org/en-US/docs/Web/API/RTCDTMFSender">MDN RTCDTMFSender</a>
+     */
+    public interface DtmfSender {
+        public val toneBuffer: String
+        public val canInsertDTMF: Boolean
```
bjhham:
```suggestion
        public val canInsertDtmf: Boolean
```

ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRtc.kt
```diff
@@ -0,0 +1,277 @@
...
+        public fun insertDTMF(tones: String, duration: Int, interToneGap: Int)
```
bjhham:
```suggestion
        public fun insertDtmf(tones: String, duration: Int, interToneGap: Int)
```

ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRtcClient.kt
```diff
@@ -0,0 +1,37 @@
...
+package io.ktor.client.webrtc
+
+import io.ktor.utils.io.*
+
+/**
+ * A multiplatform asynchronous WebRtc client for establishing peer-to-peer connections, managing media tracks.
+ */
+public class WebRtcClient(engine: WebRtcEngine) : WebRtcEngine by engine
+
+public interface WebRtcClientEngineFactory<out T : WebRtcConfig> {
```
bjhham: Needs doc and more fun imo.

```suggestion
/**
 * Client engine factory interface, used in engine implementations.
 *
 * @param T expected engine-specific configuration class
 */
public fun interface WebRtcClientEngineFactory<out T : WebRtcConfig> {
```

ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRtcEngine.kt
```diff
@@ -0,0 +1,96 @@
...
+/**
+ * Configuration for WebRtc connections.
+ *
+ * Provides settings for ICE servers, policies, and other parameters needed for WebRtc connections.
+ * @see <a href="https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/RTCPeerConnection#configuration">MDN RTCPeerConnection config</a>
+ */
+@KtorDsl
+public open class WebRtcConfig {
+    public var mediaTrackFactory: MediaTrackFactory? = null
+    public var dispatcher: CoroutineDispatcher? = null
```
bjhham: Looks like these public vars will need docs.

ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRtc.kt
```diff
@@ -0,0 +1,277 @@
...
+
+        /**
+         * @return Native platform-specific object representing this DtmfSender.
+         */
+        public fun <T> getNative(): T
```
bjhham: Looks like this type variable could be part of the interface since the implementer knows its type at compile time.  From what I can tell, it could be fairly common to access this when rendering to the UI, so it might be worth the effort to refactor.  The type variable would have to get passed all the way up to the client interface.

So in it's current state, it'd end up looking something like this:

```kotlin
class WebRtcClient<T: Track, Dtmf: DtmfSender, Rtp: RtpSender<Dtmf, Track>>
```

This would seem a bit ridiculous, but maybe we can remove the native referencing from the interfaces for `DtmfSender` and `RtpSender` which would make it:

```kotlin
class WebRtcClient<T: Track> {
    public suspend fun createPeerConnection(): WebRtcPeerConnection<T>
}
```

Then if you need to cast on the senders, you can cast the containing instance which would just have a property of the native implementation instead of the interface function.

ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRtc.kt
```diff
@@ -0,0 +1,277 @@
...
+     * Interface representing parameters for RTP transmission.
+     *
+     * @property transactionId The transaction ID for these parameters.
+     * @property codecs The codecs used for transmission.
+     * @property rtcp The RTCP parameters.
+     * @property headerExtensions The header extensions for the RTP packets.
+     * @property degradationPreference The degradation preference for the media quality.
+     * @property encodings The encoding parameters for the media.
+     *
+     * @see <a href="https://developer.mozilla.org/en-US/docs/Web/API/RTCRtpParameters">MDN RTCRtpParameters</a>
```
bjhham: I think these links ought to be markdown format as opposed to HTML tags.

ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRtcPeerConnection.kt
```diff
@@ -0,0 +1,173 @@
...
+ */
+public abstract class WebRtcPeerConnection(iceCandidatesReplay: Int, remoteTracksReplay: Int) : Closeable {
+    protected val statsFlow: MutableStateFlow<List<WebRtc.Stats>> = MutableStateFlow(listOf())
+
+    /**
+     * Flow of connection statistics.
+     */
+    public val stats: StateFlow<List<WebRtc.Stats>> = statsFlow.asStateFlow()
+
+    protected val iceCandidatesFlow: MutableSharedFlow<WebRtc.IceCandidate> = MutableSharedFlow(iceCandidatesReplay)
```
bjhham: Maybe we can move all these protected flows to the top of the class for readability.  Or better yet, you could merge the mutable flows into a single type and using delegation, so you can have something like `WebRtcPeerConnection(protected val events: MutableWebRtcConnectionEvents): WebRtcConnectionEvents by events` where the `WebRtcConnectionEvents` interface just has the public flow properties.

bjhham (commented):
Very nice progress!

ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRtcEngine.kt
```diff
@@ -0,0 +1,96 @@
...
+ * Provides common functionality for WebRtc engine implementations.
+ *
+ * @param engineName Name identifier for the engine, used in coroutine naming.
+ */
+public abstract class WebRtcEngineBase(private val engineName: String) : WebRtcEngine {
+
+    override val coroutineContext: CoroutineContext by lazy {
+        val dispatcher = config.dispatcher ?: ioDispatcher()
+        dispatcher + CoroutineName("$engineName-context")
+    }
```
bjhham: It's generally better to use delegation when implementing `CoroutineScope` and creating an instance via `CoroutineScope()`.

> Manual implementation of this interface is not recommended, implementation by delegation should be preferred instead. By convention, the context of a scope should contain an instance of a job to enforce the discipline of structured concurrency with propagation of cancellation.

The problem here is that it's missing an instance of `Job()` in the context so child jobs won't be bound to it for structured concurrency.

ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRtc.kt
zibet27: What do you think of the following approach?
We can remove all mentions of native implementations in the common srcset and add extension functions for different platforms.

public class JsWebRtcPeerConnection(
    internal val nativePeerConnection: RTCPeerConnection,
    coroutineContext: CoroutineContext,
    config: WebRtcConnectionConfig,
) : WebRtcPeerConnection(coroutineContext, config) {...}

// Js target, the same could be applied for every target
public fun WebRtcPeerConnection.getNative(): RTCPeerConnection {
    val conn = this as? JsWebRtcPeerConnection ?: error("Only JsWebRtcPeerConnection implementation is supported.")
    return conn.nativePeerConnection
} 

// So users don't need to pass any generics or do any casts.
// common srcset doesn't know about any internal type
actual fun useConnection(connection: WebRtcPeerConnection) {  
    connection.getNative().someNativeMethod() 
}

bjhham: Yeah, good thinking!

PR #4904, opened 2025-06-03
wasmjs webrtc engine for ktor-client-webrtc

zibet27:
**Subsystem**
Client

**Motivation**

**Solution**
The solution implements WasmJS WebRTC Engine. Files in the browser directory are generated by [Dukat](https://github.com/Kotlin/dukat) and just declare the browser WebRTC API. There is a lot of code that almost copies the JS engine behaviour, but can't be reused (moved to jsWasmShared target group) because JS and Wasm types are different.

### DIFF
--- a/ktor-client/ktor-client-webrtc/js/src/io/ktor/client/webrtc/utils/JsUtils.kt
+++ b/ktor-client/ktor-client-webrtc/js/src/io/ktor/client/webrtc/utils/JsUtils.kt
@@ -20,27 +20,27 @@ internal inline fun <T : Any> jsObject(init: T.() -> Unit): T {
 }
 
 internal fun WebRtcMedia.AudioTrackConstraints.toJs(): MediaTrackConstraints {
-    return MediaTrackConstraints(
-        volume = volume,
-        latency = latency,
-        sampleRate = sampleRate,
-        sampleSize = sampleSize,
-        echoCancellation = echoCancellation,
-        autoGainControl = autoGainControl,
-        noiseSuppression = noiseSuppression,
-        channelCount = channelCount,
-    )
+    return jsObject {
+        volume = this@toJs.volume
+        latency = this@toJs.latency
+        sampleRate = this@toJs.sampleRate
+        sampleSize = this@toJs.sampleSize
+        echoCancellation = this@toJs.echoCancellation
+        autoGainControl = this@toJs.autoGainControl
+        noiseSuppression = this@toJs.noiseSuppression
+        channelCount = this@toJs.channelCount
+    }
 }
 
 internal fun WebRtcMedia.VideoTrackConstraints.toJs(): MediaTrackConstraints {
-    return MediaTrackConstraints(
-        width = width,
-        height = height,
-        aspectRatio = aspectRatio,
-        facingMode = facingMode?.toJs(),
-        frameRate = frameRate,
-        resizeMode = resizeMode?.toJs(),
-    )
+    return jsObject {
+        width = this@toJs.width
+        height = this@toJs.height
+        aspectRatio = this@toJs.aspectRatio
+        facingMode = this@toJs.facingMode?.toJs()
+        frameRate = this@toJs.frameRate
+        resizeMode = this@toJs.resizeMode?.toJs()
+    }
 }
 
 /**
--- a/ktor-client/ktor-client-webrtc/wasmJs/src/io/ktor/client/webrtc/utils/JsUtils.kt
+++ b/ktor-client/ktor-client-webrtc/wasmJs/src/io/ktor/client/webrtc/utils/JsUtils.kt
@@ -25,27 +25,27 @@ internal inline fun <T : JsAny> jsObject(init: T.() -> Unit): T {
 }
 
 internal fun WebRtcMedia.AudioTrackConstraints.toJs(): MediaTrackConstraints {
-    return MediaTrackConstraints(
-        volume = volume?.toJsNumber(),
-        latency = latency?.toJsNumber(),
-        sampleRate = sampleRate?.toJsNumber(),
-        sampleSize = sampleSize?.toJsNumber(),
-        echoCancellation = echoCancellation?.toJsBoolean(),
-        autoGainControl = autoGainControl?.toJsBoolean(),
-        noiseSuppression = noiseSuppression?.toJsBoolean(),
-        channelCount = channelCount?.toJsNumber(),
-    )
+    return jsObject {
+        volume = this@toJs.volume?.toJsNumber()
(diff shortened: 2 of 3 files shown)


ktor-client/ktor-client-webrtc/wasmJs/src/io/ktor/client/webrtc/Engine.wasmJs.kt
```diff
@@ -0,0 +1,157 @@
...
+ */
+public class WasmJsWebRtcPeerConnection(
+    private val nativePeerConnection: RTCPeerConnection,
+    override val coroutineContext: CoroutineContext,
+    private val statsRefreshRate: Long,
+    iceCandidatesReplay: Int,
+    remoteTracksReplay: Int
+) : CoroutineScope, WebRtcPeerConnection(iceCandidatesReplay, remoteTracksReplay) {
+    init {
+        nativePeerConnection.onicecandidate = { event: RTCPeerConnectionIceEvent ->
```
e5l: Could you tell me if we can make an interface and provide its implementation to the nativePeerConnection (for instance like here: https://en.wikipedia.org/wiki/Strategy_pattern) instead of initializing all the fields manually?

e5l:
Huge work! Thank you for the PR, can't wait to see the demo in action

ktor-client/ktor-client-webrtc/wasmJs/src/io/ktor/client/webrtc/Engine.wasmJs.kt
zibet27: If I understand the `Strategy` pattern correctly, it's used to interchange different behaviours triggered by the same event (calling the method with the same name, etc), but here is the opposite situation. We want to provide the same behaviour triggered by "different" events. Though the current implementation duplicates some code like `launch { state.emit(newConnectionState) }` and statistics polling. What do you think about adding `PeerConnectionEventHandler` interface, implementing one handler in the `WebRtcPeerConnection` abstract class and using this handler across all implementations, so we could need to make MutableFlows private. 
Though for static polling, we could use the Strategy pattern, introduce the `StatsProvider` interface and implement it for different peer connections.

interface PeerConnectionEventHandler {
    fun onIceCandidate(candidate: WebRtc.IceCandidate)
    fun onConnectionStateChange(newState: WebRtc.ConnectionState)
    fun onAddTrack(event: WebRtcMedia.Track)
    fun onNegotiationNeeded()
    ...
}

public class WasmJsWebRtcPeerConnection(
    ....
) : CoroutineScope, WebRtcPeerConnection(iceCandidatesReplay, remoteTracksReplay) {
    init {
        nativePeerConnection.onicecandidate = { event: RTCPeerConnectionIceEvent ->
            event.candidate?.let { super.handler.onIceCandidate(it.toCommon()) }
        }
        
        // init other events
    }        
}

interface StatsProvider {
   fun getStatistics(): WebRtc.Stats
}
// omit StatsProvider implementations

ktor-client/ktor-client-webrtc/wasmJs/src/io/ktor/client/webrtc/MediaDevices.kt
```diff
@@ -0,0 +1,71 @@
...
+    override suspend fun createAudioTrack(constraints: WebRtcMedia.AudioTrackConstraints): WebRtcMedia.AudioTrack =
+        withPermissionException("audio") {
+            val streamConstrains = MediaStreamConstraints(audio = constraints.toJS())
+            val mediaStream = navigator.mediaDevices.getUserMedia(streamConstrains).await<MediaStream>()
+            return WasmJsAudioTrack(mediaStream.getAudioTracks()[0]!!, mediaStream)
+        }
+
+    override suspend fun createVideoTrack(constraints: WebRtcMedia.VideoTrackConstraints): WebRtcMedia.VideoTrack =
+        withPermissionException("video") {
+            val streamConstrains = MediaStreamConstraints(audio = constraints.toJS())
```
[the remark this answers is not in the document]
Mr3zee: Nice catch Copilot!

ktor-client/ktor-client-webrtc/wasmJs/src/io/ktor/client/webrtc/RtpSender.kt
```diff
@@ -0,0 +1,76 @@
...
+            nativeSender.setParameters(parameters.nativeRtpParameters).await<JsUndefined>()
+        }
+    }
+}
+
+public class WasmJsDtmfSender(private val nativeSender: RTCDTMFSender) : WebRTC.DtmfSender {
+    override val toneBuffer: String
+        get() = nativeSender.toneBuffer.toString()
+    override val canInsertDTMF: Boolean
+        get() = nativeSender.canInsertDTMF.toBoolean()
```
Mr3zee: Why getters and not lazy? ca the prop from JS change between invokations?

ktor-client/ktor-client-webrtc/wasmJs/src/io/ktor/client/webrtc/browser/LibDom.kt
```diff
@@ -0,0 +1,65 @@
...
+package io.ktor.client.webrtc.browser
+
+import org.w3c.dom.mediacapture.MediaDevices
+
+public typealias JsUndefined = JsAny?
+
+public open external class DOMException {
+    public val code: JsNumber
+    public val message: JsString
+    public val name: JsString
```
Mr3zee, line 16: Can it be in jsAndWasmShared module? Not sure how it works in commonization

Mr3zee (changes requested):
Please review my comments from `js` impl and apply them to this review too. Great job!

ktor-client/ktor-client-webrtc/wasmJs/src/io/ktor/client/webrtc/browser/LibDom.kt
zibet27: I would like to make them shared, so we don't need to duplicate a lot of code, but it's not possible at the moment. See https://youtrack.jetbrains.com/issue/KT-64214

turansky: Issue isn't blocker for you in fact.
All required base types we already provide in Kotlin Wrappers (`kotlin-js` and `kotlin-js-core` subprojects).

turansky:
Common (for JS and WasmJS) declarations for WebRTC will be available in Kotlin Wrappers, when https://github.com/JetBrains/kotlin-wrappers/issues/2449 will be fixed (probably this week).
@zibet27 can assist ;)

PR #4905, opened 2025-06-03
android webrtc engine + introduced Android-KMP gradle plugin

zibet27:
**Subsystem**
Client

**Motivation**

**Solution**
The solution implements Android WebRTC Engine. It is based on [webrtc-android](https://github.com/GetStream/webrtc-android), which depends on Android API (like Camera, Audio), so Android Native target couldn't be used. Instead, the [Android-KMP plugin](https://developer.android.com/kotlin/multiplatform/plugin#features) was added and configured. The Android team recommends using this one instead of the [old Gradle plugin](https://developer.android.com/build/releases/gradle-plugin) for KMP projects. The Android JVM target is not added until the Android-KMP plugin is applied.

Tests are running on the device (Android emulator) because even if we mock all Android API, org.webrtc uses a native library which is only compiled for Android devices.

### DIFF
--- a/build-logic/build.gradle.kts
+++ b/build-logic/build.gradle.kts
@@ -15,6 +15,7 @@ dependencies {
     implementation(libs.develocity)
     implementation(libs.gradleDoctor)
     implementation(libs.kotlinter)
+    implementation(libs.android.kmp.library)
 
     // A hack to make version catalogs accessible from buildSrc sources
     // https://github.com/gradle/gradle/issues/15383#issuecomment-779893192
--- a/build-logic/settings.gradle.kts
+++ b/build-logic/settings.gradle.kts
@@ -5,6 +5,7 @@
 pluginManagement {
     // Add repositories required for build-settings-logic
     repositories {
+        google()
         gradlePluginPortal()
 
         // Should be in sync with ktorsettings.kotlin-user-project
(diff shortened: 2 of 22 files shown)


wkornewald:
@zibet27 Not sure if you receive notifications for that ticket. I've added a comment here about a more general approach for adding all non-Web targets at once: https://youtrack.jetbrains.com/issue/KTOR-7958/WebRTC-Client

ktor-client/ktor-client-webrtc/ktor-client-webrtc-rs/common/src/io/ktor/client/webrtc/rs/DataChannel.kt
```diff
@@ -0,0 +1,125 @@
...
+import kotlinx.coroutines.CoroutineStart
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.runBlocking
+import uniffi.ktor_client_webrtc.DataChannel
+import uniffi.ktor_client_webrtc.DataChannelMessage
+import uniffi.ktor_client_webrtc.DataChannelObserver
+import uniffi.ktor_client_webrtc.RtcException
+
+/**
+ * WebRtc data channel implementation for the Android platform.
```
wkornewald, line 18: Not for Android only.

zibet27: Yeah, that's a typo

PR #4906, opened 2025-06-03
KTOR-8558: Fix JS tasks running with configure-on-demand enabled

osipxd:
**Subsystem**
Infrastructure

**Motivation**
[KTOR-8558](https://youtrack.jetbrains.com/issue/KTOR-8558) Can't run JS tests when configure on demand is enabled

**Solution**
The root cause is in [KT-55701](https://youtrack.jetbrains.com/issue/KT-55701/KJS-Gradle-kotlinUpgradeYarnLock-and-kotlinStoreYarnLock-dont-support-configureondemand-if-building-one-module-only). Task `packageJsonUmbrella` that is responsible for gathering package JSONs from all projects requires these projects to be evaluated to work properly.

So I add an empty task to all projects and it makes it possible to request project evaluation using `Task.dependsOn(path: String)` function.

### DIFF
--- a/build-logic/src/main/kotlin/ktorbuild.base.gradle.kts
+++ b/build-logic/src/main/kotlin/ktorbuild.base.gradle.kts
@@ -9,3 +9,5 @@ version = resolveVersion()
 
 ProjectTagsService.register(project)
 extensions.create<KtorBuildExtension>(KtorBuildExtension.NAME)
+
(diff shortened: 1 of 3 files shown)


bjhham (approved):
Seems there's some build problems on the CI, though I don't see the connection immediately 😬

PR #4912, opened 2025-06-04
KTOR-8490 Do not parse ipString

osipxd:
Follow up for #4857

Realized that we can simply return IP address bytes instead of parsing `ipString`


ktor-network/common/test/io/ktor/network/sockets/tests/InetSocketAddressTest.kt
```diff
@@ -0,0 +1,29 @@
...
+
+class InetSocketAddressTest {
+    @Test
+    fun testResolveAddress() {
+        // Address resolving is not supported on JS and WASM platforms
+        if (PlatformUtils.IS_JS || PlatformUtils.IS_WASM_JS) return
+
+        val testCases = listOf(
+            "127.0.0.1" to byteArrayOf(127, 0, 0, 1),
+            "::1" to byteArrayOf(0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1),
```
osipxd, line 27: Test failed on macOS:
> kotlin.AssertionError: Unexpected bytes for address '::1'. Array elements differ at index 0. Expected element <0>, actual element <7>.
Expected <[0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1]>, actual <[7, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]>.

I can reproduce this problem locally and the resolved address is always consists one or two random bytes followed by zeroes. `ipString` returns `700::`, and Linux tests pass, so the problem is unlikely in the way I'm getting bytes.
@e5l, @bjhham, do you have any idea why `::1` is resolved to a random address on macOS?

osipxd: Found the root cause. `getAddressInfo` is wrapped with `memScoped`, so after we exit from this function the allocated `addrinfo` might be cleared.

ktor-network/nix/src/io/ktor/network/util/NativeSocketAddressNix.kt
```diff
@@ -45,18 +49,17 @@ internal actual class NativeIPv4SocketAddress(
         }
 }
 
-@OptIn(UnsafeNumber::class)
+@OptIn(UnsafeNumber::class, ExperimentalForeignApi::class)
 internal actual class NativeIPv6SocketAddress(
     family: UByte,
-    private val rawAddress: in6_addr,
+    private val rawAddress: CValue<in6_addr>,
```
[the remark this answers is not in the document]
osipxd: It's exactly the opposite. Now I should call `.readValue()` to get `CValue` from a struct 😅

PR #4913, opened 2025-06-05
WIP: Set caPath on linuxArm64

osipxd:
**Subsystem**
`ktor-client-curl`

**Motivation**
TBD

**Solution**
TBD


osipxd:
- Fixed in #5205

PR #4916, opened 2025-06-07
KTOR-8494 Modify DI to use async resolution by default

bjhham:
**Subsystem**
Server, DI

**Motivation**
[KTOR-8494](https://youtrack.jetbrains.com/issue/KTOR-8494) DI: Support for async dependencies

This is a rework of this concept and a general refactor of the DI plugin to incorporate "first-class" support for coroutines as part of the framework.

**Solution**
Now instead of using `provideAsync` and `resolveAwait`, we can provide async functions for any initialization function and `resolve()` always suspends.

Now, instead of relying on a two-phase approach to validating the dependency tree when provided out of order, we can use concurrent module loading where consumers can suspend until a provider comes along.  This assumes that the modules themselves aren't temporally coupled in other ways, so by default this behaviour will be disabled and dependencies will need to be registered and consumed in the correct sequence as with other libraries.

I've made some general changes to the API and plugin config to better suit the new behaviour.  Of note, there's some more exposure of the supplied create functions which have been renamed to `DependencyInitializer` so it's more clear that these are the main units of functionality we're working with for each service.  The provider and resolver parts of the registry are no longer configurable because they share a common backing map, and there's already several ways to customize them, so it should be less confusing for config.

I also added the ability to the extend the dependencies map through a service locator.  This was a requirement to avoid coupling third party plugins like Koin with ours.  Now, you can declare or resolve dependencies the Koin way only, or import our plugin to get the best of both worlds.


ktor-server/ktor-server-core/common/src/io/ktor/server/application/ApplicationModules.kt
```diff
@@ -0,0 +1,35 @@
...
+        "concurrent" -> ApplicationStartupMode.CONCURRENT
+        "sequential", null -> ApplicationStartupMode.SEQUENTIAL
+        else -> error("Invalid startup mode: $text")
+    }
+
+/**
+ * Configuration for the timeout on loading modules.  Defaults to 10 seconds.
+ */
+public val ApplicationEnvironment.startupTimeout: Long get() =
+    config.propertyOrNull("ktor.application.startupTimeoutMillis")?.getString()?.toLong() ?: 10_000
```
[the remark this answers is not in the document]
bjhham: That's ok, I think NumberFormatException here is sufficient.

ktor-server/ktor-server-core/api/ktor-server-core.api
```diff
@@ -133,7 +148,7 @@ public final class io/ktor/server/application/MissingApplicationPluginException
 
 public abstract interface class io/ktor/server/application/ModuleParametersInjector {
 	public static final field Companion Lio/ktor/server/application/ModuleParametersInjector$Companion;
-	public abstract fun resolveParameter (Lio/ktor/server/application/Application;Lkotlin/reflect/KParameter;)Ljava/lang/Object;
+	public abstract fun resolveParameter (Lio/ktor/server/application/Application;Lkotlin/reflect/KParameter;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
```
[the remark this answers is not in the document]
bjhham: Thanks for checking on breaking changes, but these changes have yet to be released, so breaking changes are ok.

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DeferredDependencyMap.kt
```diff
@@ -0,0 +1,37 @@
...
+@OptIn(ExperimentalCoroutinesApi::class)
+internal fun <T> Deferred<T>.tryGetCompleted(): T? =
+    if (isCompleted) {
+        val cause = getCompletionExceptionOrNull()
+        if (cause != null) throw cause
+        getCompleted()
+    } else {
+        null
+    }
+
```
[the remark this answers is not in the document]
bjhham: I'd rather the exception is thrown if something goes wrong with `getCompleted()` here.

ktor-server/ktor-server-core/jvm/src/io/ktor/server/application/ApplicationModules.jvm.kt
```diff
@@ -0,0 +1,58 @@
...
+            Dispatchers.Default.limitedParallelism(1)
+    ) {
+        val jobs = modules.map { module ->
+            launch {
+                module(application, classLoader)
+            }
+        }
+
+        // ensure all modules are started before proceeding
+        yield()
```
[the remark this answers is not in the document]
bjhham: Introducing a latch doesn't make sense here because of the relationship between the server and the plugins.  In this case, we have a convention for any supplied modules to provide dependencies first using normal blocking calls before doing any async work so the yield has the intended effect.

ktor-server/ktor-server-core/jvm/src/io/ktor/server/application/ApplicationModules.jvm.kt
```diff
@@ -0,0 +1,58 @@
...
+    for (module in modules) {
+        module(application, classLoader)
+    }
+}
+
+internal val LoadConcurrently = ApplicationModuleLoader { application, classLoader, modules ->
+    withContext(
+        application.coroutineContext +
+            CoroutineExceptionHandler { _, e -> application.environment.log.error("Failed to load module", e) } +
+            Dispatchers.Default.limitedParallelism(1)
```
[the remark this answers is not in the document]
bjhham: We want to avoid threading issues here because we have some shared state and don't have thread-safe collections to work with.  This is why we limit parallelism.

ktor-server/ktor-server-core/common/src/io/ktor/server/application/ApplicationModules.kt
osipxd: Minor. Maybe convert it to `Duration`?
```suggestion
    config.propertyOrNull("ktor.application.startupTimeoutMillis")?.getString()?.toLong()?.milliseconds ?: 10.seconds
```

ktor-server/ktor-server-core/jvm/src/io/ktor/server/application/ApplicationModules.jvm.kt
```diff
@@ -0,0 +1,69 @@
...
+            launch {
+                module(application, classLoader)
+            }
+        }
+
+        // ensure all modules are started before proceeding
+        yield()
+
+        // all modules are either loaded or suspended at this point
+        application.monitor.raise(ApplicationModulesLoading, application)
```
osipxd, line 62: Should this event be raised also in sequential mode?

ktor-server/ktor-server-core/common/src/io/ktor/server/application/ApplicationModules.kt
```diff
@@ -0,0 +1,35 @@
+/*
+ * Copyright 2014-2025 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.server.application
+
+/**
+ * Shorthand value for `ktor.application.modules` configuration list entries.
+ */
+public val ApplicationEnvironment.moduleConfigReferences: List<String> get() =
```
nomisRev: Do we want these to be public? It seems to increase our API surface for no real reason imho. I saw that we also have these for `port` and `host`, but they're actually incorrect.. To get the actual port/host you need to get them from `resolvedConnector`, and `ApplicationEnviroment.port` will result in the wrong values.

Similarly this will return incorrect values if used in `EmbeddedServer`, so it seems like something only useful to us internally in `EngineMain`?

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyResolution.kt
```diff
@@ -236,11 +263,11 @@ public data class DependencyResolverContext(
      * Property delegation for [DependencyResolverContext] for use with the `named` shorthand for string qualifiers.
      */
     public inline operator fun <reified T> getValue(thisRef: Any?, property: KProperty<*>): T =
-        resolver.resolve(name)
+        resolver.getBlocking(DependencyKey<T>(name))
```
nomisRev, line 266: I'd be in favour to remove these for now.

e5l (approved):
Looks good!

@vnikolova, could you check api docs?

ktor-server/ktor-server-core/jvm/src/io/ktor/server/application/ApplicationModules.jvm.kt
bjhham: Yes, I was just thinking about that when I woke up this morning, true story 😂

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyResolution.kt
```diff
@@ -46,9 +47,18 @@ public interface DependencyMap {
...
+        }
+
+        public fun fromMap(map: Map<DependencyKey, Any>): DependencyMap =
+            fromLookup(map::get)
+
+        @Suppress("UNCHECKED_CAST")
+        public fun fromLookup(resolve: (DependencyKey) -> Any?): DependencyMap = object : DependencyMap {
+            override fun contains(key: DependencyKey): Boolean = resolve(key) != null
+            override fun getInitializer(key: DependencyKey): DependencyInitializer =
+                DependencyInitializer.Value(key, resolve(key))
```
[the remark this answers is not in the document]
bjhham: The current implementation is consistent with the logic of the overall system where nulls are considered to be undefined.

ktor-server/ktor-server-core/common/src/io/ktor/server/application/ApplicationModules.kt
bjhham: Yeah, that's fair, I'll make this internal.  I think the semantics in the naming are specific enough that it should always be correct, though it's unlikely people would need something like this.

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyResolution.kt
bjhham: Hmm yeah it might be confusing having the multiple ways of handling things.  My thinking was that we might not want to force all adopters to change their modules to suspend, especially since all of our docs use the blocking functions.

nomisRev: Right but unexpected blocking is seems much worse? :/

In my personal opinion I think it's fine that adoption of the _new module_ forces adoption of `suspend` modules (also new feature). Also, do we not want to make suspend modules the default in 4.x.x? I think this would help us in the right direction.

I'd happily update any docs, the KLIP document, and template to reflect these changes if that helps ☺️

osipxd:
@arnaudgiuliani, FYI

arnaudgiuliani:
> @arnaudgiuliani, FYI

Thanks, will look at it 👍

PR #4917, opened 2025-06-09
KTOR-8565 Fix for merged config properties

bjhham:
**Subsystem**
Server, Config

**Motivation**
[KTOR-8565](https://youtrack.jetbrains.com/issue/KTOR-8565) Incorrect loading of ApplicationConfig

**Solution**
There was a missing clause that caused the merged property to be interpreted wrong when merging a second time, so merging more than once would cause this problem.

### DIFF
--- a/ktor-server/ktor-server-core/common/src/io/ktor/server/config/MergedApplicationConfig.kt
+++ b/ktor-server/ktor-server-core/common/src/io/ktor/server/config/MergedApplicationConfig.kt
@@ -4,6 +4,7 @@
 
 package io.ktor.server.config
 
+import io.ktor.server.config.ApplicationConfigValue.Type.*
 import io.ktor.server.config.MapApplicationConfig.Companion.flatten
 
 /**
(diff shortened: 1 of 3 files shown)


ktor-server/ktor-server-core/common/src/io/ktor/server/config/MergedApplicationConfig.kt
```diff
@@ -30,9 +30,12 @@ public fun List<ApplicationConfig>.merge(): ApplicationConfig {
...
  * @see [withFallback]
  */
-public fun ApplicationConfig.mergeWith(other: ApplicationConfig): ApplicationConfig {
-    return MergedApplicationConfig(other, this)
-}
+public fun ApplicationConfig.mergeWith(other: ApplicationConfig): ApplicationConfig =
+    when {
+        keys().isEmpty() -> other
+        other.keys().isEmpty() -> this
+        else -> MergedApplicationConfig(other, this)
```
osipxd, line 38: Is `ApplicationConfig` immutable? If it is possible to add keys after `mergeWith` called, new behavior might be surprising.

bjhham: Yeah, it's always immutable so we should be fine here 👍

PR #4922, opened 2025-06-10
KTOR-8575 Support variable expansion syntax with curly braces

Stexxe:
https://youtrack.jetbrains.com/issue/KTOR-8575


bjhham (approved):
lgtm 👍

PR #4924, opened 2025-06-11
KTOR-4766 Add defaultRequest plugin extension

e5l:


ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/DefaultRequest.kt
```diff
@@ -231,6 +232,16 @@ public class DefaultRequest private constructor(private val block: DefaultReques
         public fun setAttributes(block: Attributes.() -> Unit) {
             attributes.apply(block)
         }
+
+        public fun <T : Any> setCapability(key: HttpClientEngineCapability<T>, capability: T) {
+            val capabilities = attributes.computeIfAbsent(ENGINE_CAPABILITIES_KEY) { mutableMapOf() }
+            capabilities[key] = capability
+        }
```
bjhham, line 239: Maybe this method ought to be internal?

ktor-client/ktor-client-cio/jvm/test/io/ktor/client/engine/cio/UnixSocketTest.kt
```diff
@@ -55,7 +55,55 @@ class UnixSocketTest {
...
+
+        val client = HttpClient(io.ktor.client.engine.cio.CIO) {
+            defaultRequest {
+                unixSocket("/tmp/test-unix-socket-default.sock/")
+            }
+        }
+
+        try {
+            server.start(wait = false)
+            delay(1000)
```
bjhham, line 93: Would be nice to avoid the delay here.

nomisRev: ~I think we should be able to use `startSuspend` without `delay`?~ I saw another test had `startSuspend` with `delay` which seems like `startSuspend` might not be working right 😅

PR #4927, opened 2025-06-11
KTOR-7234 Fix WS session closure

Mr3zee:
**Subsystem**
WebSockets

**Motivation**
`DefaultWebSocketSession`'s coroutine scope is completed improperly.
The scope is completed using `complete` method on the underlying `Job`.
From the `complete` method doc:
> This function transitions this job into _completed_ state if it was not completed or cancelled yet.
However, that if this job has children, then it transitions into _completing_ state and becomes _complete_
once all its children are [complete][isCompleted]. See [Job] for details.

So in scenario like this on a client a launched coroutine will not stop if the session was stopped from the server for any reason:

```kotlin
client.webSocket("ws://localhost/ws") {
	// shared resources for the coroutine and the rest of the body
	... 

	launch {
		// some background work
		...
	}

	// process ws data
	...
}
```

I consider this behavior breaking structured concurrency.

**Solution**
Switch `complete` to `cancel`. 

**Issues**
This is behavioral breaking change. I need your opinion on whether it is reasonable to put this change into master, considering some user scenarios may break.


e5l (approved):
you also can close outgoing channel

Mr3zee:
@e5l outgoing channel is closed fine afaics

ktor-shared/ktor-websockets/common/src/io/ktor/websocket/DefaultWebSocketSession.kt
```diff
@@ -297,6 +305,7 @@ internal class DefaultWebSocketSessionImpl(
     private suspend fun sendCloseSequence(reason: CloseReason?, exception: Throwable? = null) {
         if (!tryClose()) return
         LOGGER.trace { "Sending Close Sequence for session $this with reason $reason and exception $exception" }
+        // don't cancel because sendCloseSequence is invoked inside a child coroutine of this context
         context.complete()
```
bjhham, line 309: Mildly confusing but ok 😕

bjhham (approved):

ktor-shared/ktor-websockets/common/src/io/ktor/websocket/DefaultWebSocketSession.kt
Mr3zee: Hence the comment ahaha
Because it was confusing for me too, until I saw the state of affairs. Left it for future generations

Mr3zee:
@bjhham can I assign the PR on you to deliver it into a proper release?

PR #4928, opened 2025-06-12
KTOR-8581 Avoid passing the max parameter of the decode method

Stexxe:
https://youtrack.jetbrains.com/issue/KTOR-8581


e5l (approved):
lgtm

PR #4938, opened 2025-06-16
KTOR-6794 Send response in routing on failures

Stexxe:
Forces to send the response on failures in the routing if nothing has handled the response in the call pipeline.
https://youtrack.jetbrains.com/issue/KTOR-6794


ktor-server/ktor-server-core/common/test/io/ktor/server/application/HooksTest.kt
```diff
@@ -185,4 +187,23 @@ class HooksTest {
...
+        var responseSentCalled = false
+        routing {
+            install(createRouteScopedPlugin("Plugin") {
+                on(ResponseSent) {
+                    responseSentCalled = true
+                }
+            })
+
+            get("/") {
+                throw BadRequestException("Bad Request")
```
e5l, line 205: Could you also add a test for internal server error?

PR #4943, opened 2025-06-18
Stexxe/cors options routing wildcard

Stexxe:
https://youtrack.jetbrains.com/issue/KTOR-4499

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-cors/common/src/io/ktor/server/plugins/cors/routing/CORS.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-cors/common/src/io/ktor/server/plugins/cors/routing/CORS.kt
@@ -23,6 +23,16 @@ import io.ktor.server.plugins.cors.*
  *
  * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.server.plugins.cors.routing.CORS)
  */
+
+@Deprecated(
+    message = "???",
+    level = DeprecationLevel.ERROR,
+    replaceWith = ReplaceWith("CORS", "io.ktor.server.plugins.cors.routing.CORS")
+)
 public val CORS: RouteScopedPlugin<CORSConfig> = createRouteScopedPlugin("CORS", ::CORSConfig) {
+    buildPlugin()
+}
+
+internal val commonCORS: RouteScopedPlugin<CORSConfig> = createRouteScopedPlugin("CORS", ::CORSConfig) {
     buildPlugin()
 }
--- a/ktor-server/ktor-server-plugins/ktor-server-cors/common/src/io/ktor/server/plugins/cors/routing/CORS.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-cors/common/src/io/ktor/server/plugins/cors/routing/CORS.kt
@@ -25,9 +25,8 @@ import io.ktor.server.plugins.cors.*
  */
 
 @Deprecated(
-    message = "???",
+    message = "The plugin installation has been moved to the Route.cors method",
     level = DeprecationLevel.ERROR,
-    replaceWith = ReplaceWith("CORS", "io.ktor.server.plugins.cors.routing.CORS")
 )
 public val CORS: RouteScopedPlugin<CORSConfig> = createRouteScopedPlugin("CORS", ::CORSConfig) {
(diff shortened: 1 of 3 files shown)


ktor-server/ktor-server-plugins/ktor-server-cors/common/src/io/ktor/server/plugins/cors/routing/CORS.kt
```diff
@@ -23,6 +23,15 @@ import io.ktor.server.plugins.cors.*
  *
  * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.server.plugins.cors.routing.CORS)
  */
+
+@Deprecated(
+    message = "The plugin installation has been moved to the Route.cors method",
```
e5l: let's explain the difference between `install(CORS)` and `cors {}` in the API docs for the both symbols

ktor-server/ktor-server-tests/common/test/io/ktor/tests/server/plugins/CORSTest.kt
```diff
@@ -1275,4 +1291,292 @@ class CORSTest {
...
+            assertEquals(response.status, HttpStatusCode.Forbidden)
+        }
+    }
+
+    @Test
+    fun preflightPluginInParentHandlesChildRoutes() = testApplication {
+        routing {
+            route("/outer") {
+                cors {
+                    anyHost()
```
e5l, line 1538: this DSL looks like we're going to write routes inside it:
```
cors {
   route("/abc") {}
}
```

Let's think how to design it right

e5l (changes requested):

ktor-server/ktor-server-plugins/ktor-server-cors/common/src/io/ktor/server/plugins/cors/routing/CORS.kt
```diff
@@ -24,5 +25,9 @@ import io.ktor.server.plugins.cors.*
  * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.server.plugins.cors.routing.CORS)
  */
 public val CORS: RouteScopedPlugin<CORSConfig> = createRouteScopedPlugin("CORS", ::CORSConfig) {
+    route?.options("{cors-options-wildcard...}") {
```
e5l, line 28: Could you tell me why we need an explicit handler?

Stexxe: Otherwise, this route will not be resolved, and thus, the `onCall` handler of the CORS plugin will not be called.

ktor-server/ktor-server-core/common/src/io/ktor/server/application/ApplicationPlugin.kt
```diff
@@ -168,6 +168,10 @@ private fun <B : Any, F : Any> RoutingNode.installIntoRoute(
     val installed = plugin.install(fakePipeline, configure)
     pluginRegistry.put(plugin.key, installed)
 
+    for (child in fakePipeline.children) { // We need to transfer nodes added into the fake pipeline
+        copyChildrenRecursively(child)
```
e5l, line 172: The only place I'm concerned about. @bjhham, what do you think?

bjhham: Having all this logic for the CORS plugin does seem like a code smell for the rigidity of the routing API, but the functionality looks solid from the tests and I can't think of another way to achieve the same effect.  Maybe I'll create an issue to investigate areas for improvement in the routing API in 4.0 with this as an example.

bjhham (approved):
Looks good to me, nice tests 👍

ktor-server/ktor-server-core/common/src/io/ktor/server/application/ApplicationPlugin.kt
bjhham: New ticket for improvements here https://youtrack.jetbrains.com/issue/KTOR-8666/Routing-API-areas-for-improvement

e5l (approved):

PR #4948, opened 2025-06-19
KTOR-2934 Throw BadRequestException on query string parsing errors wi…

Stexxe:
https://youtrack.jetbrains.com/issue/KTOR-2934


e5l (approved):
lgtm

PR #4949, opened 2025-06-19
KTOR-8597 Fix serialization links to the docs

Stexxe:
https://youtrack.jetbrains.com/issue/KTOR-8597


vnikolova (approved):
Looks good! Thanks for the update 🙏

PR #4960, opened 2025-06-24
Fix footer for non index.html pages

Mr3zee:
**Subsystem**
Gradle, Dokka

**Motivation**
Footer on pages that are not index.html is wrong

**Solution**
fix it!


Mr3zee:
Oops, it was already fixed, I was on a stale branch

PR #4962, opened 2025-06-27
KTOR-8618 Fix race condition in ReaderJob.flushAndClose

Stexxe:
https://youtrack.jetbrains.com/issue/KTOR-8618

### DIFF
--- a/ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
+++ b/ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
@@ -303,7 +303,10 @@ public class ReaderJob internal constructor(
     @InternalAPI
     public suspend fun flushAndClose() {
         job.cancelChildren()
-        job.children.forEach { it.join() }
+        job.children.forEach {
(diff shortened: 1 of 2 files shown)


ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
```diff
@@ -303,7 +303,10 @@ public class ReaderJob internal constructor(
     @InternalAPI
     public suspend fun flushAndClose() {
         job.cancelChildren()
-        job.children.forEach { it.join() }
+        job.children.forEach {
+            it.cancel() // Children may appear at this point so we cancel them before joining
```
e5l, line 307: What if the new child appears during forEach iteration? Should we cancel the job before?

Stexxe: They are cancelled before in `job.cancelChildren`. 
If the new child appears during the `forEach` iteration, I suppose it won't be joined, so that the deadlock won't happen.

PR #4963, opened 2025-06-27
KTOR-8622 ForwardedHeaders: Parse parameters case insensitively

solonovamax:
**Subsystem**
Server `ForwardedHeaders` plugin

**Motivation**
https://youtrack.jetbrains.com/issue/KTOR-8622

Ensures compliance with [RFC 7239](https://datatracker.ietf.org/doc/html/rfc7239):
> The first element in this list holds information added by the first proxy that implements and uses this header field, and each subsequent element holds information added by each subsequent proxy.  Because this header field is optional, any proxy in the chain may choose not to update this header field.  Each field-value is a semicolon-separated list; this sublist consists of parameter-identifier pairs. Parameter-identifier pairs are grouped together by an equals sign. Each parameter MUST NOT occur more than once per field-value.  **The parameter names are case-insensitive.**

emphasis my own.

**Solution**
just `.lowercase()` the parameter names before they're added to the map.


solonovamax:
Those test failures seem completely unrelated to the PR, I believe they may just be flaky (only some of them were marked as flaky, however I did not touch anything relating to them)

bjhham (approved):

PR #4964, opened 2025-06-27
KTOR-8559 Thymeleaf: Allow null values in template model

solonovamax:
**Subsystem**
Server `Thymeleaf` plugin.

**Motivation**
https://youtrack.jetbrains.com/issue/KTOR-8559

Allows `null`/nullable values to be passed to the model used for thymeleaf.

imo this is a bugfix as it does not involve any breaking api changes & is just matching features with what thymeleaf already supports (ie. if you do an unchecked cast to `Map<String, Any>` and pass a null value, everything will work just fine), however the issue in youtrack was marked as "Feature" rather than "Bug", so I named my branch following that.

**Solution**
`Any` -> `Any?`


marychatte (approved):
@solonovamax thank you for the PR!
It looks good. I just added a test to verify that the solution works as expected. I’ll merge the PR once the checks finish.

PR #4967, opened 2025-06-29
KTOR-8450: add content negotiation suffix support

anton-erofeev:
**Subsystem**
Client, ktor.client.plugins.contentnegotiation

**Motivation**
https://youtrack.jetbrains.com/issue/KTOR-8450

**Solution**
added content negotiation suffix support

### DIFF
--- a/ktor-client/ktor-client-plugins/ktor-client-content-negotiation/common/src/io/ktor/client/plugins/contentnegotiation/ContentNegotiation.kt
+++ b/ktor-client/ktor-client-plugins/ktor-client-content-negotiation/common/src/io/ktor/client/plugins/contentnegotiation/ContentNegotiation.kt
@@ -19,6 +19,8 @@ import io.ktor.utils.io.*
 import io.ktor.utils.io.charsets.*
 import kotlin.reflect.*
 
+// https://datatracker.ietf.org/doc/html/rfc6839
+private val supportedSuffixTypes = setOf("json", "ber", "der", "fastinfoset", "wbxml", "zip", "xml")
 private val LOGGER = KtorSimpleLogger("io.ktor.client.plugins.contentnegotiation.ContentNegotiation")
 
 internal val DefaultCommonIgnoredTypes: Set<KClass<*>> = setOf(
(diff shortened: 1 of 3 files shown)


ktor-client/ktor-client-plugins/ktor-client-content-negotiation/common/src/io/ktor/client/plugins/contentnegotiation/ContentNegotiation.kt
```diff
@@ -150,7 +148,17 @@ public class ContentNegotiationConfig : Configuration {
...
+        override fun contains(contentType: ContentType): Boolean {
+            if (contentType.match(pattern)) {
+                return true
+            }
+
+            val value = contentType.withoutParameters().toString()
+
+            return value.startsWith(pattern.contentType) &&
+                value.endsWith("+${pattern.contentSubtype}", ignoreCase = true) &&
+                pattern.contentSubtype in supportedSuffixTypes
```
osipxd: It's nice idea, but it won't work for `+wbxml` as it should be handled as `application/vnd.wap.wbxml`. Probably we should convert `supportedSuffixTypes` into a map.
Another problem that is that RFC 6839 doesn't require content type to match "application/", so we can drop the first condition from here.
Probably, something like this would work (I haven't tested nor run this code):

```kotlin
private val supportedSuffixTypes = mapOf(
    ContentType.Application.Json -> "json",
    ContentType("application", "vnd.wap.wbxml") -> "wbxml"
    // ...
)
```
```kotlin
private fun defaultMatcher(pattern: ContentType): ContentTypeMatcher = object : ContentTypeMatcher {
    // Access the map only once when creating the matcher
    private val subtypeSuffix = supportedSuffixTypes[pattern]

    override fun contains(contentType: ContentType): Boolean {
        return contentType.match(pattern) ||
            subtypeSuffix != null && contentType.contentSubtype.endsWith("+$subtypeSuffix", ignoreCase = true)
    }
}
```

osipxd: Could you also add tests for the mentioned cases?

osipxd (changes requested):
Hey @anton-erofeev, thank you for the PR!
I'm sorry for the long delay 🙏  Please check my comment bellow

osipxd:
Hi @anton-erofeev,
Are you planning to continue working on this PR or would you be okay if I picked it up? 
Thanks for your contribution to the project!

anton-erofeev:
Hey @osipxd, sure, please feel free to pick it up

Mr3zee:
@osipxd I took over and pushed a fix to your comments, please check

osipxd (approved):
LGTM!
Could you run `./switch-base-branch.sh` before merging so we could include this fix into the next patch release?

Mr3zee:
@osipxd done

PR #4968, opened 2025-06-30
[KTOR-8627] Reorder SSE fields for streaming deserialization

kpavlov:
Ticket: [KTOR-8627](https://youtrack.jetbrains.com/issue/KTOR-8627) Change the order of SSE field serialization: put `event` before `data`

**Subsystem**
SSE: ktor-shared/ktor-sse

**Motivation**

Currently, the `data` field comes before the `event` field in the SSE payload. If the streaming parser used the `event` field as a discriminator to determine the deserializer, it must buffer the `data`. Moving `event` before `data` would allow streaming SSE parsers to avoid buffering SSE's `data` field and choosing the right deserializer before reading the data field. In the [reference](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events#event_stream_format) `event` field also comes before the `data`.

**Solution**

Change the order of field serialization to put `event` before `data`. This allows using `event` as a discriminator for streaming parsers.

### DIFF
--- a/ktor-shared/ktor-sse/common/src/io/ktor/sse/ServerSentEvent.kt
+++ b/ktor-shared/ktor-sse/common/src/io/ktor/sse/ServerSentEvent.kt
@@ -80,7 +80,8 @@ public data class TypedServerSentEvent<T>(
 }
 
 /**
- * Serialize event to string representation according to the [spec](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events#event_stream_format)
+ * Serialize [event](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events#event_stream_format)
+ * to a string representation
  */
 private fun eventToString(data: String?, event: String?, id: String?, retry: Long?, comments: String?): String {
     return buildString {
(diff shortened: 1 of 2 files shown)


ktor-shared/ktor-sse/common/src/io/ktor/sse/ServerSentEvent.kt
```diff
@@ -79,10 +79,13 @@ public data class TypedServerSentEvent<T>(
         eventToString(data?.let { serializer(it) }, event, id, retry, comments)
 }
 
+/**
+ * Serialize event to string representation according to the [spec](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events#event_stream_format)
```
marychatte: Spec doesn't specify the order of the fields, I suggest to change KDoc:
```
/**
 * Serialize [event](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events#event_stream_format) to string representation
 **/
 ```

**Review by marychatte — APPROVED** [review]

**Comment by loxal** [comment]
After this change my tests that worked just fine started failing. 

```kotlin
                            val sseChannel = eventStream.bodyAsChannel()
                            while (sseChannel.isClosedForRead.not()) { // is always `false` since Ktor v3.2.1 
                                val channelLine = sseChannel.readUTF8Line() // is always empty, since Ktor v3.2.1 
                                
```

PR #4972, opened 2025-07-01
KTOR-7910 Fix WriteChannelContent streaming in testApplication

Stexxe:
https://youtrack.jetbrains.com/issue/KTOR-7910

### DIFF
--- a/ktor-server/ktor-server-test-host/common/src/io/ktor/server/testing/TestApplicationResponse.kt
+++ b/ktor-server/ktor-server-test-host/common/src/io/ktor/server/testing/TestApplicationResponse.kt
@@ -52,7 +52,7 @@ public class TestApplicationResponse(
 
     private var responseJob: Job? = null
 
-    internal val writeContentChannel = atomic<ByteChannel?>(null)
+    internal val writeContentChannel = atomic<ByteReadChannel?>(null)
 
     /**
      * Get completed when a response channel is assigned.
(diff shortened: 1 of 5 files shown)


ktor-server/ktor-server-test-host/common/src/io/ktor/server/testing/TestApplicationResponse.kt
```diff
@@ -51,6 +52,8 @@ public class TestApplicationResponse(
 
     private var responseJob: Job? = null
 
+    internal val writeContentChannel = atomic<ByteChannel?>(null)
```
bjhham: Could we not avoid the new field and instead leverage the existing `responseChannel` and `responseJob` here?

It looks like you could work as long as the test client engine references the response channel in the new way.

bjhham (commented):
It seems

ktor-server/ktor-server-test-host/common/src/io/ktor/server/testing/TestApplicationResponse.kt
Stexxe: I can do that for the new streaming functionality (`respondWriteChannelContent`), but I don't see a way to use the `responseChannel` for the existing cases without breaking them. In my opinion, this part of `testApplication` requires a complete rewrite to make the bridge between the server and the client work properly.

marychatte (approved):

ktor-server/ktor-server-test-host/common/src/io/ktor/server/testing/TestApplicationResponse.kt
bjhham: Hmm yeah it looks that way 😞

bjhham (approved):

PR #4980, opened 2025-07-04
Fix KTOR-8232 StatusPages: OutgoingContent headers aren't available in the status handlers

marychatte:
**Subsystem**
Server, StatusPages plugin

**Motivation**
[KTOR-8232](https://youtrack.jetbrains.com/issue/KTOR-8232) 

**Solution**
Append OutgoingContent headers in the `ResponseBodyReadyForSend` hook


osipxd (approved):
LGTM

PR #4981, opened 2025-07-04
Fix KTOR-8411 CountedByteWriteChannel: fix autoFlush not working

marychatte:
**Subsystem**
Core, io

**Motivation**
[KTOR-8411](https://youtrack.jetbrains.com/issue/KTOR-8411/CountedByteWriteChannel-autoFlush-of-the-source-channel-doesnt-make-the-channel-auto-flushing)

**Solution**
A solution was suggested in [PR #4800](https://github.com/ktorio/ktor/pull/4800), but the JS/Wasm tests are hanging, so it was reverted.
This PR contains a previous solution and a proper closing of the server in a test

### DIFF
--- a/ktor-network/common/test/io/ktor/network/sockets/tests/TCPSocketTest.kt
+++ b/ktor-network/common/test/io/ktor/network/sockets/tests/TCPSocketTest.kt
@@ -4,12 +4,14 @@
 
 package io.ktor.network.sockets.tests
 
-import io.ktor.network.selector.SelectorManager
+import io.ktor.network.selector.*
 import io.ktor.network.sockets.*
 import io.ktor.utils.io.*
 import io.ktor.utils.io.CancellationException
 import kotlinx.coroutines.*
-import kotlinx.io.*
+import kotlinx.io.EOFException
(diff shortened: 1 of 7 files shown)


ktor-network/common/test/io/ktor/network/sockets/tests/TCPSocketTest.kt
```diff
@@ -288,4 +290,33 @@ class TCPSocketTest {
...
+
+        val countedWriteChannel = CountedByteWriteChannel(writeChannel)
+        countedWriteChannel.writeStringUtf8("Hello again\n")
+        val message2 = serverInput.readUTF8Line()
+        assertEquals("Hello again", message2)
+
+        serverConnection.close()
+        clientConnection.close()
+
+        server.close()
```
bjhham: Might want to use try/finally here so we get the right feedback if it fails.

bjhham (commented):
Looks like this will need to go into 3.3 since it introduces autoflush to the interface.

marychatte:
Yes, changed the target release in the task, but forgot to change the branch here 😅

PR #4983, opened 2025-07-04
KTOR-5922 Discover HTTP proxy settings on Android

Stexxe:
https://youtrack.jetbrains.com/issue/KTOR-5922


ktor-client/ktor-client-cio/jvm/src/io/ktor/client/engine/cio/Proxy.jvm.kt
```diff
@@ -0,0 +1,26 @@
...
+import java.net.InetSocketAddress
+import java.net.Proxy
+
+internal actual fun discoverHttpProxy(): ProxyConfig? {
+    var proxy: Proxy? = null
+
+    try {
+        // For Android
+        val host = System.getProperty("http.proxyHost")
+        val port = System.getProperty("http.proxyPort")?.toIntOrNull() ?: 0
```
osipxd: I guess we should rely on `ProxySelector.getDefault()` instead. There is more sophisticated logic to determine proxy host and port (see `sun.net.spi.DefaultProxySelector`).
This would require requesting a new proxy for each URL, but it doesn't seem a problem here.

osipxd (changes requested):
I think we can remove mention of "Android" in comments and in tests since they are JVM-wide system properties. And could you consider using `ProxySelector` instead of property lookup?

ktor-client/ktor-client-cio/jvm/src/io/ktor/client/engine/cio/Proxy.jvm.kt
Stexxe: At this point, we don't have the URL to choose the proxy for. Can you explain why the solution with the `ProxySelector` is better than the current one?

osipxd: > At this point, we don't have the URL to choose the proxy for.

Yes, but we could turn the field `proxy` into a lambda taking an `Url` as a parameter.

> Can you explain why the solution with the ProxySelector is better than the current one?

I see two reasons.
The first one is that the current logic is not enough. Proxy could also be configured via `proxyHost` and `socksProxyHost` system properties. And `ProxySelector` would honor non-proxy hosts configured in the system.
The second reason is that using `ProxySelector` would make CIO behavior consistent to OkHttp and Java engines as they also rely on `ProxySelector`.
As a bonus it would solve [this issue](https://youtrack.jetbrains.com/issue/KTOR-3727/Add-a-way-to-configure-a-proxy-for-each-HTTP-request) (for JVM only).

ktor-client/ktor-client-cio/jvm/src/io/ktor/client/engine/cio/Proxy.jvm.kt
```diff
@@ -0,0 +1,27 @@
...
+internal actual fun lookupGlobalProxy(url: Url): ProxyConfig? {
+    val proxies = ProxySelector.getDefault().select(url.toURI())
+
+    return if (proxies.isNotEmpty()) {
+        val proxy = proxies.first()
+
+        if (proxies.size == 1 && proxy.type() == Proxy.Type.DIRECT) {
+            null
+        } else {
+            proxies.first()
```
osipxd: Minor
```suggestion
            proxy
```

ktor-client/ktor-client-cio/jvm/src/io/ktor/client/engine/cio/Proxy.jvm.kt
```diff
@@ -0,0 +1,27 @@
...
+import java.net.Proxy
+import java.net.ProxySelector
+
+internal actual fun lookupGlobalProxy(url: Url): ProxyConfig? {
+    val proxies = ProxySelector.getDefault().select(url.toURI())
+
+    return if (proxies.isNotEmpty()) {
+        val proxy = proxies.first()
+
+        if (proxies.size == 1 && proxy.type() == Proxy.Type.DIRECT) {
```
osipxd, line 31: A comment clarifying the purpose of this `if` would be helpful. As now it is not clear without reading `ProxySelector.select(...)` documentation

PR #4984, opened 2025-07-04
Update dependencies

osipxd:
Merge Renovate PRs at once (only patch releases)


bjhham (approved):

gradle/libs.versions.toml
```diff
@@ -4,9 +4,9 @@ kotlin = "2.1.21"
 kotlinx-html = "0.12.0"
 kotlinx-io = "0.7.0"
 coroutines = "1.10.2"
-atomicfu = "0.27.0"
+atomicfu = "0.28.0"
 serialization = "1.8.1"
```
nomisRev, line 8:
```suggestion
serialization = "1.9.0"
```

`kotlin.time` is only support from 1.9.0, so I think should be bumped.

nomisRev (changes requested):

gradle/libs.versions.toml
osipxd: kotlinx-serialization depends on Kotlin 2.2.0, but we decided to not update it until the next minor release.
We use only `Clock` and do not serialize `Instant`s, so we can go without updating serialization. Meanwhile our users are free to update it if they want to (de)serialize `kotlin.time` types.

nomisRev: Sorry was too late to remove this comment after our conversation on Slack 👍

PR #4986, opened 2025-07-07
KTOR-8656 Add the kotlinx-datetime dependency back

osipxd:
**Subsystem**
Server

**Motivation**
[KTOR-8656](https://youtrack.jetbrains.com/issue/KTOR-8656) kotlinx.datetime is not available transitively in 3.2.1

**Solution**
Add the kotlinx-datetime dependency back, but do not use it. Schedule it for removal in the next minor version.


nomisRev:
~I think we should also postpone [this](https://github.com/ktorio/ktor/pull/4974/files#diff-257186d1e427375a659860fb0a526d5c6ac901c515daf3e84320eb2ae86c19c4L40) until 3.3.0 where we can bump to Kotlin 2.2, and KotlinX Serialization to 1.9.0 preventing breaking any existing code from our side.~

PR #4994, opened 2025-07-07
KTOR-8377 Keep compression headers but skip check for relevant platforms

bjhham:
**Subsystem**
Client, ContentEncoding

**Motivation**
[KTOR-8377](https://youtrack.jetbrains.com/issue/KTOR-8377) JS/WASM: response doesn't contain the Content-Hength header in a browser
[KTOR-8526](https://youtrack.jetbrains.com/issue/KTOR-8526) Darwin: The Content-Encoding header is removed since 3.0.3

**Solution**
Instead of stripping headers, we skip the checks for content length and decoding.


bjhham:
Sorry, got a little ahead of myself when I saw the jvm tests passing 😓

PR #5002, opened 2025-07-10
Update report links for APIs added in 3.2.0

osipxd:


bjhham (approved):
We should really just write a dokka plugin for this.

osipxd:
> We should really just write a dokka plugin for this.

IIUC we want report links to be shown in sources in KDoc, not only in the generated Dokka
We could write a Gradle Plugin though.

PR #5004, opened 2025-07-14
[js, wasmJs] fix crash of absence of the `process` global by providing default values

der-fruhling:
**Subsystem**
client/server: ktor-utils and ktor-server-core

**Motivation**
I was trying to use Ktor Client with Wasm/JS, noticed this issue, and just decided to fix it. 

Kotlin Client on Wasm/JS crashes when you try to use it because it's assuming that the `process` variable is defined in a browser environment and tried to read it for it's log level configuration.

**Solution**
If `process` is `null` or `undefined` (false-ish), provide a default value instead of trying to access `env`.

I intended to solve this in a way that allows defining `process` in such environments for debugging purposes, e.g.:
```html
<script>
    process = { env: { KTOR_LOG_LEVEL: 'TRACE' } }
    // load kotlin/js / kotlin/wasm and do stuff with ktor here
</script>
```

### DIFF
--- a/ktor-shared/ktor-websockets/common/src/io/ktor/websocket/DefaultWebSocketSession.kt
+++ b/ktor-shared/ktor-websockets/common/src/io/ktor/websocket/DefaultWebSocketSession.kt
@@ -157,8 +157,16 @@ internal class DefaultWebSocketSessionImpl(
 
         _extensions.addAll(negotiatedExtensions)
         runOrCancelPinger()
-        runIncomingProcessor(ponger(outgoing))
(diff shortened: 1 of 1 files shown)


bjhham:
For tracking: https://youtrack.jetbrains.com/issue/KTOR-8686/JS-WASM-error-when-process-global-is-undefined

bjhham:
Could you please run `./gradlew formatKotlin` to appease the linter?

der-fruhling:
Thanks!

Ran the linter as you requested.

PR #5009, opened 2025-07-14
KTOR-8193 Reduce blocking in channel output stream wrapper

bjhham:
**Subsystem**
Server, Core I/O

**Motivation**
[KTOR-8193](https://youtrack.jetbrains.com/issue/KTOR-8193) OutputStreamChannel block request processing

**Solution**
It is not entirely possible to avoid blocking threads when waiting for the content to be consumed, but we can optimize for payloads under 1 MB by using the buffers directly instead of using the high-level suspend API.  Since the previous fix to use a dedicated dispatcher for this work will free up the standard IO dispatcher, this should prevent thread exhaustion while maintaining good performance for smaller payloads.


ktor-io/jvm/src/io/ktor/utils/io/jvm/javaio/Blocking.kt
```diff
@@ -54,20 +57,50 @@ public fun ByteReadChannel.toInputStream(parent: Job? = null): InputStream = obj
...
     override fun write(b: ByteArray, off: Int, len: Int) {
-        runBlocking { this@toOutputStream.writeFully(b, off, off + len) }
+        throwIfClosed()
+        channelWriteBuffer.write(b, off, off + len)
+        flushIfNeeded()
+    }
+
+    @OptIn(InternalIoApi::class)
+    private fun flushIfNeeded() {
+        if (channelWriteBuffer.buffer.size >= FLUSH_THRESHOLD) {
```
osipxd, line 85: Should we check `autoFlush` here?

bjhham: Good point 👍

bjhham: Looks like the always-on flushing is important based on the failing tests 😞

osipxd: Hmm.. I don't think I understand why it stopped working. `flushIfNeeded` used by `ByteWriteChannel.writeFully` also checks `autoFlush` under the hood, so I think the root cause might be different. Might it be caused by removing `runBlocking` and suspending calls from the execution path?

bjhham: The logic is a little different for autoFlush.  It makes it so that it flushes on every write, whereas the normal behaviour is still to automatically flush when the buffer is full, which is consistent with the behaviour here.

Note the `||`:
```kotlin
public suspend fun ByteWriteChannel.flushIfNeeded() {
    rethrowCloseCauseIfNeeded()

    if (autoFlush || writeBuffer.size >= CHANNEL_MAX_SIZE) flush()
}
```

osipxd: Ah, I see. Should we also add `autoFlush ||` here?

osipxd (approved):

PR #5010, opened 2025-07-14
WebRtc Client. Introduce data channels to the common API.

zibet27:
**Subsystem**
Ktor Client WebRtc

**Motivation**
The data channel is an important part of the WebRTC protocol. I've implemented the common API and added stubs to existing engines. In the following commits, I will add real implementations.
Please review.


ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRtcPeerConnectionEvents.kt
```diff
@@ -105,6 +106,7 @@ internal class WebRtcConnectionEventsEmitter(
     private val signalingStateFlow: MutableStateFlow<WebRtc.SignalingState> =
         MutableStateFlow(WebRtc.SignalingState.STABLE)
     private val trackEventsFlow: MutableSharedFlow<TrackEvent> = MutableSharedFlow(config.remoteTracksReplay)
+    private val dataChannelsFlow: MutableSharedFlow<WebRtcDataChannel> = MutableSharedFlow(config.dataChannelsReplay)
```
bjhham: Shouldn't the data channels flow be roughly the same as media tracks?  I see you can subscribe to the close events for cleanup, but it seems oddly asymmetrical here.

Mr3zee:
@zibet27 is there a resource I can use to read about data channels?

ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRtcPeerConnectionEvents.kt
zibet27: Hmm, yeah. It makes sense. We could add `DataChannelEvent` to keep it symmetrical. Also, I just realised that subscribing to the channel state becomes redundant because it is already included in those events.

zibet27:
> @zibet27 is there a resource I can use to read about data channels?

@Mr3zee Sorry, forgot to add this [link](https://developer.mozilla.org/en-US/docs/Web/API/RTCDataChannel) to the description.

ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRtcDataChannel.kt
```diff
@@ -0,0 +1,250 @@
...
+/**
+ * Abstract class representing a network channel which can be used for bidirectional peer-to-peer transfers
+ * of arbitrary data. Every data channel is associated with an [WebRtcPeerConnection].
+ *
+ * @see [MDN RTCDataChannel](https://developer.mozilla.org/en-US/docs/Web/API/RTCDataChannel)
+ */
+public abstract class WebRtcDataChannel(
+    override val coroutineContext: CoroutineContext,
+    options: WebRtcDataChannelOptions
+) : CoroutineScope, AutoCloseable {
```
Mr3zee: Whys is it a `CoroutineScope`? And what scope is it?

ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRtcDataChannel.kt
```diff
@@ -0,0 +1,250 @@
...
+    private val messagesFlow: MutableSharedFlow<WebRtc.DataChannelMessage> = MutableSharedFlow(options.messagesReplay)
+
+    /**
+     * Flow of incoming messages on the data channel.
+     *
+     * @see [MDN message event](https://developer.mozilla.org/en-US/docs/Web/API/RTCDataChannel/message_event)
+     */
+    public val messages: SharedFlow<WebRtc.DataChannelMessage> = messagesFlow.asSharedFlow()
+
+    protected fun emitMessage(message: WebRtc.DataChannelMessage) {
```
Mr3zee: Why not `suspend` and no `launch`?

ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRtcPeerConnectionEvents.kt
```diff
@@ -156,4 +160,8 @@ internal class WebRtcConnectionEventsEmitter(
     fun emitRemoveTrack(track: WebRtcMedia.Track) {
         launch { trackEventsFlow.emit(TrackEvent.Remove(track)) }
     }
+
+    fun emitDataChannelEvent(event: DataChannelEvent) {
+        launch { dataChannelEventsFlow.emit(event) }
+    }
```
Mr3zee: Same question here

ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRtcDataChannel.kt
zibet27: The reason I did it that way is that under the hood (in browser or android library used) we listen to the events using callbacks, so there is no suspend or async, and to provide our flows API we need to create our scope somewhere to at least emit some value (and the collector will have its own scope). 
Example:

```
public class JsWebRtcPeerConnection(
    internal val nativePeerConnection: RTCPeerConnection,
    coroutineContext: CoroutineContext,
    config: WebRtcConnectionConfig
) : WebRtcPeerConnection(coroutineContext, config) {
    init {
        nativePeerConnection.onicecandidate = { event: RTCPeerConnectionIceEvent ->
            event.candidate?.let { events.emitIceCandidate(it.toKtor()) }
            // if we don't `launch` in the `emitIceCandidate`, then we would need to do it here
        }
        ....
    }
    ....
}
```

Now, each`WebRtcPeerConnection` has one scope to emit all its events, and it's created here:

```
public abstract class WebRtcEngineBase(private val engineName: String) : WebRtcEngine {
    private val job = SupervisorJob()
    private val scope = CoroutineScope(job + CoroutineName("$engineName-context"))

    override val coroutineContext: CoroutineContext
        get() = scope.coroutineContext + (config.dispatcher ?: ioDispatcher())
 ```
 
 Also, I plan to share the context of `WebRtcPeerConnection` to its datachannels. Or should I create new ones for every data channel?

Comment (Mr3zee) [reply ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRtcDataChannel.kt]: The problem I see here is inconsistency and some dangerous APIs.

First, `AutoClosable` and `CoroutineScope` are mutually excluding.
Then providing `CoroutineScope` as a public API is bad structured concurrency semantics: a user can can do `dataChannel.launch { }` and the meaning of it is ambiguous. 

There is also an inconsistency in APIs of this class:
- `send` for sending events 
- `messages: SharedFlow<*>` for receiving.

What I propose to do here is to mimmic a `Channel` - a low level APIs for events. As `WebRtcDataChannel` is also considered a low level API for RTC (and also the name similarity) this suits as a good abstraction. 

In this case you change `messages` to `receive` method, `messagesFlow` to a `Channel` and `emitMessage` to `channel.trySend` which is non-suspendible. You keep `AutoClosable` to close the channel and you provide users with APIs for the cannel configuration (buffer size, an overflow strategy and a discard strategy).

In this way you can remove scopes completely and keep APIs consistent. Same thought should be given to all other APIs that expose or use `CoroutineScope` right now.

Comment (zibet27) [reply ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRtcDataChannel.kt]: I like the idea of implementing the `Channel` interface in `WebRtcDataChannel`. However, it would require two channels under the hood, one for sending and one for receiving messages, as they have different buffering requirements.

We could make an abstraction over buffering and remove `BufferedAmountLow` event and `bufferedAmountLowThreshold` in favour of channel capacity, but I'm not sure about this because it's different to the existing APIs.

Comment (Mr3zee) [reply ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRtcDataChannel.kt]: It's not like implementing `Channel`, but more like mimicking. We can just leave `send` and `receive` methods

File: ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRtcDataChannel.kt [hunk 5]
```diff
@@ -0,0 +1,250 @@
...
+    public var messagesReplay: Int = 10
+}
+
+/**
+ * Events that can be emitted by a WebRTC data channel.
+ *
+ * @see [MDN RTCDataChannel events](https://developer.mozilla.org/en-US/docs/Web/API/RTCDataChannel#events)
+ */
+public sealed interface DataChannelEvent {
+    public val dataChannel: WebRtcDataChannel
```
bjhham: I think `data` here is redundant, you can just call the property `channel`.

ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRtcDataChannel.kt
```diff
@@ -0,0 +1,250 @@
...
+    public class Error(override val dataChannel: WebRtcDataChannel, public val reason: String) : DataChannelEvent
+}
+
+/**
+ * Abstract class representing a network channel which can be used for bidirectional peer-to-peer transfers
+ * of arbitrary data. Every data channel is associated with an [WebRtcPeerConnection].
+ *
+ * @see [MDN RTCDataChannel](https://developer.mozilla.org/en-US/docs/Web/API/RTCDataChannel)
+ */
+public abstract class WebRtcDataChannel(
```
bjhham: I think it would be more consistent here to use the nested name-spacing.

Maybe this could be an interface instead of an abstract class too, and as @Mr3zee mentioned it could implement `Channel` for the messages.

So something like:
```kotlin
object WebRtc {
    interface DataChannel: Channel<DataChannel.Message> {
        data class Message()
        sealed interface Event {}
        enum class State {}
        // etc.
    }
}
```

ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRtc.kt
```diff
@@ -255,6 +258,56 @@ public object WebRtc {
...
+    public class DataChannelMessage private constructor(
+        /**
+         * The raw data buffer containing the message content.
+         */
+        public val data: Buffer,
+        /**
+         * Whether this message contains string data (true) or binary data (false).
+         */
+        public val isString: Boolean
+    ) {
```
bjhham: Looks like there's some extra fields on this message event:
- origin
- lastEventId
- source
- ports

We ought to replace the `Buffer` type on `data` with `ByteArray` since it should be a stateless representation of the data.  It's a little weird how the spec says it can be text or binary - imo we should just always have the raw bytes then let the consumers encode / decode based on their own contracts.  We can drop the `isString` here too in that case.

ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRtcDataChannel.kt
Mr3zee: Just to be clear, I **don't** think it should implement `Channel`. It should mimick it by providing familiar APIs like `send`, `trySend` and `receive`. Channel is an coroutines/stdlib entity, and just like it's discouraged to implement `Flow`, `Channel` also shouldn't be implemented

zibet27: I experimented with that a bit and found that we can't properly implement `SendChannel` because of the different buffering logic compared to how it is done in the `Channel`. But at the same time, it's logical to provide `ReceiveChannel` interface, which could be delegated by the `private val receiveChannel: Channel<WebRtc.DataChannelMessage>`. The messages here are emitted in the `onmessage` js callback (or the same callback in the Android lib).

ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRtc.kt
zibet27: These from message event fields are not provided in the `getstream-webrtc` and `webrtc-rs` because they are not very useful.
I agree, `ByteArray` could suits better here.
I believe `isString` field could make `receiveString()` (it's not committed yet) a bit more safe because when we map received message to `DataChannelMessage`(without `isString`), we lose the "type" information. Also, the `Message` in `getstream-webrtc` and `webrtc-rs` has `isString`(or its inverse).

ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRtcDataChannel.kt
zibet27: @Mr3zee @bjhham Do you think implementing `ReceiveChannel` by delegation is a good idea?

ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRtc.kt
zibet27: Anyway, I guess the sealed interface works better for `DataChannelMessage`. Please review.

ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRtcPeerConnection.kt
```diff
@@ -27,13 +29,15 @@ public abstract class WebRtcPeerConnection private constructor(
...
         if (config.statsRefreshRate > 0) {
-            launch {
-                while (true) {
-                    delay(config.statsRefreshRate)
-                    events.emitStats(getStatistics())
+            statsRefreshScope = CoroutineScope(coroutineContext).apply {
+                launch {
+                    while (true) {
+                        delay(config.statsRefreshRate)
+                        events.emitStats(getStatistics())
```
harrytmthy: If `getStatistics()` throws an exception, this coroutine will get cancelled silently, causing stats emission to stop, which might be confusing to some users. Consider catching the exception:

```kotlin
launch {
    while (true) {
        try {
            delay(config.statsRefreshRate)
            events.emitStats(getStatistics())
        } catch (e: CancellationException) {
            throw e // Respect coroutine cancellation
        } catch (e: Exception) {
            // Optional: log or emit an error event
        }
    }
}
```

Hope this helps 🙂

ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRtcDataChannel.kt
Mr3zee: I don't think it should implement any channel for the reasons given above

ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRtcPeerConnection.kt
zibet27: Hi! In the new commit, I've updated `WebRtcConnectionConfig` so it accepts a custom `coroutinesContext`, which could be `CoroutineExceptionHandler`. Do you think we should provide a method to restart stats collection or do it automatically?

ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRtcDataChannel.kt
```diff
@@ -0,0 +1,372 @@
...
+     */
+    public abstract fun send(bytes: ByteArray)
+
+    /**
+     * Suspends until a message is available in the data channel and returns it.
+     *
+     * This method will suspend the current coroutine until a message is received.
+     * The message can be either text or binary data.
+     */
+    public suspend fun receive(): Message = receiveChannel.receive()
```
Mr3zee: If these methods (`receive...`, `tryReceive...`) are not `final`, please make `receiveChannel` protected instead of `private`

zibet27: I made it private because I thought that classes that extend `WebRtcDataChannel` would need only `trySend` in the local channel, so I left protected methods like `emitMessage` and `stopReceivingMessages`. Should I mark `receive` methods you've mentioned as final, then?

bjhham (approved):

ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRtcDataChannel.kt
Mr3zee: Oh, sorry, that's my bad, they can't be overriden, I forgot in Kotlin and need to mark them explicitly. All good then

ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRtcPeerConnection.kt
harrytmthy: Great question! I think it depends on how resilient the stats collection is intended to be.

If it's meant to be long-running (e.g. like telemetry), I would lean toward restarting it automatically after non-cancellation exceptions, possibly logging or emitting an error event on the first failure. That way, users won't be surprised by stats stopping silently.

Alternatively, if we expect more advanced control or debugging scenarios, exposing a `restartStatsCollection()` method could be a good fallback.

Either way, I think you're on the right track by letting the coroutineContext carry the handler!

bjhham (approved):

e5l (approved):

PR #5011, opened 2025-07-14
KTOR-8523 Fix for empty part parse exception

bjhham:
**Subsystem**
Server, Core

**Motivation**
[KTOR-8523](https://youtrack.jetbrains.com/issue/KTOR-8523) Intermittent "ParserException: No colon in HTTP header" when parsing multipart request

**Solution**
It's technically valid to have empty multipart segments, but our parser dies when it sees the boundary instead of headers.  The fix is to check the input for the boundary before parsing headers.


e5l (approved):
lgtm

PR #5012, opened 2025-07-15
KTOR-8681 Fix for double initialization in DI

bjhham:
**Subsystem**
Server, DI

**Motivation**
[KTOR-8681](https://youtrack.jetbrains.com/issue/KTOR-8681) DI async duplicate resolution

**Solution**
Added an atomic reference and used lazy start for the deferred result so the create function can only be called once.


bjhham:
@osipxd `init()` should only be called a single time here because we're using a lazy coroutine start and the "missed" copy will never be awaited.

PR #5013, opened 2025-07-15
KTOR-6963 Fix for Darwin engine ws frame limit

bjhham:
**Subsystem**
Client, iOS, Websockets

**Motivation**
[KTOR-6963](https://youtrack.jetbrains.com/issue/KTOR-6963) Darwin: The `maxFrameSize` option has no effect

**Solution**
We were assuming that we could set `maxFrameSize` on the session after the upgrade, but for Darwin, the property needs to be assigned _before_ `NSURLSessionWebSocketTask.resume()` is called.  So to provide access to the config value, I included the plugin details as an attribute for WS requests.

### DIFF
--- a/ktor-client/ktor-client-darwin/darwin/src/io/ktor/client/engine/darwin/internal/DarwinSession.kt
+++ b/ktor-client/ktor-client-darwin/darwin/src/io/ktor/client/engine/darwin/internal/DarwinSession.kt
@@ -5,6 +5,7 @@
 package io.ktor.client.engine.darwin.internal
 
 import io.ktor.client.engine.darwin.*
+import io.ktor.client.plugins.websocket.WEBSOCKETS_KEY
 import io.ktor.client.request.*
 import io.ktor.utils.io.*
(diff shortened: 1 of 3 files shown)


ktor-client/ktor-client-darwin/darwin/src/io/ktor/client/engine/darwin/internal/DarwinSession.kt
```diff
@@ -32,6 +33,9 @@ internal class DarwinSession(
         val (task, response) = if (request.isUpgradeRequest()) {
             val task = session.webSocketTaskWithRequest(nativeRequest)
             val response = delegate.read(task, callContext)
+            val maxFrameSize = request.attributes[WEBSOCKETS_KEY].maxFrameSize
+            task.setMaximumMessageSize(maxFrameSize / 10)
```
e5l: could you write a comment behind the magic number?

e5l (changes requested):

e5l:
please check API dump before merging

PR #5014, opened 2025-07-16
Drop broken SSE test

bjhham:
This flaky test doesn't really test anything because there's no synchronization between the server and client for SSE, and there are other tests that ensure the messages arrive in order.


osipxd (approved):
👍

PR #5015, opened 2025-07-16
KTOR-8682 sanitize infinite loop

jimirocks:
**Subsystem**
ktor-io, server

**Motivation**
The infinite loop can cause resources wasting, also as the function is not cooperative it can't be stopped without stopping the JVM.

[KTOR-8682](https://youtrack.jetbrains.com/issue/KTOR-8682) Infinite loop in ByteReadChannel.readFully

**Solution**
Throwing EOFException as it seems to be in-line with other ByteReadChannel operations.


ktor-io/jvm/src/io/ktor/utils/io/ByteReadChannelOperations.jvm.kt
```diff
@@ -122,7 +122,9 @@ public suspend fun ByteReadChannel.skipDelimiter(delimiter: ByteString) {
 public suspend fun ByteReadChannel.readFully(buffer: ByteBuffer) {
     while (buffer.hasRemaining()) {
         if (availableForRead == 0) {
-            awaitContent()
+            if(!awaitContent()) {
+                throw EOFException("Not enough bytes available: expected ${buffer.remaining()} more bytes")
+            }
         }
```
bjhham: You can remove the outer if here since the read buffer is checked in `awaitContent` as well.

So:
```
while (buffer.hasRemaining()) {
    if (!awaitContent()) {
        throw EOFException("Not enough bytes available: expected ${buffer.remaining()} more bytes")
    }
    ...
}
```

ktor-io/jvm/src/io/ktor/utils/io/ByteReadChannelOperations.jvm.kt
```diff
@@ -122,7 +122,9 @@ public suspend fun ByteReadChannel.skipDelimiter(delimiter: ByteString) {
 public suspend fun ByteReadChannel.readFully(buffer: ByteBuffer) {
```
bjhham, line 128: Could you also add some KDoc while you're here?  It's not very obvious that "fully" here refers to the byte buffer contents.

ktor-io/jvm/src/io/ktor/utils/io/ByteReadChannelOperations.jvm.kt
jimirocks: good catch - will do it

ktor-io/jvm/src/io/ktor/utils/io/ByteReadChannelOperations.jvm.kt
jimirocks: sure

jimirocks:
@bjhham I amended what you desired hopefully

bjhham (approved):

bjhham:
Please rebase your branch so we can merge.

jimirocks:
@bjhham rebased

PR #5020, opened 2025-07-26
KTOR-8712 add .yml as allowed suffix for configuration

froks:
**Subsystem**
Server Yaml Configuration 

**Motivation**
When using `-config=application.yml` to specify your configuration file, the specified configuration will silently not load, this is very confusing, since yml is a very common suffix.

The reason is, that only configuration files ending in `.yaml` are loaded.

**Solution**
Adds `.yml` as an allowed suffix for yaml configuration files.


froks:
The failing tests are unrelated to these changes

osipxd:
@froks, could you rebase the PR onto recent main?

froks:
> @froks, could you rebase the PR onto recent main?

rebased and force pushed it

PR #5031, opened 2025-07-30
KTOR-7075: Zstd compression support

tre3p:
**Subsystem**
ktor-utils module, Compression plugin

**Motivation**
This PR solves KTOR-7075 ticket by adding support for zstd compression algorithm to Compression plugin.

**Solution**
I've used [zstd-jni library](https://github.com/luben/zstd-jni) which provides JNI binding to zstd compression algorithm implementation.


tre3p:
@bjhham @osipxd Hey guys! Recreated PR cause of previous was closed

bjhham (approved):

bjhham:
Hey, sorry I tried to merge this a while back and forgot about it.  I'll ensure it goes through now and it will go into 3.4.

ApoloApps:
what's the ETA for 3.4? more or less

bjhham:
Should be released on the 17th of December.  We also have nightly [EAP builds](https://ktor.io/eap/) you can use if you'd like to try this out.

05nelsonm:
Question.

Won't this cause issue for Android consumers, given the `ktor-utils` module is JVM only, is an API dependency for pretty much all other ktor modules, and does not have the necessary `@aar` dependency as noted in the `zstd-jni` documentation?

Additionally, won't the +17.5mb of native desktop libs in the `ztsd-jni` jar be included in Android builds as dead weight, blowing up android consumer's apk sizes?

EDIT: Wouldn't it also be better as a `compileOnly` dependency and lazy loading + `throws UnsupportedOperationException` on load failure, with documentation stating that API consumers need to add the dependency for JVM or android if they want support for it?

EDIT2: Also, doesn't this PR unnecessarily break API compatibility by adding a constructor parameter to `io/ktor/server/plugins/compression/CompressionEncoderConfig` which is a data class?

bjhham:
Thanks @05nelsonm, we'll look at mitigating this before the release 🙏

PR #5032, opened 2025-07-30
Fix hits related to ^KT-78304

lunakoly:
Otherwise, `UPPER_BOUNDS_VIOLATED` is now reported.

**Subsystem**
Server: `ktor-server-core`.

**Motivation**
Makes our user projects build pass.

**Solution**
Recently we found and fixed some false negative `UPPER_BOUND_VIOLATED` diagnostics (see [KTLC-358](https://youtrack.jetbrains.com/issue/KTLC-358)), and began showing deprecation warnings for the related cases.


ktor-server/ktor-server-core/common/src/io/ktor/server/application/ApplicationPlugin.kt
```diff
@@ -195,7 +195,7 @@ private fun Route.copyChildrenRecursively(child: RoutingNode) {
     }
 }
 
-private fun <B : Any, F : Any, TSubject, TContext, P : Pipeline<TSubject, TContext>> P.addAllInterceptors(
+private fun <B : Any, F : Any, TSubject: Any, TContext: Any, P : Pipeline<TSubject, TContext>> P.addAllInterceptors(
```
cypressious:
```suggestion
private fun <B : Any, F : Any, TSubject : Any, TContext : Any, P : Pipeline<TSubject, TContext>> P.addAllInterceptors(
```

woainikk:
I've cherry picked the commit, thank you

PR #5044, opened 2025-08-14
WebRtc Client. Introduce Native engine based on WebRTC.rs

zibet27:
**Subsystem**
WebRTC Client

**Motivation**
Sometimes, we need only WebRTC data channels without capturing and sending any video or audio streams. In this case, almost all targets could be easily supported (native, JVM, etc).

**Solution**
- I have introduced [Gobley](https://gobley.dev/docs/) into the project, which allows mixing Rust and Kotlin code via [Uniffi](https://github.com/mozilla/uniffi-rs). 
- On the Rust side, I've made a wrapper over [WebRTC.rs](https://github.com/webrtc-rs/webrtc). 
- The data channel communication part works well without major differences compared to the browser API. 
- There is no default `MediaTrackFactory` provided.
- High-level `MediaStream API` abstractions (without much protocol-specific logic) are included for compatibility with Ktor's WebRTC Client common API. They are rather mock than ready-to-use and require further testing.
- Only JVM is supported yet because of cross-compilation issues I've encountered, but native should also be added soon.
- The common data channel API was a bit updated to separate the closing of the data stream and cleaning resources with `Autoclosable`
- Tests are copied from the main WebRTC module. Renegotiation tests were removed because of a bug in WebRTC.rs not firing the `negotiation_needed` event.

### DIFF
--- a/ktor-client/ktor-client-webrtc/ktor-client-webrtc-rs/api/ktor-client-webrtc-rs.klib.api
+++ b/ktor-client/ktor-client-webrtc/ktor-client-webrtc-rs/api/ktor-client-webrtc-rs.klib.api
@@ -1,5 +1,5 @@
 // Klib ABI Dump
-// Targets: [iosArm64, iosSimulatorArm64, iosX64, linuxArm64, linuxX64, macosArm64, macosX64, mingwX64]
+// Targets: [iosArm64, iosSimulatorArm64, iosX64, linuxArm64, linuxX64, macosArm64, mingwX64]
 // Rendering settings:
 // - Signature version: 2
 // - Show manifest properties: true
--- a/ktor-client/ktor-client-webrtc/ktor-client-webrtc-rs/gradle.properties
+++ b/ktor-client/ktor-client-webrtc/ktor-client-webrtc-rs/gradle.properties
@@ -2,7 +2,13 @@
 # Copyright 2014-2025 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
 #
 target.js=false
-target.tvos=false
-target.watchos=false
 target.wasmJs=false
+# not suppported by rust
+target.tvos=false
+# not suppported by rust
+target.watchos=false
+# macosX64 is disabled for now because of buidling issues
+target.macosX64=false
+# android native compilation with Gobley requires AGP and NDK,
+# so `ktor-client-webrtc` is recommended
 target.androidNative=false


ktor-client/ktor-client-webrtc/android/src/io/ktor/client/webrtc/DataChannel.kt
```diff
@@ -116,7 +116,7 @@ public class AndroidWebRtcDataChannel(
 
             override fun onBufferedAmountChange(previousAmount: Long) {
                 coroutineScope.launch(start = CoroutineStart.UNDISPATCHED) {
-                    if (previousAmount > bufferedAmountLowThreshold && bufferedAmountLowThreshold > bufferedAmount) {
+                    if (bufferedAmountLowThreshold in (bufferedAmount + 1)..<previousAmount) {
                         return@launch
                     }
```
bjhham: Is this logic right?  It seems a little odd to be checking that the low threshold is between the two recorded buffer amounts.  I would have thought it'd be `bufferedAmount - previousAmount > lowThreshold`...

ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRtcPeerConnection.kt
```diff
@@ -118,6 +120,15 @@ public abstract class WebRtcPeerConnection private constructor(
      */
     public abstract suspend fun getStatistics(): List<WebRtc.Stats>
 
+    public suspend fun awaitIceGatheringComplete() {
+        if (iceGatheringState.value == WebRtc.IceGatheringState.COMPLETE) {
+            return
+        }
+        iceGatheringState
+            .filter { it == WebRtc.IceGatheringState.COMPLETE }
+            .first() // Suspends until the first "COMPLETE" is emitted
```
bjhham: Can simplify:

```suggest
iceGatheringState.first { it == WebRtc.IceGatheringState.COMPLETE }
```

bjhham: Might also want some exception handling in case the `iceGatheringState` flow closes before complete is sent.

ktor-client/ktor-client-webrtc/ktor-client-webrtc-rs/common/rust/media/sink.rs
```diff
@@ -0,0 +1,161 @@
...
+use crate::rtc::RtcError::MediaTrackError;
+use std::io::{Seek, SeekFrom, Write};
+use std::sync::Arc;
+use tokio::sync::Mutex;
+use webrtc::media::io::h264_writer::H264Writer;
+use webrtc::media::io::ivf_reader::{IVF_FILE_HEADER_SIZE, IVFFileHeader};
+use webrtc::media::io::ivf_writer::IVFWriter;
+use webrtc::media::io::ogg_writer::OggWriter;
+
+/// Custom writer that captures data and forwards to the handler
```
bjhham:
```suggest
// Custom writer that captures data and forwards to the handler
```

ktor-client/ktor-client-webrtc/ktor-client-webrtc-rs/common/rust/media/sink.rs
```diff
@@ -0,0 +1,161 @@
...
+            version: 0,
+            header_size: IVF_FILE_HEADER_SIZE as u16,
+            four_cc: *b"VP80", // VP8 FourCC
+            width: 640,
+            height: 480,
+            timebase_denominator: 30, // 30 FPS default
+            timebase_numerator: 1,
+            num_frames: 0, // Unknown for streaming
+            unused: 0,
+        };
```
bjhham: Seems like these properties ought to come from somewhere.

ktor-client/ktor-client-webrtc/ktor-client-webrtc-rs/common/rust/media/track.rs
```diff
@@ -0,0 +1,256 @@
...
+use webrtc::track::track_local::track_local_static_sample::TrackLocalStaticSample;
+use webrtc::track::track_remote::TrackRemote;
+
+impl MediaStreamTrack {
+    pub fn from_local(track: Arc<TrackLocalStaticSample>) -> Self {
+        let kind = if track.codec().mime_type.starts_with("audio/") {
+            MediaKind::Audio
+        } else {
+            MediaKind::Video
+        };
```
bjhham: This looks like it ought to be a lookup for "audio", "video", or some unrecognized media error.

ktor-client/ktor-client-webrtc/ktor-client-webrtc-rs/common/rust/media/sink.rs
zibet27: In Rust, the `///` is called line doc comments, though I'm still not sure if it's needed here.

bjhham:
@coderabbitai review

ktor-client/ktor-client-webrtc/android/src/io/ktor/client/webrtc/DataChannel.kt
zibet27: MDN: 
A bufferedamountlow event is sent to an [RTCDataChannel](https://developer.mozilla.org/en-US/docs/Web/API/RTCDataChannel) when the number of bytes currently in the outbound data transfer buffer ([bufferedAmount](https://developer.mozilla.org/en-US/docs/Web/API/RTCDataChannel/bufferedAmount)) falls from above to below or equal the threshold specified in [bufferedAmountLowThreshold](https://developer.mozilla.org/en-US/docs/Web/API/RTCDataChannel/bufferedAmountLowThreshold).

So, we should fire the event only when the previous amount was bigger than the threshold and the current amount is lower than or equal to the threshold.
You are right, the logic and the corresponding test were incorrect. But the problem with logic is that it should be negated, and `+ 1` is not needed.

ktor-client/ktor-client-webrtc/ktor-client-webrtc-rs/build.gradle.kts
```diff
@@ -0,0 +1,40 @@
...
+ */
+import gobley.gradle.GobleyHost
+import gobley.gradle.cargo.dsl.jvm
+
+description = "Ktor WebRTC Engine based on the WebRTC.rs and Gobley"
+
+plugins {
+    id("dev.gobley.cargo") version "0.3.1"
+    id("dev.gobley.uniffi") version "0.3.1"
+    kotlin("plugin.atomicfu") version libs.versions.kotlin.get()
```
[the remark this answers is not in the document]
zibet27: Cargo plugin depends on `atomicfu`

ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRtcPeerConnection.kt
zibet27: `iceGatheringState` is a `StateFlow`, so afaik it will never end?

ktor-client/ktor-client-webrtc/android/src/io/ktor/client/webrtc/DataChannel.kt
bjhham: Oh I see, the `previousAmount` is there to ensure we don't get duplicate events when the buffer remains low.

I think I'd write it like this:

```kotlin
// guard against duplicate events
if (previousAmount < bufferedAmountLowThreshold)
    return@launch
// buffer has dropped below threshold
if (bufferedAmount < bufferedAmountLowThreshold) {
    val event = DataChannelEvent.BufferedAmountLow(this@AndroidWebRtcDataChannel)
    eventsEmitter.emitDataChannelEvent(event)
}
```

This way we keep the subject (variable) of the comparison on the left-hand side and we break up the two concepts.  It's a bit more verbose but more natural for reading.

ktor-client/ktor-client-webrtc/ktor-client-webrtc-rs/common/rust/connection.rs
```diff
@@ -0,0 +1,242 @@
...
+                    let can_receive = match transceiver.current_direction() {
+                        RTCRtpTransceiverDirection::Sendonly => false,
+                        RTCRtpTransceiverDirection::Inactive => false,
+                        _ => true,
+                    };
+                    if !can_receive {
+                        observer_ref.on_remove_track(track_ref.clone());
+                    }
+                    Box::pin(async {})
+                });
```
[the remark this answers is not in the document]
zibet27: There are not many options there to subscribe to track removal on the other peer (not our track removals). 
This event could be identified by the change of track direction after renegotiation, but it doesn't happen immediately. 
The possible solution is checking all received tracks multiple times with some interval, but I don't think it is good enough and `on_mute` event seems to be a better indicator. 
Lmk if you know better options.

ktor-client/ktor-client-webrtc/ktor-client-webrtc-rs/common/rust/senders.rs
```diff
@@ -0,0 +1,143 @@
...
+
+impl RtpEncodingParameters {
+    pub fn from_native(params: RTCRtpCodingParameters) -> Self {
+        Self {
+            rid: params.rid.to_string(),
+            ssrc: params.ssrc,
+            payload_type: params.payload_type,
+            rtx: params.rtx.ssrc,
+        }
+    }
```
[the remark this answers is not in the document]
zibet27: It is not optional

bjhham:
Could you also include build instructions in the readme / contributors guide?

It looks like the builds are failing for native targets even when they're disabled due to the missing cargo binary.

ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRtc.kt
```diff
@@ -458,16 +458,23 @@ public object WebRtc {
...
          */
-        override fun close()
+        public fun closeTransport()
+
+        /**
+         * Closes the data channel and releases all associated resources.
+         * Automatically invokes `closeTransport`.
+         * Accessing the channel after this operation could throw an exception.
+         */
+        override fun close(): Unit = closeTransport()
```
e5l:
```suggestion
        override fun close() {
            closeTransport()
        }
```

ktor-client/ktor-client-webrtc/jsAndWasmShared/src/io/ktor/client/webrtc/DataChannel.kt
```diff
@@ -63,9 +63,7 @@ public class JsWebRtcDataChannel(
         channel.bufferedAmountLowThreshold = threshold.toInt()
     }
 
-    override fun close() {
-        channel.close()
-    }
+    override fun closeTransport(): Unit = channel.close()
```
e5l:
```suggestion
    override fun closeTransport() {
        channel.close()
    }
```

zibet27:
> Please check the CI

Working on that. CI doesn't have the Rust toolchain and Android NDK installed.

ktor-client/ktor-client-webrtc/ktor-client-webrtc-rs/build.gradle.kts
```diff
@@ -49,15 +49,55 @@ kotlin {
...
+tasks.named { it == "cinteropRustLinuxArm64" }.configureEach {
     onlyIf { GobleyHost.Platform.Linux.isCurrent }
 }
-tasks.named { it == "cinteropRustLinuxArm64" }.configureEach {
+tasks.named { it == "cargoBuildLinuxArm64Release" }.configureEach {
     onlyIf { GobleyHost.Platform.Linux.isCurrent }
 }
 tasks.named { it == "cinteropRustLinuxX64" }.configureEach {
     onlyIf { GobleyHost.Platform.Linux.isCurrent }
 }
```
osipxd: You could declare sets with linux-only, windows-only and macos-only tasks. Like this:
```kotlin
val linuxOnlyTasks = setOf(
    "cargoBuildLinuxX64Release",
    "cinteropRustLinuxArm64",
    "cargoBuildLinuxArm64Release",
    "cinteropRustLinuxX64",
)

tasks.named { it in linuxOnlyTasks }.configureEach {
    onlyIf { GobleyHost.Platform.Linux.isCurrent }
}
```
another option is to depend on task name, and this can be even more flexible
```kotlin
tasks.named { it.startsWith("cargoBuildLinux") || it.startsWith("cinteropRustLinux") }.configureEach {
    onlyIf { GobleyHost.Platform.Linux.isCurrent }
}
```

settings.gradle.kts
```diff
@@ -128,7 +128,13 @@ projects {
         +"ktor-client-tests"
 
         +"ktor-client-webrtc" including {
-            +"ktor-client-webrtc-rs"
+            // Include `ktor-client-webrtc-rs` if rust compilation is enabled in `gradle.properties` for local builds
+            // or if the `RUST_COMPILATION` environment variable is set to `true` for the CI.
+            val compileRust = providers.gradleProperty("ktorbuild.rustCompilation").orNull.toBoolean() ||
+                System.getenv("RUST_COMPILATION").toBoolean()
```
osipxd: Let's refactor the code this way and add prefix `KTOR_` to the env variable
```kotlin
providers.environmentVariable("KTOR_RUST_COMPILATION")
    .orElse(providers.gradleProperty("ktorbuild.rustCompilation"))
    .orNull.toBoolean()
```

bjhham:
Looks like `WebRtcMediaTest` could be a little flaky on TC.

zibet27:
> Looks like `WebRtcMediaTest` could be a little flaky on TC.

@bjhham Yes, it is. I am thinking about the improvement. The nature of this test is flaky because we don't know exactly the number of frames after which the `AddTrack` event will be emitted, though we could try sending frames until it happens (now it's just 10). What do you think?

bjhham:
@zibet27 looks like it finishes quickly when it passes, so it could be a problem with the connection or something happening out of sequence.  It may help to add some logging.

e5l (changes requested):
Hey @zibet27, thanks for the PR. LGTM. Please check the failing tests

PR #5046, opened 2025-08-20
KTOR-8781 Fix Application.propertyAsNull() to ensure parity with Application.property()

solonovamax:
**Subsystem**
Server (core)

**Motivation**
Fix for [KTOR-8781](https://youtrack.jetbrains.com/issue/KTOR-8781)

**Solution**
Specify the generic explicitly


bjhham (approved):
Thank you for the fix!  Looks good 👍

PR #5054, opened 2025-08-22
KTOR-6700 & KTOR-5836 Serve static resources with Etag and LastModified headers, support for auto-generating ETag

marychatte:
**Subsystem**
Server, StaticContent

**Motivation**
[KTOR-6700](https://youtrack.jetbrains.com/issue/KTOR-6700)
[KTOR-5836](https://youtrack.jetbrains.com/issue/KTOR-5836)

**Solution**
1) Add support to configure [`etag`](https://www.rfc-editor.org/rfc/rfc9110#name-etag) and `lastModified` in `StaticContentConfig`:
```kotlin
staticFiles("/static", filesDir) {
    etag { resource -> ... }
    lastModified { resource -> ... }
}
```
Requires the `ConditionalHeaders` plugins to work

2) Add predifined etag generation function:
```kotlin
staticFiles("/static", filesDir) {
    etag(ETagProvider.StrongSha256)
}
```
It calculates the ETag based on the SHA256 of the content of the resource.


ktor-http/common/src/io/ktor/http/content/Versions.kt
```diff
@@ -213,18 +214,27 @@ public data class EntityTagVersion(val etag: String, val weak: Boolean) : Versio
...
+     * Checks whether two entity-tags match (weak).
+     */
+    private fun weakMatch(other: EntityTagVersion): Boolean {
+        if (this == STAR || other == STAR) return true
+        return opaque == other.opaque
     }
 
     /**
-     * Specifies `If-None-Match` logic using the [match] function.
+     * Specifies `If-None-Match` logic using the [weakMatch] function.
```
osipxd: Probably it's better not to mention the `weakMatch` function here as it is not a part of the public API

ktor-http/common/src/io/ktor/http/content/Versions.kt
```diff
@@ -175,12 +174,14 @@ public fun EntityTagVersion(spec: String): EntityTagVersion {
...
  */
 public data class EntityTagVersion(val etag: String, val weak: Boolean) : Version {
-    private val normalized: String = when {
+    private val opaque: String = when {
         etag == "*" -> etag
         etag.startsWith("\"") -> etag
         else -> etag.quote()
     }
 
+    private val normalized: String = if (weak) "W/$opaque" else opaque
```
osipxd, line 183: I saw a comment from the rabbit that the wildcard `*` should not be prefixed with `W/`. Wasn't he right?

ktor-http/common/src/io/ktor/http/content/Versions.kt
```diff
@@ -213,18 +214,27 @@ public data class EntityTagVersion(val etag: String, val weak: Boolean) : Versio
      */
     public fun match(other: EntityTagVersion): Boolean {
         if (this == STAR || other == STAR) return true
-        return normalized == other.normalized
+        if (weak || other.weak) return false
+        return opaque == other.opaque
```
osipxd: We could reuse `weakMatch` logic here like this
```kotlin
if (weak || other.weak) return false
return weakMatch(other)
```

osipxd: And probably we should add a check in the `init { }` block that `weak` is not `true` when `etag == STAR`

ktor-server/ktor-server-core/jvm/src/io/ktor/server/http/content/EntityTagMode.kt
```diff
@@ -0,0 +1,95 @@
...
+            streamSupplier().use { ins ->
+                val hex = sha256Hex(ins)
+                EntityTagVersion(hex, weak = false)
+            }
+        }
+    }
+
+    private fun keyAndStreamSupplier(resource: Any): Pair<String, () -> InputStream> = when (resource) {
+        is File -> {
+            "S:F:${resource.absolutePath}:${resource.lastModified()}:${resource.length()}" to {
```
osipxd: Is it some conventional notation for generated ETags? I failed to find any conventions, so I'm curious

ktor-server/ktor-server-core/jvm/src/io/ktor/server/http/content/EntityTagMode.kt
```diff
@@ -0,0 +1,95 @@
...
+ * routing {
+ *   staticFiles("/assets", File("assets")) {
+ *     etag(EntityTagMode.StrongSha256)
+ *   }
+ * }
+ * ```
+ *
+ * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.server.http.content.EntityTagMode)
+ */
+public object EntityTagMode {
```
osipxd: Maybe name it `EntityTagProvider`?
The only concern I have is that this name fits more for an interface. Probably `EntityTagProviders` or `EntityTagStrategies` then :D

Or we could introduce a functional interface with a companion object. Like this:

```kotlin
fun interface EntityTagProvider {

    fun provide(resource: Any): EntityTagVersion?

    companion object {
        val StrongSha256 = EntityTagProvider { ... }
}
```

However it's not a strong recommendation, more like feeling. You can keep it "as is" if you want

osipxd: Maybe we can use `ETag` instead of `EntityTag` here. I know we already have `EntityTagVersion`, but it is not wide-used public API and here the situation is different. `ETag` might be more recognizable given the name `etag` already used for the function configuring ETag generation

ktor-server/ktor-server-core/jvm/src/io/ktor/server/http/content/EntityTagMode.kt
```diff
@@ -0,0 +1,95 @@
...
+     * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.server.http.content.EntityTagMode.StrongSha256)
+     */
+    public val StrongSha256: (Any) -> EntityTagVersion? = { resource ->
+        val (key, streamSupplier) = keyAndStreamSupplier(resource)
+        etagCache.computeIfAbsent(key) {
+            streamSupplier().use { ins ->
+                val hex = sha256Hex(ins)
+                EntityTagVersion(hex, weak = false)
+            }
+        }
```
osipxd: Should we handle IO exceptions here?

ktor-server/ktor-server-tests/jvm/test/io/ktor/server/plugins/StaticContentTest.kt
```diff
@@ -1444,6 +1452,110 @@ class StaticContentTest {
...
+        expectedDate: GMTDate,
+        acceptEncoding: String? = null
+    ) {
+        val response = client.get(url) {
+            headers {
+                acceptEncoding?.let { append(HttpHeaders.AcceptEncoding, acceptEncoding) }
+            }
+        }
+        assertEquals(HttpStatusCode.OK, response.status)
+        val etag = response.headers[HttpHeaders.ETag] ?: error("no ETag")
```
osipxd: In tests it's better to use `fail`. The test will be marked as "Failed" instead of "Exited exceptionally" in the IDE
```suggestion
        val etag = response.headers[HttpHeaders.ETag] ?: fail("no ETag")
```

ktor-http/common/src/io/ktor/http/content/Versions.kt
marychatte: Yes, he is right. But `*` is marked with `weak=false`, so it can't be `"W/*"`:
```kotlin
public val STAR: EntityTagVersion = EntityTagVersion("*", false)
```

ktor-http/common/src/io/ktor/http/content/Versions.kt
marychatte: Usually, EntityTag is created using the  `parse` function, where we set `weak=false` for star:
```kotlin
public fun parseSingle(value: String): EntityTagVersion {
    if (value == "*") return STAR
```
But I added a check in init for the cases when it's created via the constructor

ktor-server/ktor-server-core/jvm/src/io/ktor/server/http/content/EntityTagMode.kt
marychatte: As far as I know, there's no conventional, standardized notation for generated ETags. The idea was to provide a variant of generation for strong ETags since weak ETags may be based on metadata and can miss some changes. Actually, they are based on the logic of the application, so there is no general option to generate weak ETags.
On the other hand, strong ETag must change on any byte-for-byte change of the representation, so calculating them based on the sha256 of the resource's content is one option.
Here `S:F:${resource.absolutePath}:${resource.lastModified()}:${resource.length()}` is just an internal namespace to avoid collisions across resource types, not a public convention. `s` is for `strong`, `f` is for `files` 😅

ktor-server/ktor-server-core/jvm/src/io/ktor/server/http/content/EntityTagMode.kt
marychatte: Thanks, I find the `ETagProvider` name the best :)

ktor-server/ktor-server-core/jvm/src/io/ktor/server/http/content/EntityTagMode.kt
osipxd: Got it! Do we need to distinct between different types (Path/File/Resource)? Probably if server changes route `staticFiles` to `staticFileSystem`, but files remain the same, we should reuse existing cache. Am I correct?
Also, can we drop the metadata (path, modified date and file length) if we anyways calculate file content sha256? I mean if a file has been moved/renamed, but the content remains the same it should be okay to use the same ETag.

However I'm not very familiar with the ETag concept, so I might be missing something.

ktor-server/ktor-server-core/jvm/src/io/ktor/server/http/content/EntityTagMode.kt
marychatte: Yes, thank you. I suggest catching the exception, logging it as a warning, and proceeding with no ETag

ktor-server/ktor-server-core/jvm/src/io/ktor/server/http/content/EntityTagMode.kt
marychatte: > Do we need to distinct between different types (Path/File/Resource)? Probably if server changes route staticFiles to staticFileSystem, but files remain the same, we should reuse existing cache. Am I correct?

You are correct, we can reuse the cache, so I will drop the prefix "S:F..." etc

> Also, can we drop the metadata (path, modified date and file length) if we anyways calculate file content sha256?

We use the key only for internal memoization, which avoids re-hashing on every request. If we remove it, we could serve a stale ETag after the file changes because the cached value would never be invalidated.

ktor-server/ktor-server-tests/common/test/io/ktor/tests/server/plugins/ConditionalHeadersTests.kt
```diff
@@ -157,4 +157,11 @@ class ETagsTest {
         }
         assertEquals(HttpStatusCode.PreconditionFailed, result.status)
     }
+
+    @Test
+    fun testInvalidWeakStarETag() = withConditionalApplication {
+        assertFailsWith<IllegalArgumentException> {
+            EntityTagVersion(EntityTagVersion.STAR.etag, weak = true)
+        }
+    }
```
osipxd: Minor. Probably `withConditionalApplication` can be removed
```suggestion
    fun testInvalidWeakStarETag() {
        assertFailsWith<IllegalArgumentException> {
            EntityTagVersion(EntityTagVersion.STAR.etag, weak = true)
        }
    }
```

osipxd (approved):
👍

PR #5058, opened 2025-08-27
KTOR-8809 Fix the use of DSL markers

osipxd:
**Subsystem**
Client, Server

**Motivation**
- [KTOR-8809](https://youtrack.jetbrains.com/issue/KTOR-8809) Fix the use of DSL markers
- [KTOR-8667](https://youtrack.jetbrains.com/issue/KTOR-8667) Regression in Ktor 3.0 routing DSL breaking nested routing calls due to missing DSL marker annotations
- [KTOR-8750](https://youtrack.jetbrains.com/issue/KTOR-8750) Jetty Jakarta: Server unable to start listening for connections since 3.2.0
- [KTOR-8501](https://youtrack.jetbrains.com/issue/KTOR-8501) Application scope is not segregated from routing handlers

**Solution**
- Add `@KtorDsl` to the `Route` and `RoutingContext` to fix routing DSL
- Add `@KtorDsl` to the `ApplicationEngine.Configuration` and `EngineConnectorBuilder`
- Revise all usages of `@KtorDsl` to ensure they're correct

_These changes are breaking so we should discuss if this PR should be merged at once, or we should extract only some of the changes. It's better to review commit-by-commit._

### DIFF
--- a/ktor-io/common/src/io/ktor/utils/io/Annotations.kt
+++ b/ktor-io/common/src/io/ktor/utils/io/Annotations.kt
@@ -108,5 +108,5 @@ public annotation class PublicAPICandidate(val version: String)
  * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.utils.io.KtorDsl)
  */
 @DslMarker
-@Target(AnnotationTarget.CLASS, AnnotationTarget.TYPEALIAS, AnnotationTarget.TYPE, AnnotationTarget.FUNCTION)
+@Target(AnnotationTarget.CLASS, AnnotationTarget.TYPEALIAS, AnnotationTarget.TYPE)
 public annotation class KtorDsl


ktor-server/ktor-server-core/common/src/io/ktor/server/routing/RoutingRoot.kt
```diff
@@ -28,7 +28,6 @@ internal val LOGGER = KtorSimpleLogger("io.ktor.server.routing.Routing")
  *
  * @param application is an instance of [Application] for this routing node.
  */
-@KtorDsl
```
osipxd, line 31: `RoutingNode` already has the annotation

ktor-server/ktor-server-jetty-jakarta/jvm/src/io/ktor/server/jetty/jakarta/JettyApplicationEngineBase.kt
```diff
@@ -42,15 +43,15 @@ public open class JettyApplicationEngineBase(
          *
          * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.server.jetty.jakarta.JettyApplicationEngineBase.Configuration.configureServer)
          */
-        public var configureServer: Server.() -> Unit = {}
+        public var configureServer: (@KtorDsl Server).() -> Unit = {}
```
osipxd, line 47: This way the DSL scope works on Kotlin 2.2

ktor-server/ktor-server-plugins/ktor-server-htmx/common/src/io/ktor/server/htmx/HxRouting.kt
```diff
@@ -23,19 +23,14 @@ public val Route.hx: HxRoute get() = HxRoute.wrap(this)
  * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.server.htmx.hx)
  */
 @ExperimentalKtorApi
-public fun Route.hx(configuration: HxRoute.() -> Unit): Route = with(HxRoute.wrap(this)) {
-    header(HxRequestHeaders.Request, "true") {
-        configuration()
-    }
-}
+public fun Route.hx(configuration: HxRoute.() -> Unit): Route = hx.apply(configuration)
```
osipxd, line 26: DSL scope has helped to find incorrect code. Nice!

ktor-server/ktor-server-core/common/src/io/ktor/server/routing/RoutingRoot.kt
```diff
@@ -173,13 +172,18 @@ public val Route.application: Application
         )
     }
 
+/**
+ * Shortcut to get an [Application] for this [RoutingContext].
+ */
+public val RoutingContext.application: Application
```
osipxd: Added this to keep usage of `application` working inside `RoutingScope`

ktor-server/ktor-server-core/common/src/io/ktor/server/routing/RoutingNode.kt
```diff
@@ -257,7 +258,7 @@ public class RoutingCall internal constructor(
 @KtorDsl
 public class RoutingContext(
     public val call: RoutingCall
-)
+) : CoroutineScope by call
```
osipxd: We have some calls of `launch` and `async` in `RoutingContext` in tests. Actually they're implicitly called on the `Application` instance. There are two ways of what to do:
1. Make `RoutingContext` implement `CoroutinesScope` to keep these calls working, but behavior might change (for better, I believe). This would also fix [KTOR-8501](https://youtrack.jetbrains.com/issue/KTOR-8501)
2. Refactor these calls to `call.launch { ... }` and `call.async { ... }`. Breaks existing code, but explicitly declares behavior change.

ktor-server/ktor-server-plugins/ktor-server-di/common/test/io/ktor/server/plugins/di/DependencyInjectionTest.kt
```diff
@@ -390,15 +381,15 @@ class DependencyInjectionTest {
...
+                    val service: GreetingService = application.dependencies.resolve()
                     call.respondText(service.hello())
                 }
                 get("/balance") {
                     val service: BankService = bankService.await()
                     call.respondText(service.balance().toString())
                 }
                 get("/bank-teller") {
-                    val service: BankTeller = dependencies.resolve()
+                    val service: BankTeller = application.dependencies.resolve()
```
osipxd, line 392: I don't know the best way to fix these ones. We can either resolve dependencies at the application level, or use `application.depenencies` in routing context

ktor-server/ktor-server-tests/common/test/io/ktor/tests/server/plugins/CORSTest.kt
```diff
@@ -1572,24 +1572,4 @@ class CORSTest {
...
-    fun routeInsideCorsPluginConfig() = testApplication {
-        routing {
-            install(CORS) {
-                route("/abc") {
-                    get {
-                        call.respond("OK")
-                    }
-                }
-            }
-        }
```
osipxd, line 1586: This is not valid anymore. Declaring route inside Plugin configuration will throw a compilation error

ktor-server/ktor-server-core/common/src/io/ktor/server/routing/RoutingNode.kt
osipxd: Okay. The current implementation of the first option is problematic. Calling `coroutineContext` on `RoutingContext` returns different `CoroutineContext` which doesn't contain some coroutine elements. So probably it's better to go with the second option, which is more breaking.

bjhham: Yeah option 2 would be best.  If I recall correctly, one of the intentional breaking changes in 3.0 was to force the explicit `call.launch` instead of having the scope in the receiver, but with the broken DSL markers it's easy to accidentally launch from the application scope 😜

ktor-server/ktor-server-plugins/ktor-server-di/common/test/io/ktor/server/plugins/di/DependencyInjectionTest.kt
bjhham: Prefixing with `application` is best here.

ktor-server/ktor-server-tests/common/test/io/ktor/tests/server/plugins/CORSTest.kt
bjhham: @Stexxe overzealous testing?

Stexxe: Maybe

ktor-server/ktor-server-plugins/ktor-server-status-pages/common/test/io/ktor/server/plugins/statuspages/StatusPagesTest.kt
```diff
@@ -293,10 +293,10 @@ class StatusPagesTest {
 
             routing {
                 get("/fail") {
-                    async { throw AsyncFailedException() }.await()
+                    call.async { throw AsyncFailedException() }.await()
```
osipxd: It seems this test is actually broken.
Failing inside `call.async` terminates the whole call `CoroutineScope` and even if we catch this exception the scope itself is cancelled and calling `call.respondText(...)` from handler re-throws the exception

osipxd: I haven't managed to fix this problem quickly, so I created an issue: [KTOR-8824](https://youtrack.jetbrains.com/issue/KTOR-8824)

ktor-server/ktor-server-core/common/src/io/ktor/server/engine/ApplicationEngine.kt
```diff
@@ -21,6 +23,7 @@ public interface ApplicationEngine {
      * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.server.engine.ApplicationEngine.Configuration)
      */
     @Suppress("MemberVisibilityCanBePrivate")
+    @KtorDsl
     public open class Configuration {
         /**
```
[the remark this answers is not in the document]
osipxd: Unfortunately it lives in `io.ktor.utils.io`

ktor-server/ktor-server-test-host/common/src/io/ktor/server/testing/TestApplication.kt
```diff
@@ -226,7 +226,6 @@ public open class TestApplicationBuilder {
      *
      * @see [testApplication]
      */
-    @KtorDsl
```
bjhham, line 229: The annotation must have worked this way at some point... There's no way we'd have repeated the same mistake this many times, right? 😅

osipxd: I don't know, but with [this reproducer](https://pl.kotl.in/8lj2HX6me) I failed to find any Kotlin version where compilation would fail for the annotated function (`scopedFunctinon2`)

PR #5062, opened 2025-08-29
KTOR-8165 Add support to obtain response body buffer in SSE

marychatte:
**Subsystem**
Client

**Motivation**
[KTOR-8165](https://youtrack.jetbrains.com/issue/KTOR-8165)

**Solution**
SSE responses are streaming, so saving the entire stream is impractical (unbounded, memory-heavy). That’s why reading `response.bodyAsText()` after an error used to fail (double-receive / already consumed).

The idea is to introduce a configurable diagnostic buffer that captures only what has already been processed by the SSE reader:
```
install(SSE) { 
   bufferPolicy = SSEBufferPolicy.LastEvents(10) 
}

// or per call:
client.sse(url, { bufferPolicy(SSEBufferPolicy.All) }) { … }
```

Snapshot is accumulated inside the session as the stream is read (no second read from the network).
On errors, we build a replayable response from this buffer (`SavedHttpCall`/`SavedHttpResponse`), so handlers can safely call `e.response?.bodyAsText()` for logging.

Options:
* Off - default
* LastLines(n)
* LastEvent / LastEvents(n): last 1 / last n completed SSE events.
* All: everything processed so far

"Handshake" failures (non-2xx or non-text/event-stream) contains full response body


ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/sse/BodySnapshotPolicy.kt
```diff
@@ -0,0 +1,103 @@
...
+import io.ktor.sse.*
+import io.ktor.utils.io.core.*
+
+/**
+ * Policy that controls how an SSE diagnostic body snapshot is captured while reading a stream.
+ *
+ * The snapshot represents already processed data (no re-reading from the network).
+ * It is intended for logging and error analysis when failures happen.
+ */
+public interface BodySnapshotPolicy {
```
bjhham: I think the word "snapshot" usually refers to more interactive recall mechanisms, whereas this would be more of a "buffer" in my opinion.  Perhaps it should be `SSEBufferPolicy`?

Also, I agree with the rabbit, this looks like it should probably be sealed.

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/sse/BodySnapshotPolicy.kt
```diff
@@ -0,0 +1,103 @@
...
+    public data object All : BodySnapshotPolicy
+}
+
+internal interface BodySnapshot {
+    fun appendLine(line: String)
+
+    fun appendEvent(event: ServerSentEvent)
+
+    fun getSnapshot(): ByteArray
+}
```
bjhham: This can also be sealed, and we can use a null object pattern to forgo the optional checks in the session.

```
internal sealed interface BodyBuffer {
    fun appendLine(line: String) {}
    fun appendEvent(event: ServerSentEvent) {}
    fun getSnapshot(): ByteArray
    
    internal class Events: BodyBuffer { /* ... */ }
    internal class Lines: BodyBuffer { /* ... */ }
    internal object Empty: BodyBuffer { /* ... */ }
}
```

bjhham (approved):
Some minor comments, but otherwise looks nice 👍

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/sse/BodySnapshotPolicy.kt
marychatte: Yes, thanks. I wasn't sure about "snapshot" naming, but "SSEBufferPolicy" sounds good. I removed all mentions of "snapshot" so now it looks like:
```kotlin
install(SSE) {
   sseBufferPolicy = SSEBufferPolicy.All
}
```

bjhham: Oh, one thing about the change, I'd go with:

```
install(SSE) {
   bufferPolicy = SSEBufferPolicy.All
}
```

Since the "sse" in the field name would be redundant in this context.

PR #5070, opened 2025-09-08
KTOR-8821 Fix auto-reload issue / exception on lambda

bjhham:
**Subsystem**
Server, Core

**Motivation**
[KTOR-8821](https://youtrack.jetbrains.com/issue/KTOR-8821) Autoreloading: it doesn't reload rebuilt classes since 3.2.0
[KTOR-8810](https://youtrack.jetbrains.com/issue/KTOR-8810) Autoreloading: JobCancellationException when app is reloaded in the debugger since 3.2.0

**Solution**
Fix 1: dynamic module wrapping for lambdas would cause errors on reloads
Fix 2: Netty was using the old application instance in the coroutine context, which caused auto-reloading to fail when debugging.

This also fixes the issue with the static resources.

### DIFF
--- a/ktor-server/ktor-server-core/jvm/src/io/ktor/server/engine/EmbeddedServerJvm.kt
+++ b/ktor-server/ktor-server-core/jvm/src/io/ktor/server/engine/EmbeddedServerJvm.kt
@@ -61,10 +61,9 @@ actual constructor(
     private val moduleInjector: ModuleParametersInjector by lazy {
         loadServiceOrNull() ?: ModuleParametersInjector.Disabled
     }
-    private val modules by lazy {
+    private val modules: List<DynamicApplicationModule> get() =
         environment.moduleConfigReferences.map(::dynamicModule) +
             rootConfig.modules.map { module -> module.toDynamicModuleOrNull() ?: module.wrapWithDynamicModule() }
-    }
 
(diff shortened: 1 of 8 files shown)


ktor-server/ktor-server-core/jvm/src/io/ktor/server/engine/EmbeddedServerJvm.kt
```diff
@@ -61,10 +61,9 @@ actual constructor(
     private val moduleInjector: ModuleParametersInjector by lazy {
         loadServiceOrNull() ?: ModuleParametersInjector.Disabled
     }
-    private val modules by lazy {
+    private val modules: List<DynamicApplicationModule> get() =
```
bjhham, line 64: I made this uncached, since it's only called once and this will allow for adding / removing modules at runtime.

bjhham:
As it turns out, the actual breaking change was that introducing suspend modules caused the blocking function references to get wrapped during casting so that they become anonymous inner classes internally.  Previously, the function reference `Application:myModule` would retain the function name and work as an argument in the `embeddedServer` function.

Maybe in 3.4, we can investigate fixing this as part of the "enhanced dev mode" story, since compose hot reload works with lambdas, the solution should work here.

PR #5071, opened 2025-09-09
KTOR-8820 Performance regression when using ContentEncoding and HttpRequestRetry since 3.2.0

osipxd:
**Subsystem**
Server

**Motivation**
[KTOR-8820](https://youtrack.jetbrains.com/issue/KTOR-8820) Performance regression when using ContentEncoding and HttpRequestRetry since 3.2.0

The bug was caused by a combination of `ContentEncoding`, `HttpRequestRetry` and `SaveBody` plugins.

1. `SaveBody`: creates a new channel on each `HttpResponse.rawContent` call
2. `ContentEnconding`: starts a coroutine for each created channel, and this coroutine exists until the entire body has been read
3. `HttpRequestRetry`: calls `rawContent` and then reads only one byte, leaving the channel not closed

As a result, the coroutine started by the `ContentEncoding` plugin hangs until the request fails by timeout (the default timeout is 15 seconds).

**Solution**

Added cancellation of the `rawContent` channel in the `HttpRetryRequest` plugin.
Also found that we were calling `rawContent` in `HttpResponse.cleanup` even for saved responses. Fixed this too

### DIFF
--- a/ktor-client/ktor-client-cio/common/src/io/ktor/client/engine/cio/Endpoint.kt
+++ b/ktor-client/ktor-client-cio/common/src/io/ktor/client/engine/cio/Endpoint.kt
@@ -40,7 +40,7 @@ internal class Endpoint(
     private val deliveryPoint: Channel<RequestTask> = Channel()
     private val maxEndpointIdleTime: Long = 2 * config.endpoint.connectTimeout
 
-    private val timeout = launch(coroutineContext + CoroutineName("Endpoint timeout($host:$port)")) {
+    private val timeout = launch(coroutineContext + CoroutineName("cio-endpoint-timeout")) {
         try {
             while (true) {
                 val remaining = (lastActivity.value + maxEndpointIdleTime) - getTimeMillis()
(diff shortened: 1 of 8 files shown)


ktor-client/ktor-client-cio/common/src/io/ktor/client/engine/cio/Endpoint.kt
```diff
@@ -40,7 +40,7 @@ internal class Endpoint(
     private val deliveryPoint: Channel<RequestTask> = Channel()
     private val maxEndpointIdleTime: Long = 2 * config.endpoint.connectTimeout
 
-    private val timeout = launch(coroutineContext + CoroutineName("Endpoint timeout($host:$port)")) {
+    private val timeout = launch(coroutineContext + CoroutineName("cio-endpoint-timeout")) {
```
osipxd, line 43: Removed metadata for consistency, but maybe I should add it back here 🤔

ktor-utils/jvm/src/io/ktor/util/EncodersJvm.kt
```diff
@@ -72,7 +77,7 @@ private fun inflate(
     source: ByteReadChannel,
     gzip: Boolean = true,
     coroutineContext: CoroutineContext
-): ByteReadChannel = GlobalScope.writer(coroutineContext) {
+): ByteReadChannel = GlobalScope.writer(coroutineContext + CoroutineName("encoder-inflate-writer")) {
```
osipxd: Should I create a global field with a pre-created `CoroutineName` to create it only once?

bjhham (approved):

PR #5072, opened 2025-09-09
KTOR-8753 Change OkHttp specific sse session to default sse session

marychatte:
**Subsystem**
Client, SSE

**Motivation**
Initially, for the `OkHttp` client, we supported their own implementation of SSE, but later, when we started to add new functionality to SSE, we encountered the limits of the `OkHttpSSESession`.
This PR changes the `OkHttpSSESession` to `DefaultClientSSESession`, so now we will have one common implementation for SSE

[KTOR-8753](https://youtrack.jetbrains.com/issue/KTOR-8753) - because it was one more SSE ticket related only to OkHttp


bjhham:
Seems the timeout test does not work on TC 👎

PR #5073, opened 2025-09-09
KTOR-8785 Fix for DI shutdown hook

bjhham:
**Subsystem**
Server, DI

**Motivation**
[KTOR-8785](https://youtrack.jetbrains.com/issue/KTOR-8785) DI JobCancellationException during cleanup

**Solution**
Turns out access to the deferred results, even without awaiting, requires the original coroutine scope.  Also, the parent scope (applcation) is destroyed before `ApplicationStopped`, so it needs to trigger on `ApplicationStopping`.


marychatte (approved):
👍

PR #5074, opened 2025-09-10
KTOR-8044 Get flaky tests under control

bjhham:
**Subsystem**
Tests

**Motivation**
There's been an increase in flaky tests on the native platforms recently.  We need them to play nice so we can update dependencies and not go insane.

**Solution**
Disabled some tests.


bjhham:
@coderabbit skip

PR #5084, opened 2025-09-16
Direct renovate to the appropriate branches

bjhham:
This should map patches to release and others to main.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -6,7 +6,7 @@
     ":label(dependencies)",
     "schedule:automergeNonOfficeHours"
   ],
-  "baseBranchPatterns": ["main","release/*"],
+  "baseBranchPatterns": ["main","/^release\\/.*/"],
   "rebaseWhen": "conflicted",
   "packageRules": [
     {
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -66,11 +67,15 @@
...
-        "minor"
-      ],
+      "description": "Patch updates go to release branch",
+      "matchUpdateTypes": ["patch"],
+      "matchBaseBranches": ["release"],
+      "automerge": true
+    },
+    {
+      "description": "Minor updates go to main branch",
+      "matchUpdateTypes": ["minor"],
```
osipxd: And major?

osipxd (commented):
How will we sync patch updates between these two branches? Merge `release` into `main` every week?

renovate.json
```diff
@@ -66,11 +67,15 @@
...
+      "matchBaseBranches": ["main"],
       "automerge": true
     },
```
osipxd, line 80: Probably it should be something like this:

```suggestion
    {
      "description": "Don't create PRs to main branch for patch updates",
      "matchUpdateTypes": ["patch"],
      "matchBaseBranches": ["main"],
      "enabled": false
},
```

renovate.json
```diff
@@ -66,11 +67,15 @@
...
+      "matchUpdateTypes": ["patch"],
+      "matchBaseBranches": ["main"],
+      "enabled": false
+    },
+    {
+      "description": "Minor/major updates go to main branch",
+      "matchUpdateTypes": ["minor", "major"],
+      "matchBaseBranches": ["main"],
       "automerge": true
     },
```
osipxd, line 80: I think this rule should also be converted to something with `enabled: false`. If I understand it correctly, `match*` properties just match dependencies and then we can change their options. This rule probably just enables automerge for minor and major versions targeted to `main`

bjhham:
> How will we sync patch updates between these two branches? Merge release into main every week?

Yes, every week or two.  We could automate creating a PR or even doing the merge.

I was also thinking we could include a tag for PRs into main that will automatically cherry-pick to release for any bug fixes that are targeted to main, wdyt?

PR #5086, opened 2025-09-16
KTOR-8592 Support serving static resources within bootJar 

marychatte:
**Subsystem**
Server, Static Content

**Motivation**
[KTOR-8592](https://youtrack.jetbrains.com/issue/KTOR-8592) Support serving static resources within bootJar (nested jars)
[KTOR-3880](https://youtrack.jetbrains.com/issue/KTOR-3880) `resources` routing is not working inside bootJar

**Solution**
`staticResources` crashes when the app runs from a Spring Boot bootJar because boot jars expose classpath resources via nested JAR URLs like: `jar:nested:/path/to/app.jar/!BOOT-INF/classes/!/static/…`

`resourceClasspathResource` tried to optimize any `jar` URL by turning it into a `JarFileContent` via `findContainingJarFile`, which assumes a single-level local `jar:file:/…!/entry`, so for nested/non-local JARs it threw:
```
IllegalArgumentException: Only local jars are supported (jar:file:)
```

The solution is only to optimize a single-level local `jar:file:/…!/entry`
For non-local jars, stream from the URL instead.
For nested jars, still throw the exception ([KTOR-8883](https://youtrack.jetbrains.com/issue/KTOR-8883))


ktor-server/ktor-server-core/jvm/src/io/ktor/server/http/content/StaticContentResolution.kt
```diff
@@ -107,15 +109,20 @@ public fun resourceClasspathResource(
     }
 }
 
-internal fun findContainingJarFile(url: String): File {
-    if (url.startsWith("jar:file:")) {
-        val jarPathSeparator = url.indexOf("!", startIndex = 9)
-        require(jarPathSeparator != -1) { "Jar path requires !/ separator but it is: $url" }
+// Only optimize single-level local 'jar:file:/...!/entry'
+internal fun findSimpleJarFile(url: String): File? {
+    if (!url.startsWith("jar:file:")) return null
```
bjhham: Could you extract `"jar:file:"` into a constant and reference it like `JAR_PREFIX.length` below as well?  The 9 is a bit of a magic number.

ktor-server/ktor-server-core/jvm/src/io/ktor/server/http/content/StaticContentResolution.kt
```diff
@@ -107,15 +109,20 @@ public fun resourceClasspathResource(
     }
 }
 
-internal fun findContainingJarFile(url: String): File {
-    if (url.startsWith("jar:file:")) {
-        val jarPathSeparator = url.indexOf("!", startIndex = 9)
-        require(jarPathSeparator != -1) { "Jar path requires !/ separator but it is: $url" }
+// Only optimize single-level local 'jar:file:/...!/entry'
```
bjhham: Could you elaborate a little more on the function comment for those unfamiliar with jar entry URL conventions?

bjhham (approved):
Couple minor comments; otherwise looks good.

PR #4971, opened 2025-07-01
Add support for suspendable OAuth providerLookup and urlProvider

chsavvaidis:
**Subsystem**
 Server, ktor-server-auth (inside ktor-server-plugins)

**Motivation**
This PR addresses the limitation in the current OAuthAuthenticationProvider which only supports non-suspending providerLookup and urlProvider configurations. This makes it difficult to support dynamic, multi-tenant OAuth setups where provider settings must be fetched asynchronously (e.g., from a database or remote service). This is especially critical for enterprise or SaaS use cases requiring per-request customization based http request parameters, etc. 

**Solution**
- Updated OAuthAuthenticationProvider to accept suspendable lambdas for both providerLookup and urlProvider.
- Inside the authentication flow, these lambdas are invoked within a coroutine context to support asynchronous fetching.
- Preserved backward compatibility with existing non-suspending configurations.
- Added tests.

PR #5087, opened 2025-09-17
Add support for suspendable OAuth providerLookup and urlProvider

chsavvaidis:
**Subsystem**
 Server, ktor-server-auth (inside ktor-server-plugins)

**Motivation**
Re-open of [#4971](https://github.com/ktorio/ktor/pull/4971).
This PR addresses the limitation in the current OAuthAuthenticationProvider which only supports non-suspending providerLookup and urlProvider configurations. This makes it difficult to support dynamic, multi-tenant OAuth setups where provider settings must be fetched asynchronously (e.g., from a database or remote service). This is especially critical for enterprise or SaaS use cases requiring per-request customization based http request parameters, etc.

**Solution**
- Updated OAuthAuthenticationProvider to accept suspendable lambdas for both providerLookup and urlProvider.
- Inside the authentication flow, these lambdas are invoked within a coroutine context to support asynchronous fetching.
- Preserved backward compatibility with existing non-suspending configurations.
- Added tests.


bjhham (approved):
lgtm 👍

PR #5089, opened 2025-09-17
KTOR-8771 Use `shutdownTimeout` instead of `shutdownGracePeriod`

marychatte:
**Subsystem**
Server

**Motivation**
[KTOR-8771](https://youtrack.jetbrains.com/issue/KTOR-8771) shutdownGracePeriod is used instead of shutdownTimeout in EmbeddedServer.stop()


bjhham (approved):
Oh, quite an oversight 🤦

PR #5094, opened 2025-09-18
KTOR-6761 Add `Apache5EngineConfig.configureConnectionManager` to simplify configuration

marychatte:
**Subsystem**
Client, Apache5

**Motivation**
[KTOR-6761](https://youtrack.jetbrains.com/issue/KTOR-6761) Apache5: Simplify configuration

**Solution**
Initially, the configuration looked like:
```kotlin
HttpClient(Apache) {
    engine {
        customizeClient {
            setMaxConnTotal(10_000)
            setMaxConnPerRoute(1_000)
        }
    }
}
```
But after upgrading to `Apache5`, the configuration started to look like:
```kotlin
HttpClient(Apache5) {
    engine {
        customizeClient {
            setConnectionManager(
                PoolingAsyncClientConnectionManagerBuilder.create()
                    .setMaxConnTotal(10_000)
                    .setMaxConnPerRoute(1_000)
                    .build()
            )
        }
    }
}
```

Because users need to create and set `ConnectionManager` now, it overrides the Ktor-managed connection manager [here](https://github.com/ktorio/ktor/blob/83ea18ec8439164333249ba6e03552ac168b7525/ktor-client/ktor-client-apache5/jvm/src/io/ktor/client/engine/apache5/Apache5Engine.kt#L98). For example, if the `HttpTimeout` plugin is used, its settings will be ignored.

The idea is to provide a `configureConnectionManager` function, so users won't override `ConnectionManager` but still can configure it.


osipxd:
> and you can point it to release/3.x if you want to include it in 3.3.1

Isn't it prohibited to add new APIs in patch versions? While this change is forward-compatible it breaks backward compatibility.

marychatte:
I have a similar question 😅 I want to include it in `3.4.0`

PR #5095, opened 2025-09-18
KTOR-8898: add netty config load: enableHttp2 and enableH2c

Komdosh:
**Subsystem**
Server, Netty Engine

**Motivation**
Not all netty config are loaded with `loadConfiguration` method.

**Solution**
Add a new config fields.


Komdosh:
@bjhham Any blockers for merging?

PR #5098, opened 2025-09-19
Resolve flaky test issues with JS

bjhham:
The task flag should fix [this problem](https://ktor.teamcity.com/buildConfiguration/Ktor_KtorMatrixJavaScriptChromeNodeJs/365329?buildTab=log&linesState=512&logView=flowAware&focusLine=46510).

And there seems to be some [flakiness with a couple of tests](https://ktor.teamcity.com/buildConfiguration/Ktor_KtorMatrixJavaScriptChromeNodeJs/365329?buildTab=log&linesState=512&logView=flowAware&focusLine=46510).

### DIFF
--- a/build-logic/src/main/kotlin/ktorbuild/targets/CiConfig.kt
+++ b/build-logic/src/main/kotlin/ktorbuild/targets/CiConfig.kt
@@ -24,6 +24,7 @@ internal fun Project.configureTestTasksOnCi() {
     // 2. To run as many tests as possible while keeping fail-fast behavior locally.
     tasks.withType<AbstractTestTask>().configureEach {
         ignoreFailures = true
+        failOnNoDiscoveredTests = false
         if (this is KotlinTest) ignoreRunFailures = true
     }
     // KotlinTestReport overwrites ignoreFailure values and fails build on test failure if this flag is disabled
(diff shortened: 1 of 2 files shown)


build-logic/src/main/kotlin/ktorbuild/targets/CiConfig.kt
```diff
@@ -24,6 +24,7 @@ internal fun Project.configureTestTasksOnCi() {
     // 2. To run as many tests as possible while keeping fail-fast behavior locally.
     tasks.withType<AbstractTestTask>().configureEach {
         ignoreFailures = true
+        failOnNoDiscoveredTests = false
```
osipxd: I don't think we should disabled this. With this flag Gradle fails if Chrome has failed to start tests for some module. Probably limiting parallelism in tests would solve this problem.

bjhham: I seeee.  I thought it was just a fluke in the cleanup because it happens near the end, but it looks like `:ktor-client-tests:jsBrowserTest` is crashing sometimes.

bjhham:
The JS tests still have the audacity to fail! 😠

bjhham:
Shutting down parallelism seems to do the trick 👍

PR #5099, opened 2025-09-19
 KTOR-7162 Add `defaultRequest { }` with optional `replace` parameter

marychatte:
**Subsystem**
Client, DefaultRequest

**Motivation**
[KTOR-7162](https://youtrack.jetbrains.com/issue/KTOR-7162) DefaultRequest: Configuration applied twice for client created with `HttpClient.config`

**Solution**
Install function combines configs: https://github.com/ktorio/ktor/blob/eda13fba68b1f4c513b4d3cb775615f2a1a9312f/ktor-client/ktor-client-core/common/src/io/ktor/client/HttpClientConfig.kt#L151

So the idea is to add a new function that doesn't combine configs and use it in `defaultRequest { ... }`

But it will ad some inconsistency so I'm not really sure:
```kotlin
val client2 = client.config {
    defaultRequest { // will override
        ...
    }
}

val client3 = client.config {
    install(DefaultRequest) {  // will NOT override
        ...
    }
}
```


marychatte:
@coderabbitai review

bjhham:
Aren't there some scenarios where you want `defaultRequest` to be additive?  Like if you want to keep the base URL but include some extra header later on?

marychatte:
@bjhham `install(DefaultRequest) { ... }` is still additive, only `defaultRequest { ... }` isn't. But I think it's too confusing to have them both with different behaviour, maybe add smth like this:
```kotlin
defaultRequest(override = true) { 
    ... 
}
```

bjhham:
Yes, I think we'll need the optional parameter there to avoid changing the default behaviour.

ktor-client/ktor-client-core/common/src/io/ktor/client/HttpClientConfig.kt
```diff
@@ -166,6 +166,26 @@ public class HttpClientConfig<T : HttpClientEngineConfig> {
...
         }
 
+        installPlugin(plugin)
+    }
+
+    /**
+     * Installs the specified [plugin] and optionally configures it using the [configure] block.
+     * If the plugin is already installed, the configuration block will override the existing configuration class.
+     */
+    internal fun <TBuilder : Any, TPlugin : Any> installAndOverride(
```
bjhham: Maybe `installOrReplace()`?  This could be public as well, since there's some other uses.

bjhham (approved):

PR #5100, opened 2025-09-19
KTOR-8901 Remove @KtorDsl from functions

osipxd:
**Subsystem**
Core

**Motivation**
[KTOR-8901](https://youtrack.jetbrains.com/issue/KTOR-8901) Remove `AnnotationTarget.FUNCTION` from `@KtorDsl`

**Solution**
This PR removes wrongly used `@KtorDsl` annotation to make it possible to remove the annotation target `FUNCTION` in the future.


bjhham (commented):
> I'd merge this into the next patch release, however this change might break code misusing the @KtorDsl annotation. The main point why I want it to be included is that this change prevents further misuse of the @KtorDsl annotation.

It sounds like you meant to select `main` for the base branch in this case...

osipxd:
We could merge all changes except for removing the target into the patch release 🤔 
And merge removing the target into the `main`.

bjhham (approved):

PR #5104, opened 2025-09-23
CORS Plugin: Access-Control-Allow-Methods header does not include default methods (GET, POST, HEAD)

### LINKED ISSUE #5103: CORS Plugin: Access-Control-Allow-Methods header does not include default methods (GET, POST, HEAD)
In Ktor 3.3.0, the CORS plugin does not correctly include the default HTTP methods (GET, POST, HEAD) in the Access-Control-Allow-Methods response header.

Current implementation in CORSConfig builds the methods header like this:

`val methodsListHeaderValue = methods.filterNot { it in CorsDefaultMethods }
    .map { it.value }
    .sorted()
    .joinToString(", ")
`

Since CorsDefaultMethods = { GET, POST, HEAD }, these methods are filtered out and not included in the final header.
As a result, browsers see an incomplete Access-Control-Allow-Methods header (often containing only OPTIONS), which breaks CORS preflight validation.

To Reproduce

Enable the CORS plugin in Ktor 3.3.0 and allow POST.

Send an OPTIONS preflight request from the browser.

Inspect the response headers.

Example output:

`Access-Control-Allow-Methods: OPTIONS
`

Expected output:

`Access-Control-Allow-Methods: GET, POST, HEAD, OPTIONS
`

Expected behavior
The Access-Control-Allow-Methods header should always include all configured methods, including the defaults (GET, POST, HEAD).

Proposed Fix
Replace the filtering logic with a distinct collection of methods:

`val methodsListHeaderValue = methods.distinct()
    .map { it.value }
    .sorted()
    .joinToString(", ")
`

This ensures that GET, POST, HEAD are included in the header along with any additional methods.

Environment

Ktor version: 3.3.0

Module: ktor-server-cors

JVM: 17

OS: (e.g. Ubuntu 22.04 / Windows 11)

Additional context
This bug prevents correct CORS preflight handling in browsers, leading to blocked requests even when methods are allowed.

morteza-69:
Fixes #5103

Problem
In Ktor 3.3.0, the CORS plugin does not correctly include the default HTTP methods (GET, POST, HEAD) in the Access-Control-Allow-Methods response header.

Current implementation in CORSConfig builds the header value like this:

`val methodsListHeaderValue = methods.filterNot { it in CorsDefaultMethods }
    .map { it.value }
    .sorted()
    .joinToString(", ")
`

Since CorsDefaultMethods = { GET, POST, HEAD }, these methods are filtered out and not included in the final header.
This results in browsers seeing an incomplete Access-Control-Allow-Methods (often only OPTIONS), which breaks CORS preflight validation.

Solution
Replaced filtering logic with a distinct collection:

`val methodsListHeaderValue = methods.distinct()
    .map { it.value }
    .sorted()
    .joinToString(", ")
`
This ensures that the default methods (GET, POST, HEAD) are included along with any additional configured methods.

Tests
Added a new test testPreflightIncludesDefaultMethods to verify that:

GET, POST, and HEAD are always present in Access-Control-Allow-Methods

Custom allowed methods (e.g., PUT) are also included

Before the fix, this test fails. After the fix, it passes ✅

Environment

Ktor version: 3.3.0

Module: ktor-server-cors

JVM: 17

OS: Ubuntu 22.04 / Windows 11

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-cors/common/src/io/ktor/server/plugins/cors/CORS.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-cors/common/src/io/ktor/server/plugins/cors/CORS.kt
@@ -54,7 +54,7 @@ internal fun PluginBuilder<CORSConfig>.buildPlugin() {
     val allowNonSimpleContentTypes: Boolean = pluginConfig.allowNonSimpleContentTypes
     val headersList = pluginConfig.headers.filterNot { it in CORSConfig.CorsSimpleRequestHeaders }
         .let { if (allowNonSimpleContentTypes) it + HttpHeaders.ContentType else it }
-    val methodsListHeaderValue = methods.filterNot { it in CORSConfig.CorsDefaultMethods }
+    val methodsListHeaderValue = methods.distinct()
         .map { it.value }
         .sorted()
         .joinToString(", ")
--- a/ktor-server/ktor-server-tests/common/test/io/ktor/tests/server/plugins/CORSTest.kt
+++ b/ktor-server/ktor-server-tests/common/test/io/ktor/tests/server/plugins/CORSTest.kt
@@ -1592,4 +1592,33 @@ class CORSTest {
             assertEquals(response.status, HttpStatusCode.Forbidden)
         }
     }
+
+    @Test
+    fun testPreflightIncludesDefaultMethods() = testApplication {
+        install(CORS) {
+            anyHost()
+            allowMethod(HttpMethod.Put)
+        }
+
+        routing {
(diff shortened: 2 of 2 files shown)


bjhham (commented):
Yep, this sounds like how it ought to work:

> By default, the CORS plugin allows the GET, POST and HEAD HTTP methods. To add additional methods, use the allowMethod function.

Thanks for the fix.

Could you please run `./gradlew formatKotlin` to get past the linter check?

bjhham:
Fix merged in another PR https://github.com/ktorio/ktor/pull/5633

PR #5110, opened 2025-10-01
KTOR-4828 Fix NumberFormatException when Content-Length value is invalid

marychatte:
**Subsystem**
Client, CIO

**Motivation**
[KTOR-4828](https://youtrack.jetbrains.com/issue/KTOR-4828) NumberFormatException when Content-Length header value contains null bytes

**Solution**
Replace `toLong()` with `toLongOrNull()` for `Content-Length`, so if parsing fails, treat it as unknown length instead of throwing. For `Connection: close`, we safely read until EOF. For  `Connection: keep-alive` with no valid `Content-Length` and no chunked, we fail. 
Almost all other engines throw an exception on invalid `Content-Length`.

### DIFF
--- a/ktor-http/common/src/io/ktor/http/HttpMessageProperties.kt
+++ b/ktor-http/common/src/io/ktor/http/HttpMessageProperties.kt
@@ -73,7 +73,7 @@ public fun HttpMessageBuilder.vary(): List<String>? = headers.getAll(HttpHeaders
  *
  * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.http.contentLength)
  */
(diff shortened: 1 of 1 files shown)


bjhham (approved):
I'm curious to know how the other engines handle this 🤔

marychatte:
@bjhham 
For both `Connection: keep-alive` and `Connection: close`:
- Throws an exception:
 `Darwin`, `Apache5`, `OkHttp`, `Js`
- Treats `Content-Length` as a valid number and processes body:
`Android`, `Java`, `Apache`
- No exception, status code 200, no headers, no body
`Curl`

PR #5111, opened 2025-10-01
Add support to configure token caching in BearerAuthProvider

ruben-caceiro-digiwest:
**Subsystem**
Client, related module

**Motivation**
[Allow non cache Bearer token](https://youtrack.jetbrains.com/issue/KTOR-8932/Allow-non-cache-Bearer-token)

**Solution**
I simply add a cache flag when the flag is false it ignores the cached token and get always a new one.


e5l:
Hey @ruben-caceiro-digiwest, sorry for the delay. The issue is fixed in `main`

PR #5112, opened 2025-10-02
KTOR-6198: Fix incorrect CloseReason in Client/WebSocket/Darwin

egeorgievminerva:
**Subsystem**
Client/WebSocket/Darwin

**Motivation**
The fix of KTOR-6198 is important so that the client application can implement custom behaviour based on the provided close code by the server.

**Solution**
Don't cancel the WebSocket job if receiveMessage() fails with exception and the WebSocket is closed. When the WebSocket is closed the callback method URLSession:webSocketTask:didCloseWithCode:reason: in the Ktor library takes care of canceling the WebSocket job after the close code is propagated to the client application. The exception in receiveMessage() was canceling the WebSocket job and because of that the close code and reason were not propagated.


bjhham (approved):
Thank you for the fix 🙏

PR #5113, opened 2025-10-03
KTOR-7824 Fix parsing of multiple headers

marychatte:
**Subsystem**
Core

**Motivation**
[KTOR-7824](https://youtrack.jetbrains.com/issue/KTOR-7824)
We want to parse something like this `X-Multi-Header: Value1, Value2` into `listOf("value1", "value2")`: 2 values not 1 as it's done now
The same behaviour can be found in [Spring](https://github.com/spring-projects/spring-framework/issues/18797)

**Solution**
For headers that are semantically lists, values are split on commas outside quotes and flattened across multiple field-lines. Known single-value/special headers are left untouched.


ktor-http/common/src/io/ktor/http/header/utils.kt
```diff
@@ -0,0 +1,106 @@
...
+
+/**
+ * Returns `true` when [name] should be treated as a comma-separated list header.
+ */
+internal fun isCommaSeparatedHeader(name: String): Boolean = name.lowercase() !in NON_LIST_HEADERS
+
+/**
+ * Splits header's values on commas when the header is treated as a comma-separated list.
+ */
+internal fun splitValuesForHeader(values: Map<String, List<String>>) = values.entries.associate { (k, v) ->
```
e5l: Please also consider specifying the type explicitly. It would also be nice to have separate method names. `splitValueorHeader` has 2 overloads with different complexity

marychatte:
@e5l 
I tried to fix all the fallen tests after the initial solution and understood that maybe it'sn't a good solution. A few points:
1. It will be a breaking change and lead to mistakes for some users' custom headers, such as `Custom-Date: Mon, 06 Oct 2025 18:17:02 GMT`  or all other custom headers where commas appear inside values.
2. It could be a breaking change for many existing apps/plugins. For example, `Cache-Control: max-age=180, public`, there are many places where logic relies on raw per-line values.
3. Spring provides a separate function for such cases, they have [`HttpHeaders.getValuesAsList`](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/http/HttpHeaders.html#getValuesAsList(java.lang.String)) which returns separated values and `HttpHeaders.get` which has the same behaviour as the current `Headers.getAll`

So, I think it's better to keep `getAll` as it is, as the lossless view, and add a separate function for convenience list behavior.

```kotlin
/**
 * Gets all values associated with the specified header [name] and splits them using the provided [separator].
 *
 * If [splitInsideQuotes] is true, the function ignores quotes entirely and splits everywhere the [separator] appears.
 * Otherwise, [separator] that occurs inside a double-quoted string isn't treated as a split point.
 *
 *  @param name The header name.
 *  @param separator The character on which to split (e.g., ',' or ';').
 *  @param splitInsideQuotes If `true`, quotes are ignored and splitting occurs at every [separator].
 *                           If `false`, separators inside quoted strings are not considered split points.
 */
public fun Headers.getSplitValues(
    name: String,
    separator: Char = ',',
    splitInsideQuotes: Boolean = false
): List<String>?
```

e5l:
> @e5l I tried to fix all the fallen tests after the initial solution and understood that maybe it'sn't a good solution. A few points:
> 
> 1. It will be a breaking change and lead to mistakes for some users' custom headers, such as `Custom-Date: Mon, 06 Oct 2025 18:17:02 GMT`  or all other custom headers where commas appear inside values.
> 2. It could be a breaking change for many existing apps/plugins. For example, `Cache-Control: max-age=180, public`, there are many places where logic relies on raw per-line values.
> 3. Spring provides a separate function for such cases, they have [`HttpHeaders.getValuesAsList`](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/http/HttpHeaders.html#getValuesAsList(java.lang.String)) which returns separated values and `HttpHeaders.get` which has the same behaviour as the current `Headers.getAll`
> 
> So, I think it's better to keep `getAll` as it is, as the lossless view, and add a separate function for convenience list behavior.
> 
> ```kotlin
> /**
>  * Gets all values associated with the specified header [name] and splits them using the provided [separator].
>  *
>  * If [splitInsideQuotes] is true, the function ignores quotes entirely and splits everywhere the [separator] appears.
>  * Otherwise, [separator] that occurs inside a double-quoted string isn't treated as a split point.
>  *
>  *  @param name The header name.
>  *  @param separator The character on which to split (e.g., ',' or ';').
>  *  @param splitInsideQuotes If `true`, quotes are ignored and splitting occurs at every [separator].
>  *                           If `false`, separators inside quoted strings are not considered split points.
>  */
> public fun Headers.getSplitValues(
>     name: String,
>     separator: Char = ',',
>     splitInsideQuotes: Boolean = false
> ): List<String>?
> ```

Agree, let's introduce new method instead of breaking change

e5l (approved):
lgtm, please check ABI and tests

PR #5116, opened 2025-10-06
Zibet27/ktor client webrtc fix close

zibet27:
**Subsystem**
WebRTC Client

**Motivation**
All peer connection implementations didn't close the inner coroutine context on close.

**Solution**
- Call `super.close()` in classes that extend `WebRtcPeerConnection`
- update `suspendCoroutine` -> `suspendCancellableCoroutine`
- extra: remove redundant `as? T ?: error(...)` in the some code parts


bjhham (approved):
Nice fix, lgtm 👍

PR #5118, opened 2025-10-08
Contributing guide. Mention `CocoaPods` installation.

zibet27:
**Subsystem**
Documentation

**Motivation**
Building on macOS requires CocoaPods.

**Solution**
Mention it in `Contributing Guide`.

### DIFF
--- a/ktor-http/api/ktor-http.api
+++ b/ktor-http/api/ktor-http.api
@@ -465,6 +465,8 @@ public final class io/ktor/http/HeadersImpl : io/ktor/util/StringValuesImpl, io/
 }
 
 public final class io/ktor/http/HeadersKt {
+	public static final fun getSplitValues (Lio/ktor/http/Headers;Ljava/lang/String;CZ)Ljava/util/List;
+	public static synthetic fun getSplitValues$default (Lio/ktor/http/Headers;Ljava/lang/String;CZILjava/lang/Object;)Ljava/util/List;
 	public static final fun headers (Lkotlin/jvm/functions/Function1;)Lio/ktor/http/Headers;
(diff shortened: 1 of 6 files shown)


CONTRIBUTING.md
```diff
@@ -46,7 +46,9 @@ The path to the android sdk should be defined in the ANDROID_HOME environment va
 sdk.dir=/Users/USER_NAME/Library/Android/sdk
 ```
 
-If targeting macOS and/or iOS, install `Xcode` and `Xcode command line tools` on macOS.
+If you target macOS and/or iOS, install `Xcode`, `Xcode command line tools`, `CocoaPods` and `Ruby` on macOS.
+An outdated Ruby version 2 is installed by default, but you should manually install a new version of it, see [Kotlin Multiplatform Docs](https://www.jetbrains.com/help/kotlin-multiplatform-dev/multiplatform-cocoapods-overview.html).
```
vnikolova: I think we can include the steps to install CocoaPods and Ruby with `rbenv`, instead of linking to the KMP docs. It's the only relevant section from the topic, right?

zibet27: @vnikolova do you mean the section of KMP docs?

vnikolova: Yes

zibet27: Yeah, this is the only section explaining how to install Cocoapods for MCP needs. 
I'm not sure if we need to copy the steps from their documentation.

vnikolova: I think it is better than redirecting because it's only 4 steps and we have similar instructions in the next section to install other libraries in "Requirements for Ktor before 3.1.0". It is ok to copy the steps exactly as they are if you want and just change the Ruby version.

PR #5120, opened 2025-10-08
Add changelog for 3.3.1

Stexxe:


CHANGELOG.md
```diff
@@ -1,3 +1,19 @@
+# 3.3.1
+> Published 8 October 2025
+
+### Improvements
+* SSE: ClientSSESession doesn't behave like a CoroutineScope ([KTOR-8440](https://youtrack.jetbrains.com/issue/KTOR-8440))
```
marychatte: It's not an improvement since we only updated KDocs, so they say the current behavior is a designed one
PR called: KTOR-8440 Add a note about SSE session lifetime in KDoc 
```suggestion
* Add a note about SSE session lifetime in KDoc ([KTOR-8440](https://youtrack.jetbrains.com/issue/KTOR-8440))
```

CHANGELOG.md
```diff
@@ -1,3 +1,19 @@
...
+* Update Kotlin to 2.2.20 ([KTOR-8896](https://youtrack.jetbrains.com/issue/KTOR-8896))
+* Support serving static resources within bootJar (nested jars) ([KTOR-8592](https://youtrack.jetbrains.com/issue/KTOR-8592))
```
marychatte: It's a bugfix, and maybe better drop `(nested jars)` since it became another new task

e5l:
We usually do

PR #5123, opened 2025-10-14
KTOR-8970 Lowercase `varyKeys` keys when deserializing from file storage

sheepmaster:
ktorio/ktor#5030 updated the `varyKeys` map to use lowercased keys in order to allow case-insensitive matching. The corresponding logic in `FileCacheStorage` has not been updated however, which can cause an exception when upgrading from an older version to 3.3.0. This is the sequence of events:

* An old entry exists in the cache on disk, with a `varyKeys` key that contains uppercase letters
* Client updates to 3.3.0 or later
* Updated client prepares for a request, and calls [`findResponse()`](https://github.com/ktorio/ktor/blob/bd41f807a02967990cc45ebe9915af6b6af36f08/ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/cache/HttpCache.kt#L366)
  * `findResponse()` uses header lookup (which is case-insensitive) to verify that the `Vary` keys match
  * Cache entry matches and the request proceeds, with revalidation
* Client sends the request to revalidate, server responds with `Not Modified`
* Client calls `findAndRefresh()` with the response, which calls [the other variant of `findResponse()`](https://github.com/ktorio/ktor/blob/bd41f807a02967990cc45ebe9915af6b6af36f08/ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/cache/HttpCache.kt#L366)
  * *That* variant calls `CacheStorage.find()`, which looks up keys directly in the `varyKeys` map (case-*sensitive*)
  * This lookup fails because the map keys are using uppercase characters, whereas the requested keys are lowercased
* This causes an `InvalidCacheStateException` to be thrown because it looks like the cache entry has disappeared.

This PR updates the code in `FileCacheStorage` that deserializes cache entries from disk to lowercase the keys in the `varyKeys` map, as that's the only other "real" source of cache entries (in-memory caches need to be populated from somewhere else).

### DIFF
--- a/ktor-client/ktor-client-core/jvm/src/io/ktor/client/plugins/cache/storage/FileCacheStorage.kt
+++ b/ktor-client/ktor-client-core/jvm/src/io/ktor/client/plugins/cache/storage/FileCacheStorage.kt
@@ -206,13 +206,13 @@ private class FileCacheStorage(
     }
 
     /**
-     * Deserialize a single CachedResponseData from the provided ByteReadChannel.
+     * Deserialize a single [CachedResponseData] from the provided [ByteReadChannel].
      *
      * Reads the cached-entry fields in the stored binary format: request URL, HTTP status and version,
      * headers, request/response/expiration timestamps, vary keys (converted to lowercase), and the body bytes.
      *
      * @param channel Source channel positioned at the start of a serialized cache entry.
-     * @return The reconstructed CachedResponseData instance.
+     * @return The reconstructed [CachedResponseData] instance.
      */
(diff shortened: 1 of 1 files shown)


sheepmaster:
@coderabbitai generate docstrings

marychatte:
@sheepmaster I noticed an additional [PR](https://github.com/ktorio/ktor/pull/5124) for the KDoc, you can include it in this PR

sheepmaster:
> @sheepmaster Nice catch, thanks for the fix! Can you please change the base branch from `ktorio:main` to `ktorio: release:3.x`?

Done!

PR #5125, opened 2025-10-15
KTOR-8936 Routing documentation API

bjhham:
**Subsystem**
Server, OpenAPI tooling

**Motivation**
[KTOR-8936](https://youtrack.jetbrains.com/issue/KTOR-8936) Routing documentation runtime API

This introduces the runtime API for annotating endpoints.  It does not include any means in particular for actually _using_ the new metadata attached to the routes, but this will come in a subsequent PR.

**Solution**
This adds the relevant parts of the OpenAPI spec as serializable Kotlin types, and a DSL for adding information to any route so that it may be read later for producing documentation.  You can find an example of what this looks like in the `RouteAnnoationApiTest`.  There's not much there currently for testing, but that will come in with later changes to complete the OpenAPI types and integrate it with the current OpenAPI / Swagger plugins.

So for context, the implementation plan is going something like this:
* [x] Introduce operation and path items for runtime documentation API (here)
* [ ] Add the complete model to our runtime API and integrate with the current plugins.
* [ ] Update compiler plugin to generate the annotation API code instead of generating the full spec

There's a fair bit of boilerplate here with the data classes that you can mostly ignore - credit to https://github.com/nomisRev/OpenAPI-kt

One key difference is that I introduced a `GenericElement` with possibility of extension so we're not bound to a particular YAML implementation.  You'll notice this is still using `kotlinx-serialization`, but it should be fine for our types, so long as we provide support for the JSON schema publishing for Jackson and Gson, which will require a couple custom functions to traverse the types with reflection - I'll create some new tickets for these.

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-routing-annotate/common/src/io/ktor/annotate/RouteAnnotationApi.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-routing-annotate/common/src/io/ktor/annotate/RouteAnnotationApi.kt
@@ -84,7 +84,7 @@ private fun RoutingNode.operationFromSelector(): Operation? {
         is OptionalParameterRouteSelector -> Operation.build {
             parameters {
                 query(paramSelector.name) {
-                    required = paramSelector is OptionalParameterRouteSelector
+                    required = paramSelector !is OptionalParameterRouteSelector
                 }
             }
         }
@@ -92,7 +92,7 @@ private fun RoutingNode.operationFromSelector(): Operation? {
         is PathSegmentOptionalParameterRouteSelector -> Operation.build {
             parameters {
                 path(paramSelector.name) {
-                    required = paramSelector is PathSegmentOptionalParameterRouteSelector
+                    required = paramSelector !is PathSegmentOptionalParameterRouteSelector
                 }
             }
         }
--- a/gradle/artifacts/publishJvmAndCommonPublications.txt
+++ b/gradle/artifacts/publishJvmAndCommonPublications.txt
@@ -261,6 +261,13 @@ io.ktor:ktor-network/.jar
 io.ktor:ktor-network/javadoc.jar
 io.ktor:ktor-network/kotlin-tooling-metadata.json
 io.ktor:ktor-network/sources.jar
+io.ktor:ktor-openapi-schema-jvm/.jar
+io.ktor:ktor-openapi-schema-jvm/javadoc.jar
+io.ktor:ktor-openapi-schema-jvm/sources.jar
+io.ktor:ktor-openapi-schema/.jar
+io.ktor:ktor-openapi-schema/javadoc.jar
+io.ktor:ktor-openapi-schema/kotlin-tooling-metadata.json
+io.ktor:ktor-openapi-schema/sources.jar
 io.ktor:ktor-resources-jvm/.jar
 io.ktor:ktor-resources-jvm/javadoc.jar
 io.ktor:ktor-resources-jvm/sources.jar
@@ -621,6 +628,13 @@ io.ktor:ktor-server-resources/.jar
 io.ktor:ktor-server-resources/javadoc.jar
 io.ktor:ktor-server-resources/kotlin-tooling-metadata.json
 io.ktor:ktor-server-resources/sources.jar
+io.ktor:ktor-server-routing-annotate-jvm/.jar
+io.ktor:ktor-server-routing-annotate-jvm/javadoc.jar
+io.ktor:ktor-server-routing-annotate-jvm/sources.jar
+io.ktor:ktor-server-routing-annotate/.jar
+io.ktor:ktor-server-routing-annotate/javadoc.jar
+io.ktor:ktor-server-routing-annotate/kotlin-tooling-metadata.json
+io.ktor:ktor-server-routing-annotate/sources.jar
 io.ktor:ktor-server-servlet-jakarta-jvm/.jar
 io.ktor:ktor-server-servlet-jakarta-jvm/javadoc.jar
 io.ktor:ktor-server-servlet-jakarta-jvm/sources.jar
--- a/ktor-server/ktor-server-test-host/common/test/TestApplicationCoroutinesTest.kt
+++ b/ktor-server/ktor-server-test-host/common/test/TestApplicationCoroutinesTest.kt
@@ -33,6 +33,6 @@ class TestApplicationCoroutinesTest {
         val mark = kotlin.time.TimeSource.Monotonic.markNow()
         client.get("/")
         val elapsed = mark.elapsedNow().inWholeMilliseconds
-        assertTrue(elapsed < 1_000, "Expected request to finish quickly, but it took $elapsed ms")
+        assertTrue(elapsed < 5_000, "Expected request to finish quickly, but it took $elapsed ms")
     }
 }
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -128,16 +128,16 @@ projects {
         +"ktor-client-test-base"
         +"ktor-client-tests"
 
-//        +"ktor-client-webrtc" including {
-//            // Include `ktor-client-webrtc-rs` if rust compilation is enabled in `gradle.properties` for local builds
-//            // or if the `RUST_COMPILATION` environment variable is set to `true` for the CI.
-//            val compileRust = providers.environmentVariable("KTOR_RUST_COMPILATION")
-//                .orElse(providers.gradleProperty("ktorbuild.rustCompilation"))
-//                .orNull.toBoolean()
-//            if (compileRust) {
-//                +"ktor-client-webrtc-rs" // requires `cargo` to be installed
-//            }
-//        }
+        +"ktor-client-webrtc" including {
+            // Include `ktor-client-webrtc-rs` if rust compilation is enabled in `gradle.properties` for local builds
+            // or if the `RUST_COMPILATION` environment variable is set to `true` for the CI.
+            val compileRust = providers.environmentVariable("KTOR_RUST_COMPILATION")
+                .orElse(providers.gradleProperty("ktorbuild.rustCompilation"))
+                .orNull.toBoolean()
+            if (compileRust) {
+                +"ktor-client-webrtc-rs" // requires `cargo` to be installed
+            }
+        }
 
         nested("ktor-client-plugins") {
             +"ktor-client-auth"


ktor-shared/ktor-openapi-schema/common/src/io/ktor/openapi/GenericElement.kt
```diff
@@ -0,0 +1,486 @@
...
+public class GenericElementWrapper<T : Any>(
+    override val element: T,
+    override val elementSerializer: KSerializer<T>,
+) : GenericElement {
+
+    @Suppress("UNCHECKED_CAST")
+    override fun <T> deserialize(serializer: KSerializer<T>): T =
+        element as? T ?: error {
+            "Cannot deserialize ${element::class} to ${serializer.descriptor.serialName}"
+        }
```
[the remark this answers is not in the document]
bjhham: This is incorrect.

ktor-shared/ktor-openapi-schema/common/src/io/ktor/openapi/GenericElement.kt
```diff
@@ -0,0 +1,486 @@
...
+ */
+public class GenericElementString(
+    override val element: String,
+    override val elementSerializer: KSerializer<String> = String.serializer(),
+) : GenericElement {
+    @Suppress("UNCHECKED_CAST")
+    override fun <T> deserialize(serializer: KSerializer<T>): T =
+        element as? T ?: error {
+            "Expected String but got ${serializer.descriptor.serialName}"
+        }
```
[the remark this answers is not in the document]
bjhham: Also wrong.

ktor-shared/ktor-openapi-schema/common/src/io/ktor/openapi/PathItem.kt
```diff
@@ -0,0 +1,65 @@
...
+     * These parameters can be overridden at the operation level, but cannot be removed there. The
+     * list MUST NOT include duplicated parameters. A unique parameter is defined by a combination of
+     * a name and location. The list can use the Reference Object to link to parameters that are
+     * defined at the OpenAPI Object's components/parameters.
+     */
+    public val parameters: List<ReferenceOr<Parameter>> = emptyList(),
+    /**
+     * Any additional external documentation for this OpenAPI document.
+     */
+    public val extensions: Map<String, GenericElement> = emptyMap(),
```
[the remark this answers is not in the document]
bjhham: This will be handled using the `ReferenceOr` wrapper type like elsewhere.

ktor-shared/ktor-openapi-schema/common/src/io/ktor/openapi/GenericElement.kt
```diff
@@ -0,0 +1,486 @@
...
+            if (descriptor.isElementOptional(newIndex)) {
+                continue
+            }
+
+            // Required property not found
+            return CompositeDecoder.DECODE_DONE
+        }
+
+        return CompositeDecoder.DECODE_DONE
+    }
```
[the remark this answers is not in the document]
bjhham: I think this is actually expected behaviour for when there's a missing required property.

ktor-shared/ktor-openapi-schema/common/src/io/ktor/openapi/Operation.kt
```diff
@@ -0,0 +1,758 @@
...
+@KeepGeneratedSerializer
+public data class Operation(
+    public val tags: List<String>? = null,
+    public val summary: String? = null,
+    public val description: String? = null,
+    public val externalDocs: String? = null,
+    public val operationId: String? = null,
+    public val parameters: List<Parameter>? = null,
+    public val requestBody: RequestBody? = null,
+    public val responses: Map<String, Response>? = null,
```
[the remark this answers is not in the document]
bjhham: I'll leave this as is because these objects can be combined from multiple sources.

ktor-shared/ktor-openapi-schema/common/src/io/ktor/openapi/Operation.kt
```diff
@@ -0,0 +1,758 @@
...
+@Serializable(Response.Companion.Serializer::class)
+@OptIn(ExperimentalSerializationApi::class)
+@KeepGeneratedSerializer
+public data class Response(
+    public val description: String,
+    public val headers: Map<String, Parameter>? = null,
+    @Serializable(ContentTypeSerializer::class)
+    public var contentType: ContentType? = null,
+    public var schema: Schema? = null,
+    public val extensions: Map<String, GenericElement>? = null
```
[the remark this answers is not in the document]
bjhham: No, I'd rather leave some things as mutable in these builder types.

ktor-shared/ktor-openapi-schema/common/src/io/ktor/openapi/ReferenceOr.kt
```diff
@@ -0,0 +1,74 @@
...
+            override fun serialize(encoder: Encoder, value: ReferenceOr<T>) {
+                when (value) {
+                    is Value -> encoder.encodeSerializableValue(dataSerializer, value.value)
+                    is Reference -> encoder.encodeSerializableValue(Reference.serializer(), value)
+                }
+            }
+
+            override fun deserialize(decoder: Decoder): ReferenceOr<T> {
+                val element: GenericElement = decoder.decodeSerializableValue(decoder.serializersModule.serializer())
+                return when {
```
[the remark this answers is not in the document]
bjhham: It is not ambiguous, the function infers the correct type and this ensures that we use the serializer from the module instead of the default.

ktor-server/ktor-server-core/common/src/io/ktor/server/routing/RouteSelector.kt
```diff
@@ -464,6 +496,8 @@ public data class PathSegmentOptionalParameterRouteSelector(
         )
     }
 
+    override fun pathString(): String = "${prefix ?: ""}{$name}${suffix ?: ""}"
+
```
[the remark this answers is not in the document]
bjhham: These path strings are not meant to be parsed by Ktor routing, they are used for descriptions like OpenAPI that do not support this syntax.

ktor-shared/ktor-openapi-schema/common/src/io/ktor/openapi/Operation.kt
```diff
@@ -0,0 +1,758 @@
...
+@Serializable(Parameter.Companion.Serializer::class)
+@OptIn(ExperimentalSerializationApi::class)
+@KeepGeneratedSerializer
+public data class Parameter(
+    public val name: String,
+    public val `in`: String,
+    public val description: String? = null,
+    public val required: Boolean = false,
+    public val deprecated: Boolean = false,
+    public val schema: Schema? = null,
```
bjhham: Looks like schema also needs to be hoisted up similar to extensions.

bjhham: Turns out I was looking at an older spec 🤦

ktor-server/ktor-server-plugins/ktor-server-routing-annotate/common/src/io/ktor/annotate/RouteAnnotationApi.kt
```diff
@@ -0,0 +1,191 @@
...
+        .mapNotNull(RoutingNode::asPathItem)
+        .fold(mutableMapOf()) { map, (route, pathItem) ->
+            map.also {
+                if (route in map) {
+                    map[route] = map[route]!! + pathItem
+                } else {
+                    map[route] = pathItem
+                }
+            }
+        }
```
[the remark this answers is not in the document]
bjhham: This is what I expected for the behaviour.  I'll update the KDoc to reflect this.

ktor-shared/ktor-openapi-schema/common/src/io/ktor/openapi/DelegateMixinSerializer.kt
```diff
@@ -0,0 +1,154 @@
...
+    baseSerializer,
+    mapOf(
+        property0.name to PropertyDelegateDescriptor(property0Serializer, property0KeyMatcher),
+        property1.name to PropertyDelegateDescriptor(property1Serializer, property1KeyMatcher),
+    ),
+    { base, extras ->
+        val (property0Value, property1Value) = extras
+        copy(base, property0Value as P0, property1Value as P1)
+    }
+)
```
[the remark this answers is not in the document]
bjhham: There should always been an equal number of elements based on the count of provided properties.

ktor-shared/ktor-openapi-schema/common/src/io/ktor/openapi/Header.kt
```diff
@@ -0,0 +1,71 @@
+/*
+ * Copyright 2014-2025 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+package io.ktor.openapi
+
+import io.ktor.utils.io.*
+import kotlinx.serialization.Serializable
```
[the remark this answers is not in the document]
bjhham: It seems you're incorrect in this matter.  Please ignore.

ktor-shared/ktor-openapi-schema/common/src/io/ktor/openapi/Header.kt
```diff
@@ -0,0 +1,71 @@
...
+public data class Header(
+    /** A short description of the header. */
+    public val description: String? = null,
+    public val required: Boolean? = null,
+    public val deprecated: Boolean? = null,
+    public val allowEmptyValue: Boolean? = null,
+    public val explode: Boolean? = null,
+    public val examples: Map<String, ReferenceOr<ExampleObject>>? = null,
+    public val schema: ReferenceOr<Schema>? = null,
+) {
```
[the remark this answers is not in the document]
bjhham: I'll avoid implementing this for now.

ktor-http/common/src/io/ktor/http/HttpMethod.kt
```diff
@@ -59,7 +61,7 @@ private val REQUESTS_WITHOUT_BODY = setOf(
     HttpMethod.Get,
     HttpMethod.Head,
     HttpMethod.Options,
-    HttpMethod("TRACE"),
+    HttpMethod.Trace,
```
osipxd, line 64: Could you also update `HttpMethodTest` to reflect this change?

ktor-server/ktor-server-core/common/src/io/ktor/server/routing/RouteSelector.kt
```diff
@@ -252,12 +252,34 @@ public abstract class RouteSelector {
     public abstract suspend fun evaluate(context: RoutingResolveContext, segmentIndex: Int): RouteSelectorEvaluation
 }
 
+/**
+ * Exposes a textual representation of the path fragment contributed by this selector.
+ * Used for tooling (e.g., documentation generation).
+ */
+public interface RoutePathComponent {
+    public fun pathString(): String
```
osipxd: Probably `RouteSelector.toPathElement()` should be updated to use this function

ktor-server/ktor-server-core/common/src/io/ktor/server/routing/RouteSelector.kt
```diff
@@ -435,6 +465,8 @@ public data class PathSegmentParameterRouteSelector(
         )
     }
 
+    override fun pathString(): String = "${prefix ?: ""}{$name}${suffix ?: ""}"
+
     override fun toString(): String = "${prefix ?: ""}{$name}${suffix ?: ""}"
```
osipxd, line 469: Minor. Maybe reuse the logic here?
```suggestion
    override fun toString(): String = pathString()
```

ktor-server/ktor-server-plugins/ktor-server-docs/common/src/io/ktor/annotate/RouteAnnotationApi.kt
```diff
@@ -0,0 +1,191 @@
...
+ * Attribute key for storing [Operation] in a [Route].
+ */
+public val EndpointAnnotationAttributeKey: AttributeKey<Operation> =
+    AttributeKey<Operation>("operation-docs")
+
+/**
+ * Annotate a [Route] with an OpenAPI [Operation].
+ */
+public fun Route.annotate(configure: Operation.Builder.() -> Unit): Route {
+    attributes[EndpointAnnotationAttributeKey] = Operation.build(configure)
```
osipxd: What if a user has called `annotate` several times on a route? Should we "append" endpoint annotations?

ktor-server/ktor-server-plugins/ktor-server-docs/common/src/io/ktor/annotate/RouteAnnotationApi.kt
```diff
@@ -0,0 +1,191 @@
...
+        }
+        is PathSegmentParameterRouteSelector,
+        is PathSegmentOptionalParameterRouteSelector -> Operation.Companion.build {
+            parameters {
+                path(paramSelector.name) {
+                    required = paramSelector is PathSegmentOptionalParameterRouteSelector
+                }
+            }
+        }
+        is HttpHeaderRouteSelector -> Operation.Companion.build {
```
osipxd: Minor. Here and above `Companion` can be omitted

ktor-server/ktor-server-plugins/ktor-server-docs/common/src/io/ktor/annotate/RouteAnnotationApi.kt
osipxd: This might be useful to move common annotations into an extensions function

ktor-server/ktor-server-plugins/ktor-server-docs/build.gradle.kts
```diff
@@ -0,0 +1,22 @@
...
+
+plugins {
+    id("ktorbuild.project.server-plugin")
+    id("kotlinx-serialization")
+}
+
+kotlin {
+    sourceSets {
+        commonMain.dependencies {
+            implementation(projects.ktorOpenapiSchema)
```
osipxd: Some types from `ktor-openapi-schema` are used in public API of this module, so I think it's better to change the dependency scope to `api`
```suggestion
            api(projects.ktorOpenapiSchema)
```

ktor-server/ktor-server-plugins/ktor-server-routing-annotate/build.gradle.kts
```diff
@@ -0,0 +1,22 @@
+/*
+ * Copyright 2014-2025 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+description = "Endpoint documentation API for Ktor"
```
osipxd, line 4: Nit. Name `ktor-server-docs` feels a bit too generic
We only have one package `io.ktor.annotate` by now, so maybe rename this module to `ktor-server-route-annotation`? Alternatively we could change the package to `io.ktor.docs.annotate` or `io.ktor.docs` if there are more APIs planned in this module.

ktor-shared/ktor-openapi-schema/common/src/io/ktor/openapi/Operation.kt
```diff
@@ -0,0 +1,1029 @@
...
+ * @property responses The list of possible responses as they are returned from executing this operation, keyed by status code.
+ * @property deprecated Marks the operation as deprecated if true.
+ * @property security A declaration of which security mechanisms can be used for this operation.
+ * @property servers An alternative server array to service this operation.
+ * @property extensions Specification-extensions for this operation (keys must start with `x-`).
+ */
+@Serializable(Operation.Companion.Serializer::class)
+@OptIn(ExperimentalSerializationApi::class)
+@KeepGeneratedSerializer
+public data class Operation(
```
osipxd, line 39: It [might be difficult](https://kotlinlang.org/docs/api-guidelines-backward-compatibility.html#avoid-using-data-classes-in-your-api) to maintain binary compatibility with data classes
Should we use a plain class instead (here and for other models)?

ktor-shared/ktor-openapi-schema/common/src/io/ktor/openapi/Operation.kt
```diff
@@ -0,0 +1,1029 @@
...
+        /**
+         * Builds an [Operation] instance using the provided DSL [configure] block.
+         */
+        public fun build(configure: Builder.() -> Unit): Operation =
+            Builder().apply(configure).build()
+
+        internal object Serializer : ExtensibleMixinSerializer<Operation>(
+            generatedSerializer(),
+            { op, extensions -> op.copy(extensions = extensions) }
+        )
```
osipxd, line 64: Minor. Can be moved out of the companion object?

bjhham:
Apologies for the size, there were more types than expected and I couldn't break it into chunks 😓

ktor-shared/ktor-openapi-schema/common/src/io/ktor/openapi/AdditionalProperties.kt
```diff
@@ -0,0 +1,63 @@
...
+ */
+@Serializable(with = AdditionalProperties.Companion.Serializer::class)
+public sealed interface AdditionalProperties {
+    /** Wrapper for a boolean flag that allows or denies extra properties on an object schema. */
+    @JvmInline
+    public value class Allowed(public val value: Boolean) : AdditionalProperties
+
+    /** Wrapper for a schema reference describing the type of additional properties. */
+    @JvmInline
+    public value class PSchema(public val value: ReferenceOr<Schema>) : AdditionalProperties
```
osipxd, line 32: What `P` stands for in the name `PSchema`?

ktor-shared/ktor-openapi-schema/jvm/test-resources/petstore.json
```diff
@@ -0,0 +1,236 @@
+{
+  "openapi": "3.0.0",
```
osipxd, line 2: Btw, what is the max supported OpenAPI version? I see that [v3.2.0 has been published](https://spec.openapis.org/oas/v3.2.0.html) in September.

ktor-shared/ktor-openapi-schema/build.gradle.kts
```diff
@@ -0,0 +1,22 @@
...
+    id("ktorbuild.project.library")
+    id("kotlinx-serialization")
+}
+
+kotlin {
+    sourceSets {
+        commonMain.dependencies {
+            implementation(projects.ktorIo)
+            implementation(projects.ktorHttp)
+            implementation(libs.kotlin.serialization)
```
osipxd: It is a Gradle plugin, so let's remove it from dependencies
```suggestion
```

ktor-shared/ktor-openapi-schema/build.gradle.kts
```diff
@@ -0,0 +1,22 @@
...
+plugins {
+    id("ktorbuild.project.library")
+    id("kotlinx-serialization")
+}
+
+kotlin {
+    sourceSets {
+        commonMain.dependencies {
+            implementation(projects.ktorIo)
+            implementation(projects.ktorHttp)
```
osipxd: Let's use `api` as `ContentType` is a part of public API
```suggestion
            api(projects.ktorHttp)
```

ktor-shared/ktor-openapi-schema/build.gradle.kts
```diff
@@ -0,0 +1,22 @@
...
+
+plugins {
+    id("ktorbuild.project.library")
+    id("kotlinx-serialization")
+}
+
+kotlin {
+    sourceSets {
+        commonMain.dependencies {
+            implementation(projects.ktorIo)
```
osipxd, line 16: It's sad that we have to depend on `ktor-io` to use annotations `@KtorDsl` and `@InternalAPI`.
Created an issue: [KTOR-8995](https://youtrack.jetbrains.com/issue/KTOR-8995) Move commonly used annotations out of ktor-io module

ktor-shared/ktor-openapi-schema/build.gradle.kts
```diff
@@ -0,0 +1,22 @@
...
+            implementation(projects.ktorHttp)
+            implementation(libs.kotlin.serialization)
+            implementation(libs.kotlinx.serialization.core)
```
osipxd, line 17: Probably, also should have `api` scope

ktor-utils/common/src/io/ktor/util/collections/TreeLike.kt
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+package io.ktor.util.collections
+
+/**
+ * Represents a tree-like hierarchical structure, where each node has a parent and a collection of child nodes.
+ *
+ * @param T The type parameter representing the node type, which must also implement [TreeLike].
+ */
+public interface TreeLike<out T : TreeLike<T>> {
```
osipxd, line 12: Should be marked as internal API?

ktor-shared/ktor-openapi-schema/common/src/io/ktor/openapi/GenericElement.kt
```diff
@@ -0,0 +1,593 @@
...
+ */
+public interface GenericElementSerialAdapter {
+    public fun <T> trySerializeToElement(encoder: Encoder, value: T, serializer: KSerializer<T>): GenericElement?
+    public fun tryDeserialize(decoder: Decoder): GenericElement?
+}
+
+/**
+ * [GenericElementSerialAdapter] for standard kotlinx-json.
+ */
+public object JsonElementSerialAdapter : GenericElementSerialAdapter {
```
marychatte, line 164: KDoc is duplicated for GenericMapDecoderAdapter

ktor-shared/ktor-openapi-schema/common/src/io/ktor/openapi/ReferenceOr.kt
```diff
@@ -0,0 +1,96 @@
...
+
+        internal class Serializer<T>(private val dataSerializer: KSerializer<T>) :
+            KSerializer<ReferenceOr<T>> {
+            @OptIn(InternalSerializationApi::class, ExperimentalSerializationApi::class)
+            override val descriptor: SerialDescriptor =
+                buildSerialDescriptor("io.ktor.openapi.ReferenceOr", SerialKind.CONTEXTUAL)
+            private val refDescriptor = buildClassSerialDescriptor("io.ktor.openapi.ReferenceOr.Reference") {
+                element<String>(RefKey)
+            }
+            private val recursiveRefDescriptor = buildClassSerialDescriptor("io.ktor.openapi.ReferenceOr.Reference") {
```
marychatte: Both `refDescriptor` and `recursiveRefDescriptor` use the same name `io.ktor.openapi.ReferenceOr.Reference`. Is it okay?

ktor-shared/ktor-openapi-schema/common/src/io/ktor/openapi/GenericElement.kt
marychatte, line 164: KDoc is duplicated for `GenericMapDecoderAdapter`

ktor-shared/ktor-openapi-schema/common/src/io/ktor/openapi/ContentTypeSerializer.kt
```diff
@@ -0,0 +1,27 @@
...
+import kotlinx.serialization.descriptors.PrimitiveKind
+import kotlinx.serialization.descriptors.PrimitiveSerialDescriptor
+import kotlinx.serialization.descriptors.SerialDescriptor
+import kotlinx.serialization.encoding.Decoder
+import kotlinx.serialization.encoding.Encoder
+
+/**
+ * Serializer for [ContentType] that uses [ContentType.toString] and [ContentType.parse]
+ */
+public class ContentTypeSerializer : KSerializer<ContentType> {
```
marychatte: Can it be a singleton?

ktor-server/ktor-server-plugins/ktor-server-docs/common/src/io/ktor/annotate/RouteAnnotationApi.kt
```diff
@@ -0,0 +1,191 @@
...
+        trace = trace ?: other.trace,
+        servers = mergeNullable(servers, other.servers) { a, b -> a + b },
+        parameters = mergeNullable(parameters, other.parameters) { a, b -> a + b },
+        extensions = mergeNullable(extensions, other.extensions) { a, b -> a + b }
+    )
+
+internal operator fun Responses.plus(other: Responses) =
+    Responses(
+        default = default ?: other.default,
+        responses = mergeNullable(responses, other.responses) { a, b -> a + b },
```
marychatte: Can it be duplicates after merging?

ktor-server/ktor-server-plugins/ktor-server-docs/common/src/io/ktor/annotate/RouteAnnotationApi.kt
```diff
@@ -0,0 +1,191 @@
...
+                "/$it"
+            }
+        }
+
+internal fun RouteSelector.segmentOrNull(): String? =
+    when (this) {
+        is RoutePathComponent -> pathString()
+        is CompositeRouteSelector -> subSelectors()
+            .filterIsInstance<RoutePathComponent>()
+            .singleOrNull()?.pathString()
```
marychatte: Why do we take only one segment and drop the other?

marychatte (approved):
LGTM👍👍

ktor-server/ktor-server-core/common/src/io/ktor/server/routing/RouteSelector.kt
bjhham: The behaviour is slightly different... The `toPathElement()` is used for our internal string representation of an endpoint, but this `pathString()` function is used for the OpenAPI paths, which doesn't support some things.

Maybe a better way of handling this would be to have the function accept an encoding enum and potentially return unsupported, so we can support other documentation formats including our own internal one.

ktor-server/ktor-server-plugins/ktor-server-docs/common/src/io/ktor/annotate/RouteAnnotationApi.kt
bjhham: Oh, good catch.  That was actually my intention.  I'll update it to append and include a test.

ktor-utils/common/src/io/ktor/util/collections/TreeLike.kt
bjhham: I was thinking it would serve in the public API for custom routing inference tools.  Since we're only providing OpenAPI / Swagger, people may want to build other extensions.

ktor-shared/ktor-openapi-schema/common/src/io/ktor/openapi/ReferenceOr.kt
bjhham: Honestly not sure where it's used, but it doesn't look right so I'll fix it 👍

ktor-server/ktor-server-plugins/ktor-server-docs/common/src/io/ktor/annotate/RouteAnnotationApi.kt
bjhham: Hmm yeah looks like they'd be overridden now, but we ought to combine details when there's collisions.

ktor-server/ktor-server-plugins/ktor-server-docs/common/src/io/ktor/annotate/RouteAnnotationApi.kt
bjhham: In the case of composite selectors it's either AND / OR, so if there's anything but a single path element here we don't know which one we can use to represent the path.  We could always traverse the tree of possible paths from these nodes and include them all, but I've never seen one of these route selectors in the wild and didn't knew they existed until now 😆. Maybe I'll add support here idk

ktor-shared/ktor-openapi-schema/build.gradle.kts
bjhham: I think maybe it's not necessary to expose it.  The most common interaction will be just the objects and builders and all the serialization logic can remain hidden internally.

ktor-shared/ktor-openapi-schema/common/src/io/ktor/openapi/Operation.kt
```diff
@@ -0,0 +1,1029 @@
+/*
+ * Copyright 2014-2025 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.openapi
+
+import io.ktor.http.*
+import io.ktor.openapi.ReferenceOr.*
+import io.ktor.utils.io.*
```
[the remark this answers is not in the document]
bjhham: No, this is correct.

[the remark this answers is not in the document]
osipxd: The rabbit is also confused)

ktor-server/ktor-server-core/common/src/io/ktor/server/routing/RouteSelector.kt
bjhham: Ended up exposing the `RoutingNode.path()` function but introducing an argument for the format which can either be the default Ktor format (replaces `toPathElement()`) or OpenAPI or any custom format.  This seems like a better abstraction.

ktor-server/ktor-server-plugins/ktor-server-routing-annotate/build.gradle.kts
bjhham: Yes, I agree, `-docs` is too general and non-descriptive...  I'll have to think a bit about the naming.  Really, this is only responsible for injecting descriptive metadata into the routing tree and has nothing to do with documentation.  Even if this isn't used for OpenAPI it's referencing its schema, so maybe it should be something like `ktor-server-openapi-metadata-routing`...

ktor-server/ktor-server-core/common/src/io/ktor/server/routing/RoutingNode.kt
```diff
@@ -273,9 +274,6 @@ public typealias RoutingHandler = suspend RoutingContext.() -> Unit
  *
  * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.server.routing.Route)
  */
-@Suppress("ktlint:standard:no-consecutive-comments")
-// TODO KTOR-8809: Uncomment the annotation
-// @KtorDsl
```
osipxd, line 278: Did you remove this by intention or accidentally?

bjhham: Did not intend to remove this 👎

PR #5130, opened 2025-10-17
Version 3.3.2-SNAPSHOT

osipxd:


bjhham (approved):

bjhham:
BTW, the `.gitattributes` didn't seem to have an effect when merging within github, but it is very easy to resolve the conflict on merges.

PR #5134, opened 2025-10-20
KTOR-2404 Add fallback logic in OAuth

marychatte:
**Subsystem**
Server, OAuth

**Motivation**
[KTOR-2404](https://youtrack.jetbrains.com/issue/KTOR-2404/Ktor-Oauth2-feature-sends-401-response-when-the-client-secret-is-invalid)

**Solution**
Ktor already supports `authenticate(optional = true)`, but that strategy is not appropriate for OAuth failure handling, because optional is defined to skip challenges only when the client provides no credentials, like “guest” requests, not real failures.

The idea is to add fallback which is triagered on `AuthenticationFailedCause.Error`, because `AuthenticationFailedCause.NoCredential` means the start of OAuth flow and `AuthenticationFailedCause.InvalidCredentials` means the restart of the auth flow

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-di/api/ktor-server-di.api
+++ b/ktor-server/ktor-server-plugins/ktor-server-di/api/ktor-server-di.api
@@ -24,6 +24,7 @@ public abstract interface class io/ktor/server/plugins/di/DependencyConflictPoli
 public final class io/ktor/server/plugins/di/DependencyConflictPolicyKt {
 	public static final fun getDefaultConflictPolicy ()Lio/ktor/server/plugins/di/DependencyConflictPolicy;
 	public static final fun getIgnoreConflicts ()Lio/ktor/server/plugins/di/DependencyConflictPolicy;
+	public static final fun getOverridePrevious ()Lio/ktor/server/plugins/di/DependencyConflictPolicy;
 }
 
 public abstract interface class io/ktor/server/plugins/di/DependencyConflictResult {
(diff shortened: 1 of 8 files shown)


ktor-server/ktor-server-plugins/ktor-server-auth/common/test/io/ktor/tests/auth/OAuth2Test.kt
```diff
@@ -517,11 +587,7 @@ class OAuth2Test {
             ).formUrlEncode()
         )
 
-        assertEquals(HttpStatusCode.Found, call.status)
-        assertNotNull(call.headers[HttpHeaders.Location])
-        assertTrue {
-            call.headers[HttpHeaders.Location]!!.startsWith("https://login-server-com/authorize")
-        }
+        assertEquals(HttpStatusCode.Unauthorized, call.status)
```
marychatte, line 590: I think we shouldn't re-challenge on an `OAuth2RedirectError` (e.g., user clicked "Deny")

ktor-server/ktor-server-plugins/ktor-server-auth/common/test/io/ktor/tests/auth/OAuth2Test.kt
```diff
@@ -363,6 +366,73 @@ class OAuth2Test {
...
+                        name = "oauth2",
+                        authorizeUrl = "http://localhost/authorize",
+                        accessTokenUrl = "http://localhost/access_token",
+                        clientId = "clientId1",
+                        clientSecret = "clientSecret1",
+                        requestMethod = HttpMethod.Post,
+                        passParamsInURL = true
+                    )
+                }
+                fallback = { cause ->
```
bjhham, line 387: I would say it could be a function to be consistent with `StaticContent`, but that would look weird here 🙁

ktor-server/ktor-server-plugins/ktor-server-auth/common/test/io/ktor/tests/auth/OAuth2Test.kt
bjhham: Will the re-challenge still kick in for other errors?  I just wanna make sure nobody could be relying on the changed behaviour.

marychatte: There can be other [reasons](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.2.1):
1. invalid_request
2. unauthorized_client
3. access_denied
4. unsupported_response_type
5. invalid_scope
6. server_error
7. temporarily_unavailable

I think for reasons 1-5 we shouldn't do a retry because we will loop. In reasons 6 and 7 I'm not sure, for me it looks like no retry...
Someone could have relied on the old behavior with an infinite loop, but it looks like a bug, and if needed, can be done like:
```kotlin
oauth("login") {
  // ...
  fallback = { cause ->
    if (cause is OAuth2RedirectError) {
      // old behavior
      respondRedirect("/login")
    }
  }
}
``` 
@bjhham wdyt?

bjhham: Yep, that makes sense to me 👍

PR #5136, opened 2025-10-20
KTOR-8975 BodyProgress: Keep the response body replayable

osipxd:
**Subsystem**
Client, BodyProgress plugin

**Motivation**
[KTOR-8975](https://youtrack.jetbrains.com/issue/KTOR-8975) Response body channel is canceled while the body is being saved when having HttpRequestRetry and onDownload

**Solution**
Fixed `withObservableDownload` to create a fresh observable channel on each `rawContent` access instead of reusing a single instance. Previously, when `HttpRequestRetry` cancelled the channel during validation (see `throwOnInvalidResponseBody`), it became unavailable for subsequent reads by `SaveBody`, causing `ClosedByteChannelException`. The fix passes a lambda to `replaceResponse` that creates a new channel on each access, making cancellation safe.


marychatte (approved):
Nice catch!

PR #5137, opened 2025-10-21
KTOR-8990 Support for raw types in DI

bjhham:
**Subsystem**
Server, DI

**Motivation**
[KTOR-8990](https://youtrack.jetbrains.com/issue/KTOR-8990) DI: API can't allow resolution with DependencyKey without Ktype

**Solution**
Introduce a new default covariance mapping so the raw type (missing KType) can be accessed.


ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/utils/Types.kt
```diff
@@ -40,3 +40,25 @@ public expect fun TypeInfo.toNullable(): TypeInfo?
...
+ *
+ * @return A new [TypeInfo] with [kotlinType] set to null if the current type has a [kotlinType],
+ *         or null if the type is already raw (kotlinType is already null).
+ */
+@InternalAPI
+public fun TypeInfo.toRawType(): TypeInfo? =
+    if (kotlinType != null) {
+        TypeInfo(type, kotlinType = null)
+    } else {
+        null
```
osipxd: It's unexpected for me that this function returns `null` if the given `TypeInfo` is already raw. I'd expect it returns `this` instead. What do you think?

bjhham: Hmm the problem there is that we want to avoid including it if it's already raw.  I'll just push the check up the stack for better readability.

osipxd (approved):

PR #5138, opened 2025-10-21
Update dependencies

osipxd:
**Motivation**
[KTOR-8916](https://youtrack.jetbrains.com/issue/KTOR-8916) Netty: java.lang.VerifyError is thrown on Android since 3.3.0
Reduce noise from Renovate.

**Solution**
Update dependencies in one PR.

_Probably we should configure renovate to create a single PR at least for all patch versions._

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -51,7 +51,7 @@ micrometer = "1.15.5"
 jansi = "2.4.2"
 typesafe = "1.4.5"
 
-mockk = "1.14.5"
+mockk = "1.14.6"
 
 java-jwt = "4.5.0"
 


gradle/libs.versions.toml
```diff
@@ -2,55 +2,56 @@
...
 kotlinx-io = "0.8.0"
 coroutines = "1.10.2"
 atomicfu = "0.29.0"
 serialization = "1.9.0"
 dokka = "2.1.0"
-kover = "0.9.1"
+kover = "0.9.3"
 kotlinter = "5.2.0"
-kotlinx-browser = "0.3.1"
+kotlinx-browser = "0.5.0"
```
osipxd, line 15: Finally, we can update this library after updating to Kotlin 2.2.20

gradle/libs.versions.toml
```diff
@@ -2,55 +2,56 @@
...
 
 json-simple = "1.1.1"
-gson = "2.13.1"
-jackson = "2.19.2"
+gson = "2.13.2"
+jackson = "2.20.0"
+jackson-annotations = "2.20"
 
-junit = "5.13.4"
+junit = "5.14.0"
```
osipxd, line 44: Can't update to JUnit 6.0.0 as it has raised the minimal required Java version to 17. I think we still need to run tests against Java 8, right?

bjhham: Yeah let's keep it at JUnit 5 for now.  I'm curious to see how the coroutine support works though...

bjhham:
> Probably we should configure renovate to create a single PR at least for all patch versions.

Is this possible???

osipxd:
> Is this possible???

Yes, I'll do it

PR #5154, opened 2025-10-23
KTOR-7639 Separate path and other selectors in traces

marychatte:
**Subsystem**
Server

**Motivation**
[KTOR-7639](https://youtrack.jetbrains.com/issue/KTOR-7639)

**Solution**
Separate selectors into `pathSelectors` and `constraintsSelectors`.
Previously, for example, `/route/LocalPortRouteSelector(port=80)/port`, now `/route/port [LocalPortRouteSelector(port=80)]`
New structure is `path [constraint1, constraint2, ...]`


osipxd:
Just thinking at loud.
1. Maybe we should introduce a supertype or interface for selectors that should be included in path. That would make it possible to correctly render custom path selectors.
2. Maybe we can reuse for this interfaces introduced in #5125 (fyi @bjhham)

PR #5159, opened 2025-10-27
KTOR-5850 Fix SendCountExceedException in HttpRequestRetry

marychatte:
**Subsystem**
Client, HttpRequestRetry

**Motivation**
[KTOR-5850](https://youtrack.jetbrains.com/issue/KTOR-5850)

**Solution**
`HttpSend` plugin has its own counter for attempts, and the default value is `20`. Users can configure it:
```kotlin
install(HttpSend) {
    maxSendCount = 25
}
```
`HttpRequestRetry` plugin also has its own value `maxRetries`, which can be bigger than `20`. The solution is to use `HttpRequestRetry.maxRetries` for `HttpSend` if it's configured

### DIFF
--- a/ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/HttpRequestRetryTest.kt
+++ b/ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/HttpRequestRetryTest.kt
@@ -544,7 +544,7 @@ class HttpRequestRetryTest {
 
         test { client ->
             assertEquals(HttpStatusCode.OK, client.get("/").status)
-            println(counter)
+            assertEquals(maxRetriesCount, counter)
         }
     }
 }


ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/HttpRequestRetryTest.kt
```diff
@@ -516,4 +516,35 @@ class HttpRequestRetryTest {
...
+                delayMillis { 0L }
+            }
+            install(HttpSend) {
+                maxSendCount = 1
+            }
+        }
+
+        test { client ->
+            assertEquals(HttpStatusCode.OK, client.get("/").status)
+            println(counter)
```
bjhham: Let's change this `println` into an assert 👀

bjhham (changes requested):

PR #5161, opened 2025-10-29
KTOR-8891 Make it possible to build without Android target

osipxd:
**Subsystem**
Infrastructure

**Motivation**
[KTOR-8891](https://youtrack.jetbrains.com/issue/KTOR-8891) Make it possible to disable Android target

**Solution**
Check for `ANDROID_HOME` or `sdk.dir` before applying Android Gradle Plugin.
Also fixed `KtorTargets` to properly handle Android target in the hierarchy.

It's better to review commit-by-commit

### DIFF
--- a/build-logic/build.gradle.kts
+++ b/build-logic/build.gradle.kts
@@ -33,5 +33,6 @@ kotlin {
 
     compilerOptions {
         allWarningsAsErrors = true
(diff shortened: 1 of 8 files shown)


osipxd:
FYI @zibet27

PR #5166, opened 2025-10-31
KTOR-8838 Fix exception handling in client cache

marychatte:
**Subsystem**
Client, Cache

**Motivation**
[KTOR-8838](https://youtrack.jetbrains.com/issue/KTOR-8838) 

**Solution**
Wrap `coroutineScope` in try/catch, so exceptions thrown inside the launch also could be caught

### DIFF
--- a/ktor-client/ktor-client-core/jvm/src/io/ktor/client/plugins/cache/storage/FileCacheStorage.kt
+++ b/ktor-client/ktor-client-core/jvm/src/io/ktor/client/plugins/cache/storage/FileCacheStorage.kt
@@ -160,9 +160,8 @@ private class FileCacheStorage(
                     channel.copyTo(output)
                 }
             }
-        } catch (cause: CancellationException) {
-            throw cause
         } catch (cause: Exception) {
+            if (cause is CancellationException) currentCoroutineContext().ensureActive()
             LOGGER.trace { "Exception during saving a cache to a file: ${cause.stackTraceToString()}" }
         } finally {
             channel.close()


ktor-client/ktor-client-core/jvm/src/io/ktor/client/plugins/cache/storage/FileCacheStorage.kt
```diff
@@ -143,21 +145,27 @@ private class FileCacheStorage(
...
                     }
-                    channel.close()
+                    channel.copyTo(output)
                 }
-                channel.copyTo(output)
             }
+        } catch (cause: CancellationException) {
+            throw cause
         } catch (cause: Exception) {
             LOGGER.trace { "Exception during saving a cache to a file: ${cause.stackTraceToString()}" }
```
osipxd: Probably we should use `ensureActive()` instead of rethrowing `CancellationException`.
```suggestion
        } catch (cause: Exception) {
            if (cause is CancellationException) currentCoroutineContext().ensureActive()
            LOGGER.trace { "Exception during saving a cache to a file: ${cause.stackTraceToString()}" }
```
See [this discussion](https://github.com/Kotlin/kotlinx.coroutines/issues/3658#issuecomment-1943562914)

marychatte: Nice article and discussion, thanks!

osipxd (approved):

PR #5167, opened 2025-10-31
WebRTC Client. Remove redundant targets.

zibet27:
**Subsystem**
WebRTC Client.

**Motivation**
We have some targets like jvm, desktopNative, tvos, etc that are published but don't have clients, which is confusing for users.

**Solution**
Remove those targets.


zibet27:
> Let's fix this also in the release branch

@osipxd Should I manually merge this branch to `release`, or will `main` be merged to it?

osipxd:
> @osipxd Should I manually merge this branch to `release`, or will `main` be merged to it?

Main will be merged into the release branch only after 3.4.0 release. So let's cherry-pick this change manually to include it in 3.3.x patch releases.

PR #5170, opened 2025-11-04
KTOR-9054 Fix application call coroutine context

bjhham:
**Subsystem**
Server, Netty

**Motivation**
[KTOR-9054](https://youtrack.jetbrains.com/issue/KTOR-9054) Netty: `application` and `call` coroutine scopes share the same job instance since 3.3.0

**Solution**
Reverted mistake made in https://github.com/ktorio/ktor/pull/5070 where I assumed the engine call would follow the job hierarchy and the pipeline context did not, but it is in fact the other way around.  It seems the change was not necessary for the auto-reloading fix so it should be fine.

I also included a test to ensure that call coroutine jobs are children of the application job and found this isn't the case all the time, so I fixed the standard Jetty engine and ignored other tests to mitigate impact.  Logged a follow up issue [KTOR-9058](https://youtrack.jetbrains.com/issue/KTOR-9058) to fix CIO.

### DIFF
--- a/ktor-server/ktor-server-tomcat-jakarta/jvm/test/io/ktor/tests/server/tomcat/jakarta/TomcatEngineTest.kt
+++ b/ktor-server/ktor-server-tomcat-jakarta/jvm/test/io/ktor/tests/server/tomcat/jakarta/TomcatEngineTest.kt
@@ -173,6 +173,9 @@ class TomcatSustainabilityTestSuite :
     override fun testBlockingConcurrency() {
         super.testBlockingConcurrency()
     }
+
+    @Ignore
+    override fun validateCallCoroutineContext() {}
 }
 
 class TomcatConfigTest : ConfigTestSuite(Tomcat)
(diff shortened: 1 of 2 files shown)


ktor-server/ktor-server-test-suites/jvm/src/io/ktor/server/testing/suites/SustainabilityTestSuite.kt
```diff
@@ -967,6 +967,32 @@ abstract class SustainabilityTestSuite<TEngine : ApplicationEngine, TConfigurati
         }
     }
 
+    @OptIn(ExperimentalCoroutinesApi::class)
+    @Test
+    open fun validateCallCoroutineContext() = runTest {
+        createAndStartServer {
+            get {
+                val applicationJob = application.coroutineContext[Job] ?: error("application job is empty")
+                val callJob = call.coroutineContext[Job] ?: error("call job is empty")
```
osipxd: Minor. Could be simplified a bit
```suggestion
                val applicationJob = application.coroutineContext.job
                val callJob = call.coroutineContext.job
```

osipxd (approved):

ktor-server/ktor-server-test-suites/jvm/src/io/ktor/server/testing/suites/SustainabilityTestSuite.kt
bjhham: Oh wow, I can't keep up with these improvements 😂

PR #5174, opened 2025-11-04
WebRTC Client. Remove redundant targets.

zibet27:
**Subsystem**
WebRTC Client

**Motivation**
[KTOR-9062](https://youtrack.jetbrains.com/issue/KTOR-9062) WebRTC Client. Remove redundant targets. 
The same as this [pull request](https://github.com/ktorio/ktor/pull/5167).

**Solution**
`git cherry-pick ...` can't help here because `release` doesn't have iOS target yet.


ktor-client/ktor-client-webrtc/gradle.properties
```diff
@@ -4,3 +4,7 @@
 target.js.nodeJs=false
 target.wasmJs.nodeJs=false
 target.android.unitTest=false
+target.androidNative=true
```
osipxd, line 7: Btw, forgot to ask in the previous PR why `androidNative` was enabled? I thought WebRTC was not implemented on this target

osipxd:
Could you create a YouTrack issue to track this change in release notes?

ktor-client/ktor-client-webrtc/gradle.properties
zibet27: Yeah, it actually can be removed at this branch.
But on main with iOS target enabled we had to keep because KMP couldn't generate klib definitions correctly on Linux. See https://jetbrains.slack.com/archives/CPM016M29/p1760952385047569
Do you think we should also remove android native here?

bjhham (approved):

PR #5176, opened 2025-11-05
KTOR-8657: deps(ktor-server-default-headers): Remove `kotlinx-datetime`

Ryuheeeei:
**Subsystem**

`ktor-server-plugins` > `ktor-server-default-headers` (`io.ktor.server.plugins.defaultheaders`)

**Motivation**

[KTOR-8657](https://youtrack.jetbrains.com/issue/KTOR-8657/Remove-kotlinx-datetime-from-ktor-server-default-headers-dependencies)
> It should never have been exposed through transitive dependencies was it is an internal implementation detail.

**Solution**

Remove kotlinx-datetime from ktor-server-default-headers dependencies

**Note**

* It seems this change should be included in next minor version. Sorry if I create PR too early. 🙏
https://youtrack.jetbrains.com/issue/KTOR-8656/kotlinx.datetime-is-not-available-transitively-in-3.2.1
> So we should bring it back and remove it only in the next minor release.


osipxd (approved):
Thank you! 👍 
It's okay to merge this PR into main. We track patch releases in a separate branch - `release/3.x`

Ryuheeeei:
@osipxd 
Thank you for your review!
However, maybe I don't have permission to merge this pull request 🙏

PR #5178, opened 2025-11-05
KTOR-5241 Deprecate Base64 encode/decode helpers, ensure strict decoding

marychatte:
**Subsystem**
Utils, Base64

**Motivation**
[KTOR-5241](https://youtrack.jetbrains.com/issue/KTOR-5241) 

**Solution**
Pull request [4721](https://github.com/ktorio/ktor/pull/4721) in `3.1.2` allowed the acceptance of both standard and URL-safe Base64 alphabets. It can be discussed if we want separate methods for that or not.
For example, Base64 utility in the stdlib says:
> That also implies that a Base64 scheme could not decode data encoded by another Base64 scheme if their alphabets differ.

And it separates the standard and URL-safe variant:
```kotlin
println(Base64.Default.encode(byteArrayOf(-1, 0, -2, 0))) // /wD+AA==
println(Base64.UrlSafe.encode(byteArrayOf(-1, 0, -2, 0))) // _wD-AA==
```

The idea is to introduce new methods that utilize the standard library instead of the previous ambiguous ones and deprecate the old ones.

[!] But maybe we don't really want to split the method and can define behavior as:
```kotlin
public fun String.decodeBase64Bytes(): ByteArray =
    runCatching { decodeStandardBase64Bytes() }.getOrElse { decodeUrlSafeBase64Bytes() }
```
So it will fail on non-Base64 strings (initial problem in ticket)


mltheuser:
Looking through the changes, I still think it would be clearer if both the ktor library and the users were to use the standard library encode/decode functions directly.

osipxd:
I think it's better to deprecate these extensions in 3.4.0
Deprecating things in a patch release might be unexpected

e5l (changes requested):
please check the CI

rururux:
I believe the safest path is to apply the Deprecated label without modifying the function's internal implementation.
​Users who haven't migrated yet will rely on the current implementation's exact behavior. By swapping out the custom logic for the standard library now, we risk introducing unintended subtle changes in behavior (despite passing tests), such as unforeseen failures in edge cases, variances in runtime performance, or changes to exception types/messages.
​To ensure maximum stability and zero unexpected breaks for existing consumers, I propose we keep the current implementation and remove it entirely in the next major version.

marychatte:
@rururux Updated implementation preserves the same decoding logic (accepting both standard and URL-safe variants), but now correctly rejects invalid input that previously produced corrupted data.

marychatte:
@e5l I checked the CI, failing tests are flaky

e5l:
lgtm, thank you for checking

PR #5179, opened 2025-11-07
KTOR-8947 Fix http/2 header name check with content-encoding

Komdosh:
**Subsystem**
Client

**Motivation**
HTTP/2 header name can contain `:` symbol, but now it throws an error.

**Solution**
Just check that the unpermitted char is not a first colon in the key

link to the issue https://youtrack.jetbrains.com/issue/KTOR-9077/Content-Encoding-HTTP-2-pseudo-header-throws-an-error

### DIFF
--- a/ktor-http/common/src/io/ktor/http/HttpHeaders.kt
+++ b/ktor-http/common/src/io/ktor/http/HttpHeaders.kt
@@ -160,7 +160,7 @@ public object HttpHeaders {
             if (ch <= ' ' || isDelimiter(ch)) {
                 // HTTP/2 pseudo header can start with `:`
                 val isPseudoHeader = index == 0 && ch == ':'
-                if(!isPseudoHeader) {
+                if (!isPseudoHeader) {
                     throw IllegalHeaderNameException(name, index)
                 }
             }


Komdosh:
KTOR-8947

osipxd (changes requested):
Hey @Komdosh, thank you for the fix! 🎉 
I think `Headers` shouldn't contain pseudo-headers and it's better to filter them out when mapping headers in a `HttpClientEngine` implementation (something similar is already done for Netty server in https://github.com/ktorio/ktor/pull/3186).

It would be also great to add a test verifying that all client engines ignore pseudo-headers. To do so we should introduce HTTP/2 endpoints in the `ktor-test-server`. I'll try to make a PR soon 🧑‍💻

Komdosh:
> Hey @Komdosh, thank you for the fix! 🎉 
> 
> I think `Headers` shouldn't contain pseudo-headers and it's better to filter them out when mapping headers in a `HttpClientEngine` implementation (something similar is already done for Netty server in https://github.com/ktorio/ktor/pull/3186).
> 
> 
> 
> It would be also great to add a test verifying that all client engines supporting HTTP/2 ignore pseudo-headers. To do so we should introduce HTTP/2 endpoints in the `ktor-test-server`. I'll try to make a PR soon 🧑‍💻 

Great, I'll close this then

osipxd:
Sorry for misunderstanding. I meant I would open a PR adding a possibility to test HTTP/2 features. You could update your PR after that if you are interested in it

PR #5181, opened 2025-11-10
Server. Detect if a request was cancelled by the client.

zibet27:
**Subsystem**
Server, Server Netty, Server CIO

**Motivation**
[KTOR-2187](https://youtrack.jetbrains.com/issue/KTOR-2187) How to detect if a request was cancelled from the client on the Ktor server

**Solution**
Route-based plugin that sets a handler for the engines to run when the request is closed (successfully or not). The handler cancels the`call.coroutineContext` which fits the structured concurrency model. 
Engines that are based on servlets don't support it. Writing a response is the only way (I found) to check if the connection is active in that case.

### DIFF
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/UserAgent.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/UserAgent.kt
@@ -30,8 +30,10 @@ public val UserAgent: ClientPlugin<UserAgentConfig> = createClientPlugin("UserAg
     val agent = pluginConfig.agent
 
     onRequest { request, _ ->
-        LOGGER.trace("Adding User-Agent header: agent for ${request.url}")
-        request.header(HttpHeaders.UserAgent, agent)
+        if (!request.headers.contains(HttpHeaders.UserAgent)) {
+            LOGGER.trace("Adding User-Agent header: agent for ${request.url}")
+            request.header(HttpHeaders.UserAgent, agent)
+        }
     }
 }
 


ktor-http/ktor-http-cio/common/src/io/ktor/http/cio/RequestResponse.kt
```diff
@@ -55,7 +56,10 @@ public class Request internal constructor(
     public val version: CharSequence,
     headers: HttpHeadersMap,
     builder: CharArrayBuilder
-) : HttpMessage(headers, builder)
+) : HttpMessage(headers, builder) {
+    @InternalAPI
+    public var onClose: (() -> Unit)? = null
```
bjhham: Maybe we can drop the field and call the attribute directly from an extension function:

```
@InternalApi
fun Request.onClose() {
    call.attributes.getOrNull(HttpRequestCloseHandlerKey)?.invoke()
}
```

ktor-server/ktor-server-test-suites/jvm/src/io/ktor/server/testing/suites/HttpRequestLifecycleTest.kt
```diff
@@ -0,0 +1,133 @@
...
+import io.ktor.utils.io.*
+import kotlinx.coroutines.*
+import kotlinx.coroutines.CancellationException
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+import kotlin.time.Duration.Companion.milliseconds
+import kotlin.time.Duration.Companion.seconds
+
+abstract class HttpRequestLifecycleTest<TEngine : ApplicationEngine, TConfiguration : ApplicationEngine.Configuration>(
```
bjhham, line 28: It seems this test is a bit flaky on the CI.  It might just need an extended timeout.

bjhham (changes requested):
Looks good, just have a couple comments.

ktor-server/ktor-server-test-suites/jvm/src/io/ktor/server/testing/suites/HttpRequestLifecycleTest.kt
zibet27: If you mean the test that failed Java 8, 11 on Linux, it's actually `io.ktor.tests.server.netty.NettyConfigurationTest.configuredChannelEventuallyClosed()[jvm]`, not my test.
Anyway, it makes sense to increase the timeout to 10 seconds (in case of high load on the CI).

bjhham: Nah, I mean this test in particular failed but the automatic retry got it through.  Unfortunately, we don't have any kind of reporting on when tests are retried so the only way to notice is by watching them as they're running 😓

zibet27: Ohh, I didn't know that. I'll look in the logs; maybe there is a real problem.

ktor-http/ktor-http-cio/common/src/io/ktor/http/cio/RequestResponse.kt
zibet27: Unfortunately, we can't access `call` from `Request`. 
On the other hand, we just need `onClose` in **CIOApplicationEngine**, but it's created in **cio.backend.ServerPipeline**. Another way to pass it could be through **ServerRequestScope**(in this case, `onClose` can be internal and won't modify public API).

zibet27: I moved it in the second commit.

bjhham: How about using something like `ServerRequestScope.coroutingContext[Job].invokeOnCompletion {}` or possibly `launchOnCancellation`?

marychatte (commented):
lgtm, I have a question, should it also work for HTTP/2 in Netty?

ktor-server/ktor-server-core/common/src/io/ktor/server/http/HttpRequestLifecycle.kt
```diff
@@ -0,0 +1,102 @@
...
+     * When `true`, cancels the call coroutine context if the other peer resets the client connection.
+     * When `false` (default), request processing continues even if the connection is closed.
+     *
+     * Example:
+     * ```kotlin
+     * install(HttpRequestLifecycle) {
+     *     cancelCallOnClose = true
+     * }
+     * ```
+     */
```
e5l, line 34: It's not clear from the documentation where the cancellation exception will be thrown. Could you please add an example of handling the exception as well?

e5l: It is also unclear what the use case for the new property is - when should someone use it?

ktor-server/ktor-server-core/common/src/io/ktor/server/http/HttpRequestLifecycle.kt
```diff
@@ -0,0 +1,102 @@
...
+ * This is particularly useful for:
+ * - Long-running requests where the client may disconnect before completion
+ * - Streaming responses where detecting disconnection allows early cleanup
+ * - Resource-intensive operations that should be canceled when the client is no longer waiting
+ *
+ * ## Example
+ *
+ * ```kotlin
+ * install(HttpRequestLifecycle) {
+ *     cancelCallOnClose = true
```
e5l, line 65: I wonder if this should be the default, and we don't need an option to change it. @bjhham, what do you think?

ktor-server/ktor-server-core/common/src/io/ktor/server/http/HttpRequestLifecycle.kt
```diff
@@ -0,0 +1,102 @@
...
+     * ```
+     */
+    public var cancelCallOnClose: Boolean = false
+}
+
+/**
+ * Internal attribute key for storing the connection close handler callback.
+ */
+@InternalAPI
+public val HttpRequestCloseHandlerKey: AttributeKey<() -> Unit> = AttributeKey<() -> Unit>("HttpRequestCloseHandler")
```
e5l, line 42: I would make a class to store the config instead of using a callback type. It will have name with clear semantic

ktor-server/ktor-server-core/common/src/io/ktor/server/http/HttpRequestLifecycle.kt
bjhham: This was the initial idea, but iirc @zibet27 mentioned it broke quite a few tests so it appears it could be quite impactful.  There's not many scenarios for keeping the call coroutine alive after the connection is lost, but they do exist.  Maybe the best approach would be to log a ticket to enable it by default in 4.0.

zibet27:
> lgtm, I have a question, should it also work for HTTP/2 in Netty?

It should, though not tested, I'll add some tests.

ktor-server/ktor-server-core/common/src/io/ktor/server/http/HttpRequestLifecycle.kt
zibet27: @e5l I think the cancellation is needed only for computation-heavy requests because it can cancel a DB insert, which is usually not what you want, even if the client is not waiting for a response.

bjhham: It would be pretty uncommon for a client to hang up before receiving a response like this though.  Maybe if the DB is under load and the client has a short timeout though... I agree though, it's probably safer to assume that there could be sensitive work being done in the call coroutine.  The main use-case for cancellation would be long-polling connections I guess.

ktor-server/ktor-server-core/common/src/io/ktor/server/http/HttpRequestLifecycle.kt
zibet27: @e5l 
We can have the `HttpRequestLifecycleHandler` class with the `onClose` method.
I'm not sure about this. Moreover, it's internal.

ktor-server/ktor-server-cio/common/src/io/ktor/server/cio/backend/ServerRequestScope.kt
```diff
@@ -42,4 +42,6 @@ public class ServerRequestScope internal constructor(
             localAddress,
             upgraded
         )
+
+    internal var onClose: (() -> Unit)? = null
```
bjhham, line 46: Hey, could you check if using `invokeOnCompletion` for the job is viable here instead of introducing the field?

zibet27: Hey! Sorry, I forgot to mention it here. `invokeOnCompletion` doesn't help in this case.

ktor-server/ktor-server-test-suites/jvm/src/io/ktor/server/testing/suites/HttpRequestLifecycleTest.kt
```diff
@@ -0,0 +1,153 @@
...
+                }
+            } catch (err: CancellationException) {
+                @OptIn(InternalAPI::class)
+                assertTrue(err.rootCause is ConnectionClosedException)
+                requestCancelled.send(requestCancelledCnt.incrementAndFetch())
+            }
+        }
+
+        fun resetRequestOnStart(request: suspend () -> Unit) = launch {
+            client = createApacheClient()
```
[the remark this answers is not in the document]
zibet27: It's an intended behaviour because we need to change the internal client. The tests run in sequence.

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationCallHandler.kt
```diff
@@ -35,9 +37,21 @@ internal class NettyApplicationCallHandler(
...
+    internal fun onConnectionClose(context: ChannelHandlerContext) {
+        if (context.channel().isActive) {
+            return
+        }
+        currentCall?.let {
+            currentCall = null
+            @OptIn(InternalAPI::class)
+            it.attributes.getOrNull(HttpRequestCloseHandlerKey)?.invoke()
+        }
+    }
```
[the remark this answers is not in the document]
zibet27: It should still be there.

e5l:
lgtm!

PR #5183, opened 2025-11-12
Server Netty. Fix rejected execution during engine stop (#8671)

zibet27:
**Subsystem**
Server Netty

**Motivation**
[KTOR-8671](https://youtrack.jetbrains.com/issue/KTOR-8671) Netty: RejectedExecutionException during shutdown on MacOS when dev mode is enabled

**Solution**

Actually, the problem is reproducible even without dev mode. The problem is that the channels were closed after the workers:
Other small improvements that didn't cause problems yet, but can be dangerous:
- Shouldn't we cancel the call coroutine when the handler is unregistered
- Stop writing a response to an inactive channel
- Don't allow the engine to stop multiple times
- make sure all `shutdown()` methods are called and the exception is logged properly(`.await()` was ignoring any errors)

PS: 
I can't reproduce the reported problem in unit tests.
I reproduced another `RejectedExecutionException` when trying to stop the server, which is writing a response, but the error is only logged (I suppose it's done by Netty's internal logger, which I can't mock properly).
That's why the PR doesn't contain any tests, verified by local Maven publication.
Do you have advice here?

### DIFF
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/UserAgent.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/UserAgent.kt
@@ -30,8 +30,10 @@ public val UserAgent: ClientPlugin<UserAgentConfig> = createClientPlugin("UserAg
     val agent = pluginConfig.agent
 
     onRequest { request, _ ->
-        LOGGER.trace("Adding User-Agent header: agent for ${request.url}")
-        request.header(HttpHeaders.UserAgent, agent)
+        if (!request.headers.contains(HttpHeaders.UserAgent)) {
+            LOGGER.trace("Adding User-Agent header: agent for ${request.url}")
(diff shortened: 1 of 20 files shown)


ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationResponse.kt
```diff
@@ -51,7 +51,7 @@ public abstract class NettyApplicationResponse(
         // because it should've been set by commitHeaders earlier
         val chunked = headers[HttpHeaders.TransferEncoding] == "chunked"
 
-        if (responseMessageSent) return
+        if (responseMessageSent || !context.channel().isActive) return
```
bjhham: Maybe a `fun canSendMessage() = !responseMessageSent && context.channel().isActive` function could be in order here.

PR #5187, opened 2025-11-12
KTOR-8924 Curl: Client sends both Transfer-Encoding and Content-Length headers

osipxd:
**Subsystem**
ktor-client-curl

**Motivation**
[KTOR-8924](https://youtrack.jetbrains.com/issue/KTOR-8924) Curl: Client sends both Transfer-Encoding and Content-Length headers for DELETE requests with body

**Solution**
`CURLOPT_INFILESIZE_LARGE` that we used to set Content-Length for all requests is effective only for PUT requests. For POST-like requests (PATCH, DELETE) `CURLOPT_POSTFIELDSIZE_LARGE` should be used


osipxd:
@coderabbitai review

PR #5192, opened 2025-11-13
KTOR-9092 Fix an exception when NettyHttp2Handler is loaded several times

osipxd:
**Subsystem**
ktor-server-netty

**Motivation**
[KTOR-9092](https://youtrack.jetbrains.com/issue/KTOR-9092) NettyHttp2Handler throws IllegalArgumentException: 'ktor.ApplicationCall' is already in use

**Solution**
Use `AttributeKey.valueOf` that doesn't thrown an exception for existing keys instead of `newInstance`


osipxd:
I want to merge this PR today to use EAP build tomorrow in another PR. The change is very minor and shouldn't break anything, so I'll merge it with an approve from the bot.
Post-review is appreciated 🙏

PR #5193, opened 2025-11-13
KTOR-8285 Allow to configure nonCancellable refresh in Bearer auth

marychatte:
**Subsystem**
Client, Bearer auth

**Motivation**
[KTOR-8285](https://youtrack.jetbrains.com/issue/KTOR-8285) Bearer Auth: request cancellation causes refresh token invalidation
If the request is cancelled (for example, due to UI lifecycle) during the refresh, Ktor drops the new token even if the server has already accepted the refresh and invalidated the old token. As a result, the client is left with an invalidated refresh token and is required to perform a full re-authentication.

**Solution**
Add an optional `nonCancellableTokenRefresh` flag to `BearerAuthConfig` and update the token refresh logic to prevent cancellation of the originating request from discarding a successfully refreshed token.


marychatte:
I noticed that the introduction of `cacheTokens` in [PR #5115](https://github.com/ktorio/ktor/pull/5115) changed the signature of `BearerAuthProvider`’s constructor, which looks like a binary-incompatible change. Since both PRs are going into 3.4.0, I included the new parameter in this constructor as well

PR #5194, opened 2025-11-14
KTOR-3019 Improve CORS plugin logging to ease troubleshooting

Stexxe:
https://youtrack.jetbrains.com/issue/KTOR-3019

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-cors/common/src/io/ktor/server/plugins/cors/CORS.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-cors/common/src/io/ktor/server/plugins/cors/CORS.kt
@@ -87,7 +87,7 @@ internal fun PluginBuilder<CORSConfig>.buildPlugin() {
             return@onCall
         }
 
-        LOGGER.trace { "${call.request.id()}: Start handler" }
+        LOGGER.info { "${call.request.id()}: Start handler" }
 
         if (!allowsAnyHost || allowCredentials) {
             call.corsVary()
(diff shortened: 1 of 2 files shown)


ktor-server/ktor-server-plugins/ktor-server-cors/common/src/io/ktor/server/plugins/cors/CORS.kt
```diff
@@ -142,11 +153,16 @@ internal fun PluginBuilder<CORSConfig>.buildPlugin() {
...
 
         if (!call.corsCheckCurrentMethod(methods)) {
-            LOGGER.trace("Respond forbidden ${call.request.uri}: method doesn't match ${call.request.httpMethod}")
+            LOGGER.trace {
+                "${call.request.id()}: CORS check fails because HTTP method ${call.request.httpMethod.value} " +
+                    "is not allowed. Allowed methods: $methods"
+            }
             call.respondCorsFailed()
             return@onCall
         }
```
bjhham, line 162: I think for all the failure scenarios we ought to use severity INFO so it's more generally presented in the logs.

ktor-server/ktor-server-plugins/ktor-server-cors/common/src/io/ktor/server/plugins/cors/CORSUtils.kt
```diff
@@ -135,3 +186,7 @@ internal fun normalizeOrigin(origin: String): String {
 
     return builder.toString()
 }
+
+internal inline fun Logger.info(message: () -> String) {
+    info(message())
```
osipxd, line 191: Should we add `if (isInfoEnabled)` condition to save memory allocations as it was done for `trace {}`?

Stexxe: That will require publishing a new API and postponing the fix until 3.4.0. So, what's better: to leave the current logging as is and accept the allocation costs, or use the TRACE level?

bjhham: Eventually it'd be a good option, but I don't think this will have a significant impact on allocations.  We use info pretty sparingly.

ewanmellor:
This logging should have been placed at TRACE level. Users of this plugin do not need to see `Start handler` for every single call in normal operation.

PR #5196, opened 2025-11-14
KTOR-9095 Darwin: Set an actual HttpProtocolVersion

osipxd:
**Subsystem**
ktor-client-darwin

**Motivation**
[KTOR-9095](https://youtrack.jetbrains.com/issue/KTOR-9095) Darwin: HttpResponse.version always returns HTTP_1_1

**Solution**
Use `URLSession:task:didFinishCollectingMetrics:` to save transaction metrics and get protocol version from it later.

### DIFF
--- a/ktor-client/ktor-client-darwin/darwin/src/io/ktor/client/engine/darwin/internal/DarwinTaskHandler.kt
+++ b/ktor-client/ktor-client-darwin/darwin/src/io/ktor/client/engine/darwin/internal/DarwinTaskHandler.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2024 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ * Copyright 2014-2025 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
  */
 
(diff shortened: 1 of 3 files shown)


ktor-client/ktor-client-darwin/darwin/src/io/ktor/client/engine/darwin/internal/DarwinTaskHandler.kt
```diff
@@ -42,15 +48,10 @@ internal class DarwinTaskHandler(
     }.channel
 
     fun receiveData(dataTask: NSURLSessionDataTask, data: NSData) {
-        if (!response.isCompleted) {
-            val result = dataTask.response as NSHTTPURLResponse
-            response.complete(result.toResponseData(requestData))
-        }
```
osipxd, line 48: This is actually wrong. Events are called in the following order:
`:didReceiveData` → `:didFinishCollectingMetrics` → `:didCompleteWithError`

`:didFinishCollectingMetrics` is the place where we get the protocol version. So I thought we can complete response only from `:didCompleteWithError`. The problem is that this event is not called until we've completely read the body. And it is a problem for streaming responses. That's why `testDownloadStreamResponseWithCancel` hangs.

I'll try to find a different solution.

PR #5197, opened 2025-11-14
KTOR-9094 Jetty Client: Support HTTP/2 over cleartext (h2c)

osipxd:
**Subsystem**
`ktor-client-jetty-jakarta`

**Motivation**
[KTOR-9094](https://youtrack.jetbrains.com/issue/KTOR-9094) Jetty Client: Support HTTP/2 over clear-text (h2c)

**Solution**
Do not pass `sslContextFactory` for HTTP requests.
I also had to add `stream.demand()` calls. Without these calls requests would hang.


osipxd:
@coderabbitai review

PR #5198, opened 2025-11-16
[KTOR-8947] Filter pseudo-header in java client

Komdosh:
**Subsystem**
Client

**Motivation**
HTTP/2 header name can contain : symbol, but now it throws an error.

**Solution**
Filter pseudo-header in java client


osipxd (approved):
Great! Thank you for your contribution! 🎉

PR #5199, opened 2025-11-18
Tests. Downgrade  version.

zibet27:
**Subsystem**
JVM Tests

**Motivation**
JVM 8 and 11 tests are failing on the CI because of a new `mockk` version that is not compatible with Java 11, see this [issue](https://github.com/mockk/mockk/issues/1433).

**Solution**
Downgrade its version, for now.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -51,7 +51,9 @@ micrometer = "1.15.5"
 jansi = "2.4.2"
 typesafe = "1.4.5"
 
-mockk = "1.14.6"
+# Version 1.14.6 is not compatible with Java 11.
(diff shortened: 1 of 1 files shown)


zibet27:
Failing test example: https://ktor.teamcity.com/buildConfiguration/Ktor_KtorMatrixCoreLinuxJava11/381501

gradle/libs.versions.toml
```diff
@@ -51,7 +51,7 @@ micrometer = "1.15.5"
 jansi = "2.4.2"
 typesafe = "1.4.5"
 
-mockk = "1.14.6"
+mockk = "1.14.5"
```
osipxd, line 56: Could you also mention the issue here in a comment? Not to update mockk accidentially

PR #5205, opened 2025-11-19
KTOR-7864 Build libcurl on CI

osipxd:
**Subsystem**
Infrastructure, ktor-client-curl

**Motivation**
[KTOR-7864](https://youtrack.jetbrains.com/issue/KTOR-7864) Build libcurl on CI
[KTOR-8339](https://youtrack.jetbrains.com/issue/KTOR-8339) Curl: `caPath` is not set by default in the Curl client on linuxArm64
[KTOR-9100](https://youtrack.jetbrains.com/issue/KTOR-9100) Curl always uses HTTP/1.1
[KTOR-9098](https://youtrack.jetbrains.com/issue/KTOR-9098) Curl: HttpResponse.version always returns HTTP_1_1

**Solution**
Build libcurl using [vcpkg](https://vcpkg.io/en/) on CI.
Added a Gradle task `:ktor-client-curl:updateLibcurl` that runs `vcpkg install` and copies resulting binaries and headers into the interop dir. This task should be run on a target host to build binaries for it, so I created [5 build types](https://ktor.teamcity.com/project/Ktor_ProjectKtorLibcurlBuild) to build libcurl on each target host.

The most difficult part was that we should use the same toolchain that Konan uses to compile Kotlin/Native code. So:
- On Linux we should use Konan's sysroot to link agains correct version of glibc.
- On Windows we should the LLVM compiler that is used by Konan

To solve this we [download Konan Toolchain](https://ktor.teamcity.com/buildConfiguration/Ktor_DownloadKonanToolchain#all-projects) and configure vcpkg to use it.

Also I had to apply commit https://github.com/curl/curl/commit/1eca08a5417740320501448b64e8584de349126c as a patch to curl to fix [linking error on Windows](https://ktor.teamcity.com/buildConfiguration/Ktor_KtorMatrixNativeWindowsX64/385019).

**Related commits:**
- https://github.com/ktorio/ktor-build/commit/7b90ddafca6f9f925db9c684f7975e21f45062db Build libcurl on CI
- https://github.com/ktorio/ktor-build/commit/409e92cf56add1b3190c1aa4cd8cadd586691d52 Libcurl: Download toolchain before building

_It's better to review commit by commit. Commit "libcurl: Copy the original curl port" should be skipped from review as well as "Update libcurl to 8.17.0" which mostly consists of libcurl headers updates_

### DIFF
--- a/ktor-client/ktor-client-curl/desktop/test/io/ktor/client/engine/curl/test/CurlHttp2Test.kt
+++ b/ktor-client/ktor-client-curl/desktop/test/io/ktor/client/engine/curl/test/CurlHttp2Test.kt
@@ -6,7 +6,10 @@ package io.ktor.client.engine.curl.test
 
 import io.ktor.client.engine.curl.*
 import io.ktor.client.tests.*
-import kotlin.test.Ignore
 
-@Ignore // KTOR-9100 Curl always uses HTTP/1.1
-class CurlHttp2Test : Http2Test<CurlClientEngineConfig>(Curl)
+class CurlHttp2Test : Http2Test<CurlClientEngineConfig>(Curl, useH2c = false) {
+
+    override fun CurlClientEngineConfig.disableCertificateValidation() {
+        sslVerify = false
+    }
+}
--- a/ktor-client/ktor-client-curl/desktop/src/io/ktor/client/engine/curl/internal/CurlAdapters.kt
+++ b/ktor-client/ktor-client-curl/desktop/src/io/ktor/client/engine/curl/internal/CurlAdapters.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2024 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ * Copyright 2014-2025 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
  */
 
 package io.ktor.client.engine.curl.internal
@@ -130,11 +130,11 @@ internal fun HttpRequestData.headersToCurl(): CPointer<curl_slist> {
     return result!!
 }
 
-@OptIn(ExperimentalForeignApi::class)
-internal fun UInt.fromCurl(): HttpProtocolVersion = when (this) {
-    CURL_HTTP_VERSION_1_0 -> HttpProtocolVersion.HTTP_1_0
-    CURL_HTTP_VERSION_1_1 -> HttpProtocolVersion.HTTP_1_1
-    CURL_HTTP_VERSION_2_0 -> HttpProtocolVersion.HTTP_2_0
+@OptIn(ExperimentalForeignApi::class, UnsafeNumber::class)
+internal fun Long.fromCurl(): HttpProtocolVersion = when (this) {
+    CURL_HTTP_VERSION_1_0.toLong() -> HttpProtocolVersion.HTTP_1_0
+    CURL_HTTP_VERSION_1_1.toLong() -> HttpProtocolVersion.HTTP_1_1
+    CURL_HTTP_VERSION_2_0.toLong() -> HttpProtocolVersion.HTTP_2_0
     // old curl fallback
     else -> HttpProtocolVersion.HTTP_1_1
 }
--- a/ktor-client/ktor-client-curl/desktop/src/io/ktor/client/engine/curl/internal/CurlMultiApiHandler.kt
+++ b/ktor-client/ktor-client-curl/desktop/src/io/ktor/client/engine/curl/internal/CurlMultiApiHandler.kt
@@ -318,6 +318,7 @@ internal class CurlMultiApiHandler : Closeable {
 
         easyHandle.apply {
             getInfo(CURLINFO_RESPONSE_CODE, httpStatusCode.ptr)
+            getInfo(CURLINFO_HTTP_VERSION, httpProtocolVersion.ptr)
             getInfo(CURLINFO_PRIVATE, responseDataRef.ptr)
         }
 
@@ -332,7 +333,7 @@ internal class CurlMultiApiHandler : Closeable {
 
             CurlSuccess(
                 httpStatusCode.value.toInt(),
-                httpProtocolVersion.value.toUInt(),
+                httpProtocolVersion.value,
                 headers,
                 responseBody
             )
--- a/ktor-client/ktor-client-curl/desktop/src/io/ktor/client/engine/curl/internal/CurlRaw.kt
+++ b/ktor-client/ktor-client-curl/desktop/src/io/ktor/client/engine/curl/internal/CurlRaw.kt
@@ -1,6 +1,6 @@
 /*
-* Copyright 2014-2021 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
-*/
+ * Copyright 2014-2025 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
 
 package io.ktor.client.engine.curl.internal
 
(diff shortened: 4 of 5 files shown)


ktor-client/ktor-client-curl/desktop/interop/include/curl/curlver.h
```diff
@@ -32,14 +32,13 @@
 
 /* This is the version number of the libcurl package from which this header
    file origins: */
-#define LIBCURL_VERSION "8.11.1"
+#define LIBCURL_VERSION "8.17.0-DEV"
```
whyoleg, line 35: Is it fine that it's the `DEV` version?

ktor-client/ktor-client-curl/desktop/interop/libcurl.def
```diff
@@ -3,13 +3,14 @@ package = libcurl
...
+# libcurl uses native Schannel for TLS on Windows
+# There is no libz by default installed on Windows, so we need to include it statically
+staticLibraries.mingw = libcurl.a libnghttp2.a libzlib.a
 compilerOpts.mingw = -DCURL_STATICLIB
+linkerOpts.mingw = -liphlpapi -lsecur32 -lcrypt32 -lws2_32
 
-# there is no need to use openssl on macos, as curl is built using out-of-the-box security framework
-staticLibraries.osx = libcurl.a
-linkerOpts.osx = -framework Security -framework SystemConfiguration
+staticLibraries.osx = libcurl.a libnghttp2.a libssl.a libcrypto.a
```
whyoleg, line 12: Is there a reason why we can't build `curl` using the macOS Security framework instead of OpenSSL?

ktor-client/ktor-client-curl/desktop/interop/include/curl/curlver.h
osipxd: It is because vcpkg checks out curl sources by tag ([for example curl-8_17_0](https://github.com/curl/curl/blob/curl-8_17_0/include/curl/curlver.h#L35)). The stable version and timestamp are set in [maketgz](https://github.com/curl/curl/blob/curl-8_17_0/scripts/maketgz) script. Which doesn't seem to be called by vcpkg
So the version is actually stable, but `LIBCURL_VERSION` and `LIBCURL_TIMESTAMP` are not correct.

ktor-client/ktor-client-curl/desktop/interop/libcurl.def
osipxd: Secure Transport support [has been dropped](https://daniel.haxx.se/blog/2025/06/11/dropping-some-tls-laggards/) in curl 8.15.0 so we have to use OpenSSL

ktor-client/ktor-client-curl/desktop/interop/include/curl/curlver.h
osipxd: And Conan [uses sources from "Releases" tab](https://github.com/conan-io/conan-center-index/blob/master/recipes/libcurl/all/conandata.yml). So version and timestamp are correctly replaced there

osipxd: Filed an issue: https://github.com/microsoft/vcpkg/issues/48495

ktor-client/ktor-client-curl/desktop/interop/include/curl/multi.h
```diff
@@ -395,9 +395,29 @@ typedef enum {
...
+
+/* - CURLMNWC_CLEAR_CONNS tells libcurl to prevent further reuse of existing
+   connections. Connections that are idle will be closed. Ongoing transfers
+   will continue with the connection they have. */
+#define CURLMNWC_CLEAR_CONNS (1L<<0)
+
+/* - CURLMNWC_CLEAR_DNS tells libcurl to prevent further reuse of existing
+   connections. Connections that are idle will be closed. Ongoing transfers
+   will continue with the connection they have. */
+#define CURLMNWC_CLEAR_DNS (1L<<0)
```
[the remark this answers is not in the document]
osipxd: Do not review `/interop/include/curl/*` files

ktor-client/ktor-client-curl/libcurl/ports/curl/dependencies.patch
```diff
@@ -0,0 +1,169 @@
...
+   include_directories(SYSTEM ${LIBGSASL_INCLUDE_DIRS})
+@@ -1446,7 +1446,7 @@ if(CURL_USE_GSSAPI)
+ 
+   set(HAVE_GSSAPI ${GSS_FOUND})
+   if(GSS_FOUND)
+-    list(APPEND CURL_LIBS ${GSS_LIBRARIES})
++    list(APPEND CURL_LIBS ${_gss_LINK_LIBRARIES})
+     list(APPEND CURL_LIBDIRS ${GSS_LIBRARY_DIRS})
+     list(APPEND LIBCURL_PC_REQUIRES_PRIVATE ${GSS_PC_REQUIRES})
+     include_directories(SYSTEM ${GSS_INCLUDE_DIRS})
```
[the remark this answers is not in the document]
osipxd: Do not review this file. It has been copied "as is" from vcpkg ports

build-logic/src/main/kotlin/ktorbuild/vcpkg/VcpkgInstall.kt
```diff
@@ -0,0 +1,146 @@
...
+        environment("TOOLCHAIN_DIR", toolchainDirectory.absolutePath)
+        environment("TOOLCHAIN_TARGET", toolchainDirectory.name.substringBefore("-gcc"))
+    }
+}
+
+// See: https://github.com/JetBrains/kotlin/blob/v2.2.21/kotlin-native/konan/konan.properties#L68-L71
+private fun getToolchainDirName(target: String): String? = when (target) {
+    "linuxX64" -> "x86_64-unknown-linux-gnu-gcc-8.3.0-glibc-2.19-kernel-4.9-2"
+    "linuxArm64" -> "aarch64-unknown-linux-gnu-gcc-8.3.0-glibc-2.25-kernel-4.9-2"
+    "mingwX64" -> "llvm-19-x86_64-windows-essentials-134"
```
zibet27, line 133: @osipxd Shouldn't it be "mingwX64" -> "msys2-mingw-w64-x86_64-2"?

osipxd: Probably it can be the mingw directory. At some point I switched to llvm toolchain on windows to use the same compiler and linker that Konan uses to build Kotlin/Native. And then I've never tried to use the mingw toolchain again.

wkornewald:
What do you think about reusing NativeBuilds (https://github.com/ensody/native-builds) or at least forking that repo, so curl + OpenSSL + HTTP/3 could be used on all native targets and potentially even JVM and so that these libs can be updated independently of Ktor's update cycle (could be important for security)? By using OpenSSL on all platforms it would also be possible to make the TLS ciphersuite and other security settings configurable in a unified way. This could become the best solution when a project has strict security requirements about TLS and/or needs the latest crypto algorithms (e.g. ML-KEM is only supported in OpenSSL 3.5). Also, in NativeBuilds I've already prepared JVM & Android support and plan to create a sample project integrating several compression libs like zstd, gzip etc. for native and JVM targets - which could also be relevant for Ktor (at least CIO).

ktor-client/ktor-client-curl/desktop/test/io/ktor/client/engine/curl/test/CurlHttp2Test.kt
```diff
@@ -6,7 +6,10 @@ package io.ktor.client.engine.curl.test
 
 import io.ktor.client.engine.curl.*
 import io.ktor.client.tests.*
-import kotlin.test.Ignore
 
-@Ignore // KTOR-9100 Curl always uses HTTP/1.1
-class CurlHttp2Test : Http2Test<CurlClientEngineConfig>(Curl)
+class CurlHttp2Test : Http2Test<CurlClientEngineConfig>(Curl, useH2c = false) {
```
bjhham, line 10: 🎉

bjhham (approved):

zibet27 (approved):

PR #5207, opened 2025-11-19
KTOR-6837 Fix client URL handling for invalid DefaultRequest.host

marychatte:
**Subsystem**
Client, DefaultRequest, CIO, Darwin

**Motivation**
[KTOR-6837](https://youtrack.jetbrains.com/issue/KTOR-6837) 

**Solution**
`DefaultRequest.host` allows setting non-host values, which can include path, query, or fragment (e.g., "httpbin.org/status"), and this produces an invalid URL where the host field contains path segments. Most engines recover by parsing the full URL string and sending a correct request, but CIO and Darwin use the host field directly and therefore fail.
Since this behaviour could be used, the solution is to add a warning now and throw an exception for an invalid host in DefaultRequest in 4.0.0. Also, normalize URLs in CIO and Darwin by reparsing from url.toString() only when the host is invalid, aligning their behavior with that of other client engines.


marychatte:
@coderabbitai review

PR #5208, opened 2025-11-19
KTOR-9096 Fix SSE cancellation leak on Darwin

marychatte:
**Subsystem**
Client, SSE

**Motivation**
[KTOR-9096](https://youtrack.jetbrains.com/issue/KTOR-9096) 

**Solution**
Cancel `content.callContext` as well


marychatte:
@coderabbitai review

PR #5212, opened 2025-11-23
Respect preconfigured OkHttp dispatcher

### LINKED ISSUE #4908: Preconfigured OkHttp dispatcher ignored
https://github.com/ktorio/ktor/blob/main/ktor-client/ktor-client-okhttp/jvm/src/io/ktor/client/engine/okhttp/OkHttpEngine.kt#L153

This does not seem intentional. The dispatcher from the preconfigred okhttp client is ignored, and a new dispatcher is created, which is configured via engine { config { ... } }.
Is there a reason for this? Willing to contribute to fix this.

e5l:
Closes #4908

## Summary
- Ignore the original "bug" from 2020 — instead, just DON'T close OkHttp. This is the better solution, is advised by OkHttp, and solves all related problems.
- Avoid overriding the dispatcher when using a preconfigured OkHttpClient
- Add a JVM test to ensure the engine keeps the preconfigured dispatcher

## Testing
- ./gradlew :ktor-client-okhttp:jvmTest --tests io.ktor.client.engine.okhttp.OkHttpEngineTest

### DIFF
--- a/ktor-client/ktor-client-okhttp/jvm/src/io/ktor/client/engine/okhttp/OkHttpEngine.kt
+++ b/ktor-client/ktor-client-okhttp/jvm/src/io/ktor/client/engine/okhttp/OkHttpEngine.kt
@@ -147,6 +147,9 @@ public class OkHttpEngine(override val config: OkHttpConfig) : HttpClientEngineB
 
     private fun createOkHttpClient(timeoutExtension: HttpTimeoutConfig?): OkHttpClient {
         val builder = (config.preconfigured ?: okHttpClientPrototype).newBuilder()
+        if (config.preconfigured == null) {
+            builder.dispatcher(Dispatcher())
+        }
         builder.apply(config.config)
         config.proxy?.let { builder.proxy(it) }
         timeoutExtension?.let {


ktor-client/ktor-client-okhttp/jvm/src/io/ktor/client/engine/okhttp/OkHttpEngine.kt
```diff
@@ -144,7 +144,9 @@ public class OkHttpEngine(override val config: OkHttpConfig) : HttpClientEngineB
     private fun createOkHttpClient(timeoutExtension: HttpTimeoutConfig?): OkHttpClient {
         val builder = (config.preconfigured ?: okHttpClientPrototype).newBuilder()
 
-        builder.dispatcher(Dispatcher())
+        if (config.preconfigured == null) {
+            builder.dispatcher(Dispatcher())
```
osipxd, line 151: Probably we should just drop this line as `OkHttp` already [creates `Dispatcher` by default](https://github.com/square/okhttp/blob/parent-5.3.2/okhttp/src/commonJvmAndroid/kotlin/okhttp3/OkHttpClient.kt#L587):
```kotlin
internal var dispatcher: Dispatcher = Dispatcher()
```

gajicm93: > Probably we should just drop this line as `OkHttp` already [creates `Dispatcher` by default](https://github.com/square/okhttp/blob/parent-5.3.2/okhttp/src/commonJvmAndroid/kotlin/okhttp3/OkHttpClient.kt#L587):
> 
> ```kotlin
> internal var dispatcher: Dispatcher = Dispatcher()
> ```

Yes, because even when copying okHttpClientPrototype, Dispatcher should be shared, as stated in the [docs](https://square.github.io/okhttp/5.x/okhttp/okhttp3/-ok-http-client/index.html):

> OkHttp performs best when you create a single OkHttpClient instance and reuse it for all of your HTTP calls. This is because each client holds its own connection pool and thread pools. Reusing connections and threads reduces latency and saves memory. Conversely, creating a client for each request wastes resources on idle pools.

The above should apply to prototype client also, new Dispatcher should NEVER be set.

gajicm93:
It's important to understand that "newBuilder()" method creates new OkHttpClient instance, but reuses original Dispatcher and ConnectionPool. (which is expected and recommended in the docs).

That is why it's important to remove setting new Dispatcher completely, even in the "prototype" instance case, to achieve that all Ktor OkHttpEngine instances reuse the same Dispatcher and ConnectionPool.

EDIT: If somebody doesn't want to share Dispatcher and ConnectionPool instances across OkHttpEngine instances, that can be achieved by explicitly setting DIFFERENT PRECONFIGURED INSTANCES when creating Engine instances, but that should not be the default behaviour.

TLDR: You should just make a "sane default", but leave all the options open to the developer, not force any of your decisions on us.

osipxd (approved):
👍

e5l:
We can't drop the setting `Dispatcher()`, because all OkHttp clients will share the same dispatcher. 

We can fall back to the default OkHttp behaviour, but it will lead to dispatcher hanging in the background even after `close()` call

osipxd:
If we use the shared dispatcher, we shouldn't call shutdown on it. Are there any downsides if we don't shutdown the dispatcher?

e5l:
it will be active even if ktor client was used for a single request and closed

osipxd:
> it will be active even if ktor client was used for a single request and closed

This matches OkHttp’s default behavior. The [official docs](https://square.github.io/okhttp/5.x/okhttp/okhttp3/-ok-http-client/index.html) say that “shutdown isn’t necessary” and that idle resources are released automatically.   Probably we also shouldn't shut it down by default.

gajicm93:
Or you can make "okHttpClientPrototype" be NOT static, and bound to Engine instance instead.

That way when you close Engine, only it's own resources get cleaned.

The whole problem actually stems from "prototype" being static, because then multiple Engines share the same OkHttp resources.

gajicm93:
"clientCache" is not static, so prototype should not be either.

And you still keep the behaviour that all cached clients share resource, but they're not shared with OTHER ENGINE INSTANCES.

gajicm93:
@osipxd @e5l If you look at history, this whole mess started because someone complained OkHttpEngine leaks resources: https://github.com/ktorio/ktor/issues/1340

Because of that, you started shutting down Dispatcher (which official OkHttp docs advise against), and that caused the following problem:

- Because "prototype" is static, closing one Engine closed resources of all others (including preconfigured OkHttp instance in case it was used).

To fix that, you implemented setting new Dispatcher every time, but that created new problem because now preconfigured Dispatcher was ignored (+ it's a resource hog to make every new client use new Dispatcher, Dispatchers SHOULD be shared).

So now you have 2 options:

1. Make "prototype" object bound (not static), and guard against closing preconfigured instances. This solves both issues of one prototype Engine closing resources of other prototype Engine, and also guards against preconfigured Engine closing resources it doesn't own.

2. Ignore original "bug" from 2020, and just DON'T close OkHttp. This is better solution and is advised by OkHttp and it solves all problems.

gajicm93:
> it will be active even if ktor client was used for a single request and closed

That is ok for preconfigured instance, because it is usually shared across the whole app.

E.g, I use single OkHttp instance for:

1. Coil
2. ConfigCat
3. ExoPlayer
4. KtorEngine (preconfigured)
5. ...

So if KtorEngine closes my Dispatcher, then even my Coil and ExoPlayer stop working.

e5l:
Logged here https://youtrack.jetbrains.com/issue/KTOR-9391/OkHttpEngine-Do-not-close-engine-on-the-client-close, will try to check

btw we're accepting PRs 🙂

PR #5214, opened 2025-11-24
Use Apache5 for server tests

zibet27:
**Subsystem**
Server JVM

**Motivation**
I've noticed that we use Apache for HTTP/2 tests, which doesn't support it. When testing proper HTTP/2 requests, I've also found [KTOR-9113](https://youtrack.jetbrains.com/issue/KTOR-9113) Netty HTTP2 server hangs on the plugin exception.

**Solution**
- Change `Apache` -> `Apache5`
- HTTP/2 and higher don't require draining the input to reuse the persistent connection, which caused the Netty hang

### DIFF
--- a/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationEngine.kt
+++ b/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationEngine.kt
@@ -280,29 +280,37 @@ public class NettyApplicationEngine(
         workerEventGroup.shutdownGracefully().sync()
     }
 
+    private inline fun <R> withStopException(crossinline block: () -> R) {
+        runCatching(block).onFailure {
+            environment.log.error("Exception thrown during engine stop", it)
+        }
+    }
+
     override fun stop(gracePeriodMillis: Long, timeoutMillis: Long) {
         cancellationJob?.complete()
         monitor.raise(ApplicationStopPreparing, environment)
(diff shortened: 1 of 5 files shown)


bjhham:
Kinda looks like Jetty just straight up doesn't work with HTTP/2 😅

zibet27:
> Kinda looks like Jetty just straight up doesn't work with HTTP/2 😅

That's actually a problem with Jetty based on Test Servlet, and the real Jetty Engine works fine.

ktor-server/ktor-server-test-base/common/src/io/ktor/server/test/base/EngineTestBase.kt
```diff
@@ -23,6 +23,10 @@ expect abstract class EngineTestBase<TEngine : ApplicationEngine, TConfiguration
     @Retention
     protected annotation class Http2Only()
 
+    @Target(AnnotationTarget.FUNCTION)
+    @Retention
+    protected annotation class NoHttp2()
```
bjhham: Maybe we can future-proof this on HTTP/3 here by calling it `Http1Only`.

zibet27: Yeah, `Http1Only` sounds better. `NoHttp2` was there in JVM before, so I moved it to common.

PS: trying to fix JettyHttp2AsyncServlet now.

PR #5220, opened 2025-11-26
KTOR-9108 Remove `Connection: keep-alive` for HTTP/2 in sse

marychatte:
**Subsystem**
Server, SSE

**Motivation**
[KTOR-9108](https://youtrack.jetbrains.com/issue/KTOR-9108) SSE: The handler adds `Connection: Keep-Alive` header, which is incompatible with HTTP/2

### DIFF
--- a/ktor-http/common/src/io/ktor/http/ContentTypes.kt
+++ b/ktor-http/common/src/io/ktor/http/ContentTypes.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2024 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ * Copyright 2014-2025 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
  */
 
 package io.ktor.http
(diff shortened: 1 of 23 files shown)


ktor-server/ktor-server-plugins/ktor-server-sse/common/src/io/ktor/server/sse/Routing.kt
```diff
@@ -178,7 +179,11 @@ private fun Route.processSSE(
     handle {
         call.response.header(HttpHeaders.ContentType, ContentType.Text.EventStream.toString())
         call.response.header(HttpHeaders.CacheControl, "no-store")
-        call.response.header(HttpHeaders.Connection, "keep-alive")
+        if (call.request.httpVersion == HttpProtocolVersion.HTTP_1_1.toString() ||
+            call.request.httpVersion == HttpProtocolVersion.HTTP_1_0.toString()
```
osipxd: Maybe check by prefix? `.startsWith("HTTP/1.")`

marychatte:
@osipxd I added a test to check the header `Connection`

PR #5228, opened 2025-12-01
KTOR-9146 Run HttpStatement.execute on the engine's dispatcher

osipxd:
**Subsystem**
Client

**Motivation**
[KTOR-9146](https://youtrack.jetbrains.com/issue/KTOR-9146) Run HttpStatement.execute on the engine's dispatcher

**Solution**
Wrap `block` with `withContext(response.coroutineContext)`

### DIFF
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/statement/HttpStatement.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/statement/HttpStatement.kt
@@ -11,6 +11,8 @@ import io.ktor.client.request.*
 import io.ktor.utils.io.*
 import kotlinx.coroutines.CompletableJob
 import kotlinx.coroutines.job
+import kotlinx.coroutines.withContext
+import kotlin.coroutines.ContinuationInterceptor
 
 /**
  * Represents a prepared HTTP request statement for [HttpClient].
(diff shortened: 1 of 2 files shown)


ktor-client/ktor-client-core/common/src/io/ktor/client/statement/HttpStatement.kt
```diff
@@ -37,25 +39,29 @@ public class HttpStatement(
...
      * @param block A suspend function that receives the [HttpResponse] for streaming.
-     * @return The result of executing [block] with the streaming [response].
+     * @return The result of executing [block] with the streaming response.
      */
     public suspend fun <T> execute(block: suspend (response: HttpResponse) -> T): T = unwrapRequestTimeoutException {
         val response = fetchStreamingResponse()
 
         try {
-            return block(response)
+            return withContext(response.coroutineContext[ContinuationInterceptor]!!) {
```
osipxd, line 62: Initially I wanted to use `response.coroutineContext` here, but it turned out to be a bad idea. For example, if a user cancels the response from within `execute`, the `CancellationException` will be thrown from `.execute {}` call.

So I take only the dispatcher. Alternatively we could take the dispatcher from `response.call.client.engine.dispatcher`, but it feels a bit long 😅

bjhham (approved):
Looks good 👍 

I was hoping this might fix a few flaky tests, but it does not seem to be the case 😓

PR #5229, opened 2025-12-01
Update commons-beanutils to 1.11.0 - resolve CVE-2025-48734

stecurran-est:
**Subsystem**

**Motivation**
Address vulnerability CVE-2025-48734.

**Solution**
Force use of commons-beanutils:1.11.0 across subprojects

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -14,15 +14,6 @@ plugins {
 logger.lifecycle("Build version: ${project.version}")
 logger.lifecycle("Kotlin version: ${libs.versions.kotlin.get()}")
 
-subprojects {
-
-    configurations.all {
-        resolutionStrategy {
-            force("commons-beanutils:commons-beanutils:1.11.0")
-        }
-    }
-}
-
 wirePackageJsonAggregationTasks()
 
 configureYarn()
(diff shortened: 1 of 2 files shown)


stecurran-est:
@bjhham Is this suitable for review for suitability to merge? Or do all Build tests need to pass? On Team City link I can see there are some failures on individual test cases listed for Build tests

build.gradle.kts
```diff
@@ -10,9 +10,19 @@ plugins {
...
 logger.lifecycle("Build version: ${project.version}")
 logger.lifecycle("Kotlin version: ${libs.versions.kotlin.get()}")
 
+subprojects {
+
+    configurations.all {
+        resolutionStrategy {
+            force("commons-beanutils:commons-beanutils:1.11.0")
+        }
+    }
```
osipxd: This adds the constraint to all pom files of all modules. Even the ones not depending on the `commons-beanutils`. I think we should instead find which dependency adds the `commons-beanutils` and either update it, or add the constraint to that specific module.

stecurran-est:
Hi @osipxd, thank you for your comment. I’ve updated the changes based on your suggestion.
commons-beanutils is pulled in transitively by ktor-server-velocity.
I’ve removed the force configuration from the root build.gradle.kts and moved it to ktor-server-velocity/build.gradle.kts.

build.gradle.kts
stecurran-est: Hi @osipxd, thank you for your comment. I’ve updated the changes based on your suggestion.
commons-beanutils is pulled in transitively by ktor-server-velocity.
I’ve removed the force configuration from the root build.gradle.kts and moved it to ktor-server-velocity/build.gradle.kts.

osipxd:
Will be fixed in #5405

PR #5230, opened 2025-12-02
Netty Server. Fix race condition during engine stop.

zibet27:
**Subsystem**
Netty Server.

**Motivation**
[KTOR-8671](https://youtrack.jetbrains.com/issue/KTOR-8671) Netty: RejectedExecutionException during shutdown on MacOS when dev mode is enabled
It was supposed to be fixed in https://github.com/ktorio/ktor/pull/5183, but my test configuration was not correct.
So, it would help if someone also tests the solution on this reproducer https://github.com/dadadom/ktor-8671-reproducer.
As I already mentioned, I didn't come up with a proper unit test reproducer.

**Solution**
Wait for connections and workers shutdown, then start the call event group shutdown.
Also, connectionPoolQuietPeriod can be 0, so we don't accept more incoming connections.


zibet27:
@marychatte, can you please check the updated implementation with your reproducer? This one spends less time doing nothing and waiting (and passes `testShutdownGracePeriod`), though it shouldn't have any race conditions.

marychatte:
@zibet27 checked, it's working

PR #5231, opened 2025-12-02
Fix crash when URLDecodeException is thrown from parseUrl

edenman:
URLBuilder.takeFrom has a catch (Throwable) but the build() method does a parameters.build() which can throw a URLDecodeException (and that is not a URLParserException so the parseUrl catch block wouldn't catch it

**Subsystem**
Common

**Motivation**
parseUrl should not crash my app

**Solution**
Catch the exception


edenman:
@bjhham thanks for the approval! The test failures don't seem related, how should I proceed?

bjhham:
Don't worry about the tests, we've got some flakiness problems atm.  I'll just rerun them until they pass then merge 👍

PR #5233, opened 2025-12-03
Server. Update KDocs for `Application.respond...`

zibet27:
**Subsystem**
Server

**Motivation**
I was confused that calling `call.respondText()` doesn't set the status code to 200 if it was assigned before.

**Solution**
Mentioned this behaviour in KDocs.

### DIFF
--- a/ktor-server/ktor-server-core/common/src/io/ktor/server/response/ApplicationResponseFunctions.kt
+++ b/ktor-server/ktor-server-core/common/src/io/ktor/server/response/ApplicationResponseFunctions.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2024 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ * Copyright 2014-2025 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
  */
 
 @file:Suppress("unused")
@@ -13,7 +13,6 @@ import io.ktor.server.util.*
 import io.ktor.util.reflect.*
 import io.ktor.utils.io.*
 import io.ktor.utils.io.charsets.*
-import io.ktor.utils.io.core.*
 import kotlinx.io.*
 import kotlin.jvm.*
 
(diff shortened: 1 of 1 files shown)


bjhham (approved):

ktor-server/ktor-server-core/common/src/io/ktor/server/response/ApplicationResponseFunctions.kt
```diff
@@ -118,7 +118,7 @@ public suspend inline fun ApplicationCall.respondRedirect(permanent: Boolean = f
  *
  * @see [io.ktor.server.response.ApplicationResponse]
  * @param contentType is an optional [ContentType], default is [ContentType.Text.Plain]
- * @param status is an optional [HttpStatusCode], default is [HttpStatusCode.OK]
+ * @param status is an optional [HttpStatusCode], default is [HttpStatusCode.OK] unless already assigned
```
vnikolova: For consistency with other KDocs, I suggest to omit the verb (is) in this and the next occurrences:

```suggestion
 * @param status An optional [HttpStatusCode], defaults to [HttpStatusCode.OK] unless already assigned
```

ktor-server/ktor-server-core/common/src/io/ktor/server/response/ApplicationResponseFunctions.kt
```diff
@@ -214,6 +214,11 @@ public suspend fun ApplicationCall.respondSource(
  * The provided [ByteWriteChannel] will be closed automatically.
  *
  * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.server.response.respondBytesWriter)
+ *
+ * @param contentType An optional [ContentType], unspecified by default
```
vnikolova, line 221: Isn't the default `contentType ` actually `ContentType.Application.OctetStream`?

zibet27: @vnikolova afaik it is used as a default in the Client's `defaultTransformers`. But on the server, it's chosen depending on the data sent, isn't it? @bjhham

bjhham: I think it will automatically populate it with octet stream down the line.  Better try it out to see though.  Sorry, missed this comment last month 😓

PR #5234, opened 2025-12-03
KTOR-5391: Document ECDSA certificate limitation for client authentication

retoo:
**Subsystem**
ktor-network-tls, ktor-client-cio

**Motivation**
Currently, the Ktor CIO engine silently ignores client certificates that use the ECDSA 

This limitation is not immediately obvious.

See also KTOR-5391.

**Solution**
Added a **Note** to the KDoc to the affected methods


Mr3zee (approved):

Mr3zee:
Thanks for the contribution! Sorry, we just found it now 😅

PR #5235, opened 2025-12-05
KTOR-9086 Default content types for OpenAPI spec

bjhham:
**Subsystem**
Server, OpenAPI

**Motivation**
[KTOR-9086](https://youtrack.jetbrains.com/issue/KTOR-9086) Read OpenAPI default content type information from ContentNegotiation plugin

**Solution**
Introduced an attribute to define the default content types that will be used for cases where it is unspecified, like `call.respond(T)`.  This is assigned upon installing `ContentNegotiation` and read later when generating the OpenAPI specification.

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-content-negotiation/common/src/io/ktor/server/plugins/contentnegotiation/ContentNegotiation.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-content-negotiation/common/src/io/ktor/server/plugins/contentnegotiation/ContentNegotiation.kt
@@ -61,6 +61,9 @@ public val ContentNegotiation: RouteScopedPlugin<ContentNegotiationConfig> = cre
 ) {
     convertRequestBody()
     convertResponseBody()
+
+    // register default content types for application metadata, used in OpenAPI
+    application.attributes[DefaultContentTypesAttribute] = pluginConfig.registrations.map { it.contentType }.distinct()
 }
 
 /**
--- a/ktor-server/ktor-server-plugins/ktor-server-routing-annotate/common/src/io/ktor/annotate/RouteAnnotationApi.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-routing-annotate/common/src/io/ktor/annotate/RouteAnnotationApi.kt
@@ -9,6 +9,7 @@ package io.ktor.annotate
 import io.ktor.http.*
 import io.ktor.openapi.*
 import io.ktor.openapi.ReferenceOr.Companion.value
+import io.ktor.server.http.content.DefaultContentTypesAttribute
 import io.ktor.server.routing.*
 import io.ktor.util.*
 import io.ktor.utils.io.*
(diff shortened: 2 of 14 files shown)


ktor-server/ktor-server-plugins/ktor-server-content-negotiation/common/src/io/ktor/server/plugins/contentnegotiation/ContentNegotiation.kt
```diff
@@ -61,6 +61,9 @@ public val ContentNegotiation: RouteScopedPlugin<ContentNegotiationConfig> = cre
 ) {
     convertRequestBody()
     convertResponseBody()
+
+    // register default content types for application metadata, used in OpenAPI
+    application.attributes[DefaultContentTypesAttribute] = pluginConfig.registrations.map { it.contentType }
```
marychatte: Maybe use `.distinct()` for some cases like this:
```kotlin
install(ContentNegotiation) { r
    register(ContentType.Application.Json, LegacyJsonConverter()) 
    register(ContentType.Application.Json, FasterJsonConverter()) 
}
```

ktor-server/ktor-server-plugins/ktor-server-routing-annotate/common/src/io/ktor/annotate/RouteAnnotationApi.kt
```diff
@@ -111,22 +118,27 @@ private fun RoutingNode.method(): HttpMethod? =
         ?.method
 
 private fun RoutingNode.operation(): Operation? {
-    // TODO KTOR-9086 get schema inference from ContentNegotiation plugin
-    val schemaInference = KotlinxJsonSchemaInference
+    val schemaInference = application.attributes.getOrNull(JsonSchemaAttributeKey)
+        ?: KotlinxJsonSchemaInference
+    val defaultContentTypes = application.attributes.getOrNull(DefaultContentTypesAttribute)
+        ?: listOf(ContentType.Application.Json)
```
marychatte, line 124: Why is `ContentType.Application.Json` the default one?

ktor-server/ktor-server-plugins/ktor-server-content-negotiation/common/src/io/ktor/server/plugins/contentnegotiation/ContentNegotiation.kt
bjhham: Ah, good call 👍

ktor-server/ktor-server-plugins/ktor-server-routing-annotate/common/src/io/ktor/annotate/RouteAnnotationApi.kt
bjhham: My thinking was that it's probably best to have _some_ default value, so users can see the responses appear in the result and make corrections accordingly.  I considered using some placeholder value so it'd be more obvious when it's not set, but decided to go with the most likely default so, in case of some missing logic, people can still stumble into the correct solution.  The only times when you'll hit the default is if you use `call.respond(T)` without installing any `ContentNegotiation` handlers, so either they've got a custom replacement or it's misconfigured and they'll hit runtime errors when calling the endpoints.

bjhham:
@zibet27 I'll include some additional content type tests in https://github.com/ktorio/ktor/pull/5245 since I'm including YAML support there anyway.

PR #5236, opened 2025-12-05
KTOR-9171 Redesign ByteReadChannel.readUTF8Line API

osipxd:
**Subsystem**
`ktor-io`

**Motivation**
[KTOR-9171](https://youtrack.jetbrains.com/issue/KTOR-9171) Redesign ByteReadChannel.readUTF8Line API

Fixes:
- [KTOR-9137](https://youtrack.jetbrains.com/issue/KTOR-9137) ByteReadChannel.readUTF8Line is inefficient for long lines
- [KTOR-4219](https://youtrack.jetbrains.com/issue/KTOR-4219) Make readUTF8LineTo return number of read symbols instead of boolean
- [KTOR-8945](https://youtrack.jetbrains.com/issue/KTOR-8945) ByteReadChannel.readUTF8Line doesn't throw TooLongLineException when the limit is reached

**Solution**
- Replaced byte-by-byte scanning with `indexOf`
- Optimized case when CR is not allowed (LF+CRLF mode)
- Added `readBuffer` caching to reduce property access overhead
- Performance scales with line length (larger lines see bigger gains)

| Scenario | 3.3.3 | 3.4.0 (Lenient) | 3.4.0 (Default) | Speedup (best) |
|----------|-------|-------------|-----------------|----------------|
| 200k × 10B lines | 21.74 ms | 10.71 ms | 9.17 ms | **2.4×** |
| 1k × 1MB lines | 3678.26 ms | 807.09 ms | 625.51 ms | **5.9×** |

Benchmark: https://github.com/ktorio/ktor-benchmarks/pull/87

I want to discuss if we should use LF + CRLF line ending mode by default. I think this mode should cover 99% of cases and it comes with performance boost. When we don't look for CR we can use only one `indexOf` so implementation becomes faster.

Currently `readUTF8Line()` and `readUTF8LineTo()` use `LineEndingMode.Any` by default and the variant of  `readUTF8LineTo` taking `lineEnding` as a parameter is marked as `@InternalAPI`.

### DIFF
--- a/ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
+++ b/ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2025 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ * Copyright 2014-2026 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
  */
 
 package io.ktor.utils.io
@@ -421,21 +421,21 @@ private const val LF: Byte = '\n'.code.toByte()
  * Reads a line of UTF-8 characters from the `ByteReadChannel`.
  * It recognizes CR, LF, and CRLF as a line delimiter.
  *
- * ## Deprecation Notes
+ * ## Deprecation
  *
- * This function is deprecated in favor of [readLineStrict] and [readLine].
+ * This function is deprecated. Use [readLineStrict] or [readLine] instead.
  *
  * Use [readLineStrict] when:
- * - Need to limit line length and throw [TooLongLineException] if the limit is reached
- * - Expect an explicit line break at the end of a line. [EOFException] will be thrown
- *   if the channel is closed before the line break
+ * - You need to limit the line length and throw [TooLongLineException] when the limit is reached.
+ * - You expect an explicit line break at the end of a line. [EOFException] will be thrown
+ *   when the channel is closed before the line break.
  *
  * Use [readLine] when:
- * - No need to limit line length
- * - An explicit line break at the end of the channel is not mandatory
+ * - You do not need to limit line length.
+ * - An explicit line break at the end of the channel is not mandatory.
  *
- * Other changes:
- * - New functions recognize LF and CRLF as a line delimiter by default. This default comes with better performance.
+ * Compared to `readUTF8Line`, the new functions differ in the following ways:
+ * - They recognize LF and CRLF as line delimiters by default, which provides better performance.
  *   To keep current behavior and recognize all line delimiters (CR, LF, and CRLF), specify [LineEnding.Lenient].
  * - [readLineStrict] accepts [Long] instead of [Int] as a limit parameter.
  *
@@ -453,8 +453,8 @@ private const val LF: Byte = '\n'.code.toByte()
  * }
  * val line = if (success) buffer.toString() else null
  * ```
- * However, we recommend to use [LineEnding.Default] if possible and verify if the case with
- * the unexpected end of line should actually be ignored. We expect the following code to be correct in most cases:
+ * However, we recommend using the default line ending behavior ([LineEnding.Default]) and
+ * do not ignore unexpected end-of-input:
  * ```
  * val line = channel.readLineStrict(limit = 1024)
  * ```
@@ -466,7 +466,7 @@ private const val LF: Byte = '\n'.code.toByte()
  * @throws TooLongLineException if max is reached before encountering a line delimiter or end of input
  */
 @Suppress("DEPRECATION")
-@Deprecated("Use readLineStrict instead. See deprecation notes for more details.")
+@Deprecated("Use readLineStrict or readLine instead. See deprecation notes for more details.")
 public suspend fun ByteReadChannel.readUTF8Line(max: Int = Int.MAX_VALUE): String? {
     val result = StringBuilder()
     val completed = readUTF8LineTo(result, max)
@@ -477,23 +477,24 @@ public suspend fun ByteReadChannel.readUTF8Line(max: Int = Int.MAX_VALUE): Strin
  * Reads a line of UTF-8 characters to the specified [out] buffer.
  * It recognizes CR, LF, and CRLF as a line delimiter.
  *
- * ## Deprecation Notes
+ * ## Deprecation
  *
- * This function is deprecated in favor of [readLineStrictTo] and [readLineTo].
+ * This function is deprecated. Use [readLineStrictTo] or [readLineTo] instead.
  *
  * Use [readLineStrictTo] when:
- * - Need to limit line length and throw [TooLongLineException] if the limit is reached
- * - Expect an explicit line break at the end of a line. [EOFException] will be thrown
- *   if the channel is closed before the line break
+ * - You need to limit the line length and throw [TooLongLineException] when the limit is reached.
+ * - You expect an explicit line break at the end of a line. [EOFException] will be thrown
+ *   when the channel is closed before the line break.
  *
  * Use [readLineTo] when:
- * - No need to limit line length
- * - An explicit line break at the end of the channel is not mandatory
+ * - You do not need to limit line length.
+ * - An explicit line break at the end of the channel is not mandatory.
  *
- * Other changes:
- * - New functions recognize LF and CRLF as a line delimiter by default. This default comes with better performance.
+ * Compared to `readUTF8LineTo`, the new functions differ in the following ways:
+ * - They recognize LF and CRLF as line delimiters by default, which provides better performance.
  *   To keep current behavior and recognize all line delimiters (CR, LF, and CRLF), specify [LineEnding.Lenient].
- * - New functions return number of appended characters instead of [Boolean] or `-1` if the channel is empty.
+ * - They return the return number of appended characters instead, or `-1` if the channel is empty, instead of
+ *   returning [Boolean].
  * - [readLineStrictTo] accepts [Long] instead of [Int] as a limit parameter.
  *
  * The direct equivalent of `readUTF8LineTo` would be:
@@ -508,8 +509,8 @@ public suspend fun ByteReadChannel.readUTF8Line(max: Int = Int.MAX_VALUE): Strin
  *     true
  * }
  * ```
- * However, we recommend to use [LineEnding.Default] if possible and verify if the case with
- * the unexpected end of line should actually be ignored. We expect the following code to be correct in most cases:
+ * However, we recommend using the default line ending behavior ([LineEnding.Default]) and
+ * do not ignore unexpected end-of-input:
  * ```
  * val success = channel.readLineTo(out, limit = 1024) >= 0
  * ```
@@ -524,7 +525,7 @@ public suspend fun ByteReadChannel.readUTF8Line(max: Int = Int.MAX_VALUE): Strin
  */
 @Suppress("DEPRECATION")
 @OptIn(InternalAPI::class)
-@Deprecated("Use readLineStrictTo instead. See deprecation notes for more details.")
+@Deprecated("Use readLineStrictTo or readLineTo instead. See deprecation notes for more details.")
 public suspend fun ByteReadChannel.readUTF8LineTo(out: Appendable, max: Int = Int.MAX_VALUE): Boolean {
     return readUTF8LineTo(out, max, lineEnding = LineEndingMode.Any)
 }


ktor-io/common/test/ReadUtf8LineTest.kt
```diff
@@ -57,4 +80,154 @@ class ReadUtf8LineTest {
...
+        while (channel.readUTF8LineTo(out) && count < numberOfLines) {
+            count++
+        }
+
+        assertEquals(numberOfLines, count)
+        assertEquals(2_088_890, out.length)
+    }
+
+    @Test
+    fun `test reading long lines completes in a reasonable time`() = runTest(timeout = 5.seconds) {
```
osipxd: This test times out. It seems too low timeout for CI agent.
Probably we should remove these two tests and just run io-benchmarks 🤔

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
```diff
@@ -473,54 +473,96 @@ public suspend fun ByteReadChannel.readUTF8LineTo(
...
+    // We should subtract 1 from lfIndex to ignore the case when CR is part of the CRLF sequence
+    val crEndIndexCorrection = if (allowsCrlf) 1L else 0L
+
+    while (!isClosedForRead) {
+        val lfIndex = if (lineEnding != LineEndingMode.CR) readBuffer.indexOf(LF) else -1
+        val crIndex = if (allowsCR) {
+            // Index of lone CR delimiter.
+            // We're interested in it only if the requested line ending mode allows CR as a line delimiter.
+            val crEndIndex = if (lfIndex >= 0L) (lfIndex - crEndIndexCorrection).coerceAtLeast(0) else Long.MAX_VALUE
+            readBuffer.indexOf(CR, endIndex = crEndIndex)
```
osipxd: Handling different line ending modes turned out to be quite annoying. I hope the code is clear with comments. Please point out any unclear moments.

ktor-io/common/test/ReadUtf8LineTest.kt
bjhham: Yeah, the CI has proven itself to be unreasonable 😅

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
bjhham: Kinda looks like it may be a bit cleaner if we just had `when(lineEnding) { ... }` with separate functions for each strategy, then we can use the direct calls internally for an optimisation where relevant.

osipxd: Do you mean create a `when` with 7 branches? (`CR + LF + CRLF`, `CR + LF`, `CR + CRLF`, `LF + CRLF`, `CR`, `LF`, `CRLF`). It's sad that I've introduced such a flexible parameter, I think some of these combinations don't make any sense 😅

bjhham: Oh yeah that would be a bit unwieldy.  I thought it might be like 4 branches 😅 

Maybe it would make sense to just have a function for CRLF + LF, then keep this more generous implementation for all other branches - then we call the CRLF + LF one directly for the better performance.

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
```diff
@@ -466,61 +545,231 @@ public suspend fun ByteReadChannel.readUTF8LineTo(out: Appendable, max: Int = In
...
-            throw IOException("Unexpected line ending $lineEndingToCheck, while expected $lineEnding")
+    var consumed = 0L
+
+    fun transferString(count: Long) {
+        if (count > 0L) {
+            val string = readBuffer.readString(count)
+            out.append(string)
+            consumed += string.length
         }
     }
```
bjhham, line 687: We could avoid one extra copy here if we introduce a `Source.readTo(Appendable)` function in `kotlinx-io`.  Maybe we can make a contribution for the next release.  If we wanted to do this without changing `kotlinx-io` it seems that would require a fair bit of duplication because most of their functions are internal in this area.

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
```diff
@@ -466,61 +545,231 @@ public suspend fun ByteReadChannel.readUTF8LineTo(out: Appendable, max: Int = In
...
-                throw TooLongLineException("Line exceeds limit of $max characters")
+            val crIndex = readBuffer.indexOf(CR, endIndex = crEndIndex)
+
+            // Sole CR in the buffer
+            if (crIndex >= 0) {
+                transferString(count = crIndex)
+                readBuffer.discard(1)
+                return consumed
             }
+        }
```
bjhham, line 729: This would be performing two scans of the buffer here...  Maybe it would be better to implement a custom `Source.indexOf()` for the two different bytes for the CR / LF case.  I think it would be cleaner to just have two separate functions instead of including the CR only case inside this one.  There'll be a bit of duplication, but it should be easier to follow.

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
```diff
@@ -466,61 +545,231 @@ public suspend fun ByteReadChannel.readUTF8LineTo(out: Appendable, max: Int = In
...
+    return consumed
+}
+
+private fun throwTooLongLineException(limit: Long) {
+    throw TooLongLineException("Line exceeds limit of $limit characters")
+}
+
+private fun throwEndOfStreamException(consumed: Long) {
+    throw EOFException("Unexpected end of stream after reading $consumed characters")
 }
```
bjhham: Looks like these ought to return Nothing.

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
osipxd: Yes, also thought about it. My plan was to implement `readStringTo` inside Ktor and measure it's performance compared to `readString`. Then go to kotlinx-io guys with this data

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
osipxd: Yes, initially I made two different functions, but it was hard to fix bugs in two places 😅 And in this case we would need to duplicate all tests for both line ending modes to verify these functions actually behave similarly.
So I thought that if this extra logic for CR handling was guarded with a simple `if` it would be easy to follow the logic ignoring these 'if's for no-CR case.
Do you think we should split these functions?

osipxd: > Maybe it would be better to implement a custom Source.indexOf() for the two different bytes for the CR / LF case

Good idea! I'll try to do it in the scope of [KTOR-9178](https://youtrack.jetbrains.com/issue/KTOR-9178)

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
```diff
@@ -466,61 +545,231 @@ public suspend fun ByteReadChannel.readUTF8LineTo(out: Appendable, max: Int = In
...
+        }
+    }
 
-        return (lineBuffer.size > 0).also { remaining ->
-            if (remaining) {
-                out.append(lineBuffer.readString())
+    if (consumed == 0L && isClosedForRead) return -1
+
+    // Defensive check. Normally the consumed count should never exceed the limit
+    if (consumed > limit) throwTooLongLineException(limit)
```
osipxd: **Question for review:** Maybe `IllegalStateException` should be thrown here, I'm not sure

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
```diff
@@ -466,61 +545,231 @@ public suspend fun ByteReadChannel.readUTF8LineTo(out: Appendable, max: Int = In
...
+ * @param out the buffer to append line to.
+ * @param lineEnding line ending mode. Accepts LF and CRLF by default.
+ *
+ * @return number of characters appended to [out], or `-1` if can't read from the channel
+ */
+@OptIn(InternalAPI::class)
+public suspend fun ByteReadChannel.readLineTo(
+    out: Appendable,
+    lineEnding: LineEnding = LineEnding.Default
+): Long {
```
osipxd, line 600: I've removed `limit` from non-strict `readLineTo` as it was not obvious what to do when we reach the limit. We shouldn't throw, but returning a truncated string doesn't make sense. It also aligns with kotlinx-io.

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
```diff
@@ -466,61 +545,231 @@ public suspend fun ByteReadChannel.readUTF8LineTo(out: Appendable, max: Int = In
...
+ * @return the line, or `null` if can't read from the channel
+ * @throws TooLongLineException if the line exceeds [limit]
+ * @throws EOFException if the channel is closed before a line break is found
+ */
+public suspend fun ByteReadChannel.readLineStrict(
+    limit: Long = Long.MAX_VALUE,
+    lineEnding: LineEnding = LineEnding.Default,
+): String? {
+    val result = StringBuilder()
+    return if (readLineStrictTo(result, limit, lineEnding) >= 0) result.toString() else null
```
osipxd, line 630: **Question for review:**
Maybe we should make it stricter and throw an exception here instead of returning `null`?
```kotlin
return if (readLineStrictTo(result, limit, lineEnding) >= 0) result.toString() else throwEndOfStreamException()
```

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
```diff
@@ -466,61 +545,231 @@ public suspend fun ByteReadChannel.readUTF8LineTo(out: Appendable, max: Int = In
...
+ * @param lineEnding line ending mode. Accepts LF and CRLF by default.
+ *
+ * @return number of characters appended to [out], or `-1` if can't read from the channel
+ * @throws TooLongLineException if the line exceeds [limit]
+ * @throws EOFException if the channel is closed before a line break is found
+ */
+@OptIn(InternalAPI::class, InternalIoApi::class)
+public suspend fun ByteReadChannel.readLineStrictTo(
+    out: Appendable,
+    limit: Long = Long.MAX_VALUE,
```
osipxd, line 655: **Question for review:** Changed the type of limit to `Long` as kotlinx-io operates with long. However I'm not sure if it's good or bad for the API. In one hand we don't limit users with `Int.MAX_VALUE`, but in other hand this might require adding `.toLong()` in existing code.

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
osipxd: Created an issue for further improvements: [KTOR-9178](https://youtrack.jetbrains.com/issue/KTOR-9178)

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
bjhham: > Do you think we should split these functions?

Maybe if splitting them out isn't viable, you could try function composition?

Something like...

```
@OptIn(InternalAPI::class, InternalIoApi::class)
public suspend fun ByteReadChannel.readLineStrictTo(
    out: Appendable,
    limit: Long = Long.MAX_VALUE,
    lineEnding: LineEnding = LineEnding.Default,
): Long {
    require(limit >= 0) { "Limit ($limit) should be non-negative" }

    return when(lineEnding) {
        LineEnding.Default -> internalReadLineTo(
            out,
            limit,
            throwOnIncompleteLine = true,
            scanForCr = DefaultScan,
            discardCr = DefaultDiscard,
        )
        LineEnding.Lenient -> internalReadLineTo(
            out,
            limit,
            throwOnIncompleteLine = true,
            scanForCr = LenientScan,
            discardCr = LenientDiscard,
        )
    }
}

@OptIn(InternalAPI::class, InternalIoApi::class)
private suspend fun ByteReadChannel.internalReadLineTo(
    out: Appendable,
    limit: Long,
    throwOnIncompleteLine: Boolean,
    scanForCr: Source.(Long, Long) -> Long,
    discardCr: Source.() -> Boolean,
): Long
```

Then you could use `inline` / `crossinline` for slightly better performance 🤷

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
bjhham: Yeah, looks like a good place for a `check()`.

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
bjhham: I'm ok with returning null here, since it plays into the common pattern for iterating on lines.

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
bjhham: I'm ok with Long here.  Best to keep it consistent, though I can't imagine anyone using a value outside the max integer range.

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
osipxd: Do you mean something like this?

```kotlin
// Default
scanForCr = { _, _ -> -1L }
discardCr = { false }

// Lenient
scanForCr = { lfIndex, limitLeft ->
    val endIndex = when (lfIndex) {
        // Subtract 1 from source.remaining to ignore the last byte,
        // when RC might be part of a CRLF sequence split into two buffers
        -1L -> minOf(limitLeft, remaining - 1)
        0L -> 0
        // Subtract 1 from lfIndex to ignore the case when CR is part of the CRLF sequence
        else -> lfIndex - 1
    }

    indexOf(CR, endIndex = endIndex)
}
discardCr = {
    discard(1)
    true
}
```

Maybe it will be enough to just introduce an inner function `scanForCr`?
```kotlin
fun Source.scanForCr(lfIndex: Long, limitLeft: Long): Long {
    // Don't scan for CR in default mode
    if (!lenientLineEnding) return -1L
    
    val endIndex = ...
    return indexOf(CR, endIndex = endIndex)
}
```

osipxd: You can check the variant with an inner function in this commit: https://github.com/ktorio/ktor/pull/5236/changes/a9ac7a8aa9f8054fdcb5d405cf01c858a04e9992

bjhham: Yeah, that's what I was thinking.  Just to shift the branching up a higher level.  It's fine with the checks inside the function, it just adds a little more complexity.

osipxd: I've failed to make this function `inline` as inline functions don't support inner functions 😢 
So I'd proceed with an inner `scanForCr` function. WDYT?

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
```diff
@@ -434,9 +417,102 @@ public suspend fun ByteReadChannel.discard(max: Long = Long.MAX_VALUE): Long {
...
+ *     channel.readLineStrictTo(out, limit = 1024, lineEnding = LineEnding.Lenient) >= 0
+ * } catch (_: EOFException) {
+ *     true
+ * }
+ * ```
+ * However, we recommend to use [LineEnding.Default] if possible and verify if the case with
+ * the unexpected end of line should actually be ignored. We expect the following code to be correct in most cases:
+ * ```
+ * val success = channel.readLineTo(out, limit = 1024) >= 0
+ * ```
```
osipxd, line 516: @bjhham, @vnikolova, could you check KDocs here? I suspect they're not actionable enough, but I don't how to improve yet keep it simple

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
```diff
@@ -434,9 +417,102 @@ public suspend fun ByteReadChannel.discard(max: Long = Long.MAX_VALUE): Long {
...
+ *
+ * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.utils.io.readUTF8Line)
+ *
+ * @param max the maximum number of characters to read. Default is [Int.MAX_VALUE].
+ * @return a string containing the line read, or null if channel is closed
+ * @throws TooLongLineException if max is reached before encountering a line delimiter or end of input
+ */
+@Suppress("DEPRECATION")
+@Deprecated("Use readLineStrict instead. See deprecation notes for more details.")
+public suspend fun ByteReadChannel.readUTF8Line(max: Int = Int.MAX_VALUE): String? {
```
osipxd, line 470: These KDoc are almost similar to the first one, but for a function returning `String`.

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
```diff
@@ -466,61 +543,249 @@ public suspend fun ByteReadChannel.readUTF8LineTo(out: Appendable, max: Int = In
...
+ * Reads a line of UTF-8 characters to the specified [out] buffer.
+ *
+ * Appends UTF-8 characters until a line break is found or the channel is exhausted. Implicit line break
+ * is assumed if the channel ends without a line break. Line break characters are not included in the result.
+ *
+ * @param out the buffer to append line to (line break characters are not appended)
+ * @param lineEnding line ending mode. Accepts LF and CRLF by default.
+ *
+ * @return number of characters appended to [out], or `-1` if can't read from the channel
+ */
```
osipxd, line 595: Also here

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
```diff
@@ -466,61 +543,249 @@ public suspend fun ByteReadChannel.readUTF8LineTo(out: Appendable, max: Int = In
...
+ * @param out the buffer to append line to (line break characters are not appended)
+ * @param limit maximum characters to append to [out]. Unlimited by default.
+ * @param lineEnding line ending mode. Accepts LF and CRLF by default.
+ *
+ * @return number of characters appended to [out], or `-1` if can't read from the channel
+ * @throws TooLongLineException if the line exceeds [limit]
+ * @throws EOFException if the channel is closed before a line break is found
+ */
+@OptIn(InternalAPI::class, InternalIoApi::class)
+public suspend fun ByteReadChannel.readLineStrictTo(
```
osipxd, line 653: And here

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
bjhham: It seems like enough information for me.  It might be good to link to the relevant kotlinx-io function docs too.

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
```diff
@@ -466,61 +543,249 @@ public suspend fun ByteReadChannel.readUTF8LineTo(out: Appendable, max: Int = In
...
+        val endIndex = when (lfIndex) {
+            // Subtract 1 from remaining to ignore the last byte,
+            // when RC might be part of a CRLF sequence split into two buffers
+            -1L -> minOf(limitLeft, remaining - 1)
+            0L -> 0
+            // Subtract 1 from lfIndex to ignore the case when CR is part of the CRLF sequence
+            else -> lfIndex - 1
+        }
+        return indexOf(CR, endIndex = endIndex)
+    }
```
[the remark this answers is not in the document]
osipxd: Currently the code is written in the way when `remaining` is never less than 1 when this function is called. And for `remaining == 1` it works fine, as `endIndex = 1 - 1 = 0` and it is a valid end index.

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
```diff
@@ -466,61 +543,250 @@ public suspend fun ByteReadChannel.readUTF8LineTo(out: Appendable, max: Int = In
...
+        if (readBuffer.buffer[count - 1] == CR) {
+            transferString(count = count - 1)
+            if (readBuffer.discardCrlfOrCr()) return consumed
+            transferString(1) // transfer the CR
+        } else {
+            // No new line separator
+            transferString(count)
+        }
+
+        if (consumed < limit && readBuffer.remaining == 0L && !awaitContent()) break
```
osipxd, line 754: Is it okay to check `readBuffer.remaining == 0L` or should I call `exhuasted()` instead?

bjhham: Looks like `readBuffer.remaining == 0` could return a false positive when using a raw source with `SourceByteReadChannel`, but this is mitigated by calling `awaitContent()` afterwards.

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
```diff
@@ -434,9 +417,102 @@ public suspend fun ByteReadChannel.discard(max: Long = Long.MAX_VALUE): Long {
...
+}
+
 /**
  * Reads a line of UTF-8 characters to the specified [out] buffer.
- * It recognizes CR, LF and CRLF as a line delimiter.
+ * It recognizes CR, LF, and CRLF as a line delimiter.
+ *
+ * ## Deprecation Notes
+ *
+ * This function is deprecated in favor of [readLineStrictTo] and [readLineTo].
```
vnikolova:
```suggestion
 * This function is deprecated. Use [readLineStrictTo] or [readLineTo] instead.
```

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
```diff
@@ -434,9 +417,102 @@ public suspend fun ByteReadChannel.discard(max: Long = Long.MAX_VALUE): Long {
...
+    val completed = readUTF8LineTo(result, max)
+    return if (!completed) null else result.toString()
+}
+
 /**
  * Reads a line of UTF-8 characters to the specified [out] buffer.
- * It recognizes CR, LF and CRLF as a line delimiter.
+ * It recognizes CR, LF, and CRLF as a line delimiter.
+ *
+ * ## Deprecation Notes
```
vnikolova:
```suggestion
 * ## Deprecation
```

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
```diff
@@ -434,9 +417,102 @@ public suspend fun ByteReadChannel.discard(max: Long = Long.MAX_VALUE): Long {
...
+ *
+ * This function is deprecated in favor of [readLineStrictTo] and [readLineTo].
+ *
+ * Use [readLineStrictTo] when:
+ * - Need to limit line length and throw [TooLongLineException] if the limit is reached
```
vnikolova:
```suggestion
 * - You need to limit the line length and throw [TooLongLineException] when the limit is reached.
```

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
```diff
@@ -434,9 +417,102 @@ public suspend fun ByteReadChannel.discard(max: Long = Long.MAX_VALUE): Long {
...
+ * - Expect an explicit line break at the end of a line. [EOFException] will be thrown
```
vnikolova:
```suggestion
 * - You expect an explicit line break at the end of a line. [EOFException] will be thrown
```

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
```diff
@@ -434,9 +417,102 @@ public suspend fun ByteReadChannel.discard(max: Long = Long.MAX_VALUE): Long {
...
+ *   if the channel is closed before the line break
+ *
+ * Use [readLineTo] when:
+ * - No need to limit line length
```
vnikolova:
```suggestion
 * - You do not need to limit line length.
```

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
```diff
@@ -434,9 +417,102 @@ public suspend fun ByteReadChannel.discard(max: Long = Long.MAX_VALUE): Long {
...
+ * - An explicit line break at the end of the channel is not mandatory
```
vnikolova:
```suggestion
 * - You do not require an explicit line break at the end of the channel.
```

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
```diff
@@ -434,9 +417,102 @@ public suspend fun ByteReadChannel.discard(max: Long = Long.MAX_VALUE): Long {
...
+ *
+ * Other changes:
```
vnikolova:
```suggestion
 * Compared to `readUTF8LineTo`, the new functions differ in the following ways:
```

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
```diff
@@ -434,9 +417,102 @@ public suspend fun ByteReadChannel.discard(max: Long = Long.MAX_VALUE): Long {
...
+ * - New functions recognize LF and CRLF as a line delimiter by default. This default comes with better performance.
```
vnikolova:
```suggestion
 * - They recognize LF and CRLF as line delimiters by default, which provides better performance.
```

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
```diff
@@ -434,9 +417,102 @@ public suspend fun ByteReadChannel.discard(max: Long = Long.MAX_VALUE): Long {
...
+ *   To keep current behavior and recognize all line delimiters (CR, LF, and CRLF), specify [LineEnding.Lenient].
+ * - New functions return number of appended characters instead of [Boolean] or `-1` if the channel is empty.
```
vnikolova:
```suggestion
 * - They return the number of appended characters, or`-1` if the channel is empty, instead of returning [Boolean].
```

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
```diff
@@ -434,9 +417,102 @@ public suspend fun ByteReadChannel.discard(max: Long = Long.MAX_VALUE): Long {
...
+ * val success = channel.readUTF8LineTo(out, max = 1024)
+ *
+ * // After
+ * val success = try {
+ *     channel.readLineStrictTo(out, limit = 1024, lineEnding = LineEnding.Lenient) >= 0
+ * } catch (_: EOFException) {
+ *     true
+ * }
+ * ```
+ * However, we recommend to use [LineEnding.Default] if possible and verify if the case with
```
vnikolova: Suggestion for added clarity and conciseness: 
" However, we recommend using the default line ending behavior ([LineEnding.Default]) and verifying unexpected end-of-input explicitly:"

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
vnikolova: I suggested some small improvements for added clarity and actionability, but it generally looks good as it is!

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
```diff
@@ -466,61 +543,250 @@ public suspend fun ByteReadChannel.readUTF8LineTo(out: Appendable, max: Int = In
...
+/**
+ * Reads a line of UTF-8 characters to the specified [out] buffer.
+ *
+ * Appends UTF-8 characters until a line break is found or the channel is exhausted. Implicit line break
+ * is assumed if the channel ends without a line break. Line break characters are not included in the result.
+ *
+ * @param out the buffer to append line to (line break characters are not appended)
+ * @param lineEnding line ending mode. Accepts LF and CRLF by default.
+ *
+ * @return number of characters appended to [out], or `-1` if can't read from the channel
```
vnikolova, line 594:
```suggestion
 * @return number of characters appended to [out], or `-1` if the channel is empty
```

PR #5241, opened 2025-12-08
CORS: Allow any header for the simmetry with anyHost

y9san9:
**Subsystem**
Server: CORS

**Motivation**
There is a function to allow any host (anyHost). Since 2016 it is allowed to pass '*' to headers as well: https://github.com/whatwg/fetch/issues/251. I expect to have anyHeader.

**Solution**
Added the function.


bjhham (changes requested):
Please introduce a test in `io.ktor.tests.server.plugins.CORSTest` for the new behaviour.

It looks like this will respond with a forbidden from using the `headers` field instead of `headerPredicates`.  Check out `io.ktor.server.plugins.cors.CORSUtilsKt#corsCheckRequestHeaders`.

Mr3zee:
Closed for inacivity

PR #5243, opened 2025-12-08
KTOR-9162 Auth API key plugin

nathanfallet:
Everything is explained in https://youtrack.jetbrains.com/issue/KTOR-9162/Auth-API-key-plugin

The idea is to provide a plugin with a configuration like this:
```kotlin
val key = environment.config.property("app.key").getString()
authentication {
    apiKey("api-key") {
        headerName = "X-Api-Key"
        validate { keyFromHeader ->
            keyFromHeader.takeIf { it == key }?.let { AppPrincipal(it) }
        }
    }
}
```

Then you can protect your endpoints like this:
```kotlin
routing {
    authenticate("api-key") {
        // ...
    }
}
```

### DIFF
--- a/gradle/artifacts/publishJvmAndCommonPublications.txt
+++ b/gradle/artifacts/publishJvmAndCommonPublications.txt
@@ -331,6 +331,13 @@ io.ktor:ktor-serialization/.jar
 io.ktor:ktor-serialization/javadoc.jar
 io.ktor:ktor-serialization/kotlin-tooling-metadata.json
 io.ktor:ktor-serialization/sources.jar
+io.ktor:ktor-server-auth-api-key-jvm/.jar
+io.ktor:ktor-server-auth-api-key-jvm/javadoc.jar
+io.ktor:ktor-server-auth-api-key-jvm/sources.jar
+io.ktor:ktor-server-auth-api-key/.jar
+io.ktor:ktor-server-auth-api-key/javadoc.jar
+io.ktor:ktor-server-auth-api-key/kotlin-tooling-metadata.json
+io.ktor:ktor-server-auth-api-key/sources.jar
 io.ktor:ktor-server-auth-jvm/.jar
 io.ktor:ktor-server-auth-jvm/javadoc.jar
 io.ktor:ktor-server-auth-jvm/sources.jar


nathanfallet:
I'll have a look at the ABI part soon, it's the first time I see this in any Kotlin project.

You can already give your feedback on the actual plugin implementation 😄

bjhham:
@nathanfallet the allocation benchmarks are just failing from a publications snapshot check due to the new artifact being introduced.

You can run `./gradlew validatePublishedArtifacts --dump` and push the new snapshot to fix this.

ktor-server/ktor-server-plugins/ktor-server-auth-api-key/common/test/io/ktor/server/auth/apikey/TestApiKeyAuth.kt
```diff
@@ -0,0 +1,154 @@
...
+import io.ktor.server.auth.*
+import io.ktor.server.response.*
+import io.ktor.server.testing.*
+import kotlinx.serialization.Serializable
+import kotlin.random.Random
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertFailsWith
+
+class TestApiKeyAuth {
```
bjhham: Could you please rename this to `ApiKeyAuthTest` to match our convention, as well as the other test classes?

ktor-server/ktor-server-plugins/ktor-server-auth-api-key/common/src/io/ktor/server/auth/apikey/ApiKeyAuth.kt
```diff
@@ -0,0 +1,125 @@
...
+
+        /**
+         * Name of the scheme used when challenge fails, see [AuthenticationContext.challenge].
+         */
+        public var authScheme: String = "apiKey"
+
+        /**
+         * Name of the header that will be used as a source for the api key.
+         */
+        public var headerName: String = "X-Api-Key"
```
bjhham: Perhaps we could expose the default values as public constants under a `ApiKeyAuth` object for a little extra tooling when using defaults - then we can remove the declaration in the tests.

nathanfallet:
I'll do update for what you commented.
Also, I saw that documentation was mentioned in the YouTrack issue, I'm okay to do it as soon as this PR is merged (in the ktor-documentation repo I guess)

Finally, for the dump, on both macOS 26.1 and Ubuntu 24.04 I got this error while running the above command:

```
> Task :ktor-io:commonizeCInterop
Exception in thread "main" java.lang.IllegalStateException: Unresolved classifier: platform/posix/pthread_mutex_t
```

Both have Java 21 installed, as well as libcurl/libncurses (as mentioned in CONTRIBUTING.md).
If any maintainer wants to run that command on my branch and commit the result (if it works for you) I'll be okay with it too.

bjhham:
Oh yes, the c commonizer stuff can cause problems sometimes.  Can you try the dump once again with `--rerun-tasks`?  If this also fails I can take over updating this.

ktor-server/ktor-server-plugins/ktor-server-auth-api-key/common/src/io/ktor/server/auth/apikey/ApiKeyAuth.kt
nathanfallet: Now available under `ApiKeyAuth.DEFAULT_HEADER_NAME`

ktor-server/ktor-server-plugins/ktor-server-auth-api-key/common/test/io/ktor/server/auth/apikey/TestApiKeyAuth.kt
nathanfallet: Done! (only two files left since I removed the examples ones as requested)

nathanfallet:
@bjhham I tried with `--rerun-tasks`, the error is gone but no file is getting updated. So if you can take over it I'll ba happy about that.
Also, there is a failing test on macOS (io.ktor.client.tests.plugins.ServerSentEventsTest.testPostRequest[macosX64]) but this might be due to a change when I merged main into my branch since I don't think a new independent module would impact other modules like this.
Let me know if this is good or not 😄

PR #5244, opened 2025-12-10
KTOR-9085 Infer security schemes from `Authentication` plugin

zibet27:
OpenAPI Security. KTOR-9085 Infer security schemes from the `Authentication` plugin.

**Subsystem**
Server OpeAPI.

**Motivation**
[KTOR-9085](https://youtrack.jetbrains.com/issue/KTOR-9085) Read OpenAPI security details from authentication plugin

**Solution**
New API:
- Introduced `securitySchemes` as an OpenAPI Spec component. It has all implementations described in the [spec](https://swagger.io/docs/specification/v3_0/authentication/)
-  Added application attribute to store all security schemes globally.
- Security schemes can be added manually or inferred from the `Authentication` plugin (we allow combining these approaches, manual has higher priority).
- `:ktor-server-routine-annotate` now depends on `ktor-server-auth` via `compileTime`. If a developer tries to infer smth from the auth plugin that isn't installed, it should throw a runtime error. It seems to be working, but I'm not sure about non-JVM platforms.
- Not sure if security schema description should be null or "general" by default(see code).

Changes to old API:
- Added optional `description` field to`AuthenticationProvider` and its descendants with new helper functions. Used for API docs only.
- Added `settings` property inside `OAuthAuthenticationProvider`, which works like `providerLookup` but static, allows generating OpenAPI docs and should be a bit faster.
-  Added `OAuth2ServerSettings.refreshUrl` and `defaultScopeDescriptions` for docs purposes.
- Made some auth plugin API public with `@InternalAPI` instead of internal.

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-auth-api-key/api/ktor-server-auth-api-key.api
+++ b/ktor-server/ktor-server-plugins/ktor-server-auth-api-key/api/ktor-server-auth-api-key.api
@@ -5,7 +5,9 @@ public final class io/ktor/server/auth/apikey/ApiKeyAuth {
 
 public final class io/ktor/server/auth/apikey/ApiKeyAuthKt {
 	public static final fun apiKey (Lio/ktor/server/auth/AuthenticationConfig;Ljava/lang/String;Ljava/lang/String;Lkotlin/jvm/functions/Function1;)V
+	public static final fun apiKey (Lio/ktor/server/auth/AuthenticationConfig;Ljava/lang/String;Lkotlin/jvm/functions/Function1;)V
 	public static synthetic fun apiKey$default (Lio/ktor/server/auth/AuthenticationConfig;Ljava/lang/String;Ljava/lang/String;Lkotlin/jvm/functions/Function1;ILjava/lang/Object;)V
+	public static synthetic fun apiKey$default (Lio/ktor/server/auth/AuthenticationConfig;Ljava/lang/String;Lkotlin/jvm/functions/Function1;ILjava/lang/Object;)V
 }
 
 public final class io/ktor/server/auth/apikey/ApiKeyAuthenticationProvider : io/ktor/server/auth/AuthenticationProvider {
--- a/ktor-server/ktor-server-plugins/ktor-server-auth-api-key/api/ktor-server-auth-api-key.klib.api
+++ b/ktor-server/ktor-server-plugins/ktor-server-auth-api-key/api/ktor-server-auth-api-key.klib.api
@@ -30,4 +30,5 @@ final object io.ktor.server.auth.apikey/ApiKeyAuth { // io.ktor.server.auth.apik
         final fun <get-DEFAULT_HEADER_NAME>(): kotlin/String // io.ktor.server.auth.apikey/ApiKeyAuth.DEFAULT_HEADER_NAME.<get-DEFAULT_HEADER_NAME>|<get-DEFAULT_HEADER_NAME>(){}[0]
 }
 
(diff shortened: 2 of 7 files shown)


ktor-server/ktor-server-plugins/ktor-server-routing-annotate/common/src/io/ktor/annotate/SecuritySchemeAnnotation.kt
```diff
@@ -0,0 +1,230 @@
...
+            )
+        }
+
+        is FormAuthenticationProvider -> HttpSecurityScheme(
+            scheme = "form",
+            description = provider.description ?: (
+                "Form-based Authentication with post parameters " +
+                    "${provider.userParamName} and ${provider.passwordParamName}"
+                )
+        )
```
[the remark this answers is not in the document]
zibet27: @bjhham What do you think is the correct scheme for form-based auth? Afaik, there is no standard way to describe it in OpenAPI. I feel `HttpSecurityScheme` is the closest here.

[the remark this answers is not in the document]
bjhham: It would need to be ignored for the OpenAPI spec here.  I think we can recommend using manual overrides for this one, since it's not standard.  Basic auth would be the best way to handle it anyway.

bjhham (commented):
Looking good!  I'll take another gander in the morning 👀

bjhham:
Seems you've got a test failure...

> expected: \<Http Digest Authentication\> but was: \<HTTP Digest Authentication\>

Very helpful assert message.

zibet27:
> Seems you've got a test failure...
> 
> > expected: <Http Digest Authentication> but was: <HTTP Digest Authentication>
> 
> Very helpful assert message.

Yeah, it's a last-minute refactor 😅. Also, some tests from `auth` were failing.

bjhham:
Now that https://github.com/ktorio/ktor/pull/5243 is merged, could you include the API key support?  There seems to be a provision in OpenAPI for it https://swagger.io/docs/specification/v3_0/authentication/api-keys/

ktor-server/ktor-server-plugins/ktor-server-auth/common/src/io/ktor/server/auth/OAuthCommon.kt
```diff
@@ -70,16 +71,19 @@ public sealed class OAuthServerSettings(public val name: String, public val vers
...
         name: String,
         public val authorizeUrl: String,
         public val accessTokenUrl: String,
+        public val refreshUrl: String? = null,
         public val requestMethod: HttpMethod = HttpMethod.Get,
         public val clientId: String,
         public val clientSecret: String,
         public val defaultScopes: List<String> = emptyList(),
+        // TODO: merge `defaultScopes` and `defaultScopeDescriptions` in the next major release
+        public val defaultScopeDescriptions: Map<String, String> = emptyMap(),
```
bjhham, line 86: Could you log a YT ticket for this and target it for 4.0.0 so we don't forget?

zibet27: Sure! I'll create it after the PR is merged.

ktor-server/ktor-server-plugins/ktor-server-auth/common/src/io/ktor/server/auth/AuthenticationInterceptors.kt
```diff
@@ -208,6 +209,12 @@ private suspend fun AuthenticationContext.executeChallenges(
     return false
 }
 
+@OptIn(InternalAPI::class)
+public fun AuthenticationConfig.allProviders(): Map<String?, AuthenticationProvider> {
```
e5l, line 213: Could you tell me why it is possible to have a nullable key?

e5l (approved):
lgtm, please consider merging after the release

ktor-server/ktor-server-plugins/ktor-server-auth/common/src/io/ktor/server/auth/AuthenticationInterceptors.kt
zibet27: In ktor-auth, the auth provider name is nullable, null == "default". 
In this PR, I just made it publicly available.

zibet27:
> lgtm, please consider merging after the release

Sure, some OpenAPI Security features are still missing.

PR #5246, opened 2025-12-10
KTOR-9102 Fix closing underlying connection in Java engine, fix sse closing

marychatte:
**Subsystem**
Client, SSE

**Motivation**
[KTOR-9102](https://youtrack.jetbrains.com/issue/KTOR-9102) SSE: Java engine does not close the underlying connection when SSE session is canceled


ktor-client/ktor-client-java/jvm/src/io/ktor/client/engine/java/JavaHttpResponseBodyHandler.kt
```diff
@@ -156,7 +159,9 @@ internal class JavaHttpResponseBodyHandler(
                 queue.close(cause)
                 subscription.getAndSet(null)?.cancel()
             } finally {
-                consumerJob.completeExceptionally(cause)
+                cause?.apply {
+                    consumerJob.completeExceptionally(cause)
+                } ?: consumerJob.complete()
```
osipxd: Minor. `.let` + function reference is probably more suitable here
```suggestion
                cause?.let(consumerJob::completeExceptionally) ?: consumerJob.complete()
```

PR #5248, opened 2025-12-15
KTOR-9166 Change logging level to TRACE to avoid excessive messages

Stexxe:
https://youtrack.jetbrains.com/issue/KTOR-9166

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-cors/common/src/io/ktor/server/plugins/cors/CORS.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-cors/common/src/io/ktor/server/plugins/cors/CORS.kt
@@ -87,7 +87,7 @@ internal fun PluginBuilder<CORSConfig>.buildPlugin() {
             return@onCall
         }
 
-        LOGGER.info { "${call.request.id()}: Start handler" }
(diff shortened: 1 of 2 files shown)


ktor-server/ktor-server-plugins/ktor-server-cors/common/src/io/ktor/server/plugins/cors/CORS.kt
```diff
@@ -116,7 +116,7 @@ internal fun PluginBuilder<CORSConfig>.buildPlugin() {
 
             OriginCheckResult.SkipCORS -> return@onCall
             OriginCheckResult.Failed -> {
-                LOGGER.info { "${call.request.id()}: CORS check fails because Origin $origin does not match" }
+                LOGGER.trace { "${call.request.id()}: CORS check fails because Origin $origin does not match" }
```
bjhham, line 119: The failure scenario is OK with info so we know when illegal access is attempted - we just need to avoid logging on valid scenarios.

Stexxe: Some messages depend on each other, so for consistency, I'd like to log on the same level.

PR #5249, opened 2025-12-15
OpenAPI Security route annotation

zibet27:
**Subsystem**
Server. OpenAPI

**Motivation**
[KTOR-9189](https://youtrack.jetbrains.com/issue/KTOR-9189) Annotate `Route`s security based on the `Authentication` plugin.

**Solution**
- generate operation security block if the selector is `AuthenticationRouteSelector`
- if any auth provider is OAuth2, include its scopes
- cache resolved security schemes instead of generating a new map on every route selector.

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-routing-annotate/common/src/io/ktor/annotate/RouteAnnotationApi.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-routing-annotate/common/src/io/ktor/annotate/RouteAnnotationApi.kt
@@ -171,47 +171,49 @@ private fun RoutingNode.operationFromSelector(): Operation? {
             }
         }
 
-        is AuthenticationRouteSelector -> {
-            val globalSchemes = application.findSecuritySchemes(useCache = true)
-            val registration = attributes.getOrNull(AuthenticateProvidersKey)
-            val strategy = registration?.strategy ?: AuthenticationStrategy.FirstSuccessful
+        else -> if (application.isAuthPluginInstalled) operationFromAuthSelector() else null
+    }
+}
 
-            Operation.build {
-                security {
-                    fun firstSuccessful() {
-                        // At least one of the schemes must succeed (OR relationship)
+private fun RoutingNode.operationFromAuthSelector(): Operation? {
+    val paramSelector = selector as? AuthenticationRouteSelector
+        ?: return null
+    val globalSchemes = application.findSecuritySchemes(useCache = true)
+    val registration = attributes.getOrNull(AuthenticateProvidersKey)
+    val strategy = registration?.strategy ?: AuthenticationStrategy.FirstSuccessful
+
+    return Operation.build {
+        security {
+            fun firstSuccessful() {
+                // At least one of the schemes must succeed (OR relationship)
+                for (providerName in paramSelector.names) {
+                    val schemeName = providerName ?: AuthenticationRouteSelector.DEFAULT_NAME
+                    requirement(schemeName, scopes = globalSchemes.scopesFor(schemeName))
+                }
+            }
+
+            when (strategy) {
+                AuthenticationStrategy.Optional -> {
+                    firstSuccessful()
+                    optional()
+                }
+
+                AuthenticationStrategy.FirstSuccessful -> {
+                    firstSuccessful()
+                }
+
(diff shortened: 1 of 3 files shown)


zibet27:
@coderabbitai review

zibet27:
@coderabbitai review

ktor-server/ktor-server-plugins/ktor-server-routing-annotate/common/src/io/ktor/annotate/SecuritySchemeAnnotation.kt
```diff
@@ -14,13 +14,20 @@ import io.ktor.server.sessions.SessionTransportCookie
...
  * Attribute key for storing OpenAPI security scheme metadata for authentication providers.
  * Maps provider names to their corresponding SecurityScheme definitions.
  */
-public val AuthenticationSecuritySchemesAttributeKey: AttributeKey<Map<String, SecurityScheme>> =
-    AttributeKey("AuthenticationSecuritySchemes")
+public val AuthSecuritySchemesAttributeKey: AttributeKey<Map<String, SecurityScheme>> =
+    AttributeKey("AuthSecuritySchemes")
+
+public val AuthSecuritySchemesCacheAttributeKey: AttributeKey<Map<String, SecurityScheme>> =
+    AttributeKey("InferredAuthSecuritySchemes")
```
bjhham: We can probably make this internal.

ktor-server/ktor-server-plugins/ktor-server-routing-annotate/common/src/io/ktor/annotate/RouteAnnotationApi.kt
```diff
@@ -49,10 +52,14 @@ public fun Route.annotate(configure: RouteAnnotationFunction): Route {
...
  * @param info The OpenAPI info object.
  * @param route The route to generate the specification for.
+ * @param inferSecurity Whether to infer security schemes from the authentication plugin. Defaults to `false`.
+ * @param inferJwtSecurity Whether to infer JWT security schemes from the authentication plugin. Defaults to `false`.
  */
 public fun generateOpenApiSpec(
     info: OpenApiInfo,
     route: RoutingNode,
+    inferSecurity: Boolean = false,
+    inferJwtSecurity: Boolean = false
```
bjhham: I missed in the other PR, but I'm not seeing why JWT should be a special case here.  What does it cost us to remove the logic for filtering it out when checking for other schemes?

zibet27: The issue is that it requires the `io.ktor:ktor-server-auth-jwt` package to be installed. Moreover, it has only jvm implementation.
We could try to catch every exception from `inferPlatformSpecificSecurityScheme`(can be `ClassNotFound` on JVM or `kotlin.native.internal.linkageerror`).

bjhham: I understand now...  Maybe we can simplify the dependency tree here by referencing the openapi schema from the auth plugins and include the logic of registering this info during plugin registration - then we can just duplicate the attribute key under the base auth plugin and the routing annotate module.  This way, users can override or register their own security details and the routing annotate plugin won't need to know any details about the different auth plugins.

zibet27: Not sure about this because `auth` will depend on OpenAPI and register schemes even if not needed.

bjhham: Yeah, ideally we could use a more generic model for the shared type that could live in auth.  Maybe even extend `AuthenticationProvider` with a function for providing metadata.  You could just stick with what you have now but I'd remove the `includeJwt` flag from the API and use `implementation` import on the jvm to avoid classpath / linkage errors.  It'd be an unnecessary import for most cases, but since it's not `api` we can still remove it later.

zibet27:
@coderabbitai review

PR #5253, opened 2025-12-16
WebRTC Client. Make `IceServer.urls` a list

zibet27:
**Subsystem**
WebRTC Client.

**Motivation**
[KTOR-9188](https://youtrack.jetbrains.com/issue/KTOR-9188) WebRTC. `IceServer.urls` should be a list.

**Solution**
Change the type of `urls`. I think it's ok, because the API is still experimental. Moreover, `IceServers` are usually initialised only once.
I wanted to add more tests, ensuring ice servers are actually considered, but it doesn't work on emulators/simulators.


ktor-client/ktor-client-webrtc/common/src/io/ktor/client/webrtc/WebRtc.kt
```diff
@@ -89,14 +89,14 @@ public object WebRtc {
...
-     * @property urls The URL(s) of the ICE server.
+     * @property urls The URLs of the ICE server.
      * @property username Optional username for the ICE server.
      * @property credential Optional credential for the ICE server.
      *
      * @see [MDN RTCIceServer](https://developer.mozilla.org/en-US/docs/Web/API/RTCIceServer)
      */
     public data class IceServer(
-        val urls: String,
+        val urls: List<String>,
```
bjhham, line 99: Could you introduce an overload constructor to mitigate the breaking change here?

zibet27: Yeah, that should work. Forgot about it :)

ktor-client/ktor-client-webrtc/jvm/test/io/ktor/client/webrtc/utils/ConnectionUtils.jvm.kt
```diff
@@ -0,0 +1,16 @@
...
+
+package io.ktor.client.webrtc.utils
+
+import io.ktor.client.webrtc.WebRtcClient
+import io.ktor.utils.io.ExperimentalKtorApi
+
+@OptIn(markerClass = [ExperimentalKtorApi::class])
+actual fun createTestWebRtcClient(): WebRtcClient {
+    TODO("Not yet implemented")
+}
```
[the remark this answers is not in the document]
zibet27: The real implementation should be merged soon. Now, the tests on JVM are disabled.

PR #5258, opened 2025-12-18
KTOR-9148 Fix logging multipart/form-data requests and binary body detection

Stexxe:
https://youtrack.jetbrains.com/issue/KTOR-9148


ktor-client/ktor-client-plugins/ktor-client-logging/common/src/io/ktor/client/plugins/logging/Logging.kt
```diff
@@ -119,74 +119,45 @@ public val Logging: ClientPlugin<LoggingConfig> = createClientPlugin("Logging",
...
+            contentType.match("application/xml") ||
+            contentType.match("application/ld+json") ||
+            contentType.match("application/xhtml+xml") ||
+            contentType.match("application/rss+xml") ||
+            contentType.match("application/atom+xml") ||
+            contentType.match("application/x-www-form-urlencoded") ||
+            contentType.match("image/svg+xml")
+        ) {
+            return true
+        }
```
Goooler, line 136: All of 

https://github.com/ktorio/ktor/blob/dba895808b79e4661eb54e6dca8dc8ac82146b68/ktor-http/common/src/io/ktor/http/ContentTypes.kt#L190-L200

except `cbor`/`octet-stream` are both plain text.

Might be great to add an extension for `ContentType`?

bjhham: +1, a helper extension would be handy.

Since this is called on each request and there's no ambiguity with the type category, we can use a when or set contains on the subtype here to improve performance on the check.

ktor-client/ktor-client-plugins/ktor-client-logging/common/src/io/ktor/client/plugins/logging/Logging.kt
```diff
@@ -119,74 +119,45 @@ public val Logging: ClientPlugin<LoggingConfig> = createClientPlugin("Logging",
...
+    }
+
     /**
      * Detects if the body is a binary data
      * @return
      * Boolean: true if the body is a binary data.
      * Long?: body size if calculated.
      * ByteReadChannel: body channel with the original data.
      */
-    suspend fun detectIfBinary(
```
bjhham, line 129: Since this is going to break support for responses with missing `Content-Type` headers, I think it would be prudent to provide an override in `LoggingConfig` for detecting if the body is binary.  In this case, the new content type check would be the default function, and can advise how to restore the old body reading check in our documentation with caveats on why it could be problematic.

Stexxe: I can make the plugin log the response body as text if the Content-Type header is not present, which will align the behavior with the default format. We can introduce the configuration method, but I wonder if anyone needs it.

ktor-client/ktor-client-plugins/ktor-client-logging/common/src/io/ktor/client/plugins/logging/Logging.kt
Stexxe: I can improve the performance of that method if you like, but I suggest introducing the helper in another PR.

ktor-client/ktor-client-plugins/ktor-client-logging/common/src/io/ktor/client/plugins/logging/Logging.kt
bjhham: > I can make the plugin log the response body as text if the Content-Type header is not present, which will align the behavior with the default format

We'd still have a behaviour change from the automatic detection here, and some may be relying on this.

I think introducing a check override in the config type is the easiest way for flexibility here.  We also have an ask for [turning on/off body logging during runtime](https://jetbrains.slack.com/archives/C07U498LLUR/p1765461151787779) so we could kill two birds with one stone.

Something like this could provide all the flexibility we need for 3.4.0:

```kotlin
public class LoggingConfig {
    public var includeBody: (HttpResponse) -> Boolean = { response -> detectIfBinary(response.contentType(), response.headers) }
}
```

Stexxe: What if the level is set to `BODY`, but `includeBody` returns `false`? If `includeBody` returns false, should the `binary body ommited` message still be logged with the OkHttp format? 
In my opinion, turning body logging off does not equal detecting the body as binary.

bjhham: It ought to be `BODY` as a global level config and `includeBody` as a per-request check for omission, so both would be required for logging the body.  And I would think `body omited` would be shown when `includeBody` is `false` and we're logging with `BODY` level.

Stexxe:
The latest fix can be found in #5270.

PR #5259, opened 2025-12-19
Append content types and mime types

Goooler:
There are a lot of commonly used types are missing.

### DIFF
--- a/ktor-http/common/src/io/ktor/http/Mimes.kt
+++ b/ktor-http/common/src/io/ktor/http/Mimes.kt
@@ -11,6 +11,7 @@ internal const val INITIAL_MIMES_LIST_SIZE: Int = 1217
 private val rawMimes: String
     get() = """
 application/acad,dwg
+application/activity+json,activity
 application/andrew-inset,ez
 application/applixware,aw
 application/arj,arj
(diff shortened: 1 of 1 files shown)


Goooler:
Addressing 9c678443ef4d9755e1ae1ae317a74273dab2e4c3 into #5260, will rebase after that PR gets merged.

ktor-http/common/src/io/ktor/http/Mimes.kt
```diff
@@ -32,6 +34,7 @@ application/clariscad,ccad
 application/commonground,dp
 application/cu-seeme,cu
 application/davmount+xml,davmount
+application/did+json,didjson
```
osipxd, line 37: It's not obvious from the name of the field, but its purpose is to associate file extensions with media types. So media types that are not associated with any file extensions shouldn't be listed here.
Could you add a clarifying comment for the field? And could you remove those of the newly added media-types that are not associated with any extension? You can consult with [this file from the httpd project](https://github.com/apache/httpd/blob/trunk/docs/conf/mime.types) to quickly check if a media type has registered file extension.

Goooler: This is generated from copilot, I never use it.

Goooler:
> Btw, we have an issue [KTOR-7587](https://youtrack.jetbrains.com/projects/KTOR/issues/KTOR-7587) "Update media types (MIME types) list". Are you interested in working on it?

Good to know it. Closing this PR for now. I'll probably address a new one after some discussions after #5260.

PR #5260, opened 2025-12-19
Resort rawMimes

Goooler:
Separated from #5259 to reduce diff.

### DIFF
--- a/ktor-http/common/src/io/ktor/http/Mimes.kt
+++ b/ktor-http/common/src/io/ktor/http/Mimes.kt
@@ -8,6 +8,7 @@ import io.ktor.util.*
 
 internal const val INITIAL_MIMES_LIST_SIZE: Int = 1217
 
+// If you add new lines, make sure they are sorted.
 private val rawMimes: String
     get() = """
 application/acad,dwg
@@ -65,11 +66,11 @@ application/java-vm,class
 application/json,json
 application/lha,lha
 application/lzx,lzx
-application/macbinary,bin
 application/mac-binary,bin
 application/mac-binhex,hqx
 application/mac-binhex40,hqx
 application/mac-compactpro,cpt
+application/macbinary,bin
 application/mads+xml,mads
 application/manifest+json,webmanifest
 application/marc,mrc
@@ -85,6 +86,7 @@ application/mets+xml,mets
 application/mime,aps
 application/mods+xml,mods
 application/mp21,m21
+application/mp4,mp4
 application/mspowerpoint,pot pps ppt ppz
 application/msword,doc dot w6w wiz word
 application/mswrite,wri
@@ -536,7 +538,6 @@ application/x-cmu-raster,ras
 application/x-cocoa,cco
 application/x-compactpro,cpt
 application/x-compress,z
-application/zip,zip
 application/x-compressed,gz tgz z zip
 application/x-conference,nsc
 application/x-cpio,cpio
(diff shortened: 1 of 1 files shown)


ktor-http/common/src/io/ktor/http/Mimes.kt
```diff
@@ -731,6 +735,7 @@ audio/x-au,au
 audio/x-gsm,gsd gsm
 audio/x-jam,jam
 audio/x-liveaudio,lam
+audio/x-matroska,mkv
```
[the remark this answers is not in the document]
Goooler: It's just moved.

ktor-http/common/src/io/ktor/http/Mimes.kt
```diff
@@ -8,6 +8,7 @@ import io.ktor.util.*
 
 internal const val INITIAL_MIMES_LIST_SIZE: Int = 1217
 
+// If you add new lines, make sure they are sorted.
```
osipxd, line 11: Thank you for you initiative!

Unfortunately we can't sort these values.
Order here matters as the first media type declared for an extension is considered to be the default one.
It affects the result of the `ContentType.defaultForFileExtension(String)` function. For example if we want it to return "video/x-matroska" for "mkv", we should place this media type before "audio/x-matroska".

I think we should instead introduce a test checking that all media types associated with a specific extension are grouped together.
```kotlin
// This would pass
video/x-matroska,mkv
audio/x-matroska,mkv

// And this wouldn't
audio/x-matroska,mkv
video/quicktime,moov mov qt
video/x-matroska,mkv
```

That test would probably detect some additional cases like "application/json" and "text/json" that are in the proper order even without rearrangement, but it is okay to rearrange them.
What do you think?

Goooler: > Order here matters as the first media type declared for an extension is considered to be the default one.

`video/x-matroska` should be `mkv`, but `audio/x-matroska` should be `mka`. This is why the bot complained at https://github.com/ktorio/ktor/pull/5260#discussion_r2633145466.

It's also listed at https://github.com/apache/httpd/blob/befb52370cf75df77751393d7c3b06146c99d8a3/docs/conf/mime.types#L1520

I created a new issue for this, see https://youtrack.jetbrains.com/issue/KTOR-9201.

osipxd: Yes, it seems like a bug. Good catch!
However the same logic might be applied to another pair of media types. For example, for the ".zip" extension the default media type should be "applicaiton/zip", so we have to maintain a non-alphabetic order here:
```kotlin
application/zip,zip
application/x-compressed,gz tgz z zip
```

ktor-http/common/src/io/ktor/http/Mimes.kt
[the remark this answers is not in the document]
Goooler: Fixed in #5262.

ktor-http/common/src/io/ktor/http/Mimes.kt
Goooler: I just saw the test cases in https://github.com/ktorio/ktor/blob/d67df05f01020955d241d62425eaddb67be905ee/ktor-http/common/test/io/ktor/tests/http/ContentTypeLookupTest.kt#L19

Reasonable for not sorting the items.

PR #5263, opened 2025-12-23
KTOR-2162 Ktor Server Jetty Jakarta enable thread pool expansion

amccague:
**Subsystem**
Server, Jetty + Jetty Jakarta

**Motivation**

https://youtrack.jetbrains.com/issue/KTOR-2162

**Solution**
Fix `ThreadPoolExecutor` in `JettyKtorHandler` to actually utilize the configured maxPoolSize (8x burst capacity) by replacing the unbounded `LinkedBlockingQueue` with `SynchronousQueue`, which has no capacity and forces thread creation up to `maxPoolSize` before blocking.

This is the same pattern used by `Executors.newCachedThreadPool()`.

**Impact**
Backpressure behavior: When the pool reaches `callGroupSize * 8` threads and all are busy, the calling Jetty worker thread blocks. However, Jetty's own `QueuedThreadPool` provides natural request buffering upstream. Clients only experience connection-level backpressure after both pools are saturated.

This does change the client experience under overload:
* Current (unbounded queue): Connections accepted indefinitely, requests queue in memory, potential OOM, eventual response timeouts.
* After fix: Connections accepted up to Jetty's buffer capacity, then new connections wait/timeout at TCP level, fail-fast(er) behavior.

Edit: The `QueuedThreadPool` implemented in Ktor is currently done with an unbounded queue as well, so the behaviour will be unchanged for clients.

Benefit for user code: If application handlers inadvertently perform blocking operations (common, in my opinion), the pool now grows to absorb the load (up to 8x) rather than serializing all requests through `callGroupSize` threads while the unbounded queue grows silently.

I would like to explore being able to monitor this threadpool closer in the future, but for now - concurrent ktor handler threads can be a proxy. Also if Jetty threads become more reliably consumed these will be easier to monitor.

### DIFF
--- a/ktor-server/ktor-server-jetty-jakarta/jvm/src/io/ktor/server/jetty/jakarta/JettyKtorHandler.kt
+++ b/ktor-server/ktor-server-jetty-jakarta/jvm/src/io/ktor/server/jetty/jakarta/JettyKtorHandler.kt
@@ -31,16 +31,21 @@ internal class JettyKtorHandler(
     private val applicationProvider: () -> Application
 ) : Handler.Abstract() {
     private val environmentName = configuration.connectors.joinToString("-") { it.port.toString() }
-    private val queue: BlockingQueue<Runnable> = LinkedBlockingQueue()
+    private val queue: BlockingQueue<Runnable> = SynchronousQueue()
+    private val rejectedExecutionHandler =
+        // Always run on caller thread when pool is full or during shutdown.
+        // Unlike `ThreadPoolExecutor.CallerRunsPolicy` which discards tasks during shutdown,
+        // this ensures cancellation can propagate through the dispatcher.
+        RejectedExecutionHandler { r, _ -> r.run() }
     private val executor = ThreadPoolExecutor(
         configuration.callGroupSize,
         configuration.callGroupSize * 8,
         THREAD_KEEP_ALIVE_TIME,
         TimeUnit.MINUTES,
-        queue
-    ) { r ->
-        Thread(r, "ktor-jetty-$environmentName-${JettyKtorCounter.incrementAndGet()}")
-    }
+        queue,
+        { r -> Thread(r, "ktor-jetty-$environmentName-${JettyKtorCounter.incrementAndGet()}") },
+        rejectedExecutionHandler
+    )
     private val dispatcher = executor.asCoroutineDispatcher()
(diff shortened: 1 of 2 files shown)


ktor-server/ktor-server-jetty-jakarta/jvm/test/io/ktor/tests/server/jetty/jakarta/JettyEngineTest.kt
```diff
@@ -35,7 +35,12 @@ class JettyHttpServerJvmTest : HttpServerJvmTestSuite<JettyApplicationEngine, Je
 }
 
 class JettySustainabilityTest :
-    SustainabilityTestSuite<JettyApplicationEngine, JettyApplicationEngineBase.Configuration>(Jetty)
+    SustainabilityTestSuite<JettyApplicationEngine, JettyApplicationEngineBase.Configuration>(Jetty) {
+
+    override fun configure(configuration: JettyApplicationEngineBase.Configuration) {
+        configuration.callGroupSize = 6
```
amccague: I don't agree with making this change to satisfy the tests, this comment is to start a discussion thread. 
Without it the following tests will fail (when the `callGroupSize` is _lower_):
- `io.ktor.server.testing.suites.SustainabilityTestSuite#testBlockingDeadlock`
- `io.ktor.server.testing.suites.SustainabilityTestSuite#testBlockingConcurrency`

Which I now see the test I wrote was a similar like test.

The question is what to do about it, 
- Potentially the multiple of the max thread pool size could be increased substantially, and/or the base is increased from parallelism (x2-3), which I'm not opposed to because on single or dual core VMs this is very low.
- Or be `min()` of a value to satisfy these tests if the tests are setting the minimum client expectations.

amccague: Resolved this with a form of `ThreadPoolExecutor.CallerRunsPolicy` on the Executor - so if the thread pool is maxed, we fall back to running on the Jetty Caller threads. This makes the Ktor Jetty Thread Pool Executor an optimisation, falling back to how a Jetty application may normally be handling requests.
In an ideal world then the Ktor pool would always be offloading Jetty.

bjhham (approved):
Thank you for the contribution and detailed explanation!  The fix looks good to me 👍

PR #5269, opened 2026-01-07
Fixes for windows platform tests

bjhham:
This ought to get our Native windows tests back into working order:

https://ktor.teamcity.com/buildConfiguration/Ktor_KtorMatrixNativeWindowsX64

I disabled the failing `ConnectionTest` for the windows platform, because it is a pretty dubious use-case with an easy workaround.  I also bumped the timeout for one SSE test because the windows host has a lot of random delays.

### DIFF
--- a/ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/ConnectionTest.kt
+++ b/ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/ConnectionTest.kt
@@ -50,7 +50,7 @@ class ConnectionTest : ClientLoader() {
     }
 
     @Test
-    fun testInvalidHostInDefaultRequest() = clientTests {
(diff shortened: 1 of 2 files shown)


zibet27:
@bjhham I agree it's a weird use case, but shouldn't we maintain consistent behaviour across all clients?

bjhham:
@zibet27 this is true, but it's lower priority than getting our tests back in order.  I've logged a ticket to fix it https://youtrack.jetbrains.com/issue/KTOR-9222/WinHttp-client-failure-with-invalid-host

The test was introduced to validate a fix for the same inconsistency with some other engines, but it looks like we just missed the windows one because of our flaky tests here 🤦

zibet27:
@bjhham Looks like some tests still fail

bjhham:
@zibet27 they are failing, yes, but sometimes they pass! 😏

PR #5274, opened 2026-01-08
WebSockets. Added frame queues backpressure configuration.

zibet27:
**Subsystem**
WebSockets (Client + Server)

**Motivation**
[KTOR-6993](https://youtrack.jetbrains.com/issue/KTOR-6993) OOM caused by unlimited incoming websocket frames

**Solution**
- Added the configuration for incoming and outgoing frame queues, which is similar to `Channel(...)` function parameters.
- Different WS implementations had different configurations, which led to different behaviours. Default values are kept.
- The testing is not straightforward because we need to emulate backpressure on these queues. Any advice here?

**Updates**
- Changed the data class to DSL.
- All configuration classes are public because we need them in other KTOR modules with internal constructors, so they can be created using DSL only.
- The default IO frame queues are unlimited across all engines
- Callback-based WS implementations don't allow a bounded incoming queue with suspension.
- Tests should be more reliable


ktor-client/ktor-client-core/js/src/io/ktor/client/plugins/websocket/JsWebSocketSession.kt
```diff
@@ -16,11 +16,13 @@ import kotlin.coroutines.*
...
     override val coroutineContext: CoroutineContext,
-    private val websocket: WebSocket
+    private val websocket: WebSocket,
+    incomingFramesConfig: ChannelConfig<Frame>,
+    outgoingFramesConfig: ChannelConfig<Frame>
 ) : DefaultWebSocketSession {
     private val _closeReason: CompletableDeferred<CloseReason> = CompletableDeferred()
-    private val _incoming: Channel<Frame> = Channel(Channel.UNLIMITED)
-    private val _outgoing: Channel<Frame> = Channel(Channel.UNLIMITED)
+    private val _incoming: Channel<Frame> = incomingFramesConfig.toChannel()
```
whyoleg: It's an **error** to use anything apart from `UNLIMITED` for `incoming` frames here, as the underlying browser API is callback-based and therefore uses `trySend`, as shown in the file below.
Using anything other than `UNLIMITED` will cause frame drops in the event of a buffer overflow.

Same in some other implementations in other engines: wasmJs, curl, darwin, windows (overall, in any implementation which uses `trySend`)

I don't have a better suggestion here, but maybe there are some underlying platform APIs that will allow us to control backpressure.

ktor-shared/ktor-websockets/common/src/io/ktor/websocket/ChannelConfig.kt
```diff
@@ -0,0 +1,67 @@
...
+    public val capacity: Int,
+    public val onBufferOverflow: BufferOverflow,
+    public val onUndeliveredElement: ((E) -> Unit)? = null
+) {
+    /**
+     * Creates a [kotlinx.coroutines.channels.Channel] using this configuration.
+     *
+     * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.utils.io.ChannelConfig.toChannel)
+     */
+    public fun <T : E> toChannel(): Channel<T> {
```
whyoleg: Minor: calling this function `toChannel` feels a bit strange, as this is more of a factory and can be called multiple times, creating different channels. So it might make sense to call it `createChannel`.

zibet27:
@whyoleg, thank you for the review! Yes, using `trySend` is an error, and I didn't change it everywhere.
Currently, we have some queues with invalid default behaviour. Like in `RawWebSocketCommon`, which has a small buffer and uses `trySend`.
Additionally, when I was debugging a flaky test, I discovered that the connection could become stuck if a `Frame.Close` is dropped. We should at least document it. I would suggest making buffers unlimited by default for all engines, but it's a breaking change. @bjhham

zibet27:
> One thing to note - be sure to enumerate the nuances for configuration in the different engines and forward this info to @vnikolova for documentation ℹ️

@bjhham, should users rely on these values?

ktor-client/ktor-client-core/js/src/io/ktor/client/plugins/websocket/JsWebSocketSession.kt
bjhham: There is a [WebSocketStream](https://developer.mozilla.org/en-US/docs/Web/API/WebSocketStream) API now we could migrate to for replacing callbacks with promises.  This might be a sizeable chunk of work though, so we might want to just include validation to enforce unlimited channels in the configuration for JS for now and create a new ticket to address support for web clients (your call, @zibet27)

bjhham:
@zibet27 wrt documentation, I was thinking just a table with each engine, whether or not back-pressure is supported, what the default is, what is accepted.  For engines where a limited channel size is not viable (like with the JS engine call back situation @whyoleg pointed out, we should include validation and ensure it's documented).

I was thinking for testing, we should include one for the suspension scenario too, so same as the dropping scenario with the limited size but it should be able to read all frames through means of deferred reads / suspension.

whyoleg:
> Yes, using trySend is an error, and I didn't change it everywhere.

TBH, it's not really an error to use `trySend` here. We just have no other choice if the underlying platform API is synchronous callback-based :)
The only way to remove usage of `trySend` is to use `send`, for example, like this:
```kotlin
val frame = ...
if(!channel.trySend(frame).isSuccess) {
  scope.launch(UNDISPATCHED) { channel.send(frame) }
}
```
But, this is not a solution, as it will also create some kind of "queue", but via coroutines Job relationships, and so can cause OOM.

So we **should** continue to use `UNLIMITED` channel + `trySend` in those engines. Additionally, some platform implementations can support various types of backpressure settings.

All in all, I don't think that a general configuration on the plugin side is possible here. IMO, merging PR in its current state might cause more issues than it solves.

ktor-client/ktor-client-core/js/src/io/ktor/client/plugins/websocket/JsWebSocketSession.kt
whyoleg: `WebSocketStream` is an experimental API, and Firefox and Safari **do not currently support it**. So, it's not a solution here, but rather an optimization. :)

My 5 cents: I would suggest implementing support for `WebSocketStream` via a separate issue/PR with the ability to fallback to the current implementation.

bjhham: Ah yes, I didn't check the browser support, definitely not an option atm.  Looks like we should just enforce `UNLIMITED` here and keep an eye on `WebSocketStream`.

zibet27:
> So we **should** continue to use `UNLIMITED` channel + `trySend` in those engines. Additionally, some platform implementations can support various types of backpressure settings.

Hmm, actually, LIMITED BUFFER + DROP on buffer overflow is also a "valid" behaviour here. 
I would disallow a limited buffer + SUSPEND for the incoming frames queue on all platforms because cause OOM even faster than an unlimited one.

> IMO, merging PR in its current state might cause more issues than it solves.
Agree, we should discuss the behaviour we want to achieve.

For the incoming queue, I see 2 "valid" options:
- unlimited buffer, can cause OOM.
- limited buffer + frame drops.

For the outgoing queue, any config is "valid" until the user knows the current config. 

I think we also need route-scoped configuration here.

zibet27:
> For engines where a limited channel size is not viable (like with the JS engine call back situation @whyoleg pointed out, we should include validation and ensure it's documented).

I think the limited size + suspend doesn't make much sense for any engine's incoming frames, because it'll cause OOM even faster than an unlimited buffer.

Hmm, is it a breaking change to make the buffer unlimited on all platforms? The config wasn't exposed, and with unlimited ones, we won't silently lose any frames.

bjhham:
> For the incoming queue, I see 2 "valid" options:
>    unlimited buffer, can cause OOM.
>    limited buffer + frame drops.

Ah yes, this is true, both are valid.  I feel like unlimited / potential OOM is a better default for feedback, since frame drops can lead to sneaky data loss which is much harder to debug.

> For the outgoing queue, any config is "valid" until the user knows the current config.

👍 

> I think we also need route-scoped configuration here.

Oh, that could be a nice feature!  I wouldn't say it's very high priority though, since it would require changing the routing API, and people generally don't have many websocket endpoints per service.  Maybe it would make sense to convert the websockets plugin to a route-scoped one for different low-level configs like this down the road.

> I think the limited size + suspend doesn't make much sense for any engine's incoming frames, because it'll cause OOM even faster than an unlimited buffer.

Maybe I'm missing something here, but I thought this was how the back pressure was supposed to work (i.e., waiting to read from the socket)??

zibet27:
> Maybe I'm missing something here, but I thought this was how the back pressure was supposed to work (i.e., waiting to read from the socket)??

Afaik, we don't use such a low-level API to read from a socket, but set callbacks for the next frame. When we receive a frame and call suspending `send`, we need even more memory than just having it in a queue because coroutines overhead.
As @whyoleg mentioned 
> But, this is not a solution, as it will also create some kind of "queue", but via coroutines Job relationships, and so can cause OOM.

In my opinion, limited buffer + suspend is useful for the `outgoing` queue to keep the frames in order without busy waiting.

bjhham:
> Afaik, we don't use such a low-level API to read from a socket, but set callbacks for the next frame.

Ah, I thought this was only the case for a few of the engines.

ktor-client/ktor-client-core/js/src/io/ktor/client/plugins/websocket/JsWebSocketSession.kt
whyoleg: Worth filing an issue for the future, though.

ktor-shared/ktor-websockets/common/src/io/ktor/websocket/ChannelConfig.kt
```diff
@@ -0,0 +1,67 @@
...
+ * A configuration for a [kotlinx.coroutines.channels.Channel].
+ * The implementation is not necessarily the default one, so don't rely on it.
+ *
+ * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.utils.io.ChannelConfig)
+ *
+ * @property capacity: channel capacity.
+ * @property onBufferOverflow: strategy to use on buffer overflow.
+ * @property onUndeliveredElement: handler for undelivered elements.
+ */
+public data class ChannelConfig<in E>(
```
whyoleg: BTW: `data class` in public API is not a good idea :)

whyoleg:
I'm curious, though, what is the use case for `limited buffer + frame drops`? In this case, we might accidentally drop `ping`/`pong`/`close` frames, resulting in an incorrect outcome.

If there are some specific use-cases for that, I would then vote for having a DSL configuration insted of `class` with properties, which will only allow to configure specific configurations, e.g.:
- `unlimited()` - uses unlimited + suspend
- `buffered(size = 8, ...)` - parameters are TBD here

ktor-shared/ktor-websockets/common/src/io/ktor/websocket/ChannelConfig.kt
zibet27: Agree, the dsl looks better, but what's wrong with `data class`?

whyoleg: Further evolution of it is not that easy :)
One of the recent examples in Dokka: https://github.com/Kotlin/dokka/pull/4398/changes
- `dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/model/doc/TagWrapper.kt` - changes are here
  - just to add one field, we need to add an additional deprecated constructor and copy with carefully selected signatures to save binary compatibility
- `dokka-subprojects/core/api/dokka-core.api` - contains binary compatible diff of ABI - it's not easy to review it, and it's easy to miss something

[One more scary example of similar change](https://github.com/Kotlin/dokka/blob/d67cf0cd96b02999192ab4c8fc229195aba4bff7/dokka-subprojects/core/src/main/kotlin/org/jetbrains/dokka/model/Documentable.kt#L599-L661) :)

And, here is something more structured: https://jakewharton.com/public-api-challenges-in-kotlin/

zibet27:
> I'm curious, though, what is the use case for `limited buffer + frame drops`? In this case, we might accidentally drop `ping`/`pong`/`close` frames, resulting in an incorrect outcome.

The use-case can be the detection of some bugs when the client stops reading incoming frames. [KTOR-6993](https://youtrack.jetbrains.com/issue/KTOR-6993) OOM caused by unlimited incoming websocket frames.
The user can set the `onUndeliveredElement` callback and limit the buffer to a big(but not unlimited) capacity to close the connection when this happens.

I'm not sure how not to lose important frames, though.

ktor-client/ktor-client-core/js/src/io/ktor/client/plugins/websocket/JsWebSocketSession.kt
zibet27: [KTOR-9228](https://youtrack.jetbrains.com/issue/KTOR-9228) WebSocketStream API support

whyoleg:
> The user can set the onUndeliveredElement callback and limit the buffer to a big(but not unlimited) capacity to close the connection when this happens.

Oh, I see, that's a nice use case. It might make sense to collect other use cases, if they exist.
If it's the only one, then maybe we can make only this behavior configurable `buffered` configuration, which doesn't support backpressure:
- `unlimited()` - uses unlimited + suspend
- `buffered(size = 8)` - buffered + suspend. Throw an error on configuration if the engine doesn't support this (just buffered channel under the hood)
- `buffered(size = 8, closeConnectionOnBufferOverflow = true)` - buffered + non-suspend. Throw an error if the buffer overflows, not during configuration but at runtime, to prevent this kind of issue. (under the hood, it will set up `onUndeliveredElement` + `DROP_LATEST` strategy)

I mean, giving the user the ability to accidentally configure dropping frames is not a good idea.

WDYT?

zibet27:
> WDYT?

Agree with these options. I would introduce Ktor's BufferOverflow Strategies (Suspend, CancelConnection) instead of `kotlinx.coroutines.channels.BufferOverflow`. 
We may also want to specify the size of the buffer content (for the outgoing queue), not only the buffer length.
That would need more effort. Do we actually need it?

bjhham:
> We may also want to specify the size of the buffer content (for the outgoing queue), not only the buffer length.
> That would need more effort. Do we actually need it?

Would be handy for low-memory deployments, but might be overkill for this fix.

ktor-client/ktor-client-darwin/darwin/src/io/ktor/client/engine/darwin/KtorNSURLSessionDelegate.kt
```diff
@@ -5,9 +5,14 @@
 package io.ktor.client.engine.darwin
 
 import io.ktor.client.engine.darwin.internal.*
+import io.ktor.client.plugins.websocket.WEBSOCKETS_KEY
```
e5l, line 8: Please consider checking if we need to update darwin-legacy engine as well: https://github.com/ktorio/ktor/tree/main/ktor-client/ktor-client-darwin-legacy

ktor-client/ktor-client-okhttp/jvm/src/io/ktor/client/engine/okhttp/OkHttpWebsocketSession.kt
```diff
@@ -98,13 +110,18 @@ internal class OkHttpWebsocketSession(
...
+        }
     }
 
     override fun onMessage(webSocket: WebSocket, text: String) {
-        super.onMessage(webSocket, text)
-        _incoming.trySendBlocking(Frame.Text(true, text.toByteArray()))
+        // start immediately to keep the order
+        launch(start = CoroutineStart.UNDISPATCHED) {
+            super.onMessage(webSocket, text)
+            _incoming.send(Frame.Text(true, text.toByteArray()))
```
e5l: Could you check if frames can be reordered under pressure?

ktor-shared/ktor-websockets/common/src/io/ktor/websocket/ChannelConfig.kt
e5l: it also looks more like internal class than a public api

ktor-client/ktor-client-okhttp/jvm/src/io/ktor/client/engine/okhttp/OkHttpWebsocketSession.kt
zibet27: Yes, if we call `launch` without `CoroutineStart.UNDISPATCHED`

ktor-client/ktor-client-winhttp/windows/src/io/ktor/client/engine/winhttp/internal/WinHttpRequest.kt
```diff
@@ -208,14 +213,19 @@ internal class WinHttpRequest(
...
-            closeRequest()
-        }
+        val wsConfig = attributes[WEBSOCKETS_KEY]
+
+        return WinHttpWebSocket(
+            hWebsocket,
+            connect,
+            callContext,
+            wsConfig.ioChannelsConfig
+        ).also { closeRequest() }
```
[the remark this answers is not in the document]
zibet27: No, WinHttp uses `receiveNextFrame`, which pulls a new frame every time. 
So, we can have only one suspended frame at a time.

ktor-server/ktor-server-plugins/ktor-server-websockets/jvm/test/io/ktor/tests/websocket/WebSocketBackpressureTest.kt
```diff
@@ -0,0 +1,137 @@
...
+                outgoing = bounded(capacity = 1, onOverflow = ChannelOverflow.CLOSE)
+            }
+        }
+        routing {
+            expectChannelOverflow {
+                createSession("/receive-all") {
+                    for (i in 1..FRAMES_COUNT) send("message $i")
+                }
+            }
+        }
```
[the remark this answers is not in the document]
zibet27: It throws

ktor-shared/ktor-websockets/jvm/src/io/ktor/websocket/RawWebSocketJvm.kt
```diff
@@ -26,27 +26,61 @@ import kotlin.properties.*
...
+    coroutineContext: CoroutineContext,
+    ioChannelsConfig: IOChannelsConfig
+): WebSocketSession = RawWebSocketJvm(
+    input,
+    output,
+    maxFrameSize,
+    masking,
+    coroutineContext,
+    ioChannelsConfig,
+)
```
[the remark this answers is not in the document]
zibet27: No, they don't.

ktor-client/ktor-client-core/js/src/io/ktor/client/plugins/websocket/JsWebSocketSession.kt
```diff
@@ -64,7 +66,9 @@ internal class JsWebSocketSession(
                     }
                 }
 
-                _incoming.trySend(frame)
+                launch(start = CoroutineStart.UNDISPATCHED) {
+                    _incoming.send(frame)
+                }
```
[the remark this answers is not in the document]
zibet27: Kotlin docs on `BufferOverflow.DROP_LATEST`:

> Leave the buffer unchanged on overflow, dropping the value that we were going to add, do not suspend.
This option can be used in rare advanced scenarios where all elements that are expected to enter the buffer are equal, so it is not important which of them get thrown away.

ktor-client/ktor-client-darwin/darwin/src/io/ktor/client/engine/darwin/KtorNSURLSessionDelegate.kt
zibet27: `DarwinLegacy` doesn't support WebSockets :(

ktor-client/ktor-client-java/jvm/src/io/ktor/client/engine/java/JavaHttpWebSocket.kt
```diff
@@ -189,20 +197,23 @@ internal class JavaHttpWebSocket(
...
 
-    override fun onPong(webSocket: WebSocket, message: ByteBuffer): CompletionStage<*> = async {
-        _incoming.trySend(Frame.Pong(message)).isSuccess
-        webSocket.request(1)
-    }.asCompletableFuture()
+    override fun onPong(webSocket: WebSocket, message: ByteBuffer): CompletionStage<*> =
+        async(start = CoroutineStart.UNDISPATCHED) {
+            _incoming.send(Frame.Pong(message))
+            webSocket.request(1)
+        }.asCompletableFuture()
```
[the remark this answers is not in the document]
zibet27: Java WebSocket documentation says receive methods (like `onText`, `onBuffer`) won't be called until  `webSocket.request(...)` is invoked. So, there will be no coroutine queue.

ktor-client/ktor-client-curl/desktop/src/io/ktor/client/engine/curl/internal/CurlWebSocketResponseBody.kt
```diff
@@ -66,7 +72,7 @@ internal class CurlWebSocketResponseBody(
             }
 
             (flags and CURLWS_PONG != 0) -> {
-                _incoming.trySend(Frame.Ping(data = buffer))
+                _incoming.trySend(Frame.Pong(data = buffer))
             }
```
bjhham: Oh, good catch 👀

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/websocket/WebSockets.kt
```diff
@@ -143,6 +150,26 @@ public class WebSockets internal constructor(
...
+         *
+         * ```kotlin
+         * ioChannels {
+         *     incoming = unlimited()
+         *     outgoing = bounded(capacity = 512, onOverflow = ChannelOverflow.SUSPEND)
+         * }
+         * ```
+         * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.client.plugins.websocket.WebSockets.Config.ioChannels)
+         */
+        public fun ioChannels(block: IOChannelsConfigBuilder.() -> Unit) {
```
whyoleg: I would suggest renaming the function to `channels` or `buffering`, and renaming `IOChannelsConfigBuilder` to `WebSocketChannelsConfig` to align with the `extensions` naming. You also probably don't need a separation between `builder` and `config`, as is the case with most Ktor configurations.
Overall, `IOChannels` sounds strange to me, especially given that it is web socket specific.

ktor-client/ktor-client-core/js/src/io/ktor/client/plugins/websocket/JsWebSocketSession.kt
```diff
@@ -138,6 +140,9 @@ internal class JsWebSocketSession(
     @OptIn(InternalAPI::class)
     override fun start(negotiatedExtensions: List<WebSocketExtension<*>>) {
         require(negotiatedExtensions.isEmpty()) { "Extensions are not supported." }
+        if (ioChannelsConfig.incoming.canSuspend) {
```
whyoleg: Curious: Why not perform this check on session creation?

ktor-client/ktor-client-core/wasmJs/src/io/ktor/client/plugins/websocket/JsWebSocketSession.kt
```diff
@@ -153,6 +160,9 @@ internal class JsWebSocketSession(
     @OptIn(InternalAPI::class)
     override fun start(negotiatedExtensions: List<WebSocketExtension<*>>) {
         require(negotiatedExtensions.isEmpty()) { "Extensions are not supported." }
+        if (ioChannelsConfig.incoming.canSuspend) {
```
whyoleg: same here

ktor-client/ktor-client-darwin/darwin/src/io/ktor/client/engine/darwin/internal/DarwinWebsocketSession.kt
```diff
@@ -141,7 +143,9 @@ internal class DarwinWebsocketSession(
                             cancel("Error receiving pong", DarwinHttpRequestException(error))
                             return@sendPingWithPongReceiveHandler
                         }
-                        _incoming.trySend(Frame.Pong(payload))
+                        launch(start = CoroutineStart.UNDISPATCHED) {
```
whyoleg: minor: it might make sense to first try to call `trySend` and only if it failed launch a coroutine - so on the fast path, there will be no coroutine allocation.

ktor-client/ktor-client-java/jvm/src/io/ktor/client/engine/java/JavaHttpWebSocket.kt
```diff
@@ -189,20 +197,23 @@ internal class JavaHttpWebSocket(
         webSocket.request(1)
     }
 
-    override fun onText(webSocket: WebSocket, data: CharSequence, last: Boolean): CompletionStage<*> = async {
-        _incoming.trySend(Frame.Text(last, data.toString().toByteArray())).isSuccess
-        webSocket.request(1)
-    }.asCompletableFuture()
+    override fun onText(webSocket: WebSocket, data: CharSequence, last: Boolean): CompletionStage<*> =
+        async(start = CoroutineStart.UNDISPATCHED) {
+            _incoming.send(Frame.Text(last, data.toString().toByteArray()))
```
whyoleg: similar to the comment above: you can try to use `trySend` first.

ktor-server/ktor-server-plugins/ktor-server-websockets/common/src/io/ktor/server/websocket/WebSockets.kt
```diff
@@ -112,6 +117,26 @@ public class WebSockets private constructor(
...
+         *
+         * ```kotlin
+         * ioChannels {
+         *     incoming = unlimited()
+         *     outgoing = bounded(capacity = 512, onOverflow = ChannelOverflow.SUSPEND)
+         * }
+         * ```
+         * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.server.websocket.WebSockets.WebSocketOptions.ioChannels)
+         */
+        public fun ioChannels(block: IOChannelsConfigBuilder.() -> Unit) {
```
whyoleg: same regarding naming

ktor-client/ktor-client-core/js/src/io/ktor/client/plugins/websocket/JsWebSocketSession.kt
zibet27: Because the thrown exception won't be propagated to the "user" coroutine.

whyoleg: Hm, then you need to do the same for all other engines?
E.g., in okhttp it's [checked on creation](https://github.com/ktorio/ktor/pull/5274/changes#diff-790087b9ea4a1ff119bba9e184da31db554f09bf2d6f766151a01d671fdf0414R50).

While it's 100% required, it would be nice to have one place to check the configuration.

zibet27: It's engine-specific, so I think it's the right place to check.
Yes, it works in OkHttp, but hangs in JS impl. 
Afaik, they work differently with coroutine contexts.
I'll try to figure out the difference.

zibet27: @whyoleg, the cause is even more interesting; the thrown exception was actually propagated to the "user" coroutine. 
But if we throw the exception before adding a callback on the current context close (to close js ws connection), the underlying WS connection will hang in memory until another peer decides to close it. Also, trying to close not yet connected connection results in an error. 
Fix: https://github.com/ktorio/ktor/pull/5274/files#diff-945a1718963cfe456236ce7744c0562f93a3d4bc55a67a552b739a396ced1746R90
The same fix was applied for the wasm impl.

ktor-client/ktor-client-winhttp/windows/src/io/ktor/client/engine/winhttp/internal/WinHttpWebSocket.kt
```diff
@@ -28,20 +29,21 @@ private object WinHttpWebSocketBuffer {
...
     private val closed = atomic(false)
     private val socketJob = Job(callContext[Job])
 
-    private val _incoming = Channel<Frame>(Channel.UNLIMITED)
-    private val _outgoing = Channel<Frame>(Channel.UNLIMITED)
-
     override val coroutineContext: CoroutineContext = callContext + socketJob
+
+    private val _incoming = Channel.from<Frame>(channelsConfig.incoming)
+    private val _outgoing = Channel.from<Frame>(channelsConfig.outgoing)
```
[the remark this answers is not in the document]
zibet27: It uses a custom `trySend` which will close the channel on overflow.

ktor-shared/ktor-websockets/common/src/io/ktor/websocket/WebSocketChannelsConfig.kt
```diff
@@ -0,0 +1,172 @@
...
+    public companion object {
+        /**
+         * A configuration with unlimited buffer.
+         *
+         * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.websocket.ChannelConfig.Companion.UNLIMITED)
+         */
+        public val UNLIMITED: ChannelConfig =
+            ChannelConfig(capacity = Channel.UNLIMITED, onOverflow = ChannelOverflow.SUSPEND)
+    }
+}
```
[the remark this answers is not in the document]
zibet27: It's correct; RENDEZVOUS and BUFFERED channels can be suspended.

ktor-client/ktor-client-core/wasmJs/src/io/ktor/client/plugins/websocket/JsWebSocketSession.kt
```diff
@@ -132,21 +158,16 @@ internal class JsWebSocketSession(
...
-        coroutineContext[Job]?.invokeOnCompletion { cause ->
-            if (cause == null) {
-                websocket.close()
-            } else {
-                // We cannot use INTERNAL_ERROR similarly to other WebSocketSession implementations here
-                // as sending it is not supported by browsers.
-                websocket.close(CloseReason.Codes.NORMAL.code, "Client failed")
-            }
+        if (channelsConfig.incoming.canSuspend) {
+            throw IllegalArgumentException("SUSPEND overflow strategy for incoming channel is not supported.")
```
osipxd: Minor.

```suggestion
        require(!channelsConfig.incoming.canSuspend) {
            "SUSPEND overflow strategy for incoming channel is not supported."
```

ktor-shared/ktor-websockets/jvm/src/io/ktor/websocket/WebSocketReader.kt
```diff
@@ -25,37 +24,42 @@ import kotlin.coroutines.*
...
  * @param maxFrameSize maximum frame size that could be read
  */
+@OptIn(InternalAPI::class)
 public class WebSocketReader(
     private val byteChannel: ByteReadChannel,
     override val coroutineContext: CoroutineContext,
     public var maxFrameSize: Long,
-    pool: ObjectPool<ByteBuffer> = KtorDefaultPool
+    pool: ObjectPool<ByteBuffer> = KtorDefaultPool,
+    queueConfig: ChannelConfig = ChannelConfig.UNLIMITED
```
osipxd, line 33: Could you also add a second constructor with the old signature and `DeprecationLevel.HIDDEN` to keep it binary compatible?

ktor-shared/ktor-websockets/jvm/src/io/ktor/websocket/WebSocketWriter.kt
```diff
@@ -15,21 +15,24 @@ import kotlin.coroutines.*
...
+ * @param queueConfig: configuration for the internal [Frame] queue.
  */
+@OptIn(InternalAPI::class)
 public class WebSocketWriter(
     private val writeChannel: ByteWriteChannel,
     override val coroutineContext: CoroutineContext,
     public var masking: Boolean = false,
-    public val pool: ObjectPool<ByteBuffer> = KtorDefaultPool
+    public val pool: ObjectPool<ByteBuffer> = KtorDefaultPool,
+    queueConfig: ChannelConfig = ChannelConfig.UNLIMITED,
```
osipxd, line 32: And here

ktor-shared/ktor-websockets/posix/src/io/ktor/websocket/RawWebSocket.kt
```diff
@@ -26,4 +26,28 @@ public actual fun RawWebSocket(
...
+ * @param channelsConfig is a [WebSocketChannelsConfig] for the incoming and outgoing [Frame] queues
+ */
+@Suppress("FunctionName")
+public actual fun RawWebSocket(
+    input: ByteReadChannel,
+    output: ByteWriteChannel,
+    maxFrameSize: Long,
+    masking: Boolean,
+    coroutineContext: CoroutineContext,
+    channelsConfig: WebSocketChannelsConfig,
```
osipxd, line 30: Theoretically you could add a default value here and then deprecate the old function with HIDDEN

ktor-shared/ktor-websockets/common/src/io/ktor/websocket/RawWebSocketCommon.kt
```diff
@@ -36,22 +35,45 @@ public expect fun RawWebSocket(
...
+ * @param channelsConfig is a [WebSocketChannelsConfig] for the incoming and outgoing [Frame] queues
+ */
+@Suppress("FunctionName")
+public expect fun RawWebSocket(
+    input: ByteReadChannel,
+    output: ByteWriteChannel,
+    maxFrameSize: Long = Int.MAX_VALUE.toLong(),
+    masking: Boolean = false,
+    coroutineContext: CoroutineContext,
+    channelsConfig: WebSocketChannelsConfig,
```
osipxd: Can we specify the old function to call the new one? So we don't have to implement actual for it?

ktor-shared/ktor-websockets/common/src/io/ktor/websocket/DefaultWebSocketSession.kt
```diff
@@ -74,37 +74,86 @@ public interface DefaultWebSocketSession : WebSocketSession {
...
+ * @param timeoutMillis timeout for pings.
+ * @param channelsConfig configuration for the I/O frame channels.
+ *
  * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.websocket.DefaultWebSocketSession)
  */
 public fun DefaultWebSocketSession(
     session: WebSocketSession,
     pingIntervalMillis: Long = PINGER_DISABLED,
     timeoutMillis: Long = 15_000L,
+    channelsConfig: WebSocketChannelsConfig,
```
osipxd: It would've been be great if the new function had `Duration` parameters

ktor-server/ktor-server-plugins/ktor-server-websockets/jvm/test/io/ktor/tests/websocket/BackpressureTest.kt
```diff
@@ -0,0 +1,137 @@
...
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+import kotlin.test.fail
+import kotlin.time.Duration.Companion.seconds
+import io.ktor.client.plugins.websocket.WebSockets as ClientWebSockets
+
+private const val FRAMES_COUNT = 100
+
+abstract class BackpressureTest {
```
osipxd: Maybe name it `WebsocketBackpressureTest`?

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/WebSocketBackpressureTest.kt
```diff
@@ -0,0 +1,144 @@
...
+private val CALLBACK_BASED_WS_CLIENTS = arrayOf("OkHttp", "JS", "Curl")
+
+class WebSocketBackpressureTest : ClientLoader() {
+
+    val Throwable?.isChannelOverflow: Boolean
+        get() = this is ChannelOverflowException || this?.cause is ChannelOverflowException
+
+    @Test
+    fun `test IO frame channels suspension`() =
+        clientTests(except(ENGINES_WITHOUT_WS, *CALLBACK_BASED_WS_CLIENTS)) {
```
osipxd: Minor. Alternatively, without the spread operator
```suggestion
        clientTests(except(ENGINES_WITHOUT_WS + CALLBACK_BASED_WS_CLIENTS)) {
```
`CALLBACK_BASED_WS_CLIENTS` should be a list

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/WebSocketBackpressureTest.kt
```diff
@@ -0,0 +1,144 @@
...
+                        incoming = bounded(capacity = 1, onOverflow = ChannelOverflow.SUSPEND)
+                        outgoing = bounded(capacity = 1, onOverflow = ChannelOverflow.SUSPEND)
+                    }
+                }
+            }
+
+            test { client ->
+                client.webSocket("$TEST_WEBSOCKET_SERVER/websockets/echo") {
+                    val sendJob = launch {
+                        for (i in 1..FRAMES_COUNT) send("message $i")
```
osipxd: Minor
```suggestion
                        repeat(FRAMES_COUNT) { i -> send("message $i") }
```

osipxd (commented):
Great work! I've found some binary compatibility issues

ktor-client/ktor-client-webrtc/ios/src/io/ktor/client/webrtc/Utils.kt
```diff
@@ -112,7 +112,7 @@ internal fun NSData.toByteArray(): ByteArray = ByteArray(length.toInt()).apply {
 }
 
 @OptIn(BetaInteropApi::class)
-internal fun ByteArray.toNSData() = memScoped<NSData> {
+internal fun ByteArray.toNSData(): NSData = memScoped<NSData> {
     return NSData.create(bytes = this@toNSData.toCValues().ptr, length = size.toULong())
```
osipxd: Is `return` required here?

ktor-shared/ktor-websockets/common/src/io/ktor/websocket/DefaultWebSocketSession.kt
```diff
@@ -82,29 +87,74 @@ public fun DefaultWebSocketSession(
...
+ * @param pingIntervalMillis interval between pings or [PINGER_DISABLED] to disable.
+ * @param timeout timeout for pings.
+ * @param channelsConfig configuration for the I/O frame channels.
+ *
+ * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.websocket.DefaultWebSocketSession)
+ */
+public fun DefaultWebSocketSession(
+    session: WebSocketSession,
+    pingIntervalMillis: Long = PINGER_DISABLED,
+    timeout: Duration = 15.seconds,
```
osipxd: Oh, I see the problem that because of the `PINGER_DISABLED` constant we can't use `Duration` for both parameters. Let's revert it back to `Long` then

ktor-client/ktor-client-webrtc/ios/src/io/ktor/client/webrtc/Utils.kt
osipxd: Btw, I've opened a PR to fix this issue
https://github.com/ktorio/ktor/pull/5314

zibet27: @osipxd after sync with main, which updated the Kotlin version, `./gradlew updateLegacyAbi` fails because of this return type.
Can you try to run it on the main branch on your machine?

osipxd: Fixed in the main

zibet27: I merged main into this branch, so it now doesn't include the fix

PR #5287, opened 2026-01-12
KTOR-8642 Enable segment L2 cache for kotlinx-io by default for Android

bjhham:
**Subsystem**
Client, Android

**Motivation**
[KTOR-8642](https://youtrack.jetbrains.com/issue/KTOR-8642) Excessive memory allocations while writing bytes into write channel of TCP/IP socket

**Solution**
We have a memory allocation problem with Android due to the default configuration for kotlinx-io on this platform.  This will automatically assign the system property responsible for enabling the L2 cache for segments if not already set.  This is the only way I could think of which would handle most cases without requiring any code changes.

### DIFF
--- a/ktor-client/ktor-client-android/jvm/test/io/ktor/client/engine/android/AndroidHttpClientTest.kt
+++ b/ktor-client/ktor-client-android/jvm/test/io/ktor/client/engine/android/AndroidHttpClientTest.kt
@@ -19,5 +19,4 @@ class AndroidHttpClientTest : HttpClientTest(Android) {
         val segmentPoolSize = System.getProperty("kotlinx.io.pool.size.bytes")
         assertEquals("2097152", segmentPoolSize, "Default segment pool should be assigned")
     }
-
 }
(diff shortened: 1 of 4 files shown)


osipxd (commented):
Nice!
I have two questions:
1. Should this property also be set for other engines that can be used on Android? CIO and OkHttp
2. What if a device doesn't have 2MB of L2 cache?

bjhham:
> Should this property also be set for other engines that can be used on Android? CIO and OkHttp

I was thinking it'd be better to avoid letting this little workaround leak over multiple clients, though the default for JVM is the 2MB here, so it at least wouldn't change anything functionally...  Maybe it'd be wise to copy it over.

> What if a device doesn't have 2MB of L2 cache?

I'll need to check on that... They would have a fix to change the property themselves and we wouldn't override, but I'm not sure what would happen if we go outside the limit on this property.  Maybe there's a way to get the process allocation and use a smarter value when space is scarce.

bjhham:
I've added some extra checks and included the assignment in OkHttp and CIO, so it should be good to go.  I included a test, but it may be mostly useless since it appears we don't have any test suites that run on Android...

PR #5293, opened 2026-01-13
KTOR-9195 Add option for filtering body with custom logic

bjhham:
**Subsystem**
Client, Logging

**Motivation**
[KTOR-9195](https://youtrack.jetbrains.com/issue/KTOR-9195) Client logging plugin improvements

Request from AI platform team for better control over how request/response bodies are logged in the client.

**Solution**
I introduced an override for the default request body logic which currently just avoids logging when binary data is encountered.

Note, this ignores multi-part bodies atm because it requires different processing.

### DIFF
--- a/ktor-client/ktor-client-plugins/ktor-client-logging/common/src/io/ktor/client/plugins/logging/LogBodyFilter.kt
+++ b/ktor-client/ktor-client-plugins/ktor-client-logging/common/src/io/ktor/client/plugins/logging/LogBodyFilter.kt
@@ -1,0 +1,221 @@
+/*
+ * Copyright 2014-2026 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.client.plugins.logging
+
+import io.ktor.http.*
+import io.ktor.utils.io.*
+import io.ktor.utils.io.charsets.*
+import io.ktor.utils.io.core.*
+import kotlinx.io.Buffer
+
+/**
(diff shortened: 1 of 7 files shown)


ktor-client/ktor-client-plugins/ktor-client-logging/common/src/io/ktor/client/plugins/logging/LogBodyFilter.kt
```diff
@@ -0,0 +1,221 @@
...
+                contentType?.charset() ?: Charsets.UTF_8
+            )
+        }
+    }
+}
+
+/**
+ * Represents the result of a body filtering process, which can determine the state of the body after processing.
+ */
+public sealed interface BodyFilterResult {
```
Stexxe, line 170: Why do we need this interface if all that is asked is to:
> enable/disable response body logging at runtime using some configuration logic.

In terms of the behavior, disabling the request and response bodies logging at runtime should be equal to setting the level to LogLevel.HEADERS and lower.

bjhham: I could have used nullable content for the result, but we've got logic for logging the reason for skipping content based on whether it's encoded or binary, etc., so the sealed type was the only way to go.  This will allow also for modifying what's logged so you do things like filter out sensitive data, or you can apply some custom logic on which endpoints you want to log.

PR #5300, opened 2026-01-14
KTOR-8316 Naming and annotations for OpenAPI runtime API

bjhham:
**Subsystem**
Server, OpenAPI routing API

**Motivation**
The current module name and function name "annotate" was not acceptable.

**Solution**
I renamed the module and function name.  Also added experimental annotation to the function, so we can defer the decision on the final API.

Naming changes:
- ktor-server-routing-annotate --> ktor-server-routing-openapi
- `annotate {}` -> `describe {}`


gradle/artifacts/publishJvmAndCommonPublications.txt
```diff
@@ -645,13 +645,13 @@ io.ktor:ktor-server-resources/.jar
...
-io.ktor:ktor-server-routing-annotate/javadoc.jar
-io.ktor:ktor-server-routing-annotate/kotlin-tooling-metadata.json
-io.ktor:ktor-server-routing-annotate/sources.jar
+io.ktor:ktor-server-routing-openapi-jvm/.jar
+io.ktor:ktor-server-routing-openapi-jvm/javadoc.jar
+io.ktor:ktor-server-routing-openapi-jvm/sources.jar
+io.ktor:ktor-server-routing-openapi/.jar
+io.ktor:ktor-server-routing-openapi/javadoc.jar
+io.ktor:ktor-server-routing-openapi/kotlin-tooling-metadata.json
+io.ktor:ktor-server-routing-openapi/sources.jar
```
osipxd, line 654: Probably other artifact dumps should also be updated

osipxd (approved):
🚀

gradle/artifacts/publishJvmAndCommonPublications.txt
bjhham: Oh right, I've just been running the one I normally use locally 🤦 

Maybe we need a CI step for checking all artifacts...

PR #5301, opened 2026-01-15
KTOR-9209 Support for Jackson 3.0

bjhham:
**Subsystem**
Client/Server, ContentNegotiation

**Motivation**
[KTOR-9209](https://youtrack.jetbrains.com/issue/KTOR-9209) Support Jackson 3

**Solution**
Introduced a new `jackson3` serialization module on account of the numerous breaking changes in the major release.

This is mostly copied over from the Jackson serialization module, but using the new API.  I think it would be premature to deprecate the original Jackson model for the time being, since Jackson 3 just came out in October.


ktor-shared/ktor-serialization/ktor-serialization-jackson3/build.gradle.kts
```diff
@@ -0,0 +1,28 @@
...
+            api(libs.jackson3.module.kotlin)
+        }
+        jvmTest.dependencies {
+            api(projects.ktorServerTestHost)
+            api(projects.ktorClientTests)
+            api(projects.ktorClientContentNegotiationTests)
+            api(projects.ktorSerializationTests)
+
+            api(libs.logback.classic)
+            api(libs.jackson3.dataformat.smile)
```
osipxd: Could you change scope for test dependencies from `api` to `implementation`?
Kotlin 2.3.0 shows warnings for test dependencies declared with `api` scope

PR #5308, opened 2026-01-17
KTOR-8316 Usability improvements for describe API

bjhham:
**Subsystem**
Server, OpenAPI Routing

**Motivation**
There were some soft spots in the DX of this API that needed addressing before the release.

**Solution**
- Added a simple `Route.hide()` function which works as the opposite of `describe()` for omitting routes from the documentation.
- Renamed and refactored some of the functions for generating the models from routes.
- Renaming and minor improvements for `OpenApiDocSource` which is used in the OpenAPI and Swagger plugins.


ktor-server/ktor-server-plugins/ktor-server-routing-openapi/common/src/io/ktor/server/routing/openapi/OpenApiDocSource.kt
```diff
@@ -21,75 +22,28 @@ import kotlinx.serialization.json.Json
...
 
     /**
      * A static string source for an OpenAPI document.
      *
      * @param content The text returned for the spec.
      */
-    public data class OpenApiDocText(
-        val content: String,
-        val contentType: ContentType = ContentType.Application.Json
+    public data class Text(
```
osipxd: Should it also be a plain class?
```suggestion
    public class Text(
```

bjhham: Yeah, there is just one destructuring reference atm I'll need to update...

PR #5310, opened 2026-01-19
KTOR-8316 Updates for 3.1.1 compatibility

bjhham:
**Subsystem**
Server, OpenAPI

**Motivation**
Found some out-of-date model references in our JSON schema.

**Solution**
1. Removed `nullable` in lieu of array base types like `type: [string, null]` or `anyOf` in the case of classes
2. `exclusiveMinimum` and `exclusiveMaximum` are now numbers instead of booleans
3. Included a fix for recursive types for kotlinx-serialization schema inference
4. Update swagger JS library version default in the plugin
5. recursiveRef / recursiveAnchor has been replaced with dynamicRef / dynamicAnchor
6. Added servers to components

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-swagger/jvm/test/io/ktor/server/plugins/swagger/SwaggerTest.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-swagger/jvm/test/io/ktor/server/plugins/swagger/SwaggerTest.kt
@@ -50,13 +50,13 @@ class SwaggerTest {
             <html>
               <head>
                 <title>Swagger UI</title>
-                <link href="https://unpkg.com/swagger-ui-dist@5.17.12/swagger-ui.css" rel="stylesheet">
-                <link href="https://unpkg.com/swagger-ui-dist@5.17.12/favicon-32x32.png" rel="icon" type="image/x-icon">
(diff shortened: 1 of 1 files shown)


ktor-shared/ktor-openapi-schema/common/src/io/ktor/openapi/OpenApiDoc.kt
```diff
@@ -330,6 +330,7 @@ public data class Components(
     public val links: Map<String, ReferenceOr<Link>>? = null,
     public val callbacks: Map<String, ReferenceOr<Callback>>? = null,
     public val pathItems: Map<String, ReferenceOr<PathItem>>? = null,
+    public val servers: Map<String, ReferenceOr<Server>>? = null,
```
zibet27, line 333: Shouldn't it be an array?
Do you think we should populate it somehow?

zibet27:
Looks like snapshot swagger tests are failing after update :(

ktor-shared/ktor-openapi-schema/common/src/io/ktor/openapi/OpenApiDoc.kt
bjhham: Nah, this is an optional one that was introduced in 3.1.1 and I don't think it's used too often.

PR #5311, opened 2026-01-19
Fix breaking change in `AuthenticationProvider.Config`

zibet27:
**Subsystem**
Server Auth

**Motivation**
I introduced breaking changes in [KTOR-9085](https://youtrack.jetbrains.com/issue/KTOR-9085)

The following code compiles with 3.3.3 but throws an error with 3.4.0-SNAPSHOT:
```kotlin
class Config : AuthenticationProvider.Config("myname") {
    
}
```

**Solution**
Add a default value for the description field.
I checked all other occurrences of `description`; it's required only in internal constructors.


osipxd:
It still seems to be binary incompatible

image

We should introduce a constructor with a single `String` parameter and `DeprecationLevel.HIDDEN`

PR #5313, opened 2026-01-19
KTOR-8821 Undeprecate blocking overloads

bjhham:
**Subsystem**
Server, Core

**Motivation**
[KTOR-8821](https://youtrack.jetbrains.com/issue/KTOR-8821) Autoreloading: module function refs not working since 3.2.0

**Solution**
Now that Kotlin 2.2.0 has been out for a little while, we should be able to deprecate these overloads, which should allow blocking function calls to pass to the embedded server without wrapping, which should fix auto-reloading here...  I will test it out now.


bjhham:
Looks like this is still not working.

PR #5318, opened 2026-01-19
KTOR-9210 Migrate Zstd into standalone module

bjhham:
**Subsystem**
Server, Compression

**Motivation**
[KTOR-9210](https://youtrack.jetbrains.com/issue/KTOR-9210) Zstd support import changes fixes

**Solution**
- Fix for missing backwards compatible constructor
- Moved zstd-jni to compileOnly so it requires the explicit library import
- Changed to actual class in relevant places to fix dokka issue


ktor-utils/build.gradle.kts
```diff
@@ -32,7 +32,7 @@ kotlin {
             implementation(projects.ktorTestDispatcher)
         }
         jvmMain.dependencies {
-            implementation(libs.zstd.jni)
+            compileOnly(libs.zstd.jni)
```
osipxd: Probably it's better to create some module with zstd support providing this dependency not to expose this library. As it is an implementation detail and can be changed later

ktor-server/ktor-server-plugins/ktor-server-compression/jvm/src/io/ktor/server/plugins/compression/Config.kt
```diff
@@ -170,7 +179,7 @@ public class CompressionEncoderBuilder internal constructor(
      *
      * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.server.plugins.compression.CompressionEncoderBuilder.compressionLevel)
      */
-    public var compressionLevel: Int = 3
+    public var compressionLevel: Int = CompressionEncoderConfig.DEFAULT_COMPRESSION_LEVEL
```
osipxd: Btw, do we plan to introduce compression level support to other compression algorithms? If not, we could introduce `ZstdEncoderBuilder` to set compression level only for zstd.

ktor-utils/build.gradle.kts
bjhham: Yeah, I kinda went for a quick-and-dirty approach 🙈  I'll migrate it out.

ktor-server/ktor-server-plugins/ktor-server-compression/jvm/src/io/ktor/server/plugins/compression/Config.kt
bjhham: I just moved it to an argument for the extension function.

Because of this:
```
public val encoders: MutableMap<String, CompressionEncoderBuilder>
```

We can't extend `CompressionEncoderBuilder` and assign it normally.  If we used an internal backing field we'd be alright though 😒

ktor-shared/ktor-encoding-zstd/build.gradle.kts
```diff
@@ -0,0 +1,24 @@
...
+
+plugins {
+    id("ktorbuild.project.library")
+}
+
+kotlin {
+    sourceSets {
+        commonMain.dependencies {
+            implementation(projects.ktorIo)
+            implementation(projects.ktorUtils)
```
osipxd, line 15: These dependencies probably should be added with `api` scope as they're part of public API

ktor-utils/build.gradle.kts
osipxd: Is this dependency still necessary?

ktor-shared/ktor-encoding-zstd/common/src/io/ktor/encoding/zstd/Zstd.kt
```diff
@@ -0,0 +1,34 @@
...
+import io.ktor.utils.io.ByteReadChannel
+import io.ktor.utils.io.ByteWriteChannel
+import kotlin.coroutines.CoroutineContext
+
+/**
+ * Implementation of [ContentEncoder] using zstd algorithm
+ *
+ * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.util.ZstdEncoder)
+ */
+public expect class ZstdEncoder : ContentEncoder {
```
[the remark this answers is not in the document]
osipxd: We can ignore such comments and update all references by running a tool

ktor-utils/build.gradle.kts
bjhham: Nope, I missed that one 👀

osipxd (approved):

PR #5320, opened 2026-01-20
Yaml config. Resolve references only once.

zibet27:
**Subsystem**
Server

**Motivation**
[KTOR-9250](https://youtrack.jetbrains.com/issue/KTOR-9250) YAML configuration: resolved value starting with $ is attempted to be resolved again

**Solution**
Resolve references only on config creation

### DIFF
--- a/ktor-server/ktor-server-config-yaml/jvm/test/YamlConfigTestJvm.kt
+++ b/ktor-server/ktor-server-config-yaml/jvm/test/YamlConfigTestJvm.kt
@@ -74,10 +74,14 @@ class YamlConfigTestJvm {
     @Test
     fun testEnvVarWithValueStartingWithDollar() = withProperty("TEST_DB_PASSWORD", $$"$123Password") {
         val content = $$"""
-                password: $TEST_DB_PASSWORD
+                password1: $TEST_DB_PASSWORD
+                password2: $TEST_DB_PASSWORD
+                password3: $password1
         """.trimIndent()
         val yaml = Yaml.default.decodeFromString<YamlMap>(content)
         val config = YamlConfig.from(yaml)
-        assertEquals($$"$123Password", config.property("password").getString())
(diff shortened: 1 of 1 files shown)


bjhham (approved):
Nice!  Seems to be a couple linting problems, but looks good.

Stexxe:
Can you please explain why it tries to resolve the resolved value? Is it because the resolved values are cached or something?

zibet27:
> Can you please explain why it tries to resolve the resolved value? Is it because the resolved values are cached or something?

The issue was that we were resolving references twice :
- once during config creation 
- and inside `propertyOrNull`

Stexxe:
> > Can you please explain why it tries to resolve the resolved value? Is it because the resolved values are cached or something?
> 
> The issue was that we were resolving references twice :
> 
> * once during config creation
> * and inside `propertyOrNull`

Why can't we do it once when the `propertyOrNull` is called to resolve the problem? 
Will the `Cycle detected in references` error be thrown if there are multiple keys that refer to the same variable?

zibet27:
> Why can't we do it once when the `propertyOrNull` is called to resolve the problem? 

We can, but I believe resolving references during initialisation and failing early is better

> Will the `Cycle detected in references` error be thrown if there are multiple keys that refer to the same variable?

This works
```kotlin
withProperty("TEST_DB_PASSWORD", $$"$123Password") {
        val content = $$"""
                password: $TEST_DB_PASSWORD
                password2: $TEST_DB_PASSWORD
                password3: $password
        """.trimIndent()
        val yaml = Yaml.default.decodeFromString<YamlMap>(content)
        val config = YamlConfig.from(yaml)
        assertEquals($$"$123Password", config.property("password").getString())
        assertEquals($$"$123Password", config.property("password2").getString())
        assertEquals($$"$123Password", config.property("password3").getString())
}
```

PR #5321, opened 2026-01-20
Merge 3.4.0

bjhham:


bjhham:
It occurred to me last night there may be more clever ways to avoid the commit duplication here...  Stay tuned.

PR #5324, opened 2026-01-21
KTOR-9263 Ignore chunked encoding extensions

bjhham:
**Subsystem**
Server, Core

**Motivation**
[KTOR-9263](https://youtrack.jetbrains.com/issue/KTOR-9263) HTTP Request Smuggling Vulnerability via Chunked Extension Quoted-String Parsing

**Solution**
- Use state-machine-style parsing for chunk size
- Ignore extensions
- Ignore CRLF's inside quotes
- Throw on \n without \r

Note, it appears that the carriage return inside quotes is not handled properly in any of our engines, so we can only fix for CIO atm.

### DIFF
--- a/ktor-http/ktor-http-cio/common/src/io/ktor/http/cio/ChunkedTransferEncoding.kt
+++ b/ktor-http/ktor-http-cio/common/src/io/ktor/http/cio/ChunkedTransferEncoding.kt
@@ -82,11 +82,8 @@ public suspend fun decodeChunked(input: ByteReadChannel, out: ByteWriteChannel) 
 }
 
 private suspend fun ByteReadChannel.skipCrLf() {
-    when (val b = readByte()) {
-        CR -> if (readByte() != LF) throw IOException("Expected LF")
-        LF -> return
-        else -> throw IOException("Expected CRLF but found 0x${b.toString(16)}")
-    }
+    if (readByte() != CR) throw IOException("Expected CR")
+    if (readByte() != LF) throw IOException("Expected LF")
 }
 
 private const val CR = '\r'.code.toByte()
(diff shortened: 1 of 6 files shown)


bjhham:
@coderabbitai review

ktor-http/ktor-http-cio/common/src/io/ktor/http/cio/ChunkedTransferEncoding.kt
```diff
@@ -65,40 +58,78 @@ public fun CoroutineScope.decodeChunked(input: ByteReadChannel, contentLength: L
...
+
+@OptIn(InternalAPI::class)
+private suspend fun parseChunkSize(input: ByteReadChannel): Long {
+    var result = 0L
+    var inExtension = false
+    var inQuotes = false
+    var afterCr = false
+    var i = 0
+    while (i++ < MAX_CHUNK_SIZE_LENGTH) {
+        val byte = input.readByte()
```
e5l: Could we write it using String.indexOf?. It will be slow with both index check and `byte by byte` reading

bjhham: My thinking was that it wouldn't be significant because the chunk size will seldom be over a few bytes long.  I can still look at optimizing though.

bjhham: Actually, we don't know how many characters to read in ahead of time to match on the CRLF so it needs to go one at a time.

Correction: we can use something like readLineStrictTo here

ktor-http/ktor-http-cio/common/src/io/ktor/http/cio/ChunkedTransferEncoding.kt
```diff
@@ -65,40 +58,78 @@ public fun CoroutineScope.decodeChunked(input: ByteReadChannel, contentLength: L
...
+        if (inQuotes && byte != QUOTE) continue
```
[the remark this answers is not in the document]
bjhham: That's the exploit which this is fixing.  It is technically legal to include CRLF inside quotes and they should be ignored.

ktor-http/ktor-http-cio/common/src/io/ktor/http/cio/ChunkedTransferEncoding.kt
bjhham: I ran some tests, and using the `kotlinx-io` buffer one byte at a time is actually much faster than using `indexOf` in this case.  Probably because behind the scenes it's roughly equivalent to iterating on a byte array.

PR #5330, opened 2026-01-25
KTOR-9421 Track streaming responses separately to fix SSE blocking flush

arkanovicz:
[KTOR-9421](https://youtrack.jetbrains.com/issue/KTOR-9421) Netty: active SSE connection blocks HTTP/2 response flushing for other requests

HTTP/2 multiplexes multiple streams on a single connection. When an SSE endpoint is active, it keeps activeRequests > 0 forever, which prevents flushIfNeeded() from flushing responses for other requests on the same connection.

This fix:
- Tracks streaming responses (SSE) separately with a dedicated counter
- Changes flush condition to allow flushing when only streaming requests remain
- Properly decrements the streaming counter when SSE connections end

### DIFF
--- a/ktor-server/ktor-server-netty/jvm/test/io/ktor/tests/server/netty/NettyEngineTest.kt
+++ b/ktor-server/ktor-server-netty/jvm/test/io/ktor/tests/server/netty/NettyEngineTest.kt
@@ -25,6 +25,7 @@ import io.netty.buffer.Unpooled
 import io.netty.handler.codec.http.HttpResponseStatus
 import io.netty.handler.codec.http2.*
 import io.netty.handler.codec.http2.Http2CodecUtil.readUnsignedInt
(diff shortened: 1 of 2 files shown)


arkanovicz:
> Could you include a test case in `HttpServerCommonTestSuite` that ensures an HTTP/2 connection will be flushed while an active SSE connection is present?

Well. This would have been quite too time consuming for me, but my coding agent said it would be happy to help, so here it is. It turned out not to be that trivial because the default apache5 server used in h2 tests does not multiplex sse and standard requests on the same h2 connection.

I checked that the test does fail without the patch.

arkanovicz:
Some builds did fail, though, but I do not have access to teamcity (do I?), so I cannot investigate further.

bjhham:
It seems to be some unrelated TLS failures.  I'll investigate.

PR #5331, opened 2026-01-26
KTOR-9211 Fix SendCountExceedException when max retries is lesser tha…

Stexxe:
https://youtrack.jetbrains.com/issue/KTOR-9211

### DIFF
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/HttpSend.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/HttpSend.kt
@@ -90,7 +90,7 @@ public class HttpSend private constructor(
                 }
                 context.setBody(content)
                 val maxRetries = context.attributes.getOrNull(MaxRetriesPerRequestAttributeKey)
-                val maxSendCount = if (maxRetries != null && maxRetries > plugin.maxSendCount) {
+                val maxSendCount = if (maxRetries != null && maxRetries >= plugin.maxSendCount) {
                     if (maxRetries < Int.MAX_VALUE) {
                         maxRetries + 1 // +1 for the initial request
                     } else {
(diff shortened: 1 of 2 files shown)


ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/HttpSend.kt
```diff
@@ -89,17 +89,17 @@ public class HttpSend private constructor(
                         .trimMargin()
                 }
                 context.setBody(content)
-                val maxRetriesFromRetryPlugin = context.attributes.getOrNull(MaxRetriesPerRequestAttributeKey)
-                val maxRetries = if (maxRetriesFromRetryPlugin != null) {
-                    if (maxRetriesFromRetryPlugin < Int.MAX_VALUE) {
-                        maxRetriesFromRetryPlugin + 1 // +1 for the initial request
+                val maxRetries = context.attributes.getOrNull(MaxRetriesPerRequestAttributeKey)
+                val maxSendCount = if (maxRetries != null && maxRetries > plugin.maxSendCount) {
```
Nek-12: Because maxSendCount is the total number of sends while maxRetries counts retries only, the condition maxRetries > plugin.maxSendCount leaves maxSendCount unchanged when maxRetries == plugin.maxSendCount, which still permits only maxSendCount - 1 retries. In that case the last retry attempt will throw SendCountExceedException (e.g., maxRetries=20 with the default maxSendCount=20 allows only 19 retries); this should compare against maxRetries + 1 or use >=.

bjhham: Perhaps something like this?

```
val maxRetries = context.attributes.getOrNull(MaxRetriesPerRequestAttributeKey)?.coerceAtMost(Integer.MAX_VALUE - 1)?.inc()
val maxSendCount = plugin.maxSendCount.coerceAtLeast(maxRetries ?: 0)
```

albertvaka:
Any chance we can get a release of this fix? This is affecting our users.

osipxd:
> Any chance we can get a release of this fix? This is affecting our users.

Yes! This fix is going to be included in 3.4.1, which is scheduled for this week.

PR #5334, opened 2026-01-27
Server Auth Jwt. Fail early on `verify` not provided, update docs

zibet27:
**Subsystem**
Server JWT

**Motivation**
[KTOR-7512](https://youtrack.jetbrains.com/issue/KTOR-7512) JWT: Docs for `validate` method claim that it's optional, but it isn't

**Solution**
- fail early if `authenticationFunction` is not provided
- updated KDocs


zibet27:
https://github.com/ktorio/ktor-documentation/pull/754

PR #5335, opened 2026-01-27
KTOR-9290 Jetty Jakarta idleTimeout does not cancel the request's job

amccague:
**Subsystem**
Jetty Jakarta Ktor Server

**Motivation**
https://youtrack.jetbrains.com/issue/KTOR-9290/Jetty-Jakarta-idleTimeout-does-not-cancel-the-requests-job

When Jetty's idle timeout expires during request processing, the timeout is effectively ignored if the server-side coroutine is performing slow work. Clients would still receive the intended status code beyond the idle timeout time.

The current tests are focussed on clients being slow to read and write.

The fix adds an idle timeout listener to the Jetty request that cancels the handler's coroutine job when the timeout fires therefore making it effective on server resources.

### DIFF
--- a/ktor-server/ktor-server-jetty-jakarta/jvm/src/io/ktor/server/jetty/jakarta/JettyKtorHandler.kt
+++ b/ktor-server/ktor-server-jetty-jakarta/jvm/src/io/ktor/server/jetty/jakarta/JettyKtorHandler.kt
@@ -67,7 +67,8 @@ internal class JettyKtorHandler(
     ): Boolean {
         try {
             val application = applicationProvider()
-            application.launch(handlerContext) {
+
+            val job = application.launch(handlerContext) {
                 val call = JettyApplicationCall(
                     application,
                     request,
@@ -82,28 +83,32 @@ internal class JettyKtorHandler(
                     pipeline.execute(call)
                     callback.succeeded()
                 } catch (cancelled: kotlinx.coroutines.CancellationException) {
-                    Response.writeError(request, response, callback, HttpStatus.GONE_410, cancelled.message, cancelled)
-                } catch (channelFailed: ChannelIOException) {
-                    callback.failed(channelFailed)
-                } catch (error: Throwable) {
-                    logError(call, error)
-                    if (!response.isCommitted) {
-                        try {
+                    if (response.isCommitted || runCatching {
                             Response.writeError(
                                 request,
                                 response,
                                 callback,
-                                HttpStatus.INTERNAL_SERVER_ERROR_500,
-                                error.message,
-                                error
+                                HttpStatus.GONE_410,
+                                cancelled.message,
+                                cancelled
                             )
-                        } catch (_: Throwable) {
-                            callback.failed(error)
-                        }
-                    } else {
-                        callback.failed(error)
+                        }.isFailure
+                    ) {
+                        callback.failed(cancelled)
                     }
+                } catch (channelFailed: ChannelIOException) {
+                    callback.failed(channelFailed)
+                } catch (error: Throwable) {
+                    logError(call, error)
+                    tryWriteError(response, request, callback, error)
                 }
+            }
+
+            request.addIdleTimeoutListener { timeoutException ->
+                job.cancel(CancellationException("Jetty idle timeout expired", timeoutException))
+                tryWriteError(response, request, callback, timeoutException)
+                // Returning true indicates that the idle timeout should be handled by the container as a fatal failure.
+                true
             }
         } catch (ex: Throwable) {
             environment.log.error("Application cannot fulfill the request", ex)
@@ -111,5 +116,26 @@ internal class JettyKtorHandler(
         }
 
         return true
+    }
+
+    private fun tryWriteError(
+        response: Response,
+        request: Request,
+        callback: Callback,
+        throwable: Throwable
+    ) {
+        if (response.isCommitted || runCatching {
+                Response.writeError(
+                    request,
+                    response,
+                    callback,
+                    HttpStatus.INTERNAL_SERVER_ERROR_500,
+                    throwable.message,
+                    throwable
+                )
+            }.isFailure
+        ) {
+            callback.failed(throwable)
+        }
     }
 }
--- a/ktor-server/ktor-server-jetty-jakarta/jvm/src/io/ktor/server/jetty/jakarta/JettyKtorHandler.kt
+++ b/ktor-server/ktor-server-jetty-jakarta/jvm/src/io/ktor/server/jetty/jakarta/JettyKtorHandler.kt
@@ -16,6 +16,7 @@ import org.eclipse.jetty.server.Request
 import org.eclipse.jetty.server.Response
 import org.eclipse.jetty.util.Callback
 import java.util.concurrent.*
+import java.util.concurrent.atomic.AtomicBoolean
 import java.util.concurrent.atomic.AtomicLong
 import kotlin.coroutines.CoroutineContext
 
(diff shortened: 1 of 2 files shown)


ktor-server/ktor-server-jetty-jakarta/jvm/src/io/ktor/server/jetty/jakarta/JettyKtorHandler.kt
```diff
@@ -82,13 +83,24 @@ internal class JettyKtorHandler(
                     pipeline.execute(call)
                     callback.succeeded()
                 } catch (cancelled: kotlinx.coroutines.CancellationException) {
-                    Response.writeError(request, response, callback, HttpStatus.GONE_410, cancelled.message, cancelled)
+                    if (response.isCommitted || runCatching {
+                            Response.writeError(
+                                request,
+                                response,
+                                callback,
+                                HttpStatus.GONE_410,
```
amccague: I should say I'm not a fan of this 410 Gone here - I can't see how a cancellation exception should imply

> the target resource is no longer available at the origin server and that this condition is likely to be permanent

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/410

408 probably fits better.. but in this context we can't be sure how this bubbled up so it could be a 500 like the others.

ktor-server/ktor-server-jetty-jakarta/jvm/src/io/ktor/server/jetty/jakarta/JettyKtorHandler.kt
```diff
@@ -82,13 +83,24 @@ internal class JettyKtorHandler(
...
+                        callback.failed(cancelled)
+                    }
                 } catch (channelFailed: ChannelIOException) {
                     callback.failed(channelFailed)
                 } catch (error: Throwable) {
                     logError(call, error)
-                    if (!response.isCommitted) {
-                        try {
+                    if (response.isCommitted || runCatching {
                             Response.writeError(
```
amccague: Similar comment above. If the response is committed then we shouldn't be able to write a response so this attempt should always fail.

amccague:
Hi @bjhham, would you consider reviewing this?

ktor-server/ktor-server-jetty-jakarta/jvm/src/io/ktor/server/jetty/jakarta/JettyKtorHandler.kt
bjhham: I'd stick with a 500 like the others in this case since we don't know the cause of the cancellation.

bjhham (commented):
I actually don't think Ktor should be handling timeouts in the server.  The request and idle timeouts are there for preventing the server from locking up from too many slow client connections, whereas slow activity on the server should be handled with timeouts in the application code.

bjhham:
Oh, disregard that, I see you're still using the idle timeout listener from the Jetty engine.

ktor-server/ktor-server-jetty-jakarta/jvm/src/io/ktor/server/jetty/jakarta/JettyKtorHandler.kt
```diff
@@ -82,34 +83,59 @@ internal class JettyKtorHandler(
...
+                    callback.failed(channelFailed)
+                } catch (error: Throwable) {
+                    logError(call, error)
+                    tryWriteError(response, request, callback, error)
                 }
             }
+
+            request.addIdleTimeoutListener { timeoutException ->
+                job.cancel(CancellationException("Jetty idle timeout expired", timeoutException))
+                tryWriteError(response, request, callback, timeoutException)
```
bjhham: It would be nice to use 408 here, though if this fires from hang ups on the server side too then we can't make any claims on whose fault it is...

ktor-server/ktor-server-jetty-jakarta/jvm/src/io/ktor/server/jetty/jakarta/JettyKtorHandler.kt
bjhham: A 503 might actually make more sense, because it might trigger a back-off in the client, and cancellations often occur when restarting too.

bjhham (changes requested):
I'd just drop the 410 for either a 500 or 503, the rest looks good.

amccague:
> Oh, disregard that, I see you're still using the idle timeout listener from the Jetty engine.

Yeah this is more about hooking in the intention that configuring the jetty idle timeout I would say has started but not finished.

Another option is making implementing this listener more accessible in the ktor user code so this behaviour can be opted in.

In our case we've seen a scenario where the idle timeouts have triggered client retries but the server work has then built up leading to resource exhaustion. More explicit timeout handling has been used but I feel this is a better default.

ktor-server/ktor-server-jetty-jakarta/jvm/src/io/ktor/server/jetty/jakarta/JettyKtorHandler.kt
amccague: The semantics of 503 sound attractive here. To quote mozilla:

> In overload cases, some server-side applications will reject requests with a 503 status when resource thresholds like memory, CPU, or connection pool limits are met. Dropping incoming requests creates backpressure that prevents the server's compute resources from being exhausted, avoiding more severe failures.

Maybe 503 is more proactive, whereas the response here is more reactive? So I could lean closer to 500 in this case, and reserve 503 for a condition if e.g. the Jetty QTP was saturated.

Side benefit of the 500 is it simplifies this code further..

ktor-server/ktor-server-jetty-jakarta/jvm/src/io/ktor/server/jetty/jakarta/JettyKtorHandler.kt
amccague: If I'm not mistaken, it would be rare (as in I can't see a clear scenario for it) for this to fire on a client stall vs server. So I would lean towards the 5XX.

ktor-server/ktor-server-jetty-jakarta/jvm/src/io/ktor/server/jetty/jakarta/JettyKtorHandler.kt
amccague: Another thought is that for an app behind a load balancer that is checking health probes, will commonly return a 503 if there's no healthy backends, so it may make investigation trickier.

bjhham: Yeah, 500 probably makes the most sense.

amccague: Done

ktor-server/ktor-server-jetty-jakarta/jvm/src/io/ktor/server/jetty/jakarta/JettyKtorHandler.kt
```diff
@@ -82,34 +83,47 @@ internal class JettyKtorHandler(
...
+                    callback,
+                    HttpStatus.INTERNAL_SERVER_ERROR_500,
+                    throwable.message,
+                    throwable
+                )
+            }.isFailure
+        ) {
+            callback.failed(throwable)
+        }
+    }
```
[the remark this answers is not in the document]
amccague: @coderabbitai Would you agree this concern is addressed by the `callbackCompleted` guard introduced in the last commit? Every call to `tryWriteError` is now wrapped in `completeCallback { ... }`, which only executes if `compareAndSet(false, true)` wins, ensuring the callback has not yet been completed when `tryWriteError` runs.

The `response.isCommitted` → `callback.failed` branch remains correct for the case where the pipeline began writing headers (committing the response) but threw before `callback.succeeded()` was reached. The callback is still incomplete in that scenario and must be failed.

The proposed early return would silently drop the callback in that case, leaving the request hanging.

bjhham (approved):

PR #5338, opened 2026-01-29
KTOR-9293 Add empty summary default for operations

bjhham:
**Subsystem**
Server, Swagger

**Motivation**
[KTOR-9293](https://youtrack.jetbrains.com/issue/KTOR-9293) OpenAPI describe needs defaults

The Swagger front-end breaks when Operation is an empty object.

**Solution**
Added an empty string summary for the `Operation` type.


ktor-shared/ktor-openapi-schema/common/src/io/ktor/openapi/Operation.kt
```diff
@@ -89,7 +89,7 @@ public data class Operation(
          *
          * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.openapi.Operation.Builder.summary)
          */
-        public var summary: String? = null
+        public var summary: String? = ""
```
zibet27, line 92: Can we make it non-nullable?

bjhham: That would be a breaking change 😞

zibet27:
Looks like `io.ktor.openapi.OperationSerializationTests.operation with tags and external docs()[jvm]` is failing

bjhham:
Yeah it seems that the empty string uncovered a bug in the YAML parser.  

It creates something like:

```
summary:
responses:
  200: yadda yadda
```

Then interpreting `responses` as being part of `summary`.  YAML is hard.

PR #5339, opened 2026-01-29
Server Auth. JWK auto-configuration

zibet27:
**Subsystem**
Server Auth 

**Motivation**
[KTOR-8595](https://youtrack.jetbrains.com/issue/KTOR-8595) Auth JWK Support (auto-discover)
Connected proposal: [KTOR-8596](https://youtrack.jetbrains.com/issue/KTOR-8596) OpenID Connect (OAuth2) auto-discover & configuration

**Solution**
 ```kotlin
         val config = httpClient.fetchOpenIdConfiguration("https://accounts.google.com")
         install(Authentication) {
              jwt {
                  validate { credential ->
                      // check if user exists in database
                      val userId = credential.subject
                      database.findUser(userId)?.let { UserPrincipal(it) }
                  }
                  jwk {
                      openIdConfig = config
                      audience = "my-app-client-id"
                  }
              }
         }
```

Added extension on `HttpClient`, `jwk` dsl (maybe better name?)

Open question: json deserialization

### CODE REVIEW

File: ktor-server/ktor-server-plugins/ktor-server-auth-jwt/jvm/test/io/ktor/server/auth/jwt/JwkDiscoveryTest.kt [hunk 1]
```diff
@@ -0,0 +1,290 @@
...
+    @SerialName("alg") val algorithm: String?,
+    @SerialName("use") val usage: String?,
+    @SerialName("n") val modulus: String?,
+    @SerialName("e") val exponent: String?
+)
+
+@Serializable
+internal data class JwksKeys(val keys: List<JwksKey>)
+
+class JwkDiscoveryTest {
```
bjhham, line 51: Could you include a test with it working in conjunction with JWT validation?

ktor-server/ktor-server-plugins/ktor-server-auth-jwt/jvm/src/io/ktor/server/auth/jwt/JWTAuth.kt
```diff
@@ -388,6 +388,56 @@ public class JWTAuthenticationProvider internal constructor(config: Config) : Au
             challenge = block
         }
 
+        /**
+         * Configures JWT authentication using OpenID Connect discovery.
+         *
+         * This function will:
+         * 1. Set up a JWT verifier with the fetched keys. Keys are cached and rate-limited by default, see [JwkProviderBuilder].
+         * 2. Validate issuer and audience claims automatically. You still need to specify `validate` function before `jwk`.
```
bjhham: > You still need to specify `validate` function before `jwk`.

Since this doesn't depend on the result of validate, I think we can mitigate the temporal coupling here by introducing a new field like `validateCheck: (JWTCredential) -> String?` that can be called before the validate.

ktor-server/ktor-server-plugins/ktor-server-auth-jwt/jvm/src/io/ktor/server/auth/jwt/JWTAuth.kt
```diff
@@ -388,6 +388,56 @@ public class JWTAuthenticationProvider internal constructor(config: Config) : Au
...
+
+            verifier(jwkProvider, issuer, config.jwtConfigure)
+
+            val customValidation = requireNotNull(authenticationFunction) {
+                "JWT auth validate function is not specified. Use jwt { validate { ... } } to fix."
+            }
+            validate { credential ->
+                val issuerMatches = credential.issuer == issuer
+                val audienceMatches = config.audience?.let { credential.audience.contains(it) } ?: true
+                if (!issuerMatches || !audienceMatches) return@validate null
```
bjhham: Would be nice to include a log message when one of these doesn't match, specifying which field (or both) doesn't match.

bjhham (changes requested):

ktor-server/ktor-server-plugins/ktor-server-auth-jwt/jvm/src/io/ktor/server/auth/jwt/JWTAuth.kt
zibet27: Wouldn't it be confusing to have both `validateCheck` and `validate`?
Or you mean it to be inside of `jwk { }`?  In that case, I guess it should work

zibet27:
https://github.com/ktorio/ktor/pull/5341

ktor-server/ktor-server-plugins/ktor-server-auth-jwt/jvm/src/io/ktor/server/auth/jwt/JWTAuth.kt
bjhham: I meant internally, you could introduce pre-validate checks.  So keep validate as it is inside jwt, but introduce an internal field so validation can be extended with pre-conditions during the execution.

bjhham (changes requested):

ktor-server/ktor-server-plugins/ktor-server-auth-jwt/jvm/test/io/ktor/server/auth/jwt/JwkDiscoveryTest.kt
zibet27: Added

ktor-server/ktor-server-plugins/ktor-server-auth-jwt/jvm/src/io/ktor/server/auth/jwt/JWTAuth.kt
```diff
@@ -167,6 +167,8 @@ public class JWTPrincipal(payload: Payload) : JWTPayloadHolder(payload)
  */
 public typealias JWTConfigureFunction = Verification.() -> Unit
 
+internal typealias PreAuthenticationFunction<C> = suspend ApplicationCall.(credentials: C) -> Boolean
```
bjhham: Let's drop the typealias here since it's only used in one field.

ktor-server/ktor-server-plugins/ktor-server-auth-jwt/jvm/src/io/ktor/server/auth/jwt/JWTAuth.kt
```diff
@@ -181,9 +183,12 @@ public class JWTAuthenticationProvider internal constructor(config: Config) : Au
...
     private val authHeader: (ApplicationCall) -> HttpAuthHeader? = config.authHeader
     private val verifier: ((HttpAuthHeader) -> JWTVerifier?) = config.verifier ?: { null }
-    private val authenticationFunction = config.authenticationFunction ?: throw IllegalArgumentException(
+    private val authenticationFunction = requireNotNull(config.authenticationFunction) {
         "JWT auth validate function is not specified. Use jwt { validate { ... } } to fix."
-    )
+    }
+
+    private val preAuthenticationFunction: PreAuthenticationFunction<JWTCredential> =
+        config.preAuthenticationFunction ?: { true }
```
bjhham: The naming is a little obscure, how about something like `verifyCredentials`?

zibet27: Yeah, the name here is not very obvious
I would rename 
`authenticationFunction` -> `validateCredentials`
`preAuthenticationFunction` -> `prevalidateCredentials`
so its clear that `prevalidateCredentials` runs just before `validateCredentials` and has roughly the same.

Using `verifyCredentials` can be confused with `JWTVerifier.verify(token: String): DecodedJWT`
What do you think?

ktor-server/ktor-server-plugins/ktor-server-auth-jwt/jvm/src/io/ktor/server/auth/jwt/JwkConfig.kt
```diff
@@ -0,0 +1,172 @@
...
+    /**
+     * Configures caching for fetched JSON Web Keys.
+     * Caching reduces the number of HTTP requests to the JWKS endpoint by storing previously fetched keys.
+     *
+     * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.server.auth.jwt.JwkConfig.cache)
+     *
+     * @param maxEntries Maximum number of keys to cache
+     * @param duration How long cached keys remain valid before being refreshed
+     */
+    public fun cache(maxEntries: Long, duration: Duration) {
```
e5l: Could we add some reasonable defaults?

e5l: please check other config methods as well

ktor-server/ktor-server-plugins/ktor-server-auth-jwt/jvm/src/io/ktor/server/auth/jwt/JWTAuth.kt
```diff
@@ -388,6 +394,71 @@ public class JWTAuthenticationProvider internal constructor(config: Config) : Au
...
+         *     }
+         * }
+         * ```
+         *
+         * @see JwkConfig
+         * @see fetchOpenIdConfiguration
+         */
+        public fun jwk(configure: JwkConfig.() -> Unit) {
+            val config = JwkConfig().apply(configure)
+            val openId = requireNotNull(config.openIdConfig) {
```
e5l: if the property is obligatory, please consider include it in the function arguments

e5l (approved):
lgtm. please check comments before merging

bjhham (approved):
All good; see others comments 👍

ktor-server/ktor-server-plugins/ktor-server-auth-jwt/jvm/src/io/ktor/server/auth/jwt/JwkConfig.kt
zibet27: `com.auth0` has default values, so we implicitly relied on them here. 
But I will explicitly set defaults, so we don't rely on `com.auth0` as much.

zibet27:
This is now implemented in the OIDC plugin

PR #5340, opened 2026-01-29
Fix get and set implementations for `Uint8Array` and `ArrayLike`

bashor:
**Subsystem**
`ktor-client-core` for  `wasm-js` target

**Motivation**
With the current implementation, when they are generated into a JS file, the resulting code is syntactically incorrect, so webpack and JS VMs reject such code even if it's unreachable.

It is mostly unobservable since the declarations are not used inside ktor, and the kotlin toolchain doesn't generate helper functions for them by default. But it breaks projects using ktor when we try to build them with wip per-module compilation.

**Solution**
Fix incorrect JS code.


bashor:
It would also be nice to have the fix in some upcoming (fix?) releases.

bjhham:
[KTOR-9316](https://youtrack.jetbrains.com/issue/KTOR-9316)

PR #5341, opened 2026-01-30
OAuth2 auto-configuration

zibet27:
**Subsystem**
Server Auth

**Motivation**
[KTOR-8596](https://youtrack.jetbrains.com/issue/KTOR-8596) OpenID Connect (OAuth2) auto-discover & configuration

Also see [KTOR-8595](https://youtrack.jetbrains.com/issue/KTOR-8595) Auth JWK Support (auto-discover)
https://github.com/ktorio/ktor/pull/5339

**Solution**
New DSL
```kotlin
val httpClient = HttpClient()
val config = httpClient.fetchOpenIdConfiguration("https://accounts.google.com")
install(Authentication) {
   oauth("test-openid", openIdConfig) {
      client = httpClient
      clientId = "test-client-id"
      clientSecret = "test-client-secret"
      urlProvider = { "http://localhost/callback" }
   }
}
```


zibet27:
@coderabbitai review

ktor-server/ktor-server-plugins/ktor-server-auth/api/ktor-server-auth.klib.api
```diff
@@ -286,16 +286,37 @@ final class io.ktor.server.auth/OAuthAuthenticationProvider : io.ktor.server.aut
 }
 
 final class io.ktor.server.auth/OpenIdConfiguration { // io.ktor.server.auth/OpenIdConfiguration|null[0]
-    constructor <init>(kotlin/String, kotlin/String) // io.ktor.server.auth/OpenIdConfiguration.<init>|<init>(kotlin.String;kotlin.String){}[0]
```
bjhham, line 289: Appears to be a breaking change in the constructor args.

ktor-server/ktor-server-plugins/ktor-server-auth/api/ktor-server-auth.api
```diff
@@ -574,9 +593,6 @@ public final class io/ktor/server/auth/OpenIdConfigurationKt {
 }
 
 public final class io/ktor/server/auth/OpenIdOAuthConfig {
-	public field clientId Ljava/lang/String;
-	public field clientSecret Ljava/lang/String;
-	public field urlProvider Lkotlin/jvm/functions/Function3;
```
bjhham: This appears to be breaking, though I'm not sure what happened.  Maybe there's a JVM annotation to retain them as fields.

ktor-server/ktor-server-plugins/ktor-server-auth/api/ktor-server-auth.klib.api
zibet27: This API has not been released yet.
Maybe it's better to make it a DSL-like non-data class, so it's easier to add extra fields in the future without breaking the API?

bjhham: Oh, my mistake I thought it was.  I'd remove the data keyword... unless if it's used in serialization.

ktor-server/ktor-server-plugins/ktor-server-auth/api/ktor-server-auth.api
zibet27: `OpenIdOAuthConfig` was introduced in this pull request.
The issue here is that I was switching branches (`git switch` instead of `git checkout`), and the `OpenIdOAuthConfig` definition appeared there.
`jwt-auto-discover` has now been updated with a valid ABI.

ktor-server/ktor-server-plugins/ktor-server-auth/api/ktor-server-auth.klib.api
zibet27: It is used for deserialization.

bjhham: Yeah it's fine then.

ktor-server/ktor-server-plugins/ktor-server-auth/api/ktor-server-auth.klib.api
```diff
@@ -286,16 +286,37 @@ final class io.ktor.server.auth/OAuthAuthenticationProvider : io.ktor.server.aut
...
+    final fun component7(): kotlin.collections/List<kotlin/String>? // io.ktor.server.auth/OpenIdConfiguration.component7|component7(){}[0]
+    final fun component8(): kotlin.collections/List<kotlin/String>? // io.ktor.server.auth/OpenIdConfiguration.component8|component8(){}[0]
+    final fun component9(): kotlin.collections/List<kotlin/String>? // io.ktor.server.auth/OpenIdConfiguration.component9|component9(){}[0]
+    final fun copy(kotlin/String = ..., kotlin/String = ..., kotlin/String = ..., kotlin/String? = ..., kotlin/String = ..., kotlin.collections/List<kotlin/String>? = ..., kotlin.collections/List<kotlin/String>? = ..., kotlin.collections/List<kotlin/String>? = ..., kotlin.collections/List<kotlin/String>? = ...): io.ktor.server.auth/OpenIdConfiguration // io.ktor.server.auth/OpenIdConfiguration.copy|copy(kotlin.String;kotlin.String;kotlin.String;kotlin.String?;kotlin.String;kotlin.collections.List<kotlin.String>?;kotlin.collections.List<kotlin.String>?;kotlin.collections.List<kotlin.String>?;kotlin.collections.List<kotlin.String>?){}[0]
```
osipxd: This is the problem with data classes :(
It will be hard to add new fields or rearrange existing ones without breaking binary compatibility

ktor-server/ktor-server-plugins/ktor-server-auth/common/src/io/ktor/server/auth/OpenIdOAuthConfig.kt
```diff
@@ -0,0 +1,147 @@
...
+ *
+ * This configuration is used with [AuthenticationConfig.oauth] when an [OpenIdConfiguration]
+ * is provided. It pre-configures OAuth endpoints from the discovery document while allowing
+ * customization of client credentials and other settings.
+ *
+ * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.server.auth.OpenIdOAuthConfig)
+ *
+ * @property openIdConfiguration The OpenID Connect discovery configuration
+ */
+public class OpenIdOAuthConfig(
```
osipxd, line 26: Could you add `@KtorDsl` annotation here? To avoid usage of other DSLs inside the `oauth { ... }` block

ktor-server/ktor-server-plugins/ktor-server-auth/api/ktor-server-auth.klib.api
zibet27: Do you think we need to make it DSL-like and manually deserialize or it's ok to keep it dataclass?

osipxd: It should still be serializable as a plain class if I understand it correctly
So, we could avoid `data class` here, and introduce `toString`, `hasCode` and `equals` if necessary

PR #5342, opened 2026-01-31
Server throughput optimizations: reduce GC overhead 

e5l:
Optimizations applied:
- Skip redundant job cancellation in NettyApplicationCall
- Lazy routing trace registration (only when TRACE logging enabled)
- CaseInsensitiveMap: open-addressing hash table (eliminates CaseInsensitiveString allocations)
- HttpProtocolVersion.parse: fast path for HTTP/1.0, HTTP/1.1, HTTP/2.0, HTTP/3.0
- StringValuesImpl: parallel arrays for zero-allocation forEach
- GMTDate: zero-allocation timestamp conversion using civil_from_days algorithm
- LazyThreadSafetyMode.NONE for request-scoped lazy properties

Results (256 byte payload, 64 concurrency):
- Download: 22,713 req/s, 5.55 MB/s
- Upload: 23,529 req/s, 5.74 MB/s
- GC overhead: reduced from ~10.5% to 0.70%

Added throughput benchmark module with:
- Request/response throughput benchmarks
- Big file transfer benchmark (achieved 10.66 Gbps with LocalFileContent, 18.48 Gbps with in-memory)
- async-profiler integration scripts

### DIFF
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/call/SavedCall.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/call/SavedCall.kt
@@ -15,19 +15,6 @@ import kotlinx.io.readByteArray
 import kotlin.coroutines.CoroutineContext
 
 /**
- * Marker interface for responses that have their body stored as a byte array.
- * This allows optimized access without copying through ByteReadChannel.
- */
-@InternalAPI
-public interface SavedResponseBody {
-    /**
-     * The response body as a byte array.
-     * Callers should not modify this array.
-     */
-    public val savedBody: ByteArray
-}
-
-/**
  * Saves the entire content of this [HttpClientCall] to memory and returns a new [HttpClientCall]
  * with the content cached in memory.
  * This can be particularly useful for caching, debugging,
@@ -80,9 +67,7 @@ internal class SavedHttpResponse(
     override val call: SavedHttpCall,
     private val body: ByteArray,
     origin: HttpResponse
-) : HttpResponse(), SavedResponseBody {
-
-    override val savedBody: ByteArray get() = body
+) : HttpResponse() {
     override val status: HttpStatusCode = origin.status
 
     override val version: HttpProtocolVersion = origin.version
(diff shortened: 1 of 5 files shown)


ktor-client/ktor-client-core/common/src/io/ktor/client/call/SavedCall.kt
```diff
@@ -8,11 +8,25 @@ import io.ktor.client.*
...
+ * Marker interface for responses that have their body stored as a byte array.
+ * This allows optimized access without copying through ByteReadChannel.
+ */
+@InternalAPI
+public interface SavedResponseBody {
+    /**
+     * The response body as a byte array.
+     * Callers should not modify this array.
+     */
+    public val savedBody: ByteArray
```
bjhham: Could use the existing hierarchy and drop the interface, like:

```
abstract class HttpResponse {
    @InternalAPI
    public open fun bodyAsByteArray(): ByteArray = body.toByteArray()
}

class SavedHttpResponse {
    override fun bodyAsByteArray(): ByteArray = savedBody
}

...

val bytes = response.bodyAsByteArray()
```

ktor-http/common/src/io/ktor/http/HttpProtocolVersion.kt
```diff
@@ -77,10 +77,14 @@ public data class HttpProtocolVersion(val name: String, val major: Int, val mino
...
         public fun parse(value: CharSequence): HttpProtocolVersion {
-            /**
-             * Format: protocol/major.minor
-             */
-                val (protocol, major, minor) = value.split("/", ".").also {
+            // Fast path: check common versions first to avoid allocation
+            if (value == "HTTP/1.1") return HTTP_1_1
+            if (value == "HTTP/1.0") return HTTP_1_0
+            if (value == "HTTP/2.0" || value == "HTTP/2") return HTTP_2_0
+            if (value == "HTTP/3.0" || value == "HTTP/3") return HTTP_3_0
```
bjhham, line 84: Wouldn't `when` be better here?

bjhham (commented):
It may be best to transfer the benchmark test to `ktor-benchmarks` to keep everything in one place.

It also seems that the response write job cancellations might not have been so redundant by the looks of the CI...  Maybe it's something else causing the timeouts idk...

ktor-client/ktor-client-core/common/src/io/ktor/client/call/SavedCall.kt
e5l: thanks, will drop

ktor-http/common/src/io/ktor/http/HttpProtocolVersion.kt
e5l: it will be the same, but look nicer. let me replace

PR #5348, opened 2026-02-04
Server Digest Auth. Support RFC 7616.

zibet27:
**Subsystem**
Server Digest Auth

**Motivation**
[KTOR-7578](https://youtrack.jetbrains.com/issue/KTOR-7578) Update Digest authentication implementation according to RFC 7616

**Solution**
- send required `qop` parameter
- support multiple hash algorithms at the same time
- enums for algorithms and qop
- Username Hashing
- support for -sess algorithms
- support of qop=auth-int

Docs need to be updated with examples using  sha256/512.

PS: How can we update DigestCredential without breaking ABI?

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-auth/jvm/src/io/ktor/server/auth/DigestAuth.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-auth/jvm/src/io/ktor/server/auth/DigestAuth.kt
@@ -1,5 +1,5 @@
 /*
-* Copyright 2014-2025 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+* Copyright 2014-2026 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
 */
 
 package io.ktor.server.auth
@@ -60,7 +60,7 @@ public class DigestAuthenticationProvider internal constructor(
         val authorizationHeader = call.request.parseAuthorizationHeader()
         val credentials = authorizationHeader?.let { authHeader ->
             if (authHeader.authScheme == AuthScheme.Digest && authHeader is HttpAuthHeader.Parameterized) {
-                authHeader.toDigestCredential(defaultCharset = charset ?: Charsets.ISO_8859_1)
+                authHeader.toDigestCredential(defaultCharset = charset)
             } else {
                 null
             }
@@ -195,7 +195,9 @@ public class DigestAuthenticationProvider internal constructor(
         public var algorithmName: String
             get() = algorithms.first().hashName
             set(value) {
-                val digestAlgorithm = DigestAlgorithm.from(value) ?: error("Unsupported digest algorithm: $value")
+                val digestAlgorithm = requireNotNull(DigestAlgorithm.from(value)) {
+                    "Unsupported digest algorithm: $value"
+                }
                 algorithms = listOf(digestAlgorithm)
             }
 
@@ -323,7 +325,7 @@ public class DigestAuthenticationProvider internal constructor(
          */
         public fun strictRfc7616Mode() {
             @Suppress("DEPRECATION")
-            if (DigestAlgorithm.MD5 in algorithms) {
+            if (DigestAlgorithm.MD5 in algorithms || DigestAlgorithm.MD5_SESS in algorithms) {
                 if (algorithms !== defaultAlgorithms) {
                     LOGGER.warn("MD5 algorithms are overridden in strictRfc7616Mode")
                 }
@@ -422,7 +424,7 @@ public fun HttpAuthHeader.Parameterized.toDigestCredential(): DigestCredential =
  *
  * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.server.auth.verifier)
  */
-@Deprecated(message = "Use [DigestCredential.verifier] without digester.")
+@Deprecated(message = "Use [DigestCredential.verifier] without digester.", level = DeprecationLevel.WARNING)
 public suspend fun DigestCredential.verifier(
     method: HttpMethod,
     digester: MessageDigest,
@@ -440,7 +442,7 @@ public suspend fun DigestCredential.verifier(
  *
  * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.server.auth.expectedDigest)
  */
-@Deprecated(message = "Use [DigestCredential.expectedDigest] without digester.")
+@Deprecated(message = "Use [DigestCredential.expectedDigest] without digester.", level = DeprecationLevel.WARNING)
 public fun DigestCredential.expectedDigest(
     method: HttpMethod,
     digester: MessageDigest,
--- a/ktor-server/ktor-server-plugins/ktor-server-auth/jvm/src/io/ktor/server/auth/DigestCredential.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-auth/jvm/src/io/ktor/server/auth/DigestCredential.kt
@@ -44,7 +44,7 @@ public class DigestCredential(
     public val charset: Charset = Charsets.UTF_8
 ) {
 
-    @Deprecated("Maintained for binary compatibility", level = DeprecationLevel.HIDDEN)
+    @Deprecated("Maintained for backwards compatibility", level = DeprecationLevel.WARNING)
     public constructor(
         realm: String,
         userName: String,
@@ -57,14 +57,22 @@ public class DigestCredential(
         cnonce: String?,
         qop: String?,
     ) : this(
-        realm, userName, digestUri, nonce, opaque, nonceCount, algorithm.toDigestAlgorithm(), response, cnonce, qop
+        realm = realm,
+        userName = userName,
+        digestUri = digestUri,
+        nonce = nonce,
+        opaque = opaque, nonceCount = nonceCount,
+        digestAlgorithm = algorithm.toDigestAlgorithm(),
+        response = response,
+        cnonce = cnonce,
+        qop = qop
     )
 
-    @Deprecated("Maintained for binary compatibility", level = DeprecationLevel.HIDDEN)
-    public val DigestCredential.algorithm: String?
+    @Deprecated("Maintained for backward compatibility", level = DeprecationLevel.WARNING)
+    public val algorithm: String?
         get() = digestAlgorithm.name
 
-    @Deprecated("Maintained for binary compatibility", level = DeprecationLevel.HIDDEN)
+    @Deprecated("Maintained for backward compatibility", level = DeprecationLevel.WARNING)
     public fun copy(
         realm: String = this.realm,
         userName: String = this.userName,
@@ -77,18 +85,18 @@ public class DigestCredential(
         cnonce: String? = this.cnonce,
         qop: String? = this.qop
     ): DigestCredential = DigestCredential(
-        realm,
-        userName,
-        digestUri,
-        nonce,
-        opaque,
-        nonceCount,
-        algorithm.toDigestAlgorithm(),
-        response,
-        cnonce,
-        qop,
-        userHash,
-        charset
+        realm = realm,
+        userName = userName,
+        digestUri = digestUri,
+        nonce = nonce,
+        opaque = opaque,
+        nonceCount = nonceCount,
+        digestAlgorithm = algorithm.toDigestAlgorithm(),
+        response = response,
+        cnonce = cnonce,
+        qop = qop,
+        userHash = userHash,
+        charset = charset
     )
 
     public fun copy(
@@ -105,48 +113,48 @@ public class DigestCredential(
         userHash: Boolean = this.userHash,
         charset: Charset = this.charset
     ): DigestCredential = DigestCredential(
-        realm,
-        userName,
-        digestUri,
-        nonce,
-        opaque,
-        nonceCount,
-        digestAlgorithm,
-        response,
-        cnonce,
-        qop,
-        userHash,
-        charset
+        realm = realm,
+        userName = userName,
+        digestUri = digestUri,
+        nonce = nonce,
+        opaque = opaque,
+        nonceCount = nonceCount,
+        digestAlgorithm = digestAlgorithm,
+        response = response,
+        cnonce = cnonce,
+        qop = qop,
+        userHash = userHash,
+        charset = charset
     )
 
-    @Deprecated("Maintained for binary compatibility", level = DeprecationLevel.HIDDEN)
+    @Deprecated("Maintained for backward compatibility", level = DeprecationLevel.WARNING)
     public operator fun component1(): String = realm
 
-    @Deprecated("Maintained for binary compatibility", level = DeprecationLevel.HIDDEN)
+    @Deprecated("Maintained for backward compatibility", level = DeprecationLevel.WARNING)
     public operator fun component2(): String = userName
 
-    @Deprecated("Maintained for binary compatibility", level = DeprecationLevel.HIDDEN)
+    @Deprecated("Maintained for backward compatibility", level = DeprecationLevel.WARNING)
     public operator fun component3(): String = digestUri
 
-    @Deprecated("Maintained for binary compatibility", level = DeprecationLevel.HIDDEN)
+    @Deprecated("Maintained for backward compatibility", level = DeprecationLevel.WARNING)
     public operator fun component4(): String = nonce
 
-    @Deprecated("Maintained for binary compatibility", level = DeprecationLevel.HIDDEN)
+    @Deprecated("Maintained for backward compatibility", level = DeprecationLevel.WARNING)
     public operator fun component5(): String? = opaque
 
-    @Deprecated("Maintained for binary compatibility", level = DeprecationLevel.HIDDEN)
+    @Deprecated("Maintained for backward compatibility", level = DeprecationLevel.WARNING)
     public operator fun component6(): String? = nonceCount
 
-    @Deprecated("Maintained for binary compatibility", level = DeprecationLevel.HIDDEN)
+    @Deprecated("Maintained for backward compatibility", level = DeprecationLevel.WARNING)
     public operator fun component7(): String = digestAlgorithm.name
 
-    @Deprecated("Maintained for binary compatibility", level = DeprecationLevel.HIDDEN)
+    @Deprecated("Maintained for backward compatibility", level = DeprecationLevel.WARNING)
     public operator fun component8(): String = response
 
-    @Deprecated("Maintained for binary compatibility", level = DeprecationLevel.HIDDEN)
+    @Deprecated("Maintained for backward compatibility", level = DeprecationLevel.WARNING)
     public operator fun component9(): String? = cnonce
 
-    @Deprecated("Maintained for binary compatibility", level = DeprecationLevel.HIDDEN)
+    @Deprecated("Maintained for backward compatibility", level = DeprecationLevel.WARNING)
     public operator fun component10(): String? = qop
 
     override fun toString(): String {
@@ -192,21 +200,12 @@ public class DigestCredential(
     }
 }
 
+internal val DigestCredential.digester
+    get() = digestAlgorithm.toDigester()
+
 // helper for deprecated methods
 private fun String?.toDigestAlgorithm(): DigestAlgorithm =
     requireNotNull(DigestAlgorithm.from(this ?: "MD5")) { "Unsupported algorithm: $this" }
-
-/**
- * Creates a [MessageDigest] instance for this algorithm.
- *
- * @return A new MessageDigest configured for this algorithm's hash function
- * @throws [java.security.NoSuchAlgorithmException] If the algorithm is not supported by the JVM
- */
-public fun DigestAlgorithm.toDigester(): MessageDigest =
-    MessageDigest.getInstance(hashName)
-
-internal val DigestCredential.digester
-    get() = digestAlgorithm.toDigester()
 
 /**
  * Converts [HttpAuthHeader] to [DigestCredential].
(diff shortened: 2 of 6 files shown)


zibet27:
@coderabbitai review

ktor-server/ktor-server-plugins/ktor-server-auth/jvm/src/io/ktor/server/auth/DigestAlgorithm.kt
```diff
@@ -0,0 +1,95 @@
...
+ *
+ * **Security Recommendation**: Use [SHA_512_256] or [SHA_512_256_SESS] for new implementations.
+ * These provide the strongest security properties. MD5 variants are deprecated and should only
+ * be used for backward compatibility with legacy systems.
+ *
+ * @property algorithmName The name of the algorithm as it appears in HTTP headers (e.g., "MD5", "SHA-256-sess")
+ * @property hashName The Java Security algorithm name used for MessageDigest (e.g., "MD5", "SHA-256")
+ * @property isSession Whether this is a session variant that incorporates nonce and cnonce into HA1
+ */
+public enum class DigestAlgorithm(
```
bjhham: It may be better to use a class with companion object fields here so users can extend with custom algorithms as needed.

ktor-http/common/src/io/ktor/http/auth/HttpAuthHeader.kt
```diff
@@ -425,7 +425,10 @@ public sealed class HttpAuthHeader(public val authScheme: String) {
             domain: List<String> = emptyList(),
             opaque: String? = null,
             stale: Boolean? = null,
-            algorithm: String = "MD5"
+            algorithm: String = "MD5",
```
osipxd: Should we change the default algorithm to SHA-256?

ktor-server/ktor-server-plugins/ktor-server-auth/jvm/src/io/ktor/server/auth/DigestAuth.kt
```diff
@@ -4,60 +4,114 @@
...
     private val nonceManager: NonceManager = config.nonceManager
 
-    private val userNameRealmPasswordDigestProvider: DigestProviderFunction = config.digestProvider
+    private val userNameRealmPasswordDigestProvider: DigestProviderFunctionV2 = config.digestProvider
 
     private val authenticationFunction: AuthenticationFunction<DigestCredential> = config.authenticationFunction
 
+    init {
+        if (config.strictRfc7616Mode) {
+            require(algorithms.none { it == DigestAlgorithm.MD5 || it == DigestAlgorithm.MD5_SESS }) {
```
osipxd: If the `algorithms` list is empty, it will pass this check, but the `defaultAlgorithm` will be MD5

ktor-server/ktor-server-plugins/ktor-server-auth/jvm/src/io/ktor/server/auth/DigestAlgorithm.kt
```diff
@@ -0,0 +1,95 @@
...
+    public val algorithmName: String,
+    public val hashName: String,
+    public val isSession: Boolean
+) {
+    /** MD5 algorithm - deprecated, use only for backward compatibility with legacy systems */
+    MD5("MD5", "MD5", false),
+
+    /** MD5 session variant - deprecated, use only for backward compatibility with legacy systems */
+    MD5_SESS("MD5-sess", "MD5", true),
```
osipxd: Let's add the `@Deprecated` annotation to these algorithms

ktor-server/ktor-server-plugins/ktor-server-auth/jvm/src/io/ktor/server/auth/DigestAlgorithm.kt
```diff
@@ -0,0 +1,95 @@
...
+
+    /** SHA-512/256 session variant - **recommended for new implementations**, provides the strongest security */
+    SHA_512_256_SESS("SHA-512-256-sess", "SHA-512/256", true);
+
+    /**
+     * Creates a [MessageDigest] instance for this algorithm.
+     *
+     * @return A new MessageDigest configured for this algorithm's hash function
+     */
+    public fun toDigester(): MessageDigest = MessageDigest.getInstance(hashName)
```
osipxd: Let's make it an extension-function this way it will be easier to move the `DigestAlgorithm` enum into the common source set (if demand appears)

ktor-server/ktor-server-plugins/ktor-server-auth/jvm/src/io/ktor/server/auth/DigestCredential.kt
```diff
@@ -0,0 +1,272 @@
...
+    val algorithm: String?,
+    val response: String,
+    val cnonce: String?,
+    val qop: String?,
+    val userHash: Boolean = false,
+    val charset: String? = null,
+) {
+    internal val digestAlgorithm = algorithm?.let { DigestAlgorithm.from(it) } ?: DigestAlgorithm.MD5
+
+    internal val digester = digestAlgorithm.toDigester()
```
osipxd: And this field also can be made an extension property

ktor-server/ktor-server-plugins/ktor-server-auth/jvm/src/io/ktor/server/auth/DigestAuth.kt
```diff
@@ -119,7 +208,75 @@ public class DigestAuthenticationProvider internal constructor(
...
+         * - `SHA-256-sess` session variant
+         * - `SHA-512-256` **recommended for new implementations** (provides the strongest security)
+         * - `SHA-512-256-sess` session variant with the strongest security
+         *
+         * When multiple algorithms are configured, the server will send multiple
+         * `WWW-Authenticate` headers, one per algorithm, allowing the client to choose.
+         *
+         * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.server.auth.DigestAuthenticationProvider.Config.algorithms)
+         */
+        public var algorithms: List<DigestAlgorithm> = listOf(DigestAlgorithm.MD5)
```
osipxd: Do you think it's possible to change the default to `SHA-256`? What will happen to existing users?

ktor-server/ktor-server-plugins/ktor-server-auth/jvm/src/io/ktor/server/auth/DigestCredential.kt
```diff
@@ -0,0 +1,272 @@
...
+ * @property opaque A string of data that should be returned by the client unchanged
+ * @property nonceCount The request counter (nc), must be sent if [qop] is specified
+ * @property algorithm The digest algorithm name (e.g., "SHA-512-256", "SHA-256", "MD5")
+ * @property response The client's digest response, consisting of hex digits
+ * @property cnonce Client nonce, must be sent if [qop] is specified
+ * @property qop Quality of protection ("auth" or "auth-int")
+ * @property userHash Whether the [userName] field contains a hashed username instead of plaintext
+ * @property charset The character encoding used (typically "UTF-8" or "ISO-8859-1")
+ */
+public data class DigestCredential(
```
osipxd: One possible way to keep binary compatibility would be to avoid using data class here and provide implementations for `component*` methods, `copy`, `hashCode`, `toString` and `equals`. It might make sense to deprecate `component` methods as well.
This feels cumbersome though

ktor-server/ktor-server-plugins/ktor-server-auth/jvm/src/io/ktor/server/auth/DigestCredential.kt
```diff
@@ -0,0 +1,272 @@
...
+    val digestUri: String,
+    val nonce: String,
+    val opaque: String?,
+    val nonceCount: String?,
+    val algorithm: String?,
+    val response: String,
+    val cnonce: String?,
+    val qop: String?,
+    val userHash: Boolean = false,
+    val charset: String? = null,
```
osipxd: Minor. Maybe make this field of type `Charset` and remove `digestCharset`?

ktor-server/ktor-server-plugins/ktor-server-auth/jvm/src/io/ktor/server/auth/DigestCredential.kt
zibet27: But then it will create a new digester every time.
Though it's not much overhead.

ktor-server/ktor-server-plugins/ktor-server-auth/jvm/src/io/ktor/server/auth/DigestAuth.kt
zibet27: It would break the compatibility.
If the client wants to use MD5(which is usually the case) `algorithms.any { it == algorithm }` will return false.
RFC7616 is not widely adopted yet. 
Even Spring doesn't support it because digest auth is usually not very secure. 
Though RFC7616 tries to fix that.

ktor-server/ktor-server-plugins/ktor-server-auth/jvm/src/io/ktor/server/auth/DigestAlgorithm.kt
zibet27: afaik RFC2617 and RFC7616 describe only those algorithms
Do we need it then?

ktor-server/ktor-server-plugins/ktor-server-auth/jvm/src/io/ktor/server/auth/DigestCredential.kt
zibet27: Ohh, I forgot that `charset` wasn't there before, so I can remove it.

ktor-http/common/src/io/ktor/http/auth/HttpAuthHeader.kt
zibet27: I would love to, but it would be a breaking change.
Maybe in version 4.

ktor-server/ktor-server-plugins/ktor-server-auth/jvm/src/io/ktor/server/auth/DigestAuth.kt
osipxd: What if we set the default to `listOf(DigestAlgorithm.SHA_256, DigestAlgorithm.MD5)`?
It should be compatible with MD5, while use SHA-256 as a default option

ktor-server/ktor-server-plugins/ktor-server-auth/jvm/src/io/ktor/server/auth/DigestAuth.kt
```diff
@@ -119,7 +208,75 @@ public class DigestAuthenticationProvider internal constructor(
...
+         * When enabled:
+         * - MD5 and MD5-sess algorithms are prohibited (SHA-256 or SHA-512-256 required)
+         * - UTF-8 charset is required
+         *
+         * **Recommendation**: Enable this mode for new implementations to enforce modern security standards.
+         * Use SHA-512-256 for optimal security when this mode is enabled.
+         *
+         * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.server.auth.DigestAuthenticationProvider.Config.strictRfc7616Mode)
+         */
+        public var strictRfc7616Mode: Boolean = false
```
osipxd: Probably we can turn this field into a method and change the default algorithms to `SHA-256` and the charset to UTF-8 when it is called (and only if these properties wasn't configured). To make it easier to use this mode

ktor-server/ktor-server-plugins/ktor-server-auth/jvm/src/io/ktor/server/auth/DigestAuth.kt
osipxd: Btw, would it be a good idea to also include `SHA_512_256` in the list by default?

zibet27: Yes, I believe it's better than `SHA_256`, because it's even more secure and faster on 64-bit platforms.

ktor-server/ktor-server-plugins/ktor-server-auth/jvm/test/io/ktor/tests/auth/DigestCurlComplianceTest.kt
```diff
@@ -0,0 +1,544 @@
...
+ * Black Box Compliance Test Suite for Server Digest Authentication.
+ *
+ * These tests verify that Ktor's implementation correctly interoperates with the curl client, ensuring compliance with:
+ * - RFC 7616 (HTTP Digest Access Authentication) - Modern standard with SHA-256/SHA-512-256
+ * - RFC 2617 (HTTP Authentication) - Legacy standard with MD5 for backward compatibility
+ */
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+class DigestCurlComplianceTest {
+
+    companion object {
```
osipxd: 💅 Minor
I believe we usually put companions at the end of the file according to Kotlin Coding Convention. And implementation details after functions they're used in. However not sure if the last rule is written down somewhere. Just intuition based on the existing codebase

ktor-server/ktor-server-plugins/ktor-server-auth/jvm/src/io/ktor/server/auth/DigestAlgorithm.kt
osipxd: Btw. If we would ever add support for different digest algorithms to clients, these enums could be reused there. In this case we might want to put them into `io.ktor.http.auth` package (common).

But I agree with Bruce. At some point some new RFC could introduce new algorithms. So `companion object` might be a better choice for API extensibility

ktor-server/ktor-server-plugins/ktor-server-auth/jvm/src/io/ktor/server/auth/DigestCredential.kt
```diff
@@ -0,0 +1,351 @@
...
+    public operator fun component1(): String = realm
+    public operator fun component2(): String = userName
+    public operator fun component3(): String = digestUri
+    public operator fun component4(): String = nonce
+    public operator fun component5(): String? = opaque
+    public operator fun component6(): String? = nonceCount
+    public operator fun component7(): String? = algorithm
+    public operator fun component8(): String = response
+    public operator fun component9(): String? = cnonce
+    public operator fun component10(): String? = qop
```
osipxd, line 158: Should we deprecate `component*` and `copy` functions?

ktor-server/ktor-server-plugins/ktor-server-auth/jvm/src/io/ktor/server/auth/DigestCredential.kt
```diff
@@ -0,0 +1,351 @@
...
+ * @property charset The character encoding used (typically "UTF-8" or "ISO-8859-1")
+ */
+public class DigestCredential(
+    public val realm: String,
+    public val userName: String,
+    public val digestUri: String,
+    public val nonce: String,
+    public val opaque: String?,
+    public val nonceCount: String?,
+    public val algorithm: String?,
```
osipxd: Btw, after making this class plain class can safely change the type of this field to `DigestAlgorithm`.
All we need to do is to:
1. Introduce a field of new type and provide `algorithm` field implementation
2. Change deprecation level for the old constructor to warning with `ReplaceWith` (for source compatibility)
3. Fix implementation of `copy`, `component7` and `toString`.

In this case new constructor could have `Charsets.UTF_8` as a default. And `Charsets.ISO_8859_1` would've been passed in the old constructor.

However it might be not so simple to come up with a good name here.

ktor-server/ktor-server-plugins/ktor-server-auth/jvm/src/io/ktor/server/auth/DigestAuth.kt
```diff
@@ -4,32 +4,52 @@
...
 public class DigestAuthenticationProvider internal constructor(
     config: Config
 ) : AuthenticationProvider(config) {
 
     private val realm: String = config.realm
 
-    private val algorithmName: String = config.algorithmName
+    private val algorithms: List<DigestAlgorithm> = config.algorithms
+
+    private val defaultAlgorithm: DigestAlgorithm = algorithms.firstOrNull() ?: DigestAlgorithm.SHA_512_256
```
osipxd: Maybe verify the list is not empty? Is there any use case when it's useful to set empty algorithms list?

ktor-server/ktor-server-plugins/ktor-server-auth/jvm/src/io/ktor/server/auth/DigestAuth.kt
```diff
@@ -143,11 +278,78 @@ public class DigestAuthenticationProvider internal constructor(
...
+         * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.server.auth.DigestAuthenticationProvider.Config.strictRfc7616Mode)
+         */
+        public fun strictRfc7616Mode() {
+            if (algorithms != defaultAlgorithms) {
+                LOGGER.warn("Defined algorithms are overridden in strictRfc7616Mode")
+            }
+            if (charset != null && charset != Charsets.UTF_8) {
+                LOGGER.warn("Defined charset is overridden in strictRfc7616Mode")
+            }
+            algorithms = strictAlgorithms
```
osipxd: User might have configured `algorithms = listOf(SHA_512_256)`. I think we shouldn't override user config in this case. Let's override user config only. if it contains MD5.
```kotlin
if (MD5 in algorithms) {
    if (algorithms !== defaultAlgorithms) {
        LOGGER.warn("...")
    }
    algorithms -= MD5
}
```
(I think `!==` should be used instead of `!=` to check it's actually the default list)

ktor-server/ktor-server-plugins/ktor-server-auth/jvm/src/io/ktor/server/auth/DigestAuth.kt
zibet27: No, I think it's better to be explicit here and throw an exception when an empty list is provided.

ktor-server/ktor-server-plugins/ktor-server-auth/jvm/src/io/ktor/server/auth/DigestCredential.kt
zibet27: We might need a copy function when resolving hashed username

ktor-server/ktor-server-plugins/ktor-server-auth/jvm/src/io/ktor/server/auth/DigestAuth.kt
```diff
@@ -119,7 +199,62 @@ public class DigestAuthenticationProvider internal constructor(
          *
          * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.server.auth.DigestAuthenticationProvider.Config.algorithmName)
          */
-        public var algorithmName: String = "MD5"
+        @Deprecated("Use algorithms instead", ReplaceWith("algorithms"))
+        public var algorithmName: String?
```
e5l: adding nullability is a source breaking change

e5l (changes requested):
please fix the source breaking changes

ktor-server/ktor-server-plugins/ktor-server-auth/jvm/src/io/ktor/server/auth/DigestAuth.kt
```diff
@@ -334,19 +322,19 @@ public class DigestAuthenticationProvider internal constructor(
          * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.server.auth.DigestAuthenticationProvider.Config.strictRfc7616Mode)
          */
         public fun strictRfc7616Mode() {
-            if (algorithms != defaultAlgorithms) {
-                LOGGER.warn("Defined algorithms are overridden in strictRfc7616Mode")
+            @Suppress("DEPRECATION")
+            if (DigestAlgorithm.MD5 in algorithms) {
```
osipxd: Or MD5_SESS

ktor-server/ktor-server-plugins/ktor-server-auth/jvm/src/io/ktor/server/auth/DigestCredential.kt
```diff
@@ -56,99 +56,154 @@ public class DigestCredential(
         response: String,
         cnonce: String?,
         qop: String?,
-    ) : this(realm, userName, digestUri, nonce, opaque, nonceCount, algorithm, response, cnonce, qop)
+    ) : this(
+        realm, userName, digestUri, nonce, opaque, nonceCount, algorithm.toDigestAlgorithm(), response, cnonce, qop
+    )
+
+    @Deprecated("Maintained for binary compatibility", level = DeprecationLevel.HIDDEN)
+    public val DigestCredential.algorithm: String?
```
osipxd: Can we drop receiver here?
Unfortunately we can't make this field hidden as it would be source-breaking (yet binary compatible) change

ktor-server/ktor-server-plugins/ktor-server-auth/jvm/src/io/ktor/server/auth/DigestCredential.kt
```diff
@@ -56,99 +56,154 @@ public class DigestCredential(
...
+        get() = digestAlgorithm.name
 
-    /**
-     * Creates a copy of this DigestCredential, optionally replacing some properties.
-     */
+    @Deprecated("Maintained for binary compatibility", level = DeprecationLevel.HIDDEN)
```
osipxd: The same here. We can't make it hidden. And for all other deprecations here we should  use `WARNING`

ktor-server/ktor-server-plugins/ktor-server-auth/jvm/src/io/ktor/server/auth/DigestCredential.kt
```diff
@@ -56,99 +56,154 @@ public class DigestCredential(
...
+        userName,
+        digestUri,
+        nonce,
+        opaque,
+        nonceCount,
+        digestAlgorithm,
+        response,
+        cnonce,
+        qop,
+        userHash,
```
osipxd: 💅  For functions containing a lot of arguments the same type it's safer to add argument names even if they match to the value:
```
realm = realm,
userName = userName,
// etc.
```
This way we will be sure that values are mapped correctly.
The same could be applied to the secondary constructor here

ktor-server/ktor-server-plugins/ktor-server-auth/jvm/src/io/ktor/server/auth/DigestCredential.kt
```diff
@@ -56,99 +56,154 @@ public class DigestCredential(
...
+private fun String?.toDigestAlgorithm(): DigestAlgorithm =
+    requireNotNull(DigestAlgorithm.from(this ?: "MD5")) { "Unsupported algorithm: $this" }
+
+/**
+ * Creates a [MessageDigest] instance for this algorithm.
+ *
+ * @return A new MessageDigest configured for this algorithm's hash function
+ * @throws [java.security.NoSuchAlgorithmException] If the algorithm is not supported by the JVM
+ */
+public fun DigestAlgorithm.toDigester(): MessageDigest =
```
osipxd: Minor. Probably should belong to `DigestAlgorithm.jvm.kt` in the `ktor-http` module

ktor-server/ktor-server-plugins/ktor-server-auth/jvm/src/io/ktor/server/auth/DigestCredential.kt
```diff
@@ -169,21 +224,26 @@ public fun HttpAuthHeader.Parameterized.toDigestCredential(
...
+    val algorithmName = parameter("algorithm") ?: "MD5"
+    val algorithm = DigestAlgorithm.from(algorithmName) ?: error("Invalid algorithm in digest authentication header")
+    val charsetParam = parameter("charset")
+    require(charsetParam == null || charsetParam == Charsets.UTF_8.name()) {
+        "Unsupported charset in digest authentication header"
+    }
 
     return DigestCredential(
-        parameter("realm")!!,
+        parameter("realm") ?: error("Missing realm parameter in digest authentication header"),
```
osipxd: Minor. `checkNotNull`/`requireNotNull` could be used

ktor-server/ktor-server-plugins/ktor-server-auth/jvm/src/io/ktor/server/auth/DigestCredential.kt
```diff
@@ -37,13 +36,14 @@ public class DigestCredential(
...
+    public val digestAlgorithm: DigestAlgorithm,
     public val response: String,
     public val cnonce: String?,
     public val qop: String?,
     public val userHash: Boolean = false,
-    public val charset: Charset = Charsets.ISO_8859_1
+    public val charset: Charset = Charsets.UTF_8
 ) {
+
     @Deprecated("Maintained for binary compatibility", level = DeprecationLevel.HIDDEN)
```
osipxd: And here WARNING should be used to preserve source compatibility

zibet27:
@osipxd When should we use `DeprecationLevel.HIDDEN` and when `DeprecationLevel.WARNING`?

osipxd:
@zibet27, when we add new parameters (with default values) this change is source compatible, but binary incompatible. In this case we can use `DeprecationLevel.HIDDEN` to preserve binary compatibility. Functions annotated with this deprecation level won't be available from sources, but still can be used by already compiled code.
Another case is when we want to keep old signature for source compatibility, and introduce a new function with different set of parameters (like with `copy` and secondary constructor here), or just want to discourage users from using this function (like with `component*` functions).

e5l (approved):
lgtm

PR #5353, opened 2026-02-05
KTOR-9318 Do not validate server certificate against handshake algorithms

osipxd:
**Subsystem**
CIO Client

**Motivation**
[KTOR-9318](https://youtrack.jetbrains.com/issue/KTOR-9318) CIO engine rejects valid certificates with unsupported signature algorithms
[KTOR-9347](https://youtrack.jetbrains.com/issue/KTOR-9347) Fix flaky client cert test

**Solution**
Root cause is an extra post-validation filter in CIO TLS handshake code: after the TrustManager verifies the chain, CIO tries to pick a suitable leaf certificate by comparing the certificate’s signature algorithm OID (`X509Certificate.sigAlgOID`, i.e., how the CA signed the certificate) against a hardcoded `SupportedSignatureAlgorithms` list meant for TLS handshake signature negotiation. This conflates two different concepts and incorrectly rejects otherwise-valid certificates whose `sigAlgOID` is not in that list.

### DIFF
--- a/ktor-server/ktor-server-test-suites/jvm/src/io/ktor/server/testing/suites/ClientCertTestSuite.kt
+++ b/ktor-server/ktor-server-test-suites/jvm/src/io/ktor/server/testing/suites/ClientCertTestSuite.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2025 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ * Copyright 2014-2026 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
  */
 
 package io.ktor.server.testing.suites
@@ -16,7 +16,6 @@ import io.ktor.server.routing.*
 import kotlinx.coroutines.debug.junit5.CoroutinesTimeout
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.runBlocking
-import kotlin.test.Ignore
 import kotlin.test.Test
 import kotlin.test.assertEquals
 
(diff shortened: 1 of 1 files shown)


ktor-network/ktor-network-tls/jvm/src/io/ktor/network/tls/TLSClientHandshake.kt
```diff
@@ -248,13 +255,7 @@ internal class TLSClientHandshake(
...
                     val manager = config.trustManager
                     manager.checkServerTrusted(x509s.toTypedArray(), exchangeType.jvmName)
-
-                    serverCertificate = x509s.firstOrNull { certificate ->
-                        SupportedSignatureAlgorithms.any {
-                            val oid = it.oid?.identifier ?: return@any false
-                            oid.equals(certificate.sigAlgOID, ignoreCase = true)
-                        }
-                    } ?: throw TLSException("No suitable server certificate received: $certs")
+                    serverCertificate = x509s.first()
```
osipxd, line 258: Don't know how to test this change
@coderabbitai do you have any ideas?

ktor-network/ktor-network-tls/common/src/io/ktor/network/tls/OID.kt
```diff
@@ -59,7 +59,7 @@ public data class OID(public val identifier: String) {
             "SHA256withECDSA" -> ECDSAwithSHA256Encryption
             "SHA384withRSA" -> RSAwithSHA384Encryption
             "SHA256withRSA" -> RSAwithSHA256Encryption
-            else -> error("Could't find OID for $algorithm")
+            else -> error("Couldn't find OID for $algorithm")
```
bjhham, line 62: 🤦

ktor-network/ktor-network-tls/jvm/src/io/ktor/network/tls/TLSClientHandshake.kt
bjhham, line 258: Wouldn't we want to keep `firstOrNull() ?: throw TLSException("No suitable server certificate received: $certs")` here?  There may be some exception handling for the `TLSException`, I'm not sure.

bjhham (approved):

ktor-network/ktor-network-tls/jvm/src/io/ktor/network/tls/TLSClientHandshake.kt
osipxd: Yes, also thought about it. And noticed there was a check `if (x509s.isEmpty()) throw TLSException("Server sent no certificate")` a few lines before

PR #5354, opened 2026-02-05
Added support for the new ABI validation DSL in KGP 2.4.0

shanshin:


woainikk:
cherry picked manually

PR #5355, opened 2026-02-05
KTOR-9267 Curl: Handle chunked websocket frames

osipxd:
**Subsystem**
kotlin-client-curl

**Motivation**
[KTOR-9267](https://youtrack.jetbrains.com/issue/KTOR-9267) Curl: Incorrect WebSockets frame fragmentation

**Solution**
Read `meta.offset` and `meta.bytesleft` to determine chunked frames. Collect frame data into a buffer and send at once.


osipxd:
@bjhham, I've implemented a fix but I'd like to open another PR for it. Could you approve this one?
Here is an issue to track: [KTOR-9327](https://youtrack.jetbrains.com/issue/KTOR-9327) Curl: Support WebSockets maxFrameSize option

PR #5356, opened 2026-02-09
KTOR-9276 Dynamic auth provider suspend authenticate

zibet27:
**Subsystem**
Server Auth

**Motivation**
[KTOR-9276](https://youtrack.jetbrains.com/issue/KTOR-9276) Make DynamicProviderConfig.authenticateFunction suspend

**Solution**
- add `DynamicProviderConfig.authenticateSuspend`

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-auth/common/test/io/ktor/tests/auth/AuthBuildersTest.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-auth/common/test/io/ktor/tests/auth/AuthBuildersTest.kt
@@ -14,10 +14,12 @@ import io.ktor.server.response.*
 import io.ktor.server.routing.*
 import io.ktor.server.sessions.*
 import io.ktor.server.testing.*
-import io.ktor.util.*
 import io.ktor.utils.io.InternalAPI
+import kotlinx.coroutines.delay
 import kotlinx.serialization.*
+import kotlin.io.encoding.Base64
 import kotlin.test.*
+import kotlin.time.Duration.Companion.milliseconds
 
 class AuthBuildersTest {
 
@@ -92,7 +94,7 @@ class AuthBuildersTest {
             header(HttpHeaders.ContentType, ContentType.Application.FormUrlEncoded.toString())
             header(
                 HttpHeaders.Authorization,
-                HttpAuthHeader.Single("basic", "second:".encodeBase64()).render()
+                HttpAuthHeader.Single("basic", Base64.encode("second:".encodeToByteArray())).render()
             )
         }.let { call ->
             assertEquals(HttpStatusCode.OK, call.status)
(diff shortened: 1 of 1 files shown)


ktor-server/ktor-server-plugins/ktor-server-auth/common/src/io/ktor/server/auth/DynamicProviderConfig.kt
```diff
@@ -14,19 +14,26 @@ public class DynamicProviderConfig(
...
+    private var authenticateFunction: ((context: AuthenticationContext) -> Unit)? = null
+    private var suspendAuthenticateFunction: (suspend (context: AuthenticationContext) -> Unit)? = null
 
     public fun authenticate(block: (context: AuthenticationContext) -> Unit) {
         authenticateFunction = block
     }
 
+    public fun authenticateSuspend(block: suspend (context: AuthenticationContext) -> Unit) {
+        suspendAuthenticateFunction = block
+    }
```
bjhham, line 26: To avoid complicating the signature here, we can instead go with something like this:

```
private lateinit var authenticateFunction: suspend (context: AuthenticationContext) -> Unit

@Deprecated("Use suspend argument", level = DeprecationLevel.HIDDEN)
public fun authenticate(block: (context: AuthenticationContext) -> Unit) {
    authenticateFunction = { ctx -> block(ctx) }
}

public fun authenticate(block: suspend (context: AuthenticationContext) -> Unit) {
    authenticateFunction = block
}
```

This might be technically breaking in some cases, but we've done it in the past for other DSLs.

zibet27: I've tried to name both of them `authenticate`, but it didn't compile, because I didn't annotate one with `@Deprecated`
Thanks, now I know deprecated changes the priority of the method)

ktor-server/ktor-server-plugins/ktor-server-auth/common/src/io/ktor/server/auth/DynamicProviderConfig.kt
```diff
@@ -14,9 +14,14 @@ public class DynamicProviderConfig(
...
 
-    private lateinit var authenticateFunction: (context: AuthenticationContext) -> Unit
+    private lateinit var authenticateFunction: suspend (context: AuthenticationContext) -> Unit
 
+    @Deprecated("Use suspend argument", level = DeprecationLevel.HIDDEN)
     public fun authenticate(block: (context: AuthenticationContext) -> Unit) {
+        authenticateFunction = { ctx -> block(ctx) }
+    }
+
+    public fun authenticate(block: suspend (context: AuthenticationContext) -> Unit) {
```
e5l, line 24: Please consider adding test for the new method

PR #5357, opened 2026-02-09
KTOR-9304 Fix ordering of parameters by order of appearance

bjhham:
**Subsystem**
Server, OpenAPI

**Motivation**
[KTOR-9304](https://youtrack.jetbrains.com/issue/KTOR-9304) OpenAPI: Order of path parameters is not preserved in the spec

**Solution**
Reversed the order of traversal for routes when populating the OpenAPI documentation model.  Before, we were going from bottom to top and retaining data, but now, we'll go top to bottom and replace values.  This fixes the positioning of the parameters.


zibet27:
Great! Please update older tests.

bjhham:
Turns out the test failure caught a mistake I made when rebasing 😄

PR #5358, opened 2026-02-10
Server Sessions. Create sessions based on call; clear session by id.

zibet27:
**Subsystem**
Server Sessions

**Motivation**
(https://youtrack.jetbrains.com/issue/KTOR-8300)

**Solution**
- new api:

```kotlin
install(Sessions) {
   cookie<TestUserSession>(cookieName, sessionStorage) {
       // identity provider based on call
       identity { call ->
         val userAgent = call.request.headers["User-Agent"]!!
         "custom-${customIdCounter++}-${userAgent.take(5)}"
       }
   }
}
```

```kotlin
// delete session by id
call.sessions.clear<SessionType>(sessionId)
```

The only problem is breaking the change in the following getter return type
```kotlin
public var sessionIdProvider: (ApplicationCall) -> String = { generateSessionId() }
        private set
```
We can't overload on return type.


ktor-server/ktor-server-plugins/ktor-server-sessions/common/src/io/ktor/server/sessions/SessionData.kt
```diff
@@ -51,6 +51,18 @@ public interface CurrentSession {
...
+     * Clears a session with the specified [sessionId] for the session [name].
+     *
+     * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.server.sessions.CurrentSession.clear)
+     *
+     * @param name the session name
+     * @param sessionId the session ID to invalidate
+     * @throws IllegalStateException if no session provider is registered with for [name]
+     * or the session provider doesn't use session IDs
+     */
+    public suspend fun clear(name: String, sessionId: String)
```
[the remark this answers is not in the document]
bjhham: Yeah providing a default here might be wise.

ktor-server/ktor-server-plugins/ktor-server-sessions/common/src/io/ktor/server/sessions/SessionsBuilder.kt
bjhham, line 351: We _can_ avoid the breaking change here with something like this:

```
private var _sessionIdProvider: (ApplicationCall?) -> String = { generateSessionId() }

public var sessionIdProvider: () -> String
    get() = { _sessionIdProvider(null) }
    private set(value) {
        _sessionIdProvider = { _ -> value() }
    }

public fun identity(f: () -> String) {
    sessionIdProvider = f
}

public fun identity(f: (ApplicationCall) -> String) {
    _sessionIdProvider = { call ->
        f(requireNotNull(call) { "ApplicationCall is required for this sessionIdProvider" })
    }
}

// internal (or used by tracker) when call is available:
internal fun sessionIdProvider(call: ApplicationCall): String = _sessionIdProvider(call)
```

Though imo it's better to just accept a breaking change for a variable that nobody would ever read anyway.

ktor-server/ktor-server-tests/jvm/test/io/ktor/server/sessions/SessionTestJvm.kt
```diff
@@ -10,6 +10,7 @@ import io.ktor.http.*
 import io.ktor.server.response.*
 import io.ktor.server.routing.*
 import io.ktor.server.testing.*
+import io.ktor.tests.server.sessions.*
```
bjhham, line 13: Erroneous import here.

ktor-server/ktor-server-plugins/ktor-server-sessions/common/src/io/ktor/server/sessions/SessionsBuilder.kt
zibet27: We ought to make the field `internal` in the next major version

ktor-server/ktor-server-tests/jvm/test/io/ktor/server/sessions/SessionTestJvm.kt
zibet27: The import is required here, because I fixed the incorrect package in https://github.com/ktorio/ktor/pull/5358/changes#diff-71a849a34e0512268e8812e6ad3b11f594496e4fc7f94aeb1c474de2ec15bd43R5

It looks like tests in the jvm target have different package than common (`io.ktor.server` instead of `io.ktor.tests.server`)
I can make them identical, but I didn't want to add noise in the PR. Then we won't need such imports

ktor-server/ktor-server-plugins/ktor-server-sessions/common/src/io/ktor/server/sessions/SessionsBuilder.kt
```diff
@@ -456,15 +461,20 @@ internal constructor(
...
+        sessionIdProvider = { f() }
+    }
+
     /**
      * A function used to provide a current session ID.
      *
      * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.server.sessions.HeaderIdSessionBuilder.sessionIdProvider)
      */
-    public var sessionIdProvider: () -> String = { generateSessionId() }
+    public var sessionIdProvider: (ApplicationCall) -> String = { generateSessionId() }
```
e5l: We have to keep this property as is; it can be proxied to the new property if needed.

I would make a new `@InternalAPI` for that. It would also be nice to avoid the functional property type and use (functional) interface, so we can rely on default arguments or overloads in the future

ktor-server/ktor-server-plugins/ktor-server-sessions/common/src/io/ktor/server/sessions/SessionTrackerById.kt
```diff
@@ -82,14 +82,26 @@ internal fun ApplicationCall.sessionId(name: String): String? {
...
  * @property storage session storage to store session
- * @property sessionIdProvider is a function that generates session IDs
+ * @property sessionIdProvider is a function that generates session IDs based on the [ApplicationCall]
  */
 public class SessionTrackerById<S : Any>(
     public val type: KClass<S>,
     public val serializer: SessionSerializer<S>,
     public val storage: SessionStorage,
-    public val sessionIdProvider: () -> String
+    public val sessionIdProvider: (ApplicationCall) -> String
```
e5l: could we also avoid this breaking change?

PR #5360, opened 2026-02-10
Move decompression to earlier phase in receivePipeline.

ascheja:
**Subsystem**
Server, compression

**Motivation**
See [KTOR-6683](https://youtrack.jetbrains.com/issue/KTOR-6683).

Currently content decoding only works in cases where the application directly receives a `ByteReadChannel`. When combining it with any default transformations (`ByteArray`, `Parameters`, `MultiPartData`, ...) the content decoding would silently not work because the Compression's `transformBody ` would already get the converted subject from the automatically installed default transformations. The same principle applies to the `ContentNegotation` plugin if it is installed before the `Compression` plugin, although in that case the transformation would likely fail.

**Solution**
Simply moves the content decoding into it's own phase right before the default `Transform` phase.

### DIFF
--- a/ktor-server/ktor-server-tests/jvm/test/io/ktor/server/plugins/CompressionTest.kt
+++ b/ktor-server/ktor-server-tests/jvm/test/io/ktor/server/plugins/CompressionTest.kt
@@ -909,7 +909,7 @@ class CompressionTest {
                 val body = call.receive<ByteArray>()
 
                 assertNull(call.request.headers[HttpHeaders.ContentEncoding])
-                assertContentEquals(compressed, body)
+                assertContentEquals(textToCompressAsBytes, body)
 
                 call.respond(textToCompressAsBytes)
             }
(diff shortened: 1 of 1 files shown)


bjhham:
Looks like this introduces a couple failures in `io.ktor.server.plugins.CompressionTest`. Please take a look 👀

ascheja:
Sorry, overlooked that test class as it was in a completely different subproject. Moved my tests over to it and adjusted the assertions for the two failing tests (both received as `ByteArray`, but expected compressed content even though request body decompression was enabled)

bjhham:
Could you please run `./gradlew formatKotlin` and push to clear the linter check?

ascheja:
Did that, but it looks like the linter check is still failing?

ascheja:
Figured it out - the problem was a line with more than 120 characters, but `./gradlew formatKotlin` didn't fix it.

ascheja:
@bjhham Is there anything missing from my side for approval?

bjhham:
Hey, thanks for the reminder, I went on vacation and forgot this one 😅 

All looks good 👍

PR #5362, opened 2026-02-12
Fix for missing oneOf in sealed types; allow for custom modules

bjhham:
**Subsystem**
Server, OpenAPI

**Motivation**
[KTOR-8913](https://youtrack.jetbrains.com/issue/KTOR-8913) OpenAPI: Polymorphic serialization is not respected by the compiler plugin
[KTOR-9330](https://youtrack.jetbrains.com/issue/KTOR-9330) OpenAPI: Cannot override kotlinx.serialization module

**Solution**
- Added oneOf for sealed types - previously only included `discriminator`.  We'll also populate each element's schema here properly so the references work.
- Extracted a class with a single `SerializersModule` field for the descriptor inference so users can supply their own.  This isn't a problem when using annotations, but people often use custom modules for overriding the default serialization behaviour.  This change should be ok for a patch because the describe API is all marked as experimental.

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-routing-openapi/common/src/io/ktor/server/routing/openapi/OperationMapping.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-routing-openapi/common/src/io/ktor/server/routing/openapi/OperationMapping.kt
@@ -159,19 +159,29 @@ public class CollectSchemaReferences(private val schemaToComponent: (JsonSchema)
 
     /**
      * We use the "title" field for referencing types to schema definitions.
+     *
+     * This applies a "depth-first" transformation on the schema to extract all nested references first.
      */
     private fun collectSchema(schema: JsonSchema): ReferenceOr<JsonSchema> {
-        return schemaToComponent(schema)?.let { ref ->
-            ReferenceOr.schema(ref)
-        } ?: ReferenceOr.value(
-            schema.copy(
-                allOf = schema.allOf?.map { it.mapToReference(::collectSchema) },
-                anyOf = schema.anyOf?.map { it.mapToReference(::collectSchema) },
-                oneOf = schema.oneOf?.map { it.mapToReference(::collectSchema) },
-                not = schema.not?.mapToReference(::collectSchema),
-                properties = schema.properties?.mapValues { (_, value) -> value.mapToReference(::collectSchema) },
-                items = schema.items?.mapToReference(::collectSchema),
-            )
+        val nestedSchema = schema.copy(
+            allOf = schema.allOf?.map { it.mapToReference(::collectSchema) },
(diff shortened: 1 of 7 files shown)


ktor-shared/ktor-openapi-schema/api/ktor-openapi-schema.api
```diff
@@ -974,8 +974,8 @@ public final class io/ktor/openapi/JsonSchemaInferenceKt {
...
-	public static synthetic fun jsonSchemaFromAnnotations$default (Ljava/util/List;Lkotlin/jvm/functions/Function1;Lio/ktor/openapi/JsonSchema$SchemaType;Ljava/lang/String;Ljava/util/List;Lio/ktor/openapi/ReferenceOr;Ljava/util/Map;Lio/ktor/openapi/AdditionalProperties;Ljava/util/List;Ljava/lang/String;Lio/ktor/openapi/JsonSchemaDiscriminator;ILjava/lang/Object;)Lio/ktor/openapi/JsonSchema;
+	public static final fun jsonSchemaFromAnnotations (Ljava/util/List;Lkotlin/jvm/functions/Function1;Lio/ktor/openapi/JsonSchema$SchemaType;Ljava/lang/String;Ljava/util/List;Lio/ktor/openapi/ReferenceOr;Ljava/util/Map;Lio/ktor/openapi/AdditionalProperties;Ljava/util/List;Ljava/lang/String;Lio/ktor/openapi/JsonSchemaDiscriminator;Ljava/util/List;)Lio/ktor/openapi/JsonSchema;
+	public static synthetic fun jsonSchemaFromAnnotations$default (Ljava/util/List;Lkotlin/jvm/functions/Function1;Lio/ktor/openapi/JsonSchema$SchemaType;Ljava/lang/String;Ljava/util/List;Lio/ktor/openapi/ReferenceOr;Ljava/util/Map;Lio/ktor/openapi/AdditionalProperties;Ljava/util/List;Ljava/lang/String;Lio/ktor/openapi/JsonSchemaDiscriminator;Ljava/util/List;ILjava/lang/Object;)Lio/ktor/openapi/JsonSchema;
```
bjhham, line 978: These include default values for the args so it's non-breaking.

osipxd: It's source-compatible, but not binary compatible. Should we keep this function with old signature and `DeprecationLevel.HIDDEN` to preserve binary compatibility as well?

bjhham: Ah yes, that's true.  I'll include the overload.

bjhham: Actually, it's marked with `@InternalAPI` so I'll forgo this.

bjhham:
Betrayed by the IDE auto-formatting on YAML :rage1:

PR #5364, opened 2026-02-13
KTOR-9334 Fix unconfined dispatcher in handler context

bjhham:
**Subsystem**
Server, Netty

**Motivation**
[KTOR-9334](https://youtrack.jetbrains.com/issue/KTOR-9334) Coroutines in route handlers are dispatched with Dispatchers.Unconfined since 3.2.0

**Solution**
Returned the netty dispatcher to the user context.  This was removed in 3.2.0 to allow for some flexibility with overriding via the application's coroutine context, but when unassigned, this would lead to the engine's dispatcher taking over.  The engine's dispatcher is unconfined, so you can get some unpredictability in the thread usage.

**Addendum**
I created a separate ticket to introduce thread affinity for Jetty, because it's always behaved this way and would require some more invasive surgery:
[KTOR-9345](https://youtrack.jetbrains.com/issue/KTOR-9345) Jetty engine call context is unconfined

### DIFF
--- a/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/CIO.kt
+++ b/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/CIO.kt
@@ -83,13 +83,15 @@ internal object NettyDispatcher : CoroutineDispatcher() {
 
     override fun dispatch(context: CoroutineContext, block: Runnable) {
         val nettyContext = context[CurrentContextKey]!!.context
-        try {
-            // This can throw an exception when dispatching coroutines
-            // after the netty channel is closed
-            nettyContext.executor().execute(block)
-        } catch (cause: Throwable) {
-            LOG.warn("Failed to dispatch; falling back to caller thread", cause)
-            block.run()
+        val executor = nettyContext.executor()
+        if (executor.isShuttingDown) {
+            Dispatchers.IO.dispatch(context, block)
+        } else {
+            try {
+                executor.execute(block)
+            } catch (cause: Throwable) {
+                LOG.error("Failed to dispatch", cause)
+            }
         }
     }
 


ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/CIO.kt
```diff
@@ -70,19 +70,26 @@ public suspend fun <T> Future<T>.suspendAwait(exception: (Throwable, Continuatio
...
+            // This can throw an exception when dispatching coroutines
+            // after the netty channel is closed
             nettyContext.executor().execute(block)
-        }
-
-        if (result.isFailure) {
-            LOG.error("Failed to dispatch", result.exceptionOrNull())
+        } catch (cause: Throwable) {
+            LOG.warn("Failed to dispatch; falling back to caller thread", cause)
+            block.run()
```
e5l: Retrying the execution block a second time can lead to unintended consequences (if there were no suspension points and the response/request was partially consumed/produced). Should we respond with an internal server error in this case and reject execution?

e5l (approved):
please check the edge case before merging

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/CIO.kt
bjhham: I noticed that this was the approach taken in the Jetty engine, and it seems to solve a few problems with structured concurrency according to the newer sustainability tests.  The main problem is that the netty executor quits before the application jobs join, so we need some kinda fallback for the shutdown sequence.  If I recall correctly, the executor is shut down early because it's tied to accepting new requests.  I'll try playing around with the fallback logic here to see if a secondary dispatcher would make more sense, or if we can avoid dropping the netty executor early.

bjhham: It just occurred to me that we can explicitly check if the executor is shutting down and delegate to the IO dispatcher in this case.  I've updated the dispatcher to do this instead, and restored the previous behavior of logging the exception for more mysterious cases.  This should avoid any blocking or problems with the shutdown sequence.

PR #5365, opened 2026-02-13
KTOR-9331 Curl: Fix segfaults when working with WebSockets

osipxd:
**Subsystem**
ktor-client-curl

**Motivation**
[KTOR-9331](https://youtrack.jetbrains.com/issue/KTOR-9331) Fix segfaults in CurlWebSocketTests

**Solution**
I've noticed two problems:
1. Callbacks: We were returning -1 from curl callbacks expecting `size_t` as a return type. Callback functions shouldn't throw any exceptions. Throwing exceptions in such context leads to undefined behavior
2. We were calling curl_ws_send from dispatcher other than "curl-dispatcher". It violated the contract that `EasyHandle` shouldn't be shared between threads. This was the most likely reason of segfaults

To fix the second problem I've refactored `CurlWebSocketSession` to use `CurlProcessor` for sending frames.

_It's better to review commit-by-commit._


ktor-client/ktor-client-core/posix/src/io/ktor/client/utils/CoroutineUtilsPosix.kt
```diff
@@ -1,20 +1,27 @@
...
  *
  * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.client.utils.clientDispatcher)
  */
+@OptIn(DelicateCoroutinesApi::class)
+@Deprecated(
+    "Use newFixedThreadPoolContext or Dispatcher.limitedParallelism instead",
+    ReplaceWith("newFixedThreadPoolContext(threads, name)", "kotlinx.coroutines.newFixedThreadPoolContext")
+)
 @InternalAPI
 public actual fun Dispatchers.clientDispatcher(
```
osipxd, line 20: I think it should be okay to deprecate this function in a patch release as it is marked with `@InternalAPI`

ktor-client/ktor-client-curl/desktop/src/io/ktor/client/engine/curl/internal/CurlMultiApiHandler.kt
```diff
@@ -349,6 +351,36 @@ internal class CurlMultiApiHandler : Closeable {
...
+                    if ((flags and CURLWS_CLOSE) == 0) {
+                        status.verify()
+                    }
+                }
+            }
+            completionHandler.complete()
+        } catch (cause: Throwable) {
+            completionHandler.completeExceptionally(cause)
+        }
+    }
```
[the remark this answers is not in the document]
osipxd: Added verification of the sent bytes count

ktor-client/ktor-client-core/posix/src/io/ktor/client/utils/CoroutineUtilsPosix.kt
```diff
@@ -1,20 +1,27 @@
...
 /**
  * Creates [CoroutineDispatcher] for client with fixed [threadCount] and specified [dispatcherName].
  *
  * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.client.utils.clientDispatcher)
  */
+@OptIn(DelicateCoroutinesApi::class)
+@Deprecated(
+    "Use newFixedThreadPoolContext or Dispatcher.limitedParallelism instead",
+    ReplaceWith("newFixedThreadPoolContext(threads, name)", "kotlinx.coroutines.newFixedThreadPoolContext")
+)
```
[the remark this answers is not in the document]
osipxd: Fixed

ktor-client/ktor-client-curl/desktop/src/io/ktor/client/engine/curl/internal/CurlCallbacks.kt
```diff
@@ -22,7 +24,7 @@ internal fun onHeadersReceived(
     size: size_t,
     count: size_t,
     userdata: COpaquePointer
-): Long {
+): size_t {
```
e5l, line 27: I would avoid using platform times in the API. They may have a different size on different platforms

e5l: please check other places as well

e5l (approved):
lgtm, please check minor comment

ktor-client/ktor-client-curl/desktop/src/io/ktor/client/engine/curl/internal/CurlCallbacks.kt
osipxd: Thanks for the note! In this case, that was actually the issue I fixed. This function isn’t part of the public API, but a callback passed into libcurl, so it’s expected to have different return types on different platforms.

PR #5371, opened 2026-02-17
KTOR-9340: Support ES2015 for testApplication

hfhbd:
**Subsystem**
Server, server-test-host

**Motivation**
KTOR-9340: Support ES2015 for testApplication

**Solution**
Make the failing `WORKING_DIRECTORY_PATH` overridable.

Open question: How do I test it? I tested it manually, but should I overwrite the js target of `server-test-host`? Or do you prefer an integration module for js tests, to also test https://youtrack.jetbrains.com/issue/KTOR-9341?

To manually test it, add this target setup in `ktor-server-test-host`:
```kotlin
js {
  compilerOptions {
    target = "es2015"
  }
}
```

### DIFF
--- a/ktor-server/ktor-server-core/common/src/io/ktor/server/application/Application.kt
+++ b/ktor-server/ktor-server-core/common/src/io/ktor/server/application/Application.kt
@@ -35,7 +35,7 @@ public class ServerConfigBuilder(
      *
      * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.server.application.ServerConfigBuilder.watchPaths)
      */
-    public var watchPaths: List<String> = listOf(WORKING_DIRECTORY_PATH)
+    public var watchPaths: List<String> = defaultWatchPaths()
 
     /**
      * Application's root path (prefix, context path in servlet container).
@@ -80,6 +80,8 @@ public class ServerConfigBuilder(
     internal fun build(): ServerConfig =
         ServerConfig(environment, modules.toList(), watchPaths, rootPath, developmentMode, parentCoroutineContext)
 }
+
+internal expect fun defaultWatchPaths(): List<String>
 
 /**
  * Core configuration for a running server.
--- a/ktor-server/ktor-server-core/jvm/src/io/ktor/server/application/Application.jvm.kt
+++ b/ktor-server/ktor-server-core/jvm/src/io/ktor/server/application/Application.jvm.kt
@@ -1,0 +1,5 @@
+package io.ktor.server.application
+
+import io.ktor.server.engine.WORKING_DIRECTORY_PATH
+
+internal actual fun defaultWatchPaths(): List<String> = listOf(WORKING_DIRECTORY_PATH)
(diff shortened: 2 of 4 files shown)


ktor-server/ktor-server-core/common/src/io/ktor/server/application/Application.kt
```diff
@@ -25,17 +25,19 @@ import kotlin.coroutines.EmptyCoroutineContext
...
     /**
      * Paths to wait for application reload.
      *
      * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.server.application.ServerConfigBuilder.watchPaths)
      */
-    public var watchPaths: List<String> = listOf(WORKING_DIRECTORY_PATH)
+    public var watchPaths: List<String>,
+) {
+
+    public constructor(environment: ApplicationEnvironment) : this(environment, listOf(WORKING_DIRECTORY_PATH))
```
osipxd: Let's keep one constructor here and mark another as hidden:
```suggestion
    public var watchPaths: List<String> = listOf(WORKING_DIRECTORY_PATH),
) {
 
    @Deprecated("Maintained for binary compatibility", DeprecationLevel.HIDDEN)
    public constructor(environment: ApplicationEnvironment) : this(environment, listOf(WORKING_DIRECTORY_PATH))
```

ktor-server/ktor-server-core/common/src/io/ktor/server/application/Application.kt
```diff
@@ -111,6 +113,19 @@ public fun serverConfig(
...
+ *
+ * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.server.application.serverConfig)
+ */
+public fun serverConfig(
+    environment: ApplicationEnvironment = applicationEnvironment {},
+    watchPaths: List<String>,
+    block: ServerConfigBuilder.() -> Unit = {}
+): ServerConfig {
+    return ServerConfigBuilder(environment, watchPaths).apply(block).build()
+}
```
osipxd: It might be not obvious that users should call this function with manually specified `watchPaths` if es2015 is enabled. I've noticed that `watchPaths` are used only on JVM. Let's introduce a function `defaultWatchPaths()` which would be defined only on JVM and would return `emptyList()` on non-JVM platforms

osipxd:
> Open question: How do I test it? I tested it manually, but should I overwrite the js target of server-test-host? Or do you prefer an integration module for js tests

~Changing JS target would be a breaking change.~ I've created an issue for this:
- [KTOR-9350](https://youtrack.jetbrains.com/issue/KTOR-9350) JS: Make ES2015 the default target

~We'll add proper testing in the scope of the KTOR-9350 not to stop you from merging this PR.~ Yes, you can set the target to es2015.

ktor-server/ktor-server-core/common/src/io/ktor/server/application/Application.kt
hfhbd: Done, but in that case, `watchPaths` could be redesigned as a JVM only code. Also, it should not be a String but a Path.

osipxd: Added a comment here not to miss this: https://youtrack.jetbrains.com/issue/KTOR-7957/Server-configuration-is-confusing#focus=Comments-27-13400828.0-0

PR #5372, opened 2026-02-17
KTOR-9203 Add non-blocking release function for multi-part content

bjhham:
**Subsystem**
Server, Core

**Motivation**
[KTOR-9203](https://youtrack.jetbrains.com/issue/KTOR-9203) CIOMultipartDataBase: Call thread is blocked when releasing file parts since 3.2.0

**Solution**
Added function for releasing multi-part contents with suspend instead of blocking.

### DIFF
--- a/ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/MultiPartFormDataTest.kt
+++ b/ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/MultiPartFormDataTest.kt
@@ -85,7 +85,7 @@ class MultiPartFormDataTest : ClientLoader() {
                     }
                     else -> fail("Unexpected part type: ${part::class.simpleName}")
                 }
-                part.dispose()
+                part.release()
             }
 
             assertTrue(textFound, "Text part not found")
--- a/ktor-http/api/ktor-http.klib.api
+++ b/ktor-http/api/ktor-http.klib.api
@@ -1414,23 +1414,25 @@ sealed class io.ktor.http.content/PartData { // io.ktor.http.content/PartData|nu
         final fun <get-headers>(): io.ktor.http/Headers // io.ktor.http.content/PartData.headers.<get-headers>|<get-headers>(){}[0]
     final val name // io.ktor.http.content/PartData.name|{}name[0]
         final fun <get-name>(): kotlin/String? // io.ktor.http.content/PartData.name.<get-name>|<get-name>(){}[0]
+    final val release // io.ktor.http.content/PartData.release|{}release[0]
+        final fun <get-release>(): kotlin.coroutines/SuspendFunction0<kotlin/Unit> // io.ktor.http.content/PartData.release.<get-release>|<get-release>(){}[0]
 
     final class BinaryChannelItem : io.ktor.http.content/PartData { // io.ktor.http.content/PartData.BinaryChannelItem|null[0]
-        constructor <init>(kotlin/Function0<io.ktor.utils.io/ByteReadChannel>, io.ktor.http/Headers) // io.ktor.http.content/PartData.BinaryChannelItem.<init>|<init>(kotlin.Function0<io.ktor.utils.io.ByteReadChannel>;io.ktor.http.Headers){}[0]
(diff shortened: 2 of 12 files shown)


ktor-http/api/ktor-http.klib.api
```diff
@@ -1439,7 +1441,7 @@ sealed class io.ktor.http.content/PartData { // io.ktor.http.content/PartData|nu
     }
 
     final class FormItem : io.ktor.http.content/PartData { // io.ktor.http.content/PartData.FormItem|null[0]
-        constructor <init>(kotlin/String, kotlin/Function0<kotlin/Unit>, io.ktor.http/Headers) // io.ktor.http.content/PartData.FormItem.<init>|<init>(kotlin.String;kotlin.Function0<kotlin.Unit>;io.ktor.http.Headers){}[0]
+        constructor <init>(kotlin/String, kotlin/Function0<kotlin/Unit>, io.ktor.http/Headers, kotlin.coroutines/SuspendFunction0<kotlin/Unit> = ...) // io.ktor.http.content/PartData.FormItem.<init>|<init>(kotlin.String;kotlin.Function0<kotlin.Unit>;io.ktor.http.Headers;kotlin.coroutines.SuspendFunction0<kotlin.Unit>){}[0]
```
osipxd, line 1447: I think we should preserve constructors with old signature and `DeprecationLevel.HIDDEN`

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/MultiPartFormDataTest.kt
```diff
@@ -85,7 +85,7 @@ class MultiPartFormDataTest : ClientLoader() {
                     }
                     else -> fail("Unexpected part type: ${part::class.simpleName}")
                 }
-                part.dispose()
+                part.release()
```
osipxd, line 88: Should the `releaseSuspend` be used here and in all other places where you've replaced `dispose` by `release`?

osipxd (changes requested):
Looks good
I have only one concern about binary compatibility

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/MultiPartFormDataTest.kt
bjhham: It's `dispose()` -> `suspend release()` on the parts and `release()` -> `suspend releaseSuspend()` on the more internal `MultipartEvent` class.  But now that you mention it, I _did_ miss a place or two...

ktor-http/common/src/io/ktor/http/content/Multipart.kt
```diff
@@ -18,16 +20,32 @@ import kotlinx.coroutines.flow.*
  * @property dispose to be invoked when this part is no longer needed
  * @property headers of this part, could be inaccurate on some engines
  */
-public sealed class PartData(public val dispose: () -> Unit, public val headers: Headers) {
+public sealed class PartData(
```
osipxd, line 24: I think we should add `internal` to the constructor of the `PartData`.
And preserve old signature with deprecation HIDDEN. Just in case. I doubt anybody have been creating `PartData` using sealed class constructor, but it's possible 🥲 

https://pl.kotl.in/ca_AULdzI

osipxd: Private constructor even better

bjhham: Actually, the compiler should pick up any attempts to extend a sealed type in a different module / package.  They'll get:

```
[SEALED_INHERITOR_IN_DIFFERENT_PACKAGE] A class can only extend a sealed class or interface declared in the same package.
[SEALED_INHERITOR_IN_DIFFERENT_MODULE] Extending sealed classes or interfaces from a different module is prohibited.
```

osipxd: Unfortunately those also can be suppressed. I think making the constructor private is the right thing we should do anyway, and providing hidden constructor for binary compatibility is optional.

bjhham: I really don't think we should be trying to handle cases where users are suppressing compiler errors.  The act of suppressing a compiler error is an acceptance of risk for unexpected behavior, and handling these kinds of scenarios sets a precedent that's hard to maintain.

osipxd (approved):

PR #5375, opened 2026-02-18
KTOR-9333 Fix for websocket inflate issues

bjhham:
**Subsystem**
Websockets

**Motivation**
[KTOR-9333](https://youtrack.jetbrains.com/issue/KTOR-9333) WebSockets: Infinite spin and potential OOM vulnerabilities in the Inflater.inflateFully method

**Solution**
- Added an option for the max inflated frame size with default of 256 MiB and relevant check during inflate process
- Added a progress check so that 0 inflation doesn't lead to an infinite loop

### DIFF
--- a/ktor-shared/ktor-websockets/jvm/src/io/ktor/websocket/internals/DeflaterUtils.kt
+++ b/ktor-shared/ktor-websockets/jvm/src/io/ktor/websocket/internals/DeflaterUtils.kt
@@ -45,6 +45,7 @@ internal fun Inflater.inflateFully(data: ByteArray): ByteArray =
 
 internal fun Inflater.inflateFully(data: ByteArray, maxOutputSize: Int): ByteArray {
     require(maxOutputSize >= 0) { "maxOutputSize should be >= 0" }
+    reset()
 
     val dataToInflate = data + EMPTY_CHUNK
     setInput(dataToInflate)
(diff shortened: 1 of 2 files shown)


ktor-shared/ktor-websockets/jvm/src/io/ktor/websocket/internals/DeflaterUtils.kt
```diff
@@ -39,19 +40,39 @@ internal fun Deflater.deflateFully(data: ByteArray): ByteArray {
...
             val limit = dataToInflate.size + bytesRead
             while (bytesRead < limit) {
                 buffer.clear()
+
+                val beforeBytesRead = bytesRead
                 val inflated = inflate(buffer.array(), buffer.position(), buffer.limit())
+                val afterBytesRead = bytesRead
+
+                if (inflated == 0 && afterBytesRead == beforeBytesRead) {
+                    throw IOException("Inflater made no progress: bytesRead=$afterBytesRead, limit=$limit")
```
KamilUrban: well that's not necessarily error by itself, right? it can mean we have consumed all input including the appended EMPTY_CHUNK, and there's no more output to produce. 
I'd say what should happen here is 
```
reset()
setInput()
buildPacket{
...
...
    while(true){
        if(finished()) break;
        if(inflated > 0){
            ...
            writeFully()
           if(writtenBytes > maxOutput) throw
           continue;
        }
       if(inflated <= 0){
           if(needsDictionary()) throw
           if(needsInput()) break;
           // if it reaches the end, it means no progress but not finished/needsInput, indicates corrupted stream
           throw
       }
    }
}
```

bjhham: Yep, good call 👍 I'll fix it

ktor-shared/ktor-websockets/jvm/test/WebSocketDeflateTest.kt
```diff
@@ -90,4 +92,54 @@ class WebSocketDeflateTest {
...
+            setDictionary(dict)
+            setInput(original)
+            finish()
+        }
+
+        val buf = ByteArray(4096)
+        val n = deflater.deflate(buf)
+        val dictCompressed = buf.copyOf(n)
+
+        val payload = dictCompressed + ByteArray(64) { 0x13 }
```
osipxd, line 111: `ByteArray(64) { 0x13 }`
Is it some special value?

bjhham: Nah, it's just some random trailing non-zero value to trigger the condition.

PR #5378, opened 2026-02-20
KTOR-5616 Ktor always adds by default an Accept-Charset header

osipxd:
**Subsystem**
Client

**Motivation**
[KTOR-5616](https://youtrack.jetbrains.com/issue/KTOR-5616) Ktor always adds by default an Accept-Charset header

**Solution**
Do not add this header unless user has registered charsets other than UTF-8 in the `Charsets` plugin.

It's better to review commit-by-commit.


osipxd:
@coderabbitai review

zibet27:
Nice!
Is Jetty on Java 8 CI failing because of [KTOR-9334](https://youtrack.jetbrains.com/issue/KTOR-9334) Coroutines in route handlers are dispatched with Dispatchers.Unconfined since 3.2.0?

osipxd:
> Is Jetty on Java 8 CI failing because of KTOR-9334

Probably, yes

PR #5380, opened 2026-02-24
KTOR-9381 GMTDate: zero-allocation timestamp conversion on JVM

e5l:
**Subsystem**
ktor-utils (GMTDate, DateJvm)

**Motivation**
[KTOR-9381](https://youtrack.jetbrains.com/issue/KTOR-9381) GMTDate: reduce allocations

Profiling shows `GregorianCalendar` + `Gregorian$Date` allocations from `GMTDate` construction account for ~864 samples per profiling run. Every HTTP response with a `Date` header triggers this allocation.

**Solution**
Replace `Calendar.getInstance()` / `GregorianCalendar` with a direct arithmetic algorithm (`civil_from_days`) that computes year/month/day from epoch milliseconds without allocating any objects. The algorithm handles leap years, month boundaries, and pre-1970 dates correctly.

Includes 9+ tests covering epoch, leap years, month boundaries, February edge cases, and pre-1970 dates.

### DIFF
--- a/ktor-utils/jvm/test/io/ktor/util/GMTDateCalendarTest.kt
+++ b/ktor-utils/jvm/test/io/ktor/util/GMTDateCalendarTest.kt
@@ -32,7 +32,7 @@ class GMTDateCalendarTest {
     }
 
     @Test
-    fun testEpochTimestamp() {
+    fun `epoch timestamp is 1970-01-01 00_00_00 GMT`() {
         val date = GMTDate(0L)
         assertEquals(1970, date.year)
         assertEquals(Month.JANUARY, date.month)
@@ -45,7 +45,7 @@ class GMTDateCalendarTest {
     }
 
     @Test
-    fun testLeapYear2000() {
+    fun `leap year 2000 Feb 29 is handled correctly`() {
         // Feb 29, 2000 12:00:00 GMT - leap year divisible by 400
         val timestamp = 951825600000L
         val date = GMTDate(timestamp)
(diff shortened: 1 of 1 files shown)


ktor-utils/jvm/src/io/ktor/util/date/DateJvm.kt
```diff
@@ -16,8 +27,73 @@ private val GMT_TIMEZONE = TimeZone.getTimeZone("GMT")
  * @param timestamp is a number of epoch milliseconds (it is `now` by default).
  */
 @Suppress("FunctionName")
-public actual fun GMTDate(timestamp: Long?): GMTDate =
-    Calendar.getInstance(GMT_TIMEZONE, Locale.ROOT)!!.toDate(timestamp)
+public actual fun GMTDate(timestamp: Long?): GMTDate {
```
osipxd, line 30: Would it make sense to use this as a common implementation for all platforms? It should be possible if replace `java.lang.Math` calls with Kotlin's alternatives

ktor-utils/jvm/test/io/ktor/util/GMTDateCalendarTest.kt
```diff
@@ -30,4 +30,186 @@ class GMTDateCalendarTest {
...
+        assertEquals(1, date.dayOfMonth)
+        assertEquals(0, date.hours)
+        assertEquals(0, date.minutes)
+        assertEquals(0, date.seconds)
+        assertEquals(WeekDay.THURSDAY, date.dayOfWeek)
+        assertEquals(1, date.dayOfYear)
+    }
+
+    @Test
+    fun `leap year 2000 Feb 29 is handled correctly`() {
```
osipxd, line 48: Probably all these tests could be replaced with one list of `Pair<Long, String>` where `Long` is the timestamp and `String` is the expected output of `toHttpDate`

osipxd (approved):
🔥 
I have only two questions. However it's okay to merge without considering them

PR #5383, opened 2026-02-24
KTOR-9384 Add bodyAsByteArray() to HttpResponse for efficient cached access

e5l:
**Subsystem**
ktor-client-core (HttpResponse, SavedHttpCall, DefaultTransform)

**Motivation**
[KTOR-9384](https://youtrack.jetbrains.com/issue/KTOR-9384) Optimize body reading for saved responses

When a response is saved (cached), reading the body still goes through `readRemaining().readByteArray()` which copies the already-cached bytes. This is wasteful for the common case where the response body is already available as a byte array.

**Solution**
- Add an `open` `bodyAsByteArray()` method to `HttpResponse` (default: `null`, meaning no cached byte array available)
- `SavedHttpCall.SavedResponse` overrides it to return the cached bytes directly
- `DefaultTransform` uses the new hook to avoid the double-copy

**API Change**: Adds `public open fun bodyAsByteArray(): ByteArray?` to `HttpResponse`. API dumps updated.

### DIFF
--- a/ktor-client/ktor-client-core/common/test/SavedCallTest.kt
+++ b/ktor-client/ktor-client-core/common/test/SavedCallTest.kt
@@ -54,6 +54,27 @@ class SavedCallTest {
     }
 
     @Test
+    fun `mutating returned byte array does not affect subsequent reads`() = runTest {
+        val expectedBody = "immutable body"
+        val client = HttpClient(MockEngine) {
+            engine {
+                addHandler { respond(expectedBody, HttpStatusCode.OK) }
+            }
+        }
+
+        val savedCall = client.get("http://localhost/test").call.save()
+
+        val firstRead = savedCall.body<ByteArray>()
+        // Mutate every byte in the returned array
+        for (i in firstRead.indices) {
+            firstRead[i] = 0
+        }
+
+        val secondRead = savedCall.body<ByteArray>()
+        assertContentEquals(expectedBody.encodeToByteArray(), secondRead)
+    }
+
+    @Test
     fun `saved call allows double receive`() = runTest {
         val expectedBody = "test body"
         val client = HttpClient(MockEngine) {


osipxd (changes requested):
After this PR, `SavedHttpResponse.bodyAsByteArray` returns the internal body array directly, so callers of the public `body<ByteArray>()` API now hold a mutable reference to the cached data. Mutating the returned array corrupts subsequent reads.
Do you think we can avoid this?

Please rebase onto the latest main. It should contain fixes for some tests that are failing here.

ktor-client/ktor-client-core/common/src/io/ktor/client/call/SavedCall.kt
```diff
@@ -82,7 +82,7 @@ internal class SavedHttpResponse(
     @OptIn(InternalAPI::class)
     override suspend fun bodyAsByteArray(channel: ByteReadChannel): ByteArray {
         channel.cancel()
-        return body
+        return body.copyOf()
```
osipxd, line 85: Could you also add the test verifying that mutation of response body doesn't affect further reads?

e5l:
Rebased onto latest main (630e201bb). The rebase was clean with no conflicts.

Regarding the allocation regression in `gzipResponseFootprint`:

The regression comes from `body.copyOf()` in `SavedHttpResponse.bodyAsByteArray` adding an extra ~2MB allocation. Before this PR, the ByteArray path went through `rawContent` → `ByteReadChannel(body)` → `toByteArray()`, which consumed the channel buffer and produced a fresh copy. Now the channel is still created (from `rawContent` in `getResponseContent()`), but `bodyAsByteArray` cancels it (wasting the ~2MB buffer allocation) and additionally allocates `body.copyOf()`.

There are a few options to resolve this:
1. **Return `body` directly** (no `copyOf()`) — fixes allocation regression and provides a genuine optimization (one fewer copy), but callers who mutate the returned ByteArray would corrupt subsequent reads
2. **Keep `body.copyOf()`** but also override `getResponseContent()` in `SavedHttpCall` to avoid the wasted channel creation — this requires deeper changes to how the pipeline receives data for saved responses
3. **Remove the `bodyAsByteArray` override** from `SavedHttpResponse` entirely — no regression, no optimization, but the hook remains available for other HttpResponse subclasses

Which approach would you prefer?

e5l:
Closing this PR — the allocation regression from body.copyOf() needs a deeper architectural rethink. Will revisit with a different approach.

PR #5387, opened 2026-02-25
KTOR-9352 Handle EC key type when JWK algorithm is null

e5l:
## Summary
- Fixes https://youtrack.jetbrains.com/issue/KTOR-9352
- `Jwk.makeAlgorithm()` assumed RSA when `algorithm` is `null`, causing `ClassCastException` for EC keys. Now checks `type` to select the correct algorithm family.

## Test plan
- Added reproducer test `KTOR-9352 makeAlgorithm with EC key type and null algorithm` that creates an EC JWK mock with `kty=EC` and `alg=null`, verifying it produces a valid ECDSA verifier
- Updated existing `verifyNullAlgorithmWithMock` to explicitly set `type = "RSA"` on its mock
- All existing tests in the module continue to pass

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-auth-jwt/jvm/test/io/ktor/server/auth/jwt/JWTAuthTest.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-auth-jwt/jvm/test/io/ktor/server/auth/jwt/JWTAuthTest.kt
@@ -402,6 +402,54 @@ class JWTAuthTest {
     }
 
     @Test
+    fun `KTOR-9352 makeAlgorithm with EC P-384 key and null algorithm`() {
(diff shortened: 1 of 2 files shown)


ktor-server/ktor-server-plugins/ktor-server-auth-jwt/jvm/test/io/ktor/server/auth/jwt/JWTAuthTest.kt
```diff
@@ -374,6 +374,81 @@ class JWTAuthTest {
         verifier.verify(token)
     }
 
+    @Test
+    fun `KTOR-9352 makeAlgorithm with EC key type and null algorithm`() {
```
osipxd, line 378: Nit. Has the issue number been included by intention here?

PR #5390, opened 2026-02-25
KTOR-1915 Add appendBlob for multipart form data from Kotlin/JS

e5l:
## Summary
- Adds JS-specific `FormBuilder.appendBlob()` extension functions that accept `org.w3c.files.Blob` (and `File`, which extends `Blob`)
- Enables browser file uploads via the Ktor client multipart form data API from Kotlin/JS
- Uses `ChannelProvider` backed by async Blob-to-ArrayBuffer conversion

Closes KTOR-1915

## Test plan
- Added 3 failing reproducer tests validating Blob content, filename/contentType headers, and Content-Length preservation
- All existing tests in ktor-client-core continue to pass across all platforms (`allTests`)

### DIFF
--- a/ktor-client/ktor-client-core/js/src/io/ktor/client/request/forms/FormDsl.js.kt
+++ b/ktor-client/ktor-client-core/js/src/io/ktor/client/request/forms/FormDsl.js.kt
@@ -2,12 +2,11 @@
  * Copyright 2014-2026 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
  */
 
-// ABOUTME: JS-specific FormBuilder extensions for appending browser Blob/File objects
-// ABOUTME: to multipart form data requests.
+// ABOUTME: JS-specific implementation of Blob-to-ChannelProvider conversion
+// ABOUTME: using dynamic JS interop for arrayBuffer() access.
 
 package io.ktor.client.request.forms
 
-import io.ktor.http.*
 import io.ktor.utils.io.*
 import kotlinx.coroutines.*
 import org.khronos.webgl.ArrayBuffer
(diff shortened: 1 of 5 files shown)


ktor-client/ktor-client-core/js/test/io/ktor/client/request/forms/FormDslJsTest.kt
```diff
@@ -0,0 +1,82 @@
...
+import kotlinx.coroutines.test.runTest
+import kotlinx.io.readByteArray
+import org.w3c.files.Blob
+import org.w3c.files.BlobPropertyBag
+import kotlin.test.*
+
+class FormDslJsTest {
+
+    @Test
+    fun `KTOR-1915 appendBlob creates form part from Blob`() = runTest {
```
osipxd: It seems we can remove KTOR-1915 from the name of the tests

ktor-client/ktor-client-core/js/src/io/ktor/client/request/forms/FormDsl.js.kt
```diff
@@ -0,0 +1,77 @@
...
+ * ```kotlin
+ * val file: File = // from <input type="file">
+ * client.submitFormWithBinaryData(url, formData {
+ *     appendBlob("document", file)
+ * })
+ * ```
+ *
+ * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.client.request.forms.appendBlob)
+ */
+public fun FormBuilder.appendBlob(key: String, blob: Blob, headers: Headers = Headers.Empty) {
```
osipxd: Was it JS-only problem? Should we add something for Wasm/JS as well?

ktor-client/ktor-client-core/js/test/io/ktor/client/request/forms/FormDslJsTest.kt
e5l: Done, removed the KTOR-1915 prefix from all test names.

ktor-client/ktor-client-core/js/src/io/ktor/client/request/forms/FormDsl.js.kt
e5l: Good question\! The wasmJs target has its own `Blob` type defined in `LibDom.kt` (`io.ktor.client.fetch.Blob` extending `JsAny`), which is different from `org.w3c.files.Blob` used by the JS target. So we can't simply move this to the `web` shared source set — wasmJs would need a separate implementation with its own Blob type. Should I add wasmJs support in this PR or leave it for a follow-up?

osipxd: If it doesn't require writing a lot of code it'd be better to introduce Wasm/JS functions here as well

osipxd (changes requested):
Would it be possible to move `Blob` into the `web` source set to make it possible to use this API shared code?

e5l:
Moved `appendBlob` to the `web` source set (shared between JS and WasmJS). Both platforms now use `org.w3c.files.Blob` from `kotlinx-browser`, which is already a dependency of the `web` source set.

The public `appendBlob` functions live in `web/src`, while the platform-specific blob-to-channel conversion (`blobChannelProvider`) uses `expect`/`actual` — the expect declaration is in `web/src` and the actual implementations remain in `js/src` and `wasmJs/src` since the byte extraction logic differs between platforms (JS uses `asDynamic().arrayBuffer()` + `Int8Array.unsafeCast`, WasmJS uses `js()` interop functions).

The WasmJS API signature changed from `io.ktor.client.fetch.Blob` to `org.w3c.files.Blob`, which is fine since this is a new API. The ABI dump now shows a single `[js, wasmJs]` entry instead of separate per-platform entries.

PR #5391, opened 2026-02-25
KTOR-8751 Fix AmbiguousDependencyException during DI shutdown

e5l:
## Summary
- Fixes [KTOR-8751](https://youtrack.jetbrains.com/issue/KTOR-8751)
- When multiple concrete types implement the same interface and are registered as DI dependencies, their covariant keys create `Ambiguous` entries for the unnamed interface key. During application shutdown, the DI plugin iterates all dependency map entries and calls `getDeferred()` on each — including these `Ambiguous` entries — which throws `AmbiguousDependencyException`.
- The fix skips `Ambiguous` and `Missing` entries during the shutdown cleanup loop, since these represent unresolvable covariant artifacts rather than actual resolved dependencies that need cleanup.

Closes KTOR-8751

## Test plan
- Added failing reproducer test that registers two concrete types implementing the same interface with a named interface binding, then verifies no `AmbiguousDependencyException` warnings occur during shutdown
- Added test verifying named dependency override in `testApplication` works correctly
- All existing tests in the module continue to pass across all platforms

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-di/common/test/io/ktor/server/plugins/di/DependencyInjectionTest.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-di/common/test/io/ktor/server/plugins/di/DependencyInjectionTest.kt
@@ -596,7 +596,7 @@ class DependencyInjectionTest {
     }
 
     @Test
-    fun `KTOR-8751 named dependency override in testApplication`() = testApplication {
+    fun `named dependency override in testApplication`() = testApplication {
         application {
             // Override the named dependency BEFORE registering the original
             dependencies.key<GreetingService>("s1") {
@@ -624,7 +624,7 @@ class DependencyInjectionTest {
     }
 
     @Test
-    fun `KTOR-8751 shutdown does not throw for ambiguous covariant entries`() = runTest {
+    fun `shutdown does not throw for ambiguous covariant entries`() = runTest {
         val errorLogs = mutableListOf<String>()
         val warnLogs = mutableListOf<String>()
         // Use DefaultConflictPolicy (as in production) to reproduce the shutdown issue


ktor-server/ktor-server-plugins/ktor-server-di/common/test/io/ktor/server/plugins/di/DependencyInjectionTest.kt
```diff
@@ -595,6 +595,86 @@ class DependencyInjectionTest {
         assertEquals(HELLO_CUSTOMER, service.hello())
     }
 
+    @Test
+    fun `KTOR-8751 named dependency override in testApplication`() = testApplication {
```
bjhham: Let's drop the YT ticket references from the function names for the sake of consistent styling.

e5l: Done.

e5l:
Addressed review feedback:
- Removed `KTOR-8751` prefix from test function names as requested by @bjhham
- Skipped: CodeRabbit bot suggestions (informational only, no real issues)
- CI failures are all in unrelated modules (server engines, client HTTP timeout/WebSocket, curl, rate limiting) — none in the DI module

e5l:
CI status update: All 9 required checks are now passing.

- Restarted flaky CI builds (Java 11, Java 21, Native macOS X64, Native Linux X64, Native Windows X64) — all were failing with unrelated infrastructure tests (TLS handshake, H2 timeouts, WebSocket, HTTP timeout tests in server engine and client modules)
- Java 11 on Linux: ✅ passed on retry
- Java 21 on Linux: ✅ passed on retry  
- Native macOS X64: ✅ passed on retry (2051 tests, 0 failures)
- Native Windows X64: ✅ passed on retry (2025 tests, 0 failures)
- Native Linux X64: still has 2 flaky WebSocket test failures (not a required check)
- Build All Core: aggregator still references old snapshot deps (not a required check)

Auto-merge is enabled and should proceed once GitHub reconciles the check statuses.

PR #5402, opened 2026-02-27
KTOR-4237 Stop returning 403 from CORS plugin when checks fail

e5l:
## Summary
- Fixes [KTOR-4237](https://youtrack.jetbrains.com/issue/KTOR-4237)
- The CORS specification does not define HTTP 403 as a valid response for failed CORS checks. Servers should simply omit CORS headers and let the browser enforce the policy. This PR removes all 403 responses from the CORS plugin:
  - Failed preflight (OPTIONS) requests now respond with 200 OK without CORS headers, so the browser blocks the subsequent actual request
  - Failed actual requests proceed to the route handler without CORS headers, so the browser blocks JavaScript access to the response
  - Removes the now-unused `respondCorsFailed()` utility function

KTOR-4237

## Test plan
- Added 4 new reproducer tests that validate the fix (committed separately as a failing test first)
- Updated all existing CORS tests to expect the new spec-compliant behavior
- All 60 CORS tests pass

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-cors/common/src/io/ktor/server/plugins/cors/CORS.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-cors/common/src/io/ktor/server/plugins/cors/CORS.kt
@@ -117,7 +117,9 @@ internal fun PluginBuilder<CORSConfig>.buildPlugin() {
             OriginCheckResult.SkipCORS -> return@onCall
             OriginCheckResult.Failed -> {
                 LOGGER.trace { "${call.request.id()}: CORS check fails because Origin $origin does not match" }
-                call.respondCorsFailed()
+                if (call.request.httpMethod == HttpMethod.Options) {
+                    call.respond(HttpStatusCode.OK)
+                }
                 return@onCall
             }
         }
@@ -130,7 +132,9 @@ internal fun PluginBuilder<CORSConfig>.buildPlugin() {
                         "${call.request.id()}: CORS check fails because the requested content type $contentType " +
                             "is not in the list of the only allowed simple types ${CORSConfig.CorsSimpleContentTypes}"
                     }
-                    call.respondCorsFailed()
+                    if (call.request.httpMethod == HttpMethod.Options) {
+                        call.respond(HttpStatusCode.OK)
+                    }
                     return@onCall
                 }
             }
@@ -157,7 +161,6 @@ internal fun PluginBuilder<CORSConfig>.buildPlugin() {
                 "${call.request.id()}: CORS check fails because HTTP method ${call.request.httpMethod.value} " +
                     "is not allowed. Allowed methods: $methods"
             }
-            call.respondCorsFailed()
             return@onCall
         }
 
@@ -234,7 +237,7 @@ private suspend fun ApplicationCall.respondPreflight(
 
     if (!corsCheckRequestMethod(methods)) {
         LOGGER.trace { "${request.id()}: Preflight: Request method check fails" }
-        respond(HttpStatusCode.Forbidden)
+        respond(HttpStatusCode.OK)
         return
     }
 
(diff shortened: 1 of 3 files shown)


e5l:
Reviewed PR status — no actionable items found:

- **Review comments**: Only CodeRabbit bot nitpick (use `assertEquals(OK)` instead of `assertNotEquals(Forbidden)` in tests). Skipped as informational.
- **CI failures**: All 3 failing checks (Native Windows X64, Native Linux X64, Build All Core) contain only pre-existing/flaky test failures in unrelated modules (`ktor-client` curl engine, network sockets). None are related to the CORS plugin changes in this PR.

No code changes needed.

e5l:
Investigated all 3 failing CI checks. All failures are pre-existing and unrelated to the CORS changes:

**Native Windows X64** (build 407111) — 2 failures:
- `CurlHttp2Test.test protocol version is HTTP 2[mingwX64]` — failing since build #3936. Curl engine returns HTTP/1.1 instead of HTTP/2.0 on Windows CI.
- `TCPSocketTest.testAwaitClosedDoesNotDeadLock[mingwX64]` — intermittent race condition: `Failed to connect to InetSocketAddress(127.0.0.1:port)`.

**Native Linux X64** (build 407108) — 2 failures:
- `CurlWebSocketTests.testWebSocketHeaders[linuxX64]` — timeout after 10s.
- `CurlWebSocketTests.testParallelSessions[linuxX64]` — timeout after 10s.

**Build All Core** (build 407116) — snapshot dependency failure aggregating the above.

**Evidence:** Both `Native Linux X64` and `Native Windows X64` are failing across all recent PRs (checked pull/5449, pull/5426, pull/5407, pull/5391, pull/5356, pull/5417, pull/5381, and release/3.x). The `CurlHttp2Test` failure dates back to build #3936. `ContentTest.testEmptyContent` has been failing since build #3972.

This PR only modifies files in `ktor-server-plugins/ktor-server-cors/` — none of the failing tests are in that module.

ktor-server/ktor-server-tests/common/test/io/ktor/tests/server/plugins/CORSTest.kt
```diff
@@ -1589,7 +1617,95 @@ class CORSTest {
...
             header(HttpHeaders.AccessControlRequestMethod, "GET")
         }.let { response ->
-            assertEquals(response.status, HttpStatusCode.Forbidden)
+            assertEquals(response.status, HttpStatusCode.OK)
+            assertNull(response.headers[HttpHeaders.AccessControlAllowOrigin])
+        }
+    }
+
+    @Test
+    fun `KTOR-4237 preflight with disallowed method should not respond with 403`() = testApplication {
```
bjhham, line 1626: Looks like Claude put the ticket numbers in the tests, best to omit.

ktor-server/ktor-server-plugins/ktor-server-cors/common/src/io/ktor/server/plugins/cors/CORS.kt
```diff
@@ -117,7 +117,9 @@ internal fun PluginBuilder<CORSConfig>.buildPlugin() {
             OriginCheckResult.SkipCORS -> return@onCall
             OriginCheckResult.Failed -> {
                 LOGGER.trace { "${call.request.id()}: CORS check fails because Origin $origin does not match" }
-                call.respondCorsFailed()
+                if (call.request.httpMethod == HttpMethod.Options) {
+                    call.respond(HttpStatusCode.OK)
+                }
                 return@onCall
```
bjhham, line 123: I don't think this is the right behavior here.  From what I understand, the server should be providing the correct headers to allow browsers to enforce CORS, and the server shouldn't enforce checks at all.  Responding with an empty 200 would just be confusing.

bjhham (changes requested):

PR #5404, opened 2026-02-27
docs: HttpRequestRetry and HttpTImeout installation order

matteomazza91:
**Subsystem**
Client and related modules

**Motivation**
Documentation Improvement

**Solution**
Documentation Improvement, similar to already documented behaviour here: https://github.com/ktorio/ktor/blob/b5f010b5b604bdef6473893e90f259de91cbe46d/ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/HttpRequestRetry.kt#L134


osipxd (approved):
Thank you!

PR #5406, opened 2026-03-01
KTOR-9372 Fix infinite loop in CharsetDecoder.decode on Native with truncated UTF-8

e5l:
## Summary
- Fixes [KTOR-9372](https://youtrack.jetbrains.com/issue/KTOR-9372)
- `CharsetDecoder.decode()` on Kotlin/Native (Linux/MinGW) enters an infinite loop when decoding truncated or malformed UTF-8 byte sequences. This happens because `iconv` returns EINVAL (incomplete multi-byte sequence) consuming zero bytes, but the decode loop continues without making progress.
- The fix detects zero-progress iterations and throws `MalformedInputException`, consistent with how illegal sequences (EILSEQ) are already handled.

KTOR-9372

## Test plan
- Added reproducer tests in `CharsetDecoderTest` covering truncated 2-byte, 3-byte, 4-byte UTF-8 sequences, and valid text followed by a truncated sequence
- All existing tests in ktor-io continue to pass on JVM
- The fix applies to both `CharsetLinux.kt` and `CharsetMingw.kt` (identical iconv-based implementations)

### DIFF
--- a/ktor-io/common/test/CharsetDecoderTest.kt
+++ b/ktor-io/common/test/CharsetDecoderTest.kt
@@ -13,7 +13,7 @@ import kotlin.test.*
 class CharsetDecoderTest {
 
     @Test
-    fun `KTOR-9372 decoding truncated UTF-8 should not cause infinite loop`() {
+    fun `decoding truncated UTF-8 should not cause infinite loop`() {
         // Incomplete 2-byte UTF-8 sequence: 0xC2 expects a continuation byte (0x80-0xBF)
         val truncated = byteArrayOf(0xC2.toByte())
         val source = buildPacket { writeFully(truncated) }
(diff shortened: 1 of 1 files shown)


ktor-io/common/test/CharsetDecoderTest.kt
```diff
@@ -0,0 +1,77 @@
...
+            0xC2.toByte() // incomplete 2-byte sequence
+        )
+        val source = buildPacket { writeFully(data) }
+
+        try {
+            val result = Charsets.UTF_8.newDecoder().decode(source)
+            // On JVM, the valid prefix should be decoded
+            assertTrue(result.startsWith("Hello"))
+        } catch (_: MalformedInputException) {
+            // Expected on native platforms
```
osipxd, line 74: Is it possible to make it consistent for all platforms?

ktor-io/common/test/CharsetDecoderTest.kt
```diff
@@ -0,0 +1,77 @@
...
+
+import io.ktor.utils.io.charsets.*
+import io.ktor.utils.io.core.*
+import kotlinx.io.*
+import kotlin.test.*
+
+class CharsetDecoderTest {
+
+    @Test
+    fun `KTOR-9372 decoding truncated UTF-8 should not cause infinite loop`() {
```
osipxd: Nit. Claude has added task number to all tests again :(

e5l: we need to adjust the problem

osipxd (changes requested):
Please remove the issue number from test names

ktor-io/common/test/CharsetDecoderTest.kt
e5l: discussed

PR #5408, opened 2026-03-02
KTOR-9344 Disable switching to engine dispatcher by default

osipxd:
**Subsystem**
Client Core

**Motivation**
[KTOR-9344](https://youtrack.jetbrains.com/issue/KTOR-9344) Flow invariant error happens after update to Ktor 3.4.0

**Solution**
Disable dispatcher switching by default on JVM, introduce a system property `io.ktor.client.statement.useEngineDispatcher` to make it possible to enable it.

### DIFF
--- a/ktor-client/ktor-client-core/api/ktor-client-core.api
+++ b/ktor-client/ktor-client-core/api/ktor-client-core.api
@@ -1552,6 +1552,10 @@ public final class io/ktor/client/statement/DefaultHttpResponse : io/ktor/client
 	public fun getVersion ()Lio/ktor/http/HttpProtocolVersion;
 }
 
+public final class io/ktor/client/statement/DispatcherSwitching_jvmKt {
+	public static final fun getUseEngineDispatcher ()Z
+}
+
 public final class io/ktor/client/statement/HttpReceivePipeline : io/ktor/util/pipeline/Pipeline {
 	public static final field Phases Lio/ktor/client/statement/HttpReceivePipeline$Phases;
 	public fun <init> ()V
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/statement/HttpStatement.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/statement/HttpStatement.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2025 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ * Copyright 2014-2026 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
  */
 
 package io.ktor.client.statement
@@ -47,8 +47,16 @@ public class HttpStatement(
      * The response object should not be accessed outside of [block] as it will be canceled upon
      * block completion.
      *
-     * The [block] is executed on the engine's dispatcher, making it safe to perform IO operations
-     * such as reading the response content and writing it into a file.
+     * ## Dispatcher Behavior
+     * On non-JVM platforms (Web, Native), the [block] is executed on the engine's dispatcher,
+     * making it safe to perform IO operations such as reading the response content and writing it into a file.
+     *
+     * On JVM, the [block] runs on the caller's dispatcher by default for backward compatibility.
+     * To enable engine dispatcher switching on JVM, set the system property:
+     * `-Dio.ktor.client.statement.useEngineDispatcher=true`
+     *
+     * **Note:** Starting from Ktor 4.0, dispatcher switching will be enabled by default on all platforms.
+     * It is recommended to opt-in early to ensure compatibility with the upcoming release.
      *
      * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.client.statement.HttpStatement.execute)
      *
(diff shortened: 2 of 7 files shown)


ktor-client/ktor-client-core/nonJvm/src/io/ktor/client/statement/DispatcherSwitching.nonJvm.kt
```diff
@@ -0,0 +1,8 @@
+/*
+ * Copyright 2014-2026 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.client.statement
+
+@PublishedApi
+internal actual val useEngineDispatcher: Boolean = true
```
osipxd, line 8: Kept switching enabled on native and web targets. Native engines used the `Unconfined` dispatcher prior 3.4.0, so switching here shouldn't be a problem. However it might be a problem for code shared between JVM and web/native 🤔

ktor-client/ktor-client-core/api/ktor-client-core.api
```diff
@@ -1552,6 +1552,10 @@ public final class io/ktor/client/statement/DefaultHttpResponse : io/ktor/client
 	public fun getVersion ()Lio/ktor/http/HttpProtocolVersion;
 }
 
+public final class io/ktor/client/statement/DispatcherSwitching_jvmKt {
+	public static final fun getUseEngineDispatcher ()Z
+}
```
osipxd, line 1557: It is internal, marked with `@PublishedApi`, so should be ok to include in a patch release

ktor-client/ktor-client-core/jvm/src/io/ktor/client/statement/DispatcherSwitching.jvm.kt
```diff
@@ -0,0 +1,12 @@
+/*
+ * Copyright 2014-2026 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.client.statement
+
+private const val USE_ENGINE_DISPATCHER_KEY: String = "io.ktor.client.statement.useEngineDispatcher"
+
+@PublishedApi
+internal actual val useEngineDispatcher: Boolean by lazy(LazyThreadSafetyMode.NONE) {
```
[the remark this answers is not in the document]
osipxd: Is it a real problem? `System.getProperty` would return the same result for any thread, so it should be okay for us to initialize it multiple times. I want to avoid synchronization overhead.

ktor-client/ktor-client-core/jvm/src/io/ktor/client/statement/DispatcherSwitching.jvm.kt
```diff
@@ -0,0 +1,12 @@
+/*
+ * Copyright 2014-2026 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.client.statement
+
+private const val USE_ENGINE_DISPATCHER_KEY: String = "io.ktor.client.statement.useEngineDispatcher"
```
osipxd, line 7: Alternatively thought about introducing some general flag like [progressive mode](https://kotlinlang.org/api/kotlin-gradle-plugin/kotlin-gradle-plugin-api/org.jetbrains.kotlin.gradle.dsl/-kotlin-common-compiler-options/progressive-mode.html) in Kotlin. It would allow us to introduce such behavior changes and enable them by default on new projects.
However, `useEngineDispatcher` is the only candidate for this progressive mode I'm aware

ktor-client/ktor-client-core/jvm/src/io/ktor/client/statement/DispatcherSwitching.jvm.kt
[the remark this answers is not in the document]
osipxd: @codex Is it a real problem? `System.getProperty` would return the same result for any thread, so it should be okay for us to initialize it multiple times. I want to avoid synchronization overhead.

Stexxe (approved):

ktor-client/ktor-client-core/common/src/io/ktor/client/statement/HttpStatement.kt
```diff
@@ -114,8 +126,16 @@ public class HttpStatement(
...
+     * ## Dispatcher Behavior
+     * On non-JVM platforms (Web, Native), the [block] is executed on the engine's dispatcher,
+     * making it safe to perform IO operations such as writing to a file.
+     *
+     * On JVM, the [block] runs on the caller's dispatcher by default for backward compatibility.
+     * To enable engine dispatcher switching on JVM, set the system property:
+     * `-Dio.ktor.client.statement.useEngineDispatcher=true`
+     *
+     * **Note:** Starting from Ktor 4.0, dispatcher switching will be enabled by default on all platforms.
+     * It is recommended to opt-in early to ensure compatibility with the upcoming release.
```
osipxd, line 138: Could you verify this? @e5l

e5l: lgtm

PR #5411, opened 2026-03-02
KTOR-9355 Deprecate HttpHeaders.AcceptCharset

osipxd:
**Subsystem**
`ktor-http`

**Motivation**
[KTOR-9355](https://youtrack.jetbrains.com/issue/KTOR-9355) Deprecate HttpHeaders.AcceptCharset

### DIFF
--- a/ktor-http/common/src/io/ktor/http/HttpHeaders.kt
+++ b/ktor-http/common/src/io/ktor/http/HttpHeaders.kt
@@ -1,8 +1,10 @@
 /*
-* Copyright 2014-2021 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
-*/
+ * Copyright 2014-2026 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
 
 package io.ktor.http
+
+import io.ktor.http.HttpHeaders.UnsafeHeadersList
 
 @Suppress(
     "unused",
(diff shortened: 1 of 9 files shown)


ktor-server/ktor-server-core/common/src/io/ktor/server/request/ApplicationRequestProperties.kt
```diff
@@ -125,13 +125,15 @@ public fun ApplicationRequest.acceptLanguageItems(): List<HeaderValue> =
...
+@Suppress("DEPRECATION")
 public fun ApplicationRequest.acceptCharset(): String? = header(HttpHeaders.AcceptCharset)
 
 /**
  * Gets the `Accept-Charset` header charsets sorted according to their qualities.
  *
  * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.server.request.acceptCharsetItems)
  */
+@Suppress("DEPRECATION")
 public fun ApplicationRequest.acceptCharsetItems(): List<HeaderValue> =
```
osipxd, line 137: Should we also deprecate these functions? My thought was that we should discourage users from adding this header, but it's okay to try to read it, because server is usually out of control.

bjhham: I think reading it in is fine, maybe exclude this one.

bjhham (commented):
Could you include some KDoc or a deprecation message that either references or summarizes the reason in the RFC note?

ktor-http/common/src/io/ktor/http/HttpHeaders.kt
```diff
@@ -15,6 +17,11 @@ public object HttpHeaders {
     // The list is taken from https://www.iana.org/assignments/message-headers/message-headers.xml#perm-headers
 
     public const val Accept: String = "Accept"
+
+    @Deprecated(
+        "The Accept-Charset request header has been deprecated in RFC 9110. " +
+            "UTF-8 should be used by default in modern applications."
+    )
     public const val AcceptCharset: String = "Accept-Charset"
```
osipxd, line 25: @bjhham, could you check if this deprecation message is enough?

bjhham: lgtm 👍

bjhham (approved):

Issue #5412, opened 2026-03-02
ZstdEncoder decode fails when source data is split into multiple zstd frames


glebArkhipov:
`ZstdEncoder().decode` throws an exception when the source data is large enough to be split into multiple chunks by the encoder:

```
Caused by: io.ktor.utils.io.ClosedByteChannelException: Destination buffer is too small
    at io.ktor.utils.io.CloseToken$wrapCause$1.invoke(CloseToken.kt:16)
    at io.ktor.utils.io.CloseToken$wrapCause$1.invoke(CloseToken.kt:16)
    at io.ktor.utils.io.CloseToken.wrapCause(CloseToken.kt:21)
    at io.ktor.utils.io.CloseToken.wrapCause$default(CloseToken.kt:16)
    at io.ktor.utils.io.ByteChannel.cancel(ByteChannel.kt:141)
    at io.ktor.utils.io.ByteWriteChannelOperationsKt$writer$job$1.invokeSuspend(ByteWriteChannelOperations.kt:193)
    at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:34)
    at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:100)
```

You can reproduce it with the following test:

```kotlin
@Test
fun testEncodeDecodeChunkedContent() = runTest {
        val string = "zstd".repeat(4098)
        val encodedReadChannel = ZstdEncoder().encode(ByteReadChannel(string.toByteArray()))
        val decodedReadChannel = ZstdEncoder().decode(encodedReadChannel)
        val decodedString = decodedReadChannel.readRemaining().readText()
        assertEquals(string, decodedString)
}
```

Also it could be reproduced in a client–server setup.
When the response body is large enough to be split into multiple zstd frames by the server, the client crashes with the same exception.

A server is configured to return zstd-compressed response body:
```kotlin
install(Compression) {
    zstd()
}
```

A client is configured to decode zstd-compressed data.
```kotlin
val client = HttpClient(CIO) {
    install(ContentEncoding) {
        customEncoder(ZstdEncoder())
    }
}
```

### Root cause

The encoder splits data in multiple chunks, compresses them, producing multiple zstd frames concatenated in the output channel.
The decoder reads several of those frames into a single buffer at once and fails to decompress them correctly because the output buffer is too small. 
The buffer size is for one frame only.

[...]

PR #5413, opened 2026-03-02
KTOR-9387: fix ZstdEncoder.decode when source data is split into multiple zstd frames

glebArkhipov:
**Subsystem**
Server, Compression

**Motivation**
The issue is described here https://github.com/ktorio/ktor/issues/5412

**Solution**
`decodeTo` now iterates through the input buffer frame by frame, using `findFrameCompressedSize` to find frame size and decompress only a single frame at a time


glebArkhipov:
Found the strange branch name
Create new PR https://github.com/ktorio/ktor/pull/5414

PR #5421, opened 2026-03-04
Server Netty. Set proper `call.coroutineContext` to fix HttpRequestLifecycle

zibet27:
**Subsystem**
Netty Server

**Motivation**
[KTOR-9343](https://youtrack.jetbrains.com/issue/KTOR-9343) HttpRequestLifecycle plugin with `cancelCallOnClose` on, cancels subsequent requests when CallLogging plugin with callIdMdc is installed
The problem is that cancellation of `call.coroutineContext` was cancelling `userAppContext` (application.coroutineContext + userContext). They were identical.

**Solution**
- Refactored Netty HTTP/1 and HTTP/2 handlers to inline call execution instead of using a separate NettyApplicationCallHandler channel handler
- Use a separate job inside of `call.coroutineContext`

### DIFF
--- a/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/http1/NettyHttp1Handler.kt
+++ b/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/http1/NettyHttp1Handler.kt
@@ -147,6 +147,10 @@ internal class NettyHttp1Handler(
         // Reserve response slot synchronously on the I/O thread for proper ordering
         responseWriter.processResponse(call)
 
+        // Defer coroutine start to the next event loop tick so that channelReadComplete() fires first
+        // This allows the response pipeline to detect that the request body is still being received and flush headers
+        // early instead of buffering them, which is required when the client waits for response headers
+        // before sending the request body
         context.executor().execute {
             val callScope = CoroutineScope(context = callContext)
             callScope.launch(start = CoroutineStart.UNDISPATCHED) {
--- a/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/http2/NettyHttp2Handler.kt
+++ b/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/http2/NettyHttp2Handler.kt
@@ -116,6 +116,9 @@ internal class NettyHttp2Handler(
         // Reserve response slot synchronously on the I/O thread for proper ordering
         responseWriter.processResponse(call)
 
+        // Defer coroutine start to the next event loop tick via context.executor().execute so that
+        // channelRead returns and Netty can deliver subsequent Http2DataFrame messages.
+        // Without this, the coroutine runs on the event loop, blocking data frame delivery and causing EOFException.
         context.executor().execute {
             val callScope = CoroutineScope(context = callContext)
             callScope.launch(start = CoroutineStart.UNDISPATCHED) {


ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/http2/NettyHttp2Handler.kt
```diff
@@ -92,17 +105,31 @@ internal class NettyHttp2Handler(
...
 
-        context.fireChannelRead(call)
+        // Reserve response slot synchronously on the I / O thread for proper ordering
+        // call.coroutineContext should be updated with the proper value later
         responseWriter.processResponse(call)
+
+        callEventGroup.execute {
+            val userScope = CoroutineScope(userCoroutineContext)
+            val callContext = CallHandlerCoroutineName + NettyDispatcher.CurrentContext(context)
+            userScope.launch(callContext, start = CoroutineStart.UNDISPATCHED) callJob@{
```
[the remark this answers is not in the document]
zibet27: `CoroutineScope(userCoroutineContext + handlerJob)` breaks call logging plugin when the call is cancelled

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/http1/NettyHttp1Handler.kt
```diff
@@ -120,10 +133,36 @@ internal class NettyHttp1Handler(
...
+            val callScope = CoroutineScope(context = callContext)
+            callScope.launch(start = CoroutineStart.UNDISPATCHED) {
+                try {
+                    if (!call.request.isValid()) {
+                        call.respondError400BadRequest()
+                        return@launch
+                    }
+                    enginePipeline.execute(call)
+                } catch (error: Throwable) {
+                    handleFailure(call, error)
```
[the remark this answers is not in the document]
zibet27: No, it's defined

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/http2/NettyHttp2Handler.kt
```diff
@@ -68,17 +69,27 @@ internal class NettyHttp2Handler(
...
+        onStreamClose(context)
+        context.fireChannelInactive()
+    }
+
+    private fun onStreamClose(context: ChannelHandlerContext) {
+        context.applicationCall?.let { call ->
+            context.applicationCall = null
+            @OptIn(InternalAPI::class)
+            call.attributes.getOrNull(HttpRequestCloseHandlerKey)?.invoke()
+        }
```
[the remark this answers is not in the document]
zibet27: It should nopt be a problem as they run on the same Netty thread

zibet27:
PS: `call.coroutineContext` has netty dispatcher which is single threaded, that's why we don't use `callEventGroup` not to change threads inside of call. https://github.com/ktorio/ktor/pull/5421/changes#diff-2fe7824f7116f099d8d4afcd45c523ae25c386423f14ff231c775a3775e5d4ebR150

ktor-server/ktor-server-cio/jvm/test/io/ktor/tests/server/cio/CIOEngineTestJvm.kt
```diff
@@ -95,4 +95,7 @@ class CIOHttpRequestLifecycleTest :
         enableSsl = false
         enableHttp2 = false
     }
+
+    @Ignore
+    override fun testPipelinedRequestsCancelledOnDisconnect() {}
```
bjhham, line 100: Maybe you could log a new issue to handle this behaviour in CIO if it falls outside the scope here.

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/http1/NettyHttp1Handler.kt
```diff
@@ -120,10 +134,36 @@ internal class NettyHttp1Handler(
     }
 
     private fun handleRequest(context: ChannelHandlerContext, message: HttpRequest) {
-        val call = prepareCallFromRequest(context, message)
+        val userAppContext = applicationProvider().coroutineContext + userContext
+        val callJob = Job(parent = userAppContext[Job])
 
+        val callContext = userAppContext + NettyDispatcher.CurrentContext(context) + callJob + CallHandlerCoroutineName
```
bjhham, line 140: The netty dispatcher ought to be present in `userContext` here already so I think `NettyDispatcher.CurrentContext(context)` can be removed.

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/http1/NettyHttp1Handler.kt
```diff
@@ -120,10 +134,36 @@ internal class NettyHttp1Handler(
...
+                    }
+                    enginePipeline.execute(call)
+                } catch (error: Throwable) {
+                    handleFailure(call, error)
+                } finally {
+                    activeCalls.remove(call)
+                    callJob.complete()
+                }
+            }
+        }
```
bjhham, line 170: Since the user context is already on the netty dispatcher, couldn't we just have:

```
CoroutineScope(callContext).launch {
    try {
        if (!call.request.isValid()) {
            call.respondError400BadRequest()
            return@launch
        }
        enginePipeline.execute(call)
    } catch (error: Throwable) {
        handleFailure(call, error)
    } finally {
        activeCalls.remove(call)
        callJob.complete()
    }
}
```

ktor-server/ktor-server-cio/jvm/test/io/ktor/tests/server/cio/CIOEngineTestJvm.kt
zibet27: Afaik, CIO doesn't support pipelining, does it?

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/http1/NettyHttp1Handler.kt
zibet27: Yeah, we could do that
The idea was that `context.executor().execute {}` makes the call more granular; 
For instance, some other event could be dispatched.
Not sure if it makes much sense here

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/http1/NettyHttp1Handler.kt
zibet27: `NettyDispatcher.CurrentContext` is created only in handlers
Or my IDE can't find all usages (which happens)

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/http1/NettyHttp1Handler.kt
bjhham: Hmm I don't think you'll get other events for this request at this point, and using the executor with a new runnable is going to introduce a little overhead.  Maybe you could play around with it to investigate if it provides material benefits.  My main concern is introducing unneeded complexity and adding more deep references to the deeper Netty API.

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/http1/NettyHttp1Handler.kt
bjhham: Oh yeah, look at that... huh.  Ok 👍

bjhham (approved):
I'll leave the context executor call up to you, though I'd advise adding a comment to explain what's going on there.  Other than that, looks like a nice fix 👍

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/http1/NettyHttp1Handler.kt
zibet27: Hmm, I've tried to remove it, but some tests were failing
For example, early flushing of headers was broken in HTTP1 and got some EOF exceptions in HTTP2

So, I've added comments on why extra dispatching is needed

bjhham: Ok, good to know, thanks for checking 👍

PR #5422, opened 2026-03-04
Update Agents.md to run tests before assemble

zibet27:
**Motivation**
Agents first run assemble and then only tests which use more tokens and take more time

**Solution**
- Mention to run tests first

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -5,11 +5,12 @@ This file is the primary playbook for agentic coding tools working in this repos
 ## Critical Workflow Rule
 
 **NEVER return control to the user without ensuring:**
-1. Code compiles: `./gradlew :module-name:assemble`
-2. Tests pass: `./gradlew :module-name:jvmTest` (and other platforms if touched)
+1. Tests pass: `./gradlew :module-name:jvmTest` (and other platforms if touched)
+2. Code compiles: `./gradlew :module-name:assemble`
 3. Code is formatted and linting passes: run `./gradlew :module-name:formatKotlin`, then `./gradlew :module-name:lintKotlin`
 4. ABI validated: `./gradlew :module-name:updateLegacyAbi` (if public/protected API changed)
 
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -5,11 +5,12 @@ This file is the primary playbook for agentic coding tools working in this repos
 ## Critical Workflow Rule
 
 **NEVER return control to the user without ensuring:**
-1. Code compiles: `./gradlew :module-name:assemble`
-2. Tests pass: `./gradlew :module-name:jvmTest` (and other platforms if touched)
+1. Tests pass: `./gradlew :module-name:jvmTest` (and other platforms if touched)
+2. Code compiles: `./gradlew :module-name:assemble`
 3. Code is formatted and linting passes: run `./gradlew :module-name:formatKotlin`, then `./gradlew :module-name:lintKotlin`
 4. ABI validated: `./gradlew :module-name:updateLegacyAbi` (if public/protected API changed)
```
osipxd, line 11: We could ask it to use a one-liner (after the list):
```bash
./gradlew :module-name:assemble :module-name:<target>Test :module-name:formatKotlin :module-name:lintKotlin
```
What do you think?

zibet27: I think we can ask it to first run tests and then use a one-liner for `assemble`, `format`, and `lint`

If we use only one one-liner and an error occurs, I guess it will try to rerun all tasks sequentially
Though it's just a guess, and I'll try to experiment with that

osipxd: By the way we could add a general rule that when agent needs to run multiple Gradle tasks it should pass all these tasks as arguments to a single `./gradlew` command.

e5l (approved):
lgtm

osipxd:
Meh. CI needs to be started manually as we have unconditional requirement for checks, but this PR contains only *.md changes, so build hasn't started.

Issue #5309, opened 2026-01-18
Certificate pinning matches against all pins instead of hostname-scoped pins


HighOverseer:
**Description**

In Darwin's CertificatePinner (ktor-client/ktor-client-darwin/darwin/src/io/ktor/client/engine/darwin/certificates/CertificatePinner.kt), findMatchingPins(hostname) is computed but not used for pins validation.

hasOnePinnedCertificate matches the server certificate chain against all configured pins, not just the pins that match the current hostname.

// applyPinning()
val matchingPins = findMatchingPins(hostname)
...
hasOnePinnedCertificate(certificates) // uses all pins

<img width="828" height="768" alt="Image" src="https://github.com/user-attachments/assets/b1c6321f-9caa-4c5d-aa86-881e2d907529" />

// hasOnePinnedCertificate()
pinnedCertificates.any { pin -> ... }

<img width="828" height="694" alt="Image" src="https://github.com/user-attachments/assets/ba061f89-2ec2-4052-a177-5faef502a683" />

**Impact**

When multiple domains are pinned, a pin for domain A can satisfy pinning for domain B, breaking hostname-scoped pin isolation.

**Expected**

Certificate validation should only consider matchingPins.

**Suggested Fix**

Pass matchingPins into hasOnePinnedCertificate and validate against that set instead of pinnedCertificates.

Stexxe:
Thank you for the report.

zibet27:
Fixed in https://github.com/ktorio/ktor/pull/5423

PR #5423, opened 2026-03-04
Darwin Client. Use correct pins for certificate pinning

zibet27:
**Subsystem**
Darwin Client

**Motivation**
See https://github.com/ktorio/ktor/issues/5309
YouTrack issue: [KTOR-9393](https://youtrack.jetbrains.com/issue/KTOR-9393) Certificate pinning matches against all pins instead of hostname-scoped pins

**Solution**
The solution is already described in the issue

## Release Notes – Version 3.4.1

* **New Features**
  * Added configurable dispatcher switching behavior for HTTP requests (enabled on non-JVM platforms, configurable on JVM via system property).
  * Enhanced JWT authentication with automatic EC key algorithm detection.

* **Bug Fixes**
  * Improved resource cleanup in native HTTP engines.
  * Enhanced error handling for socket timeout scenarios.

* **Tests**
  * Added dispatcher behavior validation tests across platforms.
  * Added JWT authentication tests for elliptic curve keys.
  * Added routing trace output validation tests.

* **Chores**
  * Applied security-focused dependency constraints.
  * Updated routing trace display format.

### DIFF
--- a/ktor-client/ktor-client-darwin/darwin/src/io/ktor/client/engine/darwin/certificates/CertificatePinner.kt
+++ b/ktor-client/ktor-client-darwin/darwin/src/io/ktor/client/engine/darwin/certificates/CertificatePinner.kt
@@ -173,8 +173,8 @@ public data class CertificatePinner(
             throw TlsPeerUnverifiedException("Unknown certificates")
         }
 
-        if (!hasOnePinnedCertificate(certificates)) {
-            val message = buildErrorMessage(certificates, hostname)
+        if (!hasOnePinnedCertificate(certificates, matchingPins)) {
+            val message = buildErrorMessage(certificates, hostname, matchingPins)
             throw TlsPeerUnverifiedException(message)
         }
         return true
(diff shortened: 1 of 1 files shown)


zibet27:
> 👍 Do you think it is possible to write a test?

@osipxd, afaik, we don't test certificate pinning anywhere
I can try to add common integration tests. Should I?

osipxd:
That would be nice
If it turns out to be impossible with our test-server, we could use test hosts from https://badssl.com/
However I'm not sure they don't rotate certificates

zibet27:
> That would be nice If it turns out to be impossible with our test-server, we could use test hosts from https://badssl.com/ However I'm not sure they don't rotate certificates

https://badssl.com/ rotates certificates, so I wouldn't rely on them
OkHttp use nice [MockWebServer](https://github.com/square/okhttp/tree/master/mockwebserver) for testing, but it's only available on jvm :(
I'll try to extend `ktor-test-server`

PR #5425, opened 2026-03-04
KTOR-9187 Darwin Client. Close session once engine is closed

zibet27:
**Subsystem**
Darwin Client

**Motivation**
[KTOR-9187](https://youtrack.jetbrains.com/issue/KTOR-9187) Darwin: reference captured in `handleChallenge` is never deinitialized, leading to a memory leak

**Solution**
- call `DarwinSession.close` which was implemented, but not used


zibet27:
@osipxd Should we merge it into `release/3.x` now?

PR #5430, opened 2026-03-09
fix: improve error message when deserializing generic type with non-@Serializable type parameter

### LINKED ISSUE #4675: Ktor client using KxS throws unhelpful error when trying to decode non-serializable class boxed in a generic container
Given that I make any request that tries to decode something to a `List<T>`, where `T` is not serializable, Ktor-client will generate an error like this:

```
kotlinx.serialization.SerializationException: Serializer for class 'List' is not found.
Please ensure that class is marked as '@Serializable' and that the serialization compiler plugin is applied.

	at kotlinx.serialization.internal.Platform_commonKt.serializerNotRegistered(Platform.common.kt:90)
	at kotlinx.serialization.SerializersKt__SerializersKt.serializer(Serializers.kt:299)
	at kotlinx.serialization.SerializersKt.serializer(Unknown Source)
	at io.ktor.serialization.kotlinx.SerializerLookupKt.serializerForTypeInfo(SerializerLookup.kt:32)
	at io.ktor.serialization.kotlinx.KotlinxSerializationConverter.deserialize(KotlinxSerializationConverter.kt:66)
	at io.ktor.serialization.ContentConverterKt$deserialize$$inlined$map$1$2.emit(Emitters.kt:51)
```

Adding `@Serializable` to `T` fixed this in my case. 

It's very frustrating and throws the user in the wrong direction. It seems to apply to all boxed types. For instance 
```kotlin
data class Foo(val bar: Int)

@Serializable
data class Box<out T>(val value: T)

fun main() {
    client.get("").body<Box<Foo>>()
}
```

Will generate a similar error stating`kotlinx.serialization.SerializationException: Serializer for class 'Box' is not found.`
Once again, making Foo serializable would fix this

---
Ktor version: 3.1.0
KxS version: 1.8.0
Kotlin version: 2.1.0

kkh725:
## Problem                                                                                                   
                                                                                                               
  Fixes #4675                                                                                                  

  When deserializing a generic type like `List<Foo>` or `Box<Foo>` where `Foo` is not marked as                
  `@Serializable`, the error message incorrectly blamed the container class:
                                                                                                               
  kotlinx.serialization.SerializationException: Serializer for class 'List' is not found.                      
  Please ensure that class is marked as '@Serializable' and that the serialization compiler plugin is applied.
                                                                                                               
  This sends the user in the wrong direction — `List` is perfectly serializable; the real problem is the type
  argument `Foo`.                                                                                              
  
  ## Root Cause                                                                                                
                  
  In `SerializerLookup.kt`, `serializerForTypeInfo` calls `module.serializerOrNull(type)` which returns `null` 
  when any type argument is not serializable. The code then falls through to `typeInfo.type.serializer()` (the
  raw class, without type arguments), which throws an error naming the container class instead of the          
  problematic type parameter.

  ## Fix

  When `serializerOrNull(type)` returns `null` and the type has arguments, inspect each argument to find which 
  ones are missing a serializer. If any are found, throw a `SerializationException` that names them directly:
                                                                                                               
  kotlinx.serialization.SerializationException: Serializer for class 'List' is not found.
  Ensure that the following type parameter is marked as '@Serializable': 'Foo'

  ## Changes

  - **`SerializerLookup.kt`**: detect non-serializable type arguments and include their names in the exception 
  message
  - **`JsonSerializationTest.kt`**: add two regression tests covering `List<NonSerializable>` and              
  `Box<NonSerializable>` cases

### DIFF
--- a/ktor-shared/ktor-serialization/ktor-serialization-kotlinx/common/src/io/ktor/serialization/kotlinx/SerializerLookup.kt
+++ b/ktor-shared/ktor-serialization/ktor-serialization-kotlinx/common/src/io/ktor/serialization/kotlinx/SerializerLookup.kt
@@ -56,14 +56,17 @@ private fun checkTypeParameters(type: KType, typeInfo: TypeInfo, module: Seriali
 
     // Format message with the failed type parameters
     val argNames = nonSerializableArgs.joinToString {
-        when(val clz = it.classifier) {
+        when (val clz = it.classifier) {
             is KClass<*> -> "'${clz.simpleName}'"
             else -> "'$it'"
         }
     }
     val (s, be) =
-        if (nonSerializableArgs.size == 1) "" to "is"
-        else "s" to "are"
+        if (nonSerializableArgs.size == 1) {
+            "" to "is"
+        } else {
+            "s" to "are"
+        }
     throw SerializationException(
         "Serializer for type argument$s $argNames $be not found for '${typeInfo.type.simpleName}'. " +
             "Ensure that the listed type$s $be marked as '@Serializable'."


kkh725:
The CI failures are unrelated to the changes in this PR.

All failing tests are in different modules (`ktor-client` timeout tests, `ktor-auth` OAuth2 tests) and have no connection to the files changed here (`SerializerLookup.kt`, `JsonSerializationTest.kt`).

- **HTTP timeout failures** (`/timeout/with-delay`, `/content/upload`): `HttpRequestTimeoutException` across CIO, Apache5, Android, OkHttp engines — caused by CI server load, not a code issue.
- **OAuth2Test.testFailedNonce[mingwX64]**: `IllegalStateException: This job has not completed yet` in `tearDown` — a known timing/flaky issue on the mingwX64 environment.

Could someone re-run the CI? Thank you!

ktor-shared/ktor-serialization/ktor-serialization-kotlinx/common/src/io/ktor/serialization/kotlinx/SerializerLookup.kt
```diff
@@ -25,7 +26,35 @@ public fun SerializersModule.serializerForTypeInfo(typeInfo: TypeInfo): KSeriali
...
+                                " not found for '${typeInfo.type.simpleName}'. " +
+                                "Ensure that the listed " +
+                                (if (nonSerializableArgs.size == 1) "type is" else "types are") +
+                                " marked as '@Serializable'."
+                        )
+                    }
+                    null
+                }
             }
         }
```
bjhham: I think we can clean this up a bit with a function extraction and include some comments for readability. 

```suggestion
                module.serializerOrNull(type) ?: checkTypeParameters(module, typeInfo)
            }
        }
        ?: module.getContextual(typeInfo.type)?.maybeNullable(typeInfo)
        ?: typeInfo.type.serializer().maybeNullable(typeInfo)
}

/**
 * Provide a better formatted exception message when type parameters are the cause of a missing serializer.
 */
private fun checkTypeParameters(module: SerializersModule, typeInfo: TypeInfo): KSerializer<*>? {
    // If there was some other cause, return null to fallback to default handling
    val type = typeInfo.kotlinType ?: return null
    val nonSerializableArgs = type.arguments
        .mapNotNull { arg ->
            try {
                arg.type?.takeIf { module.serializerOrNull(it) == null }
            } catch (_: Exception) {
                return null
            }
        }
    if (nonSerializableArgs.isEmpty()) return null

    // Format message with the failed type parameters
    val argNames = nonSerializableArgs.joinToString {
        when(val clz = it.classifier) {
            is KClass<*> -> "'${clz.simpleName}'"
            else -> "'$it'"
        }
    }
    val (s, be) =
        if (nonSerializableArgs.size == 1) "" to "is"
        else "s" to "are"

    throw SerializationException(
        "Serializer for type argument$s $argNames $be not found for '${typeInfo.type.simpleName}'. " +
            "Ensure that the listed type$s $be marked as '@Serializable'."
    )
}
```

bjhham (changes requested):
Thanks for the fix, looks like it will be very helpful!

Please check the comment on styling and also run `./gradlew formatKotlin` to resolve the linting error in the CI.

I also noticed that the solution only resolves a single layer of parameterization, so something like `List<List<NonSerializable>>` would have the same general issue.  This is a pretty uncommon scenario and introducing traversal would need to guard against cycles, so maybe it's not worth handling it.

kkh725:
@bjhham Hi, I wanted to note that all 3 CI failures appear to be unrelated to the changes in this PR.
                                                                                                                                        
 The failures are all caused by the Homebrew bump-formula-pr workflow conflicting with the autobump policy — the ktor formula is 
  already handled automatically by BrewTestBot, so any manual bump attempt gets rejected. This seems to be a pre-existing issue with the
   CI workflow configuration rather than anything introduced by this PR.                                                                
                                                                                                                                        
 Could you take a look and confirm? Happy to re-run the checks if needed. Thank you!

ktor-shared/ktor-serialization/ktor-serialization-kotlinx/common/src/io/ktor/serialization/kotlinx/SerializerLookup.kt
```diff
@@ -25,13 +26,57 @@ public fun SerializersModule.serializerForTypeInfo(typeInfo: TypeInfo): KSeriali
...
+        .mapNotNull { it.type }
+        .filter { argType ->
+            try {
+                module.serializerOrNull(argType) == null
+            } catch (_: Throwable) {
+                // If lookup itself throws, we cannot reliably determine the cause; fall through
+                lookupFailed = true
+                false
+            }
+        }
```
bjhham: Preference for short-circuit return here.

```suggestion
    val nonSerializableArgs = type.arguments
        .mapNotNull { arg ->
            try {
                arg.type?.takeIf { module.serializerOrNull(it) == null }
            } catch (_: Exception) {
                // If lookup itself throws, we cannot reliably determine the cause; fall through
                return null
            }
        }
```

ktor-shared/ktor-serialization/ktor-serialization-kotlinx/common/src/io/ktor/serialization/kotlinx/SerializerLookup.kt
```diff
@@ -25,13 +26,57 @@ public fun SerializersModule.serializerForTypeInfo(typeInfo: TypeInfo): KSeriali
...
+
+    // If any argument lookup threw an exception, defer to the default error path
+    if (lookupFailed) return null
```
bjhham:
```suggestion
if (nonSerializableArgs.isEmpty()) return null
```

ktor-shared/ktor-serialization/ktor-serialization-kotlinx/common/src/io/ktor/serialization/kotlinx/SerializerLookup.kt
```diff
@@ -25,13 +26,57 @@ public fun SerializersModule.serializerForTypeInfo(typeInfo: TypeInfo): KSeriali
...
+            "Type " +
+                (if (nonSerializableArgs.size == 1) "argument $argNames is" else "arguments $argNames are") +
+                " not serializable (used as a type parameter for '${typeInfo.type.simpleName}'). " +
+                "Ensure the listed " +
+                (if (nonSerializableArgs.size == 1) "type is" else "types are") +
+                " marked as '@Serializable'."
+        )
+    }
+
+    return null
```
bjhham: Extract variables to make the string templating more legible + conditional branching lifted out into short-circuit.

```suggestion
    // Format message with the failed type parameters
    val argNames = nonSerializableArgs.joinToString {
        when(val clz = it.classifier) {
            is KClass<*> -> "'${clz.simpleName}'"
            else -> "'$it'"
        }
    }
    val (s, be) =
        if (nonSerializableArgs.size == 1) "" to "is"
        else "s" to "are"
    throw SerializationException(
        "Serializer for type argument$s $argNames $be not found for '${typeInfo.type.simpleName}'. " +
            "Ensure that the listed type$s $be marked as '@Serializable'."
    )
```

kkh725:
Hi @bjhham, thank you for the detailed feedback!

I've addressed all three of your formatting suggestions from today:
  - Added an explanatory comment to the catch block in checkTypeParameters                                                      
  - Simplified the empty check to a single-line return null                                                                     
  - Extracted argNames and (s, be) variables and lifted the conditional branching as you suggested
                                                                                                                                
  Could you take another look when you get a chance? Thanks!

ktor-shared/ktor-serialization/ktor-serialization-kotlinx/common/src/io/ktor/serialization/kotlinx/SerializerLookup.kt
```diff
@@ -25,13 +26,50 @@ public fun SerializersModule.serializerForTypeInfo(typeInfo: TypeInfo): KSeriali
...
+        .mapNotNull { arg ->
+            try {
+                arg.type?.takeIf { module.serializerOrNull(it) == null }
+            } catch (_: Exception) {
+                // If lookup itself throws, we cannot reliably determine the cause; fall through
+                return null
+            }
+        }
+
+if (nonSerializableArgs.isEmpty()) return null
```
bjhham: Indentation 👀 
```suggestion
    if (nonSerializableArgs.isEmpty()) return null
```

kkh725:
Run `./gradlew formatKotlin` to fix linting errors introduced by applying review suggestions via GitHub web UI.

- `when(` → `when (` (keyword-spacing)
- single-line `if/else` → wrapped with braces (if-else-wrapping, multiline-if-else)

PR #5437, opened 2026-03-11
KTOR-9398 Client Core. Cancel streaming on exception 

zibet27:
**Subsystem**
Client Core

**Motivation**
[KTOR-9398](https://youtrack.jetbrains.com/issue/KTOR-9398) Streaming call not cancelled when exception is thrown from HttpStatement.execute/body

**Solution**
- add cause to `Response.cleanup`

### DIFF
--- a/ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/HttpStatementTest.kt
+++ b/ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/HttpStatementTest.kt
@@ -74,8 +74,11 @@ class HttpStatementTest : ClientLoader() {
         }
     }
 
+    // Darwin/DarwinLegacy: NSURLSession buffers the first 512 bytes before calling didReceiveResponse/didReceiveData,
+    // so the test times out waiting for enough data to arrive unless the content type is octet/stream or application/json.
+    // See: https://developer.apple.com/forums/thread/64875
     @Test
-    fun testStreamingResponseExceptionCancelsImmediately() = clientTests(except("Darwin", "DarwinLegacy")) {
+    fun testStreamingResponseExceptionCancelsImmediately() = clientTests {
         test { client ->
             val exception = assertFailsWith<IllegalStateException> {
                 withTimeout(2000) {
@@ -90,7 +93,7 @@ class HttpStatementTest : ClientLoader() {
     }
 
     @Test
-    fun testStreamingResponseExceptionInBodyCancelsImmediately() = clientTests(except("Darwin", "DarwinLegacy")) {
+    fun testStreamingResponseExceptionInBodyCancelsImmediately() = clientTests {
         test { client ->
             val exception = assertFailsWith<IllegalStateException> {
                 withTimeout(2000) {
--- a/ktor-test-server/src/main/kotlin/test/server/tests/Content.kt
+++ b/ktor-test-server/src/main/kotlin/test/server/tests/Content.kt
@@ -110,6 +110,9 @@ internal fun Application.contentTestServer() {
                 val delay = call.parameters["delay"]?.toLong() ?: 0L
                 call.respond(
                     object : OutgoingContent.WriteChannelContent() {
+                        override val contentType: ContentType
+                            get() = ContentType.Application.OctetStream
+
                         override suspend fun writeTo(channel: ByteWriteChannel) {
                             while (true) {
                                 channel.writeInt(42)


zibet27:
PS: Should we keep binary compatibility for the `internal` published API?

osipxd:
> PS: Should we keep binary compatibility for the internal published API?

Yes, since it's marked with `@PublishedApi` it can be used from compiled code as a result of caller's inlining.

ktor-server/ktor-server-test-host/jvm/src/io/ktor/server/testing/client/TestEngineWebsocketSession.kt
```diff
@@ -30,14 +29,15 @@ internal class TestEngineWebsocketSession(
...
     suspend fun run() {
-        outgoing.invokeOnClose {
-            if (it != null) {
-                socketJob.completeExceptionally(it)
-            } else {
-                socketJob.complete()
+        suspendCancellableCoroutine { cont ->
+            outgoing.invokeOnClose { ex ->
+                if (ex == null) {
+                    cont.resume(Unit)
```
e5l: can throw if cont cancelled

ktor-client/ktor-client-core/common/src/io/ktor/client/statement/HttpStatement.kt
```diff
@@ -204,16 +213,28 @@ public class HttpStatement(
...
     /**
      * Completes [HttpResponse] and releases resources.
+     *
+     * @param cause If not null, cancels the response job with this cause to immediately interrupt
+     * any pending network operations.
      */
     @PublishedApi
     @OptIn(InternalAPI::class)
-    internal suspend fun HttpResponse.cleanup() {
+    internal suspend fun HttpResponse.cleanup(cause: Throwable? = null) {
```
gervaisj: Just commenting here because I opened the ticket on YouTrack and I'm interested in the details of the fix:

Wouldn't it be better to remove the default `null` value of this `cause` param, and pass in `null` explicitly in the call in `fetchResponse()`, such as not to add a `cleanup$default` method to the published JVM API?

zibet27: Makes sense
Thank you for the comment!

gervaisj (commented):
I think it's good to go, but I have been thinking about I slight extension to the issue, if your workflow permits adding to the scope a bit.

Specifically, what if, instead of throwing an exception, `body` or `execute` simply completes? (like in the snippet below)

```kotlin
client.prepareGet("...").body { channel: ByteReadChannel ->
    while (true) {
        // imagine a line is sent by the server only every minute
        val line = channel.readLine()
        if (line == "text I'm interested in")
            return@body
    }
}
```
Then, wouldn't it be nice to have the same behavior as when throwing an exception, i.e. have the streaming call cancel immediately, without waiting for another frame to arrive?

I feel this would match the already documented behavior of `execute`:
> The response object should not be accessed outside of [block] as it will be canceled upon block completion.

A similar line is also present on `body`. Note that "block completion" here makes no distinction between exceptional and normal completion.

So I've tried to see if `cleanup` could always `cancel` the call in this commit: https://github.com/gervaisj/ktor/commit/76a525cd9dc820482ea01e943bf88e0452788552 (which is on this branch: [gervaisj/client-always-cancel-streaming](https://github.com/gervaisj/ktor/commits/gervaisj/client-always-cancel-streaming/)).

The only issue I ran across were the tests in `ServerSentEventsTest` failing, but this was because `HttpClient.processSession` (which is used by all SSE functions) did not respect `body`'s contract of not using the response after completion of the `block`. That should be fixed in the commit referenced above, along with 2 new test cases, which are basically the same as those added by @zibet27, but for normal completion of `block` instead of throwing an exception.

@zibet27 @e5l what do you think?

zibet27:
> I feel this would match the already documented behavior of `execute`:

Agree here, afaiu it's intended behaviour and "works" on some engines.  I tried to run your tests on this branch, and it succeeded on OkHttp, Apache and Android, but not on CIO and Java @e5l @bjhham?

> So I've tried to see if `cleanup` could always `cancel` the call in this commit: [gervaisj@76a525c](https://github.com/gervaisj/ktor/commit/76a525cd9dc820482ea01e943bf88e0452788552) (which is on this branch: [gervaisj/client-always-cancel-streaming](https://github.com/gervaisj/ktor/commits/gervaisj/client-always-cancel-streaming/)).

Looks good to me! 
I would create a separate pull request for it.

gervaisj:
> I tried to run your tests on this branch, and it succeeded on OkHttp, Apache and Android, but not on CIO and Java

Without cancelling the call unconditionally (like is done in my commit), then yes those tests won't work because we fall in the `complete()` branch, and so hang on `join()`, and then time out.

> I would create a separate pull request for it.

Sure!

gervaisj:
> I would create a separate pull request for it.

@zibet27 Sorry, maybe there was a misunderstanding, were you waiting for an action on my part? Is there something missing for this PR?

zibet27:
> Is there something missing for this PR?

@gervaisj No, I'm waiting for approvals

ktor-client/ktor-client-core/common/src/io/ktor/client/statement/HttpStatement.kt
```diff
@@ -199,21 +208,33 @@ public class HttpStatement(
...
     @PublishedApi
     @OptIn(InternalAPI::class)
-    internal suspend fun HttpResponse.cleanup() {
+    internal suspend fun HttpResponse.cleanup(cause: Throwable?) {
         val job = coroutineContext.job as CompletableJob
 
         job.apply {
-            complete()
+            if (cause != null) {
+                cancel(CancellationException("Exception occurred during request execution", cause))
```
osipxd: There is a chance the `cause` is already `CancellationException`. Probably we shouldn't wrap it with another `CancellationException` in this case.
Please add a test checking this behavior if you decide to fix it.

ktor-client/ktor-client-core/api/ktor-client-core.api
```diff
@@ -1622,7 +1622,8 @@ public final class io/ktor/client/statement/HttpResponsePipeline$Phases {
 
 public final class io/ktor/client/statement/HttpStatement {
 	public fun <init> (Lio/ktor/client/request/HttpRequestBuilder;Lio/ktor/client/HttpClient;)V
-	public final fun cleanup (Lio/ktor/client/statement/HttpResponse;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
+	public final fun cleanup (Lio/ktor/client/statement/HttpResponse;Ljava/lang/Throwable;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
+	public final synthetic fun cleanup (Lio/ktor/client/statement/HttpResponse;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
```
osipxd, line 1626: Let's switch the base branch to main

ktor-client/ktor-client-core/common/src/io/ktor/client/statement/HttpStatement.kt
zibet27: `HttpResponse.cleanup` is not tested directly anywhere
Should we?

e5l (approved):

PR #5445, opened 2026-03-11
KTOR-6882 Fix SuspendFunctionGun incompatibility with ES2015 generators

osipxd:
[KTOR-9405](https://youtrack.jetbrains.com/issue/KTOR-9405) Re-enable SuspendFunctionGun in client pipelines

## Summary

- Fixes `SuspendFunctionGun` (SFG) incorrectly interacting with ES2015 generator-based coroutines on Kotlin/JS
- Reverts the workaround from #4109 that disabled SFG for client pipelines

## Root cause

When `target = es2015` + `MODULE_ES` is used, Kotlin/JS compiles coroutines via ES2015 generator functions (`function*`) managed by `GeneratorCoroutineImpl`. The old `PipelineJs.kt` called interceptors through `suspendOrReturn`, which does **not** set `isRunning = true` before execution. When an interceptor called `proceed()` synchronously (or used `suspendCoroutineUninterceptedOrReturn { it.resume(Unit); COROUTINE_SUSPENDED }`), a `resumeWith()` call inside a running generator tried to call `iterator.next()` on an already-executing generator — causing a silent "Generator is already running" TypeError that broke the coroutine continuation stack.

The fix uses `startCoroutineUninterceptedOrReturn`, which routes through `startCoroutineFromGeneratorFunction` and properly sets `isRunning = true` before execution, deferring any `resumeWith()` calls via `savedResult` until after the current step completes.

## Changes

- Replaced `PipelineJs.kt` (JS) and `PipelineWasm.kt` (wasmJs) with a shared `Pipeline.web.kt` in the `web` source set using `startCoroutineUninterceptedOrReturn`
- Reverted #4109 which disabled SFG in client pipelines as a workaround
- Removed leftover `developmentMode = true` workarounds from `AuthTest`

Also fixes https://youtrack.jetbrains.com/issue/KTOR-6878
Also fixes https://youtrack.jetbrains.com/issue/KTOR-6882

### DIFF
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/HttpClient.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/HttpClient.kt
@@ -1305,28 +1305,28 @@ public class HttpClient(
      *
      * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.client.HttpClient.requestPipeline)
      */
-    public val requestPipeline: HttpRequestPipeline = HttpRequestPipeline()
+    public val requestPipeline: HttpRequestPipeline = HttpRequestPipeline(userConfig.developmentMode)
 
     /**
      * A pipeline used for processing all responses sent by the server.
      *
      * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.client.HttpClient.responsePipeline)
      */
-    public val responsePipeline: HttpResponsePipeline = HttpResponsePipeline()
+    public val responsePipeline: HttpResponsePipeline = HttpResponsePipeline(userConfig.developmentMode)
 
     /**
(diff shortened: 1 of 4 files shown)


osipxd:
After measuring allocation impact, re-enabling SuspendFunctionGun in the client pipelines does not provide a noticeable gain on JVM — it is actually slightly worse (+264 bytes/request net) because SFG itself allocates 760 bytes (object + `suspensions` array), while `DebugPipelineContext` only costs 528 bytes.

The underlying correctness fix for ES2015 generators (`Pipeline.web.kt`) has been moved into the ES modules PR (#5396). The revert of the `developmentMode = true` workaround will not be applied.

Closing this PR in favor of #5396.

PR #5455, opened 2026-03-19
KTOR-9408 Upgrade atomicfu to fix cache performance

bjhham:
**Subsystem**
Core, Caching

**Motivation**
[KTOR-9408](https://youtrack.jetbrains.com/issue/KTOR-9408) Performance regression in atomic fu affects cache

**Solution**
Downgrade atomicfu for now.  This requires some version pinning because it's also used in coroutines.


bjhham:
FYI @fzhinkin

build-logic/src/main/kotlin/ktorbuild.kmp.gradle.kts
```diff
@@ -26,6 +26,17 @@ if (!project.hasAndroidPlugin()) {
     }
 }
 
+dependencies {
+    constraints {
+        // KTOR-9408 Remove after performance issue is resolved
```
osipxd: Let's move this into `because("....")` inside the `versions { }` block. This way this info will be visible in dependencyInsight task for Ktor consumers

osipxd (approved):

osipxd:
A new version [0.32.0](https://github.com/Kotlin/kotlinx-atomicfu/blob/master/CHANGES.md#version-0320) is here. Shall we upgrade?

bjhham:
Oh, that was quick! 😄

fzhinkin:
The new version of the AFU runtime with reverted `SynchronizedObject` implementation is already available (`0.32.0`, there were some issues with publishing the Gradle plugin and I'll roll out `0.32.1` today), so it might make sense to update to that version instead.

**UPD**: oh, you already did that. :)

PR #5457, opened 2026-03-19
KTOR-9361 Fix JsWebSocketSession._closeReason completed twice

osipxd:
**Subsystem**
Client Js

**Motivation**
[KTOR-9361](https://youtrack.jetbrains.com/issue/KTOR-9361) WebSockets: `JsWebSocketSession._closeReason` is completed twice

**Solution**

In `awaitConnection()`, the combined open/error event listener was never removed after the open event fired. When an error occurred on an established connection, the stale listener tried to resume the already- completed continuation, logging "Already resumed, but proposed with update". Fixed by using `lateinit var` so the listener can remove itself on first fire.


osipxd:
@coderabbitai review
@codex review

PR #5486, opened 2026-03-20
Add nullability support for server parameter delegation

nomisRev:
**Subsystem**
Server, core

**Motivation**
Currently it is not possible to use nullable values with parameter delegation.

```kotlin
get("/nullable") {
    val nameOrNull: String? by call.queryParameters
    val ageOrNull: Int? by parameters
}
```

**Solution**

This PR add support by removing the `: Any` constraint and relying on `typeInfo.kotlinType?.isMarkedNullable` to check if null is allowed or not.


nomisRev:
I'm a bit unsure how reliable `typeInfo.kotlinType?.isMarkedNullable` is across platforms?

bjhham (approved):
Looks like an easy win 👍 

I'm pretty sure `isMarkedNullable` works cross-platform on reified KTypes.

I don't know why we didn't just implement it that way in the first place...

bjhham:
FYI logged an issue to handle property delegation in the compiler plugin:
[KTOR-9422](https://youtrack.jetbrains.com/issue/KTOR-9422) OpenAPI code inference misses property delegation

ktor-server/ktor-server-core/common/src/io/ktor/server/util/Parameters.kt
```diff
@@ -27,11 +27,18 @@ import kotlin.reflect.*
...
+ * If [R] is nullable and no values are associated with the delegated property name, returns `null`.
+ *
+ * @throws MissingRequestParameterException if no values associated with name and [R] is not nullable
  * @throws ParameterConversionException when conversion from String to [R] fails
  */
-public inline operator fun <reified R : Any> Parameters.getValue(thisRef: Any?, property: KProperty<*>): R {
-    return getOrFail<R>(property.name)
+public inline operator fun <reified R> Parameters.getValue(thisRef: Any?, property: KProperty<*>): R {
+    val typeInfo = typeInfo<R>()
+    return if (typeInfo.kotlinType?.isMarkedNullable == true && get(property.name) == null) {
```
osipxd: Alternatively we can use `null is R` if we want to avoid creating `TypeInfo` instance

ktor-server/ktor-server-core/common/src/io/ktor/server/util/Parameters.kt
```diff
@@ -27,11 +27,18 @@ import kotlin.reflect.*
  *
  * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.server.util.getValue)
  *
- * @throws MissingRequestParameterException if no values associated with name
+ * If [R] is nullable and no values are associated with the delegated property name, returns `null`.
```
osipxd: Let's move this above `[Report a problem]` link

osipxd (changes requested):

ktor-server/ktor-server-core/common/src/io/ktor/server/util/Parameters.kt
osipxd: This would probably allow us to keep `getOrFailImpl` signature untouched because of smart cast. But I'm not sure

nomisRev: Yes, I forgot about `null is R` in reified generics. Seems like a much better solution here.

e5l (approved):
lgtm

nomisRev:
@osipxd I updated the PR to move the nullability check in the actual implementation. This way we now also support nullable types for `getOrFail`.

`TypeInfo` is always being created because it's used by the underlying conversions mechanism. Perhaps the conversions mechanism can be optimised to avoid `TypeInfo` where possible but I left that for outside of this scope since it seems unrelated to my changes.

Could you review again? Thank you for the careful review and pointing it out! 🙏 We're still getting a better outcome thanks to it!

ktor-server/ktor-server-core/common/src/io/ktor/server/util/Parameters.kt
```diff
@@ -55,17 +53,20 @@ public inline fun Parameters.getOrFail(name: String): String {
...
 @PublishedApi
-internal fun <R : Any> Parameters.getOrFailImpl(name: String, typeInfo: TypeInfo): R {
-    val values = getAll(name) ?: throw MissingRequestParameterException(name)
-    return try {
-        @Suppress("UNCHECKED_CAST")
-        DefaultConversionService.fromValues(values, typeInfo) as R
-    } catch (cause: Exception) {
-        throw ParameterConversionException(name, typeInfo.type.simpleName ?: typeInfo.type.toString(), cause)
+internal fun <R> Parameters.getOrFailImpl(name: String, typeInfo: TypeInfo): R {
+    return if (typeInfo.kotlinType?.isMarkedNullable == true && get(property.name) == null) {
```
osipxd: Thank you for updating the PR!
Only one thing I've noticed. Compilation fails here: ‎`property` is unresolved in this scope (probably should be ‎`name`) 👀

osipxd (changes requested):

ktor-server/ktor-server-core/common/src/io/ktor/server/util/Parameters.kt
nomisRev: Yes, sorry about that! I was coding without loading the project 😅 Takes too much resources 💀

osipxd (approved):

PR #5497, opened 2026-03-30
Open Telemetry Support for KTOR

Subhanshu20135:
# Add native OpenTelemetry integration for server and client

## Summary

- Introduces `ktor-server-opentelemetry` and `ktor-client-opentelemetry` modules providing first-class OpenTelemetry distributed tracing, metrics, and context propagation - filling the gap alongside existing Micrometer and Dropwizard integrations.
- Server plugin creates SERVER spans for incoming requests with automatic route-aware naming, W3C Trace Context extraction, coroutine-safe context propagation, and HTTP semantic convention attributes/metrics.
- Client plugin creates CLIENT spans for outgoing requests with automatic trace context injection into headers, enabling end-to-end distributed traces when both plugins are used together.

## Motivation

Ktor has Micrometer and Dropwizard metrics plugins, but no native OpenTelemetry support. OpenTelemetry is the industry-standard, vendor-neutral observability framework for distributed tracing, metrics, and logging. Users currently must rely on the external `opentelemetry-java-instrumentation` agent or manually instrument their applications. A native plugin provides a better developer experience with Ktor-idiomatic configuration, proper coroutine context propagation, and route-aware span naming.

## Usage

```kotlin
// Server
install(OpenTelemetry) {
    openTelemetry = sdkOpenTelemetry
    captureRequestHeaders("X-Request-ID")
}

// Client
val client = HttpClient(CIO) {
    install(OpenTelemetry) {
        openTelemetry = sdkOpenTelemetry
    }
}
```

## Key design decisions

- **JVM-only modules** - follows the same pattern as `ktor-server-metrics-micrometer` since the OpenTelemetry Java SDK is JVM-only
- **Dual configuration style** - accepts either a full `io.opentelemetry.api.OpenTelemetry` instance or individual `TracerProvider`/`MeterProvider`/`TextMapPropagator` components
- **Coroutine context propagation** - the server plugin uses a `ThreadContextElement` to propagate the OTEL `Context` across coroutine dispatches, so `Context.current()` works correctly inside handlers and client spans automatically become children of server spans
- **OpenTelemetry semantic conventions** - attributes and metrics follow the stable HTTP semantic conventions (`http.request.method`, `url.path`, `http.route`, `http.server.request.duration`, etc.)
- **Existing patterns** - plugin structure, hooks, and lifecycle follow the established `MicrometerMetrics` plugin pattern (`Metrics` hook, `ResponseSent`, `CallFailed`, `RoutingRoot.RoutingCallStarted`)

## Modules added

| Module | Location | Convention plugin |
|---|---|---|
| `ktor-server-opentelemetry` | `ktor-server/ktor-server-plugins/ktor-server-opentelemetry/` | `ktorbuild.project.server-plugin` |
| `ktor-client-opentelemetry` | `ktor-client/ktor-client-plugins/ktor-client-opentelemetry/` | `ktorbuild.project.client-plugin` |

## Dependencies added

- `io.opentelemetry:opentelemetry-api:1.54.0` (API dependency for both modules)
- `io.opentelemetry:opentelemetry-sdk:1.54.0` (test only)
- `io.opentelemetry:opentelemetry-sdk-testing:1.54.0` (test only)

### DIFF
--- a/.idea/vcs.xml
+++ b/.idea/vcs.xml
@@ -24,6 +24,6 @@
     </option>
   </component>
   <component name="VcsDirectoryMappings">
-    <mapping directory="$PROJECT_DIR$" vcs="Git" />
+    <mapping directory="" vcs="Git" />
   </component>
 </project>
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -46,6 +46,7 @@ logback = "1.5.24"
 
 dropwizard = "4.2.37"
 micrometer = "1.16.2"
+opentelemetry = "1.46.0"
 
 jansi = "2.4.2"
 typesafe = "1.4.5"
@@ -210,6 +211,9 @@ okio = { module = "com.squareup.okio:okio", version.ref = "okio" }
 dropwizard-core = { module = "io.dropwizard.metrics:metrics-core", version.ref = "dropwizard" }
 dropwizard-jvm = { module = "io.dropwizard.metrics:metrics-jvm", version.ref = "dropwizard" }
 micrometer = { module = "io.micrometer:micrometer-core", version.ref = "micrometer" }
+opentelemetry-api = { module = "io.opentelemetry:opentelemetry-api", version.ref = "opentelemetry" }
+opentelemetry-sdk = { module = "io.opentelemetry:opentelemetry-sdk", version.ref = "opentelemetry" }
+opentelemetry-sdk-testing = { module = "io.opentelemetry:opentelemetry-sdk-testing", version.ref = "opentelemetry" }
 
 mustache = { module = "com.github.spullara.mustache.java:compiler", version.ref = "mustache" }
 freemarker = { module = "org.freemarker:freemarker", version.ref = "freemarker" }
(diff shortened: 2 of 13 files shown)


osipxd (changes requested):
Thanks a lot for working on this and iterating with the CodeRabbit – this is clearly a substantial and thoughtful piece of work!

Unfortunately right now I can’t move forward with merging this PR, because I’m in the middle of preparing a [KLIP](https://github.com/ktorio/ktor-klip/) for Pipeline Tracing that is intended to define the foundation and direction for first‑class tracing support in Ktor (including OpenTelemetry). Until that design is agreed on, we need to avoid introducing new tracing mechanisms that might conflict with the KLIP.
I’ll keep this PR and its design in mind while working on the KLIP.

Once the KLIP is published, there should be a clearer picture of how tracing integration should look, and we can revisit how this PR fit into that design. In the meantime, thank you again for the effort here!

osipxd:
Thanks again for your work on this 🙏 
Unfortunately I can't accept the PR. The reasoning is the same as in my comment on the health check PR: https://github.com/ktorio/ktor/pull/5498#issuecomment-2090195489.

PR #5498, opened 2026-03-30
Added Health Check Module for Ktor

Subhanshu20135:
# Add HealthCheck plugin for Kubernetes readiness and liveness probes

## Summary

- Adds a new `ktor-server-health-check` module providing a declarative DSL for configuring health check endpoints (`readiness` and `liveness`) with named, concurrent checks
- Returns structured JSON responses with per-check status and appropriate HTTP codes (200 UP / 503 DOWN), compatible with Kubernetes probes, load balancers, and any HTTP-based health polling
- Requires zero external dependencies -- no ContentNegotiation or serialization libraries needed

## Motivation

Every production Ktor application deployed to Kubernetes needs health and readiness endpoints. Currently, developers write these manually every time. Competing frameworks (Spring Actuator, Micronaut Health, Quarkus SmallRye Health) all ship this out of the box. This plugin makes it a one-liner:

```kotlin
install(HealthCheck) {
    readiness("/ready") {
        check("database") { dataSource.connection.isValid(1) }
        check("redis") { redisClient.ping(); true }
    }
    liveness("/health") {
        check("memory") { Runtime.getRuntime().freeMemory() > threshold }
    }
}
```

## Design

- **`ApplicationPlugin` with `onCall` interception** -- matches configured paths before routing, so endpoints work even without `Routing` installed
- **Concurrent check execution** -- all checks within an endpoint run via `coroutineScope { async { ... } }`, minimizing probe latency for I/O-bound checks
- **Lightweight JSON** -- responses built with `StringBuilder`, no serialization dependency
- **Multiplatform** -- implemented entirely in `commonMain`, works on JVM, Native, and JS targets

## Response Format

The plugin responds with `application/json`:

### All healthy (HTTP 200)

```json
{
  "status": "UP",
  "checks": [
    { "name": "database", "status": "UP" },
    { "name": "redis", "status": "UP" }
  ]
}
```

### One or more unhealthy (HTTP 503)

```json
{
  "status": "DOWN",
  "checks": [
    { "name": "database", "status": "UP" },
    { "name": "redis", "status": "DOWN", "error": "Connection refused" }
  ]
}
```

## Files Changed

| File | Purpose |
|---|---|
| `settings.gradle.kts` | Register `ktor-server-health-check` module |
| `ktor-server-health-check/build.gradle.kts` | Module build config using `ktorbuild.project.server-plugin` |
[...]

PR #5499, opened 2026-03-30
Added Circuit Breaker Module

Subhanshu20135:
# PR: Add Circuit Breaker client plugin

## Summary

- Adds a new `ktor-client-circuit-breaker` module implementing the [Circuit Breaker pattern](https://learn.microsoft.com/en-us/azure/architecture/patterns/circuit-breaker) for the Ktor HTTP client, filling the gap alongside `HttpRequestRetry` and `HttpTimeout` for microservice resilience.
- Implements a three-state machine (Closed / Open / Half-Open) with per-service configuration, automatic request routing, custom failure predicates, and thread-safe state transitions via coroutine `Mutex`.
- Includes 14 test cases covering all state transitions, recovery cycles, independent circuits, custom predicates, request routing, configuration validation, and exception semantics.

## Motivation

The Ktor client has retry and timeout support but no built-in circuit breaker. Without one, a single failing downstream service can exhaust connection pools and cascade failures across the call graph. This is a table-stakes pattern for microservice architectures and is present in competing frameworks (Spring Cloud Circuit Breaker, Resilience4j, Hystrix).

## API

```kotlin
val client = HttpClient(CIO) {
    install(CircuitBreaker) {
        register("payment-service") {
            failureThreshold = 5
            resetTimeout = 30.seconds
            halfOpenRequests = 3
        }
    }
}

client.get("https://payment.example.com/api/charge") {
    circuitBreaker("payment-service")
}
```

## Design decisions

- **`on(Send)` hook** -- Same interception point as `HttpRequestRetry`, giving the circuit breaker visibility into the final outcome after any inner retry logic.
- **`IllegalStateException`** -- `CircuitBreakerOpenException` extends `IllegalStateException` rather than `IOException` because a circuit breaker rejection is a deliberate policy decision, not an I/O error. This also avoids unintended retries by `HttpRequestRetry`'s default exception predicate.
- **`TimeSource` abstraction** -- An internal `timeSource` property defaults to `TimeSource.Monotonic` but can be replaced with `TestTimeSource` in tests, enabling deterministic time-based state transition tests without real delays.
- **Coroutine `Mutex`** -- Used for state transitions instead of atomics, since the state machine has multi-field invariants that need to be updated atomically.

## Files changed

| File | Change |
|---|---|
| `settings.gradle.kts` | Register `ktor-client-circuit-breaker` module |
| `.../ktor-client-circuit-breaker/build.gradle.kts` | Module build config |
| `.../circuitbreaker/CircuitBreaker.kt` | Plugin implementation (~365 lines) |
| `.../circuitbreaker/CircuitBreakerTest.kt` | 14 test cases |
| `.../api/ktor-client-circuit-breaker.api` | JVM ABI dump |
| `.../api/ktor-client-circuit-breaker.klib.api` | Klib ABI dump |
| `.../ktor-client-circuit-breaker/README.md` | Module documentation |

### DIFF
--- a/ktor-client/ktor-client-plugins/ktor-client-circuit-breaker/README.md
+++ b/ktor-client/ktor-client-plugins/ktor-client-circuit-breaker/README.md
@@ -142,5 +142,5 @@ try {
 
 ## Interaction with other plugins
 
-- **HttpRequestRetry** -- Install `HttpRequestRetry` *before* `CircuitBreaker` so the circuit breaker wraps the retry logic. This way the circuit sees the final outcome after all retries, and `CircuitBreakerOpenException` is not retried.
+- **HttpRequestRetry** -- Install `CircuitBreaker` *before* `HttpRequestRetry` so the circuit breaker wraps the retry logic. This way the circuit sees the final outcome after all retries, and `CircuitBreakerOpenException` is not retried.
 - **HttpTimeout** -- Timeout exceptions are counted as circuit breaker failures.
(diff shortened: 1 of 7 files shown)


osipxd:
Thanks for your work on this 🙏 
Unfortunately I can't accept the PR. The reasoning is the same as in my comment on the health check PR: https://github.com/ktorio/ktor/pull/5498#issuecomment-2090195489.

PR #5500, opened 2026-03-31
KTOR-9451 Support nested generic types

bjhham:
**Subsystem**
Server, OpenAPI (Schema)

**Motivation**
[KTOR-9451](https://youtrack.jetbrains.com/issue/KTOR-9451) OpenAPI schema inference not working for custom nested generics

**Solution**
We were previously only checking if the supplied `KType` was parameterized, but this logic doesn't work for multiple levels of generics.  To account for this, we need to initially traverse the type argument tree to populate a lookup of KTypes, because the generated serializers' descriptors have no way of introspecting on this.

Additionally, I noticed that generics were only supported for collection types on the reflection schema inference side, so I also fixed this.


osipxd (approved):
👍

PR #5505, opened 2026-04-04
KTOR-9463 Fix StackOverflow on self-referencing sealed types

bjhham:
**Subsystem**
Server, JSON schema inference

**Motivation**
[KTOR-9463](https://youtrack.jetbrains.com/issue/KTOR-9463) OpenAPI: schema inference StackOverflow

**Solution**
Include recursion guard for sealed type traversal as well as properties.

There was also some inconsistency in the schema references wrt qualified vs simplified names, which should be fixed now.


ktor-shared/ktor-openapi-schema/common/src/io/ktor/openapi/JsonSchemaInference.kt
```diff
@@ -294,6 +271,37 @@ public class KotlinxSerializerJsonSchemaInference(
...
+    @OptIn(InternalAPI::class)
+    private fun buildSchemaOrReference(
+        descriptor: SerialDescriptor,
+        title: String,
+        visiting: MutableSet<String>,
+        annotations: List<Annotation> = emptyList(),
+    ): ReferenceOr<JsonSchema> =
+        if (!descriptor.isContainerType() && !visiting.add(title)) {
+            ReferenceOr.schema(title)
+                .nonNullable(descriptor.isNullable)
```
osipxd: I've just noticed that `nonNullable(isNullable: Boolean)` is confusing

osipxd (approved):

ktor-shared/ktor-openapi-schema/common/src/io/ktor/openapi/JsonSchemaInference.kt
bjhham: Hmm good point.  Perhaps it should be `wrapIfNullable(isNullable: Boolean)` ?

osipxd: Yes, it's much better

PR #5511, opened 2026-04-07
#5033 Fix scoped Sessions plugin not sending Set-Cookie without call.respond()

### LINKED ISSUE #5033: Unable to update/remove session data if no response content
This is due to `BeforeSend` hook is not called when
1. No response content
2. Scoped plugin

Sample
```kotlin
    routing {
        route("test") {
            install(Sessions) { // not global
                cookie<String>(name = "token")
            }
            post("clear") {
                call.sessions.clear("token")
                if (call.request.queryParameters["c"] == "1") { // flag to set empty content
                    call.respond("")
                }
                call.response.status(HttpStatusCode.OK) // or HttpStatusCode.NoContent
            }
        }
    }
```

curl
```
> POST /test/clear?c=1 HTTP/1.1
> Host: 127.0.0.1:8080
> Accept: */*
> Cookie: token=foo
> 
< HTTP/1.1 200 OK
< Set-Cookie: token=; Max-Age=0; Expires=Thu, 01 Jan 1970 00:00:00 GMT; Path=/; HttpOnly; $x-enc=URI_ENCODING
< Content-Length: 0
< Content-Type: text/plain; charset=UTF-8
< 

> POST /test/clear HTTP/1.1
> Host: 127.0.0.1:8080
> Accept: */*
> Cookie: token=foo
> 
< HTTP/1.1 200 OK
< Content-Length: 0
< 
```

fru1tworld:
**Subsystem**
Server, ktor-server-core (routing)

**Motivation**
Fixes #5033. Scoped `Sessions` plugin doesn't send `Set-Cookie` when handler sets status without `call.respond()`.

**Solution**
Call `respond()` through `RoutingPipelineCall` when handler set status but didn't respond, so scoped sendPipeline interceptors fire.

### DIFF
--- a/ktor-server/ktor-server-core/common/src/io/ktor/server/routing/RoutingRoot.kt
+++ b/ktor-server/ktor-server-core/common/src/io/ktor/server/routing/RoutingRoot.kt
@@ -105,6 +105,9 @@ public class RoutingRoot(
         application.monitor.raise(RoutingCallStarted, routingCall)
         try {
             routingCallPipeline.execute(routingApplicationCall)
+            if (!routingApplicationCall.isHandled && routingApplicationCall.response.status() != null) {
+                routingApplicationCall.respond(routingApplicationCall.response.status()!!)
+            }
         } catch (cause: Throwable) {
             routingCall.attributes.put(routingCallKey, routingCall)
             throw cause
--- a/ktor-server/ktor-server-core/common/src/io/ktor/server/routing/RoutingRoot.kt
+++ b/ktor-server/ktor-server-core/common/src/io/ktor/server/routing/RoutingRoot.kt
@@ -105,8 +105,10 @@ public class RoutingRoot(
         application.monitor.raise(RoutingCallStarted, routingCall)
         try {
             routingCallPipeline.execute(routingApplicationCall)
-            if (!routingApplicationCall.isHandled && routingApplicationCall.response.status() != null) {
-                routingApplicationCall.respond(routingApplicationCall.response.status()!!)
+            if (!routingApplicationCall.isHandled) {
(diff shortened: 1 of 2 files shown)


ktor-server/ktor-server-core/common/src/io/ktor/server/routing/RoutingRoot.kt
```diff
@@ -103,6 +103,9 @@ public class RoutingRoot(
         application.monitor.raise(RoutingCallStarted, routingCall)
         try {
             routingCallPipeline.execute(routingApplicationCall)
+            if (!routingApplicationCall.isHandled && routingApplicationCall.response.status() != null) {
```
zibet27: Just a nitpick
I would use `let` instead of `!!` operator

```kotlin 
if (!routingApplicationCall.isHandled) {
   routingApplicationCall.response.status()?.let { status ->
      routingApplicationCall.respond(status)
   }
}
```

zibet27 (changes requested):
Nice!

ktor-server/ktor-server-core/common/src/io/ktor/server/routing/RoutingRoot.kt
fru1tworld: Good point, updated. Thanks for the review!

zibet27 (approved):
LGTM
@fru1tworld Thank you for the fix!

PS: I'll rebase the branch to `release/3.x`, so it can be included in the next patch release. Or you can do it by running the `switch-base-branch.sh` script in the project root

fru1tworld:
Rebased onto `release/3.x` using cherry-pick. Thanks for the review and the suggestion!

Issue #4720, opened 2025-03-03
HttpTimeout plugin does not work with test dispatchers


schielek:
I want to test my clients timeout behavior. To have a deterministic test which also runs as fast as possible I decided to use the kotlinx.coroutines test dispatcher which comes with a virtual clock. It seems that the ktor's HttpTimeoutPlugin does not respect the virtual clock of kotlinx coroutines. Is there an alternative way to test this or should the HttpTimeoutPlugin respect the timing behavior of kotlinx.coroutines?

```kotlin
runTest {
    val mutex = Mutex(locked = true)

    val mockEngine = MockEngine {
        mutex.withLock {
            respond("OK")
        }
    }

    val client = HttpClient(mockEngine) {
        install(HttpTimeout) {
            requestTimeoutMillis = 100
        }
    }

    launch {
        delay(200)
        mutex.unlock()
    }
    
    // Assertion error since no exception is thrown
    // Test is green when no coroutine test scope is used
    assertThrows<HttpRequestTimeoutException> {
        client.get("/")
    }
}
```

Stexxe:
Thank you for the report. I've reproduced the issue (https://youtrack.jetbrains.com/issue/KTOR-8271).

PR #5512, opened 2026-04-07
Fix HttpTimeout not respecting test dispatchers in runTest

fru1tworld:
**Subsystem**
Client

**Motivation**
https://github.com/ktorio/ktor/issues/4720

**Solution**
Capture the caller's coroutine context before launching the timeout and use it when creating the `CoroutineScope` for `applyRequestTimeout`, so that test dispatchers (virtual time) are respected.


bjhham (approved):
Thank you for the fix 🙏

PR #5515, opened 2026-04-09
Client Form DSL. Remove incorrect contracts usage

zibet27:
**Subsystem**
Client Form DSL

**Motivation**
Reported by users


zibet27:
I guess we don't need a dedicated YT issue for that

osipxd (approved):
Would it make sense to use some different contract instead? But I'm okay with removing it

zibet27:
> Would it make sense to use some different contract instead? But I'm okay with removing it

Afaiu, the `bodyBuilder` is captured by the InputProvider, which can call it zero or multiple times

PR #5520, opened 2026-04-12
Improve crypto performance

solonovamax:
**Subsystem**
ktor-utils

**Motivation**
`generateNonce()` (on the JVM platform) was using an array list and did a bunch of unnecessary allocations and that annoyed me.

Fixes [KTOR-9488](https://youtrack.jetbrains.com/issue/KTOR-9488)

**Solution**
I rewrote `generateNonce()` in such a way that doesn't use a list to decrease the number of allocations. also it doesn't do unnecessary things like clearing it or creating a copy of it just to shuffle the elements.

I've also done two other changes:
- deprecate `generateNonce()` in favour of `generateNonceSuspend()` (if I could I'd name this `generateNonce()`) and `generateNonceBlocking()`
- mark the `hex()` functions as deprecated. the kotlin stdlib now has functions for converting to/from hex, and those will be much more performance because they don't create a bunch of unnecessary strings.

both of those changes were done in separate commits so that I can drop them if needed.

### DIFF
--- a/ktor-utils/common/src/io/ktor/util/Crypto.kt
+++ b/ktor-utils/common/src/io/ktor/util/Crypto.kt
@@ -19,6 +19,10 @@ private val digits = "0123456789abcdef".toCharArray()
  *
  * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.util.hex)
  */
+@Deprecated(
+    "prefer kotlin.text.toHexString",
+    replaceWith = ReplaceWith("bytes.toHexString()", imports = ["kotlin.text.toHexString"]),
+)
 public fun hex(bytes: ByteArray): String {
     val result = CharArray(bytes.size * 2)
     var resultIndex = 0
@@ -38,6 +42,10 @@ public fun hex(bytes: ByteArray): String {
  *
  * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.util.hex)
  */
+@Deprecated(
+    "prefer kotlin.text.hexToByteArray",
+    replaceWith = ReplaceWith("s.hexToByteArray()", imports = ["kotlin.text.hexToByteArray"]),
+)
 public fun hex(s: String): ByteArray {
     val result = ByteArray(s.length / 2)
     for (idx in result.indices) {
--- a/ktor-utils/common/src/io/ktor/util/Crypto.kt
+++ b/ktor-utils/common/src/io/ktor/util/Crypto.kt
@@ -14,8 +14,6 @@ import kotlinx.io.*
 
 private val digits = "0123456789abcdef".toCharArray()
 
-internal const val NONCE_SIZE_IN_BYTES = 16
-
 /**
  * Encode [bytes] as a HEX string with no spaces, newlines and `0x` prefixes.
  *
@@ -53,20 +51,42 @@ public fun hex(s: String): ByteArray {
 }
 
 /**
- * Generates a nonce string. Could block if the system's entropy source is empty
+ * Generates a nonce string 32 characters long. Could block if the system's entropy source is empty
  *
  * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.util.generateNonce)
  */
-public expect fun generateNonce(): String
+@Deprecated(
+    "Use generateNonceBlocking in blocking contexts and generateNonceSuspend in non-blocking contexts",
+    replaceWith = ReplaceWith("generateNonceBlocking()")
+)
+public fun generateNonce(): String = generateNonceBlocking(32)
+
+/**
+ * Generates a nonce string [length] characters long. Could suspend if the system's entropy source is empty.
+ *
+ * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.util.generateNonce)
+ */
+public expect suspend fun generateNonceSuspend(length: Int = 32): String
+
+/**
+ * Generates a nonce string [length] characters long. Could block if the system's entropy source is empty.
+ *
+ * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.util.generateNonce)
+ */
+public expect fun generateNonceBlocking(length: Int = 32): String
 
 /**
  * Generates a nonce bytes of [size]. Could block if the system's entropy source is empty
  *
  * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.util.generateNonce)
  */
+@Deprecated(
+    "Use generateNonceBlocking in blocking contexts and generateNonceSuspend in non-blocking contexts",
+    replaceWith = ReplaceWith("generateNonceBlocking(size).toByteArray()", "io.ktor.utils.io.core.toByteArray")
+)
 public fun generateNonce(size: Int): ByteArray = buildPacket {
     while (this.size < size) {
-        writeText(generateNonce())
+        writeText(generateNonceBlocking())
     }
 }.readByteArray(size)
 
--- a/ktor-utils/common/src/io/ktor/util/Crypto.kt
+++ b/ktor-utils/common/src/io/ktor/util/Crypto.kt
@@ -14,6 +14,10 @@ import kotlinx.io.*
 
 private val digits = "0123456789abcdef".toCharArray()
 
+internal const val NONCE_SIZE_IN_BYTES = 16
+
+internal const val NONCE_SIZE_IN_CHARS = NONCE_SIZE_IN_BYTES * 2
+
 /**
  * Encode [bytes] as a HEX string with no spaces, newlines and `0x` prefixes.
  *
@@ -67,21 +71,21 @@ public fun hex(s: String): ByteArray {
     "Use generateNonceBlocking in blocking contexts and generateNonceSuspend in non-blocking contexts",
     replaceWith = ReplaceWith("generateNonceBlocking()")
 )
-public fun generateNonce(): String = generateNonceBlocking(32)
+public fun generateNonce(): String = generateNonceBlocking(NONCE_SIZE_IN_CHARS)
 
 /**
  * Generates a nonce string [length] characters long. Could suspend if the system's entropy source is empty.
  *
  * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.util.generateNonce)
  */
-public expect suspend fun generateNonceSuspend(length: Int = 32): String
+public expect suspend fun generateNonceSuspend(length: Int = NONCE_SIZE_IN_CHARS): String
 
 /**
  * Generates a nonce string [length] characters long. Could block if the system's entropy source is empty.
  *
  * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.util.generateNonce)
  */
-public expect fun generateNonceBlocking(length: Int = 32): String
+public expect fun generateNonceBlocking(length: Int = NONCE_SIZE_IN_CHARS): String
 
 /**
  * Generates a nonce bytes of [size]. Could block if the system's entropy source is empty
--- a/ktor-utils/jvm/src/io/ktor/util/CryptoJvm.kt
+++ b/ktor-utils/jvm/src/io/ktor/util/CryptoJvm.kt
@@ -71,7 +71,11 @@ public actual suspend fun generateNonceSuspend(length: Int): String {
         return nonce.substring(0, length)
     }
 
-    return generateNonceLoop(nonce, length)
+    return if (length <= NONCE_SIZE_IN_CHARS) {
+        seedChannel.receive().substring(0, length)
+    } else {
+        generateNonceLong(nonce, length)
+    }
 }
 
 /**
@@ -88,10 +92,16 @@ public actual fun generateNonceBlocking(length: Int): String {
 
     ensureNonceGeneratorRunning()
 
-    return runBlocking { generateNonceLoop(nonce, length) }
+    return runBlocking {
+        if (length <= NONCE_SIZE_IN_CHARS) {
+            seedChannel.receive().substring(0, length)
+        } else {
+            generateNonceLong(nonce, length)
+        }
+    }
 }
 
-private suspend fun generateNonceLoop(initial: String?, length: Int) = buildString(length) {
+private suspend fun generateNonceLong(initial: CharSequence?, length: Int) = buildString(length) {
     if (initial != null) {
         append(initial)
     }
--- a/ktor-utils/jvm/src/io/ktor/util/CryptoJvm.kt
+++ b/ktor-utils/jvm/src/io/ktor/util/CryptoJvm.kt
@@ -64,14 +64,26 @@ private value class DigestImpl(val delegate: MessageDigest) : Digest {
  */
 public actual fun generateNonce(): String {
     val nonce = seedChannel.tryReceive().getOrNull()
-    if (nonce != null) return nonce
 
-    return generateNonceBlocking()
+    if (nonce != null && nonce.length >= 32) {
+        return nonce.substring(0, 32)
+    }
+
+    return generateNonceBlocking(nonce, 32)
 }
 
-private fun generateNonceBlocking(): String {
+private fun generateNonceBlocking(initial: String?, length: Int): String {
     ensureNonceGeneratorRunning()
     return runBlocking {
-        seedChannel.receive()
+        buildString(length) {
+            if (initial != null) {
+                append(initial)
+            }
+
+            while (length > this.length) {
+                val toAppend = seedChannel.receive()
+                append(toAppend, 0, (length - this.length).coerceAtMost(toAppend.length))
+            }
+        }.substring(0, length)
     }
 }
--- a/ktor-utils/jvm/src/io/ktor/util/CryptoJvm.kt
+++ b/ktor-utils/jvm/src/io/ktor/util/CryptoJvm.kt
@@ -65,14 +65,14 @@ private value class DigestImpl(val delegate: MessageDigest) : Digest {
 public actual suspend fun generateNonceSuspend(length: Int): String {
     ensureNonceGeneratorRunning()
 
-    val nonce = seedChannel.receive()
+    val nonce = nonceChannel.receive()
 
     if (nonce.length >= length) {
         return nonce.substring(0, length)
     }
 
     return if (length <= NONCE_SIZE_IN_CHARS) {
-        seedChannel.receive().substring(0, length)
+        nonceChannel.receive().substring(0, length)
     } else {
         generateNonceLong(nonce, length)
     }
@@ -84,7 +84,7 @@ public actual suspend fun generateNonceSuspend(length: Int): String {
  * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.util.generateNonceBlocking)
  */
 public actual fun generateNonceBlocking(length: Int): String {
-    val nonce = seedChannel.tryReceive().getOrNull()
+    val nonce = nonceChannel.tryReceive().getOrNull()
 
     if (nonce != null && nonce.length >= length) {
         return nonce.substring(0, length)
@@ -94,7 +94,7 @@ public actual fun generateNonceBlocking(length: Int): String {
 
     return runBlocking {
         if (length <= NONCE_SIZE_IN_CHARS) {
-            seedChannel.receive().substring(0, length)
+            nonceChannel.receive().substring(0, length)
         } else {
             generateNonceLong(nonce, length)
         }
@@ -107,7 +107,7 @@ private suspend fun generateNonceLong(initial: CharSequence?, length: Int) = bui
     }
 
     while (length > this.length) {
-        val toAppend = seedChannel.receive()
+        val toAppend = nonceChannel.receive()
         append(toAppend, 0, (length - this.length).coerceAtMost(toAppend.length))
     }
 }.substring(0, length)
--- a/ktor-utils/jvm/src/io/ktor/util/CryptoJvm.kt
+++ b/ktor-utils/jvm/src/io/ktor/util/CryptoJvm.kt
@@ -58,32 +58,46 @@ private value class DigestImpl(val delegate: MessageDigest) : Digest {
 }
 
 /**
- * Generates a nonce string 16 characters long. Could block if the system's entropy source is empty
+ * Generates a nonce string [length] characters long. Could block if the system's entropy source is empty.
  *
- * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.util.generateNonce)
+ * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.util.generateNonceSuspend)
  */
-public actual fun generateNonce(): String {
+public actual suspend fun generateNonceSuspend(length: Int): String {
+    ensureNonceGeneratorRunning()
+
+    val nonce = seedChannel.receive()
+
+    if (nonce.length >= length) {
+        return nonce.substring(0, length)
+    }
+
+    return generateNonceLoop(nonce, length)
+}
+
+/**
+ * Generates a nonce string [length] characters long. Could block if the system's entropy source is empty.
+ *
+ * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.util.generateNonceBlocking)
+ */
+public actual fun generateNonceBlocking(length: Int): String {
     val nonce = seedChannel.tryReceive().getOrNull()
 
-    if (nonce != null && nonce.length >= 32) {
-        return nonce.substring(0, 32)
+    if (nonce != null && nonce.length >= length) {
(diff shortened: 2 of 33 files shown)


ktor-utils/web/src/io/ktor/util/Crypto.web.kt
```diff
@@ -11,12 +11,19 @@ import kotlin.js.*
...
+/**
+ * Generates a nonce string.
+ *
+ * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.util.generateNonceBlocking)
+ */
+public actual fun generateNonceBlocking(length: Int): String {
+    val buffer = ByteArray(length / 2 + 1).toJsArray()
     _crypto.getRandomValues(buffer)
-    return hex(buffer.toByteArray())
+    return buffer.toByteArray().toHexString().substring(0, length)
```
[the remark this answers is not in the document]
solonovamax: if you request a nonce that's more than 131k characters, then that's your problem.

bjhham:
Thanks for the fix!  Unfortunately, our allocations / throughput tests don't use SSH so we don't have any indication for the improvements here.  Could you provide some profiler screenshots or similar so we can get a sense of the tangible benefits in allocation savings?

bjhham (commented):
Could you replace references in the project with `generateNonceBlocking() / generateNonceSuspend()` where relevant as well?

Btw I've added a YT ticket for tracking [KTOR-9488](https://youtrack.jetbrains.com/issue/KTOR-9488)

solonovamax:
> Thanks for the fix!  Unfortunately, our allocations / throughput tests don't use SSH so we don't have any indication for the improvements here.  Could you provide some profiler screenshots or similar so we can get a sense of the tangible benefits in allocation savings?

never actually did any performance testing lol

but, there is basically zero world in which it's slower than what it did previously
the core logic is the same, I just reduced the number of allocations

but if you really want, I can pull up JMH for this or something

solonovamax:
> Could you replace references in the project with `generateNonceBlocking() / generateNonceSuspend()` where relevant as well?

sure, I can do that
pretty sure I already got all of them, but I'll double check
when adding the feature I didn't do a full import of ktor, I commented out a bunch of modules in the `settings.gradle.kts` because of the insanely long import time

> Btw I've added a YT ticket for tracking [KTOR-9488](https://youtrack.jetbrains.com/issue/KTOR-9488)

updated the PR body to reflect that

bjhham:
> when adding the feature I didn't do a full import of ktor, I commented out a bunch of modules in the settings.gradle.kts because of the insanely long import time

bjhham:
> but if you really want, I can pull up JMH for this or something

That's alright, I'll just run the tests with the Intellij profiler for results in the ticket.

ktor-utils/common/src/io/ktor/util/Crypto.kt
```diff
@@ -53,20 +59,42 @@ public fun hex(s: String): ByteArray {
 }
```
bjhham, line 63: Turns out the std lib implementation is the same, though slightly less efficient because of the extra options.  Could you revert the deprecation here?

ktor-utils/jvm/src/io/ktor/util/Nonce.kt
```diff
@@ -23,20 +24,32 @@ private const val SECURE_NONCE_COUNT = 8
 
 private const val INSECURE_NONCE_COUNT_FACTOR = 4
 
+private const val NONCE_SIZE_IN_BYTES = 16
+
+private const val RANDOM_BYTES_CHUNK_SIZE = 16
+
 internal val seedChannel: Channel<String> = Channel(1024)
```
bjhham: Since we're dealing with many substrings here, could you try using `CharSequence` and `subSequence` calls to limit copying.

ktor-utils/jvm/src/io/ktor/util/CryptoJvm.kt
```diff
@@ -58,20 +58,46 @@ private value class DigestImpl(val delegate: MessageDigest) : Digest {
...
+    if (nonce != null && nonce.length >= length) {
+        return nonce.substring(0, length)
     }
+
+    ensureNonceGeneratorRunning()
+
+    return runBlocking { generateNonceLoop(nonce, length) }
 }
+
+private suspend fun generateNonceLoop(initial: String?, length: Int) = buildString(length) {
```
bjhham: It seems the introduction of this loop increases memory allocations.

bjhham (changes requested):
After running the profiler in Intellij, I found that these changes actually increase memory allocations in both cases.

1. The std library `toHexString()` function uses the same code as in Ktor but with a few extra options and branches, increasing allocations slightly.
2. The introduction of variable length substrings and concatenations for the nonce generation seems to have increased memory allocations by about 40-50% when running the standard blocking calls found in `NonceSmokeTest`.

I would advise reverting the `hex()` function deprecation and running `NonceSmokeTest` with the profiler to resolve the increased memory allocation spots.

ktor-utils/common/src/io/ktor/util/Crypto.kt
solonovamax: at the very least, it was much better than the method added for decoding (because it called `.toString().toInt(16)` everywhere rather than using a lookup table like the stdlib implementation)
regardless, imo it should still be deprecated and everyone should instead use the one in the stdlib

I'll take a look at it later when I have a chance

ktor-utils/jvm/src/io/ktor/util/Nonce.kt
solonovamax: sure, I can change it to do that

ktor-utils/jvm/src/io/ktor/util/CryptoJvm.kt
solonovamax: it shouldn't do any additional allocations because the string builder has its size defined up-front and no more than that size is ever added to it
the only allocation it's doing there is for the string builder itself

solonovamax: I think I realized why it's generating a ton of allocations. but this will specifically only ever happen because of the test case:
because it's calling `tryReceive()` in the blocking function and it's requesting a lot of nonces in a hot loop, the majority of the time `tryReceive` will end up returning `null`, which causes it to go and allocate an entire `StringBuilder`

ktor-utils/jvm/src/io/ktor/util/Nonce.kt
solonovamax: `subSequence` just delegates straight to `substring`, at least on OpenJDK 21, so I see no need to change this.

ktor-utils/common/src/io/ktor/util/Crypto.kt
solonovamax: when running in a profiler, the stdlib function seems to allocate *slightly* less than the one from ktor.
probably not enough to make that much of a difference, though.

I have not done any benchmarks for performance with JMH

ktor-utils/jvm/src/io/ktor/util/Nonce.kt
bjhham: Ah, I forgot they changed the implementation a while back.  You could still avoid copying memory here by using a `ByteBuffer` here with slices instead of String / substring.

ktor-utils/jvm/src/io/ktor/util/CryptoJvm.kt
bjhham: Looks like it's not entirely due to the test case here.

There's a few minor issues:
1. The nonce size doesn't match the default expected, so it will always go into the slow path (16 x 2 = 32).
2. The seed channel is returning empty strings that cause the concatenation loop to spin a little.
3. There's some unnecessary substrings in here.

Looks like you can just change the default nonce size, avoid the empty strings in the channel, and add some branches to avoid the slow path.

Something like this in the blocking calls to avoid the slow path and substrings:
```kotlin
public actual fun generateNonceBlocking(length: Int): String {
    val nonce = seedChannel.tryReceive().getOrNull()

    if (nonce != null && nonce.length >= length) {
        if (nonce.length == length) return nonce
        return nonce.substring(0, length)
    }

    ensureNonceGeneratorRunning()

    return runBlocking {
        generateNonceLoop(nonce ?: seedChannel.receive(), length)
    }
}

private suspend fun generateNonceLoop(initial: String, length: Int) =
    when {
        initial.length == length -> initial
        initial.length > length -> initial.substring(0, length)
        else -> buildString(length) {
            while (length > this.length) {
                val toAppend = seedChannel.receive()
                append(toAppend, 0, (length - this.length).coerceAtMost(toAppend.length))
            }
        }
    }
```

ktor-utils/jvm/src/io/ktor/util/Nonce.kt
solonovamax: I considered using something like that, but by using a `CharArray` instead

the problem is that it would add a whole bunch of complexity to everything as those char arrays can't be reused (or else that's a race condition waiting to happen), and I'd need to add my own to/from hex functions

I think this is very much a "future work" type of thing

ktor-utils/jvm/src/io/ktor/util/CryptoJvm.kt
solonovamax: > Looks like it's not entirely due to the test case here.
> 
> There's a few minor issues:
> 
> 1. The nonce size doesn't match the default expected, so it will always go into the slow path (16 x 2 = 32).

I had realized that, and had a pending comment mentioning that which I ended up not sending.

the issue here is that on all the other platforms, the nonce would be 32 characters (because they would genetate 16 random bytes which when converted to hex would be 32 characters), but on the jvm platform it was 16 characters. so, I used as the default requested length to ensure parity, but when I was originally writing the majority of the jvm code (which was quite a few months back, I just had it sitting around), I was not aware of this fact.

however, even if the sizing did match it could still allocate a string builder when the channel is empty, which would result in increased allocations

 have a patch for this, I'm just also analyzing a couple of other places I can make improvements

> 2. The seed channel is returning empty strings that cause the concatenation loop to spin a little.

that shouldn't have too much overhead, and will only happen at the start but will be quickly corrected for.

I might add a length check, however (or I might also make the array in the nonce generator have its elements be nullable or smth)

though, if I add a length check then that will always result in iterating over the entire length of the string (I believe. actually, I'm not 100% sure it would do this, I forget if it stores the length up front or if it's only cached after the first time it's accessed. I'd have to double check. I'm writing this out on my phone right now)

> 3. There's some unnecessary substrings in here.

if you can `substring` where the start is `0` and the length is the same as the string's length, then it will just return `this`, so there should be zero allocations. I left those in as a form of safety.

> Looks like you can just change the default nonce size, avoid the empty strings in the channel, and add some branches to avoid the slow path.

I'm instead going to change the size that the nonces are split into in the generator coroutine

> Something like this in the blocking calls to avoid the slow path and substrings:

in your code, for the slow path it discards the first value it receives

I do also have an implementation somewhat similar to this, except the hot & slow paths have their own dedicated functions

ktor-utils/jvm/src/io/ktor/util/CryptoJvm.kt
```diff
@@ -58,20 +58,56 @@ private value class DigestImpl(val delegate: MessageDigest) : Digest {
...
+        return nonce.substring(0, length)
+    }
 
-    return generateNonceBlocking()
+    return if (length <= NONCE_SIZE_IN_CHARS) {
+        seedChannel.receive().substring(0, length)
+    } else {
+        generateNonceLong(nonce, length)
+    }
 }
```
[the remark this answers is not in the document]
solonovamax: > Additionally, there's a logic issue: when `nonce.length < length` but `length <= NONCE_SIZE_IN_CHARS`, the code discards the first received nonce (line 68) and fetches a new one (line 75). This wastes entropy.

can literally never happen.

ktor-utils/jvm/src/io/ktor/util/CryptoJvm.kt
```diff
@@ -58,20 +58,56 @@ private value class DigestImpl(val delegate: MessageDigest) : Digest {
...
+
     return runBlocking {
-        seedChannel.receive()
+        if (length <= NONCE_SIZE_IN_CHARS) {
+            seedChannel.receive().substring(0, length)
+        } else {
+            generateNonceLong(nonce, length)
+        }
     }
 }
```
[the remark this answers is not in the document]
solonovamax: > When `nonce` is null or too short, and `length <= NONCE_SIZE_IN_CHARS`, the code at line 97 fetches a new nonce, potentially discarding the one from `tryReceive()` if it was non-null but too short

can literally never happen.

ktor-utils/common/src/io/ktor/util/Crypto.kt
```diff
@@ -53,20 +63,42 @@ public fun hex(s: String): ByteArray {
...
 
 /**
  * Generates a nonce bytes of [size]. Could block if the system's entropy source is empty
  *
  * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.util.generateNonce)
  */
+@Deprecated(
+    "Use generateNonceBlocking in blocking contexts and generateNonceSuspend in non-blocking contexts",
+    replaceWith = ReplaceWith("generateNonceBlocking(size).toByteArray()", "io.ktor.utils.io.core.toByteArray")
+)
```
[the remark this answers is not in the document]
solonovamax: > One concern: `generateNonceBlocking(size)` returns a `String` of `size` characters. Converting that to a `ByteArray` via `toByteArray()` produces a byte array of `size` bytes (assuming ASCII hex chars). But the original `generateNonce(size: Int)` was designed to return `size` random bytes. Since the new nonce functions return hex strings, `size` characters of hex only represent `size/2` bytes of entropy. This is a semantic change.

no it literally doesn't.

solonovamax:
some notes:
- it seems that `SecureRandom` can allocate a bunch of byte arrays. for the weak random, what are your thoughts on possibly using something like `L128X1024MixRandom`, `L64X1024MixRandom`, or `Xoshiro256PlusPlus` for the weak random? I know they're not cryptographically secure, but I'm not sure if the SHA 1 PRNG is still considered secure either. these are only accessible for newer JDKs, so there'd need to be two implementations: one for newer JDKs and one for the older JDKs
- I've noticed that in the old code, there's a bug where all the nonces it was producing where 16 characters long on the JVM but 32 characters long on all other platforms. I changed it to generate 32 character long nonces on the JVM platform as well, which can cause it to *look* like it's allocating more memory. keep that in mind when profiling.
- `generateSessionId()` returns `generateNonce() + generateNonce()`. on the JVM this is equivalent to `generateNonceBlocking(32)`, however on other platforms it's equivalent to `generateNonceBlocking(64)`. I've gone with the latter.
- I've replaced the deprecated `generateNonce()` and `hex()` functions, however made the `hex()` functions in a separate commit so it can be dropped. I have not run any of the tests for the replaced hex functions (because tbh I don't want to wait 2 years), but the behaviour SHOULD be identical.

bjhham (changes requested):
Thanks for the updates!

> it seems that SecureRandom can allocate a bunch of byte arrays. for the weak random, what are your thoughts on possibly using something like L128X1024MixRandom, L64X1024MixRandom, or Xoshiro256PlusPlus for the weak random? I know they're not cryptographically secure, but I'm not sure if the SHA 1 PRNG is still considered secure either. these are only accessible for newer JDKs, so there'd need to be two implementations: one for newer JDKs and one for the older JDKs

From what I understand, we should just be calling `SecureRandom()` here and deferring to the java runtime to select the best algorithm.  I think relying on SHA1 PRNG might be a legacy approach to ensure it worked on Java < 8 platforms when these were valid targets.  If there is some memory overhead for selecting a stronger provider, it shouldn't be a problem since it's only instantiated once.  Could you replace this with `SecureRandom()`?

As for the tests, it looks like there's just one failure in `io.ktor.server.sessions.SessionTestJvm` due to the hex string in `testSessionByValueMac` being an invalid length.  I can't think of an advantage for preserving this "leniency" in the conversion function behaviour, so I would say we can just add a digit to the string and go forward with the preference of `hexToByteArray()` over `hex()`.

solonovamax:
> From what I understand, we should just be calling `SecureRandom()` here and deferring to the java runtime to select the best algorithm

yes, that is true. however, I did keep the previous implementation rather than change it, as I largely didn't see *much* of a reason too.
although, it would probably be better to first try `SecureRandom.getInstanceStrong()` for a strong instance. (I did add code to fall back to this if none of the predefined algorithms can be found, and afterwards just falling back to `SecureRandom()` like before.

> I think relying on SHA1 PRNG might be a legacy approach to ensure it worked on Java < 8 platforms when these were valid targets. If there is some memory overhead for selecting a stronger provider, it shouldn't be a problem since it's only instantiated once. Could you replace this with `SecureRandom()`?

here, there are two secure random instances:
- the "strong" random, which tries (in order):
  1. checks the `io.ktor.random.secure.random.provider` system property
  2. entries from the predefined list
  3. `SecureRandom.getInstanceStrong()`
  4. `SecureRandom()`
- the "weak random", which uses `SHA1PRNG` (though, important to note that `SHA1PRNG` is *not* in the JVM spec, so it is not required to be present in all JVMs. there is no fallback mechanism for if it does not exist)

it then uses the strong random to generate 1/4th of the random bytes and the strong random to generate the remaining 3/4ths.
the random bytes are then replace the upper half of the nonces list.
it then shuffles the list of nonces using the weak random, and sends the upper half of them to the channel. the remaining half are re-used for the next round.
it also periodically reseeds the weak random with the strong random. (every 30 seconds)

I assume the use of the weaker random is because the weaker random is supposed to be faster than the stronger random, so it can be used for the majority of the random number generation.

tbh, I'm not *too* sure how strong (or beneficial) this whole scheme is, but I kept it because I didn't want to go through the effort of looking into that & testing out other things
a lot of it kinda feels like "security through obscurity" tbh

also, I just realized that there is a *much* simpler way to calculate the size of the `randomNonces` array. I cannot remember why I calculated it the way I did, but I just realized that I can instead do `SECURE_NONCE_COUNT * 2` rather than `SECURE_NONCE_COUNT * NONCE_SIZE_IN_BYTES * INSECURE_NONCE_COUNT_FACTOR * 2 / NONCE_SIZE_IN_CHARS * 2`

also also, tbh it might be useful to increase the size of the nonce array from 16 to maybe 64 (so 32 are sent at once instead of 8) and also decrease the size of the seed channel from 1024 to something like maybe 128 to decrease the number of strings kept in memory at once. (and maybe add a system property to change it. also possibly a system property for the number of nonces stored in the array?)

> As for the tests, it looks like there's just one failure in `io.ktor.server.sessions.SessionTestJvm` due to the hex string in `testSessionByValueMac` being an invalid length. I can't think of an advantage for preserving this "leniency" in the conversion function behaviour, so I would say we can just add a digit to the string and go forward with the preference of `hexToByteArray()` over `hex()`.

interesting that it doesn't like when the string is an odd length.
this doesn't seem to be documented anywhere, and is likely an issue in upstream. I'll open an issue later when I have more time to take a look at it.
but for now, I'll modify the test and add a leading zero.

also, interesting that it hit the slow path there. quickly reading the stdlib code, I assume it hit the fast path for all the other digits but was only in the slow path for that last digit?

ktor-utils/jvm/src/io/ktor/util/Nonce.kt
```diff
@@ -28,15 +29,21 @@ internal val seedChannel: Channel<String> = Channel(1024)
...
 ) {
     val seedChannel = seedChannel
     var lastReseed = 0L
-    val previousRoundNonceList = ArrayList<String>()
+    // empty strings are fine, because we only send non-empty strings
+    val randomNonces: Array<String> = Array(
+        SECURE_NONCE_COUNT * NONCE_SIZE_IN_BYTES * INSECURE_NONCE_COUNT_FACTOR * 2 / NONCE_SIZE_IN_CHARS * 2
+    ) { "" }
+
     val secureInstance = lookupSecureRandom()
```
bjhham: nit: `strongRandom` would be a better variable name here

ktor-utils/jvm/src/io/ktor/util/Nonce.kt
```diff
@@ -28,15 +29,21 @@ internal val seedChannel: Channel<String> = Channel(1024)
...
     context = Dispatchers.Default + NonCancellable + NonceGeneratorCoroutineName,
     start = CoroutineStart.LAZY
 ) {
     val seedChannel = seedChannel
     var lastReseed = 0L
-    val previousRoundNonceList = ArrayList<String>()
+    // empty strings are fine, because we only send non-empty strings
+    val randomNonces: Array<String> = Array(
+        SECURE_NONCE_COUNT * NONCE_SIZE_IN_BYTES * INSECURE_NONCE_COUNT_FACTOR * 2 / NONCE_SIZE_IN_CHARS * 2
+    ) { "" }
```
bjhham: Since we're in the weeds here, might as well use `arrayOfNulls(length)` here for the `O(1)` init, then you just need to change the send loop body to `val nonce = randomNonces[index] ?: continue; seedChannel.send(nonce)`.

bjhham:
Thanks for looking into the random generators.  This part all looks good to me.

> also also, tbh it might be useful to increase the size of the nonce array from 16 to maybe 64 (so 32 are sent at once instead of 8) and also decrease the size of the seed channel from 1024 to something like maybe 128 to decrease the number of strings kept in memory at once. (and maybe add a system property to change it. also possibly a system property for the number of nonces stored in the array?)

Yeah, these sound like reasonable changes.  1024 does seem unnecessarily large given the use cases where nonces are needed.

> interesting that it doesn't like when the string is an odd length.
> this doesn't seem to be documented anywhere, and is likely an issue in upstream. I'll open an issue later when I have more time to take a look at it.
> but for now, I'll modify the test and add a leading zero.

> also, interesting that it hit the slow path there. quickly reading the stdlib code, I assume it hit the fast path for all the other digits but was only in the slow path for that last digit?

Looks like it always processes the bytes in pairs in the fast path, then falls back to slow when odd, then fails.

```kotlin
// in the beautifully named "hexToByteArrayShortByteSeparatorNoPrefixAndSuffix" function
if (byteSeparatorLength == 0) {
    if (numberOfChars and 1 != 0) return null
    val numberOfBytes = numberOfChars shr 1
    val byteArray = ByteArray(numberOfBytes)
    for (byteIndex in 0 until numberOfBytes) {
        byteArray[byteIndex] = parseByteAt(charIndex)
        charIndex += 2
    }
    return byteArray
}
```

bjhham (changes requested):
Changes look good here, the failing test will still need correction 👍

solonovamax:
I've been looking at it a bit more, and atm the thing which produces the most allocations is the call to `randomNonces.shuffle(weakKotlinRandom)`.
in my test runs, it produces 1.04 GB of allocations, whereas:
- `Random.nextBytes()` produces 103.28 MB + 266.34 MB of allocations
- `ByteArray.toHexString()` produces 776.47 MB of allocations
- `String.substring()` produces 947.91 MB of allocations
- `Channel.send()` produces 171.97 MB of allocations

tbh I'm not entirely sure why this way of doing it was chosen.
- why do nonces need to be preserved across iterations & shuffled? does this make it more secure? I wouldn't think so.
- why are we using two randoms? is it actually more performant to use a strong & weak random?
- if using two randoms is more performant, why are we mixing in secure & weak bytes? does this actually increase security? we're already reseeding the weak random from the strong random every iteration anyways.
- during the period reseed that happens every 30 seconds,
  - why is it being reseeded with bytes from `strongRandom.generateSeed()`, when we're already reseeding every iteration with the secure bytes? I mean, I guess we're already interleaving the bytes into the hex codes, but the nonces are shuffled before they're sent to the channel, so an attacker trying to reverse that would have a very bad time.
  - why is the seed updated twice, first with the time since the last reseed, and then bytes from `generateSeed()`?

how useful really is any of this stuff that's being done?
is there any way to get in contact with Sergey Mashkov, who originally wrote this back in 2018?

bjhham:
These are all good questions which I cannot answer 😅 

@cy6erGn0m Sergey?

bjhham:
I can see how using a channel for prepopulating nonces makes sense to avoid blocking on `SecureRandom.nextBytes()`, but the rest of it not so much.  The weak random will be more performant, but if you're calling secure random every time to seed it with some entropy then you're still waiting on the secure random anyway.

bjhham (approved):
Looks good, thanks for the fixes!

bjhham:
Perhaps in the future, I'll look into further optimizations and a more straightforward approach without the shuffling obfuscation.  I haven't been able to find any relevant resources that encourage this method so I'm inclined to believe that they might've overemphasized the performance impact of using the secure random for everything.

PR #5521, opened 2026-04-13
Add ByteArray constructor to InetSocketAddress

### LINKED ISSUE #4978: Feature InetSocketAddress constructor with ByteArray
Currently a InetSocketAddress can be constructed via a hostname

Source: SocketAddressJvm.kt

public actual constructor(hostname: String, port: Int) :
      this(java.net.InetSocketAddress(hostname, port))

Would be nice if there is a constructor with an address array like 
public actual constructor(address: ByteArray, port: Int) : {
     val inet = InetAddress.getByAddress(address)           
     this(java.net.InetSocketAddress(inet, port))
}  
// Note this is only an example (not sure if that works, only to get the idea)

fru1tworld:
**Subsystem**
Network

**Motivation**
Fixes #4978. When callers already have raw IP bytes (e.g. from a parsed packet), there is no way to construct an `InetSocketAddress` without first converting to a string.

**Solution**
Add a secondary constructor `InetSocketAddress(address: ByteArray, port: Int)` that accepts 4-byte (IPv4) or 16-byte (IPv6) arrays in network byte order. Invalid lengths throw `IllegalArgumentException`.

### DIFF
--- a/ktor-network/api/ktor-network.api
+++ b/ktor-network/api/ktor-network.api
@@ -192,6 +192,7 @@ public final class io/ktor/network/sockets/DatagramWriteChannel$DefaultImpls {
 
 public final class io/ktor/network/sockets/InetSocketAddress : io/ktor/network/sockets/SocketAddress {
 	public fun <init> (Ljava/lang/String;I)V
+	public fun <init> ([BI)V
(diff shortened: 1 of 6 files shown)


bjhham (approved):
Thanks!  LGTM 👍

Issue #5397, opened 2026-02-26
Provide first-class request parameter validation DSL to replace verbose early-return patterns


e5l:
## Problem

A very common pattern in Ktor route handlers looks like this:

```kotlin
post("/checkout") {
    val userId = call.request.cookies["userId"]
        ?: return@post call.respondText("Login required", status = HttpStatusCode.Forbidden)
    val cartId = call.request.cookies["cartId"]
        ?: return@post call.respondText("Cart ID missing", status = HttpStatusCode.Forbidden)
    val amount = call.request.queryParameters["amount"]?.toLongOrNull()
        ?: return@post call.respondText("Amount missing", status = HttpStatusCode.BadRequest)
    // ... actual business logic
}
```

This pattern has several issues:

1. **Verbose boilerplate** — every parameter extraction repeats the same `?: return@label call.respondText(...)` structure
2. **Fragile labeled returns** — `return@post` is coupled to the enclosing lambda label; refactoring (e.g. extracting to a function) breaks it silently
3. **Mixed concerns** — validation logic, error response formatting, and HTTP status selection are all tangled in one expression
4. **No structured errors** — each handler invents its own error format, making consistent API error responses difficult
5. **No accumulation** — validation stops at the first error instead of reporting all problems at once
6. **Type conversion scattered** — `toLongOrNull()`, `toIntOrNull()`, etc. are ad-hoc and unvalidated (no range checks, no format messages)

### Idea 1: Validated parameter accessors on `ApplicationCall`

Provide extension functions that throw a well-defined exception (caught by a StatusPages-like mechanism) on missing/invalid parameters:

```kotlin
post("/checkout") {
    val userId = call.requireCookie("userId") // throws MissingRequestParameterException
    val cartId = call.requireCookie("cartId")
    val amount = call.requireQueryParameter("amount") { it.toLongOrNull() } // throws with descriptive message
    // ...
}
```

This removes the early-return boilerplate entirely and centralizes error response formatting in an exception handler.

### Idea 2: Validation DSL block

A structured validation block that accumulates errors and responds with all of them:

```kotlin
post("/checkout") {
    val (userId, cartId, amount) = call.validate {
        val userId = requireCookie("userId")
        val cartId = requireCookie("cartId")
        val amount = requireQueryParameter("amount").toLong("must be a valid number")
        Triple(userId, cartId, amount)
    }
    // ... all validated, business logic here
}
```

[...]

PR #5522, opened 2026-04-13
Add requireXxx extension functions for ApplicationCall

fru1tworld:
**Subsystem**
Server, ktor-server-core

**Motivation**
Relates to #5397 (Idea 4). Route handlers repeat `?: throw`/ `?: return` boilerplate when extracting required request parameters. `Parameters.getOrFail` exists for query/form params, but there is no equivalent for headers, cookies, or path parameters.

**Solution**
Add four `requireXxx` extension functions that return a non-null value or throw `MissingRequestParameterException` (→ 400):

- `ApplicationCall.requireQueryParameter(name)`
- `ApplicationCall.requireHeader(name)`
- `ApplicationCall.requireCookie(name, encoding)`
- `RoutingCall.requirePathParameter(name)`


ktor-server/ktor-server-core/common/src/io/ktor/server/request/RequireRequestParameter.kt
```diff
@@ -0,0 +1,51 @@
+/*
+ * Copyright 2014-2025 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+@file:Suppress("unused")
```
bjhham: I suspect this suppression is not required.

bjhham (commented):
Could you also introduce a `parameterType: String` field to `MissingRequestParameterException` for better tracing of the origin of these exceptions?  Options being { "query", "path", "cookie", "header" } along with an updated exception message.

fru1tworld:
Thanks for the feedback! Removed the `@file:Suppress("unused")` and added `parameterType` field with default `"request"`. Each `require*` function now passes the appropriate type (`"query"`, `"header"`, `"cookie"`, `"path"`).

bjhham (approved):

e5l (approved):
Hey @fru1tworld, thank you for the PR. LGTM

PR #5523, opened 2026-04-14
KTOR-8914 Add function reference reflection support to DI

bjhham:
**Subsystem**
Server, DI

**Motivation**
[KTOR-8914](https://youtrack.jetbrains.com/issue/KTOR-8914) Dependency injection: read annotations in function references

**Solution**
Introduced a function reference overload for providing dependencies that handles annotations and the like.

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-di/jvm/src/io/ktor/server/plugins/di/DependencyReflectionJvm.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-di/jvm/src/io/ktor/server/plugins/di/DependencyReflectionJvm.kt
@@ -50,6 +50,20 @@ public open class DependencyReflectionJvm : DependencyReflection {
             )
     }
 
+    override suspend fun <T> call(
+        kFunction: KFunction<T>,
+        init: suspend (DependencyKey) -> Any
+    ): T {
+        val args = mapParameters(kFunction.parameters) { param ->
+            init(toDependencyKey(param))
+        }
+        return if (kFunction.isSuspend) {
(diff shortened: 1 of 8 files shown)


ktor-server/ktor-server-plugins/ktor-server-di/jvm/src/io/ktor/server/plugins/di/DependencyReflectionJvm.kt
```diff
@@ -50,6 +50,20 @@ public open class DependencyReflectionJvm : DependencyReflection {
...
+    ): T {
+        val args = mapParameters(kFunction.parameters) { param ->
+            init(toDependencyKey(param))
+        }
+        return if (kFunction.isSuspend) {
+            kFunction.callSuspendBy(args)
+        } else {
+            kFunction.callBy(args)
+        }
+    }
```
osipxd, line 65: It seems we can introduce this as an extension function if don't want to break possible `DependencyReflection` implementations

osipxd (approved):
👍 
I have one optional comment

ktor-server/ktor-server-plugins/ktor-server-di/jvm/src/io/ktor/server/plugins/di/DependencyReflectionJvm.kt
bjhham: Hmm I think in this case, being forced to include the new function would be the easiest path for custom implementations, since they'd want to override this functionality as well.

PR #5525, opened 2026-04-14
KTOR-8989 Close or cancel engine only when the client reference count…

Stexxe:
https://youtrack.jetbrains.com/issue/KTOR-8989

The API dump was updated because an internal property has been introduced, which is used only within the `ktor-client-core` module.

### DIFF
--- a/ktor-client/ktor-client-core/api/ktor-client-core.api
+++ b/ktor-client/ktor-client-core/api/ktor-client-core.api
@@ -165,6 +165,8 @@ public final class io/ktor/client/engine/HttpClientEngine$DefaultImpls {
 }
 
 public abstract class io/ktor/client/engine/HttpClientEngineBase : io/ktor/client/engine/HttpClientEngine {
+	public static final synthetic field clientRefCount$FU$internal Ljava/util/concurrent/atomic/AtomicIntegerFieldUpdater;
+	public synthetic field clientRefCount$internal I
 	public fun <init> (Ljava/lang/String;)V
 	public fun close ()V
 	public fun getCoroutineContext ()Lkotlin/coroutines/CoroutineContext;
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/HttpClient.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/HttpClient.kt
@@ -15,6 +15,7 @@ import io.ktor.util.*
 import io.ktor.utils.io.*
 import io.ktor.utils.io.core.*
 import kotlinx.atomicfu.atomic
+import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.CompletableJob
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Job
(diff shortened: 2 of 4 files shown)


ktor-client/ktor-client-core/api/ktor-client-core.api
```diff
@@ -165,6 +165,8 @@ public final class io/ktor/client/engine/HttpClientEngine$DefaultImpls {
 }
 
 public abstract class io/ktor/client/engine/HttpClientEngineBase : io/ktor/client/engine/HttpClientEngine {
+	public static final synthetic field clientRefCount$FU$internal Ljava/util/concurrent/atomic/AtomicIntegerFieldUpdater;
+	public synthetic field clientRefCount$internal I
```
bjhham, line 169: It does seem a little unsettling that internal fields appear as public synthetic fields in the ABI snapshot.  I'm not sure if this means there's an issue in the ABI snapshotter or if Java consumers would see these because there's no equivalent.  Either way, I'm inclined to ignore it for now.

ktor-client/ktor-client-core/common/src/io/ktor/client/HttpClient.kt
bjhham, line 1280: Looks like we can remove this property since it's only used in the constructor.

ktor-client/ktor-client-core/common/src/io/ktor/client/HttpClient.kt
```diff
@@ -1292,6 +1285,26 @@ public class HttpClient(
...
     ) : this(engine, userConfig) {
         this.manageEngine = manageEngine
+
+        if (this.manageEngine) {
+            if (engine is HttpClientEngineBase) {
+                engine.clientRefCount.incrementAndGet()
+            }
+
+            coroutineContext[Job]!!.invokeOnCompletion { cause ->
+                val shouldClose = engine !is HttpClientEngineBase || engine.clientRefCount.decrementAndGet() <= 0
```
bjhham, line 1295: I think closing engines that don't implement `HttpClientEngineBase` immediately could cause some confusion for custom engines, since they'll be reverting to the behaviour in the bug.

Maybe it would make more sense to do something like:

```kotlin
if (manageEngine) {
    require(engine !is HttpClientEngineBase) {
        "manageEngine property is not supported by engine ${engine::class.simpleName}"
    }
}
```

Stexxe: I'd keep the current behavior for custom engines rather than throwing an exception because it's too intrusive.

ktor-client/ktor-client-core/common/src/io/ktor/client/HttpClient.kt
Stexxe: It's still used as a property within the `HttpClient.config` method.

bjhham: Oh I see 👀

bjhham (approved):

osipxd (approved):

PR #5527, opened 2026-04-15
HTTP/3 for Netty engine

bjhham:
**Subsystem**
Server, Netty

**Motivation**
[KTOR-9498](https://youtrack.jetbrains.com/issue/KTOR-9498) Support HTTP/3 for Netty

**Solution**
Introduced HTTP/3 codec with relevant configurations.  It started out as an exploratory session to see how different it would be to the HTTP/2 implementation, then it seemed to be pretty doable.

### DIFF
--- a/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationEngine.kt
+++ b/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationEngine.kt
@@ -30,6 +30,7 @@ import io.netty.handler.codec.http.HttpObjectDecoder
 import io.netty.handler.codec.http.HttpServerCodec
 import io.netty.handler.codec.quic.QuicSslContext
 import io.netty.handler.codec.quic.QuicSslContextBuilder
+import io.netty.handler.codec.quic.QuicTokenHandler
 import kotlinx.coroutines.CompletableJob
 import kotlinx.coroutines.asCoroutineDispatcher
 import java.net.BindException
@@ -164,6 +165,19 @@ public class NettyApplicationEngine(
         public var enableHttp3: Boolean = false
 
         /**
+         * The [QuicTokenHandler] used to generate and validate QUIC retry tokens
+         * when HTTP/3 is enabled. By default, a secure HMAC-SHA256-based handler
+         * is used that cryptographically signs tokens with a randomly generated key
+         * and rejects forged or expired tokens.
+         *
+         * Callers may replace this with a custom [QuicTokenHandler] implementation
+         * to use a different signing strategy or integrate with external token services.
+         *
+         * Only takes effect when [enableHttp3] is `true`.
+         */
+        public var quicTokenHandler: QuicTokenHandler = HmacQuicTokenHandler()
+
+        /**
          * Default function to configure Netty's
          */
         private fun defaultHttpServerCodec() = HttpServerCodec(
@@ -293,7 +307,8 @@ public class NettyApplicationEngine(
                     callEventGroup,
                     userContext,
                     configuration.runningLimit,
-                    quicSslContext
+                    quicSslContext,
+                    configuration.quicTokenHandler
                 )
             )
         }
--- a/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationEngine.kt
+++ b/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationEngine.kt
@@ -236,6 +236,9 @@ public class NettyApplicationEngine(
     }
     private val http3Bootstraps: List<Bootstrap> by lazy {
         if (!configuration.enableHttp3) return@lazy emptyList()
+        require(configuration.connectors.any { it is EngineSSLConnectorConfig }) {
+            "Netty HTTP/3 requires at least one SSL connector. Add an SSL connector or disable enableHttp3."
+        }
         configuration.connectors
             .filterIsInstance<EngineSSLConnectorConfig>()
             .map { createHttp3Bootstrap(it) }
--- a/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationEngine.kt
+++ b/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationEngine.kt
@@ -28,6 +28,7 @@ import io.netty.channel.socket.nio.NioDatagramChannel
 import io.netty.channel.socket.nio.NioServerSocketChannel
 import io.netty.handler.codec.http.HttpObjectDecoder
 import io.netty.handler.codec.http.HttpServerCodec
+import io.netty.handler.codec.quic.QuicServerCodecBuilder
 import io.netty.handler.codec.quic.QuicSslContext
 import io.netty.handler.codec.quic.QuicSslContextBuilder
 import io.netty.handler.codec.quic.QuicTokenHandler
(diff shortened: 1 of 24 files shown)


bjhham:
@coderabbitai review

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/http3/Http3FrameAdapter.kt
```diff
@@ -0,0 +1,28 @@
...
+import kotlinx.coroutines.channels.*
+
+internal suspend fun ReceiveChannel<Http3DataFrame>.http3frameLoop(bc: ByteWriteChannel) {
+    try {
+        while (true) {
+            val message = receive()
+            val content = message.content() ?: Unpooled.EMPTY_BUFFER
+
+            transferByteBuf(content, bc)
+            content.release()
```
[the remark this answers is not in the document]
bjhham: Why do you recommend calling `release()` on the message variable when working with `Http3DataFrame` but not `Http2DataFrame` as referenced in `HttpFrameAdapter.kt`?

bjhham:
@codex review

bjhham:
@codex review

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/http3/Http3FrameAdapter.kt
```diff
@@ -0,0 +1,31 @@
...
+import io.ktor.server.netty.http.*
+import io.ktor.utils.io.*
+import io.netty.buffer.*
+import io.netty.handler.codec.http3.*
+import kotlinx.coroutines.channels.*
+
+internal suspend fun ReceiveChannel<Http3DataFrame>.http3frameLoop(bc: ByteWriteChannel) {
+    try {
+        while (true) {
+            val message = receive()
```
e5l, line 16: Http3DataFrame is reference counted. Should we call for explicit release https://netty.io/4.2/api/io/netty/handler/codec/http3/Http3DataFrame.html?

e5l: same should apply in http2

e5l (approved):
lgtm, please check comments before merging

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/http/HttpMultiplexedConnectionPoint.kt
```diff
@@ -0,0 +1,106 @@
...
+        get() = httpVersion
+
+    override val uri: String
+        get() = pseudoPath?.toString() ?: "/"
+
+    @Deprecated(
+        "Use localHost or serverHost instead",
+        level = DeprecationLevel.ERROR
+    )
+    override val host: String
```
zibet27, line 41: Can we just delete this field as it's internal?

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/http3/HmacQuicTokenHandler.kt
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    override fun maxTokenLength(): Int = TIMESTAMP_LENGTH + HMAC_LENGTH + Quic.MAX_CONN_ID_LEN
+
+    private fun computeHmac(timestamp: Long, address: InetSocketAddress, dcidBytes: ByteBuffer): ByteArray {
+        val mac = Mac.getInstance("HmacSHA256")
+        mac.init(secretKey)
+
+        KtorDefaultPool.useInstance { timestampBuffer ->
+            timestampBuffer.clear()
```
zibet27: Shouldn't the timestampBuffer be already clear?

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/http3/Http3FrameAdapter.kt
bjhham: Claude assured me that `content.release()` worked transitively, although I do not see how it came to this conclusion now that you mention it...

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/http/HttpMultiplexedConnectionPoint.kt
bjhham: It's overriding a public interface ☹️

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/http3/HmacQuicTokenHandler.kt
bjhham: Yeah, looks like this can be dropped 👍

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/http/HttpMultiplexedConnectionPoint.kt
zibet27: Ohh, makes sense then

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/http3/Http3FrameAdapter.kt
bjhham: Looks like the frame is actually reused and calling `release()` on it here will break things.

zibet27: I guess we should add `retain()` calls to keep it alive where we need it

bjhham: Actually, looks like both release / retain just delegate to the content, so `content.release()` has the same effect as `message.release()`.  It might actually be best to call `message.release()` and avoid `content.release()` in case the implementation changes.

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationEngine.kt
```diff
@@ -146,6 +158,76 @@ public class NettyApplicationEngine(
...
+         * when HTTP/3 is enabled. By default, a secure HMAC-SHA256-based handler
+         * is used that cryptographically signs tokens with a randomly generated key
+         * and rejects forged or expired tokens.
+         *
+         * Callers may replace this with a custom [QuicTokenHandler] implementation
+         * to use a different signing strategy or integrate with external token services.
+         *
+         * Only takes effect when [enableHttp3] is `true`.
+         */
+        public var quicTokenHandler: QuicTokenHandler = HmacQuicTokenHandler()
```
osipxd: What if we add an function taking `configure` lambda? Something like this:

```kotlin
// Short form
http3()

// Or when additional configuration is necessary
http3 {
    quicMaxIdleTimeoutMillis = ...
}
```
Or will it be too inconsistent with the rest config?

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationEngine.kt
```diff
@@ -146,6 +158,76 @@ public class NettyApplicationEngine(
...
+
+        /**
+         * Maximum idle timeout for QUIC connections in milliseconds.
+         * If no data is exchanged within this period, the connection is closed.
+         *
+         * Only takes effect when [enableHttp3] is `true`.
+         */
+        public var quicMaxIdleTimeoutMillis: Long = 30_000
```
osipxd: Let's use `Duration` instead of `Long` for all new APIs. Probably we should save this somewehere as a rule.
@coderabbitai could you remember this?

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/cio/NettyHttpResponsePipeline.kt
```diff
@@ -207,9 +211,6 @@ internal class NettyHttpResponsePipeline(
             else -> -1
         }
 
-        call.isStreamingResponse = true
-        httpHandlerState.streamingResponses.incrementAndGet()
```
osipxd, line 211: This seems to be a rebase problem. In the main branch we don't check `contentType?.contains("text/event-stream")` and execute this code unconditionally here instead

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationEngine.kt
bjhham: Yeah it makes sense for it to be scoped.

bjhham: Maybe it should be `enableHttp3` for consistency though...

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/http3/NettyHttp3Configuration.kt
```diff
@@ -0,0 +1,72 @@
...
+     * Callers may replace this with a custom [QuicTokenHandler] implementation
+     * to use a different signing strategy or integrate with external token services.
+     */
+    public var quicTokenHandler: QuicTokenHandler = HmacQuicTokenHandler()
+
+    /**
+     * Maximum idle timeout for QUIC connections.
+     * If no data is exchanged within this period, the connection is closed.
+     */
+    public var quicMaxIdleTimeoutMillis: Duration = 30.seconds
```
osipxd: 💅 
```suggestion
    public var quicMaxIdleTimeout: Duration = 30.seconds
```

osipxd (approved):

PR #5528, opened 2026-04-16
Fix awaitContent returning true for closed empty channel

### LINKED ISSUE #4941: awaitContent returns incorrect value when channel is closed with empty buffer
## Description

The `awaitContent` function in `RawSourceChannel` incorrectly returns `true` when the channel is already closed but the buffer is empty. This violates the contract of `awaitContent`, which should return `false` when no more data can arrive.

## Current Behavior
When `closedToken != null`, the function immediately returns `true` regardless of buffer state.

## Expected Behavior  
When the channel is closed and insufficient data is available in the buffer, `awaitContent` should return `false` to indicate that the minimum required data cannot be satisfied.

## Impact
This causes callers to attempt reads that immediately hit EOF, as they're misled into thinking sufficient data is available.

## Context
This issue was identified during migration PR #4940 but kept unchanged to minimize migration scope. The bug exists in both the original JVM implementation and the new common implementation.

## References
- PR: https://github.com/ktorio/ktor/pull/4940
- Discussion: https://github.com/ktorio/ktor/pull/4940#discussion_r2152115539
- Reporter: @bishiboosh

fru1tworld:
**Subsystem**
ktor-io

**Motivation**
Fixes #4941. `RawSourceChannel.awaitContent()` returned `true` when closed with empty buffer.

### DIFF
--- a/ktor-io/jvm/src/io/ktor/utils/io/jvm/javaio/Reading.kt
+++ b/ktor-io/jvm/src/io/ktor/utils/io/jvm/javaio/Reading.kt
@@ -64,7 +64,10 @@ internal class RawSourceChannel(
         get() = buffer
 
     override suspend fun awaitContent(min: Int): Boolean {
-        if (closedToken != null) return true
+        if (closedToken != null) {
+            closedCause?.let { throw it }
+            return buffer.remaining >= min
+        }
 
         withContext(coroutineContext) {
             var result = 0L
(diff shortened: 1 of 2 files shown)


bjhham (approved):
Thanks!  This one looks like it could lead to some pretty unpleasant behaviour!

I've added [a ticket to YouTrack](https://youtrack.jetbrains.com/issue/KTOR-9500) for reference, this will be released in 3.5.0 next month.

PR #5529, opened 2026-04-16
KTOR-9497 Preventing a fatal crash in DarwinSession on close

LionZXY:
**Subsystem**
Client, ktor-client-darwin

**Motivation**
`DarwinSession.execute()` doesn't check the `closed` flag before calling `NSURLSession.dataTaskWithRequest()` /`webSocketTaskWithRequest()`. When `close()` races with an in-flight `execute()`, the invalidated `NSURLSession` throws an Objective-C `NSException` — which Kotlin/Native cannot catch, resulting in an unconditional `SIGABRT` crash. See [KTOR-9497](https://youtrack.jetbrains.com/issue/KTOR-9497/).

**Solution**
Add reproducer tests for both HTTP and WebSocket code paths that call `DarwinSession.execute()` on an already-closed session, verifying the failure is a catchable Kotlin exception rather than a process crash.

### DIFF
--- a/ktor-client/ktor-client-darwin/darwin/src/io/ktor/client/engine/darwin/internal/DarwinSession.kt
+++ b/ktor-client/ktor-client-darwin/darwin/src/io/ktor/client/engine/darwin/internal/DarwinSession.kt
@@ -9,10 +9,13 @@ import io.ktor.client.request.*
 import io.ktor.utils.io.*
 import io.ktor.utils.io.core.*
 import kotlinx.atomicfu.*
+import kotlinx.atomicfu.locks.SynchronizedObject
+import kotlinx.atomicfu.locks.synchronized
 import kotlinx.cinterop.*
 import kotlinx.coroutines.*
 import platform.Foundation.*
 import kotlin.coroutines.*
+import kotlin.coroutines.cancellation.CancellationException
 
 @OptIn(UnsafeNumber::class)
 internal class DarwinSession(
@@ -20,6 +23,7 @@ internal class DarwinSession(
     requestQueue: NSOperationQueue?
 ) : Closeable {
     private val closed = atomic(false)
+    private val closedLock = SynchronizedObject()
 
     private val sessionAndDelegate = config.sessionAndDelegate ?: createSession(config, requestQueue)
     private val session = sessionAndDelegate.first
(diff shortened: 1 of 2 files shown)


ktor-client/ktor-client-darwin/darwin/test/DarwinEngineTest.kt
```diff
@@ -275,6 +276,50 @@ class DarwinEngineTest : ClientEngineTest<DarwinClientEngineConfig>(Darwin) {
         assertSame(thrownException, challengeException, "Expected exception to be rethrown")
     }
 
+    // Issue: KTOR-9497
```
bjhham: nit: we don't tag our tests with issue numbers

ktor-client/ktor-client-darwin/darwin/src/io/ktor/client/engine/darwin/internal/DarwinSession.kt
```diff
@@ -20,6 +22,7 @@ internal class DarwinSession(
     requestQueue: NSOperationQueue?
 ) : Closeable {
     private val closed = atomic(false)
+    private val closedLock = SynchronizedObject()
```
bjhham: nit: I'd call this sessionLock, since the session is the shared resource it is guarding

ktor-client/ktor-client-darwin/darwin/test/DarwinEngineTest.kt
LionZXY: Fixed, thank you!

ktor-client/ktor-client-darwin/darwin/src/io/ktor/client/engine/darwin/internal/DarwinSession.kt
LionZXY: I think so too. Fixed. Thanks for the good param name

bjhham (approved):

ktor-client/ktor-client-darwin/darwin/src/io/ktor/client/engine/darwin/internal/DarwinSession.kt
```diff
@@ -55,9 +59,24 @@ internal class DarwinSession(
...
+        if (closed.value) return
+        synchronized(closedLock) {
+            if (!closed.compareAndSet(expect = false, update = true)) return
+            session.finishTasksAndInvalidate()
+        }
+    }
+
+    private fun cancelIfClosed() {
+        if (closed.value) throw CancellationException("Darwin session is closed")
     }
```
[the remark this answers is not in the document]
osipxd: I think the rabbit is wrong here and initial implementation was correct. We should treat this case as a normal situation and cancel coroutine instead of throwing an exception that would crash consumer's code.
But I need second option. @bjhham what do you think?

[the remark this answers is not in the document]
osipxd: @coderabbitai that comment was for you as well.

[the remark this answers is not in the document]
osipxd: Fixed in #5533

PR #5530, opened 2026-04-16
Gradle: Light sync mode

osipxd:
**Subsystem**
Infrastructure

**Motivation**
Make it easier for contributors to work on JVM/web-only tasks.

**Solution**
Light sync mode is enabled by default.
When working on native targets, it is possible to include specific targets or use full sync:
```properties
# Light mode with macosArm64 target included
ktorbuild.syncMode=light+macosArm64
# Full mode. Include all targets
ktorbuild.syncMode=full
```

In light mode native targets are ignored so import becomes significantly faster and consumes less memory.

**Hot sync benchmarks:**
image

_Better to review commit-by-commit_

### DIFF
--- a/build-logic/src/main/kotlin/ktorbuild.publish.gradle.kts
+++ b/build-logic/src/main/kotlin/ktorbuild.publish.gradle.kts
@@ -6,6 +6,10 @@ import com.vanniktech.maven.publish.MavenPublishBaseExtension
 import ktorbuild.*
 import ktorbuild.internal.*
 import ktorbuild.internal.publish.*
+import org.gradle.api.Project
+import org.gradle.api.publish.maven.tasks.AbstractPublishToMaven
+import org.gradle.kotlin.dsl.*
+import org.gradle.plugins.signing.Sign
 
 plugins {
     id("com.vanniktech.maven.publish")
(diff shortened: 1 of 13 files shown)


gradle.properties
```diff
@@ -87,10 +87,18 @@ android.useAndroidX=true
...
 # A username to be shown in build scans. Set to '<default>' to show the real username.
 #ktor.develocity.username=<default>
 # A hostname to be shown in build scans
 #ktor.develocity.hostname=
+
+# Change build mode:
+# - default: Builds the project as usual
+# - light: Excludes native targets from the project model to reduce memory consumption and IDE sync time.
+#   Should be used only for local development when working on JVM/web targets.
+#ktorbuild.buildMode=default
```
bjhham: Could we default this to light mode and toggle it on as appropriate in CI?  I think it would be a better experience for most folks looking to clone the project for the first time.

bjhham (approved):
Should we include some instruction under AGENTS.md as well?

build-logic/src/main/kotlin/ktorbuild.publish.gradle.kts
```diff
@@ -54,11 +58,14 @@ publishing {
...
 
 plugins.withId("ktorbuild.kmp") {
-    tasks.withType<AbstractPublishToMaven>().configureEach {
-        val os = ktorBuild.os.get()
-        // Workaround for https://github.com/gradle/gradle/issues/22641
-        val predicate = provider { isAvailableForPublication(publication.name, os) }
-        onlyIf("Available for publication on $os") { predicate.get() }
+    // Don't allow cross-compilation on CI, but it is okay to use it locally
+    if (ktorBuild.isCI.get()) {
+        tasks.withType<AbstractPublishToMaven>().configureEach {
```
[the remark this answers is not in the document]
osipxd: It's fine

gradle.properties
osipxd: Good idea, I'll prepare changes

osipxd:
> Should we include some instruction under AGENTS.md as well?

Now when this option affects only IDE sync I think it's okay not to include it in AGENTS.md.

gradle.properties
osipxd: Changed `buildMode` -> `syncMode` and enabled light mode by default

bjhham (approved):

PR #5532, opened 2026-04-17
Server bearer auth. Set principal with provider name

zibet27:
**Subsystem**
Server auth

**Motivation**
All other providers set the principal specifying the provider name
Found this during new typesafe DSL testing


bjhham (approved):
Nice catch 👍

PR #5533, opened 2026-04-17
Follow-up: KTOR-9497 Preventing a fatal crash in DarwinSession on close

osipxd:
Follow-up for #5529 

1. Fix exception type
2. Port the fix to Darwin legacy


bjhham (approved):
👍

PR #5466, opened 2026-03-19
KTOR-8364 Fix flaky Darwin engine test timeouts on macOS

e5l:
[KTOR-8364](https://youtrack.jetbrains.com/issue/KTOR-8364) Fix flaky tests in DarwinEngineTest and DarwinLegacyEngineTest

## Summary
- Fixes https://github.com/ktorio/ktor/issues/5462
- The `withTimeout(1000)` (1 second) in `DarwinEngineTest` and `DarwinLegacyEngineTest` runBlocking tests was too tight for CI environments under load, causing intermittent `TimeoutCancellationException` failures on macOS Arm64
- Increased timeout to 30 seconds, which is generous enough for loaded CI while still catching actual hangs

Closes #5462

## Test plan
- All affected tests (`testRequestInRunBlockingDispatchersDefault`, `testRequestInRunBlockingDispatchersUnconfined`, `testRequestInRunBlockingDispatchersNone`, `testRequestInRunBlocking`) pass locally on macosArm64
- No public API changes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #5534, opened 2026-04-19
KTOR-8364 Increase withTimeout in Darwin engine tests from 1s to 30s

fru1tworld:
**Subsystem**
Client

**Motivation**
Fixes #5475. Port of #5466 (`release/3.x`) to `main`.

`withTimeout(1000)` (1 second) in Darwin engine tests is too tight for CI under load, causing flaky `TimeoutCancellationException` on macOS Arm64.

**Solution**
Increase timeout to `30.seconds` in `DarwinEngineTest` (3 tests) and `DarwinLegacyEngineTest` (1 test).


bjhham:
We'll just merge release/3.x in a couple of days when 3.4.3 is released, thank you.

PR #5537, opened 2026-04-20
KTOR-9506 Add prefixItems to JsonSchema

bjhham:
**Subsystem**
Server, OpenAPI

**Motivation**
[KTOR-9506](https://youtrack.jetbrains.com/issue/KTOR-9506) OpenAPI: Support prefixItems in JsonSchema for tuple type definitions

**Solution**
Added the prefixItems property.

Since this API is marked under the experimental annotation, we can omit the need to introduce a backfill constructor on the data class.


zibet27 (approved):
lgtm :)

PR #5538, opened 2026-04-20
KTOR-470 Sessions. Add sendOnlyIfModified option to skip unchanged session resend

fru1tworld:
**Subsystem**
Server Sessions

**Motivation**
https://youtrack.jetbrains.com/issue/KTOR-470

Sessions plugin resends `Set-Cookie` on every response even when session data hasn't changed. This wastes bandwidth and prevents caching proxies (e.g. Cloudflare) from caching static assets that share the same response.

**Solution**
Add an opt-in `sendOnlyIfModified` property to `CookieSessionBuilder` and `HeaderSessionBuilder`. When enabled, the plugin compares `newValue` to `oldValue` using structural equality (`equals()`) and skips `Set-Cookie` / header resend if unchanged.

Default is `false` — existing behavior (always resend for cookie expiration refresh) is preserved.

```kotlin
install(Sessions) {
    cookie<MySession>("SESSION") {
        sendOnlyIfModified = true
    }
}
```

Session classes should implement `equals()` (e.g. `data class`) for correct comparison.

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-sessions/common/src/io/ktor/server/sessions/SessionProvider.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-sessions/common/src/io/ktor/server/sessions/SessionProvider.kt
@@ -15,7 +15,7 @@ import kotlin.reflect.*
  * @param transport specifies the [SessionTransport] for this provider
  * @param tracker specifies the [SessionTracker] for this provider
  * @param sendOnlyIfModified when set to `true`, session data is not re-sent if unchanged from the incoming value.
- * This avoids unnecessary `Set-Cookie` headers but prevents cookie expiration refresh.
+ * This avoids unnecessary session response headers. For cookie sessions, this also prevents cookie expiration refresh.
  * Session classes should properly implement `equals()` for this to work correctly.
  * Default: `false`.
  * @property name session name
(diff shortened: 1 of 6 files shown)


ktor-server/ktor-server-plugins/ktor-server-sessions/common/src/io/ktor/server/sessions/SessionProvider.kt
```diff
@@ -23,6 +23,16 @@ public class SessionProvider<S : Any>(
...
 ) {
+    /**
+     * When set to `true`, session data is not re-sent if unchanged from the incoming value.
+     * This avoids unnecessary `Set-Cookie` headers but prevents cookie expiration refresh.
+     * Session classes should properly implement `equals()` for this to work correctly.
+     * Default: `false`.
+     *
+     * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.server.sessions.SessionProvider.sendOnlyIfModified)
+     */
+    public var sendOnlyIfModified: Boolean = false
```
bjhham: Looks like it would be more consistent to include this property in the constructor.

fru1tworld:
Moved `sendOnlyIfModified` to the constructor as suggested.

Also added a `!==` check to prevent in-place mutations from being silently skipped, with a test.

fru1tworld:
Applied: made KDoc transport-neutral and added sendOnlyIfModified to toString().

bjhham (approved):
Thanks for the contribution!  Always nice to see these ancient tickets close 😄

PR #5539, opened 2026-04-20
Rename Jackson.streamRequestBody

zibet27:
**Subsystem**
Shared

**Motivation**
[KTOR-9503](https://youtrack.jetbrains.com/issue/KTOR-9503) The JacksonConverter.streamRequestBody property name is confusing

**Solution**
Rename `streamRequestBody` -> `streamBody`

### DIFF
--- a/ktor-shared/ktor-serialization/ktor-serialization-jackson/jvm/src/JacksonConverter.kt
+++ b/ktor-shared/ktor-serialization/ktor-serialization-jackson/jvm/src/JacksonConverter.kt
@@ -27,12 +27,12 @@ import kotlin.text.*
  * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.serialization.jackson.JacksonConverter)
  *
  * @param objectMapper a configured instance of [ObjectMapper]
- * @param streamRequestBody if set to true, will stream request body, without keeping it whole in memory.
+ * @param streamBody if set to true, will stream request/response body, without keeping it whole in memory.
  * This will set `Transfer-Encoding: chunked` header.
  */
 public class JacksonConverter(
     private val objectMapper: ObjectMapper = jacksonObjectMapper(),
-    private val streamRequestBody: Boolean = true
+    private val streamBody: Boolean = true,
 ) : ContentConverter {
 
     override suspend fun serialize(
@@ -41,7 +41,7 @@ public class JacksonConverter(
         typeInfo: TypeInfo,
         value: Any?
     ): OutgoingContent {
-        if (!streamRequestBody && typeInfo.type != Flow::class) {
+        if (!streamBody && typeInfo.type != Flow::class) {
             return TextContent(
                 objectMapper.writeValueAsString(value),
                 contentType.withCharsetIfNeeded(charset)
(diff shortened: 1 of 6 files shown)


zibet27:
Looks like this doesn't change ABI, but breaks named parameters

ktor-shared/ktor-serialization/ktor-serialization-jackson/jvm/src/JacksonConverter.kt
```diff
@@ -27,12 +27,12 @@ import kotlin.text.*
...
  *
  * @param objectMapper a configured instance of [ObjectMapper]
- * @param streamRequestBody if set to true, will stream request body, without keeping it whole in memory.
+ * @param streamBody if set to true, will stream request/response body, without keeping it whole in memory.
  * This will set `Transfer-Encoding: chunked` header.
  */
 public class JacksonConverter(
     private val objectMapper: ObjectMapper = jacksonObjectMapper(),
-    private val streamRequestBody: Boolean = true
+    private val streamBody: Boolean = true,
```
osipxd, line 35: if we want to keep source compatibility we could introduce an overload with old name and `DeprecationLevel.WARNING`. We could even remove the default value for `streamRequestBody` there.

osipxd: And it'll probably require adding a dummy `Unit` parameter to resolve ambiguity.

bjhham (commented):
I think we'll wanna target this to 4.0.0 to avoid breaking the parameter name.

ktor-shared/ktor-serialization/ktor-serialization-jackson/jvm/src/JacksonConverter.kt
zibet27: What can we do with such functions?

```kotlin
public fun Configuration.jackson(
    contentType: ContentType = ContentType.Application.Json,
    streamBody: Boolean = true,
    block: ObjectMapper.() -> Unit = {}
) {
```

Also, add one more parameter?
If we just change the name, it will be an overload error

bjhham:
Disregard me, sounds like Osip's got some ideas 😄

ktor-shared/ktor-serialization/ktor-serialization-jackson/jvm/src/JacksonConverter.kt
zibet27: > And it'll probably require adding a dummy `Unit` parameter to resolve ambiguity.

Sorry, didn't see this one

ktor-shared/ktor-serialization/ktor-serialization-jackson/jvm/src/JacksonConverter.kt
```diff
@@ -27,21 +27,28 @@ import kotlin.text.*
...
 ) : ContentConverter {
 
+    @Deprecated("Use constructor(objectMapper, streamBody) instead", level = DeprecationLevel.WARNING)
```
osipxd: Here we could provide `ReplaceWith("JacksonConverter(objectMapper, streamBody = streamRequestBody)")` for quick migration

ktor-shared/ktor-serialization/ktor-serialization-jackson/jvm/src/JacksonConverter.kt
```diff
@@ -174,6 +181,28 @@ public fun Configuration.jackson(
...
+ *
+ *
+ * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.serialization.jackson.jackson)
+ *
+ * @param contentType the content type to send with a request
+ * @param streamRequestBody if set to true, will stream request body, without keeping it whole in memory.
+ * This will set `Transfer-Encoding: chunked` header.
+ * @param block a configuration block for [ObjectMapper]
+ */
+@Deprecated("Use jackson(contentType, streamBody, block) instead", level = DeprecationLevel.WARNING)
```
osipxd: Same here `ReplaceWith("jackson(contentType, streamBody = streamRequestBody, block)")`

osipxd (commented):
It seems the rabbit is right about one missing `@Deprecated` annotation. Please check it

bjhham (approved):

osipxd (approved):

PR #5540, opened 2026-04-20
Fix WebSocket pingInterval ignored in OkHttp engine

Programistich:
**Subsystem**

Client/Server, OkHttp engine, WebSockets plugin

**Motivation**

Fixes [KTOR-4752](https://youtrack.jetbrains.com/issue/KTOR-4752/OkHttp-Websockets-pinging-doesnt-work): `pingIntervalMillis` configured in the `WebSockets` plugin is completely ignored when using the OkHttp engine.

Previously, the `pingIntervalMillis` setter in `OkHttpWebsocketSession` intentionally threw a `WebSocketException` — pinging was considered OkHttp's own responsibility, not the plugin's. This created an inconsistency: every other engine (CIO, Darwin) honors the plugin-level `pingIntervalMillis`, while OkHttp silently ignored it.

The root cause: `OkHttpWebsocketSession` implements `DefaultWebSocketSession` directly, causing `convertSessionToDefault()` in the WebSockets plugin to return it as-is — skipping all plugin-level configuration including `pingIntervalMillis`. At that time there was no way to inject the plugin's value into the session.

**This is now fixable** because `WebSocketsConfig` (including `pingIntervalMillis`) is already available at the point where the session is constructed in `OkHttpEngine.executeWebSocketRequest()` — we simply pass it through to the session constructor.

**Solution**

1. **`OkHttpEngine.kt`** — Pass `wsConfig.pingIntervalMillis` when constructing `OkHttpWebsocketSession`
2. **`OkHttpWebsocketSession.kt`** — Accept `pingIntervalMillis` as a constructor parameter (defaults to `engine.pingIntervalMillis` for backward compatibility); remove the throwing setter

### DIFF
--- a/ktor-client/ktor-client-okhttp/jvm/src/io/ktor/client/engine/okhttp/OkHttpEngine.kt
+++ b/ktor-client/ktor-client-okhttp/jvm/src/io/ktor/client/engine/okhttp/OkHttpEngine.kt
@@ -90,7 +90,7 @@ public class OkHttpEngine(override val config: OkHttpConfig) : HttpClientEngineB
             webSocketFactory = config.webSocketFactory ?: engine,
             engineRequest = engineRequest,
             coroutineContext = callContext,
-            channelsConfig = wsConfig.channelsConfig
+            wsConfig = wsConfig,
         ).apply { start() }
 
         val originResponse = session.originResponse.await()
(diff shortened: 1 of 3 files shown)


bjhham (approved):
Thank you for the contribution!  Looks good to me 👍

PR #5546, opened 2026-04-21
Fix hanging testProducingWriteChannelContentOnScale

osipxd:
**Subsystem**
Test Infrastructure

**Motivation**
Sometimes this test [could hang for 2 hours](https://ktor.teamcity.com/test/-580217390588598146?currentProjectId=Ktor_ProjectKtorCore) and then build got killed
Normal execution takes also quite long — up to 4 minutes

**Solution**
- Use `runTest` with timeout instead of `runBlocking`
- Drop usage of `encodeBase64` - it was ~15% of execution time
- Refactor `TestEncoder` to drop `runBlocking` usage

### DIFF
--- a/ktor-client/ktor-client-apache/jvm/test/io/ktor/client/engine/apache/RequestProducerTest.kt
+++ b/ktor-client/ktor-client-apache/jvm/test/io/ktor/client/engine/apache/RequestProducerTest.kt
@@ -6,25 +6,33 @@ package io.ktor.client.engine.apache
 
 import io.ktor.client.request.*
 import io.ktor.http.*
-import io.ktor.http.HttpHeaders
 import io.ktor.http.content.*
 import io.ktor.util.*
 import io.ktor.utils.io.*
-import kotlinx.coroutines.*
-import kotlinx.coroutines.test.*
-import org.apache.http.*
-import org.apache.http.nio.*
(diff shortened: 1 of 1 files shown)


ktor-client/ktor-client-apache/jvm/test/io/ktor/client/engine/apache/RequestProducerTest.kt
osipxd: This is the only hot spot now, but If I understand it correctly, it's intentional:

Image

One-by-one byte writing has been introduced in https://github.com/ktorio/ktor/pull/4231

osipxd:
Now it's faster but not always 🤔 

image

ktor-client/ktor-client-apache/jvm/test/io/ktor/client/engine/apache/RequestProducerTest.kt
bjhham: I remember this one... I don't think it's necessary to keep this behaviour.  It's necessary to have a test that fills the byte channel buffer a few times so we know the flush is working with streaming, but we don't need the byte-by-byte writing.  Might as well use a 4k buffer or similar here.

bjhham (approved):
👍

PR #5550, opened 2026-04-23
always complete or cancel responseReady in NettyApplicationResponse

larshagencognite:
**Subsystem**
ktor-server-netty

**Motivation**
We have observed hangs between ktor request finishing, and client getting the response, causing the client to time out. This problem started with ktor 3.4.0. 

This change introduced a possibility of not completing the promise when channel is inactive: https://github.com/ktorio/ktor/pull/5181.

I do not have clear evidence that this is the root cause, but I think the new behavior in this PR is an improvement. 

**Solution**
When channel is inactive, cancel the response instead of leaving the promise in a non-completed state.

### DIFF
--- a/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationResponse.kt
+++ b/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationResponse.kt
@@ -55,7 +55,7 @@ public abstract class NettyApplicationResponse(
         val chunked = headers[HttpHeaders.TransferEncoding] == "chunked"
 
         if (!canRespond) {
-            cancelInChannelNotActive()
+            cancelIfChannelNotActive()
             return
         }
 
@@ -118,7 +118,7 @@ public abstract class NettyApplicationResponse(
 
     internal fun sendResponse(chunked: Boolean = true, content: ByteReadChannel) {
         if (!canRespond) {
-            cancelInChannelNotActive()
+            cancelIfChannelNotActive()
             return
         }
 
@@ -151,7 +151,7 @@ public abstract class NettyApplicationResponse(
         // while close only does flush() and doesn't terminate connection
     }
 
-    private fun cancelInChannelNotActive() {
+    private fun cancelIfChannelNotActive() {
         if (!context.channel().isActive) {
             cancel()
         }
--- a/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationResponse.kt
+++ b/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationResponse.kt
@@ -33,9 +33,6 @@ public abstract class NettyApplicationResponse(
 
     internal var responseChannel: ByteReadChannel = ByteReadChannel.Empty
 
-    private val canRespond: Boolean
-        get() = !responseMessageSent && context.channel().isActive
-
     override suspend fun respondOutgoingContent(content: OutgoingContent) {
         try {
             super.respondOutgoingContent(content)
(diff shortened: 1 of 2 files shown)


larshagencognite:
@coderabbitai review

ktor-server/ktor-server-netty/jvm/test/io/ktor/tests/server/netty/NettySpecificTest.kt
```diff
@@ -171,6 +172,68 @@ class NettySpecificTest {
...
+
+            // Give Netty time to process the channel-inactive event
+            delay(300.milliseconds)
+
+            // Now let the handler try to respond to the already-closed channel
+            shouldRespond.complete(Unit)
+
+            // finishedEvent must be resolved (success or failure) — if responseReady is
+            // never completed, this deferred hangs forever and the test times out
+            withTimeout(5.seconds) { callFinished.await() }
```
bjhham: Our CI is pretty sensitive to flakiness with regards to timeouts.  It's likely we'll need to bump this up to 20 seconds or some other huge figure and maybe tweak the initial delay.  The use of `CompletableDeferred` looks good to mitigate most of this though 👍

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationResponse.kt
```diff
@@ -54,7 +51,12 @@ public abstract class NettyApplicationResponse(
         // because it should've been set by commitHeaders earlier
         val chunked = headers[HttpHeaders.TransferEncoding] == "chunked"
 
-        if (!canRespond) return
+        if (responseMessageSent) return
+
+        if (!context.channel().isActive) {
+            cancel()
+            return
+        }
```
bjhham, line 60: I would keep the common check here just to avoid any future mistakes from duplication, but using a function like:

```
private fun responseChannelUnavailable(): Boolean =
        responseMessageSent || (!context.channel().isActive && run { cancel(); true })
```

bjhham (commented):
Looks very likely, thanks for looking into it!

zibet27 (approved):

bjhham:
I've logged an issue in YouTrack for tracking [KTOR-9524](https://youtrack.jetbrains.com/issue/KTOR-9524)

Please let me know if there's any changes you'd make to the description there.

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationResponse.kt
larshagencognite: I changed it to use a common `cancelIfChannelNotActive`, is that better? I would like to avoid a method that seems purely read-only, but has a cancel side-effect.

ktor-server/ktor-server-netty/jvm/test/io/ktor/tests/server/netty/NettySpecificTest.kt
larshagencognite: I increased it to 20 seconds

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationResponse.kt
bjhham: Yeah that works for me 👍

bjhham (approved):

ktor-server/ktor-server-netty/jvm/test/io/ktor/tests/server/netty/NettySpecificTest.kt
```diff
@@ -171,6 +172,68 @@ class NettySpecificTest {
...
+
+            // Give Netty time to process the channel-inactive event
+            delay(300.milliseconds)
+
+            // Now let the handler try to respond to the already-closed channel
+            shouldRespond.complete(Unit)
+
+            // finishedEvent must be resolved (success or failure) — if responseReady is
+            // never completed, this deferred hangs forever and the test times out
+            withTimeout(20.seconds) { callFinished.await() }
```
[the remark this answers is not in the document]
larshagencognite: @coderabbitai I have now added more explicit polling, so we know the channel is inactive before proceeding to send response

larshagencognite:
The failing tests look unrelated. Is there anything more I need to do for this PR?

bjhham:
Nope, you're all set!  I'll merge it now, thanks!

PR #5552, opened 2026-04-23
Update Gradle and Develocity

osipxd:
And also report sync mode in build scan so we could aggregate data later

image


bjhham (approved):
Neat feature 👍

PR #5553, opened 2026-04-23
Improve static content handling

solonovamax:
**Subsystem**
ktor-server module

**Motivation**

Fixes:
- [KTOR-9526](https://youtrack.jetbrains.com/issue/KTOR-9526)
- [KTOR-8326](https://youtrack.jetbrains.com/issue/KTOR-8326) (correctly)
- [KTOR-8672](https://youtrack.jetbrains.com/issue/KTOR-8672)

Also a couple of things that I felt were lacking with the static file handling.

**Solution**

Changes:
- Adds support for zstd & deflate static compressed files (imo, this should really be a class with a companion object instead of an array, but that's a backwards incompatible change)
- Slightly improves performance of static file handling by:
  - using arrays for extensions/pre compressed file types
  - store an array of excludes instead of recursing into the old exclude
- Rejects precompressed static files if the base file does not exist. This is a change in the previous behaviour. Previously, if `test.txt.br` existed and `test.txt` did not and a request was made for `test.txt` with support for brotli compression, the request would succeed. imo, this behaviour does not make sense, so I changed it to instead only succeed if `test.txt` also exists.
  the reason for changing this is because this would result in clients with different `Accept` headers having different results when requesting a resource. some would see the resource, some wouldn't.
  - removes some tests in SinglePageApplicationTest which were failing due to this. I don't think those tests made much sense anyways, and due to the change in behaviour they no longer make sense to keep around.
- clean up a bunch of the tests in `StaticContentTest`
  - merges many of the tests that are for the different static content types (`File`/`Path`/resources), to ensure that behaviour is always tested on all three.
- when determining which precompressed file to send, send the one with the smallest size.
- make sure the `Vary` header is applied correctly. previously, it would not get added to static content if `staticFileSystem` was used.
- remove `SensitivityWatchEventModifier`.
  - [JDK-8303188](https://bugs.openjdk.org/browse/JDK-8303188)
  - [JDK-8303175](https://bugs.openjdk.org/browse/JDK-8303175)
  - [JDK-8303471](https://bugs.openjdk.org/browse/JDK-8303471)
  - [JDK-8303413](https://bugs.openjdk.org/browse/JDK-8303413)
  - openjdk/jdk#12746
  - openjdk/jdk@ae290541d17be77bbedf56a4592102ead4e4691e
  - openjdk/jdk#12795
  - [KTOR-1647](https://youtrack.jetbrains.com/issue/KTOR-1647)
- Cache the default file in memory for `staticFileSystem`
- add an overload of `staticFileSystem` which takes `Path` for the base path & index, instead of `String`. note: these overloads are ambiguous, so this change is not appropriate for a patch update.
- fix RFC 9110 compliance for content encodings, which requires encodings with `q=0` be rejected and encodings be matched case-insensitively. (currently does not respond with a 415 when `identity` has `q=0`, in order to mirror the behaviour of the `Compression` plugin)

my current implementation does *not* check exclusion rules for the default file. I felt this was reasonable, as it would be odd that something were configured as the default but had an exclusion rule that filtered it out.
would you like for me to also always check the exclusion rule for the default file?

some remaining things I want to do before this is merged but haven't gotten the chance to:
- [x] add some more tests for static content
  - [x] add test for the new way of configuring static content, as ~~all~~ a lot of the tests currently use the old way
  - [x] add test for new behaviour with static content & precompressed content
  - [x] add test for default file caching
- [x] add a way to "filter" calls (distinct from "exclude": "exclude" will cause it to return a 403, whereas "filter" will just silently pass over it)
- [x] add caching for the default file
  - [x] for the `Path` variant
  - [x] for the `File` variant
  - [x] ~~for the `URL`/resource variant~~ not sure I can do this without a bunch of effort, as registering a watcher for a resource would be quite annoying.
- [x] add an overload for `staticFileSystem` for `Path`
- [x] add a function `staticPaths`, which just directly calls `staticFileSystem`. (I personally prefer the name `staticPaths` to `staticFileSystem


solonovamax:
I also have some a change which stores the contents of the default file (& all precompressed versions) in a byte array in memory, and then watch the files for changes.
this allows for any requests which end up serving the default file to avoid any IO requests as it's already cached in memory

if you'd like, I can add these changes to the PR.

solonovamax:
Also, I tried running `:ktor-server:updateKotlinAbi`, however it did not change the api dump file at all, even though I've modified the `CompressedFileType`. not sure why this happened, I'll look into it later.

ktor-server/ktor-server-core/jvm/src/io/ktor/server/http/content/PreCompressed.kt
```diff
@@ -213,44 +284,50 @@ internal suspend fun ApplicationCall.respondStaticResource(
...
-            etag.provide(bestCompressionFit.url)?.let { versions += it }
-            lastModified(bestCompressionFit.url)?.let { versions += LastModifiedVersion(it) }
-        }
+
+        val content = PreCompressedResponse(bestCompressionFit.content, bestCompressionFit.compression)
+            .provideVersions(etag, lastModified, bestCompressionFit.url)
+
         respond(content)
-        return
     }
```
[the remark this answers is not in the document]
solonovamax: no it doesn't.

ktor-server/ktor-server-core/jvm/src/io/ktor/server/http/content/PreCompressed.kt
```diff
@@ -94,116 +149,132 @@ internal fun bestCompressionFit(
...
-    compressedTypes: List<CompressedFileType>?,
+    compressedTypes: Array<CompressedFileType>,
     contentType: (Path) -> ContentType = { ContentType.defaultForPath(it) },
     cacheControl: (Path) -> List<CacheControl> = { emptyList() },
     modify: suspend (Path, ApplicationCall) -> Unit = { _, _ -> },
     lastModified: (Path) -> GMTDate? = { null },
     etag: ETagProvider = ETagProvider { null },
 ) {
+    if (!requestedPath.exists())
+        return
```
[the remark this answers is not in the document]
solonovamax: checked elsewhere. not an issue.

ktor-server/ktor-server-core/jvm/src/io/ktor/server/http/content/PreCompressed.kt
```diff
@@ -94,116 +149,132 @@ internal fun bestCompressionFit(
...
+            continue
+
+        val compressed = "$resource.${compressedType.extension}"
+        val resolved = call.application.resolveResource(compressed, packageName) { url ->
+            val requestPath = url.path.replace(
+                Regex("${Regex.escapeReplacement(compressed.substringAfterLast(File.separator))}$"),
+                resource.substringAfterLast(File.separator)
+            )
+            contentType(URL(url.protocol, url.host, url.port, requestPath))
+        } ?: continue
```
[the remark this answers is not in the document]
solonovamax: not my code, not my issue.

ktor-server/ktor-server-core/jvm/src/io/ktor/server/http/content/PreCompressed.kt
```diff
@@ -37,50 +38,148 @@ internal val Route.staticContentEncodedTypes: List<CompressedFileType>?
...
     acceptEncoding: List<HeaderValue>,
-    compressedTypes: List<CompressedFileType>?
-): CompressedFileType? {
-    val acceptedEncodings = acceptEncoding.map { it.value }.toSet()
-    // We respect the order in compressedTypes, not the one in Accept header
-    return compressedTypes
-        ?.filter { it.encoding in acceptedEncodings }
-        ?.firstOrNull { File("${file.absolutePath}.${it.extension}").isFile }
+    compressedTypes: Array<CompressedFileType>
+): Pair<File, CompressedFileType>? {
```
bjhham, line 89: I would leave the current implementation of going by the configured order of compression types.  Sorting by file size could impact performance since checking file size is a system call.  We could make this a configurable function too so users can provide custom heuristics.

ktor-server/ktor-server-core/jvm/src/io/ktor/server/http/content/PreCompressed.kt
```diff
@@ -26,8 +25,10 @@ import kotlin.io.path.*
  * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.server.http.content.CompressedFileType)
  */
 public enum class CompressedFileType(public val extension: String, public val encoding: String = extension) {
+    ZSTD("zst", "zstd"),
     BROTLI("br"),
-    GZIP("gz", "gzip")
+    GZIP("gz", "gzip"),
+    DEFLATE("deflate"),
 }
```
bjhham, line 31: Might be better to deprecate this enum and introduce an interface with the two properties.  This way, people can define their own extensions for pre-compressed files and this would not introduce a source-breaking change on consumers exhaustive when expressions.

bjhham (commented):
I noticed there was a small increase to the memory allocations in the file responses:

```
Request consumes 37.19 KB, expected 34.58 KB. 
      Difference: 2673 < 4249 (allowed)
      Consumed 37.19 KB on request
      Expected 34.58 KB on request
      Extra    2673 bytes on request
    
    Increased locations:
      StaticContentResolution.kt                    1776    (10000        -->        11776)
      RoutingResolveTrace.kt                         664    (0            -->          664)
      Paths.kt                                       216    (216          -->          432)
      RoutingResolveContext.kt                       120    (1104         -->         1224)
      NettyApplicationCall.kt                         72    (240          -->          312)
      StaticContent.kt                                40    (528          -->          568)
      NettyHttpResponsePipeline.kt                    29    (7705         -->         7734)
      NettyConnectionPoint.kt                         24    (0            -->           24)
    
    Decreased locations:
      PreCompressed.kt                              -264    (728          -->          464)
      ByteChannel.kt                                  -3    (1960         -->         1957)
      ByteWriteChannelOperations.kt                   -1    (785          -->          784)
```

This might be due to the file size checking, I'm not sure.

Please let me know when you're satisfied with the remaining tasks in your checklist 👀

ktor-server/ktor-server-core/jvm/src/io/ktor/server/http/content/PreCompressed.kt
solonovamax: I would argue that there is a higher cost to increased network usage compared to the cost of a system call

ktor-server/ktor-server-core/jvm/src/io/ktor/server/http/content/PreCompressed.kt
solonovamax: previously, in [KTOR-8672](https://youtrack.jetbrains.com/issue/KTOR-8672), I argued that it should be a data class with some fields in the companion object:

```kotlin
data class CompressedFileType(val extension: String, val encoding: String = extension) {
    companion object {
        val BROTLI = CompressedFileType("br")
        val GZIP = CompressedFileType("gz", "gzip")
        val DEFLATE = CompressedFileType("deflate")
        val ZSTD = CompressedFileType("zst", "zstd")
    }
}
```

I think this approach is better than an interface. I don't see any need for an interface, as there won't really be any methods on it.

however, this would basically result in all methods being doubled.

the problem is, I'm not entirely sure what it would be called. I would *like* to use `CompressedFileType`, but that's taken.

~~also, if I'm doing that then iirc there was a couple of other pieces of the api I was dissatisifed with, so I'd also likely look at changing those in the new version.~~

if you would like, I could do this.

bjhham: The interface would just be there to avoid a breaking change without needing to duplicate every function with new arguments - the enum would just implement the interface so current references would be valid.  Instead of using the data class we can just use data objects that implement with the two fields.

ktor-server/ktor-server-core/jvm/src/io/ktor/server/http/content/PreCompressed.kt
bjhham: Yes, but most of the time the compression methods are already in the correct order, this would be an unnecessary operation that could be interpreted as a performance regression.  It's better to make this configurable.

solonovamax: sure, I can add a toggle for it

ktor-server/ktor-server-core/jvm/src/io/ktor/server/http/content/PreCompressed.kt
solonovamax: hmmm, I see

in that case I'd probably still go with a data class, but just have the data class implement the interface

solonovamax:
> I noticed there was a small increase to the memory allocations in the file responses:
> 
> ```
> Request consumes 37.19 KB, expected 34.58 KB. 
>       Difference: 2673 < 4249 (allowed)
>       Consumed 37.19 KB on request
>       Expected 34.58 KB on request
>       Extra    2673 bytes on request
>     (See stdout + build/allocations/* files for details)
>     
>     Increased locations:
>       StaticContentResolution.kt                    1776    (10000        -->        11776)
>       RoutingResolveTrace.kt                         664    (0            -->          664)
>       Paths.kt                                       216    (216          -->          432)
>       RoutingResolveContext.kt                       120    (1104         -->         1224)
>       NettyApplicationCall.kt                         72    (240          -->          312)
>       StaticContent.kt                                40    (528          -->          568)
>       NettyHttpResponsePipeline.kt                    29    (7705         -->         7734)
>       NettyConnectionPoint.kt                         24    (0            -->           24)
>     
>     Decreased locations:
>       PreCompressed.kt                              -264    (728          -->          464)
>       ByteChannel.kt                                  -3    (1960         -->         1957)
>       ByteWriteChannelOperations.kt                   -1    (785          -->          784)
> ```
> 
> This might be due to the file size checking, I'm not sure.

what did you use to benchmark this? was this a specific tool, or are you just writing down the differences you noticed?

ktor-server/ktor-server-core/jvm/src/io/ktor/server/http/content/PreCompressed.kt
solonovamax: I'm looking at this right now, and there's three ways I could do this.
1. leave it in the current state, where it sorts by file size
2. add a boolean toggle for this, so it either sorts by file size or uses the previous behaviour
3. allows users to provide a custom function for sorting

the 3rd option seems to be rather complicated to do, due to the different scenarios:
- using `File`? it needs a `(file: File, acceptEncoding: List<HeaderValue>, compressedTypes: Array<CompressedFileType>) -> Pair<File, CompressedFileType>?`
- using `Path`? it needs a `(fileSystem: FileSystemPaths, path: Path, acceptEncoding: List<HeaderValue>, compressedTypes: Array<CompressedFileType>) -> Pair<Path, CompressedFileType>?`
- using `URL`? it needs a `(call: ApplicationCall, resource: String, packageName: String?, acceptEncoding: List<HeaderValue>, compressedTypes: Array<CompressedFileType>, contentType: (URL) -> ContentType) -> CompressedResource?`
- is the content cached in-memory (currently only applicable for `Path`, but I intend to add support for `File` and maybe `URL`)? it needs a `(acceptEncoding: List<HeaderValue>, compressedFiles: Array<Pair<CachedStaticFile, CompressedFileType>>) -> Pair<CachedStaticFile, CompressedFileType>?`

I don't really see a reasonable way to add the 3rd option without a ton of effort.

I think if someone is really that concerned about performance, they would likely want to just vendor everything in order to simplify the code in a bunch of places to eliminate features they don't need/use.

ktor-server/ktor-server-core/jvm/src/io/ktor/server/http/content/PreCompressed.kt
solonovamax: tbh, I'm not so sure on this

although it is technically breaking, how many users are *actually* using it for an exhaustive when expression? doing a search on github for `/CompressedFileType.[A-Z_]+\s*->/ AND (language:kotlin)`, I can find nothing. searching instead for `/CompressedFileType.[A-Z_]+/ AND (language:kotlin)`, the *only* usages I see are either it being directly passed to `precompressed()` or just forks of ktor.

I also can't really think of a good name for the interface/child (data) class that doesn't just result in things being more confusing.

if I did add an interface, I'd probably want to add some additional things in order to support [Compressed dictionary transport](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Compression_dictionary_transport):
- an `addHeaders()` method
- a `findResource(): Resource?` method

though I feel if someone really wants to add custom compression formats, then they're probably better off just vendoring all this code.
That's what I had done for my improvements, prior to making this PR.

solonovamax:
While reading through RFC 9110, I a few cases that ktor does not currently comply with:

### [12.5.3. Accept-Encoding](https://httpwg.org/specs/rfc9110.html#field.accept-encoding)

> A server tests whether a content coding for a given representation is acceptable using these rules:
>
> 1. If no Accept-Encoding header field is in the request, any content coding is considered acceptable by the user agent.
> 2. If the representation has no content coding, then it is acceptable by default unless specifically excluded by the Accept-Encoding header field stating either "identity;q=0" or "*;q=0" without a more specific entry for "identity".
> 3. If the representation's content coding is one of the content codings listed in the Accept-Encoding field value, then it is acceptable unless it is accompanied by a qvalue of 0. (As defined in [Section 12.4.2](https://httpwg.org/specs/rfc9110.html#quality.values), a qvalue of 0 means "not acceptable".)

currently, if a client makes a request with `Accept-Encoding` which includes `identity;q=0` or `*;q=0` and another encoding is not matched, then it *will* respond with the identity encoding, which does not comply with RFC 9110.
Further, if a request is made with something like `gzip;q=0`, then the server might respond with gzip-encoded content.

### [8.4.1. Content Codings](https://httpwg.org/specs/rfc9110.html#content.codings)

> All content codings are case-insensitive and ought to be registered within the "HTTP Content Coding Registry", as described in [Section 16.6](https://httpwg.org/specs/rfc9110.html#content.coding.extensibility)

currently, ktor uses case-sensitive matching for the Accept-Encoding header.

solonovamax:
From some profiling, any allocations/cpu time introduced by `bestCompressionFit()` is completely negligible.
Several orders of magnitude more cpu time & allocations are performed in methods like `respond()`. Even just `request.acceptEncodingItems()`, which is passed as a parameter to `bestCompressionFit()` uses an order of magnitude more allocations & cpu time.
Adjusting `bestCompressionFit()` to use less allocations/be faster is an irrelevant micro-optimization.

I have also measured a decrease in cpu time and memory allocations from a quick profile of all the changes as a whole (although I am not 100% sure the decrease in cpu time is due to the changes and not because of other processes consuming cpu, the difference is sufficiently large that I believe it is not caused by other processes)
~~though, there also seems to be some random amount of variance in the memory allocations that I literally cannot explain~~

ktor-server/ktor-server-core/jvm/src/io/ktor/server/http/content/PreCompressed.kt
bjhham: Yeah, we can keep the enum it's fine 👍

bjhham:
> what did you use to benchmark this? was this a specific tool, or are you just writing down the differences you noticed?

It's one of our checks here https://ktor.teamcity.com/buildConfiguration/Ktor_AllocationTests/435986

Which runs the tests in here for measuring memory allocations https://github.com/ktorio/ktor-benchmarks/tree/main/allocation-benchmark

> While reading through RFC 9110, I a few cases that ktor does not currently comply with...

I'll log a separate ticket for these.

ktor-server/ktor-server-core/jvm/src/io/ktor/server/http/content/PreCompressed.kt
bjhham: A boolean flag works.

We can use something like:
```
preCompressed(CompressedFileType.BROTLI, alwaysUseSmallest = true)
```

solonovamax:
> > what did you use to benchmark this? was this a specific tool, or are you just writing down the differences you noticed?
> 
> It's one of our checks here https://ktor.teamcity.com/buildConfiguration/Ktor_AllocationTests/435986

hmm, I can't view that, it gives me a 403 when I try to open it.

> > While reading through RFC 9110, I a few cases that ktor does not currently comply with...
> 
> I'll log a separate ticket for these.

I did also add a commit (699bb3a7fbb520249ca8cd02233d94846cbba316) which (partially, see commit description) fixes this

bjhham:
> hmm, I can't view that, it gives me a 403 when I try to open it.

Turns out we've never enabled guest access to our benchmarks or test builds!  You should be able to view it now.

solonovamax:
This should all more or less be ready for a merge now, I don't think there's anything left to do.

bjhham (approved):

PR #5557, opened 2026-04-28
Fix auto-reload leaving server stopped after failed reload

### LINKED ISSUE #4717: Auto-reloading breaks server if server request is made too quickly after code change
**Ktor version/setup**
Ktor Project Generator: Ktor 3.1.1, Netty, Gradle Kotlin, Routing plugin

**Issue**
I am trying to set up a local environment for developing a Ktor application with auto-reloading. The auto-reloading has been working, but it seems like if a code change is made and then a request to the server is sent immediately after before the compiler is finished, the server breaks.

Usually if the auto-reloading works properly, the following lines are outputted:
```
INFO Application - Changes in application detected.
INFO Application - Application auto-reloaded in 0.019 seconds.
```
However if the issue is reproduced then only the first line is outputted and the server does not respond to the request.

**Steps to reproduce**
New Ktor project from Ktor Project Generator with the settings above.
Enable auto-reload following these steps: https://ktor.io/docs/server-auto-reload.html
Run `./gradlew build --continuous` and `./gradlew run`
Make a change in code to the router
Immediately send a request to the server (you might have to send 2-3 in a row)

fru1tworld:
**Subsystem**
Server core, auto-reload

**Motivation**
Fixes #4717. When auto-reload fires while a Gradle compile is still in progress, `createApplication()` may throw — leaving `applicationInstance` as `null`, so every subsequent request fails with `EmbeddedServer was stopped` until the JVM is restarted.

**Solution**
- Switch reload to a create-then-swap pattern in a new private `reloadApplication()` helper: create the new application first, dispose the previous one only on success, restore the previous watch keys on failure.
- `reload()` rethrows so explicit callers still see the error; the auto-reload path in `currentApplication()` catches and logs, keeping the previously loaded application alive.

### DIFF
--- a/ktor-server/ktor-server-core/jvm/src/io/ktor/server/engine/EmbeddedServerJvm.kt
+++ b/ktor-server/ktor-server-core/jvm/src/io/ktor/server/engine/EmbeddedServerJvm.kt
@@ -91,16 +91,16 @@ actual constructor(
     }
 
     /**
-     * Reload application: destroy it first and then create again
+     * Reload application: build a new instance first, then dispose of the previous one.
+     *
+     * If the new application cannot be created (for example, because the user code throws during
+     * module loading), the previous instance is preserved and the failure is rethrown.
      *
      * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.server.engine.EmbeddedServer.reload)
      */
     public fun reload() {
         applicationInstanceLock.write {
-            destroyApplication()
-            val (application, classLoader) = createApplication()
-            applicationInstance = application
-            applicationClassLoader = classLoader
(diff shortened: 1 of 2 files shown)


bjhham (commented):
Thanks for the fix!

It looks like one of our new tests has started failing independently.  I'll look into ignoring this in main for now.

bjhham (approved):

jg-hot:
This PR appears to cause breaking changes in the `koin-ktor` plugin. See details [here](https://github.com/InsertKoinIO/koin/issues/2461).

bjhham:
Thanks for the heads up @jg-hot, sounds like a regression.  I'll look into it.

PR #5560, opened 2026-04-28
Add reproduction for 3.4.3 Netty request handler executing on worker group

dpolugic:
**Subsystem**
Server

**Motivation**
3.4.3 introduced a bug where request handlers execute on the worker group instead of the call group when using Netty.

I've created an issue here: [KTOR-9542](https://youtrack.jetbrains.com/issue/KTOR-9542/Netty-server-runs-requests-on-worker-group-after-3.4.3)

I'd guess that [KTOR-9531](https://youtrack.jetbrains.com/issue/KTOR-9531/Netty-server-intermittently-drops-requests-after-upgrading-to-3.4.3) is at least somewhat related.

Appears to have been introduced in https://github.com/ktorio/ktor/pull/5421

**Solution**
This PR isn't intended to be merged, only to demonstrate the issue.


dpolugic:
Merged and fixed in https://github.com/ktorio/ktor/pull/5602 🫶

PR #5561, opened 2026-04-28
Fix flaky tests

osipxd:
**Subsystem**
Test Infrastructure

**Motivation**
Make tests a bit more stable

**Solution**
- `HttpTimeoutTest.testGetAfterTimeout`: Remove streaming and expect `HttpRequestTimeoutException` instead of `CancellationException`. We have separate tests for streaming get and for this test it is not what we're testing.
- `RateLimitTest.testRemovesUnusedRateLimitersOnRefill`: Remove intermittent delays causing status `OK` instead of `TooManyRequests`
- `RunTestWithDataTest`: Increase test duration to give a twice larger window to cover OS scheduling overhead.
- `WebSocketTest.testAuthenticationWithValidRefreshToken`: This one was legitimately failing on Curl because of [KTOR-9539](https://youtrack.jetbrains.com/issue/KTOR-9539)
- `NettySpecificTest`: Fix Netty call hang when channel becomes inactive before response is sent ([KTOR-9536](https://youtrack.jetbrains.com/issue/KTOR-9536)).
- `WebSocketTest.testHttpRequestAfterWebSocketClose`: Fix ABA pointer-reuse bug in Curl WebSocket cancellation ([KTOR-9540](https://youtrack.jetbrains.com/issue/KTOR-9540)).
- `DependencyInjectionTest.async support`: Replace timing-based delays with a `Job` barrier to make resolution order deterministic without relying on real-time scheduling.

_It's better to review commit-by-commit_

### DIFF
--- a/ktor-client/ktor-client-curl/desktop/src/io/ktor/client/engine/curl/CurlClientEngine.kt
+++ b/ktor-client/ktor-client-curl/desktop/src/io/ktor/client/engine/curl/CurlClientEngine.kt
@@ -48,7 +48,7 @@ internal class CurlClientEngine(
 
             val status = HttpStatusCode.fromValue(status)
 
-            val responseBody: Any = if (data.isUpgradeRequest()) {
+            val responseBody: Any = if (data.isUpgradeRequest() && status == HttpStatusCode.SwitchingProtocols) {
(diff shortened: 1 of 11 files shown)


bjhham:
Everything passes 🥲

bjhham (approved):
Love to see it 💖

Issue #1030, opened 2019-03-20
Netty engine still print annoying exceptions


kobe2000:
### Ktor Version
1.1.3
### Ktor Engine Used(client or server and name)
Netty - Firefox
### JVM Version, Operating System and Relevant Context
Windows10 JDK8
### Feedback

When I use Firefox to access web site build by ktor, exceptions are often printed in server logs as following:
```
2019-03-20 14:35:59.083 [nettyWorkerPool-3-2] DEBUG Application - I/O operation failed
java.io.IOException: 你的主机中的软件中止了一个已建立的连接。
	at sun.nio.ch.SocketDispatcher.read0(Native Method)
	at sun.nio.ch.SocketDispatcher.read(SocketDispatcher.java:43)
	at sun.nio.ch.IOUtil.readIntoNativeBuffer(IOUtil.java:223)
	at sun.nio.ch.IOUtil.read(IOUtil.java:192)
	at sun.nio.ch.SocketChannelImpl.read(SocketChannelImpl.java:380)
	at io.netty.buffer.PooledUnsafeDirectByteBuf.setBytes(PooledUnsafeDirectByteBuf.java:288)
	at io.netty.buffer.AbstractByteBuf.writeBytes(AbstractByteBuf.java:1108)
	at io.netty.channel.socket.nio.NioSocketChannel.doReadBytes(NioSocketChannel.java:345)
	at io.netty.channel.nio.AbstractNioByteChannel$NioByteUnsafe.read(AbstractNioByteChannel.java:148)
	at io.netty.channel.nio.NioEventLoop.processSelectedKey(NioEventLoop.java:645)
	at io.netty.channel.nio.NioEventLoop.processSelectedKeysOptimized(NioEventLoop.java:580)
	at io.netty.channel.nio.NioEventLoop.processSelectedKeys(NioEventLoop.java:497)
	at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:459)
	at io.netty.util.concurrent.SingleThreadEventExecutor$5.run(SingleThreadEventExecutor.java:884)
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
	at java.lang.Thread.run(Thread.java:748)
```

Akhu:
Got the same issue, here in mac OS or Docker environment with JDK 1.8. 
[...]

PR #5563, opened 2026-04-29
KTOR-646 Log client-disconnect IOException at TRACE in Netty HTTP/1 handler

fru1tworld:
**Subsystem**
Server, Netty

**Motivation**
[KTOR-646](https://youtrack.jetbrains.com/issue/KTOR-646) / #1030 — `NettyHttp1Handler.exceptionCaught` logs every `IOException` at DEBUG, so routine client disconnects ("Connection reset by peer") show up as noise in default DEBUG logs.

**Solution**
Lower the log level to TRACE. Information stays available for deep debugging without polluting DEBUG output.

### DIFF
--- a/ktor-server/ktor-server-netty/jvm/test/io/ktor/tests/server/netty/NettySpecificTest.kt
+++ b/ktor-server/ktor-server-netty/jvm/test/io/ktor/tests/server/netty/NettySpecificTest.kt
@@ -4,6 +4,10 @@
 
 package io.ktor.tests.server.netty
 
+import ch.qos.logback.classic.Level
+import ch.qos.logback.classic.Logger
+import ch.qos.logback.classic.spi.ILoggingEvent
+import ch.qos.logback.core.read.ListAppender
 import io.ktor.client.*
 import io.ktor.client.engine.cio.*
 import io.ktor.client.plugins.*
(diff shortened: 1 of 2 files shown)


ktor-server/ktor-server-netty/jvm/test/io/ktor/tests/server/netty/NettySpecificTest.kt
```diff
@@ -171,6 +181,53 @@ class NettySpecificTest {
         }
     }
 
+    @Test
+    fun `KTOR-646 client-disconnect IOException is logged at TRACE not DEBUG`() {
```
bjhham: Best to simplify the test name here to describe the high-level behavior like `client disconnect is logged` and leave the level as an implementation detail.  We generally avoid referencing YouTrack issue numbers in fixes too.

bjhham (changes requested):
Thanks for the fix.  One small comment wrt naming

fru1tworld:
Thanks for the review! Addressed both points in d6162313b:

- Renamed the test to `client disconnect is logged at trace level` and dropped the YouTrack number from both the test name and the logger name.
- Removed the redundant `detachAppender` call before assertions (kept the one in `finally`).

bjhham (approved):
Looks good, thanks!

PR #5566, opened 2026-04-30
KTOR-9544 Apache: Attach response body channel to callContext

osipxd:
**Subsystem**
ktor-client-apache5

**Motivation**
[KTOR-9544](https://youtrack.jetbrains.com/issue/KTOR-9544) Apache: body channel not cancelled when caller scope is cancelled

**Solution**
Attach the response body channel to `callContext`'s job directly (matching CIO/OkHttp/Apache pattern) so caller-scope cancellation propagates the exact cause to `closedCause`. Previously the channel was attached to an intermediate `consumerJob`, causing cause wrapping that made `closedCause` unreachable.

Also silently discard data in `consume()` when the channel is closed instead of throwing. Throwing (even `IOException` per the interface contract) triggers Apache's error-recovery path mid-body-stream, which either corrupts connection pool state or causes a `RejectedExecutionException` on an already-shutdown scheduler.


osipxd:
Superseded by #5575

PR #5565, opened 2026-04-29
Typed Auth DSL

zibet27:
**Subsystem**
Server

**Motivation**
(https://github.com/ktorio/ktor-klip/pull/6)

**Solution**
The KLIP document was aligned with the final implementation

PR #5567, opened 2026-04-30
Add typed OAuth with session support

zibet27:
**Subsystem**
Server Auth

**Motivation**
https://github.com/ktorio/ktor-klip/blob/zibet27/auth-3.5/proposals/0006-auth-3.5.md

**Why `oauthWithSession`**

The existing OAuth provider is useful for completing the OAuth callback and obtaining an `OAuthAccessTokenResponse`, but it is easy to read it as a provider that can protect application routes directly.

That is misleading for typical web login flows: OAuth authenticates only the callback request. It does not persist the login state, create an application principal, or protect later requests. Applications still need to map the token response to their own session/principal and then use session authentication for protected routes.

`oauthWithSession` makes this pattern explicit:

- OAuth handles the redirect/callback flow.
- The callback maps the OAuth token response to an application session.
- Protected routes authenticate via that session and expose a typed principal.

This keeps the OAuth provider focused on OAuth, while giving users a typed, end-to-end login flow that matches how OAuth is commonly used in server-side applications.

This is still a question if we want to just provide typed DSL for `oauth` or `oauthWithSession`, as shown in this PR


osipxd:
Could you pull the latest main? Failing tests seem to be fixed in the main branch.

ktor-server/ktor-server-plugins/ktor-server-auth/common/src/io/ktor/server/auth/typesafe/OAuthWithSession.kt
```diff
@@ -0,0 +1,359 @@
...
+ *
+ * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.server.auth.typesafe.oauthWithSession)
+ *
+ * @param name name that identifies the OAuth-with-session scheme.
+ * @param principalFactory maps a session value to the route principal.
+ * @param configure configures OAuth, session storage, and session creation.
+ * @return a compound scheme for [oauthCallback] and [authenticateWith].
+ */
+@ExperimentalKtorApi
+public inline fun <reified S : Any, reified P : Any> oauthWithSession(
```
e5l: the seprate method looks adhoc for me, could we make it more unified and have session support inside the regular one?

@nomisRev, what do you think?

e5l (changes requested):

ktor-server/ktor-server-plugins/ktor-server-auth/common/src/io/ktor/server/auth/typesafe/OAuthWithSession.kt
zibet27: We can
I didn't include the typed `oauth` scheme in the https://github.com/ktorio/ktor/pull/5565
`oauthWithSession` is named like this to make clear that session integration is provided by default

ktor-server/ktor-server-plugins/ktor-server-auth/common/test/io/ktor/tests/auth/typesafe/OAuthFlowTest.kt
```diff
@@ -0,0 +1,197 @@
...
+                clientId = "test-client-id",
+                clientSecret = "test-client-secret",
+                requestMethod = HttpMethod.Post,
+            )
+        }.withSessions<OAuthPrincipal, OAuthSession> {
+            storage {
+                cookie(scheme = it)
+            }
+            sessionCreator = { token ->
+                val oauth2 = token as OAuthAccessTokenResponse.OAuth2
```
nomisRev: Why is this cast needed here? We should have typed access the `accessToken` and the other information from the successfully logged in principal, right?

zibet27: Yeah, but the type of token is `OAuthAccessTokenResponse`, not oauth2

zibet27: Actually, this is a bug
I've reduced the token type to `OAuthAccessTokenResponse.OAuth2`, but the session creator gets raw `OAuthAccessTokenResponse` (I forgot it or lost it at some point)
Good catch!

nomisRev:
@e5l I am wondering if we might need to rethink some of these APIs a little bit because it still won't be straightforward for a developer to allow secure access from a server, there are a lot of different moving parts here. There was a secondary goal besides introducing "typed authentication" which was to simplify setting up secure OAuth with session.

Even though the PR is small it's kind-of tricky to wrap my head around but user needs to separately setup:
 - login route
 - callback route (not huge fan of the API/signature in this PR)
 - oauthflow.withSessions { }
    - Verify sessions with JWK 
 
And/or is the goal to tackle this in https://github.com/ktorio/ktor-klip/pull/7? In that case this can still be considered low-level, but we might want to deviate more from the current (old) design. Perhaps we should consider skipping this all together, and jumping straight to KLIP 7 to provide a high-level typed DSL? 🤔

WDYT @zibet27?

zibet27:
> * callback route (not huge fan of the API/signature in this PR)

Also, not 100% sure about this one
The idea here is to show that the OAuth flow is not a "proper" authentication scheme and should be used only for such callbacks

> And/or is the goal to tackle this in [ktorio/ktor-klip#7](https://github.com/ktorio/ktor-klip/pull/7)? In that case, this can still be considered low-level

Yeah, this PR is a low-level API for easier migration from the existing API and a base for OIDC

> but we might want to deviate more from the current (old) design

Personally, I am open to more changes. This is a new API anyway
Just not sure which direction you want to move

> Perhaps we should consider skipping this all together, and jumping straight to KLIP 7 to provide a high-level typed DSL?

So you mean, not to add Typesafe OAuth DSL and only keep OIDC?
We could omit this atm and see if people would really need it, though it feels too inconsistent for me

ktor-server/ktor-server-plugins/ktor-server-auth/common/src/io/ktor/server/auth/typesafe/OAuthFlow.kt
```diff
@@ -0,0 +1,409 @@
...
+ *
+ * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.server.auth.typesafe.oauthCallback)
+ *
+ * @param flow OAuth 2.0 flow.
+ * @param path route path used for both the initial OAuth redirect and the callback.
+ * @param onSuccess handler invoked after OAuth token exchange succeeds.
+ * @return the route that contains the OAuth callback.
+ */
+@ExperimentalKtorApi
+public fun Route.oauthCallback(
```
nomisRev: Maybe this could be slightly nicer signature?

```kotlin
context(route: Route)
fun OAuth2Flow.oauthCallback(path: String, onSuccess: ...): Route
```

nomisRev:
Okay, I gave this some more thought and it might make sense to have this intermediate APIs. Perhaps for people that have a more complex setup with the current low-level APIs that will not be able to easily migrate to the higher-level OIDC support. It might indeed feel inconsistent, and to be honest I'm personally open to removing or deprecating the "old" code although that'd be a **huge** breaking change. Maybe deprecate in Ktor 4 and in 10 years remove in Ktor 5.0 😅

I think what bothers me a little bit here is that the routes are still "implicitly" linked, and without them it doesn't work at all. So you have the OAuth2 configured but it's incomplete. So maybe we just need to:

 - Combine `urlProvider` with the callback handler somehow?
   - There are some annoyances with `urlProvider` we might also want to solve like [this sanitation in foodies](https://github.com/nomisRev/foodies/blob/1e2da649af6e5036016499a7b4802bc295d19e64/webapp/src/main/kotlin/io/ktor/foodies/webapp/security/Security.kt#L237).
 - Introduce `login` path to the configuration rather than using the "empty handler" approach that we now have.
 
This unifies the OAuth setup a little bit better in my opinion but I think it breaks the current `oauthCallback` mechanism with -and without sessions but I'm not sure we're missing out of a lot there or maybe I overlooked something?

zibet27:
@coderabbitai review

nomisRev (approved):
I like this new API a lot more. Great work @zibet27 👏 👏

zibet27:
> I like this new API a lot more. Great work Pantus Oleh (@zibet27) 👏 👏

Great! Thank you!
I'll merge it into [typed-auth-dsl](https://github.com/ktorio/ktor/tree/typed-auth-dsl) (you still can leave comments there)

e5l (approved):
lgtm

Issue #1924, opened 2020-06-04
Support async DNS resolve in Client


e5l:

oleg-larshin:
Please check the <a href="https://youtrack.jetbrains.com/issue/KTOR-351">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

bjhham:
Provided by @fru1tworld

Issue #1678, opened 2020-02-25
CIO DNS resolver with timeout!!


xfront:
**Subsystem**
CIO http client

**Is your feature request related to a problem? Please describe.**
request dns resolve timeout

**Describe the solution you'd like**
custom implement aysnc dns resolver!!

**Motivation to include to ktor**
block too long on dns resolve because the java dns impl without timeout!

oleg-larshin:
Please check the <a href="https://youtrack.jetbrains.com/issue/KTOR-462">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

bjhham:
A new async DNS resolver has been graciously provided by @fru1tworld in #5577

PR #5570, opened 2026-05-01
KTOR-9554  Add `dns` config to OkHttp engine

fru1tworld:
**Subsystem**
Client / OkHttp engine

**Motivation**
[KTOR-462](https://youtrack.jetbrains.com/issue/KTOR-462) / #1678 / #1924

OkHttp exposes a `Dns` interface but `OkHttpConfig` does not surface it as a first-class option. Users currently have to drop down to `engine { config { dns(...) } }` to inject a custom resolver (DNS-over-HTTPS, host overrides for tests, etc.).

**Solution**
Add `OkHttpConfig.dns: Dns?` and apply it in `OkHttpEngine.createOkHttpClient` right after `proxy`, mirroring the existing pattern. Defaults to `null` → `Dns.SYSTEM` is used as before.

```kotlin
HttpClient(OkHttp) {
    engine {
        dns = Dns { hostname -> listOf(InetAddress.getByName("127.0.0.1")) }
    }
}
```

Scoped to the OkHttp engine; the broader unified resolver abstraction (KTOR-462) is out of scope here. The same per-engine pattern can follow for Apache5 in a separate PR.

### DIFF
--- a/ktor-client/ktor-client-okhttp/api/ktor-client-okhttp.api
+++ b/ktor-client/ktor-client-okhttp/api/ktor-client-okhttp.api
@@ -12,10 +12,12 @@ public final class io/ktor/client/engine/okhttp/OkHttpConfig : io/ktor/client/en
 	public final fun addNetworkInterceptor (Lokhttp3/Interceptor;)V
 	public final fun config (Lkotlin/jvm/functions/Function1;)V
 	public final fun getClientCacheSize ()I
+	public final fun getDns ()Lokhttp3/Dns;
 	public final fun getDuplexStreamingEnabled ()Z
 	public final fun getPreconfigured ()Lokhttp3/OkHttpClient;
 	public final fun getWebSocketFactory ()Lokhttp3/WebSocket$Factory;
 	public final fun setClientCacheSize (I)V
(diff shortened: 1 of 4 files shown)


bjhham (approved):
Thanks for the contribution!

I wouldn't say it covers the ticket, however.  The ticket refers to the address resolution in `ktor-network` which is used in the CIO engine.  I'll create a new ticket with reference to supporting DNS configuration in `OkHttp` that better reflects the work done here.

bjhham (approved):
Thanks for the contribution!

I wouldn't say it covers the ticket, however.  The ticket refers to the address resolution in `ktor-network` which is used in the CIO engine.  I'll create a new ticket with reference to supporting DNS configuration in `OkHttp` that better reflects the work done here.

bjhham:
Here's the new ticket to reflect the update to the OkHttp engine:

[KTOR-9554](https://youtrack.jetbrains.com/issue/KTOR-9554) DNS configuration for OkHttp client engine

fru1tworld:
Thanks for the review!
Updated to reference KTOR-9554 in the commit/title and addressed the resource-leak feedback

PR #5572, opened 2026-05-03
#5518 Add eventProvider to Heartbeat for dynamic SSE events

### LINKED ISSUE #5518: Can Ktor SSE heartbeat support a function to get the value instead of a fixed one? I want to include a timestamp when sending heartbeats.
Currently, the SSE heartbeat event is a fixed value set at configuration time. I need to include a timestamp on each heartbeat, so the heartbeat event should be generated by a suspend function or lambda at runtime, not a static value. Please allow something like:

Thank you for your understanding and cooperation! 🙏

fru1tworld:
**Subsystem**
Server, ktor-server-sse

**Motivation**
Fixes #5518. `Heartbeat.event` is captured at config time, so every tick emits the same payload — no way to embed a fresh timestamp or counter without abandoning the built-in `heartbeat()` helper.

**Solution**
Add `Heartbeat.eventProvider: (suspend () -> ServerSentEvent)?`. When set, the heartbeat loop invokes it on each tick; otherwise falls back to `event` (existing behavior).

```kotlin
heartbeat {
    period = 30.seconds
    eventProvider = { ServerSentEvent(data = "ts=${Clock.System.now()}") }
}
```


ktor-server/ktor-server-plugins/ktor-server-sse/common/test/io/ktor/server/sse/ServerSentEventsTest.kt
```diff
@@ -335,6 +335,89 @@ class ServerSentEventsTest {
...
+    fun testHeartbeatEventProviderTakesPrecedenceOverEvent() = testApplication {
+        install(SSE)
+        routing {
+            sse {
+                heartbeat {
+                    period = 10.milliseconds
+                    event = ServerSentEvent(data = "static")
+                    eventProvider = {
+                        ServerSentEvent(data = "dynamic")
+                    }
```
bjhham, line 391: It's a little confusing, but I can see why it's necessary to preserve backwards compatibility.

bjhham (approved):
Thanks for the contribution 👍 

Nice to see some documentation filled in too.

bjhham:
Oh, looks like there's a conflict with main now.  Would you mind resolving this?

fru1tworld:
Thanks for the review — fixed!

PR #5575, opened 2026-05-04
KTOR-9546 Fix streaming responses cancellation

osipxd:
**Subsystem**
`ktor-io`, `ktor-client-core`, `ktor-client-apache5`

**Motivation**
[KTOR-9546](https://youtrack.jetbrains.com/issue/KTOR-9546) HttpClient: cancelling ByteReadChannel body does not propagate to engine
[KTOR-9544](https://youtrack.jetbrains.com/issue/KTOR-9544) Apache: body channel not cancelled when caller scope is cancelled

**Solution**
When a caller cancels a streaming `ByteReadChannel` response body, the cancellation did not propagate back to the engine's active network transfer. The root cause was in `DefaultTransform`: the `ByteReadChannel` case wraps the engine body in a `writer { body.copyTo(channel) }` coroutine and returns its output channel to the caller. Cancelling the returned channel had no effect on body, so `copyTo` stayed blocked in `awaitContent()` — leaving
`responseJobHolder` alive.

Two fixes applied in `DefaultTransform`:

1. `body.attachWriterJob(writerJob)` — when the caller cancels the returned channel, `body` (the engine channel) is cancelled immediately, unblocking
`copyTo`'s `awaitContent()`.
2. `body.rethrowCloseCauseIfNeeded()` after `copyTo` returns — handles the race where `copyTo` exits its while loop normally via `!isClosedForRead` (rather than throwing), meaning the catch block never fires and the cancellation cause would otherwise be silently dropped.

New `ByteReadChannel.attachWriterJob` API introduced to support this — wires cancel propagation from a writer's output channel back to a source ByteReadChannel

Also fixed:
- `testBodyChannelCancelledWhenCallerScopeIsCancelled` is now enabled for Apache5 (as KTOR-9544 also fixed with this fix)
- `testDownloadStreamChannelWithCancel` is un-ignored and given a proper assertion
- `testGetWithCancellation` tightened to catch the regression: checks cleanup within 500ms instead of relying on the 10s default
- `testErrorOnResponseCoroutine` re-enabled for CIO

### DIFF
--- a/ktor-client/ktor-client-apache5/jvm/src/io/ktor/client/engine/apache5/ApacheResponseConsumer.kt
+++ b/ktor-client/ktor-client-apache5/jvm/src/io/ktor/client/engine/apache5/ApacheResponseConsumer.kt
@@ -77,17 +77,16 @@ internal class BasicResponseConsumer(private val dataConsumer: ApacheResponseCon
     }
 }
 
-@OptIn(InternalCoroutinesApi::class)
 internal class ApacheResponseConsumer(
     parentContext: CoroutineContext,
     private val requestData: HttpRequestData
 ) : AsyncEntityConsumer<Unit>, CoroutineScope {
 
-    private val consumerJob = Job(parentContext[Job])
+    private val consumerJob = Job(parentContext.job)
     override val coroutineContext: CoroutineContext = parentContext + consumerJob
 
-    private val channel = ByteChannel().also {
-        it.attachJob(consumerJob)
+    private val channel = ByteChannel().apply {
+        attachJob(consumerJob)
     }
 
     @Volatile
@@ -101,12 +100,6 @@ internal class ApacheResponseConsumer(
     private val capacity = atomic(channel.availableForWrite)
 
     init {
-        coroutineContext[Job]?.invokeOnCompletion(onCancelling = true) { cause ->
-            if (cause != null) {
-                responseChannel.cancel(cause)
-            }
-        }
-
         launch(CoroutineName("apache-response-consumer")) {
             for (message in messagesQueue) {
                 when (message) {
@@ -140,9 +133,11 @@ internal class ApacheResponseConsumer(
     }
 
     override fun consume(src: ByteBuffer) {
-        if (channel.isClosedForWrite) {
-            channel.closedCause?.let { throw it }
-        }
+        // Silently discard when the channel is closed (e.g. caller scope was cancelled).
+        // Throwing here (even IOException per the interface contract) causes Apache to invoke its
+        // error-recovery path mid-body-stream, which either corrupts the connection pool state or
+        // triggers a retry on an already-shutdown scheduler (RejectedExecutionException).
+        if (channel.isClosedForWrite) return
         messagesQueue.trySend(src.copy())
     }
 
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/DefaultTransform.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/DefaultTransform.kt
@@ -5,7 +5,7 @@
 package io.ktor.client.plugins
 
 import io.ktor.client.*
-import io.ktor.client.call.checkContentLength
+import io.ktor.client.call.*
 import io.ktor.client.request.*
 import io.ktor.client.statement.*
 import io.ktor.http.*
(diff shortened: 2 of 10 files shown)


ktor-client/ktor-client-apache5/jvm/src/io/ktor/client/engine/apache5/ApacheResponseConsumer.kt
```diff
@@ -101,12 +100,6 @@ internal class ApacheResponseConsumer(
     private val capacity = atomic(channel.availableForWrite)
 
     init {
-        coroutineContext[Job]?.invokeOnCompletion(onCancelling = true) { cause ->
-            if (cause != null) {
-                responseChannel.cancel(cause)
-            }
-        }
```
osipxd, line 108: `attachJob` does exactly the same with only one difference: `onCancelling = false`.
So I decided to remove this `cancel` call duplicate and check if any test fails.

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/DefaultTransform.kt
```diff
@@ -96,10 +98,11 @@ public fun HttpClient.defaultTransformers() {
                 // the response job could be already completed, so the job holder
                 // could be canceled immediately, but it doesn't matter
                 // since the copying job is running under the client job
-                val responseJobHolder = Job(response.coroutineContext[Job])
+                val responseJobHolder = Job(response.coroutineContext.job)
                 val channel: ByteReadChannel = writer(this@defaultTransformers.coroutineContext) {
                     try {
                         body.copyTo(channel, limit = Long.MAX_VALUE)
+                        body.rethrowCloseCauseIfNeeded()
```
osipxd, line 105: The question is: should we rethrow the cause here or from `copyTo`?
We have only two functions that rethrow a cause: `readRemaining`, `readTo`. All other functions doesn't rethrow the cause. Is it intended behavior?

ktor-io/common/src/io/ktor/utils/io/ByteChannelUtils.kt
```diff
@@ -27,3 +28,12 @@ public fun ByteChannel.attachJob(job: Job) {
...
 }
+
+/**
+ * Ensures that when the [WriterJob]'s output channel is canceled, this [ByteReadChannel] is also canceled.
+ */
+public fun ByteReadChannel.attachWriterJob(writerJob: WriterJob) {
+    (writerJob.channel as? ByteChannel)?.invokeOnClose { cause ->
+        if (cause != null) cancel(cause)
+    }
+}
```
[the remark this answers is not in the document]
osipxd: Added `@InternalAPI` to it

ktor-io/common/src/io/ktor/utils/io/ByteChannelUtils.kt
```diff
@@ -4,18 +4,19 @@
...
  * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.utils.io.attachJob)
  */
+@OptIn(InternalCoroutinesApi::class)
 public fun ByteChannel.attachJob(job: Job) {
-    job.invokeOnCompletion {
-        if (it != null) {
-            cancel(it)
-        }
+    job.invokeOnCompletion(onCancelling = true) { cause ->
+        if (cause != null) cancel(cause as? CancellationException ?: CancellationException(cause.message, cause))
```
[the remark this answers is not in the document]
osipxd: Removed this

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/DefaultTransform.kt
bjhham: I think actually this function should be made private to `ByteChannel` and just be called in `awaitContent()`.  It's already called upon entering the function, but there's no cause check when exiting, so if the writer dies then the cause never gets propagated.

I would say we can leave the call here, and tackle the aforementioned refactor in a separate ticket since it's even more central than this 😅

bjhham (approved):
Looks good, well investigated 👍

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/DefaultTransform.kt
osipxd: Created one: [KTOR-9561](https://youtrack.jetbrains.com/issue/KTOR-9561) ByteReadChannel.copyTo does not propagate source closedCause on normal exit

amccague:
I'm suspecting this is the cause of a regression with the OpenTelemetry instrumentation, I've filed here: https://github.com/open-telemetry/opentelemetry-java-instrumentation/issues/18779

Would you agree? I think the fix _should_ be in the OTEL instrumentation but pointing it out here in case you disagree.

osipxd:
> I'm suspecting this is the cause of a regression with the OpenTelemetry instrumentation

Hm.. interesting. However I think this PR is not related to that regression, as the line `Job(response.coroutineContext.job)` was already there, this PR just changed it a bit.

My main suspect is https://github.com/ktorio/ktor/pull/5512 which has been reverted in https://github.com/ktorio/ktor/pull/5661 and the fix is going to be included in 3.5.1.

PR #5576, opened 2026-05-04
Fix data race on shared BadRequestPacket Source in CIO server

e5l:
## Summary

`ServerPipeline` defined a file-private `BadRequestPacket: Source` built once and shared across all CIO server connections. The null-message branch of `String?.toBadRequestPacket()` called `BadRequestPacket.buffer.copy()` from per-connection coroutines concurrently. `Buffer.copy()` walks the segment chain calling `Segment.sharedCopy()`, which performs an unsynchronized check-then-set of `Segment.copyTracker` — a non-`@Volatile` field whose contract ([`kotlinx-io Segment.kt:108-110`](https://github.com/Kotlin/kotlinx-io/blob/master/core/common/src/Segment.kt#L108-L110)) explicitly forbids concurrent segment mutation.

Replace the shared `Source` with a cached `ByteArray` (mirroring `BadRequestPacketCommon` directly above) plus a per-call `buildBadRequestPacket()` factory that returns a fresh `Buffer`. `toBadRequestPacket()` now calls the factory; no `Source` is shared between coroutines.

- Production fix: ~7 lines in `ServerPipeline.kt`.
- Regression test: `BadRequestPacketSourceTest` races 16 threads × 10,000 iterations on the factory and asserts via reflection that each returned `Buffer` has its own fresh, non-shared segment. Fails on unfixed code (segments share a `copyTracker` after `Buffer.copy()`); passes on the fix.

## Test plan

- [x] `./gradlew :ktor-server-cio:assemble`
- [x] `./gradlew :ktor-server-cio:jvmTest` (incl. new `BadRequestPacketSourceTest`)
- [x] `./gradlew :ktor-server-cio:formatKotlin :ktor-server-cio:lintKotlin`
- [x] `./gradlew :ktor-server-cio:checkKotlinAbi` (no public ABI change; `buildBadRequestPacket` is `internal` for testability)
- [ ] CI to verify other targets (Native/JS unaffected — change is JVM-test-only on the test side; production change is in `common/`)

## Notes

- `internal fun buildBadRequestPacket()` exists primarily to give the regression test a stable handle on the production factory. Could be folded back into `toBadRequestPacket()`'s null branch if preferred — open to feedback.
- No public API surface changed.
- YouTrack issue still to be filed; happy to link `KTOR-NNNNN` once it's up.

### DIFF
--- a/ktor-server/ktor-server-cio/common/src/io/ktor/server/cio/backend/ServerPipeline.kt
+++ b/ktor-server/ktor-server-cio/common/src/io/ktor/server/cio/backend/ServerPipeline.kt
@@ -16,6 +16,7 @@ import kotlinx.coroutines.*
 import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.channels.ReceiveChannel
+import kotlinx.io.Buffer
 import kotlinx.io.IOException
 import kotlinx.io.InternalIoApi
 import kotlinx.io.Source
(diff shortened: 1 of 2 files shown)


bjhham (approved):
lgtm 👍

PR #5571, opened 2026-05-01
KTOR-9559 Add `dnsResolver` config to Apache5 engine

fru1tworld:
**Subsystem**
Client / Apache5 engine

**Motivation**
[KTOR-9559](https://youtrack.jetbrains.com/issue/KTOR-9559) (umbrella: [KTOR-462](https://youtrack.jetbrains.com/issue/KTOR-462)) / #1678 / #1924

Apache HttpClient 5 has a `DnsResolver` interface but `Apache5EngineConfig` does not surface it as a first-class option. Users currently have to drop down to `engine { configureConnectionManager { setDnsResolver(...) } }` to inject a custom resolver.

**Solution**
Add `Apache5EngineConfig.dnsResolver: DnsResolver?` and apply it on the `PoolingAsyncClientConnectionManagerBuilder` before `configureConnectionManager`. Defaults to `null` → Apache's default resolver is used as before.

```kotlin
HttpClient(Apache5) {
    engine {
        dnsResolver = SystemDefaultDnsResolver.INSTANCE
    }
}
```

Follow-up to the OkHttp `dns` config PR (#5570); same per-engine pattern. The unified resolver abstraction (KTOR-462) remains out of scope.

PR #5577, opened 2026-05-05
Add async DNS resolvers to CIO engine

fru1tworld:
**Subsystem**
Client / CIO engine

**Motivation**
[KTOR-462](https://youtrack.jetbrains.com/issue/KTOR-462) / #1678 / #1924

CIO has no first-class DNS hook, and the JVM `InetAddress` path has no timeout — a slow server hangs requests indefinitely. OkHttp (#5570) and Apache5 (#5571) only needed config exposure; #1678 asks CIO for an async resolver with configurable timeout, so this PR adds both.

**Solution**
- `CIOEngineConfig.dnsResolver: (suspend (String) -> List<String>)?` — applied in `Endpoint.connect()` before `InetSocketAddress`. Bypassed for proxy / Unix socket. TLS SNI uses the original hostname.
- `JvmDnsResolver()` (jvm) — `InetAddress.getAllByName()` on `Dispatchers.IO` via `runInterruptible`. Async + cancellable, preserves `/etc/hosts`, mDNS, NSS.
- `CioDnsResolver(server, port, timeout)` (common) — RFC 1035 over UDP via ktor-network. Parallel A + AAAA, IPv6 first per RFC 6724 / 8305. JVM/Native/Posix; UDP unsupported on JS/Wasm.

```kotlin
HttpClient(CIO) {
    engine {
        dnsResolver = JvmDnsResolver()             // OS resolver, async + timeout
        // or CioDnsResolver(server = "1.1.1.1")   // pure UDP, bypass OS
    }
}
```

Defaults to `null` (existing ktor-network path). `CioDnsResolver` is intentionally minimal — A/AAAA only, no CNAME chain, no TC/TCP fallback, no caching/retry.


bjhham (changes requested):
Great work 👌 

Looks like the common code introduced a reliance on pipelining which is currently unsupported, however.  You'll need a new condition to get the DNS tests passing on web  and native it seems:

```
kotlin.IllegalStateException: Pipelining is not supported in native/js/wasm CIO
  at kotlin.Throwable#<init>(Unknown Source)
  at kotlin.Exception#<init>(Unknown Source)
  at kotlin.RuntimeException#<init>(Unknown Source)
  at kotlin.IllegalStateException#<init>(Unknown Source)
  at io.ktor.client.engine.cio.ConnectionPipeline#<init>(/opt/buildAgent/work/8d547b974a7be21f/ktor-client/ktor-client-cio/nonJvm/src/io/ktor/client/engine/cio/ConnectionPipeline.nonJvm.kt:23)
  at io.ktor.client.engine.cio.Endpoint.$createPipelineCOROUTINE$5.invokeSuspend#internal(/opt/buildAgent/work/8d547b974a7be21f/ktor-client/ktor-client-cio/common/src/io/ktor/client/engine/cio/Endpoint.kt:186)
```

fru1tworld:
Thanks for Review! 

Moved the three dnsResolver integration tests to jvm/test in [c17efa2f3](https://github.com/ktorio/ktor/pull/5577/commits/c17efa2f3c227f9282d405af2575a72bc3c1c8ba)

bjhham (approved):

PR #5578, opened 2026-05-05
KTOR-9549: add @JsName to avoid toRaw collision with Vue in Vite bundles

osipxd:
**Subsystem**
Client JS

**Motivation**
[KTOR-9549](https://youtrack.jetbrains.com/issue/KTOR-9549) Kotlin/JS: ktor-ktor-client-core.mjs is incompatible with Vite: toRaw naming conflict and yield* failure in SSR module runner

**Solution**
Rename the problematic function in JS code


bjhham (approved):
Interesting issue 👍

PR #5579, opened 2026-05-06
KTOR-9168 Add regression test for case-insensitive Set-Cookie merging in CIO

fru1tworld:
## Subsystem
Client / CIO

## Motivation
KTOR-9168 (CIO dropping case-different `Set-Cookie` headers) was inadvertently fixed by #5379. Add a wire-level regression test to guard the case-insensitive dedup behavior.

## Solution
Test only — no production code changes.

Closes KTOR-9168


bjhham (approved):
Thanks, looks good 👍

PR #5581, opened 2026-05-06
KTOR-9183 Forward original CancellationException in attachToUserJob

fru1tworld:
**Subsystem**
Client / Core

**Motivation**
[KTOR-9183](https://youtrack.jetbrains.com/issue/KTOR-9183) — `attachToUserJob`'s cleanup handler substitutes a generic `CancellationException` for the user job's cause, dropping `TimeoutCancellationException`. As a result `withTimeout` / `withTimeoutOrNull` wrapping a client call no longer recognises its own timeout.

**Solution**
Forward the original `CancellationException` (including subtypes) when the cause already is one; wrap into `CancellationException(message)` only for non-cancellation throwables.

### DIFF
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/engine/Utils.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/engine/Utils.kt
@@ -9,6 +9,7 @@ import io.ktor.http.*
 import io.ktor.http.content.*
 import io.ktor.util.*
 import io.ktor.utils.io.*
+import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.InternalCoroutinesApi
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.currentCoroutineContext
@@ -105,8 +106,7 @@ internal suspend inline fun attachToUserJob(callJob: Job) {
 
     val cleanupHandler = userJob.invokeOnCompletion(onCancelling = true) { cause ->
         cause ?: return@invokeOnCompletion
-        val cancellation = cause as? kotlinx.coroutines.CancellationException
-            ?: kotlinx.coroutines.CancellationException(cause.message)
+        val cancellation = cause as? CancellationException ?: CancellationException(cause.message)
         callJob.cancel(cancellation)
     }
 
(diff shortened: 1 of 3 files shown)


ktor-client/ktor-client-core/jvm/test/AttachToUserJobTimeoutTest.kt
```diff
@@ -0,0 +1,35 @@
...
+        val callJob = Job()
+        val callJobCause = CompletableDeferred<Throwable?>()
+        callJob.invokeOnCompletion { callJobCause.complete(it) }
+
+        try {
+            withTimeout(50) {
+                attachToUserJob(callJob)
+                awaitCancellation()
+            }
+        } catch (_: TimeoutCancellationException) {
```
osipxd: Let's use `assertThrows<TimeoutCancellationException> { ... }`
This way we could remove the `callJobCause` variable and simplify the test

ktor-client/ktor-client-core/jvm/test/AttachToUserJobTimeoutTest.kt
```diff
@@ -0,0 +1,35 @@
...
+import kotlinx.coroutines.CompletableDeferred
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.TimeoutCancellationException
+import kotlinx.coroutines.awaitCancellation
+import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.withTimeout
+import kotlin.test.Test
+import kotlin.test.assertTrue
+
+class AttachToUserJobTimeoutTest {
```
osipxd: Let's call it a bit less specific `AttachToUserJobTest`

ktor-client/ktor-client-core/common/src/io/ktor/client/engine/Utils.kt
```diff
@@ -105,7 +105,9 @@ internal suspend inline fun attachToUserJob(callJob: Job) {
 
     val cleanupHandler = userJob.invokeOnCompletion(onCancelling = true) { cause ->
         cause ?: return@invokeOnCompletion
-        callJob.cancel(kotlinx.coroutines.CancellationException(cause.message))
+        val cancellation = cause as? kotlinx.coroutines.CancellationException
+            ?: kotlinx.coroutines.CancellationException(cause.message)
```
osipxd: Can the `CancellationException` be imported?

osipxd (changes requested):
Thank you for the PR! Overall it looks good I only have a couple of stylistic comments

fru1tworld:
Thanks for the feedback! Fixed it!

osipxd (approved):

osipxd:
Nicely done! 🏆

PR #5582, opened 2026-05-06
Disable cache redirector

bjhham:
To unblock our builds for now.

### DIFF
--- a/build-settings-logic/src/main/kotlin/ktorsettings.cache-redirector.settings.gradle.kts
+++ b/build-settings-logic/src/main/kotlin/ktorsettings.cache-redirector.settings.gradle.kts
@@ -10,7 +10,7 @@ import java.net.URI
 private val ProviderFactory.isCIRun: Provider<Boolean>
     get() = gradleProperty("teamcity").map { true }
         .orElse(environmentVariable("TEAMCITY_VERSION").map { true })
-        .orElse(providers.isCIRun)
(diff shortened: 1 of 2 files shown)


build-settings-logic/src/main/kotlin/ktorsettings.cache-redirector.settings.gradle.kts
```diff
@@ -10,7 +10,7 @@ import java.net.URI
 private val ProviderFactory.isCIRun: Provider<Boolean>
     get() = gradleProperty("teamcity").map { true }
         .orElse(environmentVariable("TEAMCITY_VERSION").map { true })
-        .orElse(false)
+        .orElse(providers.isCIRun)
```
osipxd: Looks like an accidential change in the wrong place

osipxd:
```suggestion
        .orElse(false)
```

osipxd (changes requested):

build-settings-logic/src/main/kotlin/ktorsettings.cache-redirector.settings.gradle.kts
bjhham: Just not on my game today 🤦

osipxd (approved):

PR #5584, opened 2026-05-06
KTOR-9567 Flaky UnixSockets on Windows: WSAEOPNOTSUPP from bind()

osipxd:
**Subsystem**
ktor-network

**Motivation**
[KTOR-9567](https://youtrack.jetbrains.com/issue/KTOR-9567) Flaky UnixSockets on Windows: WSAEOPNOTSUPP from bind()

**Solution**

The root cause is unclear. However, two bugs were found in the Windows socket setup path in `tcpBind`, either or both of which may cause the intermittent failure:

**1. `nonBlocking()` return value silently discarded**

In `ConnectUtilsNative.kt`, both `tcpConnect` and `tcpBind` call `nonBlocking(descriptor)` without checking the return value:
```kotlin
nonBlocking(descriptor)  // return value ignored
```
If `ioctlsocket(FIONBIO)` fails and leaves the socket in a bad state, `bind()` subsequently returns `WSAEOPNOTSUPP`. The error was invisible because no `.check()` was present.

**2. `ktor_setsockopt` passes wrong bytes on Windows**

`SocketUtilsWindows.kt` had a broken `optval` conversion:
```kotlin
__optval?.reinterpret<ByteVar>()?.toKString()  // wrong
```
The K/N Windows binding for `setsockopt` annotates `optval` as `@CCall.CString String?`, meaning K/N applies UTF-8 encoding before passing the pointer. For an `IntVar` with value `1` (bytes `01 00 00 00`), `toKString()` reads until the null byte and produces a 1-character string. K/N passes a 2-byte buffer (`01 00`) while `optlen` is 4 — a 2-byte buffer overread, causing undefined behavior. The extra bytes read come from adjacent stack memory. This would explain the non-deterministic failure rate.

Applied fixes for both, and will monitor this test behavior after the fixes.

### DIFF
--- a/ktor-network/posix/src/io/ktor/network/sockets/ConnectUtilsNative.kt
+++ b/ktor-network/posix/src/io/ktor/network/sockets/ConnectUtilsNative.kt
@@ -7,9 +7,10 @@ package io.ktor.network.sockets
 import io.ktor.network.selector.*
 import io.ktor.network.util.*
 import io.ktor.utils.io.errors.*
-import kotlinx.cinterop.*
+import kotlinx.cinterop.ExperimentalForeignApi
+import kotlinx.cinterop.convert
 import kotlinx.io.IOException
-import platform.posix.*
+import platform.posix.SOCK_STREAM
 
 private const val DEFAULT_BACKLOG_SIZE = 50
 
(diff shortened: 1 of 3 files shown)


bjhham (approved):
Nicely investigated, tests are green 🙌

PR #5585, opened 2026-05-07
KTOR-9483 Curl: Implement response body backpressure

osipxd:
**Subsystem**
ktor-client-curl

**Motivation**
[KTOR-9483](https://youtrack.jetbrains.com/issue/KTOR-9483) Curl: backpressure implementation is never used
[KTOR-9527](https://youtrack.jetbrains.com/issue/KTOR-9527) Curl: Freeze when receiving large responses

**Solution**
Replace `runBlocking` with non-blocking writes via `writeBuffer`/`flushWriteBuffer`. When the channel's flush buffer reaches `CHANNEL_MAX_SIZE`, return `WRITEFUNC_PAUSE` to pause the specific easy handle. A background coroutine then calls `flush()`, suspending until the consumer drains the buffer, after  which `curl_easy_pause(CURLPAUSE_CONT)` resumes the transfer.

Also adds `ByteChannel.flushNeeded` (`@InternalAPI`) and a slow-consumer download test.

### DIFF
--- a/ktor-client/ktor-client-curl/desktop/src/io/ktor/client/engine/curl/internal/CurlHttpResponseBody.kt
+++ b/ktor-client/ktor-client-curl/desktop/src/io/ktor/client/engine/curl/internal/CurlHttpResponseBody.kt
@@ -7,67 +7,71 @@ package io.ktor.client.engine.curl.internal
 import io.ktor.client.engine.curl.internal.Libcurl.WRITEFUNC_ERROR
 import io.ktor.client.engine.curl.internal.Libcurl.WRITEFUNC_PAUSE
 import io.ktor.utils.io.*
-import kotlinx.atomicfu.atomic
+import io.ktor.utils.io.core.*
 import kotlinx.cinterop.ByteVar
 import kotlinx.cinterop.CPointer
 import kotlinx.cinterop.ExperimentalForeignApi
 import kotlinx.cinterop.convert
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Job
+import kotlinx.coroutines.cancel
 import kotlinx.coroutines.launch
-import kotlinx.coroutines.runBlocking
 import platform.posix.size_t
+import kotlin.concurrent.Volatile
+import kotlin.coroutines.CoroutineContext
 
 internal class CurlHttpResponseBody(
-    private val callContext: Job,
-    private val onUnpause: () -> Unit
-) : CurlResponseBodyData {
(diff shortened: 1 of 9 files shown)


ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/DownloadTest.kt
```diff
@@ -23,6 +27,24 @@ class DownloadTest : ClientLoader() {
         }
     }
 
+    @Test
+    fun testDownloadWithSlowConsumer() = clientTests(timeout = 10.seconds) {
```
osipxd, line 31: This test doesn't check if back pressure is properly implemented, but at least checks that client doesn't fail in slow consumer scenario.

ktor-client/ktor-client-curl/desktop/src/io/ktor/client/engine/curl/internal/CurlHttpResponseBody.kt
```diff
@@ -7,67 +7,70 @@ package io.ktor.client.engine.curl.internal
...
+                throw cause
             } catch (_: Throwable) {
                 // no op, error will be handled on next write on cURL thread
             } finally {
+                paused = false
                 onUnpause()
             }
         }
-
-        return WRITEFUNC_PAUSE
```
osipxd, line 66: It turned out that libcurl replays last chunk for which we returned `WRITEFUNC_PAUSE`, so we should return `WRITEFUNC_PAUSE` before reading data, not after it.

ktor-io/common/src/io/ktor/utils/io/ByteChannel.kt
```diff
@@ -30,6 +31,10 @@ public class ByteChannel(public override val autoFlush: Boolean = false) : ByteR
     @Volatile
     private var flushBufferSize = 0
 
+    @InternalAPI
+    public val flushNeeded: Boolean
```
osipxd: This is the main subject to discuss here. Is it okay to add this field? Are there any other ways to check that flush is needed, I'm not aware of?
Another possible name is `hasFreeSpace` (in line with `awaitFreeSpace`).

bjhham: I think `flushNeeded` is a little confusing and exposes some implementation details.  I'd go with `hasFreeSpace` or `isFull`.

bjhham (approved):

PR #5588, opened 2026-05-08
KTOR-9629 Close RawSourceChannel when Job is canceled

osipxd:
**Subsystem**
ktor-io

**Motivation**
[KTOR-9629](https://youtrack.jetbrains.com/issue/KTOR-9629) RawSourceChannel: coroutine cancellation is not propagated to the RawSource

After merging https://github.com/ktorio/ktor/pull/5575, one engine is still flaky - Android ([test history](https://ktor.teamcity.com/test/-5690675197340020900?currentProjectId=Ktor_ProjectKtorCore), look at test runs with duration longer than 2s).

**Solution**
Android engine uses `InputStream.toByteReadChannel` and the resulting channel is not attached to any `Job`.
I've made it attached to the job `RawSourceChannel` creates, so if `parent: CoroutineContext` has `Job`, cancelling that job would cancel its children and trigger `closeSource`.

### DIFF
--- a/ktor-io/jvm/src/io/ktor/utils/io/jvm/javaio/Reading.kt
+++ b/ktor-io/jvm/src/io/ktor/utils/io/jvm/javaio/Reading.kt
@@ -9,13 +9,12 @@ import io.ktor.utils.io.CloseToken.Companion.wrapCause
 import io.ktor.utils.io.core.*
 import io.ktor.utils.io.pool.*
 import kotlinx.coroutines.*
+import kotlinx.coroutines.CancellationException
 import kotlinx.io.*
 import kotlinx.io.Buffer
-import kotlinx.io.EOFException
-import kotlinx.io.IOException
-import java.io.*
-import java.nio.*
-import kotlin.coroutines.*
+import java.io.InputStream
+import java.nio.ByteBuffer
+import kotlin.coroutines.CoroutineContext
 
 /**
  * Open a channel and launch a coroutine to copy bytes from the input stream to the channel.
@@ -46,7 +45,7 @@ public fun InputStream.toByteReadChannel(
 
 internal class RawSourceChannel(
     private val source: RawSource,
-    private val parent: CoroutineContext
+    parent: CoroutineContext
 ) : ByteReadChannel {
     private var closedToken: CloseToken? = null
     private val buffer = Buffer()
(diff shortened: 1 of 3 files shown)


ktor-io/jvm/src/io/ktor/utils/io/jvm/javaio/Reading.kt
```diff
@@ -63,6 +61,12 @@ internal class RawSourceChannel(
     override val readBuffer: Source
         get() = buffer
 
+    init {
+        job.invokeOnCompletion { cause ->
+            if (cause != null) closeSource(cause)
```
osipxd: Alternatively we could call `attachJob` on the channel in all places where it's necessary.

bjhham: I think this approach should be fine.  Wrapping a source in a `ByteChannel` should be roughly analogous to wrapping an `InputStream` with a reader, and readers generally close the stream when closed, and closing when the job completes should add some extra safety.

Though it looks like it's causing some problems in the tests 😬

osipxd: It was because of wrapping `CancellationException` with `IOException`. It looks like `CancellationException` fits better there, so I've replaced `IOException` with `CancellationException`

ktor-io/jvm/src/io/ktor/utils/io/jvm/javaio/Reading.kt
```diff
@@ -46,7 +45,7 @@ public fun InputStream.toByteReadChannel(
 
 internal class RawSourceChannel(
     private val source: RawSource,
-    private val parent: CoroutineContext
+    parent: CoroutineContext
 ) : ByteReadChannel {
     private var closedToken: CloseToken? = null
```
osipxd, line 50: Should we use `atomic<CloseToken?>(null)` here similar to how `ByteChannel` is implemented?

bjhham: `RawSourceChannel` isn't thread-safe so it's not necessary.

bjhham (approved):

PR #5591, opened 2026-05-08
KTOR-8709 Allow custom WebSocket close codes in OkHttp engine

fru1tworld:
**Subsystem**
Client / OkHttp engine

**Motivation**
[KTOR-8709](https://youtrack.jetbrains.com/issue/KTOR-8709): `OkHttpWebsocketSession.isReserved()` rejected any close code outside the IANA-registered `CloseReason.Codes` enum, silently replacing application-private codes (4000–4999) and library codes (3000–3999) with `INTERNAL_ERROR` (1011). Peers never observed the requested code.

**Solution**
Replace the enum-membership check with the RFC 6455 §7.4.1 reserved range (1004–1006 and 1015–2999). Codes 3000–4999 now flow through unchanged; reserved codes still fall back to the default to avoid OkHttp's `IllegalArgumentException` from `WebSocketProtocol.validateCloseCode`.


bjhham (approved):
Thank you again for another contribution! 🙏

PR #5593, opened 2026-05-08
Update libcurl to 8.20.0 

osipxd:
**Subsystem**
ktor-client-curl

**Motivation**
Build new version of libcurl and fix [KTOR-9460](https://youtrack.jetbrains.com/issue/KTOR-9460) Curl: Linux build error with Ktor 3.4.2

**Solution**

- Curl 8.20.0+ requires 'non-http' feature to be declared for websockets. That's why `libcurl.a` has grown by ~24% (+200..300 KB)
- Fix linux-arm64 base triplet location
- Add `-pFIC` flag for Linux triplets
- Build new version of libcurl


osipxd:
Will be able to update to 8.20.0 after this PR: https://github.com/microsoft/vcpkg/pull/51597

bjhham (approved):

PR #5594, opened 2026-05-08
KTOR-8906 Bypass Dispatchers.IO for OutputStreamContent on servlet engines

amccague:
**Subsystem**
Servlet Servers, Jackson/Gson serialization

**Motivation**
In 3.x, `io.ktor.utils.io.jvm.javaio.PollersKt` was removed from ktor-io, causing the `withBlocking` bridge to always redispatch to `Dispatchers.IO`.

https://youtrack.jetbrains.com/issue/KTOR-8906

**Solution**
Restores Ktor 2.x behavior where serialization on servlet engines ran in-place on the request thread without consuming `Dispatchers.IO` threads.

Servlet-based engines (Jetty, Tomcat) now write `OutputStreamContent` directly to the native `OutputStream` via a new `@InternalAPI` overload, bypassing both the IO dispatch and the runBlocking bridge entirely.

Also removes the dead `isParkingAllowed` reflection that has been a no-op since the PollersKt class was deleted in the ktor-io 3.x rewrite, along with the redundant `ServerJacksonBlockingTest`/`ServerGsonBlockingTest` tests that existed solely to exercise that mechanism.


ktor-server/ktor-server-jetty-jakarta/jvm/src/io/ktor/server/jetty/jakarta/JettyApplicationCall.kt
```diff
@@ -186,6 +188,17 @@ public class JettyApplicationCall(
...
 
+        @OptIn(InternalAPI::class)
+        override suspend fun respondWriteChannelContent(content: OutgoingContent.WriteChannelContent) {
+            if (content is OutputStreamContent) {
+                Content.Sink.asOutputStream(response).use { stream ->
+                    content.writeTo(stream)
+                }
+                return
+            }
+            super.respondWriteChannelContent(content)
```
bjhham: nit: when would look more idiomatic here


```suggestion
            when(content) {
                is OutputStreamContent -> {
                    Content.Sink.asOutputStream(response).use { stream ->
                        content.writeTo(stream)
                    }
                }
                else -> {
                    super.respondWriteChannelContent(content)
                }
            }
```

ktor-server/ktor-server-jetty-jakarta/jvm/test/io/ktor/tests/server/jetty/jakarta/JettySerializationIOExhaustionTest.kt
```diff
@@ -0,0 +1,82 @@
...
+                    }.awaitAll()
+
+                    results.forEach { response ->
+                        assertEquals(HttpStatusCode.OK, response.status)
+                    }
+                }
+            }
+        } finally {
+            client.close()
+        }
```
bjhham: nit: may as well use `use` in these tests

```suggestion
        HttpClient(CIO) {
            install(io.ktor.client.plugins.contentnegotiation.ContentNegotiation) {
                jackson {}
            }
            install(HttpTimeout) {
                requestTimeoutMillis = 60_000
            }
        }.use { client ->
            withTimeout(60.seconds) {
                coroutineScope {
                    val results = (1..concurrentRequests).map { i ->
                        async(Dispatchers.Default) {
                            client.post("http://127.0.0.1:$port/json/jackson") {
                                contentType(ContentType.Application.Json)
                                setBody(mapOf("request" to i, "data" to "value$i"))
                            }
                        }
                    }.awaitAll()

                    results.forEach { response ->
                        assertEquals(HttpStatusCode.OK, response.status)
                    }
                }
            }
        }
```

bjhham (changes requested):
Just a couple of style comments, and could you please run `./gradlew formatKotlin` to make the linter happy?

Other than that it looks good, thank you for the fix!

amccague:
> Just a couple of style comments, and could you please run `./gradlew formatKotlin` to make the linter happy?
> 
> Other than that it looks good, thank you for the fix!

Needed to fix a stackoverflow error on running the command, done!

bjhham (approved):
Thanks!

PR #5596, opened 2026-05-11
KTOR-8151 Allow opting out of MeterFilter registration in MicrometerMetrics

fru1tworld:
**Subsystem**
Server, ktor-server-metrics-micrometer

**Motivation**
[KTOR-8151](https://youtrack.jetbrains.com/issue/KTOR-8151) The plugin always registers a `MeterFilter`, triggering a Micrometer warning when meters already exist on the registry before installation.

**Solution**
Add `registerDistributionStatisticConfig: Boolean` (default `true`) to `MicrometerMetricsConfig`. Setting it to `false` skips the `MeterFilter` registration. Mirrors `DropwizardMetricsConfig.registerJvmMetricSets`.

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-metrics-micrometer/jvm/src/io/ktor/server/metrics/micrometer/MicrometerMetrics.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-metrics-micrometer/jvm/src/io/ktor/server/metrics/micrometer/MicrometerMetrics.kt
@@ -127,6 +127,21 @@ public class MicrometerMetricsConfig {
     public var distributionStatisticConfig: DistributionStatisticConfig =
         DistributionStatisticConfig.Builder().percentiles(0.5, 0.9, 0.95, 0.99).build()
 
+    /**
+     * Controls whether the plugin registers a [MeterFilter] that applies [distributionStatisticConfig]
+     * to its request timers.
+     *
+     * Set to `false` if you register meters with [registry] before installing this plugin: Micrometer
+     * emits a warning when a [MeterFilter] is configured after a [Meter] has already been registered.
+     * When disabled, configure distribution statistics on the [registry] yourself before installing
+     * the plugin, otherwise [distributionStatisticConfig] has no effect.
+     *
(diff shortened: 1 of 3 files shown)


bjhham (approved):
Looks good, thanks 👍

ktor-server/ktor-server-plugins/ktor-server-metrics-micrometer/jvm/src/io/ktor/server/metrics/micrometer/MicrometerMetrics.kt
```diff
@@ -127,6 +127,21 @@ public class MicrometerMetricsConfig {
...
+     * Set to `false` if you register meters with [registry] before installing this plugin: Micrometer
+     * emits a warning when a [MeterFilter] is configured after a [Meter] has already been registered.
+     * When disabled, configure distribution statistics on the [registry] yourself before installing
+     * the plugin, otherwise [distributionStatisticConfig] has no effect.
+     *
+     * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.server.metrics.micrometer.MicrometerMetricsConfig.registerDistributionStatisticConfig)
+     *
+     * @see [MicrometerMetrics]
+     */
+    public var registerDistributionStatisticConfig: Boolean = true
```
osipxd, line 143: @bjhham, what do you think if we make `distributionStatisticConfig` nullable instead, and setting it to `null` would disable `meterFilter` registering?

bjhham: Technically that would be a breaking change on the property's return type, though DX-wise it makes more sense to have the registration tied to the config item.  I'm good either way.

fru1tworld:
Thanks! I'd lean toward the boolean — nullable would be a breaking change to the property's return type, and a dedicated flag reads more clearly than `null`-as-disabled. Happy to switch if you prefer otherwise.

bjhham:
I think we can still with the flag.  Maybe in a future release we can reevaluate the API more generally.

PR #5598, opened 2026-05-11
KTOR-9519 Migrate compiler plugin to ktor

bjhham:
**Subsystem**
Compiler Plugin

**Motivation**
[KTOR-9519](https://youtrack.jetbrains.com/issue/KTOR-9519) Migrate compiler plugin to the main repository

The motivation here is to ensure that the tests are always in sync with Ktor, so we can avoid running into failures when we try to release, which happened in the last release after updates to the OpenAPI generation.

**Solution**
- Moved module over to Ktor at the top level.
- Replaced library references with project references. 
- Reuse ktor project plugins etc.

See also https://github.com/ktorio/ktor-build-plugins/pull/288


bjhham:
@coderabbitai ignore

osipxd:
Could you run `./gradlew validatePublishedArtifacts --dump publishJvmAndCommonPublications` to update dumps?

gradle/libs.versions.toml
```diff
@@ -107,6 +107,12 @@ kotlin-serialization = { module = "org.jetbrains.kotlin:kotlin-serialization", v
 kotlin-stdlib = { module = "org.jetbrains.kotlin:kotlin-stdlib", version.ref = "kotlin" }
 kotlin-test = { module = "org.jetbrains.kotlin:kotlin-test", version.ref = "kotlin" }
 kotlin-test-junit5 = { module = "org.jetbrains.kotlin:kotlin-test-junit5", version.ref = "kotlin" }
+kotlin-compiler = { module = "org.jetbrains.kotlin:kotlin-compiler", version.ref = "kotlin" }
+kotlin-compilerEmbeddable = { module = "org.jetbrains.kotlin:kotlin-compiler-embeddable", version.ref = "kotlin" }
+kotlin-compilerTestFramework = { module = "org.jetbrains.kotlin:kotlin-compiler-internal-test-framework", version.ref = "kotlin" }
```
osipxd, line 112: 💅  Minor. (Worth applying only if you will change anything else as it would require changing dependencies in the compiler plugin project)

```suggestion
kotlin-compiler-embeddable = { module = "org.jetbrains.kotlin:kotlin-compiler-embeddable", version.ref = "kotlin" }
kotlin-compiler-testFramework = { module = "org.jetbrains.kotlin:kotlin-compiler-internal-test-framework", version.ref = "kotlin" }
```

osipxd (approved):
Good idea! 👍

PR #5600, opened 2026-05-11
KTOR-8443 Respect parameters in contentType and accept route selectors

fru1tworld:
**Subsystem**
ktor-server

**Motivation**
https://youtrack.jetbrains.com/issue/KTOR-8443

**Solution**
Preserve Content-Type parameters when matching `contentType` and `accept` route selectors.

### DIFF
--- a/ktor-server/ktor-server-core/common/src/io/ktor/server/routing/RouteSelector.kt
+++ b/ktor-server/ktor-server-core/common/src/io/ktor/server/routing/RouteSelector.kt
@@ -831,9 +831,9 @@ private fun ContentType.isCompatibleWith(other: ContentType): Boolean = when {
 
 private fun HeaderValue.toContentType(): ContentType {
     val contentType = ContentType.parse(value)
-    return if (params.isEmpty())
+    return if (params.isEmpty()) {
         contentType
-    else {
+    } else {
         ContentType(
             contentType.contentType,
             contentType.contentSubtype,
--- a/ktor-server/ktor-server-core/common/src/io/ktor/server/routing/RouteSelector.kt
+++ b/ktor-server/ktor-server-core/common/src/io/ktor/server/routing/RouteSelector.kt
@@ -708,7 +708,7 @@ internal data class ContentTypeHeaderRouteSelector(
         val parsedHeaders = parseAndSortContentTypeHeader(headers)
 
         val header = parsedHeaders.firstOrNull { header ->
-            contentTypes.any { it.isCompatibleWith(ContentType.parse(header.value)) }
+            contentTypes.any { header.toContentType().match(it) }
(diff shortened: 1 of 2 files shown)


ktor-server/ktor-server-core/common/src/io/ktor/server/routing/RouteSelector.kt
```diff
@@ -827,3 +827,8 @@ private fun ContentType.isCompatibleWith(other: ContentType): Boolean = when {
     this.contentSubtype == "*" -> other.match(this)
     else -> this.match(other)
 }
+
+private fun HeaderValue.toContentType(): ContentType =
+    params.asSequence()
+        .filter { it.name != "q" }
+        .fold(ContentType.parse(value)) { acc, p -> acc.withParameter(p.name, p.value) }
```
bjhham: Since this can be a performance-sensitive spot in the code, we can avoid unnecessary allocations with something like:

```suggestion
private fun HeaderValue.toContentType(): ContentType {
    val contentType = ContentType.parse(value)
    return if (params.isEmpty())
        contentType
    else {
        ContentType(
            contentType.contentType,
            contentType.contentSubtype,
            params.filter { it.name != "q" }
        )
    }
}
```

bjhham (changes requested):
Thanks for the fix!  One little optimization

ktor-server/ktor-server-core/common/src/io/ktor/server/routing/RouteSelector.kt
fru1tworld: Thanks for the review!

Great insight on object creation overhead

bjhham (approved):

ktor-server/ktor-server-core/common/src/io/ktor/server/routing/RouteSelector.kt
```diff
@@ -827,3 +827,16 @@ private fun ContentType.isCompatibleWith(other: ContentType): Boolean = when {
     this.contentSubtype == "*" -> other.match(this)
     else -> this.match(other)
 }
+
+private fun HeaderValue.toContentType(): ContentType {
+    val contentType = ContentType.parse(value)
+    return if (params.isEmpty())
+        contentType
+    else {
```
bjhham: Looks like the linter is displeased about the braceless if branch:

```suggestion
    return if (params.isEmpty()) {
        contentType
    } else {
```

bjhham (changes requested):

fru1tworld:
Thanks For Feedback !

bjhham (approved):

bjhham:
@fru1tworld could you please rebase onto `release/3.x`?

PR #5602, opened 2026-05-12
Netty request handling fixes

bjhham:
**Subsystem**
Server, Netty

**Motivation**
- [KTOR-9531](https://youtrack.jetbrains.com/issue/KTOR-9531) Netty server intermittently drops requests after upgrading to 3.4.3
- [KTOR-9542](https://youtrack.jetbrains.com/issue/KTOR-9542) Netty: The request handler runs on worker event loop instead of call event loop since 3.4.3
- Another issue with updating from Netty 4.2.9.Final to 4.2.12.Final in Ktor 3.4.3: because of some classpath conflict issues, some libraries would encounter runtime problems with Netty actually failing to respond when `netty-codec` was imported from 4.1.*. This was wrongfully attributed to KTOR-9531 above.

**Solution**
- Worker loop fix: handlers now take a `callEventGroup` and pin one `EventExecutor` per channel via an `AttributeKey`, dispatch the deferred coroutine start onto it. `NettyDispatcher.CurrentContext` carries that pinned executor so suspensions stay on the call-event-group thread.
- Discarded event fix: added a `@Sharable` `NettyHttp1ApplicationCallSink` appended in `channelActive` after `RequestBodyHandler` and after user `channelPipelineConfig` handlers, swallowing `NettyHttp1ApplicationCall` messages so they don't reach Netty's tail.
- Classpath shenanigans fix: added `netty-codec` as a dependency to avoid clashing with older versions imported from `netty-grpc` and the like

### DIFF
--- a/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/http2/NettyHttp2Handler.kt
+++ b/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/http2/NettyHttp2Handler.kt
@@ -80,7 +80,9 @@ internal class NettyHttp2Handler(
         // "Discarded inbound message" warnings for calls that pass through any user-added
         // channelPipelineConfig handlers. The call lifecycle is driven by startHttp2, so the sink
         // only needs to drop the call without further action.
-        context.pipeline().addLast(NettyHttp2ApplicationCallSink)
+        if (context.pipeline().get(NettyHttp2ApplicationCallSink::class.java) == null) {
+            context.pipeline().addLast(NettyHttp2ApplicationCallSink)
+        }
         context.fireChannelActive()
     }
 
--- a/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/http2/NettyHttp2Handler.kt
+++ b/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/http2/NettyHttp2Handler.kt
@@ -19,7 +19,6 @@ import io.netty.channel.ChannelHandlerContext
 import io.netty.channel.ChannelInboundHandlerAdapter
 import io.netty.handler.codec.http2.*
 import io.netty.util.AttributeKey
-import io.netty.util.concurrent.EventExecutor
 import io.netty.util.concurrent.EventExecutorGroup
 import kotlinx.coroutines.*
 import java.lang.reflect.Field
@@ -104,7 +103,7 @@ internal class NettyHttp2Handler(
 
     private fun startHttp2(context: ChannelHandlerContext, headers: Http2Headers) {
         val callJob = Job(parent = userCoroutineContext[Job])
-        val callExecutor = pinnedCallExecutor(context)
+        val callExecutor = pinnedCallExecutor(context, callEventGroup)
         val callContext = userCoroutineContext +
             NettyDispatcher.CurrentContext(context, callExecutor) +
             callJob +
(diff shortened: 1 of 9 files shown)


ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/http2/NettyHttp2Handler.kt
```diff
@@ -28,6 +29,7 @@ import kotlin.coroutines.CoroutineContext
 internal class NettyHttp2Handler(
```
zibet27, line 30: I was playing with my own discarded event fix

I found that NettyHttp2Handler needs an event sink as well

Also, I saw that RequestBodyHandler can be added to the pipeline multiple times

zibet27 (changes requested):

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/http2/NettyHttp2Handler.kt
bjhham: Ok, thanks!

bjhham: > Also, I saw that RequestBodyHandler can be added to the pipeline multiple times

Not sure what you're referring to here...  It's only added in one place.

zibet27: Yeah, but from what I saw, that function can be called multiple times (not sure why), and you can see that some handlers are duplicated

Here is my test:

```kotlin

// omit http2 tests
    @Test
    @Http1Only
    fun http1PipelineHandlersAreUnique() = runTestWithRealTime {
        val pipelineHandlerNames = AtomicReference<List<String>>(emptyList())

        enableSsl = false
        channelPipelineConfig = {
            addLast(
                "customHandler",
                object : ChannelInboundHandlerAdapter() {
                    override fun channelRead(ctx: ChannelHandlerContext, msg: Any) {
                        if (msg is NettyHttp1ApplicationCall) {
                            pipelineHandlerNames.set(ctx.pipeline().names())
                        }
                        super.channelRead(ctx, msg)
                    }
                }
            )
        }

        try {
            createAndStartServer {
                get("/") {
                    call.respondText("OK")
                }
            }

            withUrl("/") {
                assertEquals(HttpStatusCode.OK, status)
                assertEquals("OK", bodyAsText())
            }

            assertPipelineHandlersAreUnique(pipelineHandlerNames.get())
        } finally {
        ...
        }
    }
```

That should fail atm

bjhham: Ohh interessting 🤔

zibet27: Btw, I can share all the tests that I've added

bjhham: Updated the custom pipeline config test to verify this and fixed up the handlers.

ktor-server/ktor-server-netty/jvm/test/io/ktor/tests/server/netty/NettySpecificTest.kt
```diff
@@ -352,4 +363,210 @@ class NettySpecificTest {
...
+                callEventGroup.any { it.inEventLoop(thread) },
+                "Handler ran on '${thread.name}', not on any call event group thread"
+            )
+        } finally {
+            server.stopSuspend()
+        }
+    }
+
+    @Test
+    fun `no messages are discarded from the netty pipeline for HTTP1`() = noDiscardedMessagesTest(enableH2c = false) {
```
zibet27, line 412: Does it always fail without a fix?

zibet27 (approved):
Nice!
Hope we won't have more issues with that

PR #5606, opened 2026-05-13
Merge main to release

bjhham:


osipxd:
The release branch is 1 commit ahead of main, so maybe we could merge that commit (with merge commit) into main and then just set release branch HEAD to the main HEAD?

bjhham:
Yeah, that works too.

bjhham:
Going to merge release to main and re-point release instead.

PR #5620, opened 2026-05-14
Fix SSE (CIO) GET content length handling (KTOR-9588)

paksv:
Subsystem
Client, ktor-client-cio, ktor-client-java, SSE

Motivation
Body-less SSE GET requests through CIO and Java could emit `Content-Length: 0` because the SSE content wrapper hid the original `OutgoingContent.NoContent` semantics. That makes Ktor send an explicit zero content length for a request that should be treated as body-less.

Solution
- Detect wrapped no-content bodies when preparing CIO and Java client requests.
- Omit `Content-Length` for body-less SSE GET requests while preserving SSE wrapper headers.
- Add regression coverage for CIO and Java engines and preserve content length behavior for SSE POST requests with a real body.

Verification
- Targeted SSE regression tests for CIO and Java.
- `./gradlew :ktor-client-cio:jvmTest`
- `./gradlew :ktor-client-java:jvmTest`
- `./gradlew :ktor-client-cio:formatKotlin :ktor-client-cio:lintKotlin`
- `./gradlew :ktor-client-java:formatKotlin :ktor-client-java:lintKotlin`

Notes
Full multiplatform assemble was not run locally because this environment lacks Xcode for Apple targets. No public/protected API changed, so ABI dumps were not updated.

### DIFF
--- a/ktor-client/ktor-client-java/jvm/src/io/ktor/client/engine/java/JavaHttpRequest.kt
+++ b/ktor-client/ktor-client-java/jvm/src/io/ktor/client/engine/java/JavaHttpRequest.kt
@@ -50,10 +50,19 @@ internal fun HttpRequestData.convertToHttpRequest(callContext: CoroutineContext)
             }
         }
 
-        method(method.value, body.convertToHttpRequestBody(callContext))
+        if (method == HttpMethod.Get && body.getUnwrapped() is OutgoingContent.NoContent) {
+            GET()
+        } else {
+            method(method.value, body.convertToHttpRequestBody(callContext))
+        }
     }
 
     return builder.build()
+}
+
+private fun OutgoingContent.getUnwrapped(): OutgoingContent = when (this) {
+    is OutgoingContent.ContentWrapper -> delegate().getUnwrapped()
+    else -> this
 }
 
 @OptIn(DelicateCoroutinesApi::class)
--- a/ktor-client/ktor-client-java/jvm/src/io/ktor/client/engine/java/JavaHttpRequest.kt
+++ b/ktor-client/ktor-client-java/jvm/src/io/ktor/client/engine/java/JavaHttpRequest.kt
@@ -50,7 +50,11 @@ internal fun HttpRequestData.convertToHttpRequest(callContext: CoroutineContext)
             }
         }
 
-        method(method.value, body.convertToHttpRequestBody(callContext))
+        if (method == HttpMethod.Get && body.isEmpty()) {
+            GET()
+        } else {
+            method(method.value, body.convertToHttpRequestBody(callContext))
+        }
     }
 
     return builder.build()
--- a/ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/plugins/ServerSentEventsTest.kt
+++ b/ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/plugins/ServerSentEventsTest.kt
@@ -384,6 +384,19 @@ class ServerSentEventsTest : ClientLoader() {
     }
 
     @Test
+    fun testGetRequestWithoutBodyDoesNotSendContentLength() = clientTests(only("CIO", "Java")) {
+        config {
+            install(SSE)
+        }
+
+        test { client ->
+            client.sse("$TEST_SERVER/sse/headers") {
+                assertEquals("", incoming.single().data)
+            }
+        }
+    }
+
+    @Test
     fun testRequestBody() = clientTests {
         config {
             install(SSE)
@@ -400,6 +413,24 @@ class ServerSentEventsTest : ClientLoader() {
             }) {
                 assertEquals(contentType, call.request.contentType()?.withoutParameters())
                 assertEquals(body, incoming.single().data)
+            }
+        }
+    }
+
+    @Test
+    fun testPostRequestWithBodySendsContentLength() = clientTests(only("CIO", "Java")) {
+        config {
+            install(SSE)
+        }
+
+        val body = "hello"
+        test { client ->
+            client.sse({
+                method = HttpMethod.Post
+                url("$TEST_SERVER/sse/echo-headers")
+                setBody(body)
+            }) {
+                assertEquals(body.length.toString(), incoming.single().data)
             }
         }
     }
--- a/ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/plugins/ServerSentEventsTest.kt
+++ b/ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/plugins/ServerSentEventsTest.kt
@@ -403,6 +403,38 @@ class ServerSentEventsTest : ClientLoader() {
         }
     }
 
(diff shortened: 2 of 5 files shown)


ktor-test-server/src/main/kotlin/test/server/tests/ServerSentEvents.kt
```diff
@@ -73,13 +73,27 @@ internal fun Application.serverSentEvents() {
...
                     }
                 )
             }
+            post("/echo-headers") {
+                call.respondSseEvents(
+                    flow {
+                        emit(SseEvent(call.request.headers[HttpHeaders.ContentLength] ?: ""))
+                    }
+                )
+            }
```
osipxd: Let's replace two routes (`get("/headers")` and `post("/echo-headers")`) with one:

```kotlin
route("/content-length") {
    handle {
        val events = flowOf(
            SseEvent(call.request.headers[HttpHeaders.ContentLength].orEmpty())
        )
        call.respondSseEvents(events)
    }
}
```

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/plugins/ServerSentEventsTest.kt
```diff
@@ -404,6 +417,24 @@ class ServerSentEventsTest : ClientLoader() {
         }
     }
 
+    @Test
+    fun testPostRequestWithBodySendsContentLength() = clientTests(only("CIO", "Java")) {
```
osipxd: It would be great to run these tests for all engines
```suggestion
    fun testPostRequestWithBodySendsContentLength() = clientTests {
```

I see that `Apache5` fails. If you don't have capacity for fixing it, you can skip it with `except("Apache5")`

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/plugins/ServerSentEventsTest.kt
```diff
@@ -383,6 +383,19 @@ class ServerSentEventsTest : ClientLoader() {
         }
     }
 
+    @Test
+    fun testGetRequestWithoutBodyDoesNotSendContentLength() = clientTests(only("CIO", "Java")) {
```
osipxd: It'd be great to put these tests one after another as they're very similar

ktor-client/ktor-client-java/jvm/src/io/ktor/client/engine/java/JavaHttpRequest.kt
```diff
@@ -50,12 +50,21 @@ internal fun HttpRequestData.convertToHttpRequest(callContext: CoroutineContext)
             }
         }
 
-        method(method.value, body.convertToHttpRequestBody(callContext))
+        if (method == HttpMethod.Get && body.getUnwrapped() is OutgoingContent.NoContent) {
+            GET()
```
osipxd, line 54: Could you check if `except("Java")` can be removed from `testRequestHasContentLength` after this change?

osipxd (changes requested):
Thank you for the fix! It looks good overall, I have just a couple of stylistic comments

Mr3zee:
Hey @paksv ! Are there any updates on the PR?

ktor-test-server/src/main/kotlin/test/server/tests/ServerSentEvents.kt
paksv: Done — both routes are replaced with a single method-agnostic `route("/content-length") { handle { ... } }` as suggested, and both tests now use it.

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/plugins/ServerSentEventsTest.kt
paksv: Both tests now run for all engines. `Apache5` was failing for exactly the same reason (`hasContent = requestData.body !is OutgoingContent.NoContent` in `ApacheRequestProducer`), so I fixed it here as well instead of excluding it — it now uses `!requestData.body.isEmpty()` too.

The only exclusion left is `except("Java")` on the GET test, which is needed on JDK < 19 (see the reply about `testRequestHasContentLength`).

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/plugins/ServerSentEventsTest.kt
paksv: Moved — `testGetRequestWithoutBodyDoesNotSendContentLength` and `testPostRequestWithBodySendsContentLength` are now adjacent.

ktor-client/ktor-client-java/jvm/src/io/ktor/client/engine/java/JavaHttpRequest.kt
paksv: Checked, and unfortunately `except("Java")` has to stay: the failure is not caused by Ktor but by [JDK-8283544](https://bugs.openjdk.org/browse/JDK-8283544), which is fixed only in JDK 19+. On an older JDK the JDK HTTP client sends `Content-Length: 0` even for a request built with `HttpRequest.Builder.GET()`, so `testRequestHasContentLength` still fails there.

I verified this by running the tests with `-Ptest.jdk=17`: with the exclusion removed the Java engine fails, and it passes on the default (21) toolchain. For the same reason I added the same `except("Java")` + comment to the new `testGetRequestWithoutBodyDoesNotSendContentLength` test. Once the minimum supported JDK for the Java engine is 19+, both exclusions can be dropped together.

paksv:
Rebased onto the latest `main` (single commit) and addressed all review comments:

- **Test server**: `get("/headers")` + `post("/echo-headers")` replaced with a single method-agnostic `route("/content-length") { handle { ... } }`.
- **Tests**: `testGetRequestWithoutBodyDoesNotSendContentLength` and `testPostRequestWithBodySendsContentLength` are now adjacent and run for **all** engines (no more `only("CIO", "Java")`).
- **Apache5**: fixed the same bug there (`hasContent` now uses `!requestData.body.isEmpty()`), so no engine has to be excluded for that reason.
- **Java engine**: dropped the duplicated private `getUnwrapped()` and reused the existing `@InternalAPI OutgoingContent.isEmpty()`; body-less GET uses `HttpRequest.Builder.GET()`.
- `except("Java")` on `testRequestHasContentLength` has to stay — the failure comes from [JDK-8283544](https://bugs.openjdk.org/browse/JDK-8283544) (fixed only in JDK 19+), verified with `-Ptest.jdk=17`. The same exclusion + comment was added to the new GET test.

Verified: `:ktor-client-tests:jvmTest`, `:ktor-client-cio:jvmTest`, `:ktor-client-java:jvmTest`, `:ktor-client-apache5:jvmTest`, plus `lintKotlin`/`assemble` for the touched modules.

Mr3zee (approved):

osipxd (approved):

PR #5624, opened 2026-05-15
KTOR-9589 Fix duplicate PROPERTY_SETTER target in InternalAPI annotation

riednic:
**Subsystem**
ktor-io

**Motivation**
Fixes a duplication in the `@InternalAPI` annotation where `AnnotationTarget.PROPERTY_SETTER` was listed twice in the `@Target` declaration, making one of the entries redundant. Refers to `KTOR-9589` (https://youtrack.jetbrains.com/issue/KTOR-9589).

**Solution**
Removed the duplicate AnnotationTarget.PROPERTY_SETTER entry from the `@Target` declaration of `@InternalAPI` in Annotations.kt.

### DIFF
--- a/ktor-io/common/src/io/ktor/utils/io/Annotations.kt
+++ b/ktor-io/common/src/io/ktor/utils/io/Annotations.kt
@@ -22,7 +22,6 @@ package io.ktor.utils.io
     AnnotationTarget.PROPERTY,
     AnnotationTarget.FIELD,
     AnnotationTarget.CONSTRUCTOR,
-    AnnotationTarget.PROPERTY_SETTER,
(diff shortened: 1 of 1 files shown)


bjhham (approved):

zibet27:
I think this is a typo and suggest updating the duplicated line to `AnnotationTarget.PROPERTY_GETTER`

`@InternalAPI` also makes sense for getters

riednic:
@zibet27 

I was wondering if it is just a typo or duplicated, but it makes sense to have `@InternalApi` also for gettes, i updated the PR accordingly.

zibet27:
Thank you, @riednic!
Looks good to me!

zibet27 (approved):

PR #5625, opened 2026-05-15
KTOR-9617 Fix OutputStreamContent and WriterContent exhausting Dispatchers.IO

osipxd:
**Subsystem**
IO

**Motivation**
[KTOR-9617](https://youtrack.jetbrains.com/issue/KTOR-9617) OutputStreamContent and WriterContent can exhaust Dispatchers.IO
[KTOR-9626](https://youtrack.jetbrains.com/issue/KTOR-9626) Do not use blocking bridges on Dispatchers.IO (partially fixed)

**Solution**
Do not close the channel inside the `runBlocking`. Similarly to how it's done for other `OutgoingContent.WriteChannelContent` implementations the channel closing is handled in `BaseApplicationResponse.respondWriteChannelContent`.

### DIFF
--- a/ktor-http/jvm/src/io/ktor/http/content/BlockingBridge.kt
+++ b/ktor-http/jvm/src/io/ktor/http/content/BlockingBridge.kt
@@ -13,6 +13,22 @@ import kotlinx.coroutines.withContext
 import java.io.OutputStream
 import java.io.Writer
 
+private const val BLOCKING_BRIDGE_PARALLELISM_PROPERTY_NAME = "io.ktor.blocking.bridge.parallelism"
+
+/**
+ * Dispatcher used by Ktor blocking bridges.
+ *
+ * Blocking bridges expose synchronous APIs over suspending Ktor I/O primitives. Their operations may block
+ * waiting for backpressure or suspending I/O completion. This dispatcher gives such bridges a dedicated parallelism
+ * budget so they don't exhaust shared [Dispatchers.IO] capacity used by user code and other Ktor internals.
+ *
+ * The parallelism can be configured with [BLOCKING_BRIDGE_PARALLELISM_PROPERTY_NAME].
+ */
+private val BlockingBridgeDispatcher = Dispatchers.IO.limitedParallelism(
+    parallelism = System.getProperty(BLOCKING_BRIDGE_PARALLELISM_PROPERTY_NAME)?.toIntOrNull() ?: 64,
+    name = "ktor-blocking-bridge",
+)
+
 /**
  * Executes [block] with this channel represented as a blocking [OutputStream].
  *
@@ -23,7 +39,7 @@ import java.io.Writer
  * The stream passed to [block] is owned by this function and must not be used after [block] returns.
  */
 internal suspend inline fun ByteWriteChannel.withBlockingOutputStream(
-    dispatcher: CoroutineDispatcher = Dispatchers.IO,
+    dispatcher: CoroutineDispatcher = BlockingBridgeDispatcher,
     crossinline block: suspend (OutputStream) -> Unit,
 ) {
     withContext(dispatcher) {
@@ -43,7 +59,7 @@ internal suspend inline fun ByteWriteChannel.withBlockingOutputStream(
  */
 internal suspend inline fun ByteWriteChannel.withBlockingWriter(
     charset: Charset,
-    dispatcher: CoroutineDispatcher = Dispatchers.IO,
+    dispatcher: CoroutineDispatcher = BlockingBridgeDispatcher,
     crossinline block: suspend (Writer) -> Unit,
 ) {
     withContext(dispatcher) {
--- a/ktor-http/jvm/src/io/ktor/http/content/BlockingBridge.kt
+++ b/ktor-http/jvm/src/io/ktor/http/content/BlockingBridge.kt
@@ -4,18 +4,61 @@
 
 package io.ktor.http.content
 
+import io.ktor.utils.io.*
+import io.ktor.utils.io.charsets.*
+import io.ktor.utils.io.jvm.javaio.*
+import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.withContext
+import java.io.OutputStream
+import java.io.Writer
 
 /**
- * Redispatches [block] onto [Dispatchers.IO] for blocking I/O.
+ * Executes [block] with this channel represented as a blocking [OutputStream].
  *
- * This is used by non-blocking engines (CIO, Netty) where the calling thread is an event-loop
- * thread that must not block. Servlet-based engines bypass this entirely via
- * [OutputStreamContent.writeTo] with a native [java.io.OutputStream].
+ * Blocking operations are dispatched to [dispatcher] so they don't block event loop threads. This function doesn't
+ * close the channel after [block] finishes: the caller that owns the [ByteWriteChannel] lifecycle is responsible for
+ * closing it from a suspending context.
+ *
+ * The stream passed to [block] is owned by this function and must not be used after [block] returns.
  */
-internal suspend fun withBlocking(block: suspend () -> Unit) {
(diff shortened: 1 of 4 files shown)


amccague:
This looks intended to fix https://youtrack.jetbrains.com/issue/KTOR-8906 for the event loop engines.

ktor-http/jvm/src/io/ktor/http/content/BlockingBridge.kt
```diff
@@ -4,18 +4,60 @@
...
+ * close the channel after [block] finishes: the caller that owns the [ByteWriteChannel] lifecycle is responsible for
+ * closing it from a suspending context.
+ *
+ * The stream passed to [block] is owned by this function and must not be used after [block] returns.
  */
-internal suspend fun withBlocking(block: suspend () -> Unit) {
-    withContext(Dispatchers.IO) {
-        block()
+internal suspend inline fun ByteWriteChannel.withBlockingOutputStream(
+    dispatcher: CoroutineDispatcher = Dispatchers.IO,
```
e5l: should we use different dispatcher here?

e5l (dismissed):

ktor-http/jvm/src/io/ktor/http/content/BlockingBridge.kt
```diff
@@ -4,18 +4,60 @@
...
+}
+
+private fun OutputStream.nonClosing(): OutputStream = NonClosingOutputStream(this)
+
+private class NonClosingOutputStream(private val delegate: OutputStream) : OutputStream() {
+    override fun write(b: Int) = delegate.write(b)
+    override fun write(b: ByteArray) = delegate.write(b)
+    override fun write(b: ByteArray, off: Int, len: Int) = delegate.write(b, off, len)
+    override fun flush() = delegate.flush()
+    override fun close() = Unit
```
bjhham, line 79: Why is it that only the `close()` can exhaust threads?  I would think that all blocking operations would have the potential since they're using `runBlocking`.

bjhham (commented):
Could you introduce a test as well?

ktor-http/jvm/src/io/ktor/http/content/BlockingBridge.kt
bjhham: Maybe this could be used in conjunction with this https://github.com/ktorio/ktor/pull/5009 idk

osipxd: The problem with `closeAndFlush` is that its combination with `ByteWriteChannel.onClose` (which accepts suspend function) leads to a deadlock preventing releasing IO dispatcher threads. So in real life the problem arises when `CoroutineScope.reader(channel: ByteChannel)` implementation calls `channel.onClose { job.join() }` and that job never completes because `Dsipatchers.IO` is exhausted.

I'll try to write a test demonstrating this. Similar to the one described in the YouTrack issue.

ktor-http/jvm/src/io/ktor/http/content/BlockingBridge.kt
```diff
@@ -4,18 +4,67 @@
...
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.withContext
+import java.io.OutputStream
+import java.io.Writer
+
+/** A separate dispatcher for blocking content operations to prevent exhausting Dispatchers.IO. */
+private val BlockingContentDispatcher = Dispatchers.IO.limitedParallelism(
+    parallelism = System.getProperty("io.ktor.blockingContentParallelism")?.toIntOrNull() ?: 64,
+    name = "ktor-blocking-content",
+)
```
amccague: Correct me if your understanding differs: the default Dispatchers.IO is limited to 64 already unless cores count is >64. Therefore this won't prevent exhaustion, assuming the defaults, UNLESS `io.ktor.blockingContentParallelism` is set to less than 64?

osipxd: `Dispatchers.IO` [is elastic](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-i-o.html), so `limitedParallelism` views are not restricted by its parallelism. That means `BlockingContentDispatcher` won't spend `Dispatchers.IO` capacity.

amccague: Read the manual moment, appreciate the response.

ktor-http/jvm/src/io/ktor/http/content/OutputStreamContent.kt
```diff
@@ -23,13 +23,7 @@ public class OutputStreamContent(
...
 
     override suspend fun writeTo(channel: ByteWriteChannel) {
-        withBlocking {
-            // use block should be inside because closing OutputStream is blocking as well
-            // and should not be invoked in a epoll/kqueue/reactor thread
-            channel.toOutputStream().use { stream ->
-                stream.body()
-            }
-        }
+        channel.withBlockingOutputStream(block = body)
```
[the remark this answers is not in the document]
osipxd: Actually `"The stream would be automatically closed after [body] finish."` comment is the same for all  `OutgoingContent.WriteChannelContent` implementations, so I'd keep it "as is" in this PR

ktor-server/ktor-server-test-host/jvm/src/io/ktor/server/testing/client/TestEngineWebsocketSession.kt
```diff
@@ -49,3 +50,43 @@ internal class TestEngineWebsocketSession(
...
+
+        if (register) {
+            delegate.invokeOnClose { cause ->
+                val handlersToInvoke = synchronized(lock) {
+                    closed = true
+                    closeCause = cause
+                    handlers.toList().also { handlers.clear() }
+                }
+                handlersToInvoke.forEach { it(cause) }
+            }
```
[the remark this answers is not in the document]
osipxd: Oops. This shouldn't be included in the PR 👀

ktor-http/jvm/src/io/ktor/http/content/BlockingBridge.kt
```diff
@@ -4,18 +4,67 @@
...
+import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.withContext
+import java.io.OutputStream
+import java.io.Writer
+
+/** A separate dispatcher for blocking content operations to prevent exhausting Dispatchers.IO. */
+private val BlockingContentDispatcher = Dispatchers.IO.limitedParallelism(
+    parallelism = System.getProperty("io.ktor.blockingContentParallelism")?.toIntOrNull() ?: 64,
+    name = "ktor-blocking-content",
```
osipxd: I haven't found any conventions regarding system property naming. I've found at least two styles:
- `io.ktor.client.statement.useEngineDispatcher` (introduced by me, btw)
- `ktor.internal.cio.disable.chararray.pooling`

So which style is the correct one?
Also I'm seeking for a better name for this property.

ktor-http/jvm/src/io/ktor/http/content/BlockingBridge.kt
osipxd: Added a separate dispatcher with 64 threads by default

ktor-http/jvm/src/io/ktor/http/content/BlockingBridge.kt
bjhham: I think name-spaced as you have it should be the convention based on what I've seen in other libraries like `kotlinx.io`.  I can't think of a better name atm 🤷

bjhham (approved):

ktor-http/jvm/src/io/ktor/http/content/BlockingBridge.kt
osipxd: Renamed this property to `io.ktor.blocking.bridge.parallelism` to align naming style with kotlinx libraries

PR #5626, opened 2026-05-18
KTOR-8870 Don't close unopened file in readChannel

fru1tworld:
**Subsystem**
ktor-utils

**Motivation**
[KTOR-8870](https://youtrack.jetbrains.com/issue/KTOR-8870)

When `File.readChannel` is called for a nonexistent file, the writer fails before evaluating the lazy `RandomAccessFile`. `invokeOnCompletion` is then the first to touch it, re-throwing `FileNotFoundException` and surfacing as an uncaught `CompletionHandlerException` on a worker thread.

**Solution**
Hold the lazy directly, close only when `isInitialized()`, and swallow close errors with `runCatching`. Adds a regression test that asserts no exception leaks via `CoroutineExceptionHandler`.


bjhham (approved):
Thanks!  Could you please change the base to `release/3.x`?

PR #5627, opened 2026-05-18
KTOR-8883 Support nested jars in static resources

fru1tworld:
**Subsystem**
Server, Static Content

**Motivation**
[KTOR-8883](https://youtrack.jetbrains.com/issue/KTOR-8883) Support nested jars in static resources

**Solution**
Follow-up to [#5086](https://github.com/ktorio/ktor/pull/5086). `findContainingJarFile` now returns `null` for nested `jar:file:` URLs (Spring Boot 2.x bootJar, WAR/EAR) so `resourceClasspathResource` falls back to `URIFileContent` instead of throwing. `java.util.jar.JarFile` cannot open nested archives; environments with a registered URL handler can still stream the entry.


bjhham (commented):
Could you demonstrate how this might work with the URI file content fallback, possibly with a high-level test?

bjhham (approved):
Looks good, thanks!

PR #5629, opened 2026-05-18
KTOR-9590 Improved JSON schema inference context

bjhham:
**Subsystem**
Server, OpenAPI, Schema inference

**Motivation**
[KTOR-9590](https://youtrack.jetbrains.com/issue/KTOR-9590) OpenAPI Contextual JSON schema inference support

**Solution**
Previously, we'd only infer the default content type from the content negotiation plugin.  Now, we'll register the schema inference automatically from the serializers module.  Also, since this is a route-scoped plugin, we now use route attributes for assigning the inference method.

Note, I haven't done the same for the other JSON libraries.  This would be a useful feature in the future, though we don't have the relevant adapters atm.


osipxd (approved):
Nicely done 👍

PR #5631, opened 2026-05-18
Add common constants for HTTP auth algorithms

zibet27:
**Subsystem**
Common

**Motivation**
Reuse these constants for SAML, OIDC and digest auth

### DIFF
--- a/ktor-http/common/src/io/ktor/http/auth/KeyAlgorithm.kt
+++ b/ktor-http/common/src/io/ktor/http/auth/KeyAlgorithm.kt
@@ -1,0 +1,29 @@
+/*
+ * Copyright 2014-2026 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.http.auth
+
+import kotlin.jvm.JvmInline
+
+/**
+ * Represents a cryptographic key algorithm family.
+ *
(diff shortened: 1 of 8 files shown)


zibet27:
@coderabbitai review

bjhham (approved):
Good call 👍

ktor-http/common/src/io/ktor/http/auth/KeyAlgorithm.kt
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * Represents a cryptographic key algorithm family.
+ *
+ * @property name The common name of the key algorithm family.
+ */
+@JvmInline
+public value class KeyAlgorithm(public val name: String) {
+    public companion object {
+        /** RSA key algorithm family. */
+        public val RSA: KeyAlgorithm = KeyAlgorithm("RSA")
```
osipxd, line 18: I know we've recently added `@JvmField` annotation to other companion object fields. Should we use it here as well?
cc @bjhham

zibet27: Here it fails to compile with `@JvmField`
> [INAPPLICABLE_JVM_FIELD] JvmField cannot be applied to a property of a value class type.

PR #5633, opened 2026-05-18
Fix default methods filtering in CORS plugin (#5103)

### LINKED ISSUE #5103: CORS Plugin: Access-Control-Allow-Methods header does not include default methods (GET, POST, HEAD)

HMH-3080:
Fixes #5103

This PR ensures that default HTTP methods (`GET`, `POST`, `HEAD`) are included explicitly in the `Access-Control-Allow-Methods` header instead of being filtered out. Modern browsers strictly require these methods to be present in the preflight (OPTIONS) response header when custom configurations (like non-simple Content-Types) trigger a preflight validation.

Changes :
- Removed `.filterNot { it in CORSConfig.CorsDefaultMethods }` from `methodsListHeaderValue` computation.
- Added a dedicated unit test `testDefaultMethodsIncludedInAllowMethodsHeader` to verify the fix.

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-cors/common/src/io/ktor/server/plugins/cors/CORS.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-cors/common/src/io/ktor/server/plugins/cors/CORS.kt
@@ -54,10 +54,7 @@ internal fun PluginBuilder<CORSConfig>.buildPlugin() {
     val allowNonSimpleContentTypes: Boolean = pluginConfig.allowNonSimpleContentTypes
     val headersList = pluginConfig.headers.filterNot { it in CORSConfig.CorsSimpleRequestHeaders }
         .let { if (allowNonSimpleContentTypes) it + HttpHeaders.ContentType else it }
-    val methodsListHeaderValue = methods.filterNot { it in CORSConfig.CorsDefaultMethods }
-        .map { it.value }
-        .sorted()
-        .joinToString(", ")
+    val methodsListHeaderValue = methods.map { it.value }.sorted().joinToString(", ")
     val maxAgeHeaderValue = pluginConfig.maxAgeInSeconds.let { if (it > 0) it.toString() else null }
     val exposedHeaders = when {
         pluginConfig.exposedHeaders.isNotEmpty() -> pluginConfig.exposedHeaders.sorted().joinToString(", ")
--- a/ktor-server/ktor-server-tests/common/test/io/ktor/tests/server/plugins/CORSTest.kt
+++ b/ktor-server/ktor-server-tests/common/test/io/ktor/tests/server/plugins/CORSTest.kt
@@ -1592,4 +1592,33 @@ class CORSTest {
             assertEquals(response.status, HttpStatusCode.Forbidden)
         }
     }
+
+    @Test
+    fun testDefaultMethodsIncludedInAllowMethodsHeader() = testApplication {
+        install(CORS) {
+            allowMethod(HttpMethod.Post)
+            anyHost()
+        }
+
+        routing {
+            post("/") {
+                call.respond("OK")
+            }
+        }
+
+        val response = client.options("/") {
+            header(HttpHeaders.Origin, "https://example.com")
+            header(HttpHeaders.AccessControlRequestMethod, "POST")
+        }
+
+        assertEquals(HttpStatusCode.OK, response.status)
+        val allowMethodsHeader = response.headers[HttpHeaders.AccessControlAllowMethods]
+
+        assertNotNull(allowMethodsHeader, "Access-Control-Allow-Methods header should not be null")
+        assertTrue(
+            allowMethodsHeader.contains("POST"),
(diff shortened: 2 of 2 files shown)


bjhham (commented):
Thanks!

Please run `./gradlew formatKotlin` and submit the result to make the linter happy.

bjhham (approved):

HMH-3080:
@bjhham 
Done! Ran ./gradlew formatKotlin, fixed the unexpected blank line, and pushed the updated code. Thanks for the review!

bjhham (commented):
I just noticed there was already a PR submitted for this a while ago 😅 

https://github.com/ktorio/ktor/pull/5104

HMH-3080:
@bjhham 

Thanks for pointing that out! I looked into #5104 and noticed it’s been open since last year. Also, it introduces an unnecessary .sorted() call on every preflight request computation, which adds a tiny but avoidable overhead.

My PR keeps the code simpler, includes a robust unit test for this case, and I've already formatted the code as you requested. I’d love it if we could move forward with this cleaner implementation if possible! 🙌

ktor-server/ktor-server-plugins/ktor-server-cors/common/src/io/ktor/server/plugins/cors/CORS.kt
```diff
@@ -54,10 +54,7 @@ internal fun PluginBuilder<CORSConfig>.buildPlugin() {
     val allowNonSimpleContentTypes: Boolean = pluginConfig.allowNonSimpleContentTypes
     val headersList = pluginConfig.headers.filterNot { it in CORSConfig.CorsSimpleRequestHeaders }
         .let { if (allowNonSimpleContentTypes) it + HttpHeaders.ContentType else it }
-    val methodsListHeaderValue = methods.filterNot { it in CORSConfig.CorsDefaultMethods }
-        .map { it.value }
-        .sorted()
-        .joinToString(", ")
+    val methodsListHeaderValue = methods.map { it.value }.sorted().joinToString(", ")
```
bjhham, line 57: One thing I noticed in the other PR is they included a `distinct()` here - wondering if we'll end up with duplicates?

ktor-server/ktor-server-tests/common/test/io/ktor/tests/server/plugins/CORSTest.kt
```diff
@@ -1592,4 +1592,32 @@ class CORSTest {
...
+        }
+
+        assertEquals(HttpStatusCode.OK, response.status)
+        val allowMethodsHeader = response.headers[HttpHeaders.AccessControlAllowMethods]
+
+        assertNotNull(allowMethodsHeader, "Access-Control-Allow-Methods header should not be null")
+        assertTrue(
+            allowMethodsHeader.contains("POST"),
+            "Expected Access-Control-Allow-Methods to include POST, but got: $allowMethodsHeader"
+        )
```
bjhham: I think we should have an assertEquals here to ensure the default methods are included as well in a predictable way.

HMH-3080:
@bjhham 
You're absolutely right to wonder about that! However, since the underlying methods collection in CORSConfig is backed by a HashSet<HttpMethod>, the items are already guaranteed to be unique. Adding .distinct() would just introduce an unnecessary overhead.

As for the predictability, I agree. I will add back the .sorted() call to ensure a consistent header value order, and update the test case to use assertEquals with the expected full string. Pushing the updates shortly

PR #5634, opened 2026-05-19
KTOR-9561 Rethrow closedCause after copyTo

fru1tworld:
**Subsystem**
ktor-io

**Motivation**
[KTOR-9561](https://youtrack.jetbrains.com/issue/KTOR-9561) ByteReadChannel.copyTo does not propagate source closedCause on normal exit

**Solution**
Call `rethrowCloseCauseIfNeeded()` in both `copyTo` overloads, matching `readTo`/`readRemaining`.

### DIFF
--- a/ktor-io/common/src/io/ktor/utils/io/ByteChannel.kt
+++ b/ktor-io/common/src/io/ktor/utils/io/ByteChannel.kt
@@ -96,9 +96,14 @@ public class ByteChannel(public override val autoFlush: Boolean = false) : ByteR
         sleepWhile(Slot::Read) {
             flushBufferSize + _readBuffer.size < min && _closedCause.value == null
         }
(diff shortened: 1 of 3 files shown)


bjhham (changes requested):
Could you try the fix proposed in the YT ticket:

> rethrowCloseCauseIfNeeded() should be made private to ByteChannel and called at the end of awaitContent() (after sleepWhile exits), so that any cancellation cause is always propagated regardless of whether the caller was suspended or executing between iterations. This would make copyTo (and all other readers built on awaitContent()) correct by default.

> This is a more central refactor than the immediate workaround and should be addressed in a dedicated change.

This would ensure all looped flows like this are covered.

bjhham (approved):
@fru1tworld thx, please rebase and resolve conflicts

PR #5635, opened 2026-05-19
Allow external test services during startup

zibet27:
**Subsystem**
Server Auth

**Motivation**

Needed for testing [KTOR-5001](https://youtrack.jetbrains.com/issue/KTOR-5001) Add OpenID Connect (OIDC) support on client and server

**Solution**
This fixes `testApplication` so external services registered with `externalServices { ... }` are available while the tested application is starting.

Previously, calls made during startup could fail because the delegated test client engine was not yet wired to the external services routing. This PR makes the delegation available early enough and adds a regression test covering startup-time external service calls.


bjhham (approved):
Nice 👍

e5l (approved):

PR #5636, opened 2026-05-19
KTOR-9425 Use cipher block size for session IV

fru1tworld:
**Subsystem**
ktor-server-sessions

**Motivation**
[KTOR-9425](https://youtrack.jetbrains.com/issue/KTOR-9425) SessionTransportTransformerEncrypt init block uses wrong IV size for AES-256

**Solution**
Use cipher block size instead of `encryptionKeySize` for IV generation in `init` and `transformWrite`.

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-sessions/jvm/src/io/ktor/server/sessions/SessionTransportTransformerEncrypt.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-sessions/jvm/src/io/ktor/server/sessions/SessionTransportTransformerEncrypt.kt
@@ -58,7 +58,10 @@ public class SessionTransportTransformerEncrypt(
 
     // AES CBC IV length always equals the cipher block size (16 bytes for AES),
     // regardless of the key length. Using encryptionKeySize here used to break AES-256.
-    private val ivSize: Int = Cipher.getInstance("$encryptAlgorithm/CBC/PKCS5PADDING").blockSize
+    private val ivSize: Int = when (encryptionKeySpec.algorithm.uppercase()) {
+        "AES" -> 16
+        else -> Cipher.getInstance("$encryptAlgorithm/CBC/PKCS5PADDING").blockSize
+    }
 
     // Check that input keys are right
     init {
--- a/ktor-server/ktor-server-plugins/ktor-server-sessions/jvm/src/io/ktor/server/sessions/SessionTransportTransformerEncrypt.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-sessions/jvm/src/io/ktor/server/sessions/SessionTransportTransformerEncrypt.kt
@@ -56,9 +56,13 @@ public class SessionTransportTransformerEncrypt(
      */
     public val encryptionKeySize: Int get() = encryptionKeySpec.encoded.size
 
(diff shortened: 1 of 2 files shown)


ktor-server/ktor-server-plugins/ktor-server-sessions/jvm/src/io/ktor/server/sessions/SessionTransportTransformerEncrypt.kt
```diff
@@ -56,9 +56,13 @@ public class SessionTransportTransformerEncrypt(
      */
     public val encryptionKeySize: Int get() = encryptionKeySpec.encoded.size
 
+    // AES CBC IV length always equals the cipher block size (16 bytes for AES),
+    // regardless of the key length. Using encryptionKeySize here used to break AES-256.
+    private val ivSize: Int = Cipher.getInstance("$encryptAlgorithm/CBC/PKCS5PADDING").blockSize
```
bjhham: We can avoid the cipher lookup here for AES since it's always 128 bits.

```suggestion
    // AES CBC IV length always equals the cipher block size (16 bytes for AES),
    // regardless of the key length. Using encryptionKeySize here used to break AES-256.
    private val ivSize: Int = when(encryptionKeySpec.algorithm.uppercase()) {
        "AES" -> 16
        else -> Cipher.getInstance("$encryptAlgorithm/CBC/PKCS5PADDING").blockSize
    }
```

ktor-server/ktor-server-plugins/ktor-server-sessions/jvm/test/io/ktor/tests/sessions/SessionTransportTransformerEncryptTest.kt
```diff
@@ -0,0 +1,38 @@
...
+    @Test
+    fun `round-trip preserves payload for AES-128, AES-192 and AES-256`() {
+        for (keyLength in intArrayOf(16, 24, 32)) {
+            val transformer = SessionTransportTransformerEncrypt(ByteArray(keyLength), ByteArray(32))
+            val original = "session-value-$keyLength"
+            val encoded = transformer.transformWrite(original)
+            val decoded = transformer.transformRead(encoded)
+            assertEquals(original, decoded, "round-trip failed for AES-${keyLength * 8}")
+        }
+    }
```
bjhham, line 38: Let's include a non-AES test here as well.

```
    @Test
    fun `uses cipher block size for non-AES algorithm (DESede)`() {
        val encryptionKeySpec = SecretKeySpec(ByteArray(24), "DESede")
        val signKeySpec = SecretKeySpec(ByteArray(32), "HmacSHA256")

        val transformer = SessionTransportTransformerEncrypt(
            encryptionKeySpec = encryptionKeySpec,
            signKeySpec = signKeySpec,
        )

        val original = "non-aes-session-value"
        val encoded = transformer.transformWrite(original)
        val iv = encoded.substringBefore('/').hexToByteArray()

        assertEquals(8, iv.size, "DESede should use an 8-byte IV matching its block size")
        assertEquals(original, transformer.transformRead(encoded))
    }
```

bjhham (changes requested):

bjhham (approved):

fru1tworld:
thanks for feedback 😃

PR #5638, opened 2026-05-19
KTOR-9386 Show actual Unix socket path in startup log

fru1tworld:
**Subsystem**
ktor-server-core, ktor-server-cio

**Motivation**
[KTOR-9386](https://youtrack.jetbrains.com/issue/KTOR-9386) Misleading log output for Unix Domain Socket servers: "Responding at unix://0.0.0.0:80"

**Solution**
Add optional `EngineConnectorConfig.addressDescription` (default `null`). `UnixSocketConnectorBuilder` overrides it to `unix://$socketPath`; `EmbeddedServer` uses it when non-null.


bjhham (approved):
Thanks, I've been meaning to get around to this 😄

PR #5642, opened 2026-05-19
Align JS toolchain between local machines and CI

osipxd:
**Subsystem**
Infrastructure

**Motivation**
CI fails because of `yarn.lock` mismatch caused by using different versions of Yarn.

**Solution**
Use the same version of Yarn on CI and locally.


.devcontainer/Dockerfile
```diff
@@ -56,17 +56,18 @@ ENV PATH="$local_bin:$PATH"
...
 
 RUN <<EOT
   # Install NVM
-  curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v$nvm_version/install.sh | bash
-  source $NVM_DIR/nvm.sh
+  curl -o- "https://raw.githubusercontent.com/nvm-sh/nvm/v${nvm_version}/install.sh" | bash
+  source "$NVM_DIR/nvm.sh"
 
-  # Install Node and Yarn Classic
-  nvm install "$node_version"
```
osipxd, line 68: Not needed anymore. When `NODE_VERSION` is specified, nvm's install.sh automatically installs this Node.js version

osipxd:
JS and WasmJS tests fail, but not because of this PR.
We have one failing test in `:ktor-client-webrtc`. It might be related to coroutines update. I'm going to fix this in a separate PR.

bjhham (approved):

PR #5643, opened 2026-05-20
KTOR-9601 Fix nullable value class schema generation

sya-ri:
**Subsystem**  
Server/OpenAPI, `ktor-openapi-schema`, `ktor-compiler-plugin`

**Motivation**  
Fixes [KTOR-9601](https://youtrack.jetbrains.com/issue/KTOR-9601).

Ktor OpenAPI correctly emits nullable primitive schemas such as `String?` as `type: ["string", "null"]`, but it drops nullability for nullable `@JvmInline value class` properties. For example, `SessionToken?` was being emitted as `type: "string"` instead of `["string", "null"]`.

This affects generated request/response schemas and makes nullable value-class fields appear non-null in OpenAPI output.

**Solution**  
Preserve nullability when unwrapping inline/value-class descriptors during schema generation.

Instead of unwrapping a nullable inline descriptor and losing the original nullability, the schema now reapplies nullability to the unwrapped primitive schema type. This keeps behavior unchanged for non-null value classes while fixing cases like:

- `SessionToken` -> `type: "string"`
- `SessionToken?` -> `type: ["string", "null"]`
- non-null value class with nullable underlying value -> nullable schema is preserved

Added a minimal OpenAPI snapshot test case in `ktor-compiler-plugin/testData/openapi/NullableValueClasses.kt` covering request/response DTOs for:
- nullable plain string
- non-null value class
- nullable value class
- non-null value class wrapping nullable string
- nullable integer-backed value class


ktor-compiler-plugin/testData/openapi/NullableValueClasses.actual.json
```diff
@@ -0,0 +1,99 @@
...
+    "paths": {
+        "/nullable-value-classes": {
+            "post": {
+                "summary": "",
+                "requestBody": {
+                    "content": {
+                        "application/json": {
+                            "schema": {
+                                "$ref": "#/components/schemas/ExampleRequest"
+                            }
```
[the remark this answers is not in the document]
sya-ri: Fixed in ca3bbb5b86ded7e1377c7522aa35d394f9b9265c

ktor-shared/ktor-openapi-schema/common/src/io/ktor/openapi/JsonSchemaInference.kt
```diff
@@ -385,6 +385,17 @@ public fun JsonSchema.wrapIfNullable(isNullable: Boolean): JsonSchema =
...
     }
 
+private fun JsonSchema.withNullableType(isNullable: Boolean): JsonSchema =
+    if (!isNullable) {
+        this
+    } else {
+        when (val schemaType = type) {
+            is JsonType -> copy(type = schemaType.wrapIfNullable(true))
+            is SchemaType.AnyOf -> this
+            null -> wrapIfNullable(true)
```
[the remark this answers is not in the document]
sya-ri: Fixed in 04b6d275491654067f10956f0983fac6b99c1fae

ktor-compiler-plugin/testData/openapi/NullableValueClasses.actual.json
```diff
@@ -0,0 +1,113 @@
+{
```
bjhham: Could you delete this `actual` file?  I recently migrated the compiler plugin but forgot to transfer the `.gitignore` entries for these artifacts.

ktor-shared/ktor-openapi-schema/common/src/io/ktor/openapi/JsonSchemaInference.kt
```diff
@@ -385,6 +385,22 @@ public fun JsonSchema.wrapIfNullable(isNullable: Boolean): JsonSchema =
         this
     }
 
+@OptIn(InternalAPI::class)
+private fun JsonSchema.withNullableType(isNullable: Boolean): JsonSchema =
```
bjhham: Could you instead modify `fun JsonSchema.wrapIfNullable(Boolean): JsonSchema` to include this logic?  It seems like this should be something that's consistently applied, and it will simplify the current `oneOf` usage in the models.

bjhham (changes requested):
Thanks!  Had a couple minor comments.

ktor-shared/ktor-openapi-schema/common/src/io/ktor/openapi/JsonSchemaInference.kt
bjhham: Also, it looks like the test should be included in:

```
io.ktor.openapi.reflect.AbstractSchemaInferenceTest
```

Since the fix was in the inference code and not the compiler plugin.  This will also check if we're applying consistent logic with the reflection inference.

ktor-compiler-plugin/testData/openapi/NullableValueClasses.actual.json
sya-ri: ~~Removed the generated `actual.json` snapshot from version control and added a `.gitignore` entry for `ktor-compiler-plugin/testData/openapi/*.actual.json` so these generated artifacts do not get committed again.~~

cfa64fef29fd081dfbe89d63f8b31c317c35e9fd

ktor-shared/ktor-openapi-schema/common/src/io/ktor/openapi/JsonSchemaInference.kt
sya-ri: Moved the nullable handling into `JsonSchema.wrapIfNullable()` so the logic is applied consistently after value-class unwrapping, including primitive and `AnyOf` schema types. I also added a shared regression test to `AbstractSchemaInferenceTest` so both inference implementations verify nullable value classes in the same place.

5e3d82f5820ee5142b795fca4593ac5a813b51e9

ktor-compiler-plugin/testData/openapi/NullableValueClasses.actual.json
sya-ri: The .gitignore change from this PR is no longer needed, because the base branch now already ignores generated *.actual.json files more broadly. I dropped the PR-specific ignore entry during the rebase and kept the upstream rule instead.

f9581bacbc9b78d5ec2825acbb709cb0099861ab

sya-ri:
@bjhham Thank you for the review. I addressed the review comments and rebased onto the latest `release/3.x`.

bjhham (approved):
Looks good!

If you'd please resolve the new conflict, I'll merge for inclusion in 3.5.1

sya-ri:
@bjhham I resolved the conflict. It should be ready to merge now :rocket:

PR #5645, opened 2026-05-20
Align default test timeout

osipxd:
**Subsystem**
Test Infrastructure

**Motivation**
- Hanging JS tests fail after 30 seconds because of browser connection timeout applied by Karma, but wee see a misleading error for the whole module instead of individual test failures:
   ```
   There are test sources present and no filters are applied, but the test task did not discover any tests to execute.
   This is likely due to a misconfiguration. Please check your test configuration. If this is not a misconfiguration,
   this error can be disabled by setting the 'failOnNoDiscoveredTests' property to false.
   ```
- CI detects hanging tests to late, wasting time

**Solution**
Make default timeout equal to the browser disconnect timeout (30 sec). This also makes CI to better detect hanging tests.


ktor-shared/ktor-test-base/common/src/io/ktor/test/TestBuilders.kt
```diff
@@ -0,0 +1,31 @@
...
+import kotlinx.coroutines.test.TestResult
+import kotlinx.coroutines.test.TestScope
+import kotlin.coroutines.CoroutineContext
+import kotlin.coroutines.EmptyCoroutineContext
+import kotlin.time.Duration
+import kotlinx.coroutines.test.runTest as runTestImpl
+
+/**
+ * An override of [kotlinx.coroutines.test.runTest] using our [DEFAULT_TEST_TIMEOUT].
+ * This function should be used to aligne timeout across all tests.
```
osipxd:
```suggestion
 * This function should be used to align timeout across all tests.
```

bjhham (approved):
Good call 👍 

Could you merge release/3.x to main after merging this?  Otherwise, we could be dealing with an excess of conflicts from all the import cleanups.

osipxd:
Yes, will do

PR #5650, opened 2026-05-22
Fix webrtc test race condition

zibet27:
**Subsystem**
WebRTC Client

**Motivation**
[KTOR-9603](https://youtrack.jetbrains.com/issue/KTOR-9603) Structured concurrency is violated in WebRTC tests

Previously, background tasks (listening for events, etc.) were stored in the `tasks` mutable list and then cancelled when the test was done.

**Solution**
- add `BackgroundTasksScope` and `withBackgroundTasks` to explicitly show the block that can have background tasks
- add context parameters support for tests only to nicely pass `BackgroundTasksScope`

Though I'm not 100% sure it follows all structured concurrency rules


ktor-client/ktor-client-webrtc/build.gradle.kts
```diff
@@ -59,21 +59,21 @@ kotlin {
             api(libs.kotlinx.serialization.core)
         }
 
-        commonTest.dependencies {
-            implementation(libs.kotlin.test)
-            implementation(project(":ktor-test-dispatcher"))
+        commonTest {
+            compilerOptions {
```
osipxd: These compiler options are actually applied to the `kotlin {}` block. I think the best way to apply them only to test tasks is to configure test compilation tasks:

```kotlin
tasks.withType<KotlinCompilationTask<*>>().configureEach {
    if (name.contains("Test", ignoreCase = true)) {
        compilerOptions {
            freeCompilerArgs.add("-Xexpect-actual-classes")
        }
    }
}
```

ktor-client/ktor-client-webrtc/build.gradle.kts
```diff
@@ -59,21 +59,21 @@ kotlin {
...
+                freeCompilerArgs.add("-Xcontext-parameters")
+            }
+
+            dependencies {
+                implementation(libs.kotlin.test)
+                implementation(project(":ktor-test-dispatcher"))
```
osipxd: Let's use ktor-test-base here 
```suggestion
                implementation(projects.ktorTestBase)
```

ktor-client/ktor-client-webrtc/common/test/io/ktor/client/webrtc/utils/ConnectionUtils.kt
```diff
@@ -24,62 +21,53 @@ expect fun createTestWebRtcClient(): WebRtcClient
...
+        coroutineContext.cancelChildren()
+    }
+}
+
+suspend fun CoroutineScope.withBackgroundTasks(block: suspend BackgroundTasksScope.() -> Unit) {
+    val scope = BackgroundTasksScope(parent = this)
+    try {
+        block(scope)
+    } finally {
+        scope.stopBackgroundTasks()
```
osipxd: Since `stopBackgroundTasks` should be used only from here I think we can inline it and leave `BackgroundTasksScope` without any members

kotlin-js-store/yarn.lock
```diff
@@ -2407,7 +2578,12 @@ wrappy@1:
   resolved "https://registry.yarnpkg.com/wrappy/-/wrappy-1.0.2.tgz#b5243d8f3ec1aa35f1364605bc0d1036e30ab69f"
   integrity sha512-l4Sp/DRseor9wL6EvV2+TuQn63dMkPjZ/sp9XkghTEbV9KlPS1xUsZ3u7/IQO4wxtcFB4bgpQPRcR3QCvezPcQ==
 
-ws@8.20.1, ws@^8.20.0:
+ws@8.18.3, ws@^8.18.3:
+  version "8.18.3"
+  resolved "https://registry.yarnpkg.com/ws/-/ws-8.18.3.tgz#b56b88abffde62791c639170400c93dcb0c95472"
+  integrity sha512-PEIGCY5tSlUt50cqyMXfCzX+oOPqN0vuGqWzbcJ2xvnkzkq46oOpz7dQaTDBdfICb4N14+GARUDw2XV2N4tvzg==
+
+ws@8.20.1:
```
osipxd: It's weird, but on CI `yarn.lock` is generated without this entry 🤔 
Locally I don't have this either

osipxd (changes requested):
Thank you for the fix! I have some minor comments

ktor-client/ktor-client-webrtc/build.gradle.kts
zibet27: Looks like `ktorTestBase` not compatible with AGP :(

```
:ktor-client-webrtc:androidDeviceTest: Could not resolve org.jetbrains.kotlin:kotlin-test-junit:2.3.21.
```

ktor-client/ktor-client-webrtc/build.gradle.kts
zibet27: Good catch!
I didn't check what `this` is (

osipxd (approved):
👍

PR #5652, opened 2026-05-22
KTOR-9606 Fix KotlinxSerializationConverter deserialization of empty channel

Stexxe:
https://youtrack.jetbrains.com/issue/KTOR-9606

### DIFF
--- a/ktor-shared/ktor-serialization/ktor-serialization-kotlinx/common/src/io/ktor/serialization/kotlinx/KotlinxSerializationConverter.kt
+++ b/ktor-shared/ktor-serialization/ktor-serialization-kotlinx/common/src/io/ktor/serialization/kotlinx/KotlinxSerializationConverter.kt
@@ -62,6 +62,8 @@ public class KotlinxSerializationConverter(
             return ext.deserialize(charset, typeInfo, ByteReadChannel(contentPacket)) ?: continue
         }
 
+        if (contentPacket.exhausted()) return null
+
         val serializer = format.serializersModule.serializerForTypeInfo(typeInfo)
 
         try {
--- a/ktor-shared/ktor-serialization/ktor-serialization-kotlinx/common/src/io/ktor/serialization/kotlinx/KotlinxSerializationConverter.kt
+++ b/ktor-shared/ktor-serialization/ktor-serialization-kotlinx/common/src/io/ktor/serialization/kotlinx/KotlinxSerializationConverter.kt
@@ -57,10 +57,10 @@ public class KotlinxSerializationConverter(
     override suspend fun deserialize(charset: Charset, typeInfo: TypeInfo, content: ByteReadChannel): Any? {
         val contentPacket = content.readRemaining()
 
-        val fromExtension = extensions.asFlow()
-            .map { it.deserialize(charset, typeInfo, ByteReadChannel(contentPacket)) }
-            .firstOrNull { it != null || contentPacket.exhausted() }
-        if (extensions.isNotEmpty() && (fromExtension != null || contentPacket.exhausted())) return fromExtension
+        for (ext in extensions) {
+            if (contentPacket.exhausted()) return null
+            return ext.deserialize(charset, typeInfo, ByteReadChannel(contentPacket)) ?: continue
+        }
 
         val serializer = format.serializersModule.serializerForTypeInfo(typeInfo)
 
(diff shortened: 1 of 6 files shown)


ktor-shared/ktor-serialization/ktor-serialization-kotlinx/common/src/io/ktor/serialization/kotlinx/KotlinxSerializationConverter.kt
```diff
@@ -58,13 +55,14 @@ public class KotlinxSerializationConverter(
...
     override suspend fun deserialize(charset: Charset, typeInfo: TypeInfo, content: ByteReadChannel): Any? {
+        val contentPacket = content.readRemaining()
+
         val fromExtension = extensions.asFlow()
-            .map { it.deserialize(charset, typeInfo, content) }
-            .firstOrNull { it != null || content.isClosedForRead }
-        if (extensions.isNotEmpty() && (fromExtension != null || content.isClosedForRead)) return fromExtension
+            .map { it.deserialize(charset, typeInfo, ByteReadChannel(contentPacket)) }
+            .firstOrNull { it != null || contentPacket.exhausted() }
+        if (extensions.isNotEmpty() && (fromExtension != null || contentPacket.exhausted())) return fromExtension
```
bjhham: We can improve readability and performance here with a loop:

```suggestion
        for (ext in extensions) {
            if (contentPacket.exhausted()) return null
            return ext.deserialize(charset, typeInfo, ByteReadChannel(contentPacket)) ?: continue
        }
        if (contentPacket.exhausted()) return null
```

bjhham (changes requested):

ktor-shared/ktor-serialization/ktor-serialization-kotlinx/common/src/io/ktor/serialization/kotlinx/KotlinxSerializationConverter.kt
```diff
@@ -58,13 +55,14 @@ public class KotlinxSerializationConverter(
...
-        val fromExtension = extensions.asFlow()
-            .map { it.deserialize(charset, typeInfo, content) }
-            .firstOrNull { it != null || content.isClosedForRead }
-        if (extensions.isNotEmpty() && (fromExtension != null || content.isClosedForRead)) return fromExtension
+        val contentPacket = content.readRemaining()
+
+        for (ext in extensions) {
+            if (contentPacket.exhausted()) return null
+            return ext.deserialize(charset, typeInfo, ByteReadChannel(contentPacket)) ?: continue
+        }
```
osipxd, line 63: Should we also add `if (contentPacket.exhausted()) return null` after the loop? For the case when extensions list is empty. The test for such a case could look like the following:

 ```kotlin                                                                                                                                                  
   @Test                                                                                                                                                    
   fun `returns null for empty channel without extensions`() = runTest {                                                                                    
       val channel = ByteChannel()                                                                                                                          
       launch {                                                                                                                                             
           delay(200.milliseconds)                                                                                                                          
           channel.close()                                                                                                                                  
       }                                                                                                                                                    
                                                                                                                                                            
       val converter = KotlinxSerializationConverter(EmptyExtensionStringFormat)                                                                            
       val result = converter.deserialize(Charsets.UTF_8, typeInfo<Payload?>(), channel)                                                                    
                                                                                                                                                            
       assertNull(result)                                                                                                                                   
   }                                                                                                                                                        
                                                                                                                                                            
   private object EmptyExtensionStringFormat : StringFormat {                                                                                               
       override val serializersModule: SerializersModule = EmptySerializersModule()                                                                         
                                                                                                                                                            
       override fun <T> encodeToString(serializer: SerializationStrategy<T>, value: T): String =                                                            
           error("Not used")                                                                                                                                
                                                                                                                                                            
       override fun <T> decodeFromString(deserializer: DeserializationStrategy<T>, string: String): T =                                                     
           fail("Empty content should be handled before decoding, got: '$string'")                                                                          
   }                                                                                                                                                        
 ```

osipxd (changes requested):
Offtop:
It feels weird that we have three ways receive payload:
1. `call.receive<Payload>()`
2. `call.receiveNullable<Payload>()`
3. `call.receiveNullable<Payload?>()`

But this is out of the scope of this PR.

ktor-shared/ktor-serialization/ktor-serialization-kotlinx/common/src/io/ktor/serialization/kotlinx/KotlinxSerializationConverter.kt
bjhham: +1

osipxd (approved):

osipxd:
It seems that my comment was actually wrong. This commit https://github.com/ktorio/ktor/commit/eb14d5347ca7dafdcc58d2c110aff1d895668f9e has added `extensions.isNotEmpty()` intentionally, so last commit should be probably dropped

bjhham (approved):

PR #5654, opened 2026-05-24
WebRTC Client: retain iOS delegate to keep messages flowing

teancom:
**Subsystem**
ktor-client-webrtc (iOS specific)

**Motivation**
We're converting the Music Assistant mobile app over to ktor's webrtc library. As soon as we had a 'large' WebRTC message (we pass album art over the WebRTC data channel, as a for-instance) the RTCDataChannel.delegate and RTCPeerConnection.delegate would get reclaimed and callbacks / messages would silently stop flowing. This is because iOS's `RTCDataChannel.delegate` and `RTCPeerConnection.delegate` are declared `@property(nonatomic, weak)`. The iOS implementation passed an anonymous `NSObject` to those setters without keeping a Kotlin-side strong reference.

**Solution**
Store the delegate in a private property on the wrapper class so it lives as long as the wrapper does. `IosWebRtcConnection` uses a `val` since the delegate is created in `init`; `IosWebRtcDataChannel` uses a `lateinit var` because `setupEvents()` is called after construction, with a `check` to enforce the single-assignment invariant.

Also, we added `IosDelegateRetentionTest` which forces `GC.collect()` after the channel is established and verifies both directions of traffic, plus that the peer connection still reaches CONNECTED.

### DIFF
--- a/ktor-client/ktor-client-webrtc/ios/src/io/ktor/client/webrtc/DataChannel.kt
+++ b/ktor-client/ktor-client-webrtc/ios/src/io/ktor/client/webrtc/DataChannel.kt
@@ -25,6 +25,11 @@ public class IosWebRtcDataChannel(
     receiveOptions: DataChannelReceiveOptions
 ) : WebRtcDataChannel(receiveOptions) {
 
+    // Apple's `RTCDataChannel.delegate` is `weak` — keep a strong reference on the
+    // Kotlin side so the anonymous delegate object stays alive for the connection's lifetime.
+    @Suppress("unused")
+    private lateinit var retainedDelegate: NSObject
+
     override val id: Int?
         get() = nativeChannel.channelId.let { if (it < 0) null else it }
 
@@ -88,7 +93,8 @@ public class IosWebRtcDataChannel(
     }
 
     internal fun setupEvents(eventsEmitter: WebRtcConnectionEventsEmitter) {
-        nativeChannel.setDelegate(object : RTCDataChannelDelegateProtocol, NSObject() {
+        check(!::retainedDelegate.isInitialized) { "setupEvents() must be called at most once per channel." }
+        val delegate = object : RTCDataChannelDelegateProtocol, NSObject() {
             override fun dataChannel(dataChannel: RTCDataChannel, didChangeBufferedAmount: ULong) =
                 runInConnectionScope {
                     // guard against duplicate events
@@ -123,7 +129,9 @@ public class IosWebRtcDataChannel(
             ) = runInConnectionScope {
                 emitMessage(didReceiveMessageWithBuffer.toKtor())
             }
-        })
+        }
+        retainedDelegate = delegate
+        nativeChannel.setDelegate(delegate)
     }
 }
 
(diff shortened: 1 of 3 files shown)


ktor-client/ktor-client-webrtc/ios/src/io/ktor/client/webrtc/DataChannel.kt
```diff
@@ -25,6 +25,11 @@ public class IosWebRtcDataChannel(
     receiveOptions: DataChannelReceiveOptions
 ) : WebRtcDataChannel(receiveOptions) {
 
+    // Apple's `RTCDataChannel.delegate` is `weak` — keep a strong reference on the
+    // Kotlin side so the anonymous delegate object stays alive for the connection's lifetime.
+    @Suppress("unused")
+    private lateinit var retainedDelegate: NSObject
```
zibet27: I would avoid lateinit var here and use a nullable field instead
Also, suppress is not needed as the delegate is checked in `setupEvents`

ktor-client/ktor-client-webrtc/ios/test/io/ktor/client/webrtc/IosDelegateRetentionTest.kt
```diff
@@ -0,0 +1,149 @@
...
+    ) { jobs ->
+        client.createPeerConnection().use { pc1 ->
+            client.createPeerConnection().use { pc2 ->
+                val pc2DataChannels = pc2.dataChannelEvents.collectToChannel(this, jobs)
+
+                val local = pc1.createDataChannel("delegate-retention-local")
+                connect(pc1, pc2, jobs)
+
+                // Wait for pc2 to receive the remote channel.
+                val remote = withTimeout(5000) {
```
zibet27: `kotlin.Duration` would be a bit nicer here

ktor-client/ktor-client-webrtc/ios/src/io/ktor/client/webrtc/PeerConnection.kt
```diff
@@ -31,8 +31,15 @@ public class IosWebRtcConnection(
 ) : WebRtcPeerConnection(coroutineContext, config) {
     internal val peerConnection: RTCPeerConnection
 
+    // Apple's `RTCPeerConnection.delegate` is `weak` — keep a strong reference on the
+    // Kotlin side so the anonymous delegate object stays alive for the connection's lifetime.
+    @Suppress("unused")
+    private val retainedDelegate: NSObject
```
zibet27: The same as for datachannel
Btw, we can also null it on close, so further events don't arrive
PS: nulling the delegate at `IosWebRtcDataChannel` would break some logic as it's "async"

zibet27 (changes requested):
@teancom Thank you for the PR!
It also fixes the existing `receiveRemoteTracks` test, which I noticed started failing at some point
We don't test the iOS target on the CI :(

Just left some nitpick comments

zibet27:
[KTOR-9565](https://youtrack.jetbrains.com/issue/KTOR-9565) WebRTC: Peer stops receiving messages after a while on iOS

teancom:
Thanks for reviewing so quickly! I addressed your comments, let me know if there's anything else I can do. We're very excited to be able to switch over to ktor - it's going to greatly simplify our codebase and help us out a ton. 🙇

zibet27 (approved):
lgtm

zibet27:
> We're very excited to be able to switch over to ktor - it's going to greatly simplify our codebase and help us out a ton. 🙇

That's awesome!
Let us know if you find any other bugs or need more features

osipxd (approved):

PR #5655, opened 2026-05-25
KTOR-9615 Apache5: Close the underlying request on job cancel

osipxd:
**Subsystem**
`ktor-client-apache5`

**Motivation**
[KTOR-9615](https://youtrack.jetbrains.com/issue/KTOR-9615) Apache5: The underlying request is not aborted when coroutine Job is cancelled

**Solution**
Cancel the request `Future` when coroutine job is cancelled.

The result can be observed [here](https://ktor.teamcity.com/test/1628683479014970233?currentProjectId=Ktor_ProjectKtorCore). Previously this test succeeded only after retry.


zibet27 (approved):
Nice!

PR #5657, opened 2026-05-25
KTOR-9595 Fix ktor-network compatibility with jsBrowser and wasmJsBrowser

osipxd:
**Subsystem**
`ktor-network`

**Motivation**
[KTOR-9595](https://youtrack.jetbrains.com/issue/KTOR-9595) CIO: The engine imports `node:net` statically and breaks wasmJsBrowser webpack build

**Solution**
Use the approach with `await import` similar to [this commit](https://github.com/Kotlin/kotlinx-benchmark/commit/f7248c542778f87a8876e4c91913f6515b3c8368).

### DIFF
--- a/ktor-network/js/src/io/ktor/network/sockets/nodejs/node.net.js.kt
+++ b/ktor-network/js/src/io/ktor/network/sockets/nodejs/node.net.js.kt
@@ -5,8 +5,26 @@
 package io.ktor.network.sockets.nodejs
 
 import io.ktor.network.sockets.*
+import kotlinx.coroutines.await
 import org.khronos.webgl.ArrayBuffer
 import org.khronos.webgl.Int8Array
+import kotlin.js.Promise
+
+internal actual suspend fun loadNodeNet(): NodeNet = nodeNetPromise.await()
+    ?: throw UnsupportedOperationException("Module node:net is not available. Please verify that you are using Node.js")
+
+private val nodeNetPromise: Promise<NodeNet?> by lazy {
+    loadNodeModule(nodeNetModuleName())
+}
+
+// Keep the module name behind a function so browser bundlers don't resolve `node:net` statically.
+private fun nodeNetModuleName() = "node:net"
+
+@JsFun(
+    "moduleName => ((typeof process !== 'undefined') && process.release.name === 'node')" +
+        " ? import(moduleName) : Promise.resolve(null)"
+)
+private external fun loadNodeModule(moduleName: String): Promise<NodeNet?>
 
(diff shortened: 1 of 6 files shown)


ktor-network/js/src/io/ktor/network/sockets/nodejs/node.net.js.kt
```diff
@@ -5,8 +5,26 @@
...
+private val nodeNetPromise: Promise<NodeNet?> by lazy {
+    loadNodeModule(nodeNetModuleName())
+}
+
+// Keep the module name behind a function so browser bundlers don't resolve `node:net` statically.
+private fun nodeNetModuleName() = "node:net"
+
+@JsFun(
+    "moduleName => ((typeof process !== 'undefined') && process.release && (process.release.name === 'node'))" +
+        " ? import(moduleName) : Promise.resolve(null)"
```
bjhham: Could use optional chaining to clean up these functions a little `process?.release?.name == 'node'`

bjhham (approved):

ktor-network/web/src/io/ktor/network/sockets/SocketEngine.web.kt
```diff
@@ -12,18 +12,24 @@ internal actual suspend fun tcpConnect(
     selector: SelectorManager,
     remoteAddress: SocketAddress,
     socketOptions: SocketOptions.TCPClientSocketOptions
-): Socket = suspendCancellableCoroutine { cont ->
-    val socket = nodeNet.createConnection(CreateConnectionOptions(remoteAddress, socketOptions))
-    SocketContext(socket, remoteAddress, null).initiate(cont)
+): Socket {
+    val nodeNet = loadNodeNet()
```
zibet27, line 16: Can we also use a lazy value here?

osipxd: Unfortunately not as I failed to make `loadNodeNet` non suspending in JS. But `nodeNetPromise` is lazy, so it'll be evaluated only once.

ktor-network/js/src/io/ktor/network/sockets/nodejs/node.net.js.kt
osipxd: Thanks! I've managed to replace it with `(typeof process !== 'undefined') && process.release?.name === 'node'`, but can't get rid of `typeof process !== 'undefined'`

UPD: Ooops. `JsFun` fails to parse optional chaining :(

zibet27 (approved):

PR #5661, opened 2026-05-26
KTOR-9596 Fix timeout failure regression

### LINKED ISSUE #5651: HttpTimeout regression in 3.5.0: requests wait full timeout when KtorClientTelemetry (OTel) plugin is installed
## Summary

When both `HttpTimeout` and `KtorClientTelemetry` (OpenTelemetry) plugins are installed on a Ktor 3.5.0 HTTP client, every request takes exactly `requestTimeoutMillis` before completing — even though the server responds immediately.

This is a regression from Ktor 3.4.3 where the same combination works correctly.

## Reproduction

Standalone repro project: https://github.com/anderssv/ktor-timeout-repro

```bash
git clone https://github.com/anderssv/ktor-timeout-repro
cd ktor-timeout-repro
./gradlew test
```

Results:
- ✅ `request WITHOUT OTel plugin completes quickly` — PASSES
- ❌ `request WITH OTel plugin hits timeout` — FAILS (takes exactly 3000ms)
- ❌ `concurrent requests WITH OTel plugin all hit timeout` — FAILS

## Root Cause

Ktor 3.5.0 changed `HttpTimeout` to launch the timeout coroutine in `CoroutineScope(callerContext)`. OpenTelemetry's `KtorClientTelemetryUtil.installSpanEnd` does:

```kotlin
scope.launch { job.join() }
```

in the receive pipeline, which keeps the call's `Job` alive. The timeout fires because the job appears to still be running even after the response is received.

## Versions

- Ktor: 3.5.0 (works on 3.4.3)
- OpenTelemetry Instrumentation: 2.28.1-alpha (also tested with 2.23.0-alpha — same result)
- Kotlin: 2.3.21
- Engines tested: OkHttp, Apache (same behavior on both)

## Workaround

Stay on Ktor 3.4.x, or remove `KtorClientTelemetry` from clients that also use `HttpTimeout`.

### LINKED ISSUE #5678: Regression in 3.5.0: Android Ktor client request is cancelled when WebSocket/STOMP subscription is active
### Ktor version

Broken: `3.5.0`

Working:
- `3.4.3`
- `3.3.3`

### Platform

Android.

### Engine / client construction

The client is constructed without explicitly passing an engine:

```kotlin
HttpClient {
    install(ContentNegotiation)
    install(HttpTimeout)
    install(WebSockets)
    install(Logging)
}
```

So this is using Ktor's default engine selection for the Android build.

### Description

After upgrading Ktor from `3.4.3` to `3.5.0`, a REST request made immediately after starting a WebSocket/STOMP subscription is consistently cancelled with `JobCancellationException`. This might have nothing to do with Krossbow that I use for STOMP connection , but worth mentioning for full disclousre

The same code works correctly on `3.4.3` and `3.3.3`.

The REST request reaches the server and the server returns HTTP 200. With logging enabled, I can see the response reaching the Ktor client, but before my code receives the response, the request coroutine is cancelled.

This happens deterministically on every app launch in this flow.

### Setup

The client uses:

- `ktor-client-websockets`
- `ContentNegotiation`
- `HttpTimeout`
- `Logging`
- Krossbow STOMP adapter over Ktor WebSockets

The startup flow is roughly:

1. Connect WebSocket/STOMP.
2. Emit a subscription payload for a matter/topic.
3. Start collecting messages from the WebSocket subscription.
4. Immediately perform a REST GET for the same matter's initial messages.
5. The server returns HTTP 200.
6. Ktor 3.5.0 cancels the REST request before user code receives the response.

### Expected behavior

The REST request should complete normally and return the HTTP 200 response body, as it does with Ktor `3.4.3` and `3.3.3`.

### Actual behavior

The REST request is cancelled with:

```text
JobCancellationException: StandaloneCoroutine was cancelled
```

My code never reaches the line after the REST call.

### Relevant logs

With Ktor `3.5.0`:

```text
TRACE_MESSAGES before getMessages matter=4
TRACE_MESSAGES repo getAndInsertMessages ENTER matter=4
TRACE_MESSAGES networkCall ENTER responseType=List job=FlowCoroutine{Active} active=true

TRACE_MESSAGES stomp flatMapLatest payload=SubscriptionPayload(roomId=10, matterId=4, terminate=false, type=Matter)
TRACE_MESSAGES stomp callbackFlow START payload=SubscriptionPayload(roomId=10, matterId=4, terminate=false, type=Matter)
TRACE_MESSAGES stomp subscription job START payload=SubscriptionPayload(roomId=10, matterId=4, terminate=false, type=Matter)

Raw Message : SharedSocketMessage(... messageType=ATTENDANCE_UPDATE, matterId=4 ...)

HTTP Client: RESPONSE: 200
HTTP Client: METHOD: GET
HTTP Client: FROM: http://<host>/room/matter/4/messages

TRACE_MESSAGES networkCall CANCELLED responseType=List type=JobCancellationException message=StandaloneCoroutine was cancelled
TRACE_MESSAGES combine CANCELLED matter=4 type=JobCancellationException message=StandaloneCoroutine was cancelled
```

### Notes

This does not appear to be caused by lifecycle or collection cancellation:

- The parent app composable remains in composition.
- The parent socket manager scope remains active.
- The StateFlow subscription count does not drop to zero.
- The issue happens consistently when the app flow is stable.
- Downgrading only Ktor from `3.5.0` to `3.4.3` fixes the problem.

I also tested reducing Ktor logging from `LogLevel.ALL` to `LogLevel.HEADERS`; the issue still occurred.

I also tested using separate `HttpClient` instances for REST and WebSocket/STOMP; the issue still occurred.

### Minimal version matrix

```text
ktor = "3.3.3" -> works
ktor = "3.4.3" -> works
ktor = "3.5.0" -> broken
```

### Question

Was there a change in Ktor `3.5.0` around Android client engine selection, WebSockets, call cancellation, or body channel handling that could cause an active WebSocket receive path to cancel a concurrent REST request coroutine?

bjhham:
**Subsystem**
Client, HttpTimeout

**Motivation**
- [KTOR-9596](https://youtrack.jetbrains.com/issue/KTOR-9596) HttpTimeout: HttpRequestTimeoutException on any request with runTest and request timeout defined since 3.5.0

**Solution**
The behaviour introduced in https://github.com/ktorio/ktor/pull/5512 moved the killer task to the caller's context, but this causes more significant problems when working with virtual time and realtime.  I've reverted this back to use the client engine's dispatcher.


osipxd (approved):
👍 
Should we also reopen the previous issue?

bjhham:
>  Should we also reopen the previous issue?

I've just closed it as designed instead, since the fix is invalid behaviour.  If users want virtual time to effect the timeout plugin, they'll need to use the test engine or supply the test dispatcher to the client.

PR #5662, opened 2026-05-26
KTOR-9568 New limits for compression

bjhham:
**Subsystem**
Server, Compression / Client, ContentEncoding

**Motivation**
[KTOR-9568](https://youtrack.jetbrains.com/issue/KTOR-9568) ContentEncoding/Compression: Decompression bomb via chained Content-Encoding

**Solution**
Introduce suitable limits for compression chaining and decompressed payload size.  For now, payload size is disabled to avoid breakage, and compression chaining is limited to 2, since there's no good use case for 3+.


bjhham:
@coderabbitai review

ktor-client/ktor-client-plugins/ktor-client-encoding/common/src/ContentEncoding.kt
```diff
@@ -15,9 +15,28 @@ import io.ktor.util.*
...
 private val LOGGER = KtorSimpleLogger("io.ktor.client.plugins.compression.ContentEncoding")
 
+/**
+ * Default maximum number of chained content encodings allowed in a `Content-Encoding` response
+ * header.
+ *
+ * Helps mitigate "decompression bomb" / DoS attacks where a small response chains many
+ * encodings to expand into a huge payload.
+ */
+public const val DEFAULT_MAX_ENCODING_CHAIN_LENGTH: Int = 2
```
osipxd: Hm.. `io.ktor.server.plugins.CompressionTest.testDecoding()` doesn't pass this limit

ktor-client/ktor-client-plugins/ktor-client-encoding/api/ktor-client-encoding.api
```diff
@@ -1,14 +1,24 @@
+public final class io/ktor/client/plugins/compression/ContentEncodingChainTooLongException : java/lang/IllegalStateException {
```
osipxd: Since this PR introduces new APIs we have to target it to main, I'm afraid

ktor-server/ktor-server-plugins/ktor-server-compression/jvm/src/io/ktor/server/plugins/compression/Compression.kt
```diff
@@ -89,6 +179,13 @@ private suspend fun ContentDecoding.Context.decode(call: PipelineCall, options:
         return
     }
     val encoding = parseHeaderValue(encodingRaw)
+    if (maxEncodingChainLength > 0 && encoding.size > maxEncodingChainLength) {
+        LOGGER.trace(
+            "Rejecting decompression for ${call.request.uri} because the Content-Encoding chain " +
+                "of length ${encoding.size} exceeds the configured limit of $maxEncodingChainLength."
+        )
```
osipxd: Minor. `trace { "..." }` could be used here

ktor-server/ktor-server-plugins/ktor-server-compression/jvm/src/io/ktor/server/plugins/compression/Compression.kt
```diff
@@ -31,6 +34,86 @@ internal val LOGGER = KtorSimpleLogger("io.ktor.server.plugins.compression.Compr
...
+ */
+public const val DEFAULT_MAX_ENCODING_CHAIN_LENGTH: Int = 2
+
+/**
+ * Default maximum size, in bytes, of decompressed request content. Defaults to -1 (disabled).
+ *
+ * Helps mitigate "decompression bomb" / DoS attacks where a small compressed body
+ * decompresses to a huge payload.
+ */
+public const val DEFAULT_MAX_DECODED_CONTENT_LENGTH: Long = -1
```
osipxd: Probably we should put constants, exceptions and wrappers into some common module for encoding shared between server and client. WDYT?

osipxd (changes requested):
👍 
Please see the comments. I think we should re-target this PR to main

ktor-client/ktor-client-plugins/ktor-client-encoding/api/ktor-client-encoding.api
bjhham: I was thinking we could make an exception for this since it's needed to patch a vulnerability.

osipxd: We could make only plugin config properties public while keeping constants and exceptions internal. This would make it safer for a patch version with a cost of a bit worse DX. And this would also leave us an opportunity to move shared entities into some shared module in the next minor version.

bjhham: Yeah, that makes sense.  I'll minimize the public API as much as possible.

bjhham: I'll actually roll back the client-side handling too.  It was mentioned in the ticket but it's not necessary for protection.

osipxd (approved):

PR #5663, opened 2026-05-26
KTOR-9492 Support require parameter functions in compiler plugin

bjhham:
**Subsystem**
Compiler plugin

**Motivation**
[KTOR-9492](https://youtrack.jetbrains.com/issue/KTOR-9492) OpenAPI plugin: handle requireXxx functions in compiler plugin

**Solution**
Introduced new cases to the compiler plugin's parameter inference.


osipxd (approved):
👍

zibet27 (approved):

PR #5664, opened 2026-05-27
KTOR-9620 Some issues related to digest authentication in Ktor 3.5.0

zibet27:
**Subsystem**
Server

**Motivation**
[KTOR-9620](https://youtrack.jetbrains.com/issue/KTOR-9620) Some issues related to digest authentication in Ktor 3.5.0

[KTOR-9624](https://youtrack.jetbrains.com/issue/KTOR-9624) Digest Auth: "Unsupported charset in digest authentication header" for a charset name in lowercase
The RFC7616 says only the "UTF-8" value is allowed, but parameters in HTTP are case-insensitive

[KTOR-9624](https://youtrack.jetbrains.com/issue/KTOR-9624) Digest Auth: "Unsupported charset in digest authentication header" for a charset name in lowercase

[KTOR-9630](https://youtrack.jetbrains.com/issue/KTOR-9630) Digest Auth: The plugin sends incorrect nonce when server responds with multiple WWW-Authenticate headers

Solution:

- The server now accepts charset=utf-8 in lowercase
- The server now emits a separate WWW-Authenticate header for each Digest algorithm instead of joining them into one ambiguous comma-separated value
- The client now rejects challenges whose algorithm parameter doesn't match its configured algorithmName, so it picks the nonce from the correct challenge rather than the first one.

### DIFF
--- a/ktor-client/ktor-client-plugins/ktor-client-auth/common/src/io/ktor/client/plugins/auth/providers/DigestAuthProvider.kt
+++ b/ktor-client/ktor-client-plugins/ktor-client-auth/common/src/io/ktor/client/plugins/auth/providers/DigestAuthProvider.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2025 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ * Copyright 2014-2026 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
  */
 
 package io.ktor.client.plugins.auth.providers
@@ -10,6 +10,7 @@ import io.ktor.client.statement.*
 import io.ktor.http.*
 import io.ktor.http.auth.*
 import io.ktor.util.*
+import io.ktor.util.logging.trace
 import io.ktor.utils.io.*
 import io.ktor.utils.io.charsets.*
 import io.ktor.utils.io.core.*
@@ -129,24 +130,24 @@ public class DigestAuthProvider(
 
     override fun isApplicable(auth: HttpAuthHeader): Boolean {
         if (auth !is HttpAuthHeader.Parameterized || auth.authScheme != AuthScheme.Digest) {
-            LOGGER.trace("Digest Auth Provider is not applicable for $auth")
+            LOGGER.trace { "Digest Auth Provider is not applicable for $auth" }
             return false
         }
 
         val newNonce = auth.parameter("nonce") ?: run {
-            LOGGER.trace("Digest Auth Provider can not handle response without nonce parameter")
+            LOGGER.trace { "Digest Auth Provider can not handle response without nonce parameter" }
             return false
         }
         val newQop = auth.parameter("qop")
         val newOpaque = auth.parameter("opaque")
 
         val newRealm = auth.parameter("realm") ?: run {
-            LOGGER.trace("Digest Auth Provider can not handle response without realm parameter")
+            LOGGER.trace { "Digest Auth Provider can not handle response without realm parameter" }
             return false
         }
         @Suppress("DEPRECATION_ERROR")
         if (newRealm != realm && realm != null) {
-            LOGGER.trace("Digest Auth Provider is not applicable for this realm")
+            LOGGER.trace { "Digest Auth Provider is not applicable for this realm" }
             return false
         }
 
@@ -154,7 +155,7 @@ public class DigestAuthProvider(
         val challengeAlgorithm = auth.parameter("algorithm") ?: "MD5"
         @Suppress("DEPRECATION_ERROR")
         if (!challengeAlgorithm.equals(algorithmName, ignoreCase = true)) {
-            LOGGER.trace("Digest Auth Provider is not applicable for algorithm $challengeAlgorithm")
+            LOGGER.trace { "Digest Auth Provider is not applicable for algorithm $challengeAlgorithm" }
             return false
         }
 
--- a/ktor-client/ktor-client-plugins/ktor-client-auth/common/src/io/ktor/client/plugins/auth/providers/DigestAuthProvider.kt
+++ b/ktor-client/ktor-client-plugins/ktor-client-auth/common/src/io/ktor/client/plugins/auth/providers/DigestAuthProvider.kt
@@ -150,6 +150,14 @@ public class DigestAuthProvider(
             return false
         }
 
+        // Per RFC 7616 §3.3: a missing algorithm parameter defaults to MD5.
+        val challengeAlgorithm = auth.parameter("algorithm") ?: "MD5"
+        @Suppress("DEPRECATION_ERROR")
+        if (!challengeAlgorithm.equals(algorithmName, ignoreCase = true)) {
+            LOGGER.trace("Digest Auth Provider is not applicable for algorithm $challengeAlgorithm")
+            return false
+        }
+
         serverNonce.value = newNonce
         qop.value = newQop
         opaque.value = newOpaque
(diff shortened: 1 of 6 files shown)


ktor-client/ktor-client-plugins/ktor-client-auth/common/test/io/ktor/client/plugins/auth/AuthTest.kt
```diff
@@ -323,20 +369,20 @@ class AuthTest : ClientLoader() {
...
-    fun testUsesFreshTokenIfAvailable() = testSuspend {
-        val request1FinishMonitor = Job()
-        val request2StartMonitor = Job()
-        var refreshCount = 0
-        val client = HttpClient(MockEngine) {
+    fun testUsesFreshTokenIfAvailable() = runTest {
+        val request1FinishMonitor1 = Job(parent = coroutineContext.job)
+        val request2StartMonitor1 = Job(parent = coroutineContext.job)
+        var refreshCount1 = 0
+        val client1 = HttpClient(MockEngine) {
```
osipxd: It seems like IDE has added suffix `1` to all variables after replacement `testSuspend` with `runTest`

ktor-client/ktor-client-plugins/ktor-client-auth/common/test/io/ktor/client/plugins/auth/AuthTest.kt
```diff
@@ -616,41 +658,41 @@ class AuthTest : ClientLoader() {
...
-                assertEquals("OK", second)
-            }
-            jobs += GlobalScope.launch {
-                val second = client.get("$TEST_SERVER/auth/bearer/second").bodyAsText()
-                assertEquals("OK", second)
-            }
-            jobs += GlobalScope.launch {
-                val second = client.get("$TEST_SERVER/auth/bearer/second").bodyAsText()
-                assertEquals("OK", second)
+            withContext(Dispatchers.Default) {
```
osipxd, line 682: If switching to Default dispatcher is not necessary, `coroutineScope { ... }` could be used here

ktor-client/ktor-client-plugins/ktor-client-auth/common/test/io/ktor/client/plugins/auth/AuthTest.kt
```diff
@@ -99,6 +99,51 @@ class AuthTest : ClientLoader() {
...
+                        } else {
+                            respond("OK", HttpStatusCode.OK)
+                        }
+                    }
+                }
+            }
+
+            test { client ->
+                val response = client.get("/")
+                val auth = response.call.request.headers[HttpHeaders.Authorization]!!
```
osipxd: Minor. `!!` operator is usually can be avoided
```suggestion
                val auth = response.call.requireHeader(HttpHeaders.Authorization)
```

ktor-client/ktor-client-plugins/ktor-client-auth/common/test/io/ktor/client/plugins/auth/AuthTest.kt
```diff
@@ -99,6 +99,51 @@ class AuthTest : ClientLoader() {
...
+                val nonce = Regex("""nonce="([^"]+)"""").find(auth)!!.groupValues[1]
+                assertEquals("md5-nonce", nonce)
```
osipxd: Let's simplify this check:
```kotlin
assertContains(nonce, "nonce=\"md5-nonce\"")
```

ktor-client/ktor-client-plugins/ktor-client-auth/common/test/io/ktor/client/plugins/auth/DigestProviderTest.kt
```diff
@@ -62,6 +62,49 @@ class DigestProviderTest {
         }
     }
 
+    @Test
+    fun `isApplicable rejects challenge whose algorithm does not match the configured one`() = runTest {
+        if (!PlatformUtils.IS_JVM) return@runTest
+
+        val sha512Challenge = parseAuthorizationHeader(
+            """Digest algorithm=SHA-512-256, realm="realm", nonce="sha-nonce""""
+        )!!
```
osipxd: Let's use `assertNotNull(sha512Challenge)` to get a proper assertion failure instead of NPE
```suggestion
        val sha512Challenge = parseAuthorizationHeader(
            """Digest algorithm=SHA-512-256, realm="realm", nonce="sha-nonce""""
        )
        assertNotNull(sha512Challenge)
```

ktor-server/ktor-server-plugins/ktor-server-auth/jvm/test/io/ktor/tests/auth/DigestRFC7616Test.kt
```diff
@@ -426,6 +427,37 @@ class DigestRFC7616Test {
         assertTrue(authInfo.contains("rspauth=\"$expectedRspauth\""))
     }
 
+    @Test
+    fun `KTOR-9623 server sends one WWW-Authenticate header per algorithm`() = testApplication {
```
osipxd: I think it's better to remove issue IDs from test names

ktor-client/ktor-client-plugins/ktor-client-auth/common/src/io/ktor/client/plugins/auth/providers/DigestAuthProvider.kt
```diff
@@ -150,6 +150,14 @@ public class DigestAuthProvider(
             return false
         }
 
+        // Per RFC 7616 §3.3: a missing algorithm parameter defaults to MD5.
+        val challengeAlgorithm = auth.parameter("algorithm") ?: "MD5"
+        @Suppress("DEPRECATION_ERROR")
+        if (!challengeAlgorithm.equals(algorithmName, ignoreCase = true)) {
+            LOGGER.trace("Digest Auth Provider is not applicable for algorithm $challengeAlgorithm")
```
osipxd: Minor optimization. `trace { "..." }` could be used here

osipxd (changes requested):
Fixes look good, I have comments regarding tests

Stexxe (approved):

ktor-server/ktor-server-plugins/ktor-server-auth/jvm/test/io/ktor/tests/auth/DigestRFC7616Test.kt
zibet27: Agree
I thought I removed all of these names

ktor-client/ktor-client-plugins/ktor-client-auth/common/test/io/ktor/client/plugins/auth/AuthTest.kt
zibet27: This client call
It looks like we didn't add requireHeader for the client
Or I have an old branch version

ktor-client/ktor-client-plugins/ktor-client-auth/common/test/io/ktor/client/plugins/auth/AuthTest.kt
zibet27: I think the reason for using `GlobalScope` before was to run these jobs in parallel
Afaik, the test dispatcher is single-threaded:

```kotlin
test { client ->
            refreshRequestsCount.update { 0 }
            client.get("$TEST_SERVER/auth/bearer/first").bodyAsText()

            val dispatcher = coroutineContext[ContinuationInterceptor] as? CoroutineDispatcher
            println(dispatcher)

            withContext(Dispatchers.Default) {
```

prints `Dispatchers.Default.limitedParallelism(1)`

bjhham (approved):

ktor-client/ktor-client-plugins/ktor-client-auth/common/test/io/ktor/client/plugins/auth/AuthTest.kt
osipxd: Yes, my mistake. We don't have these shortcuts for `HttpResponse` or `HttpClientCall`

osipxd (approved):

PR #5680, opened 2026-06-04
KTOR-8695 Fix windows sending us 300 years in the future

bjhham:
**Subsystem**
Core, Native

**Motivation**
[KTOR-8695](https://youtrack.jetbrains.com/issue/KTOR-8695) GMTDate() being offset by 369 years on Windows

**Solution**
`GetSystemTimeAsFileTime` returns 100-ns intervals since 1601-01-01 UTC (FILETIME epoch), but the API contract requires milliseconds since the Unix epoch (1970-01-01 UTC). The previous code divided by 10000 without subtracting the epoch offset.

### DIFF
--- a/ktor-utils/mingwX64/src/io/ktor/util/date/DateMingw.kt
+++ b/ktor-utils/mingwX64/src/io/ktor/util/date/DateMingw.kt
@@ -14,6 +14,10 @@ internal actual fun system_time(tm: CValuesRef<tm>?): Long {
     return _mkgmtime(tm).convert()
 }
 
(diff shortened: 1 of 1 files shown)


zibet27 (approved):
Back to the present :)

Btw, I think the comment from the issue regarding deprecating this in favour of `kotlin.time`(when it becomes non-experimental) looks to be the right direction for me

https://youtrack.jetbrains.com/issue/KTOR-8695/GMTDate-being-offset-by-369-years-on-Windows#focus=Comments-27-13859670.0-0

PR #5683, opened 2026-06-04
KTOR-9646 Update compiler plugin to 2.4.0

bjhham:
Looks like there were no breaking changes in the compiler plugin itself this time, but instead some binary incompatibility within the compiler instrumentation itself.

I noticed the test snapshots were out of date and that CI was not running the compiler plugin tests automatically, so I just added a task that will trigger during the normal suite.

Note, I'm only updating to 2.4.0 for the compiler plugin, so we can be more deliberate about updating it project-wide.  I'll also test this against 2.3.21 to see if it's backwards compatible.

### DIFF
--- a/ktor-compiler-plugin/testData/openapi/MarkdownOptions.expected.json
+++ b/ktor-compiler-plugin/testData/openapi/MarkdownOptions.expected.json
@@ -196,7 +196,10 @@
     "components": {
         "schemas": {
             "Contact": {
-                "type": "object",
+                "type": [
+                    "object",
+                    "null"
+                ],
                 "title": "Contact",
                 "properties": {
                     "name": {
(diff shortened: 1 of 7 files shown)


ktor-compiler-plugin/testData/openapi/MarkdownOptions.expected.json
```diff
@@ -196,7 +196,10 @@
     "components": {
         "schemas": {
             "Contact": {
-                "type": "object",
+                "type": [
```
zibet27, line 199: Is it a breaking change from our side?

bjhham: These were actually changed in another fix but the compiler plugin tests were not being executed as part of the full suite before.  I'm just updating the snapshots and including the tests in our test suite here - it's unrelated to the Kotlin version.

zibet27 (approved):

PR #5684, opened 2026-06-05
Improve WebRTC data channel iteration, exceptions

zibet27:
**Subsystem**
WebRTC Client

**Motivation**
[KTOR-9610](https://youtrack.jetbrains.com/issue/KTOR-9610) WebRtcPeerConnection::close causes ClosedReceiveChannelException for other peer

**Solution**
- When the channel is closed, `ReceiveChannel.receive()` throws a general closed channel exception with a stacktrace pointing at `close()`, which is confusing for users; throw WebRtc.ClosedException instead
- add an iterator over the data channel, so users don't need to catch such exceptions and iterate safely
- throw common `WebRtc.IOException` in `WebRtc.Datachannel.send()` so users can catch it in common code


bjhham (approved):
lgtm 👍

PR #5685, opened 2026-06-06
KTOR-8524 Trigger application lifecycle events for ServletApplicationEngine in WAR deployments

Seongju-Lee:
**Subsystem**
 Server: ktor-server-servlet, ktor-server-servlet-jakarta
(regression tests in ktor-server-jetty, ktor-server-jetty-jakarta)

**Motivation**
[KTOR-8524](https://youtrack.jetbrains.com/issue/KTOR-8524/ServletApplicationEngine-Application-events-arent-triggered)  

`ApplicationStarted` / `ApplicationStopped` are meant to fire when the application starts and stops, consistently across deployment modes. In the embedded modes (`embeddedServer(...)`) this already works: the events fire when Ktor calls `start()` / `stop()`.

In a WAR deployment, however, `ServletApplicationEngine` self-bootstraps an `EmbeddedServer` and ties its lifecycle to the servlet's `init()` / `destroy()` callbacks. Because a servlet is initialized lazily (without `load-on-startup`, on the first request), the application lifecycle ends up bound to the wrong moment:

- `ApplicationStarted` fires only after the first request, not at deployment time.
- `ApplicationStopped` is never fired if the application is undeployed without serving any request, so resources released in `ApplicationStopped` handlers leak.

In a WAR, "the application starts" is the web application context being initialized (deploy time) — not the first request. `load-on-startup` would work around this, but it is a deployment-descriptor setting owned by the user, so the library cannot guarantee it; the lifecycle has to be correct by default.

**Solution**
Bind the application lifecycle to the servletContext instead of the servlet's lazy `init()` / `destroy()`:
- `KtorServletContextListener` starts/stops the embedded server on `contextInitialized` / `contextDestroyed`, so `ApplicationStarted` /`ApplicationStopped` fire at deploy/undeploy regardless of `load-on-startup`.
- `KtorServletContainerInitializer` (auto-registered via `META-INF/services`) registers the listener, so no user configuration is required. Marked `@InternalAPI`.
- `ServletApplicationEngine` reuses the listener-managed server and skips its own start/stop. The servlet stays lazily initialized, falls back to self-bootstrapping when no initializer ran, and the whole thing is a no-op when an embedded engine owns the lifecycle.

Applied to both `ktor-server-servlet` (javax) and `ktor-server-servlet-jakarta` with regression tests in the servlet and Jetty modules.

### DIFF
--- a/ktor-server/ktor-server-servlet-jakarta/jvm/src/io/ktor/server/servlet/jakarta/KtorServletContainerInitializer.kt
+++ b/ktor-server/ktor-server-servlet-jakarta/jvm/src/io/ktor/server/servlet/jakarta/KtorServletContainerInitializer.kt
@@ -1,0 +1,96 @@
+/*
+ * Copyright 2014-2026 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.server.servlet.jakarta
+
+import io.ktor.server.application.*
+import io.ktor.server.servlet.jakarta.ServletApplicationEngine.Companion.ApplicationAttributeKey
+import io.ktor.server.servlet.jakarta.ServletApplicationEngine.Companion.ApplicationEnginePipelineAttributeKey
+import io.ktor.utils.io.InternalAPI
+import jakarta.servlet.ServletContainerInitializer
+import jakarta.servlet.ServletContext
+import jakarta.servlet.ServletContextEvent
+import jakarta.servlet.ServletContextListener
+
+/**
+ * Registers [KtorServletContextListener] so a servlet-hosted Ktor application is started and stopped in
+ * lockstep with the web application context lifecycle.
+ *
+ * Registered automatically via `META-INF/services/jakarta.servlet.ServletContainerInitializer`, so no
+ * user configuration is required. It is a no-op when the application is driven by an embedded engine,
+ * which injects [ServletApplicationEngine.ApplicationAttributeKey] and owns the lifecycle itself.
+ *
+ * It must be `public` so the servlet container can instantiate it via `ServiceLoader`, but it is
+ * internal infrastructure and not intended to be referenced directly.
+ *
+ * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.server.servlet.jakarta.KtorServletContainerInitializer)
+ */
+@InternalAPI
+public class KtorServletContainerInitializer : ServletContainerInitializer {
+    override fun onStartup(classes: MutableSet<Class<*>>?, ctx: ServletContext) {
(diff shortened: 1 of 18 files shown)


ktor-server/ktor-server-servlet-jakarta/jvm/src/io/ktor/server/servlet/jakarta/KtorServletContainerInitializer.kt
```diff
@@ -0,0 +1,96 @@
...
+    override fun onStartup(classes: MutableSet<Class<*>>?, ctx: ServletContext) {
+        // Embedded engine mode owns the lifecycle; do not register the listener.
+        if (ctx.getAttribute(ApplicationAttributeKey) != null) return
+        ctx.addListener(KtorServletContextListener::class.java)
+    }
+}
+
+/**
+ * Starts and stops a servlet-hosted Ktor application together with the web application context, so the
+ * lifecycle events fire at deploy/undeploy regardless of `load-on-startup` (see KTOR-8524).
```
bjhham: Please omit the YouTrack ticket numbers in production code comments in favour of more detailed descriptions.

ktor-server/ktor-server-servlet/jvm/src/io/ktor/server/servlet/KtorServletContainerInitializer.kt
```diff
@@ -0,0 +1,96 @@
...
+    override fun onStartup(classes: MutableSet<Class<*>>?, ctx: ServletContext) {
+        // Embedded engine mode owns the lifecycle; do not register the listener.
+        if (ctx.getAttribute(ApplicationAttributeKey) != null) return
+        ctx.addListener(KtorServletContextListener::class.java)
+    }
+}
+
+/**
+ * Starts and stops a servlet-hosted Ktor application together with the web application context, so the
+ * lifecycle events fire at deploy/undeploy regardless of `load-on-startup` (see KTOR-8524).
```
bjhham: Here too 👀

bjhham (commented):
Thanks for the contribution!  Really nice work.  I just had a small comment about the comments 😄

ktor-server/ktor-server-servlet-jakarta/jvm/src/io/ktor/server/servlet/jakarta/KtorServletContainerInitializer.kt
Seongju-Lee: Done! swapped the ticket number for a description of the behavior. Thanks!

bjhham (approved):

PR #5686, opened 2026-06-06
Support alternative `fetch` implementations in JS client

cedrickcooke:
**Subsystem**
Ktor client `JS` engine

**Motivation**
Some JavaScript libraries, such as [AWS WAF](https://docs.aws.amazon.com/waf/latest/developerguide/waf-javascript-api.html), require using an alternative method (eg, `AwsWafIntegration.fetch`) instead of the global `fetch`. While it's possible to work around this for some libraries by monkey-patching `fetch` to include that call, trying to do this with `AwsWafIntegration` causes an infinite loop because their method calls the global `fetch` internally.

**Solution**
Adds a `fetch` variable to `JsClientEngineConfig`. By default nothing changes and the global `fetch` is used, but you can opt-in to having the client call an alternative wrapper.

### DIFF
--- a/ktor-client/ktor-client-core/api/ktor-client-core.klib.api
+++ b/ktor-client/ktor-client-core/api/ktor-client-core.klib.api
@@ -1948,6 +1948,9 @@ final class io.ktor.client.engine.js/JsError : kotlin/Throwable { // io.ktor.cli
 open class io.ktor.client.engine.js/JsClientEngineConfig : io.ktor.client.engine/HttpClientEngineConfig { // io.ktor.client.engine.js/JsClientEngineConfig|null[0]
     constructor <init>() // io.ktor.client.engine.js/JsClientEngineConfig.<init>|<init>(){}[0]
 
+    final var fetch // io.ktor.client.engine.js/JsClientEngineConfig.fetch|{}fetch[0]
+        final fun <get-fetch>(): kotlin/Function2<kotlin/String, io.ktor.client.fetch/RequestInit?, kotlin.js/Promise<org.w3c.fetch/Response>> // io.ktor.client.engine.js/JsClientEngineConfig.fetch.<get-fetch>|<get-fetch>(){}[0]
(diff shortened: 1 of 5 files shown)


bjhham (commented):
Looks good, thanks for the contribution!

Could you just provide a little more details on the AWS WAF library and an example of how this feature ought to be used?  These details will be necessary for the release notes and documentation.

cedrickcooke:
> Looks good, thanks for the contribution!
>
> Could you just provide a little more details on the AWS WAF library and an example of how this feature ought to be used? These details will be necessary for the release notes and documentation.

Sure! At a pretty high level, AWS WAF is Amazon's firewall service. One of its features ("intelligent threat mitigation") acts as a middleman between the client and server to help distinguish between the app that's intended to consume an API and bots/bad-actors. While the middleman is _usually_ transparent, sometimes it issues a "challenge" where the client needs to run some code to prove that it's not a bot and then re-issue the request with the result of that challenge. It's possible to implement this manually, but the recommended setup is to use their integration library.

Integrating this is pretty simple in raw JavaScript. AWS WAF provides a `challenge.js` script that you embed directly in your HTML. Then, you replace all your calls to `fetch` with `AwsWafIntegration.fetch`. When you call that function, if the server returns a challenge script, it will silently run it and retry the request, completing the `Promise` with the result of the second request instead. Copying directly from their docs, the HTML side of this looks like:

```html
<head>
<script type="text/javascript" src="protection pack (web ACL) integration URL/challenge.js" defer></script>
</head>
```

This HTML integration will still need to be included in the page that runs your Kotlin/JS. As far as the Kotlin integration, you'll need to define some `external`s and configure the client. That'll end up looking something like this:

```kotlin
public external object AwsWafIntegration {
    public fun fetch(input: String, init: RequestInit? = definedExternally): Promise<org.w3c.fetch.Response>
}

public fun sample() {
    val client = HttpClient(Js) {
        engine {
            fetch = AwsWafIntegration::fetch
        }
    }
    // Use the client
}
```

I did try and keep the library code here flexible/independent of AWS WAF in case anything else uses this same pattern, 🤷

bjhham (approved):
Added a ticket in YT for reference:
[KTOR-9654](https://youtrack.jetbrains.com/issue/KTOR-9654) Override fetch in JS engine

Should be out with 3.6.0 in August

PR #5689, opened 2026-06-08
Fix KTOR-9621 RateLimit plugin is bypassed

zibet27:
**Subsystem**
Server

**Motivation**
[KTOR-9621](https://youtrack.jetbrains.com/issue/KTOR-9621) RateLimit plugin is bypassed when the nested authenticate block rejects the request

**Solution**
- Auth ran in the Authenticate phase (after Plugins) while rate limit ran in a BeforeCall phase (before Call). 
- Failed auth responded with 401 and set call.isHandled, so the rate limiter’s guard exited early
- Move the rate-limit hook to the Plugins phase, so it always runs before authentication, regardless of DSL nesting (`rateLimit { authenticate { … } }` or the reverse)


bjhham (approved):
👍

Stexxe (approved):

PR #5692, opened 2026-06-11
OIDC Plugin. Support PKCE

zibet27:
**Subsystem**
Server OIDC

**Motivation**
[KTOR-5001](https://youtrack.jetbrains.com/issue/KTOR-5001) Add OpenID Connect (OIDC) support on client and server

**Solution**
- Adds PKCE (RFC 7636) to the OIDC authorisation code flow
- Sends code_challenge / code_challenge_method=S256 on authorize and code_verifier on token exchange
- Stores the code verifier in the encrypted state cookie alongside the nonce
- Adds disablePkce() for legacy OpenID Providers that reject PKCE parameters

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-auth-oidc/jvm/src/io/ktor/server/auth/oidc/OidcConfig.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-auth-oidc/jvm/src/io/ktor/server/auth/oidc/OidcConfig.kt
@@ -473,14 +473,16 @@ public class OidcOAuthConfig<P : Any> internal constructor(
     public var fetchUserInfo: Boolean = false
 
     /**
-     * Symmetric key used to encrypt the in-flight OAuth state cookie carrying `state` and `nonce`
-     * between the login redirect and the callback.
+     * Symmetric key used to encrypt the in-flight OAuth state cookie carrying `state`, `nonce`, and the PKCE code
+     * verifier between the login redirect and the callback.
      *
      * Required in production. In development mode an ephemeral key is generated when not set.
      *
      * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.server.auth.oidc.OidcOAuthConfig.stateEncryptionKey)
      */
     public var stateEncryptionKey: OidcStateEncryptionKey? = null
+
+    internal var pkceEnabled: Boolean = true
 
     /**
      * Configures the OAuth callback route URI.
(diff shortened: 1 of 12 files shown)


ktor-server/ktor-server-plugins/ktor-server-auth-oidc/jvm/src/io/ktor/server/auth/oidc/OidcConfig.kt
```diff
@@ -473,15 +473,17 @@ public class OidcOAuthConfig<P : Any> internal constructor(
...
+     * Symmetric key used to encrypt the in-flight OAuth state cookie carrying `state`, `nonce`, and the PKCE code
+     * verifier between the login redirect and the callback.
      *
      * Required in production. In development mode an ephemeral key is generated when not set.
      *
      * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.server.auth.oidc.OidcOAuthConfig.stateEncryptionKey)
      */
     public var stateEncryptionKey: OidcStateEncryptionKey? = null
 
+    internal var pkceEnabled: Boolean = true
```
bjhham: Looks like it might make more sense to introduce a sealed type instead of a flag?

```kotlin
sealed class CodeChallengeMethod {
    abstract val method: String

    object Pkce: CodeChallengeMethod {
        val verifierLength: Int = 64
        override val method: String = "S256"
    }
}
```

And `public var codeChallengeMethod: CodeChallengeMethod? = CodeChallengeMethod.Pkce`?

Not sure if we can expect new methods to be introduced, but it seems like it was designed that way.

bjhham (commented):
Looks good, just the one comment wrt the API.

ktor-server/ktor-server-plugins/ktor-server-auth-oidc/jvm/src/io/ktor/server/auth/oidc/OidcConfig.kt
zibet27: Then null would mean disabled PKCE?

Actually, there is a "plain" case, but it's highly discouraged  https://datatracker.ietf.org/doc/html/rfc7636#section-4.2

bjhham: Yeah, that's what I was thinking.

Either:
```
codeChallengeMethod = null
```

And perhaps a more generic `disableCodeChallenge()` over `disablePkse()`

We can omit `Plain` for now then potentially add it later easily.  There could be some demand for testing or something.

zibet27: Should it be `sealed`?
It's very unlikely that anyone will use `when` on it, but it still will be breaking change if we add one more case

bjhham: Though if it's not sealed then adding to the API could be breaking.  I think we're fine either way though.

bjhham (approved):

PR #5695, opened 2026-06-11
OAuth 2.0 Protected Resource Metadata to the OIDC plugin

zibet27:
**Subsystem**
Server OIDC

**Motivation**
[KTOR-5001](https://youtrack.jetbrains.com/issue/KTOR-5001) Add OpenID Connect (OIDC) support on client and server

**Solution**
- New `protectedResource(resource) { }` DSL
- Serves metadata at `/.well-known/oauth-protected-resource{path}` derived from the resource identifier URL
- Auto-derives authorization_servers, scopes_supported, and bearer_methods_supported from configured providers unless explicitly overridden
- Bearer authentication failures include a resource_metadata parameter in WWW-Authenticate when protected resource metadata is enabled

This should be the last PR for the OIDC plugin

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-auth-oidc/jvm/src/io/ktor/server/auth/oidc/OidcProtectedResource.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-auth-oidc/jvm/src/io/ktor/server/auth/oidc/OidcProtectedResource.kt
@@ -1,0 +1,92 @@
+/*
+ * Copyright 2014-2026 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.server.auth.oidc
+
+import io.ktor.serialization.kotlinx.json.*
+import io.ktor.server.application.*
+import io.ktor.server.plugins.contentnegotiation.*
+import io.ktor.server.response.*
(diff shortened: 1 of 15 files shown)


ktor-server/ktor-server-plugins/ktor-server-auth-oidc/jvm/src/io/ktor/server/auth/oidc/OidcProtectedResource.kt
```diff
@@ -0,0 +1,92 @@
...
+        val format = Json {
+            explicitNulls = false
+            encodeDefaults = true
+        }
+        json(format)
+    }
+
+    val metadata by lazy { buildProtectedResourceMetadata(config, providers()) }
+    val path = buildResourceMetadataRoutePath(config.resource)
+    get(path) { call.respond(metadata) }
```
bjhham: Might want to double check this, but it looks like you're installing `ContentNegotiation` on the root here.

I believe it should be:

```kotlin
route(path) {
    install(ContentNegotiation) { /* ... */ }

    get { call.respond(metadata) }
}
```

bjhham (commented):
Looks good - just the one item regarding the use of `ContentNegotiation`.

zibet27:
@coderabbitai review

PR #5698, opened 2026-06-12
KTOR-5009 Add a way to prevent automatic Accept header injection

MARKOCEVRLJAKOVIC:
### Target Ticket
Refers to YouTrack issue: [KTOR-5009](https://youtrack.jetbrains.com/issue/KTOR-5009) (ContentNegotiation: Add a way to prevent changing Accept and Content-Type headers)

### Description
This PR introduces a way to opt-out of automatic `Accept` header injection performed by the `ContentNegotiation` client plugin. 

Currently, the plugin appends registered content types to the request's `Accept` header even if the user has explicitly specified a custom `Accept` header. 

To resolve this, a new configuration property `addAcceptHeaderOnlyIfNonePresent` has been added to `ContentNegotiationConfig`. 
- When set to `true`, the plugin will completely bypass appending fallback headers if any `Accept` header is already configured on the outgoing request.
- The default value is set to `false` to maintain backward compatibility and preserve original behavior.

### Changes
- **`ContentNegotiation.kt`**: Added the `addAcceptHeaderOnlyIfNonePresent` property and wrapped the request registration block in a guard condition inside `convertRequest()`.
- **API Dumps**: Updated JVM and KLIB api files to reflect binary compatibility for the new public API property.
- **`ContentNegotiationTests.kt`**: Added two unit tests covering cases where the option is enabled, one when a user header is present (skips injection) and one when it is missing (performs default injection).

### DIFF
--- a/ktor-client/ktor-client-plugins/ktor-client-content-negotiation/common/src/io/ktor/client/plugins/contentnegotiation/ContentNegotiation.kt
+++ b/ktor-client/ktor-client-plugins/ktor-client-content-negotiation/common/src/io/ktor/client/plugins/contentnegotiation/ContentNegotiation.kt
@@ -65,7 +65,13 @@ public fun interface ContentTypeMergeStrategy {
          */
         public val Default: ContentTypeMergeStrategy = ContentTypeMergeStrategy { registered, headers ->
             registered.asSequence().filter { contentType ->
-                headers.none { h -> ContentType.parse(h).match(contentType) }
+                headers.none { h ->
+                    try {
+                        ContentType.parse(h).match(contentType)
+                    } catch (e: BadContentTypeFormatException) {
+                        false
+                    }
+                }
             }
         }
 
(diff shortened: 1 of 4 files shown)


ktor-client/ktor-client-plugins/ktor-client-content-negotiation/common/src/io/ktor/client/plugins/contentnegotiation/ContentNegotiation.kt
```diff
@@ -66,6 +66,16 @@ public class ContentNegotiationConfig : Configuration {
...
 
+    /**
+     * If set to true, the plugin will not add Accept headers for registered content types
+     * if the request already has any Accept header set.
+     * Useful when working with APIs that are strict about Accept header values.
+     * Default is false to preserve backward compatibility.
+     *
+     * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.client.plugins.contentnegotiation.ContentNegotiationConfig.addAcceptHeaderOnlyIfNonePresent)
+     */
+    public var addAcceptHeaderOnlyIfNonePresent: Boolean = false
```
bjhham: Maybe we could generalize this a little more with something like:

```kotlin
public var acceptHeaderMergeStrategy: ContentTypeMergeStrategy = LargestQParameter // User may change to AlwaysKeepExisting or something

// ...

fun interface ContentTypeMergeStategy {
  fun mergeContentType(
          requestRegistrations: List<ContentNegotiationConfig.ConverterRegistration>,
          providedHeaders: List<String>
      ): Sequence<ContentType>
}
```

This would allow for any custom logic and avoid too many granular flags in the API.

MARKOCEVRLJAKOVIC:
Thanks for the suggestion! I've updated the PR to replace the boolean flag with a ContentTypeMergeStrategy, which provides a more flexible and extensible API.

bjhham (approved):
Looks good 👍

PR #5701, opened 2026-06-15
KTOR-9661 HttpCache: InvalidCacheStateException is thrown on 304 when…

zibet27:
**Subsystem**
Client HttpCache

**Motivation**
[KTOR-9661](https://youtrack.jetbrains.com/issue/KTOR-9661) HttpCache: InvalidCacheStateException is thrown on 304 when no cached entry matches varyKeys

**Solution**
- Added a validator-based fallback when matching a 304 to a stored entry https://www.rfc-editor.org/info/rfc9111/#section-4.3.4
- Freshen stored entries by merging the 304 response headers and updating `varyKeys/expiry`; remove the stale entry when `varyKeys` change

### DIFF
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/cache/HttpCache.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/cache/HttpCache.kt
@@ -1,6 +1,6 @@
 /*
-* Copyright 2014-2021 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
-*/
+ * Copyright 2014-2026 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
 
 @file:Suppress("DEPRECATION")
 
@@ -43,14 +43,8 @@ internal val LOGGER = KtorSimpleLogger("io.ktor.client.plugins.HttpCache")
  * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.client.plugins.cache.HttpCache)
  */
 public class HttpCache private constructor(
-    @Deprecated(
-        "This will become internal",
-        level = DeprecationLevel.ERROR
-    ) @Suppress("DEPRECATION_ERROR") internal val publicStorage: HttpCacheStorage,
-    @Deprecated(
-        "This will become internal",
-        level = DeprecationLevel.ERROR
-    ) @Suppress("DEPRECATION_ERROR") internal val privateStorage: HttpCacheStorage,
+    @Suppress("DEPRECATION_ERROR") internal val publicStorage: HttpCacheStorage,
+    @Suppress("DEPRECATION_ERROR") internal val privateStorage: HttpCacheStorage,
     private val publicStorageNew: CacheStorage,
     private val privateStorageNew: CacheStorage,
     private val useOldStorage: Boolean,
@@ -236,21 +230,11 @@ public class HttpCache private constructor(
 
                 if (response.status == HttpStatusCode.NotModified) {
                     LOGGER.trace { "Not modified response for ${response.call.request.url}, replying from cache" }
-                    val responseFromCache =
-                        plugin.findAndRefresh(response.call.request, response) ?: throw InvalidCacheStateException(
-                            response.call.request.url
-                        )
-                    if (responseFromCache.varyKeys().size != response.varyKeys().size) {
-                        LOGGER.warn(
-                            "Vary header mismatch on cached response for ${response.call.request.url}. " +
-                                "Received 304 Not Modified with Vary: ${response.varyKeys()} " +
-                                "but cached response has Vary: ${responseFromCache.varyKeys()}. " +
-                                "According to RFC 7232 §4.1 and RFC 9111 §4.1, " +
-                                "the server must include the full Vary header in 304 responses. " +
-                                "Falling back to missing cache logic. " +
-                                "Consider reporting this issue to the server maintainers."
-                        )
-                    }
+                    val responseFromCache = refreshNotModifiedResponse(
+                        response.call.request,
+                        response,
+                        plugin::findAndRefresh,
+                    )
 
                     scope.monitor.raise(HttpResponseFromCache, responseFromCache)
                     proceedWith(responseFromCache)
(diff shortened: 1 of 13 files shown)


ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/cache/storage/UnlimitedCacheStorage.kt
```diff
@@ -28,6 +28,20 @@ internal class UnlimitedCacheStorage : HttpCacheStorage() {
...
+
+    internal fun remove(url: Url, varyKeys: Map<String, String>) =
+        store[url]?.remove(varyKeys) { it.varyKeys }
+}
+
+private inline fun <E> MutableSet<E>.remove(
+    varyKeys: Map<String, String>,
+    crossinline entryVaryKeys: (E) -> Map<String, String>
+) {
+    val entriesToRemove = filter { entry ->
```
osipxd: Probably `removeAll` is more efficient here as it filters the map in place without creating a filtered copy.
Or was it intentionally changed to filter + forEach?

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/cache/storage/HttpCacheStorage.kt
```diff
@@ -237,3 +241,22 @@ public class CachedResponseData(
         body = body
     )
 }
+
+private val mergeCacheExcludedHeaders = setOf(
+    HttpHeaders.ContentLength,
+    HttpHeaders.Connection,
```
osipxd, line 247: The spec says:
> The Connection header field **and fields whose names are listed in it**

Should we also filter out these fields?

And [Section 7.6.1](https://www.rfc-editor.org/info/rfc9110/#section-7.6.1) lists some more fields to be excluded: Proxy-Connection, Keep-Alive, TE, Transfer-Encoding, Upgrade. The list is not exhaustive though.

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/cache/storage/HttpCacheStorage.kt
```diff
@@ -237,3 +241,22 @@ public class CachedResponseData(
...
+    HttpHeaders.ProxyAuthenticate,
+    HttpHeaders.ProxyAuthenticationInfo,
+    HttpHeaders.ProxyAuthorization,
```
osipxd, line 259: From the spec:
> MUST NOT be stored, **unless the cache incorporates the identity of the proxy into the cache key**

Just wanted to ensure we don't incorporate the proxy into the cache key.

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/cache/storage/HttpCacheStorage.kt
```diff
@@ -237,3 +241,22 @@ public class CachedResponseData(
         body = body
     )
 }
+
+private val mergeCacheExcludedHeaders = setOf(
```
osipxd: It seems this filter should be applied not only on refreshing, but also before storage (except for Content-Length) 🤔 Am I right?

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/cache/storage/HttpCacheStorage.kt
```diff
@@ -237,3 +241,22 @@ public class CachedResponseData(
...
+
+internal fun Headers.merge(other: Headers): Headers = Headers.build {
+    appendAll(this@merge)
+    for (name in other.names()) {
+        if (mergeCacheExcludedHeaders.any { it.equals(name, ignoreCase = true) }) {
+            continue
+        }
+        remove(name)
+        other.getAll(name)?.let { values -> appendAll(name, values) }
+    }
```
osipxd: Let's use `other.entries()` here. This way we could avoid an extra `getAll`:


```suggestion
    for ((name, values) in other.entries()) {
        if (mergeCacheExcludedHeaders.any { it.equals(name, ignoreCase = true) }) {
            continue
        }
        remove(name)
        appendAll(name, values)
    }
```

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/cache/HttpCache.kt
```diff
@@ -340,27 +324,34 @@ public class HttpCache private constructor(
...
+        response: HttpResponse,
         url: Url,
         request: HttpRequest
-    ): CachedResponseData? = when {
-        varyKeys.isNotEmpty() -> {
-            storage.find(url, varyKeys)
-        }
+    ): CachedResponseData? {
+        val varyKeys = response.varyKeys().takeIf { it.isNotEmpty() }
+        varyKeys?.let { find(url, varyKeys) }?.let { return it }
```
osipxd, line 346: Probably without `.takeIf { it.isNotEmpty() }` this would be a bit simpler:

```kotlin
val varyKeys = response.varyKeys()
find(url, varyKeys)?.let { return it }
```

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/plugins/CacheTest.kt
```diff
@@ -856,6 +857,173 @@ class CacheTest : ClientLoader() {
...
+        val etag = "\"v1\""
+        val url = Url("https://example.com/x")
+        val publicStorage = CacheStorage.Unlimited()
+        config {
+            install(HttpCache) {
+                publicStorage(publicStorage)
+            }
+            engine {
+                addHandler { request ->
+                    if (request.headers.contains(HttpHeaders.IfNoneMatch)) {
```
osipxd: Probably it would be better to invert a condition here and in other places: `HttpHeaders.IfNoneMatch !in request.headers`. This way initial response with status OK would go first and 304 after it.

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/plugins/CacheTest.kt
```diff
@@ -856,6 +857,173 @@ class CacheTest : ClientLoader() {
...
+            }
+        }
+
+        test { client ->
+            val first = client.get(url).bodyAsText()
+            assertEquals("body", first)
+
+            // The 304 has no validator, but the single stored response is freshened.
+            val second = client.get(url).bodyAsText()
+            assertEquals("body", second)
```
osipxd: Should we check here that this response was actually freshened?

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/cache/storage/UnlimitedCacheStorage.kt
zibet27: Yes
It wasn't working on Kotlin Native with `removeAll`
But I see that it's our impl, and it has bugs

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/cache/HttpCache.kt
zibet27: @osipxd What should find return on empty keys (ideomatically)? 
Atm, it returns the first entry with the specified URL because of `.all {}` used under the hood

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/cache/storage/UnlimitedCacheStorage.kt
osipxd: Are you going to fix the implementation here or leave it for another PR?

zibet27: Afaiu, `ConcurrentSet` is only used by the HTTP cache plugin, so I'll include it here

zibet27:
PS: You can review the last changes commit by commit; most changes are just tests and deduplication. I can restore the tests, though

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/cache/storage/HttpCacheStorage.kt
```diff
@@ -242,21 +242,60 @@ public class CachedResponseData(
...
+
+private val proxyAuthHeaders = listOf(
     HttpHeaders.ProxyAuthenticate,
     HttpHeaders.ProxyAuthenticationInfo,
     HttpHeaders.ProxyAuthorization,
 )
 
+private val staticStorageExcludedHeaders = hopByHopHeaders + proxyAuthHeaders
+
+private val connectionOptionsWithoutFieldNames = listOf("close", "keep-alive", "upgrade")
```
osipxd, line 264: Minor (since this list is tiny) but we could use `setOf` here for `O(1)` lookup

ktor-client/ktor-client-core/common/test/CacheHeaderFilteringTest.kt
```diff
@@ -0,0 +1,82 @@
...
+import io.ktor.client.plugins.cache.storage.*
+import io.ktor.http.*
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertNull
+
+class CacheHeaderFilteringTest {
+
+    @Test
+    fun filterForCacheStorageStripsHopByHopAndProxyAuthButKeepsContentLength() {
```
osipxd, line 14: Minor. As it is a new file it'd be better to use test names in backticks

osipxd (approved):
Nicely done 👍 
I have only two minor comments

osipxd:
Interesting. ConcurrentSet implementation throws `ConcurrentModificationException: null` on JS

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/cache/storage/HttpCacheStorage.kt
zibet27: Lookups on small lists are faster than on sets

zibet27:
> Interesting. ConcurrentSet implementation throws `ConcurrentModificationException: null` on JS

We would need to align JS ConcurrentMap with Native one to return a new copy on every `.keys` access

PR #5702, opened 2026-06-16
KTOR-9663 Upgrade to Kotlin 2.4

osipxd:
**Subsystem**
Infrastructure, Core

**Motivation**
[KTOR-9663](https://youtrack.jetbrains.com/issue/KTOR-9663) Upgrade to Kotlin 2.4

**Solution**
Update Kotlin, bump API level and Language level


osipxd:
"Check ABI" fails because it can't correctly generate ABI dumps on a Linux machine. We should switch to macOS agents, probably

PR #5703, opened 2026-06-17
KTOR-9632 Fix merge engine capabilities in the DefaultRequest handler

Stexxe:
https://youtrack.jetbrains.com/issue/KTOR-9632

### DIFF
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/DefaultRequest.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/DefaultRequest.kt
@@ -94,24 +94,23 @@ public class DefaultRequest private constructor(private val block: DefaultReques
                 }
                 val defaultUrl = defaultRequest.url.build()
                 mergeUrls(defaultUrl, context.url)
-                defaultRequest.attributes.allKeys.forEach {
-                    if (it == ENGINE_CAPABILITIES_KEY) {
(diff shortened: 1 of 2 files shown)


ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/DefaultRequest.kt
bjhham: Would be more idiomatic and readable with a `when` here, like so:

```suggestion
                defaultRequest.attributes.allKeys.forEach { key ->
                    when(key) {
                        ENGINE_CAPABILITIES_KEY -> {
                            context.attributes.computeIfAbsent(ENGINE_CAPABILITIES_KEY) {
                                mutableMapOf()
                            }.apply {
                                putAll(defaultRequest.attributes[ENGINE_CAPABILITIES_KEY] - keys)
                            }
                        }
                        !in context.attributes -> {
                            @Suppress("UNCHECKED_CAST")
                            context.attributes.put(key as AttributeKey<Any>, defaultRequest.attributes[key])
                        }
                    }
                }
```

Stexxe: I've changed the `if` to `when`, however, I don't want to use `putAll` because it requires allocation of a new list.

bjhham (approved):

osipxd (approved):

PR #5707, opened 2026-06-19
KTOR-9252 Incomplete gzip response causes client/server to hang

osipxd:
**Subsystem**
ktor-utils, content encoding

**Motivation**
[KTOR-9252](https://youtrack.jetbrains.com/issue/KTOR-9252): malformed or incomplete gzip content can leave the JVM `Inflater` waiting for more input after the source channel has already reached EOF. The decoder then keeps trying to inflate without making progress, which can hang client/server content decoding and burn CPU.

**Solution**
Detect incomplete compressed input after EOF and fail with `EOFException` instead of continuing the inflate loop. Added a focused JVM regression test that decodes malformed gzip directly through `GZip.decode`.

Validation: `:ktor-utils:jvmTest --tests io.ktor.tests.utils.EncodersJvmTest.testMalformedGzipFails --rerun` passed, 1 test executed.

### DIFF
--- a/ktor-utils/jvm/src/io/ktor/util/EncodersJvm.kt
+++ b/ktor-utils/jvm/src/io/ktor/util/EncodersJvm.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2019 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ * Copyright 2014-2026 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
  */
 
 package io.ktor.util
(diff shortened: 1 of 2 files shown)


ktor-utils/jvm/src/io/ktor/util/EncodersJvm.kt
```diff
@@ -139,6 +140,7 @@ private fun inflate(
         readBuffer.flip()
 
         while (!inflater.finished()) {
+            if (inflater.needsInput()) throw EOFException("Compressed input is incomplete.")
```
osipxd: For review: this code fragment goes after `source.isClosedForRead == true`, so `needsInput()` would never be satisfied.

zibet27 (approved):
lgtm 👍

bjhham (approved):

ktor-utils/jvm/src/io/ktor/util/EncodersJvm.kt
osipxd: Actually Codex is correct here and `needsInput()` can be `true` in a valid state. So changed this condition to `inflated == 0 && inflater.needsInput()`

PR #5708, opened 2026-06-21
KTOR-9670: Honor Accept-Encoding q=0 and most-specific matching

nickjn92:
KTOR-9670

## What
The server `Compression` plugin ignored `Accept-Encoding` quality values when choosing a response content-coding. It:
- never excluded a coding the client refused with `q=0`,
- never let an explicitly named coding override a `*` wildcard, and
- matched coding tokens case-sensitively.

For example, a request with `Accept-Encoding: gzip;q=0` still received a gzip-compressed response.

## Why
These contradict HTTP content negotiation (RFC 7231 §5.3.4): `q=0` means "not acceptable", the most specific `Accept-Encoding` entry wins, and coding tokens are case-insensitive.

## How
`encode()` now resolves each configured encoder's *effective* quality from its most specific matching entry (an explicit name overrides `*`, compared case-insensitively), drops codings whose effective quality is `0`, and orders the remaining encoders by quality then configured priority. The existing priority tie-break and the uncompressed fallback (when no encoder is eligible) are preserved. No public API change.

## Tests
New `CompressionAcceptEncodingTest` (15 cases) covers `q=0` exclusion, explicit-over-wildcard precedence (including positive qualities), case-insensitive matching, priority tie-breaks, and the uncompressed fallback. Existing `CompressionTest` (47) and the CIO engine `CompressionTestSuite` remain green.

Pre-flight (`formatKotlin`, `lintKotlin`, `checkKotlinAbi`, `assemble`) all pass; ABI unchanged.

## Risk
Behavior change for clients that send `q=0` or relied on the previous case-sensitive matching. No public API or signature changes.

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-compression/jvm/src/io/ktor/server/plugins/compression/Compression.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-compression/jvm/src/io/ktor/server/plugins/compression/Compression.kt
@@ -216,16 +216,13 @@ private suspend fun ContentDecoding.Context.decode(
     }
 }
 
+private class AcceptedEncoder(val config: CompressionEncoderConfig, val quality: Double)
+
 private fun ContentEncoding.Context.encode(call: PipelineCall, options: CompressionOptions) {
     if (call.response.isSSEResponse()) {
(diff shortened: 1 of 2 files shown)


bjhham (approved):
Looks good, thanks 👍

PR #5709, opened 2026-06-22
KTOR-9606 Fix ConverterTest for web targets

bjhham:
**Subsystem**
Serialization

**Motivation**
Looks like we missed a test failure introduced in https://github.com/ktorio/ktor/pull/5652

**Solution**
Returning null on empty content is usually invalid, so I moved the fix higher up the pipeline by checking if the content is empty with a suspension `awaitContent()`.  Now, nullables will properly return null for empty bodies.


Stexxe (approved):

osipxd (changes requested):
Now [one protobuf test](https://github.com/ktorio/ktor/commit/eb14d5347ca7dafdcc58d2c110aff1d895668f9e) started failing. That was the reason why this check was done "only if extensions not empy". But probably that proguard fix should be different 🤔

bjhham:
Checking protobuf serialization test failure 👀

bjhham:
I think I'll look at an alternative fix for the original issue.  Returning nulls on empty input seems to introduce inconsistency, and could be considered invalid behaviour.

osipxd (approved):
Nicely done!

PR #5710, opened 2026-06-22
KTOR-8766 Fix multiple invocations of close from DI

bjhham:
**Subsystem**
Server, DI

**Motivation**
[KTOR-8766](https://youtrack.jetbrains.com/issue/KTOR-8766) DI: The close method is called twice on cleanup of AutoCloseable

**Solution**
The multiple invocations were caused by iterating on "implicit" keys in the DI registry map.  Replaced the instance check with an exhaustive when so that we can be more explicit here, importantly avoiding all non-source keys.

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyInjection.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyInjection.kt
@@ -163,18 +163,23 @@ public val DI: ApplicationPlugin<DependencyInjectionConfig> =
             }
             monitor.subscribe(ApplicationStopping) {
                 for ((key, initializer) in dependencyMap.entries.reversed()) {
-                    if (initializer is DependencyInitializer.Ambiguous ||
-                        initializer is DependencyInitializer.Missing
-                    ) {
-                        continue
-                    }
-                    try {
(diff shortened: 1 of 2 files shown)


ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyInjection.kt
```diff
@@ -163,18 +163,23 @@ public val DI: ApplicationPlugin<DependencyInjectionConfig> =
...
+                        is DependencyInitializer.Implicit -> continue
+
+                        is DependencyInitializer.Explicit,
+                        is DependencyInitializer.Value -> {
+                            try {
+                                val instance = registry.getDeferred<Any?>(key).tryGetCompleted() ?: continue
+                                registry.shutdownHooks[key]?.invoke(instance)
+                                onShutdown(key, instance)
+                            } catch (e: Throwable) {
+                                if (e !is CancellationException) {
```
zibet27: Should we cancel the coroutine scope if e is CancellationException?

bjhham: Hmm yeah looks like it.

zibet27 (approved):

PR #5712, opened 2026-06-22
KTOR-9671: Fix semicolon character encoding for old darwin versions

nikialeksey:
**Subsystem**
ktor-client-darwin

**Motivation**
[KTOR-9671](https://youtrack.jetbrains.com/issue/KTOR-9671) Darwin: Semicolons in URL path are sanitized

I faced an issue, when the url like `https://example.com/segment1;param=value/segment2/` was encoded on darwin platform and become `https://example.com/segment1%3Bparam=value/segment2/`. Android/jvm platform work correctly, they don't sanitize the semicolon character (which actually is a valid character according to [RFC-3986](https://www.rfc-editor.org/info/rfc3986/#section-3.3)) for url path.

@osipxd reproduced it on old CI; I reproduced it when running ktor on 16.4 iOS simulator.

**Solution**
Just extend allowed character set with semicolon if it is not there (based on a [suggestion](https://github.com/ktorio/ktor/pull/5699#pullrequestreview-4545410108)).

### DIFF
--- a/ktor-client/ktor-client-darwin/darwin/src/io/ktor/client/engine/darwin/internal/DarwinUrlUtils.kt
+++ b/ktor-client/ktor-client-darwin/darwin/src/io/ktor/client/engine/darwin/internal/DarwinUrlUtils.kt
@@ -8,11 +8,21 @@ import io.ktor.http.*
 import io.ktor.util.*
 import platform.Foundation.*
 
+// Old version of macOS/iOS might not allow semicolon ';' in path for
+// compatibility with RFC 1808 (obsolete). However, ';' is a valid character according to RFC-3986.
+private val pathAllowedCharacters = if (NSCharacterSet.URLPathAllowedCharacterSet.characterIsMember(';'.code.toUShort())) {
+    NSCharacterSet.URLPathAllowedCharacterSet
+} else {
+    (NSCharacterSet.URLPathAllowedCharacterSet.mutableCopy() as NSMutableCharacterSet).apply {
+        addCharactersInString(";")
+    }
+}
+
 internal fun Url.toNSUrl(): NSURL {
     val userEncoded = encodedUser.orEmpty().isEncoded(NSCharacterSet.URLUserAllowedCharacterSet)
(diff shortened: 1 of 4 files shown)


osipxd (commented):
Thank you for the fix!
Could you also apply it in the `ktor-client-darwin-legacy` module?

ktor-client/ktor-client-darwin/darwin/src/io/ktor/client/engine/darwin/internal/DarwinUrlUtils.kt
```diff
@@ -8,11 +8,21 @@ import io.ktor.http.*
...
 
+// Old version of macOS/iOS might not allow semicolon ';' in path for
+// compatibility with RFC 1808 (obsolete). However, ';' is a valid character according to RFC-3986.
+private val pathAllowedCharacters = if (NSCharacterSet.URLPathAllowedCharacterSet.characterIsMember(';'.code.toUShort())) {
+    NSCharacterSet.URLPathAllowedCharacterSet
+} else {
+    (NSCharacterSet.URLPathAllowedCharacterSet.mutableCopy() as NSMutableCharacterSet).apply {
+        addCharactersInString(";")
+    }
+}
```
osipxd: Our linter doesn't like max  line length here. Adding a line break after the assignment operator would solve this:

```suggestion
private val pathAllowedCharacters: NSCharacterSet =
    if (NSCharacterSet.URLPathAllowedCharacterSet.characterIsMember(';'.code.toUShort())) {
        NSCharacterSet.URLPathAllowedCharacterSet
    } else {
        (NSCharacterSet.URLPathAllowedCharacterSet.mutableCopy() as NSMutableCharacterSet).apply {
            addCharactersInString(";")
        }
    }
```

nikialeksey: Thanks! Done

nikialeksey:
> Could you also apply it in the `ktor-client-darwin-legacy` module?

@osipxd Yep, done. Check it please

osipxd (approved):
Nicely done 👍  Thank you for the fix!

PR #5713, opened 2026-06-23
Make huge post test less huge

bjhham:
Looks like this test is flaky on CI from timeouts on the JVM platform.  I'll see if a smaller payload will resolve the issue - since our default buffer size is 1024^2, exceeding that should be all that's needed for exercising the relevant code.


osipxd (approved):
330ms instead of 30 seconds for the test execution. Nice :)

PR #5714, opened 2026-06-23
Fix cache misses with content negotiation 3x

matteomazza91:
**Subsystem**
Client, HttpCache

**Motivation**
When a `Vary` header contains multiple values, requests always generate cache-misses and  `If-None-Match` / `If-Modified-Since` are never sent

**Solution**
Centralize multi-values header serialization in one place to force the same serialization approach (i.e. use `,` instead of `;`) 

**Use Case**

When `HttpCache` and `ContentNegotiation` are both installed, responses with `Accept` in the `Vary` header (e.g. `Vary: Accept`) are **never served from cache** if the client specifies a specific Accept header.

e.g. to make requests to github, the client should use `Accept: application/vnd.github+json` but the ContentNegotiation plugin also unconditionally appends an `Accept` for every registered codec:

```kotlin
registrations.forEach { request.accept(it.contentTypeToSend) }
```

When the caller also sets `Accept` (e.g. `accept(ContentType.parse("application/vnd.github+json"))`), the request ends up with **two** `Accept` values. The stored varyKey and the lookup key then differ only in separator:

| | Value |
|---|---|
| **Stored** (`joinToString(",")`) | `"application/vnd.github+json,application/json"` |
| **Lookup** (`joinToString(";")`) | `"application/vnd.github+json;application/json"` |

## Ktor version

Reproduced on **3.5.0-dev** (main branch as of 2026-05-15) and **2.3.7**.

In 2.3.7: `varyKeys()` stored values using `Headers.get()` (first value only) while `mergedHeadersLookup` used `getAll().joinToString(";")`. An earlier fix changed `get()` to `getAll().joinToString(",")` but left the separator in `mergedHeadersLookup` as `";"`, so the mismatch persists.

## Platform / engine

JVM — the bug is in `ktor-client-core`, not engine-specific.

### Option A — Checkout the failing tests from this fork

```bash
git clone https://github.com/ktorio/ktor.git
cd ktor
# Check out the commit that adds the three failing tests:
git checkout 6b806fb2d

# Run only the relevant test class (uses MockEngine, no server needed):
./gradlew :ktor-client-core:jvmTest \
  --tests "HttpCacheTest.varyKeysStoredWithCommaSeparatorButMergedHeadersLookupUsesSemicolon" \
  --tests "HttpCacheTest.freshCacheableEntryWithVaryAcceptIsNeverServedFromCacheWhenContentNegotiationIsInstalled" \
  --tests "HttpCacheTest.freshCacheableEntryWithVaryAcceptIsServedFromCacheWhenContentNegotiationIsNotInstalled"
```

Expected: 2 tests fail (the unit test and the integration test), 1 passes (the counterpart without `ContentNegotiation`).

## Expected behaviour

The second GET should be served from in-memory cache (`max-age=60` has not expired). `serverCallCount == 1`.

## Actual behaviour

`serverCallCount == 2`. The cache entry is stored after the first request but **never matched** on subsequent requests.

### DIFF
--- a/ktor-client/ktor-client-core/build.gradle.kts
+++ b/ktor-client/ktor-client-core/build.gradle.kts
@@ -34,6 +34,7 @@ kotlin {
             implementation(projects.ktorTestBase)
             implementation(projects.ktorClientMock)
             implementation(projects.ktorServerTestHost)
+            implementation(projects.ktorClientContentNegotiation)
         }
     }
 }
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/cache/HttpCache.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/cache/HttpCache.kt
@@ -395,8 +395,8 @@ internal fun mergedHeadersLookup(
         }
 
         else -> {
-            val value = content.headers.getAll(header) ?: allHeadersExtractor(header) ?: emptyList()
-            value.joinToString(";")
+            val value = content.headers.getAll(header) ?: allHeadersExtractor(header)
+            value.joinHeaderValues()
         }
     }
 }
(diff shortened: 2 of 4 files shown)


bjhham (approved):

bjhham:
For reference in YT:
[KTOR-9673](https://youtrack.jetbrains.com/issue/KTOR-9673) Caching always missing with Content Negotiation

PR #5715, opened 2026-06-23
Test UNIX sockets on all platforms where they are supported

whyoleg:
**Subsystem**
CIO/Unix Sockets

**Motivation**
UNIX sockets are properly supported by Node.js, making it possible to test all related functionality not only on the JVM.

**Solution**
"commonize" tests: now they run on all targets where ktor-network/CIO is supported. :)

**NOTE**
Because of https://github.com/Kotlin/kotlinx-io/issues/345, we can't run those tests on JS ([internal thread](https://jetbrains.slack.com/archives/CB3CMFE68/p1782143468169379)), as Ktor enables ES modules for tests. I've mentioned the issue directly in the code, but I could also create a KTOR issue if needed.

PR #5717, opened 2026-06-23
KTOR-9674: Support UDP sockets via Node.js APIs

whyoleg:
**Subsystem**
ktor-network: UDP

**Motivation**
TCP sockets are supported by all targets. UDP sockets were supported by JVM/native targets.
With this PR, ktor-network supports **all** socket types (TCP, UDP, UNIX) on **all** targets.

**Solution**
Implement sockets via Node.js UDP APIs, similar to those for TCP. Though those are provided in two different node modules, they also have entirely different APIs. The implementation also differs a lot.

I would suggest reviewing commit-by-commit, where the first few commits are preparatory, allowing us to share more code between implementations and reducing the number of abstractions (by dropping `SocketContext`) to simplify the code flow.

Also, I've taken the liberty of moving the `DNSResolverTest` tests to common, since all targets are now supported.

~~BTW, should we create an issue about it? The only request for this was in the KotlinLang Slack [thread](https://kotlinlang.slack.com/archives/C0A974TJ9/p1779598121658479).~~
[KTOR-9674](https://youtrack.jetbrains.com/issue/KTOR-9674)


bjhham:
Here's a YouTrack ticket for reference:
[KTOR-9674](https://youtrack.jetbrains.com/issue/KTOR-9674) UDP support for web targets

ktor-network/common/src/io/ktor/network/sockets/DatagramSendChannelBase.kt
```diff
@@ -0,0 +1,124 @@
...
+            if (handler === CLOSED_INVOKED) break
+            if (handler == null) {
+                if (onCloseHandler.compareAndSet(null, CLOSED)) break
+                continue
+            }
+
+            require(onCloseHandler.compareAndSet(handler, CLOSED_INVOKED))
+            handler(closedCause.value)
+            break
+        }
```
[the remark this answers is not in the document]
whyoleg: Overall, I agree, but this was a copy-pasted code in JVM/native, so it's better to do it in a separate follow-up PR.

ktor-network/wasmJs/src/io/ktor/network/sockets/nodejs/node.wasmJs.kt
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+package io.ktor.network.sockets.nodejs
+
+internal actual fun JsError.toThrowable(): Throwable = Error(message)
+internal actual fun Throwable.toJsError(): JsError? = jsError(message)
+
+private fun jsError(message: String?): JsError = js("(new Error(message))")
+
+internal actual fun jsTypeOf(a: JsAny): String = js("(typeof obj)")
```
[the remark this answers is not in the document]
whyoleg: Oops... The only place where it really matters is in UNIX sockets for addresses. I will add a test for this after https://github.com/ktorio/ktor/pull/5715 is merged.

ktor-network/web/src/io/ktor/network/sockets/SocketEngine.udp.web.kt
```diff
@@ -0,0 +1,139 @@
...
+private class DatagramSocketImpl(
+    override val socketContext: CompletableJob,
+    private val incomingChannel: Channel<Datagram>,
+    private val socket: DgramSocket,
+    remote: SocketAddress?,
+) : BoundDatagramSocket, ConnectedDatagramSocket {
+    override val incoming: ReceiveChannel<Datagram> get() = incomingChannel
+    override val outgoing: SendChannel<Datagram> = DatagramSendChannel(socket, this, remote)
+    override val localAddress: SocketAddress get() = socket.address().toSocketAddress()
+    override val remoteAddress: SocketAddress get() = socket.remoteAddress().toSocketAddress()
```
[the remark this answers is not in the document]
whyoleg: It follows the same pattern as UDP sockets in other targets.
While working on this PR, I had an idea for how to improve this, but it would require a big change to the sockets, so I've postponed it to avoid blocking this and breaking anything. I might return to it in future PRs.

ktor-network/web/src/io/ktor/network/sockets/SocketEngine.udp.web.kt
```diff
@@ -0,0 +1,139 @@
...
+                address = target?.hostname
+            ) { error ->
+                if (error != null) {
+                    cont.resumeWithException(IOException("Failed to send datagram", error.toThrowable()))
+                } else {
+                    cont.resume(Unit)
+                }
+            }
+        }
+    }
```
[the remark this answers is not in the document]
whyoleg: I could replace it with `require`.
Node.js doesn't support UDP-over-unix sockets.

ktor-network/jvm/src/io/ktor/network/sockets/DatagramSendChannel.kt
```diff
@@ -8,53 +8,51 @@ import io.ktor.network.selector.*
...
+                if (result) {
+                    element.packet.discard()
+                }
+            }
         }
-
-        closeAndCheckHandler()
-
         return true
     }
```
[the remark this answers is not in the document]
whyoleg: FYI (to both agent and maintainers): even before the PR, `trySend` implementations in JVM and POSIX were not really correct as per the contract of the `SendChannel.trySend` function. First of all, they could throw an error, while `trySend` should never throw. Second, what is reported above by the agent - we could return `ChannelResult.success` even in case we didn't really send the datagram.

Possible solutions:
- Fix those two issues by returning `false` from `trySendImpl` in case of an error or when the datagram was not written.
- Mimic the behavior of a Node.js-based implementation and always return `false`.

It really depends on how useful `trySend` is for datagrams. I could implement any of the solutions, but it's probably better to do this in a separate PR, as in the reply above (https://github.com/ktorio/ktor/pull/5717#discussion_r3461312219), so that this PR will be mostly about the web targets.

ktor-network/jvm/test/io/ktor/network/sockets/tests/TestUtilsJvm.kt
```diff
@@ -4,6 +4,13 @@
...
 
+import java.util.*
+
 internal actual fun Throwable.isPosixException(): Boolean = false
 
 actual fun initSocketsIfNeeded() {}
+
+actual fun isJvmWindows(): Boolean {
+    val os = System.getProperty("os.name", "unknown").lowercase(Locale.getDefault())
+    return os.contains("win")
```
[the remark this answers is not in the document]
whyoleg: It was just moved from another file ktor-network/jvm/test/io/ktor/network/sockets/tests/UDPSocketTestJvm.kt, it's fine.

ktor-client/ktor-client-cio/common/test/CioDnsResolverTest.kt
```diff
@@ -6,30 +6,26 @@ package io.ktor.client.engine.cio
...
 
-    @Test
-    fun testJvmDnsResolverResolvesLocalhost() = runBlocking {
-        val ips = JvmDnsResolver().invoke("localhost")
-        assertTrue(ips.isNotEmpty(), "localhost must resolve to at least one address")
-        assertTrue(
-            ips.any { it == "127.0.0.1" || it == "0:0:0:0:0:0:0:1" || it == "::1" },
-            "expected loopback in $ips",
-        )
+    private fun runTestWithNoDelaySkipping(testBody: suspend CoroutineScope.() -> Unit): TestResult = runTest {
```
osipxd: I think `runTestWithRealTime` could be used here

ktor-client/ktor-client-cio/jvm/test/io/ktor/client/engine/cio/JvmDnsResolverTest.kt
```diff
@@ -0,0 +1,47 @@
...
+        val ips = JvmDnsResolver().invoke("localhost")
+        assertTrue(ips.isNotEmpty(), "localhost must resolve to at least one address")
+        assertTrue(
+            ips.any { it == "127.0.0.1" || it == "0:0:0:0:0:0:0:1" || it == "::1" },
+            "expected loopback in $ips",
+        )
+    }
+
+    @Test
+    fun testJvmDnsResolverIsCooperativelyCancellable() = runBlocking {
```
osipxd: Minor. Let's replace `runBlocking` here as well

ktor-network/js/src/io/ktor/network/sockets/nodejs/node.js.kt
```diff
@@ -0,0 +1,38 @@
...
+
+@JsFun(
+    "moduleName => ((typeof process !== 'undefined') && process.release.name === 'node')" +
+        " ? import(/* webpackIgnore: true */'node:net') : Promise.resolve(null)"
+)
+private external fun loadNodeNetModule(): Promise<NodeNet?>
+
+@JsFun(
+    "moduleName => ((typeof process !== 'undefined') && process.release.name === 'node')" +
+        " ? import(/* webpackIgnore: true */'node:dgram') : Promise.resolve(null)"
```
osipxd: Interesting. I wasn't able to use the same workaround as on wasmJs here because JS was ignoring/removing the `webpackIgnore: true` comment. But now I don't see this failure on CI 🤔

ktor-network/web/src/io/ktor/network/sockets/SocketEngine.udp.web.kt
```diff
@@ -5,16 +5,135 @@
...
+        val incomingDatagrams: Channel<Datagram> = Channel(Channel.UNLIMITED)
+
+        cont.invokeOnCancellation {
+            socket.close()
+            incomingDatagrams.cancel()
+            socketContext.cancel()
+        }
+        socketContext.invokeOnCompletion {
+            socket.close()
+            incomingDatagrams.cancel()
```
osipxd, line 53: Just curious. Won't `socket.close` and `incomingDatagrams.cancel` be called twice? First time from `cont.invokeOnCancellation` and second time from `socketContext.invokeOnCompletion`. However it should be fine as close/cancel should be idempotent.

osipxd (commented):
Thank you for the PR and thanks for commonizing things! Looks nice 👍 
The only concern I have is about browser-compatibility on JS (see my comment). I'll check it and will get back with results

ktor-network/js/src/io/ktor/network/sockets/nodejs/node.js.kt
osipxd: Yes, I see the error in logs (even though the build passes successfully):

```kotlin
Module build failed: UnhandledSchemeError: Reading from "node:dgram" is not handled by plugins (Unhandled scheme).
Webpack supports "data:" and "file:" URIs by default.
You may need an additional plugin to handle "node:" URIs.
    at ktor/build/js/node_modules/webpack/lib/NormalModule.js:1023:10
    at Hook.eval [as callAsync] (eval at create (ktor/build/js/node_modules/tapable/lib/HookCodeFactory.js:33:10), <anonymous>:6:1)
    at Hook.CALL_ASYNC_DELEGATE [as _callAsync] (ktor/build/js/node_modules/tapable/lib/Hook.js:20:14)
    at Object.processResource (ktor/build/js/node_modules/webpack/lib/NormalModule.js:1019:8)
    at processResource (ktor/build/js/node_modules/loader-runner/lib/LoaderRunner.js:220:11)
    at iteratePitchingLoaders (ktor/build/js/node_modules/loader-runner/lib/LoaderRunner.js:171:10)
    at runLoaders (ktor/build/js/node_modules/loader-runner/lib/LoaderRunner.js:398:2)
    at NormalModule._doBuild (ktor/build/js/node_modules/webpack/lib/NormalModule.js:1004:3)
    at NormalModule.build (ktor/build/js/node_modules/webpack/lib/NormalModule.js:1193:15)
    at ktor/build/js/node_modules/webpack/lib/Compilation.js:1547:12

Module build failed: UnhandledSchemeError: Reading from "node:net" is not handled by plugins (Unhandled scheme).
Webpack supports "data:" and "file:" URIs by default.
You may need an additional plugin to handle "node:" URIs.
    at ktor/build/js/node_modules/webpack/lib/NormalModule.js:1023:10
    at Hook.eval [as callAsync] (eval at create (ktor/build/js/node_modules/tapable/lib/HookCodeFactory.js:33:10), <anonymous>:6:1)
    at Object.processResource (ktor/build/js/node_modules/webpack/lib/NormalModule.js:1019:8)
    at processResource (ktor/build/js/node_modules/loader-runner/lib/LoaderRunner.js:220:11)
    at iteratePitchingLoaders (ktor/build/js/node_modules/loader-runner/lib/LoaderRunner.js:171:10)
    at runLoaders (ktor/build/js/node_modules/loader-runner/lib/LoaderRunner.js:398:2)
    at NormalModule._doBuild (ktor/build/js/node_modules/webpack/lib/NormalModule.js:1004:3)
    at NormalModule.build (ktor/build/js/node_modules/webpack/lib/NormalModule.js:1193:15)
    at ktor/build/js/node_modules/webpack/lib/Compilation.js:1547:12
    at NormalModule.needBuild (ktor/build/js/node_modules/webpack/lib/NormalModule.js:1534:32)
```
 
Let's try to use the previous solution with module names extracted into functions:

```kotlin
// Keep module names behind a function so browser bundlers don't resolve `node:` modules statically.
private fun nodeNetModuleName() = "node:net"
private fun nodeDgramModuleName() = "node:dgram"

@JsFun(
    "moduleName => ((typeof process !== 'undefined') && process.release.name === 'node')" +
        " ? import(moduleName) : Promise.resolve(null)"
)
private external fun loadNodeModule(moduleName: String): Promise<JsAny?>
```

This would also require casting `JsAny` to the actual type, but I think it's fine.

I'll check why our tests doesn't fail and the error is only printed in the log.

ktor-network/js/src/io/ktor/network/sockets/nodejs/node.js.kt
```diff
@@ -0,0 +1,38 @@
...
+)
+private external fun loadNodeGramModule(): Promise<NodeDgram?>
```
osipxd: Also noticed a minor typo here `Gram` -> `Dgram`. But this function will be removed if the fix for browser bundler applied

osipxd (changes requested):

ktor-client/ktor-client-cio/common/test/CioDnsResolverTest.kt
whyoleg: Nice idea!

The most interesting thing, though, is that it's me who added it. :)
image

BTW, the reason I missed it is that there are two places the test builders are currently at:
- `runTest` is in [ktor-test-base](https://github.com/ktorio/ktor/blob/54b57960d9a7a6802f336dc8197b55d78d90b617/ktor-shared/ktor-test-base/common/src/io/ktor/test/TestBuilders.kt#L18) module
- `runTestWithRealTime` is on [ktor-test-dispatcher](https://github.com/ktorio/ktor/blob/54b57960d9a7a6802f336dc8197b55d78d90b617/ktor-test-dispatcher/common/src/TestCommon.kt#L41) module

Do you mind, if in a separate PR, I will move `runTestWithRealTime` to `ktor-test-base` (deprecating the declaration in `ktor-test-dispatcher`) and replace all the usages with the ones from `io.ktor.test` package, as well as mark both `testSuspend` and `runTestWithRealTime` with `ERROR` deprecation level?

And so, `ktor-test-dispatcher` could be dropped completely in Ktor 4? WDYT? Do you also, by any chance, have a way to track things that could be done in the next MAJOR release?

ktor-network/js/src/io/ktor/network/sockets/nodejs/node.js.kt
whyoleg: Thanks for the investigation! Yeah, I've tried to make the code more similar and haven't found any errors during running tests. I will revert to what you have proposed.

ktor-network/web/src/io/ktor/network/sockets/SocketEngine.udp.web.kt
whyoleg: Hm, yeah, nice catch!
Though it could only happen if the coroutine in which the connect/bind was called is canceled, which is rare in practice.
Still, I've removed the duplication because the Node.js UDP API's `socket.close()` call is not idempotent! It will throw an exception if called twice... In TCP, `socket.destroy` will not throw.

ktor-network/jvm/src/io/ktor/network/sockets/DatagramSendChannel.kt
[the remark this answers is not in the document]
whyoleg: I've saved this to my notes and will create a PR to fix this and the other comment above in a separate PR later.

ktor-client/ktor-client-cio/common/test/CioDnsResolverTest.kt
bjhham: > Do you also, by any chance, have a way to track things that could be done in the next MAJOR release?

We don't have anything atm that I'm aware of, I'll spin up a doc.

osipxd (approved):
Great work!

ColtonIdle:
Thank you @olegvelikanov I was missing this feature for my project!

PR #5723, opened 2026-06-26
test-server: Fix deprecations

osipxd:


bjhham (approved):

osipxd:
We can't update tes test server to 3.5.1, yet. Because it depends on kotlinx.coroutines 1.11.0 and Gradle provides 1.10.2.

PR #5724, opened 2026-06-28
 KTOR-7290: Support wasm-wasi target in most of the modules (no engine yet)

whyoleg:
In some ways, the PR is similar to #5720, but now we not only commonize things but also add new implementations for `wasm-wasi` via minimal platform APIs. Every commit migrates one module; the biggest changes are in `io` and `utils`. `ktor-server-core` is also currently in bad shape, as most of the functionality there is not yet supported and might need changes once sockets are implemented properly.

Here are some notes collected during migration, so as to share the context:
- We need to update NodeJS to at least v24.4.0, as Kotlin uses [wasi.finalizeBindings](https://nodejs.org/api/wasi.html#wasifinalizebindingsinstance-options).
  - Should I extract it to a separate PR? The current version is from the Kotlin master branch.
- kotlinx-html is not published for wasm-wasi for some reason ([internal thread](https://jetbrains.slack.com/archives/CUHBZQSCC/p1782141177378069))
  - Could be un-ignored after this PR is merged.
- for some reason, I wasn't able to add `kotlinx-datetime` in the utils module to just `wasmWasiMain` source-set - IDE code was red, so I've added it to `commonMain` for now.
  - TBD what to do here properly.
- Is 'INFO' fine as the log level? Should we support configuration via `ENV` variables? It will be hard to test the implementation, as we can't mutate them in Wasm.
- cio/network has ignored tests because no sockets implemented yet - **this is the next thing I will work on.**
- `checkContentLength` and `shouldDecode` should really be "capabilities" of engines and not expect/actual, as currently they will work incorrectly for the CIO engine.
  - Should we create an issue about it?
- `platformAddShutdownHook` - should we throw instead of ignoring the hook, as it's not supported?
- `EmbeddedServer` - for now, just a copy of `web`, I will recheck after WASI server is ready - same for `EngineTestBase`.
- I've done some commonization of attributes, but I left `ConcurrentSafeAttributes` in place. Really, we could drop it too; the difference is in one wrapper class - WDYT?

Regarding some bigger changes than just commonization:
- `io`:
  - charsets logic on web rewritten a bit to only use `TextDecoder`; `TextEncoder` is dropped in favor of plain calls, as anyway, it supports only UTF8.
  - charsets logic is now fully commonized for `web`, as well as mostly covers wasm-wasi (except for the decoding part, which existed before).
  - AFAIK: wasm is LE by default, so `ByteOrder` uses it.
- `utils`:
  - attributes commonized a bit
  - `GMTDate` is implemented via kotlinx-datetime - overall, the same implementation can be used on all platforms.
  - random impl is based on [CK implementation](https://github.com/whyoleg/cryptography-kotlin/blob/14daa2f20842bc83e9d8d2d498aa3dda5d9db1c8/cryptography-random/src/wasmWasiMain/kotlin/CryptographyRandom.wasmWasi.kt).

**Next step: sockets via WASI, which will un-ignore most of the tests.**


whyoleg:
BTW, am I right that for this PR the TC configuration should also be updated to run `wasm-wasi` tests now, since those are done in separate build steps?

ktor-utils/common/src/io/ktor/util/PlatformUtils.kt
```diff
@@ -40,6 +41,7 @@ public sealed class Platform {
     public data object Native : Platform()
     public data class Js(val jsPlatform: JsPlatform) : Platform()
     public data class WasmJs(val jsPlatform: JsPlatform) : Platform()
+    public data object WasmWasi : Platform()
```
whyoleg, line 44: THIS IS A SOURCE BREAKING CHANGE

ktor-utils/api/ktor-utils.klib.api
```diff
@@ -62,6 +62,7 @@ final enum class io.ktor.util.logging/LogLevel : kotlin/Enum<io.ktor.util.loggin
     enum entry DEBUG // io.ktor.util.logging/LogLevel.DEBUG|null[0]
     enum entry ERROR // io.ktor.util.logging/LogLevel.ERROR|null[0]
     enum entry INFO // io.ktor.util.logging/LogLevel.INFO|null[0]
+    enum entry NONE // io.ktor.util.logging/LogLevel.NONE|null[0]
```
whyoleg, line 65: THIS IS A SOURCE BREAKING CHANGE because of commonization - can be easily reverted

ktor-utils/nonJvmAndPosix/src/io/ktor/util/Attributes.nonJvmAndPosix.kt
```diff
@@ -12,8 +12,9 @@ import kotlin.js.*
  * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.util.Attributes)
  */
 @JsName("AttributesJsFn")
-public actual fun Attributes(concurrent: Boolean): Attributes = AttributesJs()
+public actual fun Attributes(concurrent: Boolean): Attributes = HashMapAttributes()
 
+@Deprecated("Use `Attributes` function", level = DeprecationLevel.WARNING)
 public class AttributesJs : Attributes {
```
whyoleg, line 18: move back to `web`

ktor-server/ktor-server-core/wasmWasi/src/io/ktor/server/engine/ShutdownHook.wasmWasi.kt
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright 2014-2025 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.server.engine
+
+// it's not possible to set up hook
+internal actual val SHUTDOWN_HOOK_ENABLED = false
+
+internal actual fun EmbeddedServer<*, *>.platformAddShutdownHook(stop: () -> Unit) {}
```
whyoleg, line 10: throw an error here

ktor-server/ktor-server-core/wasmWasi/src/io/ktor/server/engine/EnvironmentUtils.wasmWasi.kt
```diff
@@ -0,0 +1,17 @@
+/*
+ * Copyright 2014-2026 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.server.engine
+
+internal actual fun getKtorEnvironmentProperties(): List<Pair<String, String>> = emptyList()
```
whyoleg, line 7: throw an error in every function

build-logic/src/main/kotlin/ktorbuild/targets/KtorTargets.kt
```diff
@@ -203,18 +204,26 @@ abstract class KtorTargets @Inject internal constructor(
...
                 }
 
+                // has `runBlocking` and IO
                 group("jvmAndPosix") {
                     withJvm()
                     group("posix")
                 }
 
+                // no `runBlocking` and IO
+                group("nonJvmAndPosix") {
```
whyoleg, line 214: potentially rename to `nonJvmOrPosix`?

ktor-utils/common/src/io/ktor/util/PlatformUtils.kt
whyoleg: This change should be reverted.
Depends on https://youtrack.jetbrains.com/issue/KTOR-9693

ktor-utils/api/ktor-utils.klib.api
whyoleg: This change should be reverted.

build-logic/src/main/kotlin/ktorbuild/targets/KtorTargets.kt
bjhham: Technically would be `notJvmOrPosix` if you wanted to be grammatically correct since `non-` is a single-word prefix.  We could also shorten these both down to a portmanteau like `Javix` idk 🤷

bjhham (commented):
Looks good (besides existing comments and conflicts) 👍

PR #5679, opened 2026-06-03
KTOR-7468 - Add Kotlin Uuid conversion support to DefaultConversionService

nomisRev:
**Subsystem**
ktor-utils

**Motivation**
Currently we do not support Kotlin Uuid in `DefaultConversionService`.

**Solution**
Add support for Kotlin Uuid in `DefaultConversionService`.

~Created as draft since this should await Kotlin 2.4.0 version bump, and before merging we should remove `@OptIn(ExperimentalUuidApi::class)`~

PR #5728, opened 2026-07-01
Add support for Byte, UByte, UShort, UInt, ULong, and Kotlin Uuid conversions in DefaultConversionService

nomisRev:
**Subsystem**
Server

**Motivation**
Currently in `DefaultConversionService` we did not support `Byte`, `java.lang.Byte`, `UByte`, `UShort`, `UInt` or `ULong`.
We were also missing `Uuid` in `toValues` in #5679

**Solution**
Added support for these types in the existing `DefaultConversionService`.


bjhham (approved):
Thanks!  This will be quite handy with Exposed!

zibet27:
> ReferenceError: crypto is not defined

Code looks good; please check this error in the JS target

PR #5734, opened 2026-07-06
KTOR-9683 Close ActorSelectorManager when its job is canceled

osipxd:
**Subsystem**
Network

**Motivation**
[KTOR-9683](https://youtrack.jetbrains.com/issue/KTOR-9683) Sockets: Canceling a SelectorManager job doesn't close it properly

**Solution**
Close `ActorSelectorManager` when its parent job starts cancelling, while preserving the existing unintercepted selector wakeup path. Added a JVM regression test for parent-job cancellation.

Validation: `:ktor-network:jvmTest --tests io.ktor.network.selector.ActorSelectorManagerTest --rerun` passed, 3 tests.


bjhham (approved):
Nice

PR #5738, opened 2026-07-09
KTOR-9698 Include discriminator property in sealed subtype schemas

sya-ri:
**Subsystem**
OpenAPI schema inference

**Motivation**
Sealed types include their discriminator field in serialized JSON, but the generated subtype schemas only included the regular subtype properties. This made each `oneOf` variant miss the discriminator property that is present in the payload.

**Solution**
When building schemas for sealed types, add the discriminator property to each inline subtype schema as a single-value string enum. Kotlinx serialization uses the sealed element serial name, and reflection uses the same subclass name that is already used in the discriminator mapping.

Existing subtype properties are preserved, and schemas that are emitted as references remain references.

**Tests**
- Updated schema snapshots for sealed types with default and custom discriminators.

**Issue**
KTOR-9698

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-routing-openapi/jvm/test-resources/expected/openapi.json
+++ b/ktor-server/ktor-server-plugins/ktor-server-routing-openapi/jvm/test-resources/expected/openapi.json
@@ -54,11 +54,18 @@
         "type": "object",
         "title": "DM",
         "required": [
+          "type",
           "id",
           "content",
           "timestamp"
         ],
         "properties": {
+          "type": {
+            "type": "string",
+            "enum": [
+              "io.ktor.server.routing.openapi.Message.DM"
+            ]
+          },
           "id": {
             "type": "integer"
           },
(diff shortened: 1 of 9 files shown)


bjhham (approved):
Thanks for the fix, looks good! 👍

sya-ri:
@bjhham Thanks. I removed the helper method because, during self-review, it felt too small to justify extracting.

sya-ri:
This is ready for review!

bjhham (changes requested):
Looks like you just missed a couple of model updates for the assertions in `DescribeRouteTest`

Could you also rebase onto `release/3.x` so that it will be included in the patch later this month?

bjhham:
Oh, also please run `./gradlew formatKotlin` to get past the linter check 🙏

sya-ri:
Thank you for the review. I've rebased this PR onto `release/3.x` and updated the tests.

bjhham (approved):
Looks good, thanks!

PR #5739, opened 2026-07-09
Fix a hidden bug with javaClass

zhelenskiy:
**Subsystem**
Client

**Motivation**
`something::javaClass.name` is not the name of the class of something but `"javaClass"`

**Solution**
Replace property reference with property access.


bjhham (approved):
Oof

PR #5741, opened 2026-07-10
KTOR-9702 Resolve duplication between readRemaining and readBuffer

bjhham:
**Subsystem**
Core, I/O

**Motivation**
[KTOR-9702](https://youtrack.jetbrains.com/issue/KTOR-9702) Duplicate code in ByteReadChannel

**Solution**
We have `readRemaining` and `readBuffer` that do the exact same thing.  Since `readBuffer()` returns `Buffer`, it's slightly more useful, though the naming is a bit confusing with the `readBuffer` property.  Alternatively, we could use `readRemaining` as the implementation and change its return type to `Buffer` since it's a sub-type.

Also, I looked through either implementation and chose the one that's a bit faster, and inlined a couple of functions after reading through the code in `kotlinx-io`.


osipxd (approved):
Nice catch 👍

PR #5742, opened 2026-07-10
KTOR-9703 Optimization for String functions

bjhham:
**Subsystem**
Core, I/O

**Motivation**
[KTOR-9703](https://youtrack.jetbrains.com/issue/KTOR-9703) Inefficient text reading function in DefaultTransform

**Solution**
While I was looking at the other issue regarding duplicate functions in `ByteReadChannelOperations`, I noticed we had some weird code for reading text in `DefaultTransform` that creates an input stream reader and decodes from there.  It's a lot more efficient to transfer to a `kotlinx-io` `Buffer` and use `readString(Charset)`.  We also have needlessly prevented decoding other charsets in the web targets when we have some multi-platform decoder stuff available.


osipxd (approved):
👍

PR #5743, opened 2026-07-11
[KTOR-9704] Optimized encoding flow for UTF-8

inemtsev:
Resolves [KTOR-9704](https://youtrack.jetbrains.com/issue/KTOR-9704) (full analysis, benchmarks, and an alternate-solution proposal live in the ticket).

**Subsystem**
Core, ktor-io

**Motivation**

`String.toByteArray(UTF_8)` in ktor-io encodes through Kotlin's validating char-by-char loop. On the JVM this bypasses `String.getBytes`, which is intrinsic-accelerated: with compact strings, encoding ASCII is a SIMD scan plus an array copy. Since `TextContent` encodes eagerly at construction, every kotlinx/Gson response, every `respondText`, and Jackson with `streamBody = false` pay this cost proportionally to body size.

**Solution**
Per the review discussion on the ticket, this adopts the non-validating variant (solution 2): the UTF-8 branch calls `encodeToByteArray()` with no arguments, restoring the KTOR-4224 form. On the JVM that compiles to `String.getBytes(UTF_8)`; JS/Wasm/Native keep their stdlib encoder.

**Behavior note**: a string containing an unpaired surrogate now encodes with a replacement character instead of throwing `CharacterCodingException`. This matches the decode direction (which already replaces), the non-UTF-8 encode branch on JVM (which already replaces via `getBytes`), the Kotlin stdlib default, and Ktor 2.x behavior (encode never threw before the migration).

**Benchmarks**:
Results (JMH, JDK 21, avgt, 2 forks, through `String.toByteArray`):

| Payload | Before | After | Speedup |
|---|---|---|---|
| 960 B JSON | 1,909 ns | 26 ns | 73x |
| 16 KB JSON | 32,237 ns | 686 ns | 47x |
| 100 KB JSON | 197,189 ns | 2,337 ns | 84x |
| 16 KB with emoji | 33,223 ns | 11,074 ns | 3.0x |

End to end (Dockerized Netty, 4 CPUs, oha, 64 connections; the 100 KB rows verified over 3 repetitions per side with alternating order, spread < 0.5%):

| Endpoint | Before | After | Delta | 
|---|---|---|---|
| kotlinx GET, 100 KB list | 9,885 req/s | 21,969 req/s | 2.2x | 
| Gson GET, 100 KB list | 7,762 req/s | 13,602 req/s | +75% | 
| kotlinx GET, 960 B object | 61,478 req/s | 65,650 req/s | +7% | 
| kotlinx POST, receive-only (control) | 59,517 req/s | 60,318 req/s | within noise |


ktor-io/common/src/io/ktor/utils/io/core/Strings.kt
```diff
@@ -9,7 +9,10 @@ import kotlinx.io.*
 import kotlin.math.*
 
 public fun String.toByteArray(charset: Charset = Charsets.UTF_8): ByteArray {
-    if (charset == Charsets.UTF_8) return encodeToByteArray(throwOnInvalidSequence = true)
+    // Non-validating on purpose: unpaired surrogates are replaced instead of throwing,
+    // matching the decode direction, and on the JVM the argument-less overload compiles
+    // to the intrinsic-accelerated String.getBytes (KTOR-9704).
+    if (charset == Charsets.UTF_8) return encodeToByteArray()
```
bjhham: You can remove this comment, the git and ticket history should be sufficient.

bjhham (changes requested):

ktor-io/common/src/io/ktor/utils/io/core/Strings.kt
inemtsev: @bjhham Done

bjhham (approved):

osipxd (approved):

PR #5744, opened 2026-07-11
KTOR-9759 Reject whitespace in URL host during parsing

### LINKED ISSUE #5718: parseUrl() accepts urls with spaces in them
Using `io.ktor:ktor-http` version 3.5.0 and given the invalid url "http:// example.com" (incl. the space inbetween), parseUrl() returns an Url-instance instead of null.

````kotlin
val url = parseUrl("http:// example.com")
// url should be null, instead it is a valid Url-Object
````

### PR DESCRIPTION
**Pull request by codepawpaw**
**Subsystem**
Ktor HTTP (ktor-http) — URL parsing (URLParser / parseUrl, URLBuilder.takeFrom). Affects both client and server since both rely on this shared URL parsing.

**Motivation**
parseUrl accepted malformed URLs whose host component contained whitespace (e.g. http:// example.com, http://exa mple.com, http://example.com :8080), returning a Url instead of rejecting them. A host with spaces is not a valid authority, and silently accepting it can lead to surprising downstream behavior (bad requests, unexpected host values). parseUrl is documented to return null for invalid URLs, so these inputs should be rejected rather than parsed.

Fixes KTOR-9718.

**Solution**
In URLParser.fillHost, validate the extracted host and reject it if it contains any whitespace character:

val hostValue = urlString.substring(startIndex, colonIndex)
require(hostValue.none { it.isWhitespace() }) {
    "Host cannot contain whitespace characters: $hostValue"
}
host = hostValue
The thrown IllegalArgumentException is caught by parseUrl, which then returns null as documented. Added UrlTest.testParseUrlWithSpaceInHost covering a space before the host, inside the host, and before the port.

### DIFF
--- a/ktor-http/common/src/io/ktor/http/URLParser.kt
+++ b/ktor-http/common/src/io/ktor/http/URLParser.kt
@@ -206,7 +206,11 @@ private fun URLBuilder.parseFragment(urlString: String, startIndex: Int, endInde
 private fun URLBuilder.fillHost(urlString: String, startIndex: Int, endIndex: Int) {
     val colonIndex = urlString.indexOfColonInHostPort(startIndex, endIndex).takeIf { it > 0 } ?: endIndex
 
-    host = urlString.substring(startIndex, colonIndex)
+    val hostValue = urlString.substring(startIndex, colonIndex)
+    require(hostValue.none { it.isWhitespace() }) {
+        "Host cannot contain whitespace characters: \"$hostValue\""
+    }
+    host = hostValue
 
     port = if (colonIndex + 1 < endIndex) {
         urlString.substring(colonIndex + 1, endIndex).toInt()
--- a/ktor-http/common/test/io/ktor/tests/http/UrlTest.kt
+++ b/ktor-http/common/test/io/ktor/tests/http/UrlTest.kt
@@ -364,6 +364,13 @@ class UrlTest {
     }
 
     @Test
+    fun testParseUrlWithSpaceInHost() {
+        assertEquals(null, parseUrl("http:// example.com"))
+        assertEquals(null, parseUrl("http://exa mple.com"))
+        assertEquals(null, parseUrl("http://example.com :8080"))
+    }
+
+    @Test
     fun testAboutUrl() {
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Review by bjhham — APPROVED** [review]
The only problem right now is that I can't find KTOR-9718 in YouTrack.  Maybe it was deleted?

File: ktor-http/common/src/io/ktor/http/URLParser.kt [hunk 1]
```diff
@@ -206,7 +206,11 @@ private fun URLBuilder.parseFragment(urlString: String, startIndex: Int, endInde
 private fun URLBuilder.fillHost(urlString: String, startIndex: Int, endIndex: Int) {
     val colonIndex = urlString.indexOfColonInHostPort(startIndex, endIndex).takeIf { it > 0 } ?: endIndex
 
-    host = urlString.substring(startIndex, colonIndex)
+    val hostValue = urlString.substring(startIndex, colonIndex)
+    require(hostValue.none { it.isWhitespace() }) {
+        "Host cannot contain whitespace characters: $hostValue"
```
bjhham: Let's wrap the value in the message with quotes so whitespace at the end of the string is visible.

bjhham: Alternatively, would it make sense to just encode whitespace characters instead?

codepawpaw: I think  rejecting is the correct behavior for the host. Encoding would produce a broken host.  If we encoded exa mple.com into exa%20mple.com, we'd hand a hostname containing %20 to DNS resolution. No such host exists and %20 is not valid in a registered name (RFC 3986 reg-name)

osipxd (commented):
Could you run `./switch-base-branch.sh` so we could include this fix into the next patch release?

codepawpaw:
> Could you run `./switch-base-branch.sh` so we could include this fix into the next patch release?

Done, rebased onto release/3.x via switch-base-branch.sh

osipxd:
YouTrack issue: [KTOR-9759](https://youtrack.jetbrains.com/issue/KTOR-9759) Reject whitespace in URL hosts during parsing

PR #5745, opened 2026-07-11
KTOR-8755 Percent-encode Unicode characters in parsed URL paths

EarzuChan:
**Subsystem**

HTTP / URL parsing; Client / CIO

**Motivation**

Fixes [KTOR-8755](https://youtrack.jetbrains.com/issue/KTOR-8755/CIO-The-engine-doesnt-encode-Unicode-symbols-like-U2013-in-request-URL).

When a URL string contains raw non-ASCII characters in its path, `URLParser` currently copies the path directly into `encodedPathSegments`.

For example:

```text
https://assets.example.org/library/annual–review/cover-image.webp
```

This leaves `Url.encodedPath` containing the raw en dash even though the property represents an encoded path. CIO then writes `Url.fullPath` directly into the HTTP request line, which can cause servers to reject the request with `400 Bad Request`.

Browsers and clients such as OkHttp send the en dash as `%E2%80%93`.

**Solution**

Percent-encode paths while parsing URL strings in `URLParser`, before they are stored as encoded path segments.

The parser uses `encodeURLPath(encodeEncoded = false)`, which:

- encodes raw Unicode and other characters that are not valid in a URI path;
- preserves existing `%HH` sequences to avoid double encoding;
- keeps decoded path segments available through `rawSegments`.

The same normalization is applied to absolute URLs, relative URLs, and all supported `file:` URL forms. Query parameters and fragments remain unchanged.

With the URL model producing a correctly encoded path, no CIO-specific production workaround is required.

Tests cover:

- Unicode characters in absolute and relative URL paths;
- preservation of existing percent-encoded characters;
- separation of path encoding from query and fragment parsing;
- `file:/`, `file://host/`, and `file:///` URL forms;
- origin-form and proxy absolute-form CIO request targets.

**Testing**

- `./gradlew :ktor-http:jvmTest`
- `./gradlew :ktor-client-cio:jvmTest --tests "io.ktor.client.engine.cio.CIORequestLineTest"`
- `./gradlew :ktor-http:formatKotlin :ktor-client-cio:formatKotlin`
- `./gradlew :ktor-http:lintKotlin :ktor-client-cio:lintKotlin`

### DIFF
--- a/ktor-client/ktor-client-cio/common/test/CIORequestLineTest.kt
+++ b/ktor-client/ktor-client-cio/common/test/CIORequestLineTest.kt
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2014-2026 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.client.engine.cio
+
+import io.ktor.client.request.*
+import io.ktor.client.utils.*
+import io.ktor.http.*
+import io.ktor.test.*
+import io.ktor.util.*
+import io.ktor.utils.io.*
+import kotlinx.coroutines.*
+import kotlin.test.*
+
+internal class CIORequestLineTest {
+    @Test
+    fun `use encoded path in origin-form request target`() = runTest {
+        assertEquals(
+            "GET /library/annual%E2%80%93review/cover-image.webp?size=large&download= HTTP/1.1",
+            requestLine("https://assets.example.org/library/annual–review/cover-image.webp?size=large&download=")
+        )
+    }
+
+    @Test
+    fun `use encoded path in absolute-form request target`() = runTest {
+        assertEquals(
+            "GET https://assets.example.org/library/annual%E2%80%93review/cover-image.webp HTTP/1.1",
+            requestLine("https://assets.example.org/library/annual–review/cover-image.webp", overProxy = true)
+        )
+    }
+
+    @OptIn(InternalAPI::class)
+    private suspend fun requestLine(url: String, overProxy: Boolean = false): String {
+        val output = ByteChannel(autoFlush = true)
+        val request = HttpRequestData(
+            url = Url(url),
+            method = HttpMethod.Get,
+            headers = Headers.Empty,
+            body = EmptyContent,
+            executionContext = Job(),
+            attributes = Attributes()
+        )
+
+        writeHeaders(request, output, overProxy)
+        return output.readLineStrict()!!
+    }
+}
--- a/ktor-http/common/test/io/ktor/tests/http/URLParserTest.kt
+++ b/ktor-http/common/test/io/ktor/tests/http/URLParserTest.kt
@@ -1,0 +1,72 @@
+/*
+ * Copyright 2014-2026 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.tests.http
+
+import io.ktor.http.*
+import kotlin.test.*
+
+internal class URLParserTest {
+    @Test
+    fun `encode unicode characters in absolute path`() {
+        val url = Url("https://assets.example.org/library/annual–review/cover-image.webp")
+
+        assertEquals("/library/annual%E2%80%93review/cover-image.webp", url.encodedPath)
+        assertEquals(listOf("", "library", "annual–review", "cover-image.webp"), url.rawSegments)
+        assertEquals(
+            "https://assets.example.org/library/annual%E2%80%93review/cover-image.webp",
+            url.toString()
+        )
+    }
+
+    @Test
+    fun `preserve percent-encoded characters in path`() {
+        val url = Url("https://assets.example.org/library/annual%E2%80%93review/cover-image.webp")
+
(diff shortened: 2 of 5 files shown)


ktor-client/ktor-client-cio/common/src/io/ktor/client/engine/cio/utils.kt
```diff
@@ -54,7 +54,13 @@ internal suspend fun writeHeaders(
 
     try {
         val normalizedUrl = if (url.rawSegments.isEmpty()) URLBuilder(url).apply { encodedPath = "/" }.build() else url
-        val urlString = if (overProxy) normalizedUrl.toString() else normalizedUrl.fullPath
+        val encodedPath = normalizedUrl.encodedPath.encodeURLPath(encodeEncoded = false)
```
bjhham: Perhaps this fix would be better in `URLParser`?  Encoding on `encodedPath` seems like the mistake was made earlier in the flow.

ktor-client/ktor-client-cio/common/test/CIORequestLineTest.kt
```diff
@@ -0,0 +1,64 @@
...
+import io.ktor.utils.io.*
+import kotlinx.coroutines.*
+import kotlin.test.*
+
+class CIORequestLineTest {
+    @Test
+    fun `encode unicode characters in request path`() = runTest {
+        assertEquals(
+            "GET /Jiu-Yan-cosplay-2B-%E2%80%93-NierAutomata-43_result.webp HTTP/1.1",
+            requestLine("http://example.com/Jiu-Yan-cosplay-2B-–-NierAutomata-43_result.webp")
```
bjhham: Please use more generic paths in the test data 🙈

ktor-client/ktor-client-cio/common/src/io/ktor/client/engine/cio/utils.kt
EarzuChan: thanks. i ll check into it, and use a more generic path in the test data

ktor-client/ktor-client-cio/common/test/CIORequestLineTest.kt
EarzuChan: Updated. The tests and PR description now use the generic `https://assets.example.org/library/annual–review/cover-image.webp` example.

ktor-client/ktor-client-cio/common/src/io/ktor/client/engine/cio/utils.kt
EarzuChan: Good point. I moved the encoding to `URLParser`, where string paths enter the URL model. Parsed URLs now expose a correctly encoded `encodedPath` to every consumer, while existing `%HH` sequences are preserved. The CIO-specific production workaround has been removed, and parser-level tests now cover absolute, relative, query/fragment, and file URL forms.

ktor-client/ktor-client-cio/common/test/CIORequestLineTest.kt
```diff
@@ -0,0 +1,48 @@
...
+import io.ktor.client.request.*
+import io.ktor.client.utils.*
+import io.ktor.http.*
+import io.ktor.test.*
+import io.ktor.util.*
+import io.ktor.utils.io.*
+import kotlinx.coroutines.*
+import kotlin.test.*
+
+class CIORequestLineTest {
```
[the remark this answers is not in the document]
EarzuChan: Updated `CIORequestLineTest` to use `internal` visibility.

ktor-http/common/test/io/ktor/tests/http/URLParserTest.kt
```diff
@@ -0,0 +1,72 @@
...
+ */
+
+package io.ktor.tests.http
+
+import io.ktor.http.*
+import kotlin.test.*
+
+internal class URLParserTest {
+    @Test
+    fun encodeUnicodeCharactersInAbsolutePath() {
```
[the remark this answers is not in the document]
EarzuChan: Updated all test methods in `URLParserTest` to use descriptive backtick-enclosed names.

bjhham (approved):

bjhham:
Looks like there's a test failure here in the non-jvm tests, though I've no idea how it could be related...

```
io.ktor.tests.server.application.ApplicationRequestHeaderTest#an_application_that_handles_requests_to_foo_with_parameters
```

EarzuChan:
> Looks like there's a test failure here in the non-jvm tests, though I've no idea how it could be related...
> 
> ```
> io.ktor.tests.server.application.ApplicationRequestHeaderTest#an_application_that_handles_requests_to_foo_with_parameters
> ```

@bjhham 

thanks for spotting this. I took another look at the odd Native CI failure you mentioned, and it turned out to be a small oversight in the parser logic rather than an unrelated failure. I've fixed it and pushed an updated commit. Could you please take another look when you have a chance? :D

bjhham (approved):

EarzuChan:
@bjhham It seems this PR can be merged, can you merge it for me 😊

PR #5746, opened 2026-07-13
Netty engine allocation improvements

bjhham:
**Subsystem**
Server, Netty

**Motivation**
Regular analysis of engine allocations has uncovered a few spots for improvements.

The cleanup of the response writer job seems so be the source of many of our flaky tests.  I've ran the tests a few times now and they pass completely each time. 💪 

**Solution**
- Re-use connection coroutine elements instead of re-allocating every time
- Avoid manual join for writer job in favour of structured concurrency

### DIFF
--- a/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/http3/NettyHttp3Handler.kt
+++ b/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/http3/NettyHttp3Handler.kt
@@ -25,14 +25,14 @@ internal class NettyHttp3Handler(
     private val userCoroutineContext: CoroutineContext,
     runningLimit: Int
 ) : Http3RequestStreamInboundHandler(), CoroutineScope {
-    private val handlerJob = SupervisorJob(userCoroutineContext[Job])
+    // Parent [Job] for per-call [Job]s. Cached to avoid re-running `userCoroutineContext[Job]` per request.
+    private val parentJob: Job? = userCoroutineContext[Job]
+
+    private val handlerJob = SupervisorJob(parentJob)
 
     // Connection-stable portion of the per-call coroutine context. Cached so each request only needs
     // to combine it with the per-stream dispatcher and the per-call [Job].
     private val staticCallContext: CoroutineContext = userCoroutineContext + CallHandlerCoroutineName
-
-    // Parent [Job] for per-call [Job]s. Cached to avoid re-running `userCoroutineContext[Job]` per request.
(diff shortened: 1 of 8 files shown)


ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/http3/NettyHttp3Handler.kt
```diff
@@ -27,6 +27,16 @@ internal class NettyHttp3Handler(
 ) : Http3RequestStreamInboundHandler(), CoroutineScope {
     private val handlerJob = SupervisorJob(userCoroutineContext[Job])
 
+    // Connection-stable portion of the per-call coroutine context. Cached so each request only needs
+    // to combine it with the per-stream dispatcher and the per-call [Job].
+    private val staticCallContext: CoroutineContext = userCoroutineContext + CallHandlerCoroutineName
+
+    // Parent [Job] for per-call [Job]s. Cached to avoid re-running `userCoroutineContext[Job]` per request.
+    private val parentJob: Job? = userCoroutineContext[Job]
```
osipxd: Minor. Let's put this declaration before `handlerJob`. This way it would be able to use `parentJob` when creating `handlerJob`

osipxd (approved):
Nice!

bjhham:
I think I'll re-target this one to release.  It should be quite beneficial for general stability around closing connections.

EDIT: Actually, I'll just cherry-pick over to release, because this effects HTTP/3 also which is only in main.

zibet27 (approved):
Great!

Curious to see some becnhmarks
I was looking on artifacts on the CI, but it's not very clear there)

bjhham:
Oh, yeah it's not a _huge_ difference for allocations, but it helps:

```
Request consumes 8.06 KB, expected 8.55 KB. 
  Difference: -495 < 1050 (allowed)
  Consumed 8.06 KB on request
  Expected 8.55 KB on request
  Saved    495 bytes on request

Increased locations:
  NettyHttpResponsePipeline.kt                    41    (809          -->          850)

Decreased locations:
  NettyHttp1Handler.kt                          -248    (968          -->          720)
  NettyApplicationCall.kt                       -152    (240          -->           88)
  SuspendFunctionGun.kt                         -136    (352          -->          216)
```

For reference, you can find these under the Allocation Test job, then `Build Log > Gradle Step 2/2 > benchmarks.ServerAllocationsTest > helloWorldFootprint`.  Not exactly the most user-friendly way to find it.  Maybe it'd be best to build a summary report for the artifacts instead of just the dump.

PR #5747, opened 2026-07-14
KTOR-9707 Do not cancel body erroneously

bjhham:
**Subsystem**
Client

**Motivation**
[KTOR-9707](https://youtrack.jetbrains.com/issue/KTOR-9707) SaveBody performance regression

**Solution**
This kind stranger ran some client benchmarks against the underlying engine implementations and found this issue:

https://github.com/ragboyjr/ktor-benchmarks/tree/dockerized-benchmarks/docker-benchmark#high-level-results

Calling cancel on the response body is creating a new `IOException` every time which was eating up a lot of resources.  We can instead just check if it's already closed / reuse the original cause if there was some I/O error.


osipxd (approved):
Nicely done 👍

PR #5748, opened 2026-07-14
KTOR-9690 Fix graceful shutdown regression in Netty

bjhham:
**Subsystem**
Server, Netty

**Motivation**
[KTOR-9690](https://youtrack.jetbrains.com/issue/KTOR-9690) Netty: in-flight requests are aborted on stop(), ignoring shutdown grace period

**Solution**
Wait for connection event group to shutdown before moving onto the other event groups.  This should prevent live request handlers from being needlessly aborted.


osipxd (approved):
👍

PR #5750, opened 2026-07-14
KTOR-4110: Fix WebSocket session structured concurrency

ark-1:
Make DefaultWebSocketSessionImpl treat the Job of the session it wraps as a parent for its own Job.
Also add a test for it.

**Subsystem**
Client, Server (WebSockets shared module)

**Motivation**
[KTOR-4110](https://youtrack.jetbrains.com/issue/KTOR-4110) Websockets: Websocket handler doesn't support structured concurrency

**Solution**
Make DefaultWebSocketSessionImpl treat the Job of the session it wraps as a parent for its own Job.

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-websockets/jvm/test/io/ktor/tests/websocket/WebSocketTest.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-websockets/jvm/test/io/ktor/tests/websocket/WebSocketTest.kt
@@ -24,6 +24,7 @@ import kotlinx.io.readByteArray
 import kotlin.random.Random
 import kotlin.test.*
 import kotlin.time.Duration.Companion.seconds
+import kotlin.use
 
 @CoroutinesTimeout(30_000)
 class WebSocketTest {
@@ -503,6 +504,34 @@ class WebSocketTest {
             val wsRawRoute: Route = webSocketRaw("/ws-raw") {}
             assertNotNull(wsRawRoute)
         }
+    }
+
+    @Test
+    fun `WebSocket session structured concurrency`() {
+        val messages = mutableListOf<String>()
+        testApplication {
+            install(WebSockets)
+            routing {
+                webSocket("/ws") {
+                    launch(CoroutineName("/ws handler")) {
+                        for (message in incoming) {
+                            // we need to block non-cooperatively while the whole application gets canceled
+                            // application shutdown should pause until this coroutine is terminated
+                            Thread.sleep(200)
+                            messages.add(message.data.toString(Charsets.UTF_8))
+                        }
+                    }
+                }
+            }
+            client.config {
+                install(io.ktor.client.plugins.websocket.WebSockets)
+            }.use {
+                it.webSocket("/ws") {
+                    outgoing.send(Frame.Text("Hello"))
+                }
+            }
+        }
+        assertContentEquals(expected = listOf("Hello"), actual = messages)
     }
 }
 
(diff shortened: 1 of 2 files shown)


ktor-server/ktor-server-plugins/ktor-server-websockets/jvm/test/io/ktor/tests/websocket/WebSocketTest.kt
```diff
@@ -504,6 +505,34 @@ class WebSocketTest {
             assertNotNull(wsRawRoute)
         }
     }
+
+    @Test
+    fun `WebSocket session structured concurrency`() {
```
zibet27, line 510: `WebSocket session structured concurrency` feels too general

ktor-shared/ktor-websockets/common/src/io/ktor/websocket/DefaultWebSocketSession.kt
```diff
@@ -146,9 +146,9 @@ internal class DefaultWebSocketSessionImpl(
     private val pinger = atomic<SendChannel<Frame.Pong>?>(null)
     private val closeReasonRef = CompletableDeferred<CloseReason>()
 
-    private val context = Job()
+    private val context = Job(raw.coroutineContext[Job])
```
zibet27, line 149: Good catch!

It was like that before PR #5274 where I changed it to make `WebSocketBackpressureTest.test incoming frame channel overflow` pass, which was wrong

From what I see atm, the problem failing `WebSocketBackpressureTest` is incorrect error propagation and test flakiness itself

ktor-server/ktor-server-plugins/ktor-server-websockets/jvm/test/io/ktor/tests/websocket/WebSocketTest.kt
```diff
@@ -504,6 +505,34 @@ class WebSocketTest {
...
+            }
+            client.config {
+                install(io.ktor.client.plugins.websocket.WebSockets)
+            }.use {
+                it.webSocket("/ws") {
+                    outgoing.send(Frame.Text("Hello"))
+                }
+            }
+        }
+        assertContentEquals(expected = listOf("Hello"), actual = messages)
```
zibet27, line 534: The assertion doesn't guarantee that the problem described in https://youtrack.jetbrains.com/issue/KTOR-4110/Websockets-Websocket-handler-doesnt-support-structured-concurrency#focus=Comments-27-6762844.0-0 is fixed, because it doesn't wait for a period to ensure no more messages are added

This test case is tied to the `incoming` channel which could be correctly closed and propagate cancellation, but the actual `CoroutineName("/ws handler")` won't get cancelled automatically
Though I'm not saying this is definitely a case here 
 
Also, it doesn't always pass on my machine :( 
Sometimes it fails with `Cancellation exception`, and sometimes with the message list being empty

zibet27 (changes requested):
The fix is good, but it doesn't fix whole "structured concurrency" thing

I believe we need some discussion on error/cancellation propagation and concurrency guarantees on the websocket session scope and it's channels, then properly document it
Also, our impl feels too confusing and error-prone for me
Though it's not the scope of PR 
@bjhham @Mr3zee wdyt

So I suggest naming this one `WebSocket session cancellation cancels child jobs` and create more issues

bjhham:
> So I suggest naming this one WebSocket session cancellation cancels child jobs and create more issues

We can also test that the session waits for child jobs to complete too right?  That's what the ticket references.  It looks like it should be possible to modify the test with a bit of locking to validate the behaviour and fix the flakiness.

PR #5754, opened 2026-07-15
KTOR-9585 Fix schema title trimming

bjhham:
**Subsystem**
Server, OpenAPI

**Motivation**
[KTOR-9585](https://youtrack.jetbrains.com/issue/KTOR-9585) OpenAPI: JsonSchema title is truncated when it contains a dot

**Solution**
Instead of trimming to the last dot, we trim up to the last lowercase section of the dot chain.  This retains the parent for nested classes and avoids truncating custom titles when camel-case.  This is a bit of a proxy fix for the issue described in the bug.


ktor-compiler-plugin/build.gradle.kts
```diff
@@ -131,6 +131,9 @@ tasks {
         group = "verification"
         configureCompilerPluginTest()
         systemProperty("testSamples.replaceSnapshots", "true")
+        testClassesDirs = sourceSets.test.get().output.classesDirs
+        classpath = sourceSets.test.get().runtimeClasspath
+        outputs.upToDateWhen { false }
```
bjhham, line 136: Not sure if it was the gradle update or the Kotlin update, but this update snapshots target stopped working.  Had to manually assign these to get it back in order.

ktor-server/ktor-server-plugins/ktor-server-routing-openapi/common/src/io/ktor/server/routing/openapi/DescribeRoute.kt
```diff
@@ -2,8 +2,6 @@
  * Copyright 2014-2025 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
  */
 
-@file:OptIn(InternalAPI::class)
-
```
bjhham, line 6: Noticed this was no longer required.

ktor-server/ktor-server-plugins/ktor-server-routing-openapi/common/src/io/ktor/server/routing/openapi/OpenApiRoutes.kt
```diff
@@ -26,7 +26,7 @@ public fun Sequence<Route>.mapToPathItemsAndSchema(): Pair<Map<String, PathItem>
     val pathItems = mapToPathItems(
         PopulateMediaTypeDefaults + CollectSchemaReferences { schema ->
             val title = schema.title ?: return@CollectSchemaReferences null
-            val unqualifiedTitle = title.substringAfterLast('.')
+            val unqualifiedTitle = title.replace(Regex("\\b[a-z]\\w+\\."), "")
```
bjhham: Actual fix here.

ark-1:
So, this is a heuristic, right? Starts with a lowercase letter => package (drop), with an uppercase letter => enclosing class (retain). Do we have no way of accurate tracking of what's what? If not, seems fine, but maybe document it a bit more explicitly.

bjhham:
Yep, it's just a heuristic for now.  I don't think reworking the internal model to account for different title sources would be worth it at this time.  I'll review the docs for OpenAPI generation.  I'm not sure we have anything on the title inference atm.

ark-1 (approved):

PR #5755, opened 2026-07-15
KTOR-9712 Netty HTTP/3: create Http3ServerConnectionHandler per QUIC connection

ShiinaSekiu:
Fixes [KTOR-9712](https://youtrack.jetbrains.com/issue/KTOR-9712).

## Problem

With `enableHttp3()`, the first QUIC connection to the server works, but every connection after it — concurrent or opened after the first one closed — never completes its handshake and times out. The HTTP/3 listener effectively serves a single QUIC connection per process lifetime (one browser session; the second visitor gets nothing):

```
[probe] conn1 (fresh listener): connected in 821ms
[probe] conn1: GET /i -> ok
[probe] conn2 (while conn1 still open): CONNECT TIMED OUT after 5s
[probe] conn3 (after conn1 closed):     CONNECT TIMED OUT after 5s
```

`NettyHttp3ChannelInitializer` passes a **single** `Http3ServerConnectionHandler` instance to `QuicServerCodecBuilder.handler(...)`, which installs it into the pipeline of every incoming `QuicChannel`. The handler owns the connection's control-stream and QPACK state and is explicitly non-sharable — [`Http3ConnectionHandler.isSharable()` always returns `false`](https://github.com/netty/netty/blob/4.2/codec-http3/src/main/java/io/netty/handler/codec/http3/Http3ConnectionHandler.java#L235-L241) — so the second connection's pipeline initialization fails and its handshake never proceeds.

## Fix

Create the handler per connection in a `ChannelInitializer<QuicChannel>`, matching the netty HTTP/3 server example. The shared `NettyHttp3RequestStreamInitializer` is a `ChannelInitializer` and stays shared.

After the fix, the same probe:

```
[probe] conn1 (fresh listener): connected in 16ms
[probe] conn2 (while conn1 still open): connected in 15ms
[probe] conn3 (after conn1 closed): connected in 26ms
(all three serve requests)
```

## Tests

Added `NettyHttp3MultipleConnectionsTest`: it opens a second connection while the first is still open, and another one after the first has closed, and asserts that both connect and serve requests. Fails on current main, passes with this change. The existing `NettyHttp3Test` suite still passes.

### DIFF
--- a/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/http3/NettyHttp3ChannelInitializer.kt
+++ b/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/http3/NettyHttp3ChannelInitializer.kt
@@ -10,6 +10,7 @@ import io.netty.channel.ChannelInitializer
 import io.netty.channel.socket.DatagramChannel
 import io.netty.handler.codec.http3.Http3
 import io.netty.handler.codec.http3.Http3ServerConnectionHandler
+import io.netty.handler.codec.quic.QuicChannel
 import io.netty.handler.codec.quic.QuicSslContext
 import java.util.concurrent.TimeUnit
 import kotlin.coroutines.CoroutineContext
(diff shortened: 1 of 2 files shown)


bjhham (changes requested):
Thanks for the contribution!

The fix and the test looks good, I'll just need you to run:

```
./gradlew formatKotlin
```

And push the resulting changes to get past the linter check.

bjhham (approved):

bjhham:
Thanks for the fix!

PR #5760, opened 2026-07-16
KTOR-9715 Netty HTTP/3: dispatch application calls via the call event group

ShiinaSekiu:
Fixes [KTOR-9715](https://youtrack.jetbrains.com/issue/KTOR-9715).

## Problem

Since KTOR-9542, HTTP/1 and HTTP/2 dispatch application handler code onto an executor pinned from `callEventGroup`, keeping user code off the Netty I/O threads. The HTTP/3 handler does neither: `startHttp3` builds the call context with the single-argument `NettyDispatcher.CurrentContext(context)` — which defaults to the channel's own event-loop executor — and launches the call coroutine via `context.executor()` as well.

This hurts more on HTTP/3 than it would elsewhere: each SSL connector binds a **single** `DatagramChannel`, and netty-codec-quic drives all QUIC connections and streams of that connector on that one channel's event loop. Blocking user code (JDBC, file IO, CPU-heavy serialization) therefore stalls not just its own stream but the whole HTTP/3 listener, including QUIC handshakes of brand-new, unrelated connections:

```
[repro] fresh connection + GET /instant with idle server: 45ms
[repro] fresh connection + GET /instant while another connection's handler blocks: 1525ms   // ≈ the full 1500 ms Thread.sleep
```

The same application code behaves fine over HTTP/1/2 on the same server, which makes this a confusing performance trap once HTTP/3 is enabled.

## Fix

Mirror the HTTP/1/2 model (KTOR-9542): pass `callEventGroup` down to `NettyHttp3Handler` (via `NettyHttp3ChannelInitializer` / `NettyHttp3RequestStreamInitializer`) and build the call context with `NettyDispatcher.CurrentContext(context, pinnedCallExecutor(context, callEventGroup))`, launching the call coroutine on that executor as `startHttp2` does.

After the fix, the same probe:

```
[repro] fresh connection + GET /instant with idle server: 77ms
[repro] fresh connection + GET /instant while another connection's handler blocks: 68ms   // no stall
```

## Tests

Added `NettyHttp3CallExecutorTest`: one connection's handler blocks in `Thread.sleep`, and the test asserts that a brand-new QUIC connection completes `GET /instant` in roughly baseline time. Fails on current main, passes with this change. The existing HTTP/3 suites (`NettyHttp3Test`, `NettyHttp3MultipleConnectionsTest`, `NettyHttp3ApplicationResponseTest`, `NettyHttp3LocalConnectionPointTest` — 34 tests total) still pass.

### DIFF
--- a/ktor-server/ktor-server-netty/jvm/test/io/ktor/tests/server/netty/NettyHttp3CallExecutorTest.kt
+++ b/ktor-server/ktor-server-netty/jvm/test/io/ktor/tests/server/netty/NettyHttp3CallExecutorTest.kt
@@ -1,0 +1,213 @@
+/*
+ * Copyright 2014-2026 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package io.ktor.tests.server.netty
+
+import io.ktor.server.application.*
+import io.ktor.server.netty.*
+import io.ktor.server.response.*
+import io.ktor.server.routing.*
+import io.ktor.server.test.base.*
+import io.ktor.utils.io.*
+import io.netty.bootstrap.Bootstrap
+import io.netty.channel.*
+import io.netty.channel.nio.NioEventLoopGroup
+import io.netty.channel.socket.nio.NioDatagramChannel
+import io.netty.handler.codec.http3.*
+import io.netty.handler.codec.quic.*
+import java.net.InetSocketAddress
+import java.util.concurrent.CountDownLatch
+import java.util.concurrent.LinkedBlockingQueue
+import java.util.concurrent.TimeUnit
+import kotlin.system.measureTimeMillis
+import kotlin.test.Test
+import kotlin.test.assertTrue
+
+/**
+ * Regression test: application handler code must not run on the QUIC event loop.
+ *
+ * Without dispatching calls to an executor pinned from the call event group (as HTTP/1 and
+ * HTTP/2 do since KTOR-9542), user handler code runs on the QUIC event loop. All QUIC
+ * connections of a connector share a single DatagramChannel driven by one event loop, so any
+ * blocking user code (JDBC, file IO) stalls the entire HTTP/3 listener, including QUIC
+ * handshakes of unrelated new connections.
+ */
+class NettyHttp3CallExecutorTest :
+    EngineTestBase<NettyApplicationEngine, NettyApplicationEngine.Configuration>(Netty) {
+
+    init {
+        enableSsl = true
+        // FreePorts probes with TCP sockets, but HTTP/3 binds UDP on sslPort; on Windows,
+        // TCP-free ports often fall into Hyper-V/WSL UDP excluded ranges and fail the bind.
+        // Pin ports outside the dynamic range for stability.
+        port = 24430
+        sslPort = 24431
+    }
+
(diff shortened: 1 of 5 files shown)


ktor-server/ktor-server-netty/jvm/test/io/ktor/tests/server/netty/NettyHttp3CallExecutorTest.kt
```diff
@@ -0,0 +1,213 @@
...
+            }
+        }
+
+        // warmup: the first QUIC connection pays one-time JVM/native initialization costs
+        withHttp3Client { quic -> sendHttp3Request(quic, "/instant") }
+
+        val baselineMs = measureTimeMillis {
+            withHttp3Client { quic -> sendHttp3Request(quic, "/instant") }
+        }
+        println("[repro] fresh connection + GET /instant with idle server: ${baselineMs}ms")
```
bjhham, line 75: Please remove any printlns from the test to avoid cluttering the build outputs.

ktor-server/ktor-server-netty/jvm/test/io/ktor/tests/server/netty/NettyHttp3CallExecutorTest.kt
```diff
@@ -0,0 +1,213 @@
...
+
+            pending.responseQueue.poll(5, TimeUnit.SECONDS) // drain the /block response
+        }
+
+        assertTrue(
+            stalledMs - baselineMs < 1000,
+            "an unrelated new HTTP/3 connection was stalled (${stalledMs}ms vs ${baselineMs}ms baseline) " +
+                "by blocking user code on another connection: user code runs on the shared QUIC event " +
+                "loop instead of an executor from the call event group (cf. HTTP/1/2, KTOR-9542)"
+        )
```
bjhham, line 102: Could you try to find a way to test this that doesn't rely on timing checks?  It's likely this would introduce some flakiness in CI due to random delays from parallel tests.

ktor-server/ktor-server-netty/jvm/test/io/ktor/tests/server/netty/NettyHttp3CallExecutorTest.kt
```diff
@@ -0,0 +1,213 @@
...
+class NettyHttp3CallExecutorTest :
+    EngineTestBase<NettyApplicationEngine, NettyApplicationEngine.Configuration>(Netty) {
+
+    init {
+        enableSsl = true
+        // FreePorts probes with TCP sockets, but HTTP/3 binds UDP on sslPort; on Windows,
+        // TCP-free ports often fall into Hyper-V/WSL UDP excluded ranges and fail the bind.
+        // Pin ports outside the dynamic range for stability.
+        port = 24430
+        sslPort = 24431
```
bjhham, line 45: Maybe we could introduce some more scalable logic here for binding on the UDP range instead of using a specific port.

bjhham (changes requested):
Thanks for the fix!  It must have arisen from the issues being fixes out of order.  I left a couple of comments regarding the test that could use some attention.

bjhham:
Made relevant changes in https://github.com/ktorio/ktor/pull/5786

PR #5762, opened 2026-07-17
KTOR-9688 RateLimit: No access to principal when wrapped with authentication since 3.5.1

zibet27:
**Subsystem**

**Motivation**
[KTOR-9688](https://youtrack.jetbrains.com/issue/KTOR-9688) RateLimit: No access to principal when wrapped with authentication since 3.5.1

**Solution**
Move Authentication, Rate Limiting, CSRF and RequestBodyLimit to a separate Guards phase so the execution order of these plugins is defined by DSL nesting

PS: PR adds some new API and has `main` as a base for now. At the same time, it's a regression and would be nice to release it in the next minor version, so I could make those declarations internal in the minor version and change them to public in a 3.6


ktor-server/ktor-server-core/common/src/io/ktor/server/application/ApplicationCallPipeline.kt
```diff
@@ -21,6 +21,7 @@ public open class ApplicationCallPipeline public constructor(
     Setup,
     Monitoring,
     Plugins,
+    Guards,
```
bjhham: Would "Validation" or "Checks" provide more clarity for this phase?  I don't think I've ever used "Guard" outside of the context of short-circuiting a function or a loop...

bjhham (changes requested):
> PS: PR adds some new API and has main as a base for now. At the same time, it's a regression and would be nice to release it in the next minor version, so I could make those declarations internal in the minor version and change them to public in a 3.6

Yeah, make it so! ☝️ 💪 

Could you target this to 3.5.2 with the internal visibility and log a feature ticket to expose the new phase?  It'll need some documentation for plugin devs too.

ktor-server/ktor-server-core/common/src/io/ktor/server/application/ApplicationCallPipeline.kt
zibet27: I guess "Validation" should also work here

zibet27:
[KTOR-9716](https://youtrack.jetbrains.com/issue/KTOR-9716) Make ApplicationCallPipeline.ApplicationPhase.Validators public in 3.6.0

bjhham (approved):

osipxd (approved):

PR #5765, opened 2026-07-20
KTOR-9677 Check if a WatchKey is shared before cancelling it when reloading

Stexxe:
https://youtrack.jetbrains.com/issue/KTOR-9677


bjhham (approved):
👍

PR #5766, opened 2026-07-20
KTOR-9679 Fix reading incomplete UTF-8 strings from the buffer

Stexxe:
https://youtrack.jetbrains.com/issue/KTOR-9679
The current implementation now uses the number of bytes as the limit instead of the number of characters. This could be a breaking change, although the current implementation does not work properly with multi-byte characters and the limit, for example:
```kt
val channel = ByteReadChannel("привет\n")
println(channel.readLineStrict(6)) // throws TooLongLineException
```

### DIFF
--- a/ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
+++ b/ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
@@ -760,6 +760,23 @@ private suspend fun ByteReadChannel.internalReadLineTo(
             transferString(1) // transfer the CR
         } else {
             // No new line separator
+            if (count in 1..4 && !isClosedForRead) {
+                val tempSource = readBuffer.peek()
+                val incompleteCp = try {
+                    tempSource.readCodePointValue()
+                    false
+                } catch (_: EOFException) {
+                    true
+                } finally {
+                    tempSource.close()
+                }
+
+                if (incompleteCp) {
+                    awaitContent((count + 1).toInt())
+                    continue
+                }
+            }
+
             transferString(count)
         }
 
--- a/ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
+++ b/ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
@@ -687,12 +687,12 @@ private suspend fun ByteReadChannel.internalReadLineTo(
     if (isClosedForRead) return -1
 
     var consumed = 0L
+    val outBuffer = Buffer()
 
-    fun transferString(count: Long) {
+    fun transferBytes(count: Long) {
         if (count > 0L) {
-            val string = readBuffer.readString(count)
-            out.append(string)
-            consumed += string.length
+            outBuffer.write(readBuffer, count)
+            consumed += count
         }
     }
 
@@ -735,49 +735,40 @@ private suspend fun ByteReadChannel.internalReadLineTo(
 
         // Sole CR in the buffer
         if (crIndex >= 0) {
-            transferString(count = crIndex)
+            transferBytes(crIndex)
             readBuffer.discard(1)
+            out.append(outBuffer.readString())
             return consumed
         }
 
         // Fast path. LF or CRLF in the buffer
         if (lfIndex == 0L) {
             readBuffer.discard(1)
+            out.append(outBuffer.readString())
             return consumed
         }
         if (lfIndex > 0) {
             val isCrlf = if (readBuffer.buffer[lfIndex - 1] == CR) 1L else 0L
-            transferString(count = lfIndex - isCrlf)
+            transferBytes(lfIndex - isCrlf)
             readBuffer.discard(1 + isCrlf)
+            out.append(outBuffer.readString())
             return consumed
         }
 
         val count = minOf(limitLeft, readBuffer.remaining)
         // Check for the corner case when the last byte in the buffer is CR, and LF is in the next buffer
         if (readBuffer.buffer[count - 1] == CR) {
-            transferString(count = count - 1)
-            if (readBuffer.discardCrlfOrCr()) return consumed
-            transferString(1) // transfer the CR
+            transferBytes(count - 1)
+
+            if (readBuffer.discardCrlfOrCr()) {
+                out.append(outBuffer.readString())
+                return consumed
+            } else {
+                transferBytes(1)
+            }
         } else {
             // No new line separator
-            if (count in 1..4 && !isClosedForRead) {
-                val tempSource = readBuffer.peek()
-                val incompleteCp = try {
-                    tempSource.readCodePointValue()
-                    false
-                } catch (_: EOFException) {
-                    true
-                } finally {
-                    tempSource.close()
-                }
-
-                if (incompleteCp) {
-                    awaitContent((count + 1).toInt())
-                    continue
-                }
-            }
-
-            transferString(count)
+            transferBytes(count)
         }
 
         if (consumed < limit && readBuffer.remaining == 0L && !awaitContent()) break
@@ -798,14 +789,20 @@ private suspend fun ByteReadChannel.internalReadLineTo(
         when (readBuffer.buffer[0]) {
             LF -> {
                 readBuffer.discard(1)
+                out.append(outBuffer.readString())
                 return consumed
             }
 
-            CR -> if (readBuffer.discardCrlfOrCr()) return consumed
+            CR -> if (readBuffer.discardCrlfOrCr()) {
+                out.append(outBuffer.readString())
+                return consumed
+            }
         }
 
         throwTooLongLineException(limit)
     }
+
+    out.append(outBuffer.readString())
 
     if (throwOnIncompleteLine) throwEndOfStreamException(consumed)
     return consumed
--- a/ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
+++ b/ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
@@ -687,21 +687,21 @@ private suspend fun ByteReadChannel.internalReadLineTo(
     if (isClosedForRead) return -1
 
     var consumed = 0L
-    var outBufferCreated = false
-    val outBuffer: Buffer by lazy {
-        outBufferCreated = true
-        Buffer()
-    }
+    var outBuffer: Buffer? = null
 
     fun transferBytes(count: Long) {
         if (count > 0L) {
+            if (outBuffer == null) {
+                outBuffer = Buffer()
+            }
+
             outBuffer.write(readBuffer, count)
             consumed += count
         }
     }
 
-    fun completeLine(count: Long) {
-        if (!outBufferCreated) {
+    fun transferLine(count: Long) {
+        if (outBuffer == null) {
             if (count > 0L) {
                 out.append(readBuffer.readString(count))
                 consumed += count
@@ -751,7 +751,7 @@ private suspend fun ByteReadChannel.internalReadLineTo(
 
         // Sole CR in the buffer
         if (crIndex >= 0) {
-            completeLine(crIndex)
+            transferLine(crIndex)
             readBuffer.discard(1)
 
             return consumed
@@ -759,13 +759,13 @@ private suspend fun ByteReadChannel.internalReadLineTo(
 
         // Fast path. LF or CRLF in the buffer
         if (lfIndex == 0L) {
-            completeLine(lfIndex)
+            transferLine(lfIndex)
             readBuffer.discard(1)
             return consumed
         }
         if (lfIndex > 0) {
             val isCrlf = if (readBuffer.buffer[lfIndex - 1] == CR) 1L else 0L
-            completeLine(lfIndex - isCrlf)
+            transferLine(lfIndex - isCrlf)
             readBuffer.discard(1 + isCrlf)
             return consumed
         }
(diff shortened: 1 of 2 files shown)


ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
```diff
@@ -760,6 +760,23 @@ private suspend fun ByteReadChannel.internalReadLineTo(
...
+            if (count in 1..4 && !isClosedForRead) {
+                val tempSource = readBuffer.peek()
+                val incompleteCp = try {
+                    tempSource.readCodePointValue()
+                    false
+                } catch (_: EOFException) {
+                    true
+                } finally {
+                    tempSource.close()
+                }
```
bjhham: I think you can avoid the exception handling here and just check the last byte to see if it's in the middle of a UTF-8 code point:

```kotlin
val incompleteCp = (readBuffer.buffer[count - 1].toInt() and 0x80) != 0
```

The whole thing could be written as:

```kotlin
// prevent partial character reads at boundary
if (count in 1..4 &&
    !isClosedForRead &&
    (readBuffer.buffer[count - 1].toInt() and 0x80) != 0
) {
    awaitContent((count + 1).toInt())
    continue
}
```

ktor-io/common/test/ReadLineTest.kt
```diff
@@ -326,4 +328,53 @@ class ReadLineTest {
         }
         assertTrue(exhausted(), "Not all lines were read")
     }
+
+    @Test
+    fun `code point boundary`() = runTest {
+        val channel = ByteChannel()
+        backgroundScope.launch {
+            channel.writeFully(byteArrayOf(0xE3.toByte(), 0x83.toByte()))
```
osipxd, line 343: Let's change this test so it covers the concern pointed by CodeRabbit:

```suggestion
            val prefix = "Hello ".encodeToByteArray()                                                                                                                                                                                                                                                                                
            channel.writeFully(prefix + byteArrayOf(0xE3.toByte(), 0x83.toByte()))
```

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
```diff
@@ -687,12 +687,12 @@ private suspend fun ByteReadChannel.internalReadLineTo(
     if (isClosedForRead) return -1
 
     var consumed = 0L
+    val outBuffer = Buffer()
```
bjhham: What was wrong with the previous implementation that just checked for UTF character interruptions?  Adding a new intermediate buffer will increase memory allocations...

Stexxe: The previous solution works only if the incomplete UTF-8 sequence is at the beginning of the buffer.

bjhham: If we're to transfer to a buffer then it'd be best to just return the buffer as the line content and ignore the appendable at this level and append where it's called from.

Something like...

```kotlin
@OptIn(InternalAPI::class, InternalIoApi::class)
private suspend fun ByteReadChannel.internalReadLine(
    limit: Long,
    lenientLineEnding: Boolean,
    throwOnIncompleteLine: Boolean,
): Buffer? {
    val readBuffer = readBuffer // Get readBuffer once per line
    if (readBuffer.exhausted()) awaitContent()
    if (isClosedForRead) return null

    var consumed = 0L
    val outBuffer = Buffer()

    fun transferBytes(count: Long) {
        if (count > 0L) {
            outBuffer.write(readBuffer, count)
            consumed += count
        }
    }

    fun Source.scanForSoleCr(lfIndex: Long, limitLeft: Long): Long {
        // Don't scan for CR in default mode
        if (!lenientLineEnding) return -1

        val endIndex = when (lfIndex) {
            // Subtract 1 from remaining to ignore the last byte,
            // when RC might be part of a CRLF sequence split into two buffers
            -1L -> minOf(limitLeft, remaining - 1)

            0L -> 0

            // Subtract 1 from lfIndex to ignore the case when CR is part of the CRLF sequence
            else -> lfIndex - 1
        }
        return indexOf(CR, endIndex = endIndex)
    }

    // Should be called only if buffer[0] = CR
    suspend fun Source.discardCrlfOrCr(): Boolean {
        if ((remaining >= 2 || awaitContent(min = 2)) && buffer[1] == LF) {
            discard(2)
            return true
        }

        if (lenientLineEnding) {
            discard(1)
            return true
        }

        return false
    }

    while (consumed < limit && !isClosedForRead) {
        val limitLeft = limit - consumed
        val lfIndex = readBuffer.indexOf(LF, endIndex = limitLeft)
        val crIndex = readBuffer.scanForSoleCr(lfIndex, limitLeft)

        // Sole CR in the buffer
        if (crIndex >= 0) {
            transferBytes(crIndex)
            readBuffer.discard(1)
            return outBuffer
        }

        // Fast path. LF or CRLF in the buffer
        if (lfIndex == 0L) {
            readBuffer.discard(1)
            return outBuffer
        }
        if (lfIndex > 0) {
            val isCrlf = if (readBuffer.buffer[lfIndex - 1] == CR) 1L else 0L
            transferBytes(lfIndex - isCrlf)
            readBuffer.discard(1 + isCrlf)
            return outBuffer
        }

        val count = minOf(limitLeft, readBuffer.remaining)
        // Check for the corner case when the last byte in the buffer is CR, and LF is in the next buffer
        if (readBuffer.buffer[count - 1] == CR) {
            transferBytes(count - 1)

            if (readBuffer.discardCrlfOrCr()) {
                return outBuffer
            } else {
                transferBytes(1)
            }
        } else {
            // No new line separator
            transferBytes(count)
        }

        if (consumed < limit && readBuffer.remaining == 0L && !awaitContent()) break
    }

    if (consumed == 0L && isClosedForRead) return null

    check(consumed <= limit) {
        "Consumed bytes exceed the limit: $consumed > $limit. It's an implementation bug, please report it."
    }
    if (consumed == limit) {
        // We can't read data anymore
        if (limit == Long.MAX_VALUE) throw TooLongLineException("Max line length exceeded")
        // There is no more data
        if (readBuffer.remaining == 0L && !awaitContent()) throwEndOfStreamException(consumed)

        // Corner case: line ending is right after the limit
        when (readBuffer.buffer[0]) {
            LF -> {
                readBuffer.discard(1)
                return outBuffer
            }

            CR -> if (readBuffer.discardCrlfOrCr()) {
                return outBuffer
            }
        }

        throwTooLongLineException(limit)
    }

    if (throwOnIncompleteLine) throwEndOfStreamException(consumed)
    return outBuffer
}
```

@osipxd you added line readings to the I/O benchmarks right?  Maybe we can check it there?

Stexxe: Yes. Or we can have a `Buffer` as a parameter to which the bytes will be appended.

bjhham: Also an option 👍

osipxd: > [Osip Fatkullin (@osipxd)](https://github.com/osipxd), you added line readings to the I/O benchmarks right?

Yes, it's [here](https://github.com/ktorio/ktor-benchmarks/blob/main/io-benchmarks/src/jmh/kotlin/io/ktor/benchmarks/ReadLineBenchmark.kt)

Stexxe: @bjhham, the question is where will that out buffer be stored between the `readLine` calls?

bjhham: It's getting fully consumed when converted to the string so it shouldn't create much overhead for garbage collection.  If we want to optimize, we could just use a buffer directly instead of using an `Appendable` / `StringBuilder` in some cases.

osipxd: I agree that making `Buffer` a parameter instead of `Appendable` here would make the function cleaner.
The logic looks good. I'll run the benchmark and update my review after that.

osipxd (changes requested):
### `readLineStrictTo` profiling: 3.5.1 vs 3.5.2

For `200,000` lines of `10` characters, average time increased from
`8.839 ± 0.033 ms/op` to `12.629 ± 0.087 ms/op` (**+42.9%**).

#### CPU profile

| CPU leaf/frame family | 3.5.1 | 3.5.2 | Change |
|---|---:|---:|---:|
| `kotlinx.io.Buffer.write` | 0.03% | 5.98% | +5.95 pp |
| `kotlinx.io.Segment.split` | not sampled | 2.80% | +2.80 pp |
| `kotlinx.io.Buffer.recycleHead` | not sampled | 1.61% | +1.61 pp |
| `Arrays.copyInto` | 0.03% | 1.50% | +1.47 pp |
| `kotlinx.io.Segment.pop` | not sampled | 1.33% | +1.33 pp |
| `SegmentPool.*` | 0.03% | 1.61% | +1.58 pp |
| Segment-pool `VarHandle`/`AtomicReferenceArray` operations | not significant | 6.14% | ~+6.14 pp |
| **Selected buffer/segment overhead** | **~0.1%** | **~21.0%** | **~+20.9 pp** |

#### Allocation profile

| Allocated class | 3.5.1 share | 3.5.2 share | Change |
|---|---:|---:|---:|
| `kotlinx.io.Buffer` | not significant | 12.76% | +12.76 pp |
| `internalReadLineTo` coroutine state | 44.68% | 38.79% | -5.89 pp |
| `char[]` | 18.49% | 16.00% | -2.49 pp |
| `byte[]` | 14.73% | 13.04% | -1.69 pp |
| `Long` | 11.09% | 9.77% | -1.32 pp |
| `String` | 11.00% | 9.64% | -1.36 pp |   

---

I think we could avoid creating the intermediate buffer for the main case when the line fits one buffer. Something like this:

```kotlin
var consumed = 0L
var outBuffer: Buffer? = null

fun transferBytes(count: Long) {
    if (count <= 0L) return
    
    val buffer = outBuffer ?: Buffer().also { outBuffer = it }
    buffer.write(readBuffer, count)
    consumed += count
}

fun completeLine(count: Long): Long {
    val buffer = outBuffer
    if (buffer == null) {
        // Fast path: the entire line is already available.
        if (count > 0L) {
            out.append(readBuffer.readString(count))
            consumed += count
        }
    } else {
        // Slow path: bytes were accumulated across channel boundaries.
        transferBytes(count)
        out.append(buffer.readString())
    }

    return consumed
}
```

And then call `completeLine` in all places where currently `transferBytes` + `return` used

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
```diff
@@ -781,15 +803,21 @@ private suspend fun ByteReadChannel.internalReadLineTo(
         when (readBuffer.buffer[0]) {
             LF -> {
                 readBuffer.discard(1)
+                out.append(outBuffer.readString())
```
osipxd: `out.append(outBuffer.readString())` here and later can be replaced with `completeLine(0)`.
That's why I thought it'd be useful to return the `consumed` value from `completeLine`, because completeLine is meant to be a terminate function and we always want to return `consumed` bytes count. So we could just write `return completeLine(0)`

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
```diff
@@ -687,12 +687,28 @@ private suspend fun ByteReadChannel.internalReadLineTo(
     if (isClosedForRead) return -1
 
     var consumed = 0L
+    var outBufferCreated = false
+    val outBuffer: Buffer by lazy {
```
osipxd: Unfortunately `by lazy` also adds unnecessary allocations to hot path by creating an instance of `kotlin.Lazy`. A nullable value would be better here.
If we keep all interactions with the `outBuffer` inside `transferBytes` and `completeLine`, it shouldn't hard to handle nullability.

ktor-io/common/src/io/ktor/utils/io/ByteReadChannelOperations.kt
```diff
@@ -735,32 +751,38 @@ private suspend fun ByteReadChannel.internalReadLineTo(
...
             return consumed
         }
 
         val count = minOf(limitLeft, readBuffer.remaining)
         // Check for the corner case when the last byte in the buffer is CR, and LF is in the next buffer
         if (readBuffer.buffer[count - 1] == CR) {
-            transferString(count = count - 1)
-            if (readBuffer.discardCrlfOrCr()) return consumed
-            transferString(1) // transfer the CR
+            completeLine(count - 1)
```
osipxd: Probably it's better to rename this function to `transferLine` if it's used in this context. Or just use `transferByte` here and `transferLine` before `return`

osipxd (changes requested):

bjhham (approved):
The function is getting a little hairy, but the readLine benchmark looks good:

```
# 3.5.1

Benchmark                   (lineEnding)         (params)  Mode  Cnt    Score    Error  Units
ReadLineBenchmark.readLine       Default     200_000 x 10  avgt   10    9,769 ±  0,385  ms/op
ReadLineBenchmark.readLine       Default  1_000 x 1048576  avgt   10  596,025 ± 24,686  ms/op
ReadLineBenchmark.readLine       Lenient     200_000 x 10  avgt   10   12,704 ±  0,562  ms/op
ReadLineBenchmark.readLine       Lenient  1_000 x 1048576  avgt   10  790,964 ± 37,509  ms/op

# 3.5.2

Benchmark                   (lineEnding)         (params)  Mode  Cnt    Score    Error  Units
ReadLineBenchmark.readLine       Default     200_000 x 10  avgt   10    9,806 ±  0,517  ms/op
ReadLineBenchmark.readLine       Default  1_000 x 1048576  avgt   10  595,537 ± 21,909  ms/op
ReadLineBenchmark.readLine       Lenient     200_000 x 10  avgt   10   12,615 ±  0,280  ms/op
ReadLineBenchmark.readLine       Lenient  1_000 x 1048576  avgt   10  799,284 ± 42,752  ms/op
```

osipxd (approved):

PR #5768, opened 2026-07-21
Fix KTOR-9722  cannot be initialized with a congested  pool

zibet27:
**Subsystem**
Client

**Motivation**
[KTOR-9722](https://youtrack.jetbrains.com/issue/KTOR-9722) `DigestAuthProvider` cannot be initialized with a congested `Dispatchers.Default` pool

**Solution**
- generate a nonce in `DigestAuthProvider` lazily using suspend version
- generate nonce sequentially in `generateNonceBlocking` if there are no already generated 

The second change is questionable. Depending on what we expect from "Blocking" functions. 
`generateNonceBlocking` was still using coroutines and could cause a deadlock.
I tried profiling `NonceSmokeTest`, and it shows a slight increase in memory usage (~7.2 MB -> 8 MB).

### DIFF
--- a/ktor-utils/jvm/src/io/ktor/util/Nonce.kt
+++ b/ktor-utils/jvm/src/io/ktor/util/Nonce.kt
@@ -8,6 +8,7 @@ import kotlinx.coroutines.*
 import kotlinx.coroutines.channels.*
 import org.slf4j.*
 import java.security.*
+import java.util.concurrent.Executors
 import kotlin.random.*
 
 private val logger = LoggerFactory.getLogger("io.ktor.util.random")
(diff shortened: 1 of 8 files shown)


ktor-utils/jvm/src/io/ktor/util/Nonce.kt
bjhham, line 48: I think it'd be best to put this job on `Dispatchers.IO` as well.  This secure random could block threads too.

bjhham (changes requested):
Just one little comment.

Also, I was wondering if it would make sense here to replace the `generateBlocking` call in `StatelessHmacNonceManager`?  That could maybe go into main since it'd require an API change.

ktor-utils/jvm/src/io/ktor/util/Nonce.kt
zibet27: I was wondering why it wasn't on `Dispatchers.IO` and found this one https://youtrack.jetbrains.com/issue/KTOR-8829

But, they didn't add a regression test in the fix

zibet27:
> That could maybe go into main since it'd require an API change.

I was also looking at `StatelessHmacNonceManager`, and actually, we can change the default behaviour without an API change

zibet27:
> we can change the default behaviour without an API change

Anyway, it would be nice to pass suspend `public val nonceGenerator: suspend () -> String = { ... }`

ktor-utils/jvm/src/io/ktor/util/Nonce.kt
bjhham: Oh, interesting.  Maybe we should just create a standalone dispatcher with a single thread executor so it's not on either?

zibet27: I think it would better fit here, as afaiu `Dispatchers` are intended for structured concurrency, which doesn't work well with long running background processes

bjhham (approved):

Mr3zee:
@zibet27 please merge if all is ok

PR #5770, opened 2026-07-22
Clean up a whole bunch of small things

solonovamax:
**Subsystem**
Yes.

**Motivation**
I kept seeing a bunch of tiny things that needed to be cleaned up but weren't really large enough to warrant their own dedicated PR, so I'm just making this one PR which does a bunch of them at once.
There should be no behavioural changes.

**Solution**
I fixed the things that needed to be cleaned up.

Each of the commits are rather small and targeted (although many change a large number of lines, but it's just the same thing over and over). If necessary, I can rebase and drop commits. (it might actually be best to drop bccce7b71f806641380ed61a1750ef9e7c725f2e, as an iterator is allocated where previously it wasn't. possibly using `.indices` instead, which does not allocate an iterator)


ktor-server/ktor-server-plugins/ktor-server-sessions/jvm/src/io/ktor/server/sessions/SessionSerializerReflection.kt
```diff
@@ -87,7 +87,6 @@ internal class SessionSerializerReflection<T : Any>(
                 { it },
                 {
                     when (it.kind) {
-                        @OptIn(ExperimentalContextParameters::class)
```
bjhham, line 90: Looks like removing this caused the build to fail.

solonovamax: yeah, I noticed that. I was just using IntelliJ to detect any unnecessary `@OptIn` annotations. it is probably just bugged and didn't detect it in this specific case.

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/websocket/WebSocketContent.kt
```diff
@@ -9,17 +9,13 @@ import io.ktor.http.*
...
 private const val WEBSOCKET_VERSION = "13"
 private const val NONCE_SIZE = 16
 
 @OptIn(InternalAPI::class)
 internal class WebSocketContent : ClientUpgradeContent() {
-    private val nonce: String = buildString {
-        val nonce = generateNonce(NONCE_SIZE)
-        append(Base64.encode(nonce))
-    }
+    private val nonce: String = generateNonceBlocking(NONCE_SIZE)
```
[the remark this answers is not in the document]
solonovamax: this is not a real issue.

solonovamax:
I am genuinely unsure as to what could have caused those build errors.

The errors seem to be happening because the [`RouteField.Tag`](https://github.com/ktorio/ktor/blob/ae3f397f747b542e97cc57c54692cbe9c983234a/ktor-compiler-plugin/src/io/ktor/openapi/routing/RouteField.kt#L69) is not serializing properly..?
the only changes I have made in any files under the `ktor-compiler-plugin` paths where removing imports. I have gone and triple-checked, and they were most definitely unused.
I'm at a loss for what the cause of this was.
I'm going to try bisecting it to figure out what the cause is.

solonovamax:
okay, it seems it's definitely not caused by any of my changes. the same failures happen in the commit I based my branch off of (8694912d55fe9729432f44ac20c15d4b80175f4f). looking at the commits, I think it would have been fixed in b11731d227c8b07f1bdd1685fdeb853dd5d5b0de.

bjhham (approved):
Thanks for the tidying!

Please take a look at the conflicts, and we'll merge it before the release next week.

solonovamax:
> Please take a look at the conflicts, and we'll merge it before the release next week.

sure, I'll try to get around to it by the end of the weekend.

bjhham:
Resolving conflicts in another branch for merging

PR #5774, opened 2026-07-24
Fix KTOR-9729: Expose Response Body for Failed Websocket Connections

rameshvarun:
**Subsystem**
Websocket Client

**Motivation**
This fixes [KTOR-9729](https://youtrack.jetbrains.com/issue/KTOR-9729/Expose-HTTP-status-response-body-on-failed-WebSocket-handshake), which I reported. The gist is that there's no way to read the response body from a rejected WebSocket connection. It's typical for servers to send information about why the request was rejected in the body.

**Solution**
- Added a nullable `response: HttpResponse?` property to `WebSocketException`, mirroring `SSEClientException`.
- On a failed handshake, the response-pipeline interceptor now `save()`s the call and attaches it to the exception.
  - The saved call is tagged with an internal attribute so that reading `exception.response.bodyAsText()` doesn't re-enter the interceptor and throw again.
- This solution works on CIO, with the status code and body being preserved.
  -  For OkHTTP and Java engines, it only works for 401 responses, and only returns the status code / headers with the body being lost.

### DIFF
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/websocket/WebSockets.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/websocket/WebSockets.kt
@@ -293,6 +293,10 @@ public class WebSockets internal constructor(
 public class WebSocketException(
     message: String,
     cause: Throwable?,
+    /**
+     * The HTTP response from a failed handshake, exposing the status, headers, and body returned
+     * by the server. Availability varies by engine.
+     */
     public val response: HttpResponse?,
 ) : IllegalStateException(message, cause) {
     // required for backwards binary compatibility
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/websocket/WebSockets.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/websocket/WebSockets.kt
@@ -229,9 +229,17 @@ public class WebSockets internal constructor(
                     return@intercept
                 }
                 if (status != HttpStatusCode.SwitchingProtocols) {
+                    val failedResponse = try {
+                        context.save().response
+                    } catch (cause: Exception) {
+                        LOGGER.trace { "Failed to read response body of failed WebSocket handshake: $cause" }
+                        null
+                    }
                     @Suppress("ktlint:standard:max-line-length")
                     throw WebSocketException(
-                        "Handshake exception, expected status code ${HttpStatusCode.SwitchingProtocols.value} but was ${status.value}"
+                        "Handshake exception, expected status code ${HttpStatusCode.SwitchingProtocols.value} but was ${status.value}",
+                        cause = null,
+                        response = failedResponse,
                     )
                 }
                 if (session !is WebSocketSession) {
@@ -273,7 +281,12 @@ public class WebSockets internal constructor(
     }
 }
 
-public class WebSocketException(message: String, cause: Throwable?) : IllegalStateException(message, cause) {
+public class WebSocketException(
+    message: String,
+    cause: Throwable?,
+    public val response: HttpResponse?,
+) : IllegalStateException(message, cause) {
     // required for backwards binary compatibility
-    public constructor(message: String) : this(message, cause = null)
+    public constructor(message: String) : this(message, cause = null, response = null)
+    public constructor(message: String, cause: Throwable?) : this(message, cause, response = null)
 }
--- a/ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/websocket/WebSockets.kt
+++ b/ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/websocket/WebSockets.kt
@@ -24,6 +24,10 @@ public val WEBSOCKETS_KEY: AttributeKey<WebSockets> = AttributeKey<WebSockets>("
 
 internal val LOGGER = KtorSimpleLogger("io.ktor.client.plugins.websocket.WebSockets")
 
+// Marks a call whose failed-handshake response has already been captured, so reading that
+// response's body (via WebSocketException.response) doesn't re-enter the handshake handling below.
+private val FAILED_HANDSHAKE_RESPONSE_KEY = AttributeKey<Unit>("WebSocketFailedHandshakeResponse")
+
 /**
  * Indicates if a client engine supports WebSockets.
  *
@@ -228,9 +232,14 @@ public class WebSockets internal constructor(
                     LOGGER.trace { "Skipping non-websocket response from ${context.request.url}: $requestContent" }
                     return@intercept
                 }
+                if (context.attributes.contains(FAILED_HANDSHAKE_RESPONSE_KEY)) {
+                    // Reading the body of an already-captured failed handshake (WebSocketException.response);
+                    // let the default transformers produce the body instead of handling the handshake again.
+                    return@intercept
+                }
                 if (status != HttpStatusCode.SwitchingProtocols) {
                     val failedResponse = try {
-                        context.save().response
+                        context.save().also { it.attributes.put(FAILED_HANDSHAKE_RESPONSE_KEY, Unit) }.response
                     } catch (cause: Exception) {
                         LOGGER.trace { "Failed to read response body of failed WebSocket handshake: $cause" }
                         null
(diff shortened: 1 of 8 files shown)


ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/websocket/WebSockets.kt
```diff
@@ -228,10 +232,23 @@ public class WebSockets internal constructor(
...
                     return@intercept
                 }
+                if (context.attributes.contains(FAILED_HANDSHAKE_RESPONSE_KEY)) {
+                    // Reading the body of an already-captured failed handshake (WebSocketException.response);
+                    // let the default transformers produce the body instead of handling the handshake again.
+                    return@intercept
+                }
                 if (status != HttpStatusCode.SwitchingProtocols) {
+                    val failedResponse = try {
+                        context.save().also { it.attributes.put(FAILED_HANDSHAKE_RESPONSE_KEY, Unit) }.response
```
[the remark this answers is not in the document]
rameshvarun: There's precedent here since it looks like SSEClientException also calls `.save()`, thereby materializing the response.

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/websocket/WebSockets.kt
```diff
@@ -273,7 +290,12 @@ public class WebSockets internal constructor(
     }
 }
 
-public class WebSocketException(message: String, cause: Throwable?) : IllegalStateException(message, cause) {
+public class WebSocketException(
+    message: String,
+    cause: Throwable?,
+    public val response: HttpResponse?,
+) : IllegalStateException(message, cause) {
```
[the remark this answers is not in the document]
rameshvarun: KDoc added.

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/websocket/WebSockets.kt
```diff
@@ -228,10 +232,23 @@ public class WebSockets internal constructor(
...
+                    // let the default transformers produce the body instead of handling the handshake again.
+                    return@intercept
+                }
                 if (status != HttpStatusCode.SwitchingProtocols) {
+                    val failedResponse = try {
+                        context.save().also { it.attributes.put(FAILED_HANDSHAKE_RESPONSE_KEY, Unit) }.response
+                    } catch (cause: Exception) {
+                        LOGGER.trace { "Failed to read response body of failed WebSocket handshake: $cause" }
+                        null
+                    }
```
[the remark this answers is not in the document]
zibet27: Good call here

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/websocket/WebSockets.kt
```diff
@@ -228,10 +232,23 @@ public class WebSockets internal constructor(
...
                     @Suppress("ktlint:standard:max-line-length")
                     throw WebSocketException(
-                        "Handshake exception, expected status code ${HttpStatusCode.SwitchingProtocols.value} but was ${status.value}"
+                        "Handshake exception, expected status code ${HttpStatusCode.SwitchingProtocols.value} but was ${status.value}",
```
zibet27, line 249: Just a nitpick
We could split it here to omit suppression

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/WebSocketTest.kt
```diff
@@ -578,6 +578,47 @@ class WebSocketTest : ClientLoader(except(ENGINES_WITHOUT_WS)) {
         }
     }
 
+    @Test
+    fun testFailedHandshakeExposesResponse() = clientTests(only("CIO")) {
```
zibet27, line 582: Afaiu, it's possible to receive the body from the failed handshake on Java and OkHttp if handled properly

PS: Also, I found that this test case exposes our bug with an internal lateinit variable in the Java Client

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/websocket/WebSockets.kt
```diff
@@ -273,7 +290,16 @@ public class WebSockets internal constructor(
     }
 }
 
-public class WebSocketException(message: String, cause: Throwable?) : IllegalStateException(message, cause) {
+public class WebSocketException(
+    message: String,
+    cause: Throwable?,
+    /**
```
zibet27, line 296: Minor: I would put the field Kdoc on top of the class

zibet27 (changes requested):
Hi @rameshvarun! Thank you for the PR!
The code looks good!

Providing the response only on CIO feels too inconsistent for me, so I would try to construct a best-effort response on other platforms.
It can potentially expose implementation inconsistencies between our engines and require a bit more work here.

Would you like to cover more engines? Or we can finish it ourselves

@bjhham @ark-1 wdyt

Mr3zee:
Hey @rameshvarun ! Do you plan to continue the work on this PR?

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/websocket/WebSockets.kt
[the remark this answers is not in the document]
rameshvarun: Added catch + rethrow exception

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/websocket/WebSockets.kt
rameshvarun: Split error string in two and removed lint

ktor-client/ktor-client-core/common/src/io/ktor/client/plugins/websocket/WebSockets.kt
rameshvarun: Moved the field Kdoc to a class level string with an "@property" tag.

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/WebSocketTest.kt
rameshvarun: I'm iterating on some Claude proposals. The one thing I see is the response body always needs to be materialized, but I think that's inherent due to the Exception object outliving the WebSocket block.

rameshvarun:
@Mr3zee, @zibet27 - Added support for OkHttp and Java engines:
- All three engines require materializing the response body.
- The Java implementation relies on a JDK [implementation detail](https://github.com/openjdk/jdk/blob/71e04e3d85db379a718e440e587f31812090265a/src/java.net.http/share/classes/jdk/internal/net/http/websocket/OpeningHandshake.java#L185). But it's fail open so you just get no response body if that HttpResponse return type ever changes.

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/WebSocketTest.kt
rameshvarun: Added OkHTTP and Java engine support

ark-1:
Current state:
* Curl: throws WebSocketException without the body data (just headers + status).
* Darwin: throws WebSocketException on 401 with the status but without the headers or the body (just status), otherwise the engine throws another exception without the response data at all.
* WinHttp: the engine throws another exception without the response data at all.
* JS, WasmJs: throws WebSocketException without the response data at all.

We'd like to get the same consistent behavior for Curl, Darwin and WinHttp engines within this PR. For JS and WasmJs engines this behavior seems impossible to support. The test should say "except" instead of "only" then.

It makes sense to subclass WebSocketException with WebSocketHandshakeException, which will be thrown on all hanshake errors whether we have the response data or not. Then, the KDoc for this exception would document the behavior.

@rameshvarun, please tell if you want to continue working on this PR. There's also an option that you do part of the listed work and we'll finish up afterwards, if you want.

rameshvarun:
@ark-1 I have to resign from this PR as I don't have time currently to investigate and test the full engine suite. Feel free to close this PR. Appreciate the feedback and hope this feature makes it into Ktor.

ark-1:
No problem, we'll finish this up ourselves. Thank you for the contribution!

PR #5779, opened 2026-07-29
Make switch-base-branch.sh fork-aware

Mr3zee:
## Subsystem
Build/Tooling

## Motivation
`switch-base-branch.sh` assumed the branch being rebased lives in `ktorio/ktor` and that `origin` is the canonical repository. Running it on a PR branch that comes from a fork (either as the fork owner, or as a maintainer who checked the PR out with `gh pr checkout`) pushed the rewritten branch to the wrong repository, and could rebase it onto a fork's outdated copy of `main`/`release/*`.

## Solution
- **Push to the repository the branch belongs to.** The push target comes from `branch.<name>.pushRemote`, then `remote.pushDefault`, then `branch.<name>.remote`, and is repointed at the PR's head repository when they disagree — so it works both for fork owners (`origin` is the fork) and for maintainers reviewing a contributor's PR. Named remotes are preferred over URLs when one exists.
- **Always read base branches from the canonical repository.** `upstream` if configured, otherwise `origin`, or `origin`'s parent when `origin` turns out to be a fork. When the canonical repository has no remote configured, its base branches are fetched into private `refs/switch-base/*` refs instead of trusting the fork's copies.
- **Offer to switch the PR base on GitHub.** After a successful force-push, the script asks `y/n` and runs `gh pr edit <n> --repo <canonical> --base <target>`; it prints the command as a hint when the push was declined or `gh` is unavailable. Several open PRs with the same head branch are detected and left alone.
- **Stop early when a fork is involved and `gh` is not.** Forks are only visible through the GitHub CLI, so without it the script asks whether a fork is involved and fails with an explanation (plus the manual setup that makes it work) rather than silently rebasing onto the wrong bases and pushing to the wrong repository.
- **Safer force-push.** `--force-with-lease` now carries an explicit `<ref>:<sha>` lease read via `git ls-remote`, so the check also applies to remotes given as a URL, which have no remote-tracking refs.

`--dry-run` stays non-interactive: the new fork question is auto-answered `no` so the rest of the plan is still previewed.

## Testing
Verified with a unit harness (stubbed `git`/`gh`, 88 assertions) covering remote/slug resolution, base-source selection, push-target resolution, PR disambiguation, the PR base switch and the new fork guard, plus end-to-end runs against a throwaway fork pair and a mock PR: fork-owner run, maintainer `gh pr checkout` run, stale-lease rejection, declined push, `gh` unavailable, and runs with the fork's base branches deliberately made stale to prove the canonical tips are used.

### DIFF
--- a/switch-base-branch.sh
+++ b/switch-base-branch.sh
@@ -30,13 +30,28 @@ DRY_RUN=false
 MAIN_BRANCH="main"
 RELEASE_BRANCH=""
 CURRENT_BRANCH=""
-REMOTE=""
-PUSH_REMOTE="origin"
+BASE_REMOTE=""       # Remote name or URL of the canonical repository, never a fork
+BASE_LABEL=""        # Human-readable form of BASE_REMOTE
+BASE_REFS="refs/switch-base" # Base branches fetched when the canonical repo has no remote
+MAIN_REF=""          # Ref of 'main' in the canonical repository
+RELEASE_REF=""       # Ref of 'release/*' in the canonical repository
+TARGET_REF=""        # Ref the branch is rebased onto
+UPSTREAM_REPO=""     # 'owner/repo' of the canonical repository, used for gh calls
+PUSH_REMOTE="origin" # Remote name or URL the feature branch belongs to
+PUSH_LABEL=""        # Human-readable form of PUSH_REMOTE
+PUSH_ARGS=()         # 'git push' arguments, built once the remote state is known
+PUSHED=false
+REMOTE_HEAD=""       # Remote branch SHA before the rewrite, used for --force-with-lease
 CURRENT_BASE=""
 TARGET_BASE=""
 MERGE_BASE=""
 COMMITS_COUNT=0
 BACKUP_BRANCH=""
+GH_AVAILABLE=false
+PR_NUMBER=""
+PR_BASE=""
+PR_URL=""
+PR_HEAD_REPO=""
 
 # ============================================================================
 # Functions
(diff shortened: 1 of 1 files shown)


osipxd:
Correction to the PR description after the latest review changes:

The implementation has been simplified. The current description’s references to private `refs/switch-base/*` refs, PR disambiguation, push-remote precedence with disagreement correction, and stopping when `gh` is unavailable are no longer accurate.

The current behavior is:

- Base branches are always fetched from a configured remote whose URL points to `ktorio/ktor`. If none exists, the script offers to add `upstream`.
- The push destination comes from the PR head repository when `gh` can resolve an open PR. Otherwise, the script uses Git’s configured push destination for the current branch.
- A branch without a configured push destination remains local; the script does not implicitly fall back to `origin`.
- `gh` is optional. Git rebasing and pushing continue to work without it, while PR lookup and base retargeting are skipped or shown as manual actions.
- The remote feature branch is synchronized only when necessary and only after confirmation.
- Existing remote branches are protected with an explicit force-with-lease using the SHA returned by `git ls-remote`.
- Canonical base refs are fetched before the feature branch is modified, while the rebase plan is calculated after any required synchronization.
- A backup is created before synchronization when synchronization is required; otherwise, it is created only after the rebase plan is confirmed.
- Both rebases disable `rebase.updateRefs` so the backup remains usable.
- Local rebase completion is checked before push or PR updates proceed.
- The PR base is offered for retargeting only after the rewritten branch has been pushed successfully.

Validated with focused shell harnesses covering canonical and push-remote resolution, conditional synchronization, local-only branches, backup timing, explicit force-with-lease handling, rebase completion, and push/PR failure reporting, together with `bash -n`, `git diff --check`, real PR lookups, and real Git pull/push command forms.

osipxd (approved):

PR #5780, opened 2026-07-30
KTOR-9645 Introduce curated client engine defaults

bjhham:
**Subsystem**
Client

**Motivation**
[KTOR-9645](https://youtrack.jetbrains.com/issue/KTOR-9645) Client curated multi-platform facade module

**Solution**
Added a bundle artifact that includes the (subjectively) best engine for every platform.


ktor-client/ktor-client-defaults/common/src/io/ktor/client/defaults/DefaultEngine.kt
```diff
@@ -0,0 +1,36 @@
...
+ * | Linux                         | `Curl`   |
+ * | Windows (MinGW)               | `WinHttp`|
+ * | Darwin (iOS/macOS/tvOS/watchOS) | `Darwin` |
+ * | Android Native                | `CIO`    |
+ * | JS and Wasm/JS                | `Js`     |
+ *
+ * Prefer it over `HttpClient()` when the engine must be deterministic:
+ * ```kotlin
+ * val client = HttpClient(DefaultEngine)
+ * ```
```
nomisRev: `HttpClient() != HttpClient(DefaultEngine)` when the project has i.e. `implementation(ktorLibs.client.apache5)` configured for the module?

If so it would be nice to clarify here in the documentation.

nomisRev (approved):

nomisRev (changes requested):
Sorry, I had already approved but on second thought do we need `DefaultEngine` to be public at all? Or do we need it at all?

We expose the engines through `api` which allows downstream projects to pick it up through just `HttpClient()`. Do we have any ambition to deprecate `HttpClient` in favor of `HttpClient(DefaultEngine)`, and potentially remove it? Does using `HttpClient(DefaultEngine)` provide any meaningful (performance) benefits on certain platforms?

The reason I ask is because I think we should recommend using this for setting up KMP clients but it seems to avoid having to use `@ExperimentalKtorApi` we should still recommend `HttpClient()`.

ktor-client/ktor-client-defaults/common/src/io/ktor/client/defaults/DefaultEngine.kt
bjhham: In this case, the `HttpClient()` function would throw an exception because there's more than one engine available iirc.

bjhham:
@nomisRev the only benefit of having the `DefaultEngine` is that it avoids runtime errors when you have more than one engine available for a given platform.  This shouldn't be a problem if you're only using the "defaults" dependency, and people can still use the `HttpClient()` no-arg function.  I'm not sure if there'd be much need for it though, so not opposed to dropping it.

ktor-client/ktor-client-defaults/common/src/io/ktor/client/defaults/DefaultEngine.kt
osipxd: Unfortunately `HttpClient()` currently silently accepts first engine from the classpath. We should probably show  a warning at least

bjhham: Oh I see, that seems worse 😅

nomisRev: @osipxd do you know what dictates "first" in "silently accepts first engine from the classpath"? _normally_ it should be the user defined dependency rather than the transitive one but it's fragile and unreliable, right?

I guess with [Expect with defaults](https://github.com/Kotlin/KEEP/blob/main/proposals/KEEP-0466-expect-defaults.md) we have a nicer solution where _if a user wants to override_ they can do:

```kotlin
expect fun client(): HttpClient = HttpClient() { configure() }

fun HttpClientConfig<*>.configure() = TODO()

// jvm
actual fun client(): HttpClient = HttpClient(Apache5) { configure() }
```

I don't think it's currently possible without expect with defaults but instead of `DefaultEngine` I would be in favor of having a good way of overriding a single target's engine in a KMP project.

**Note:** It seems the only target this is useful for is JVM/Android since `Js`, `Darwin`, `WinHttp` , and perhaps even `Curl` don't make much sense to override (unless CIO evolves?).

bjhham: Looks like we ought to deprecate this one:

```
@InternalAPI
public inline fun <reified T : Any> loadServiceOrNull(): T? = loadServicesAsSequence<T>().firstOrNull()
```

And instead use `loadServices(): List<T>` and force the consumer to handle ambiguity.  We could even use some kind of ranking for "preferred" engines, so that if a user imports CIO in their common it doesn't randomly clobber other imports.  Ideally, we'd want specific target > common, but I doubt there's any magic that could work there.

nomisRev: > And instead use loadServices(): List<T> and force the consumer to handle ambiguity. 

Any idea what this might look like from the user side since this is currently entirely internal? 🤔 I cannot imagine anything that might work in common without having the target specific information available except something label based.

> We could even use some kind of ranking for "preferred" engines

This would ✅ this for me: "I would be in favor of having a good way of overriding a single target's engine in a KMP project.".

ktor-client/ktor-client-core/jvm/src/io/ktor/client/HttpClientJvm.kt
```diff
@@ -34,13 +36,43 @@ public actual fun HttpClient(
...
+            val selectedEngine = topEngines.first()
+            if (topEngines.size > 1) {
+                KtorSimpleLogger("HttpClient")
+                    .warn(
+                        buildString {
+                            append("Multiple engines found: ")
+                            appendLine(topEngines.joinToString { it.name })
+                            appendLine("\tUsing the first: ${selectedEngine.name}")
+                        }
+                    )
```
Mr3zee: INFO for which engine was selected and why if no warn

ktor-client/ktor-client-core/jvm/src/io/ktor/client/HttpClientJvm.kt
```diff
@@ -34,13 +36,53 @@ public actual fun HttpClient(
...
+        0 -> error(
+            "Failed to find HTTP client engine implementation: consider adding client engine dependency. " +
+                "See https://ktor.io/docs/http-client-engines.html"
+        )
+
+        1 -> engineContainers.single().factory
+
+        else -> {
+            val maxPriority = engineContainers.maxOf { it.priority }
+            val topEngines = engineContainers.filter { it.priority == maxPriority }
```
Mr3zee: Comparison of floating point doubles

ktor-client/ktor-client-core/jvm/src/io/ktor/client/HttpClientJvm.kt
```diff
@@ -13,8 +14,9 @@ import java.util.*
  * Constructs an asynchronous [HttpClient] using optional [block] for configuring this client.
  *
  * The [HttpClientEngine] is selected from the dependencies using [ServiceLoader].
- * The first found implementation that provides [HttpClientEngineContainer] service implementation is used.
- * An exception is thrown if no implementations found.
+ * All [HttpClientEngineContainer] implementations found on the classpath are ranked by [HttpClientEngineContainer.priority]
+ * and the one with the highest priority is used. An exception is thrown if no implementations are found
+ * or if multiple implementations share the same highest priority.
```
Mr3zee: Kdoc is incorrect about the exception for multiple highest

ktor-client/ktor-client-core/posix/src/io/ktor/client/HttpClient.kt
```diff
@@ -5,13 +5,17 @@
...
 import io.ktor.utils.io.*
 
 /**
  * Constructs an asynchronous [HttpClient] using optional [block] for configuring this client.
  *
- * The [HttpClientEngine] is selected from the dependencies.
- * https://ktor.io/docs/http-client-engines.html
+ * The [HttpClientEngine] is selected from the registered [engines].
+ * All registered engine factories are ranked by their priority and the one with the highest priority is used.
+ * An exception is thrown if no engines are registered or if multiple engines share the same highest priority.
```
Mr3zee: Same here

ktor-client/ktor-client-core/posix/src/io/ktor/client/HttpClient.kt
```diff
@@ -20,7 +24,44 @@ public actual fun HttpClient(
 ): HttpClient = HttpClient(FACTORY, block)
 
 @OptIn(InternalAPI::class)
-private val FACTORY = engines.firstOrNull() ?: error(
-    "Failed to find HTTP client engine implementation: consider adding client engine dependency. " +
-        "See https://ktor.io/docs/http-client-engines.html"
-)
+private val FACTORY = run {
```
Mr3zee, line 26: Seems like code duplication

Mr3zee (commented):
@bjhham I didn't find the DefaultEngine from the PR description

bjhham:
> [Bruce Hamilton (@bjhham)](https://github.com/bjhham) I didn't find the DefaultEngine from the PR description

@nomisRev mentioned that it'd be better to just integrate with our current approach and avoid this, so it has been removed.

Mr3zee (approved):

nomisRev (approved):

PR #5785, opened 2026-07-31
KTOR-9761 Cancel response lost to CIO writer timeout

osipxd:
**Subsystem**
Server, CIO

**Motivation**
Fix [KTOR-9761](https://youtrack.jetbrains.com/issue/KTOR-9761), where the CIO pipeline writer's timed receive can claim a response and then lose it to prompt cancellation before its continuation resumes. The response is no longer in the actor buffer, so a large response can suspend its request handler indefinitely without a reader.

**Solution**
Configure the response actor channel with `onUndeliveredElement` so a response claimed by a cancelled receive is cancelled explicitly. Add a deterministic test dispatcher that controls continuations and timeout callbacks independently and reproduces the receive-timeout interleaving.

Also use an explicit cancellation cause for buffered responses, allocating the fallback exception only when a pending response exists.

### DIFF
--- a/ktor-server/ktor-server-cio/common/src/io/ktor/server/cio/backend/ServerPipeline.kt
+++ b/ktor-server/ktor-server-cio/common/src/io/ktor/server/cio/backend/ServerPipeline.kt
@@ -41,7 +41,12 @@ public fun CoroutineScope.startServerConnectionPipeline(
     timeout: Duration,
     handler: HttpRequestHandler
 ): Job = launch(HttpPipelineCoroutine) {
-    val actorChannel = Channel<ByteReadChannel>(capacity = 3)
+    val actorChannel = Channel<ByteReadChannel>(
+        capacity = 3,
+        onUndeliveredElement = { response ->
+            response.cancel(IOException("Response was not delivered"))
+        }
+    )
     startServerPipelineWriter(actorChannel, timeout, connection)
 
     val requestContext = RequestHandlerCoroutine + Dispatchers.Unconfined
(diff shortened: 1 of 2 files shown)


ktor-server/ktor-server-cio/common/src/io/ktor/server/cio/backend/ServerPipeline.kt
```diff
@@ -202,13 +207,11 @@ internal fun CoroutineScope.startServerPipelineWriter(
...
+    var cancellationCause: Throwable? = cause
     while (true) {
         val response = tryReceive().getOrNull() ?: break
-        if (cause == null) {
-            response.cancel()
-        } else {
-            response.cancel(cause)
-        }
+        cancellationCause = cancellationCause ?: IOException("Response writer has stopped")
+        response.cancel(cancellationCause)
```
Mr3zee, line 214:
```kotlin
val cancellationCause: Throwable = cause ?: IOException("Response writer has stopped")
while (true) {
    val response = tryReceive().getOrNull() ?: break
    response.cancel(cancellationCause)
}
```

Mr3zee (approved):
lgtm, maybe one change

ktor-server/ktor-server-cio/common/src/io/ktor/server/cio/backend/ServerPipeline.kt
osipxd: I didn't want to allocate `IOException` if we don't have any responses to cancel.

Mr3zee: Oh, that case. Alright, fair

PR #5788, opened 2026-08-02
KTOR-9769 Introduce FormFieldLimitExceededException with headers and field name

RaySmith-ttc:
### Subsystem
Server, ktor-http-cio

### Motivation

Currently, when calling `call.receiveMultipart(formFieldLimit = ...)` and a form field exceeds the configured limit during stream parsing, Ktor throws a low-level, generic `kotlinx.io.IOException`.

#### Non-Informative Low-Level Error Messages:
When reading a multipart stream, `ByteChannelScanner` previously threw internal scanner errors like:
```text
IOException: Limit of 100 bytes exceeded while searching for "\r\n--boundary..."
```
This error message only reflects internal buffer scanning mechanics and provides zero context about which multipart part or form field caused the limit violation.

#### Practical Use-Cases:
1. **Targeted User Feedback**: Web applications processing multipart forms need to return informative API responses identifying the exact failing field (e.g., `"Form field limit exceeded for field 'photo'"`).
2. **Structured Observability & Logging**: When using Ktor's `StatusPages` plugin or application logging, developers require a strongly-typed exception exposing structured properties (`fieldName` and `headers`) rather than trying to parse uninformative low-level scanner strings.

### Solution
- **New Exception Class**: Introduced `FormFieldLimitExceededException` in `io.ktor.http.content` extending `kotlinx.io.IOException` for backward compatibility.
- **Structured Properties**:
    - `headers: Headers`: Exposes all HTTP headers of the part that exceeded the limit.
    - `fieldName: String?`: Helper property safely extracting the field name from the `Content-Disposition` header.
- **Informative Default Error Message**: Automatically formats the error message to specify the field name when present (`Form field limit exceeded for field '$name'`), falling back to full headers (`Form field limit exceeded for part with headers: $headers`) when unavailable.
- **CIO Parser Integration & Refactoring**:
    - Extracted `HttpHeadersMap.snapshot()` to `HttpHeadersMap.kt` as an internal extension, allowing it to be reused across both `CIOMultipartDataBase.kt` and `Multipart.kt`.
    - Updated M`ultipart.kt` to catch limit exceptions and throw `FormFieldLimitExceededException` populated with the part's Headers snapshot.

### DIFF
--- a/ktor-http/ktor-http-cio/common/src/io/ktor/http/cio/Multipart.kt
+++ b/ktor-http/ktor-http-cio/common/src/io/ktor/http/cio/Multipart.kt
@@ -6,6 +6,7 @@ package io.ktor.http.cio
 
 import io.ktor.http.*
 import io.ktor.http.cio.internals.*
+import io.ktor.http.content.*
 import io.ktor.utils.io.*
 import io.ktor.utils.io.core.*
 import kotlinx.coroutines.CompletableDeferred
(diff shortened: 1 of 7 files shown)


ktor-http/ktor-http-cio/common/src/io/ktor/http/cio/Multipart.kt
```diff
@@ -151,10 +152,15 @@ private suspend fun parsePartBodyImpl(
...
-        else -> throwLimitExceeded(contentLength, limit)
+    val byteCount = try {
+        when (val contentLength = headers["Content-Length"]?.parseDecLong()) {
+            null -> input.readUntil(boundaryPrefixed, output, limit, ignoreMissing = true)
+            in 0L..limit -> input.copyTo(output, contentLength) + input.skipIfFoundReadCount(boundaryPrefixed)
+            else -> throwLimitExceeded(contentLength, limit)
+        }
+    } catch (cause: IOException) {
+        if (cause is FormFieldLimitExceededException) throw cause
+        throw FormFieldLimitExceededException(headers.snapshot(), cause = cause)
```
RaySmith-ttc, line 163: Perhaps we should ensure that the exception originated specifically from `ByteChannelScanner`, e.g. `cause .message?.contains("exceeded while searching for")`

PR #5791, opened 2026-08-03
KTOR-9775 Close request body sink after duplex streaming request to release the connection for OkHttp

yakivy:
**Subsystem**
OkHttpClient

**Motivation**
When `duplexStreamingEnabled = true` in the OkHttp engine, the connection used by a duplex streaming request is never released back to the connection pool. The exchange never completes from OkHttp's perspective, so the connection never becomes idle, and the next request that reuses the pooled HTTP/2 connection hangs indefinitely.

**Solution**
Close the request body sink (`sink.use { ... }`) once the body has been fully written or the writer coroutine is cancelled. This  lets OkHttp complete the exchange and return the connection to the pool. `RequestBody.writeTo` has no defined semantics for closing the sink, but for duplex bodies OkHttp doesn't close it itself (in contrast to non-duplex requests), so closing it in `writeTo` seems the only mechanism available.

For reference `aws.smithy.kotlin.runtime.http.engine.okhttp.StreamingRequestBody` uses the same approach.

### DIFF
--- a/ktor-client/ktor-client-okhttp/jvm/src/io/ktor/client/engine/okhttp/StreamRequestBody.kt
+++ b/ktor-client/ktor-client-okhttp/jvm/src/io/ktor/client/engine/okhttp/StreamRequestBody.kt
@@ -33,7 +33,7 @@ internal class StreamRequestBody(
                     sink.use {
                         try {
                             channel.copyTo(it.outputStream().asByteWriteChannel())
-                        } catch (cause: Throwable) {
+                        } catch (cause: CancellationException) {
                             // A failing flush() completes the request body, while close() alone would
                             // throw ProtocolException for a partially written fixed-length body
                             // without releasing the connection.
--- a/ktor-client/ktor-client-okhttp/jvm/src/io/ktor/client/engine/okhttp/StreamRequestBody.kt
+++ b/ktor-client/ktor-client-okhttp/jvm/src/io/ktor/client/engine/okhttp/StreamRequestBody.kt
@@ -31,7 +31,15 @@ internal class StreamRequestBody(
                 try {
                     val channel = block()
                     sink.use {
-                        channel.copyTo(it.outputStream().asByteWriteChannel())
+                        try {
+                            channel.copyTo(it.outputStream().asByteWriteChannel())
+                        } catch (cause: Throwable) {
+                            // A failing flush() completes the request body, while close() alone would
+                            // throw ProtocolException for a partially written fixed-length body
+                            // without releasing the connection.
+                            runCatching { it.flush() }
+                            throw cause
+                        }
                     }
                 } catch (cause: IOException) {
                     throw cause
--- a/ktor-client/ktor-client-okhttp/jvm/src/io/ktor/client/engine/okhttp/StreamRequestBody.kt
+++ b/ktor-client/ktor-client-okhttp/jvm/src/io/ktor/client/engine/okhttp/StreamRequestBody.kt
@@ -30,7 +30,9 @@ internal class StreamRequestBody(
             CoroutineScope(callContext).launch(Dispatchers.IO) {
                 try {
                     val channel = block()
-                    channel.copyTo(sink.outputStream().asByteWriteChannel())
+                    sink.use {
+                        channel.copyTo(it.outputStream().asByteWriteChannel())
+                    }
                 } catch (cause: IOException) {
                     throw cause
                 } catch (cause: Throwable) {
--- a/ktor-client/ktor-client-okhttp/jvm/test/io/ktor/client/engine/okhttp/OkHttpHttp2Test.kt
+++ b/ktor-client/ktor-client-okhttp/jvm/test/io/ktor/client/engine/okhttp/OkHttpHttp2Test.kt
@@ -11,11 +11,17 @@ import io.ktor.client.tests.*
 import io.ktor.http.content.*
 import io.ktor.utils.io.*
 import kotlinx.coroutines.CompletableDeferred
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.withTimeoutOrNull
+import okhttp3.ConnectionPool
 import okhttp3.Protocol
 import kotlin.test.Test
 import kotlin.test.assertEquals
 import kotlin.test.assertFailsWith
+import kotlin.test.assertTrue
 import kotlin.test.fail
+import kotlin.time.Duration.Companion.milliseconds
+import kotlin.time.Duration.Companion.seconds
 
 class OkHttpHttp2Test : Http2Test<OkHttpConfig>(OkHttp) {
     override fun OkHttpConfig.enableHttp2() {
@@ -53,6 +59,45 @@ class OkHttpHttp2Test : Http2Test<OkHttpConfig>(OkHttp) {
                 """.trimIndent(),
                 response.trim()
             )
+        }
+    }
+
+    @Test
+    fun testDuplexStreamingConnectionReleasedAfterRequest() = testClient {
+        val connectionPool = ConnectionPool()
+        configureClient {
+            engine {
+                duplexStreamingEnabled = true
+                config { connectionPool(connectionPool) }
+            }
+        }
+
+        test { client ->
+            val inputChannel = ByteChannel(true)
+            client
+                .preparePost("/echo/stream") {
+                    setBody(inputChannel)
+                }
+                .execute {
+                    val outputChannel = it.bodyAsChannel()
+                    val buffer = StringBuilder()
+                    (0..2).forEach { i ->
+                        inputChannel.writeStringUtf8("client: $i\n")
+                        inputChannel.flush()
+                        outputChannel.readLineStrictTo(buffer)
+                        buffer.append('\n')
+                    }
+                    buffer.toString()
+                }
+
+            assertTrue(connectionPool.connectionCount() > 0, "Expected the request to use the injected pool")
+
+            val released = withTimeoutOrNull(5.seconds) {
+                while (connectionPool.idleConnectionCount() < connectionPool.connectionCount()) {
+                    delay(50.milliseconds)
+                }
+            } != null
+            assertTrue(released, "Connection should be released after request")
         }
     }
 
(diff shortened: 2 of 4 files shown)


ktor-client/ktor-client-okhttp/jvm/src/io/ktor/client/engine/okhttp/StreamRequestBody.kt
```diff
@@ -30,7 +30,9 @@ internal class StreamRequestBody(
             CoroutineScope(callContext).launch(Dispatchers.IO) {
                 try {
                     val channel = block()
-                    channel.copyTo(sink.outputStream().asByteWriteChannel())
+                    sink.use {
+                        channel.copyTo(it.outputStream().asByteWriteChannel())
+                    }
```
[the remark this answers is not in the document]
yakivy: I don't think this issue is valid in case of cancellation. Check non duplex logic that closes the body: `okhttp3.internal.http.CallServerInterceptor`, it doesn't handle coroutine cancellation case. But if you can come up with a test setup where connection release fails, I would be happy to update the logic.

ktor-client/ktor-client-okhttp/jvm/test/io/ktor/client/engine/okhttp/OkHttpHttp2Test.kt
```diff
@@ -56,6 +62,45 @@ class OkHttpHttp2Test : Http2Test<OkHttpConfig>(OkHttp) {
...
+                }
+
+            assertTrue(connectionPool.connectionCount() > 0, "Expected the request to use the injected pool")
+
+            val released = withTimeoutOrNull(5.seconds) {
+                while (connectionPool.idleConnectionCount() < connectionPool.connectionCount()) {
+                    delay(50.milliseconds)
+                }
+            } != null
+            assertTrue(released, "Connection should be released after request")
```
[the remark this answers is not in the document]
yakivy: I've refactored the asserts, should be more clear now.

ktor-client/ktor-client-okhttp/jvm/src/io/ktor/client/engine/okhttp/StreamRequestBody.kt
```diff
@@ -30,7 +30,9 @@ internal class StreamRequestBody(
             CoroutineScope(callContext).launch(Dispatchers.IO) {
                 try {
                     val channel = block()
-                    channel.copyTo(sink.outputStream().asByteWriteChannel())
+                    sink.use {
```
osipxd, line 33: This fixes the unknown-length case, but cancellation of a partially written duplex body with a declared content length still appears to retain the connection. I reproduced it by adapting the new connection-pool test to use an open `ReadChannelContent` with `contentLength = 1024`. After the response scope exits, `idleConnectionCount()` remains zero. OkHttp’s `RequestBodySink.close()` throws `ProtocolException("unexpected end of stream")` before marking the request body complete.

Could we cover this case as well and ensure cancellation releases fixed-length duplex bodies?

ktor-client/ktor-client-okhttp/jvm/test/io/ktor/client/engine/okhttp/OkHttpHttp2Test.kt
```diff
@@ -56,6 +61,46 @@ class OkHttpHttp2Test : Http2Test<OkHttpConfig>(OkHttp) {
...
+                .execute {
+                    val outputChannel = it.bodyAsChannel()
+                    val buffer = StringBuilder()
+                    (0..2).forEach { i ->
+                        inputChannel.writeStringUtf8("client: $i\n")
+                        inputChannel.flush()
+                        outputChannel.readLineStrictTo(buffer)
+                        buffer.append('\n')
+                    }
+                    buffer.toString()
```
osipxd: Minor: this can be simplified while also verifying the echoed response

```suggestion
                    repeat(3) { index ->
                        inputChannel.writeStringUtf8("client: $index\n")
                        inputChannel.flush()
                        assertEquals("server: client: $index", outputChannel.readLineStrict())
                    }
```

ktor-client/ktor-client-okhttp/jvm/test/io/ktor/client/engine/okhttp/OkHttpHttp2Test.kt
```diff
@@ -56,6 +61,46 @@ class OkHttpHttp2Test : Http2Test<OkHttpConfig>(OkHttp) {
...
+                }
+
+            assertEquals(connectionPool.connectionCount(), 1, "Expected the request to use the injected pool")
```
osipxd: Minor: the expected and actual arguments are reversed.
```suggestion
            assertEquals(1, connectionPool.connectionCount(), "Expected the request to use the injected pool")
```

ktor-client/ktor-client-okhttp/jvm/test/io/ktor/client/engine/okhttp/OkHttpHttp2Test.kt
```diff
@@ -56,6 +61,46 @@ class OkHttpHttp2Test : Http2Test<OkHttpConfig>(OkHttp) {
...
+
+            val idleConnections = withTimeoutOrNull(5.seconds) {
+                while (connectionPool.idleConnectionCount() < connectionPool.connectionCount()) {
+                    delay(50.milliseconds)
+                }
+                connectionPool.idleConnectionCount()
+            } ?: 0
+            assertEquals(idleConnections, 1, "Connection should be released after request")
```
osipxd: We have a utility function `waitForCondition` that could be used in this case:

```kotlin
waitForCondition("connection to be released", timeout = 5.seconds) {
    connectionPool.idleConnectionCount() == connectionPool.connectionCount()
}
assertEquals(1, connectionPool.idleConnectionCount(), "Connection should be released after request")

```

osipxd (changes requested):
@yakivy, Thank you for investigating and fixing the leak!

I’m requesting changes for one remaining request-body lifecycle case: cancellation after partially writing a duplex body with a declared content length can still leave the connection allocated.

I left a comment with the reproduction. The remaining test-cleanup comments are minor and non-blocking.

osipxd:
YouTrack issue: [KTOR-9775](https://youtrack.jetbrains.com/issue/KTOR-9775) OkHttp: duplex streaming does not release the connection after the request completes

ktor-client/ktor-client-okhttp/jvm/test/io/ktor/client/engine/okhttp/OkHttpHttp2Test.kt
yakivy: Fixed

ktor-client/ktor-client-okhttp/jvm/test/io/ktor/client/engine/okhttp/OkHttpHttp2Test.kt
yakivy: Fixed

ktor-client/ktor-client-okhttp/jvm/test/io/ktor/client/engine/okhttp/OkHttpHttp2Test.kt
yakivy: Fixed

ktor-client/ktor-client-okhttp/jvm/src/io/ktor/client/engine/okhttp/StreamRequestBody.kt
yakivy: IMO this issue should be solved in the OkHttp repo. There is no direct mechanism of closing the body from the Sink API. Smithy version doesn't address this problem too. I've found indirect way to close the body: `flush` closes the body on content length exception for example, but implementation becomes pretty obscure. Wdyt?

ktor-client/ktor-client-test-base/common/src/io/ktor/client/test/base/WaitUtils.kt
```diff
@@ -2,7 +2,7 @@
  * Copyright 2014-2025 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
  */
 
-package io.ktor.client.tests.utils
+package io.ktor.client.test.base
```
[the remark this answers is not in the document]
yakivy: @osipxd do you need me to do this or is it a part of CI/CD pipeline?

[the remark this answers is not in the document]
osipxd: The Rabbit is actually wrong here. Test modules are excluded from API dumps, so no action needed

ktor-client/ktor-client-okhttp/jvm/src/io/ktor/client/engine/okhttp/StreamRequestBody.kt
osipxd: I've created an issue in the OkHttp repo: https://github.com/lysine-dev/okhttp/issues/9648

Regarding the workaround, we probably could narrow this catch to `CancellationException`. The workaround relies on cancellation having already reset the HTTP/2 stream, so `flush()` fails and completes the request side. For unrelated producer failures, `flush()` may succeed and only transmit buffered data without completing the request. I verified that the fixed-length cancellation test still passes with the narrower catch.

settings.gradle.kts
```diff
@@ -130,7 +130,7 @@ projects {
         +"ktor-client-test-base"
         +"ktor-client-tests"
 
-        +"ktor-client-webrtc" including {
+        /*+"ktor-client-webrtc" including {
```
osipxd: Looks like an accidentally commited local change

yakivy: Fixed

ktor-client/ktor-client-okhttp/jvm/src/io/ktor/client/engine/okhttp/StreamRequestBody.kt
yakivy: Cool, thanks!
Narrowed down the catch scope.

osipxd (approved):
Nicely done, thank you 🏆 
I'll merge it this PR soon

PR #5797, opened 2026-08-05
KTOR-9819 Curl: Requests leak curl_slist and response StableRef

Stream29:
**Subsystem**

Client Engine/Curl

**Motivation**

I found that the Curl engine causes memory leak.
[KTOR-9819](https://youtrack.jetbrains.com/issue/KTOR-9819) Curl: Requests leak curl_slist and response StableRef

**Solution**

- Free each request's `curl_slist` correctly.
- Dispose the response `StableRef` correctly.
- Add test for this behavior.

### DIFF
--- a/ktor-client/ktor-client-curl/desktop/src/io/ktor/client/engine/curl/internal/CurlMultiApiHandler.kt
+++ b/ktor-client/ktor-client-curl/desktop/src/io/ktor/client/engine/curl/internal/CurlMultiApiHandler.kt
@@ -21,10 +21,14 @@ import platform.posix.size_tVar
 @OptIn(ExperimentalForeignApi::class)
 private class RequestHolder(
     val responseCompletable: CompletableDeferred<CurlSuccess>,
+    val requestHeaders: CPointer<curl_slist>,
+    val responseDataRef: StableRef<CurlResponseBuilder>,
     val requestWrapper: StableRef<CurlRequestBodyData>,
     val responseWrapper: StableRef<CurlResponseBodyData>,
 ) {
     fun dispose() {
+        curl_slist_free_all(requestHeaders)
+        responseDataRef.dispose()
         requestWrapper.dispose()
         responseWrapper.dispose()
     }
@@ -82,6 +86,8 @@ internal class CurlMultiApiHandler : Closeable {
         val requestWrapper = setupUploadContent(easyHandle, request)
         val requestHolder = RequestHolder(
             deferred,
+            request.headers,
+            responseDataRef.asStableRef(),
             requestWrapper.asStableRef(),
             responseWrapper.asStableRef()
         )
@@ -317,8 +323,6 @@ internal class CurlMultiApiHandler : Closeable {
         httpStatusCode: Long,
         proxyCode: CURLproxycode,
     ): CurlFail? {
-        curl_slist_free_all(request.headers)
-
         if (message != CURLMSG.CURLMSG_DONE) {
             return CurlFail(
                 IllegalStateException("Request $request failed: $message")
--- a/ktor-client/ktor-client-curl/desktop/src/io/ktor/client/engine/curl/internal/CurlMultiApiHandler.kt
+++ b/ktor-client/ktor-client-curl/desktop/src/io/ktor/client/engine/curl/internal/CurlMultiApiHandler.kt
@@ -127,7 +127,22 @@ internal class CurlMultiApiHandler : Closeable {
             request.caInfo?.let { option(CURLOPT_CAINFO, it) }
         }
 
-        curl_multi_add_handle(multiHandle, easyHandle).verify()
+        try {
+            curl_multi_add_handle(multiHandle, easyHandle).verify()
+        } catch (cause: Throwable) {
+            activeHandles.remove(easyHandle)
+            try {
+                try {
+                    responseData.responseBody.close(cause)
+                } finally {
+                    responseData.headersBytes.close()
+                }
+            } finally {
+                curl_easy_cleanup(easyHandle)
+                requestHolder.dispose()
+            }
+            throw cause
+        }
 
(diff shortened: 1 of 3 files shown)


ktor-client/ktor-client-curl/desktop/src/io/ktor/client/engine/curl/internal/CurlMultiApiHandler.kt
osipxd: Could we free request.headers if `curl_easy_init()` fails? The `curl_slist` has already been allocated, and no `RequestHolder` will take ownership of it:
```kotlin
val easyHandle = curl_easy_init() ?: run {                                                                                                   
    curl_slist_free_all(request.headers)                                                                                                     
    error("Could not initialize an easy handle")                                                                                             
}
```

ktor-client/ktor-client-curl/desktop/src/io/ktor/client/engine/curl/internal/CurlMultiApiHandler.kt
```diff
@@ -121,7 +127,22 @@ internal class CurlMultiApiHandler : Closeable {
             request.caInfo?.let { option(CURLOPT_CAINFO, it) }
         }
 
-        curl_multi_add_handle(multiHandle, easyHandle).verify()
+        try {
+            curl_multi_add_handle(multiHandle, easyHandle).verify()
+        } catch (cause: Throwable) {
+            activeHandles.remove(easyHandle)
```
osipxd: Could we add the holder to `activeHandles` only after `curl_multi_add_handle`, immediately before `return`? Then `activeHandles.remove(easyHandle)` becomes unnecessary.

ktor-client/ktor-client-curl/desktop/src/io/ktor/client/engine/curl/internal/CurlMultiApiHandler.kt
```diff
@@ -121,7 +127,22 @@ internal class CurlMultiApiHandler : Closeable {
             request.caInfo?.let { option(CURLOPT_CAINFO, it) }
         }
 
-        curl_multi_add_handle(multiHandle, easyHandle).verify()
+        try {
+            curl_multi_add_handle(multiHandle, easyHandle).verify()
+        } catch (cause: Throwable) {
```
osipxd: Could we also include we also include `setupMethod`, `setupUploadContent`, and all `option` calls in this try-catch since they use `verify` under the hood?
This requires moving `CurlRequestBodyData` construction out of `setupUploadContent`, so its `StableRef` is owned by RequestHolder before setup begins. The resulting shape could be:
```kotlin
// NativeUtils.kt
// Optionally. I've noticed that in some cases it's simpler to use StableRef, so it'd be nice to
// add a convenience extension for creating it
internal inline fun <T : Any> T.toStableRef(): StableRef<T> = StableRef.create(this)                                                                                                                                                                               
internal inline fun <T : Any> T.asStablePointer(): COpaquePointer = toStableRef().asCPointer()

// CurlMultiApiHandler.kt
val responseDataRef = responseData.toStableRef()
val responseWrapper = responseBody.toStableRef()
val requestWrapper = CurlRequestBodyData(
    body = request.content,
    callContext = request.callContext,
    onUnpause = { unpauseEasyHandle(easyHandle) },
).toStableRef()

val requestHolder = RequestHolder(...)

try {
    setupMethod(easyHandle, request.method, request.contentLength)
    setupUploadContent(easyHandle, requestWrapper.asCPointer()) // <-- Pass requestWrapper here

    // ... options ...

    curl_multi_add_handle(multiHandle, easyHandle).verify()
} catch (cause: Throwable) {
    // ...
}
```

osipxd (changes requested):
Hey @Stream29, thank you for finding an fixing this!
I found a few additional paths where resources may still leak. Please see the comments

osipxd:
YouTrack issue: [KTOR-9819](https://youtrack.jetbrains.com/issue/KTOR-9819) Curl: Requests leak curl_slist and response StableRef

osipxd:
We’re planning to release the next version next week, and it would be helpful to include this fix. If you don’t have time to address the review comments before then, please let me know and I can help finish the remaining changes.

ktor-client/ktor-client-curl/desktop/src/io/ktor/client/engine/curl/internal/CurlMultiApiHandler.kt
Stream29: Done in 227ba4bd5. `request.headers` is now freed when `curl_easy_init()` returns null.

ktor-client/ktor-client-curl/desktop/src/io/ktor/client/engine/curl/internal/CurlMultiApiHandler.kt
Stream29: Done in 227ba4bd5. The holder is now added to `activeHandles` only after `curl_multi_add_handle(...).verify()` succeeds.

ktor-client/ktor-client-curl/desktop/src/io/ktor/client/engine/curl/internal/CurlMultiApiHandler.kt
Stream29: Done in 227ba4bd5. `CurlRequestBodyData` is now wrapped in a `StableRef` owned by `RequestHolder` before setup begins, and the cleanup block covers `setupMethod`, `setupUploadContent`, all `option` calls, and `curl_multi_add_handle`. I also added the suggested `toStableRef` helper.

Stream29:
Done.

osipxd (approved):
Nicely done. Thanks!

PR #5798, opened 2026-08-05
KTOR-7009 Keep CIO pipeline connection open

osipxd:
**Subsystem**
CIO client engine

**Motivation**
[KTOR-7009](https://youtrack.jetbrains.com/issue/KTOR-7009) With HTTP pipelining enabled, CIO closes the connection output after writing a request and can fail to parse the response with an unexpected EOF.

**Solution**
Propagate the pipeline's `closeChannel` setting when writing the request body so the connection remains open for subsequent requests. Add a regression test that sends two requests through the same pipeline before receiving the first response.

### DIFF
--- a/ktor-client/ktor-client-cio/common/src/io/ktor/client/engine/cio/utils.kt
+++ b/ktor-client/ktor-client-cio/common/src/io/ktor/client/engine/cio/utils.kt
@@ -6,7 +6,7 @@ package io.ktor.client.engine.cio
 
 import io.ktor.client.call.*
 import io.ktor.client.engine.*
-import io.ktor.client.plugins.websocket.WEBSOCKETS_KEY
+import io.ktor.client.plugins.websocket.*
 import io.ktor.client.request.*
 import io.ktor.client.utils.*
 import io.ktor.http.*
@@ -28,8 +28,8 @@ internal suspend fun writeRequest(
     overProxy: Boolean,
     closeChannel: Boolean = true
 ) = withContext(callContext) {
-    writeHeaders(request, output, overProxy, closeChannel)
-    writeBody(request, output, callContext)
+    writeHeaders(request, output, overProxy = overProxy, closeChannelOnError = closeChannel)
+    writeBody(request, output, callContext, closeChannel)
 }
 
 @OptIn(InternalAPI::class)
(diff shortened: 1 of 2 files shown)


ktor-client/ktor-client-cio/common/src/io/ktor/client/engine/cio/utils.kt
ark-1: Out of the scope of this change, but the name is a bit misleading. I'd call this parameter `closeChannelOnError`.

ktor-client/ktor-client-cio/common/src/io/ktor/client/engine/cio/utils.kt
```diff
@@ -29,7 +29,7 @@ internal suspend fun writeRequest(
     closeChannel: Boolean = true
 ) = withContext(callContext) {
     writeHeaders(request, output, overProxy, closeChannel)
-    writeBody(request, output, callContext)
+    writeBody(request, output, callContext, closeChannel)
```
ark-1, line 32: I suggest passing boolean (or any other primitive) parameters by name. Even if it adds a little repetition. Otherwise you won't even notice a possible mix-up (like accidentally swapping `overProxy` and `closeChannel` in the call above).

ark-1 (approved):

ktor-client/ktor-client-cio/common/src/io/ktor/client/engine/cio/utils.kt
```diff
@@ -28,16 +28,16 @@ internal suspend fun writeRequest(
     overProxy: Boolean,
     closeChannel: Boolean = true
 ) = withContext(callContext) {
-    writeHeaders(request, output, overProxy, closeChannel)
-    writeBody(request, output, callContext)
+    writeHeaders(request, output, overProxy = overProxy, closeChannelOnError = closeChannel)
+    writeBody(request, output, callContext, closeChannel)
```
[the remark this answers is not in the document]
osipxd: It's a different issue. Tracked here:
[KTOR-9774](https://youtrack.jetbrains.com/issue/KTOR-9774) CIO pipelining corrupts requests with non-empty bodies

PR #5799, opened 2026-08-06
KTOR-9051 Implement lazy initialization of SLF4J logger to prevent disk access

Stexxe:
https://youtrack.jetbrains.com/issue/KTOR-9051

### DIFF
--- a/ktor-utils/jvm/src/io/ktor/util/logging/KtorSimpleLoggerJvm.kt
+++ b/ktor-utils/jvm/src/io/ktor/util/logging/KtorSimpleLoggerJvm.kt
@@ -5,6 +5,258 @@
 package io.ktor.util.logging
 
 import org.slf4j.*
+import java.util.concurrent.CompletableFuture
+import java.util.concurrent.Executors
 
 @Suppress("FunctionName")
-public actual fun KtorSimpleLogger(name: String): Logger = LoggerFactory.getLogger(name)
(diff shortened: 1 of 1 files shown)


ktor-utils/jvm/src/io/ktor/util/logging/KtorSimpleLoggerJvm.kt
```diff
@@ -6,5 +6,264 @@ package io.ktor.util.logging
...
+
 @Suppress("FunctionName")
-public actual fun KtorSimpleLogger(name: String): Logger = LoggerFactory.getLogger(name)
+public actual fun KtorSimpleLogger(name: String): Logger = if (!isAndroid) {
+    LoggerFactory.getLogger(name)
+} else {
+    object : org.slf4j.Logger {
+        private val logger by lazy {
+            LoggerFactory.getLogger(name)
+        }
```
bjhham, line 23: I wonder if you could do something like

```
val logger by lazy { LoggerFactory.getLogger(name) }
return object : Logger by logger {}
```

for the same effect.  I'm not sure if the delegation would unwrap the lazy though.

Stexxe: I don't want to risk relying on such unwrapping.

bjhham (approved):

PR #5800, opened 2026-08-06
KTOR-9778 Replace boxed char sets with ASCII bitmasks in ktor-http hot paths

inemtsev:
**Subsystem**
Client/Server, `ktor-http`

**Motivation**
I noticed that several hot paths in `ktor-http` classify characters using `Set<Char>` / `Set<Byte>` membership. Every one of those lookups boxes the character and does a hash lookup, and they run on every character on every URL that gets encoded, every header name that gets validated, and every parameterized header (like `Content-Type`) that gets rendered, on both client and server. 

Resolves [KTOR-9778](https://youtrack.jetbrains.com/issue/KTOR-9778).

**Solution**
I added a small internal `AsciiBitSet`: two `Long`s giving 128 bits, one per ASCII character, so a membership check is just two shifts and an AND. No boxing, no hashing, and it works on all KMP targets. This is the same trick .NET uses: [BitVector256](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/SearchValues/BitVector256.cs) behind `SearchValues`, and Kestrel's [HttpCharacters](https://github.com/dotnet/aspnetcore/blob/main/src/Shared/ServerInfrastructure/HttpCharacters.cs) validates header names/values the same way.

With that in place:

* `Codecs.kt` classifies via bitmasks and iterates UTF-8 bytes directly instead of going through `CharsetEncoder` + packet; `encodeURLPath` only touches the encoder for actual non-ASCII characters, and `%XX` digits are appended in place instead of allocating a `String` per byte.
* `HttpHeaders.checkHeaderName` does a single bitmask probe instead of a control-char check plus a linear scan through a 15-char string literal.
* `HeaderValueWithParameters` uses a bitmask for its separator set.

No public API/ABI changes; everything touched is private or internal.

I benchmarked old vs new with a small same-JVM A/B harness, included as a test (median of 10 alternating rounds, JDK 21, arm64; allocations measured via `ThreadMXBean.getThreadAllocatedBytes`):

| operation | old ns/op | new ns/op | speedup | old B/op | new B/op |
|---|---|---|---|---|---|
| encodeURLPathPart | 92.6 | 49.9 | 1.9x | 239 | 157 |
| encodeURLQueryComponent | 93.8 | 36.4 | 2.6x | 351 | 184 |
| encodeURLParameter | 125.1 | 40.3 | 3.1x | 419 | 184 |
| checkHeaderName | 30.3 | 13.2 | 2.3x | 0 | 0 |

As expected, the biggest wins are in the encoders that were paying for both the boxed lookups and the encoder/`String` allocations. The remaining allocations are just the result string and one byte-array copy, which the API contract requires anyway.


bjhham (changes requested):
Thanks for the optimization!   This one has crossed my mind quite a few times too.

Could you remove the benchmark test and (if you'd like) update the tests under https://github.com/ktorio/ktor-benchmarks/tree/main/io-benchmarks in a separate PR?

inemtsev:
Sure 🚀 

I removed the benchmark test from this PR and moved it to the other repo. The numbers are in the PR description for reference.

bjhham (approved):

PR #5801, opened 2026-08-07
KTOR-9774 Serialize CIO pipelined request bodies

osipxd:
**Subsystem**
Client, CIO

**Motivation**
[KTOR-9774](https://youtrack.jetbrains.com/issue/KTOR-9774) CIO can start writing the next pipelined request before the previous request body writer finishes. This allows two request writers to use the same connection concurrently and can corrupt HTTP/1.1 request framing.

**Solution**
Return the asynchronous body writer job from `writeBody` and `writeRequest`, and wait for it in the JVM connection pipeline before advancing to the next request. Response processing and dedicated-connection body writing remain concurrent.

Add a regression test that verifies two non-empty pipelined request bodies are written sequentially and received intact.

### DIFF
--- a/ktor-client/ktor-client-cio/jvm/src/io/ktor/client/engine/cio/ConnectionPipeline.kt
+++ b/ktor-client/ktor-client-cio/jvm/src/io/ktor/client/engine/cio/ConnectionPipeline.kt
@@ -53,7 +53,7 @@ internal actual class ConnectionPipeline actual constructor(
                     throw cause
                 }
 
-                writeRequest(task.request, networkOutput, task.context, overProxy, closeChannel = false)
+                writeRequest(task.request, networkOutput, task.context, overProxy, closeChannel = false)?.join()
                 networkOutput.flush()
             }
         } catch (_: ClosedChannelException) {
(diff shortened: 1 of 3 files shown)


ktor-client/ktor-client-cio/jvm/src/io/ktor/client/engine/cio/ConnectionPipeline.kt
```diff
@@ -53,7 +53,7 @@ internal actual class ConnectionPipeline actual constructor(
                     throw cause
                 }
 
-                writeRequest(task.request, networkOutput, task.context, overProxy, closeChannel = false)
+                writeRequest(task.request, networkOutput, task.context, overProxy, closeChannel = false)?.join()
                 networkOutput.flush()
```
bjhham, line 57: Maybe we could just use `withContext()` instead of launch / join here to reduce overhead and keep things sequential.

osipxd: Do you mean making `writeBody` suspend and replacing its `launch` with `withContext`?
The problem is that `writeBody` is also used outside of pipelining, where it needs to run concurrently with `readResponse`.

bjhham (approved):
👍

PR #5802, opened 2026-08-08
KTOR-9817 Retain borrowed CF references in CertificatePinner before bridging

RafaRuiz:
`CertificatePinner` passes six references it does not own to `CFBridgingRelease` on every TLS handshake:

- `publicKeyType` and `publicKeySize` come from `CFDictionaryGetValue`, which follows the Core Foundation Get Rule — the dictionary owns them.
- `kSecAttrKeyTypeRSA` and `kSecAttrKeyTypeECSECPrimeRandom` are Security framework constants, released twice each (`checkValidKeyType` and `getAsn1HeaderBytes`).

The corrupted objects are values of the `SecKeyCopyAttributes` dictionary, so nothing fails at the handshake. The process dies later, when the Kotlin/Native GC finalizes that dictionary and `-[__NSDictionaryM dealloc]` releases values that are already gone:

```
objc_object::release  <-  cow_cleanup  <-  -[__NSDictionaryM dealloc]
  <-  Kotlin_ObjCExport_releaseAssociatedObject  <-  GC finalizer processor
```

This reproduces reliably on watchOS (`arm64_32`): a few seconds after any pinned request, on an Apple Watch Series 8 running watchOS 26.6, with Ktor 3.5.1. It is masked on `arm64`, where those values land on tagged pointers and `release` is a no-op, and it never appears in the simulator if the pinner is not installed there. Verified fixed on-device with this patch.

Each call now retains the borrowed reference before handing ownership to `CFBridgingRelease`, leaving the retain count balanced. No behaviour or API change.

### DIFF
--- a/ktor-client/ktor-client-darwin-legacy/darwin/src/io/ktor/client/engine/darwin/certificates/LegacyCertificatePinner.kt
+++ b/ktor-client/ktor-client-darwin-legacy/darwin/src/io/ktor/client/engine/darwin/certificates/LegacyCertificatePinner.kt
@@ -303,9 +303,9 @@ public data class LegacyCertificatePinner(
         return publicKeyRef.use {
             val publicKeyAttributes = SecKeyCopyAttributes(publicKeyRef)
             val publicKeyTypePointer = CFDictionaryGetValue(publicKeyAttributes, kSecAttrKeyType)
-            val publicKeyType = CFBridgingRelease(publicKeyTypePointer) as NSString
+            val publicKeyType = CFBridgingRelease(CFRetain(publicKeyTypePointer)) as NSString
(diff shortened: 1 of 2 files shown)


osipxd (changes requested):
Thank you for the fix! It looks correct. The same ownership issue also exists in the `ktor-client-darwin-legacy` module. Could you apply the equivalent fix to `LegacyCertificatePinner`?

osipxd:
YouTrack issue: [KTOR-9817](https://youtrack.jetbrains.com/issue/KTOR-9817) Darwin CertificatePinner over-releases borrowed Core Foundation references

RafaRuiz:
Done, thanks.

osipxd (approved):
It was fast 😄  Thank you!

Issue #5694, opened 2026-06-11
[Bug] multipart/form-data is not processed on Kotlin/Native (CannotTransformContentToTypeException)


Desertniy:
## Description
When handling incoming HTTP requests with `Content-Type: multipart/form-data` on Kotlin/Native, the standard method `call.receiveMultipart()` fails to parse the request body. Client headers are formed correctly, and the server properly identifies the content type, but an exception is thrown during data reading.

The exact same code with identical dependencies works stably on JVM. The issue manifests exclusively on Native targets and has been reproduced across several consecutive library versions.

## Steps to reproduce
1. Start a Ktor server for Kotlin/Native target (linuxX64).
2. Create an endpoint:
```
put("/upload") {
    val multipart = call.receiveMultipart()
    // processing...
}
```
3. Send a PUT request with a valid header Content-Type: multipart/form-data; boundary=... and a form-data body.
4. Observe the error

### Expected behavior
The request is parsed successfully. receiveMultipart() returns an object containing all form parts and files, behaving identically to JVM.

### Actual behavior
An exception is thrown:
```
io.ktor.server.plugins.CannotTransformContentToTypeException: Cannot transform this request's content to io.ktor.http.content.MultiPartData
```
The server recognizes the correct Content-Type, but the standard receiver cannot complete processing the request body.

## Workaround found
The issue was bypassed by replacing the standard receiver with direct channel reading and manually initializing the instance of MultiPartData:
```
CIOMultipartDataBase(routingContext.call.coroutineContext, routingContext.call.receiveChannel(), routingContext.call.request.header(HttpHeaders.ContentType)!!, routingContext.call.request.contentLength(), formFieldLimit)
```

## Environment & Versions

Ktor version: 3.0.3, 3.4.0, 3.5.0 (issue persists across all tested versions)
Kotlin version: 2.3.21
Target: linuxX64
Client used for testing: Curl, Postman

hiSandog:
A useful isolation step would be to test the same Native endpoint with both `POST` and `PUT`, and with a single text part before adding a file part. That would distinguish a method-routing or receive-pipeline issue from a boundary/file-stream parsing problem, especially since the identical dependency set succeeds on JVM.

PR #5804, opened 2026-08-10
KTOR-9784 Support multipart on non-JVM server targets

gkdis6:
**Subsystem**
Server, `ktor-server-core`

**Motivation**

KTOR-9784, reported as #5694.

`call.receiveMultipart()` fails on non-JVM server targets:

```
io.ktor.server.plugins.CannotTransformContentToTypeException:
Cannot transform this request's content to io.ktor.http.content.MultiPartData
```

The default receive transformation has no `MultiPartData` branch in `commonMain`;
it falls through to `defaultPlatformTransformations`, whose non-JVM actual returns
`null`. The body then stays a `ByteReadChannel`, and `PipelineCall` throws. This is
a server-core wiring gap rather than an engine problem: the Content-Type is parsed
correctly and `call.receiveChannel()` works.

The same root cause produces a second symptom: `call.receiveParameters()` with a
`multipart/form-data` body reaches the common `Parameters::class` branch, which calls
`multiPartData(channel)` directly and hits
`error("Multipart is not supported on non JVM platforms")`, failing with
`IllegalStateException`.

The non-JVM stub used to be accurate. When the Native CIO server was added
(`0de7948fb`, KTOR-746), the CIO multipart parser lived in `ktor-http-cio/jvm/` only,
and the same was still true when JS/Wasm server targets were commonized
(`1115fae6f`). `4933074bb` (KTOR-6632, shipped in 3.1.0) moved
`CIOMultipartDataBase` and `parseMultipart` to `ktor-http-cio/common/`, leaving only
`discardBlocking()` as expect/actual, so that the **client** could receive multipart
data on every target. The server side was not updated, and the stub was later
relocated to `nonJvm` (`d0d6a7e83`) without being re-evaluated.

**Solution**

Move the JVM `multiPartData` implementation to `commonMain` and add a
`MultiPartData::class` branch to the common transformation.

- No new dependency: `ktor-server-core` already has `api(projects.ktorHttpCio)` in
  `commonMain`.
- No public API changes: every declaration involved is `internal`, so ABI dumps are
  unaffected.
- The parser this now reaches is the same one `ktor-client-core` has used from
  `commonMain` since 3.1.0.

**Tests**

The existing multipart tests in `TestEngineMultipartTest` run assertions inside a
call interceptor, where the pipeline swallows failures, so they passed on non-JVM
targets despite multipart being broken. Two tests were added that assert on the
response instead:

- `testReceiveMultipartFormItem` — the `receiveMultipart()` path
- `testReceiveParametersFromMultipart` — the `receiveParameters()` path

Both fail on `jsNodeTest` without this change and pass with it.

`testMultipartBiggerThanLimitFails` had an `if (!PlatformUtils.IS_JVM) return` guard,
an artifact of multipart being JVM-only; it is removed so the `formFieldLimit` path
is covered on the newly working targets. The guard on `testMultipartIsNotTruncated`
is kept, since that test allocates 42 MB.

Verified locally:

- `:ktor-server-tests:jsNodeTest` — RED before the change, GREEN after
- `:ktor-server-core:jvmTest`, `:ktor-server-tests:jvmTest` — pass
- `:ktor-server-tests:linuxX64TestKlibrary`, `:ktor-server-core:linuxX64MainKlibrary` — compile
- `:ktor-server-core:lintKotlin`, `:ktor-server-tests:lintKotlin` — pass

Kotlin/Native tests were not executed: no Kotlin/Native-capable host was available
(the machine used lacks Xcode, and a Linux aarch64 host is not a supported
Kotlin/Native host platform). Native coverage here is compile-only, and CI should be
the deciding signal for `linuxX64Test`. `checkKotlinAbi` could not run for the same
Xcode reason; the change is `internal`-only, so no dump updates are expected.

### DIFF
--- a/ktor-server/ktor-server-core/common/src/io/ktor/server/engine/DefaultTransform.kt
+++ b/ktor-server/ktor-server-core/common/src/io/ktor/server/engine/DefaultTransform.kt
@@ -5,6 +5,8 @@
 package io.ktor.server.engine
 
 import io.ktor.http.*
+import io.ktor.http.cio.*
+import io.ktor.http.cio.internals.*
 import io.ktor.http.content.*
 import io.ktor.server.application.*
 import io.ktor.server.http.content.*
@@ -16,6 +18,7 @@ import io.ktor.util.pipeline.*
 import io.ktor.utils.io.*
 import io.ktor.utils.io.charsets.*
 import io.ktor.utils.io.core.*
+import kotlinx.coroutines.*
 import kotlinx.io.*
 
 internal val LOGGER = KtorSimpleLogger("io.ktor.server.engine.DefaultTransform")
@@ -45,6 +48,8 @@ public fun ApplicationReceivePipeline.installDefaultTransformations() {
             ByteReadChannel::class -> null
 
             ByteArray::class -> channel.toByteArray()
+
+            MultiPartData::class -> multiPartData(channel)
 
             Parameters::class -> {
                 val contentType = withContentType(call) { call.request.contentType() }
@@ -99,7 +104,25 @@ internal expect suspend fun PipelineContext<Any, PipelineCall>.defaultPlatformTr
     query: Any
 ): Any?
 
-internal expect fun PipelineContext<*, PipelineCall>.multiPartData(rc: ByteReadChannel): MultiPartData
+@OptIn(InternalAPI::class)
+internal fun PipelineContext<*, PipelineCall>.multiPartData(rc: ByteReadChannel): MultiPartData {
+    val contentType = call.request.header(HttpHeaders.ContentType)
+        ?: throw UnsupportedMediaTypeException(null)
+
+    val contentLength = call.request.header(HttpHeaders.ContentLength)?.toLong()
+
+    try {
+        return CIOMultipartDataBase(
+            coroutineContext + Dispatchers.Unconfined,
+            rc,
+            contentType,
+            contentLength,
+            formFieldLimit = call.formFieldLimit
+        )
+    } catch (_: UnsupportedMediaTypeExceptionCIO) {
+        throw UnsupportedMediaTypeException(ContentType.parse(contentType))
+    }
+}
 
 internal inline fun <R> withContentType(call: PipelineCall, block: () -> R): R = try {
     block()
--- a/ktor-server/ktor-server-core/jvm/src/io/ktor/server/engine/DefaultTransformJvm.kt
+++ b/ktor-server/ktor-server-core/jvm/src/io/ktor/server/engine/DefaultTransformJvm.kt
@@ -4,19 +4,13 @@
 
 package io.ktor.server.engine
 
-import io.ktor.http.*
-import io.ktor.http.cio.*
-import io.ktor.http.cio.internals.*
-import io.ktor.http.content.*
 import io.ktor.server.application.*
-import io.ktor.server.plugins.UnsupportedMediaTypeException
 import io.ktor.server.request.*
 import io.ktor.util.pipeline.*
 import io.ktor.utils.io.*
 import io.ktor.utils.io.charsets.*
 import io.ktor.utils.io.jvm.javaio.*
 import io.ktor.utils.io.streams.*
-import kotlinx.coroutines.*
 import kotlinx.io.*
 import java.io.*
 
@@ -27,28 +21,7 @@ internal actual suspend fun PipelineContext<Any, PipelineCall>.defaultPlatformTr
 
     return when (call.receiveType.type) {
         InputStream::class -> receiveGuardedInputStream(channel)
-        MultiPartData::class -> multiPartData(channel)
         else -> null
-    }
-}
-
-@OptIn(InternalAPI::class)
-internal actual fun PipelineContext<*, PipelineCall>.multiPartData(rc: ByteReadChannel): MultiPartData {
-    val contentType = call.request.header(HttpHeaders.ContentType)
-        ?: throw UnsupportedMediaTypeException(null)
-
-    val contentLength = call.request.header(HttpHeaders.ContentLength)?.toLong()
-
-    try {
-        return CIOMultipartDataBase(
-            coroutineContext + Dispatchers.Unconfined,
-            rc,
-            contentType,
-            contentLength,
-            formFieldLimit = call.formFieldLimit
-        )
-    } catch (_: UnsupportedMediaTypeExceptionCIO) {
-        throw UnsupportedMediaTypeException(ContentType.parse(contentType))
     }
 }
 
--- a/ktor-server/ktor-server-core/nonJvm/src/io/ktor/server/engine/DefaultTransform.nonJvm.kt
+++ b/ktor-server/ktor-server-core/nonJvm/src/io/ktor/server/engine/DefaultTransform.nonJvm.kt
@@ -4,19 +4,14 @@
 
 package io.ktor.server.engine
 
-import io.ktor.http.content.*
 import io.ktor.server.application.*
 import io.ktor.util.pipeline.*
-import io.ktor.utils.io.*
 import io.ktor.utils.io.charsets.*
 import kotlinx.io.*
 
 internal actual suspend fun PipelineContext<Any, PipelineCall>.defaultPlatformTransformations(
     query: Any
 ): Any? = null
-
-internal actual fun PipelineContext<*, PipelineCall>.multiPartData(rc: ByteReadChannel): MultiPartData =
-    error("Multipart is not supported on non JVM platforms")
 
 internal actual fun Source.readTextWithCustomCharset(charset: Charset): String =
     error("Charset $charset is not supported on non JVM platforms")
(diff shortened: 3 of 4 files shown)


kdelay:
I ran into #5694 as well and had prototyped the same change on `release/3.x`, so a few notes that may save some triage time.

### Most of the red board looks unrelated to this patch

No multipart test fails in any of the failing configurations:

- **JavaScript on Chrome/Node.js** ([439156](https://ktor.teamcity.com/buildConfiguration/Ktor_KtorMatrixJavaScriptChromeNodeJs/439156)): a single failure, `ServerSentEventsTest.testCancellingUnderlyingConnection[js, browser]`. The same test failed on the default branch in [437888](https://ktor.teamcity.com/buildConfiguration/Ktor_KtorMatrixJavaScriptChromeNodeJs/437888).
- **Native Windows X64** ([439155](https://ktor.teamcity.com/buildConfiguration/Ktor_KtorMatrixNativeWindowsX64/439155)): 46 failures, all in the client `WebSocketTest`, `CacheTest`, `CacheLegacyStorageTest` and `HttpTimeoutTest` suites. That configuration also goes red on the default branch by itself ([438597](https://ktor.teamcity.com/buildConfiguration/Ktor_KtorMatrixNativeWindowsX64/438597), [438525](https://ktor.teamcity.com/buildConfiguration/Ktor_KtorMatrixNativeWindowsX64/438525), [438506](https://ktor.teamcity.com/buildConfiguration/Ktor_KtorMatrixNativeWindowsX64/438506)).
- **Allocation tests** ([439144](https://ktor.teamcity.com/buildConfiguration/Ktor_AllocationTests/439144)): `ClientCallAllocationTest.helloWorldFootprint` and `ServerCallAllocationTest.fileResponseFootprint`, which have been failing on the default branch since Aug 7 ([439007](https://ktor.teamcity.com/buildConfiguration/Ktor_AllocationTests/439007), [438996](https://ktor.teamcity.com/buildConfiguration/Ktor_AllocationTests/438996), [438995](https://ktor.teamcity.com/buildConfiguration/Ktor_AllocationTests/438995)).
- **Java 11 on Linux** ([439147](https://ktor.teamcity.com/buildConfiguration/Ktor_KtorMatrixCoreLinuxJava11/439147)) and **Build All Core** ([439160](https://ktor.teamcity.com/buildConfiguration/Ktor_KtorCore_All/439160)): CIO/OkHttp/Netty socket timeouts and WebSocket tests.

One failure in that last group is worth a closer look rather than a re-run: `io.ktor.server.http.ApplicationRequestContentTest.testDoubleReceiveRaceCondition` fails with `EOFException: Buffer exhausted before skipping 11 bytes` inside `DoubleReceive`. It is the only failure that sits in the receive pipeline this PR changes, and it did not appear in the last five default-branch runs of either configuration, so I would not assume it is flaky without one confirming run.

### Base branch

`CONTRIBUTING.md` points bug fixes at `release/*`, and this targets `main`. It matters more than usual here because the two bases need different diffs: on `release/3.x` the non-JVM actuals of `io.ktor.server.engine` are already split into `posix/src/.../DefaultTransformNix.kt` and `web/src/.../DefaultTransform.web.kt`, so there is no single `nonJvm/src/.../engine/DefaultTransform.nonJvm.kt` to delete the stub from. Two shapes are possible there — moving `multiPartData` into `jvmAndPosix` fixes Native and leaves the `web` actual erroring, while commonizing it as you do here also enables JS/Wasm. Deciding early which branch this lands on (`./switch-base-branch.sh`) avoids doing the work twice.

For what it is worth on the Native side: I built the `jvmAndPosix` variant on `release/3.x` and ran `:ktor-server-core:linuxX64Test` in a Linux x86_64 container. `receiveMultipart()` and `receiveParameters()` both work against a `multipart/form-data` body once the transform is wired up (120 tests, 0 failures), which matches your green `Native Linux X64` and `Native macOS Arm64` checks.

### One question on the tests

You dropped the `if (!PlatformUtils.IS_JVM) return` guard from `testMultipartBiggerThanLimitFails` but kept it on `testMultipartIsNotTruncated`. Is that deliberate because of the 42 MB form field on JS/Native, or should the second one be enabled now too?

gkdis6:
Thanks — the triage on the unrelated failures is genuinely helpful, and the base-branch point is well taken.

## `testDoubleReceiveRaceCondition`

I looked at this one rather than re-running it, and I don't think this patch can reach it.

`ApplicationRequestContentTestJvm.kt` exercises `receiveText()` under `DoubleReceive`, so `call.receiveType.type` is `String::class`. The only behavioural change in this PR is a new `MultiPartData::class` branch in the common `when` and the same branch being removed from the JVM actual; for any other receive type both the old and new code fall through to `defaultPlatformTransformations` and then the `AfterTransform` phase exactly as before. There is no shared state between the multipart path and the text path.

I ran the test 20 times locally (`:ktor-server-tests:jvmTest --tests "...testDoubleReceiveRaceCondition" --rerun`), 20/20 green, so I could not reproduce it here — which I realise is weak evidence given it is a 100-way concurrency test and my machine is not the CI runner.

The stack trace from build 439147 puts it in `DoubleReceive`'s cache teardown, not in the receive pipeline:

```
java.io.EOFException: Buffer exhausted before skipping 11 bytes.
	at kotlinx.io.Buffer.skip(Buffer.kt:293)
	at io.ktor.utils.io.core.ByteReadPacketKt.discard(ByteReadPacket.kt:73)
	at io.ktor.utils.io.ByteReadChannelOperationsKt.discard(ByteReadChannelOperations.kt:414)
	at io.ktor.server.plugins.doublereceive.MemoryCache$dispose$1.invokeSuspend(ByteArrayCache.kt:55)
	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:100)
```

`ByteArrayCache.kt:55` is the `GlobalScope.launch { reader.discard() }` inside `MemoryCache.dispose()`, which runs from the `ResponseSent` hook. `read()` returns that same `reader` channel while `fullBody` is still null (line 49), so the request coroutine and the `GlobalScope` discard can end up consuming one channel concurrently, and `ByteReadChannel.discard()` is check-then-act over the shared read buffer:

```kotlin
val count = minOf(remaining, readBuffer.remaining)
readBuffer.discard(count)   // -> request(count); skip(count)
```

The exception is thrown from a detached `GlobalScope` coroutine, so it is not tied to any particular request — which fits a 100-way concurrency test surfacing it occasionally.

The history for this test on `Ktor_KtorMatrixCoreLinuxJava11` supports that. Of the last 40 runs it failed 3 times, on three unrelated branches:

| build | branch | status |
|-------|--------|--------|
| 439378 | `pull/5009` | FAILURE |
| 439147 | `pull/5804` (this PR) | FAILURE |
| 436981 | `pull/5775` | FAILURE |

All 37 others passed, including every `main` and `release/3.x` run in that window — which is why checking only the default branch makes it look new. Nothing in `ktor-server-double-receive` has changed recently either; the last commit touching it is #5720, and only to relocate `FileCache` between source sets.

## Base branch

You are right about the layout — I confirmed it on `origin/release/3.x`:

```
ktor-server-core/jvm/src/io/ktor/server/engine/DefaultTransformJvm.kt
ktor-server-core/posix/src/io/ktor/server/engine/DefaultTransformNix.kt
ktor-server-core/web/src/io/ktor/server/engine/DefaultTransform.web.kt
```

so the stub lives in two places there, not one. Worth noting that `CIOMultipartDataBase` and `parseMultipart` are already in `ktor-http-cio/common/` on `release/3.x` too, so the commonised shape is available on that branch as well — the choice there is genuinely between your `jvmAndPosix` variant (Native only) and commonising (Native + JS/Wasm).

My preference is to keep this on `main`, for two reasons:

1. This enables functionality that was explicitly unsupported since the Native server was introduced, rather than restoring behaviour that regressed. `CONTRIBUTING.md` points *bug fixes* at `release/*`, and I read "make multipart work on targets where it never worked" as minor-release material even though there is no public API change.
2. The commonised shape is the better end state, and it leaves no target still throwing.

The call is the maintainers', though. If this should go to `release/3.x`, just tell me first which variant you want there — A (`jvmAndPosix`, Native only) or B (commonised, all targets). A leaves `web` throwing.

Also: thank you for running `:ktor-server-core:linuxX64Test` in an x86_64 container. I had no Kotlin/Native-capable host available, so my Native coverage was compile-only and I was relying on `jsNodeTest` as a proxy. Your 120/0 result closes the gap I flagged in the PR description.

## The test guards

Deliberate, yes — `testMultipartIsNotTruncated` builds a 42 MB form field, which I did not want to run on Node or Native just to widen coverage. `testMultipartBiggerThanLimitFails` uses 1000 bytes, so un-guarding it was cheap and it does cover the `formFieldLimit` path on the newly working targets.

One loose end I left alone: `testMultipartIsNotTruncated` still contains

```kotlin
if (response.status == HttpStatusCode.UnsupportedMediaType) {
    return@testApplication
}
```

which only existed because multipart was JVM-only, and is now dead on every target. If you would like that test enabled off-JVM I would rather do it with a smaller payload and drop that bail-out at the same time — say the word and I will add it here, otherwise I will leave it out of this PR's scope.

bjhham (approved):
Thanks for the contribution!  Looks good to me! 🎉

PR #5805, opened 2026-08-10
Fix KTOR-9760 Digest Auth: URI and HA2 are empty for a URL without a …

zibet27:
**Subsystem**
Client

**Motivation**
[KTOR-9760](https://youtrack.jetbrains.com/issue/KTOR-9760) Digest Auth: URI and HA2 are empty for a URL without a path

**Solution**
Use the request target that is actually sent by the client

### DIFF
--- a/ktor-client/ktor-client-plugins/ktor-client-auth/common/src/io/ktor/client/plugins/auth/providers/DigestAuthProvider.kt
+++ b/ktor-client/ktor-client-plugins/ktor-client-auth/common/src/io/ktor/client/plugins/auth/providers/DigestAuthProvider.kt
@@ -183,8 +183,11 @@ public class DigestAuthProvider(
         val credentials = tokenHolder.loadToken() ?: return
         val credential = makeDigest("${credentials.username}:$realm:${credentials.password}")
 
+        // Clients send "/" when the URL path is empty (RFC 9112); digest uri/HA2 must match that target.
+        val requestTarget = if (url.rawSegments.isEmpty()) "/${url.fullPath}" else url.fullPath
+
         val start = credential.toHexString()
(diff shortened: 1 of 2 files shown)


ktor-client/ktor-client-plugins/ktor-client-auth/common/src/io/ktor/client/plugins/auth/providers/DigestAuthProvider.kt
```diff
@@ -183,8 +183,11 @@ public class DigestAuthProvider(
         val credentials = tokenHolder.loadToken() ?: return
         val credential = makeDigest("${credentials.username}:$realm:${credentials.password}")
 
+        // Clients send "/" when the URL path is empty (RFC 9112); digest uri/HA2 must match that target.
+        val requestTarget = if (url.rawSegments.isEmpty()) "/${url.fullPath}" else url.fullPath
```
bjhham, line 187: Couldn't this just be `if (url.rawSegments.isEmpty()) "/" else url.fullPath` ?

zibet27: I had "/" at first, but found that `url.fullPath` can include query parameters

zibet27: The test covers non-empty query parameters

bjhham: Ohh right, these URL properties are confusing.

bjhham (approved):

PR #5811, opened 2026-08-12
KTOR-9787 Remove experimental flag for HTMX

bjhham:
**Subsystem**
Server, HTMX

**Motivation**
[KTOR-9787](https://youtrack.jetbrains.com/issue/KTOR-9787) Remove experimental flag for HTMX

Since this hasn't changed much at all since its introduction, I think we can safely remove the experimental flag now.

**Solution**
Removed the experimental flag.


osipxd (approved):
Btw, should we add KDocs to all public functions?

bjhham:
Good idea, I'll look through them and add docs where missing.

PR #5812, opened 2026-08-12
Fix KTOR-9793 The new DI scope is cancelled during hot reload

zibet27:
**Subsystem**
Server

**Motivation**
KTOR-9793 The new DI scope is cancelled during hot reload

**Solution**
- Identity-filter application events against the owning application
- Use `finally` blocks to guarantee DI-scope cancellation followed by subscription disposal


bjhham (changes requested):
Maybe let's close this and fix it in the embedded server since it seems like this assumption is made everywhere 😢

zibet27:
> Maybe let's close this and fix it in the embedded server since it seems like this assumption is made everywhere 😢

Agree

PR #5814, opened 2026-08-13
Accept optional whitespace around commas in the Range header

kdelay:
**Subsystem**

Server / `ktor-http` (`parseRangesSpecifier`), which is what `PartialContent` uses to read the `Range` header.

**Motivation**

`Range` carries a comma-separated list, and RFC 9110 [5.6.1](https://www.rfc-editor.org/rfc/rfc9110#section-5.6.1)
expands `1#element` to `element *( OWS "," OWS element )`, so `bytes=0-499, 1000-1499` is well-formed.

`parseRangesSpecifier` splits on `,` and hands each element to `toLong()` untrimmed:

```kotlin
val allRanges = allRangesString.split(',').map {
    ...
    ContentRange.Bounded(from.toLong(), to.toLong())
```

`" 1000"` throws `NumberFormatException`, the surrounding `catch (e: Throwable)` classifies the header as
syntactically incorrect, and the whole header is dropped rather than the one element.

Measured on `release/3.x` with the `PartialContentTest` harness (1100 byte body), the only difference between
the two requests being the space after the comma:

| `Range` | status | `Content-Range` |
| --- | --- | --- |
| `bytes=0-0,1-2` | 206 Partial Content | `bytes 0-2/1100` |
| `bytes=0-0, 1-2` | 200 OK | absent |

So a client that writes a space after the comma silently gets the full body instead of a 206.

No YouTrack ticket exists for this one, so there is no `KTOR-{NUM}` prefix.

**Solution**

Trim each list element before parsing it, and cover it with two tests in `RangesTest`
(space after the comma, and tab/space on both sides).

This covers only the OWS of the list rule. Empty list elements (`bytes=0-0,,1-2`) are still rejected;
that is a separate leniency and out of scope here.

Verification, with `JAVA_HOME` set to JDK 21:

- `./gradlew :ktor-http:jvmTest` — 294 tests, 0 failures.
- `./gradlew :ktor-http:lintKotlin` and `./gradlew :ktor-http:compileKotlinJs` pass.
- No public signature change, so there is no `api/*.api` update in the diff.

Reverting only the production hunk and keeping the tests fails, so the tests do pin the fix:

```
RangesTest[jvm] > testParseWhitespaceAfterComma()[jvm] FAILED
    org.opentest4j.AssertionFailedError: expected: <bytes=0-499,1000-1499> but was: <null>
RangesTest[jvm] > testParseWhitespaceAroundComma()[jvm] FAILED
    org.opentest4j.AssertionFailedError: expected: <bytes=9500-,-500> but was: <null>
```

Bug fix without API changes, so the base branch is `release/3.x` per the branching strategy in CONTRIBUTING.
The affected line is identical on `main`.

### DIFF
--- a/ktor-http/common/src/io/ktor/http/Ranges.kt
+++ b/ktor-http/common/src/io/ktor/http/Ranges.kt
@@ -75,12 +75,18 @@ public sealed class ContentRange {
  * Parse `Range` header value
  *
  * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.http.parseRangesSpecifier)
+ *
+ * @param rangeSpec the raw `Range` header value, for example `bytes=0-499, 1000-1499`.
+ * Optional whitespace (space or horizontal tab) around the commas separating the range set is accepted,
+ * as allowed by the list syntax of RFC 9110 5.6.1.
+ * @return the parsed specifier, or `null` if [rangeSpec] is syntactically incorrect or unsatisfiable
  */
 public fun parseRangesSpecifier(rangeSpec: String): RangesSpecifier? {
     try {
         val (unit, allRangesString) = rangeSpec.chomp("=") { return null }
-        // A range set is a comma-separated list, and RFC 9110 5.6.1 allows optional whitespace around the commas
-        val allRanges = allRangesString.split(',').map { it.trim() }.map {
+        // A range set is a comma-separated list, and RFC 9110 5.6.1 allows optional whitespace (SP / HTAB)
+        // around the commas
+        val allRanges = allRangesString.split(',').map { it.trim(' ', '\t') }.map {
             if (it.startsWith("-")) {
                 ContentRange.Suffix(it.removePrefix("-").toLong())
             } else {
--- a/ktor-http/common/src/io/ktor/http/Ranges.kt
+++ b/ktor-http/common/src/io/ktor/http/Ranges.kt
@@ -79,7 +79,8 @@ public sealed class ContentRange {
 public fun parseRangesSpecifier(rangeSpec: String): RangesSpecifier? {
     try {
         val (unit, allRangesString) = rangeSpec.chomp("=") { return null }
-        val allRanges = allRangesString.split(',').map {
+        // A range set is a comma-separated list, and RFC 9110 5.6.1 allows optional whitespace around the commas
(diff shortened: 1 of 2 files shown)


ktor-http/common/src/io/ktor/http/Ranges.kt
bjhham: Let's remove the unnecessary comment and merge the chained maps here.

```suggestion
        val allRanges = allRangesString.split(',').map { segment ->
            val range = segment.trim(' ', '\t')
            if (range.startsWith("-")) {
                ContentRange.Suffix(range.removePrefix("-").toLong())
            } else {
                val (from, to) = range.chomp("-") { "" to "" }
                when {
                    to.isNotEmpty() -> ContentRange.Bounded(from.toLong(), to.toLong())
                    else -> ContentRange.TailFrom(from.toLong())
                }
            }
        }
```

ktor-http/common/src/io/ktor/http/Ranges.kt
```diff
@@ -75,11 +75,18 @@ public sealed class ContentRange {
  * Parse `Range` header value
  *
  * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.http.parseRangesSpecifier)
+ *
+ * @param rangeSpec the raw `Range` header value, for example `bytes=0-499, 1000-1499`.
+ * Optional whitespace (space or horizontal tab) around the commas separating the range set is accepted,
+ * as allowed by the list syntax of RFC 9110 5.6.1.
```
bjhham: Don't need to reference this in the KDoc.

bjhham (changes requested):
Looks good, couple of nitpicks.

ktor-http/common/src/io/ktor/http/Ranges.kt
kdelay: Applied your suggestion as is in 2cc075d. The item is trimmed inside the single map now and the comment is gone.

ktor-http/common/src/io/ktor/http/Ranges.kt
kdelay: Dropped those two lines in 2cc075d, so the KDoc keeps only the `@param`/`@return` description.

bjhham (approved):

PR #5816, opened 2026-08-13
Keep only one Application running in Hot Reload

zibet27:
**Subsystem**
Server Hot Reload

**Motivation**
[KTOR-9792](https://youtrack.jetbrains.com/issue/KTOR-9792) Application lifecycle events affect other reload generations

https://github.com/ktorio/ktor/issues/4717

**Previous Solution**
The solution for https://github.com/ktorio/ktor/issues/4717 unnecessarily had 2 applications running during hot reload, which is breaking application monitoring, because most of them don't filter the target event application.

**Proposed Solution**
Revert the previous fix; first, stop the current app and then try to start the new one; if it fails, the server should still be active and watch for changes, though failing all incoming requests.

### DIFF
--- a/ktor-server/ktor-server-core/jvm/src/io/ktor/server/engine/EmbeddedServerJvm.kt
+++ b/ktor-server/ktor-server-core/jvm/src/io/ktor/server/engine/EmbeddedServerJvm.kt
@@ -2,6 +2,8 @@
  * Copyright 2014-2025 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
  */
 
+@file:OptIn(InternalAPI::class)
+
 package io.ktor.server.engine
 
 import io.ktor.events.*
@@ -27,6 +29,7 @@ import java.util.concurrent.locks.ReentrantReadWriteLock
 import kotlin.concurrent.getOrSet
 import kotlin.concurrent.read
 import kotlin.concurrent.write
+import kotlin.time.Duration.Companion.milliseconds
 
 private typealias ApplicationModule = suspend Application.() -> Unit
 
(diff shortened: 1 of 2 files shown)


ktor-server/ktor-server-core/jvm/src/io/ktor/server/engine/EmbeddedServerJvm.kt
```diff
@@ -61,9 +64,10 @@ actual constructor(
     private val moduleInjector: ModuleParametersInjector by lazy {
         loadServiceOrNull() ?: ModuleParametersInjector.Disabled
     }
-    private val modules: List<DynamicApplicationModule> get() =
-        environment.moduleConfigReferences.map(::dynamicModule) +
-            rootConfig.modules.map { module -> module.toDynamicModuleOrNull() ?: module.wrapWithDynamicModule() }
+    private val modules: List<DynamicApplicationModule>
+        get() =
+            environment.moduleConfigReferences.map(::dynamicModule) +
+                rootConfig.modules.map { module -> module.toDynamicModuleOrNull() ?: module.wrapWithDynamicModule() }
```
bjhham: Odd formatting change... 🤔

ktor-server/ktor-server-core/jvm/src/io/ktor/server/engine/EmbeddedServerJvm.kt
```diff
@@ -105,35 +110,33 @@ actual constructor(
...
-                    "Auto-reload failed; continuing to serve the previously loaded application.",
-                    cause,
-                )
+                environment.log.error("Auto-reload failed.", cause)
+                throw cause
             }
         }
 
         return@read applicationInstance ?: error("EmbeddedServer was stopped")
     }
```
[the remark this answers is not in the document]
bjhham: Actually, not a bad idea @zibet27 😄 🐇

ktor-server/ktor-server-core/jvm/src/io/ktor/server/engine/EmbeddedServerJvm.kt
```diff
@@ -105,35 +110,33 @@ actual constructor(
...
 
     private fun currentApplication(): Application = applicationInstanceLock.read {
-        val currentApplication = applicationInstance ?: error("EmbeddedServer was stopped")
-
         if (!rootConfig.developmentMode) {
-            return@read currentApplication
+            return@read applicationInstance ?: error("EmbeddedServer was stopped")
```
bjhham: Maybe the error message should be something like "Application failed to reload; check log for details"?  From what I can tell this would be the only scenario where we can expect this to happen.

bjhham (approved):
Looks good, a couple of minor comments there

bjhham (approved):

PR #5817, opened 2026-08-15
KTOR-9825 Unescape quoted-pair at the end of a header parameter value

kdelay:
**Subsystem**
Core (ktor-http)

**Motivation**
`parseHeaderValue` unescaped a backslash in a quoted parameter value only when it sat more than two characters from the end of the header, so an escaped character in the tail kept its backslash. A header Ktor renders itself then does not survive a round trip:

```kotlin
val rendered = ContentDisposition.File.withParameter("filename", "a\"").toString()
// file; filename="a\""
parseHeaderValue(rendered)  // filename=a\"  (expected a")
```

**Solution**
Unescape whenever a character follows the backslash. An unterminated quoted value still falls back to the raw substring, so `justValue;x="abc\q` parses to `"abc\q` as before.

Verified on JDK 21: `:ktor-http:jvmTest` (new tests fail before the change), `:ktor-client-core:jvmTest`, `:ktor-server-core:jvmTest`, four affected server plugins, `formatKotlin`/`lintKotlin`.

### DIFF
--- a/ktor-http/common/src/io/ktor/http/HttpHeaderValueParser.kt
+++ b/ktor-http/common/src/io/ktor/http/HttpHeaderValueParser.kt
@@ -217,7 +217,7 @@ private fun parseHeaderValueParameterValueQuoted(value: String, start: Int): Pai
                 return position + 1 to builder.toString()
             }
 
-            currentChar == '\\' && position < value.lastIndex - 2 -> {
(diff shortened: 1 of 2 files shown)


zibet27 (approved):
Good catch!
@kdelay Thank you for the contribution!

zibet27:
I will create a YouTrack issue for this PR

PR #5822, opened 2026-08-20
KTOR-9818 Performance improvements for Netty HTTP/3

bjhham:
**Subsystem**
Server, Netty

**Motivation**
[KTOR-9818](https://youtrack.jetbrains.com/issue/KTOR-9818) Netty HTTP/3 Performance Improvements

**Solution**

Benchmarks showed the HTTP/3 connector running about 25x slower than HTTP/2. The main cause was that the whole endpoint ran on a single thread: Netty registers every QUIC connection and stream on the event loop of the datagram socket that received it, so with one bound UDP socket all packet I/O, TLS crypto and QPACK for every connection ran on one event loop (150-200% CPU on a 64 thread box, vs 5000%+ for HTTP/2). A few per-connection and per-response overheads sat on top of that.

Changes:

* Bind multiple UDP sockets per SSL connector with SO_REUSEPORT so the kernel spreads connections across event loops. Defaults to workerGroupSize sockets on Linux, which is the only OS that load balances unicast UDP this way, and 1 elsewhere. Configurable via the new `udpSocketCount` option. On the NIO transport the JDK socket option is used instead of UnixChannelOption (which NIO silently ignores); it's resolved reflectively because the field is Java 9+ and we compile against the Java 8 API.
* When more than one socket is bound, a shared `QuicCodecDispatcher` is installed on all of them. It encodes the socket index into server connection ids and re-dispatches datagrams that the kernel hashed to the wrong socket, e.g. after connection migration or NAT rebinding. It's applied after `configureQuicServerCodec` so user config can't break the routing.
* Address validation (stateless Retry) is now opt-in. `quicTokenHandler` defaults to null, matching Netty's own default. Previously every handshake paid an extra round trip plus a second Initial processing pass. HmacQuicTokenHandler is now public for anyone who wants validation back, and it caches Mac instances per thread instead of creating one per token.
* HTTP/3 responses now end the stream with QuicStreamChannel.shutdownOutput() instead of flush + close, so the FIN coalesces with the last data packet instead of going out as a separate datagram. This also fixed a latent double close in handleLastResponseMessage that would have queued a second FIN under the new scheme.
* New `udpReceiveBufferSize` / `udpSendBufferSize` options (SO_RCVBUF / SO_SNDBUF), and GSO is enabled on epoll (up to 10 packets per syscall, falls back to NONE without kernel support).
* Stream exceptions are logged at debug instead of error, same as the HTTP/2 handler. Client resets during load tests were producing a stack trace per stream.

Testing: NettyHttp3MultiSocketTest runs the whole NettyHttp3Test suite with udpSocketCount = 2, covering the dispatcher and the NIO reuseport path. On macOS all traffic lands on one socket, but the double bind and dispatcher-managed connection id generation are still exercised; on Linux CI the kernel actually balances. Full jvmTest, lint and ABI checks pass.

API surface (all new in 3.6.0): the new NettyHttp3Configuration options, quicTokenHandler made nullable, and HmacQuicTokenHandler promoted from internal to public.

Background reading for the less obvious parts:
* SO_REUSEPORT balancing: https://lwn.net/Articles/542629/
* Retry / address validation trade-off: RFC 9000 section 8.1
  https://www.rfc-editor.org/rfc/rfc9000.html#section-8.1
* QuicCodecDispatcher (added in Netty QUIC 0.0.59): see Http3ServerExample and
  QuicServerSoReusePortExample in the netty repo
* UDP GSO: https://blog.cloudflare.com/accelerating-udp-packet-transmission-for-quic/

### DIFF
--- a/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationEngine.kt
+++ b/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationEngine.kt
@@ -41,6 +41,7 @@ import java.net.StandardSocketOptions
 import java.security.PrivateKey
 import java.security.cert.X509Certificate
 import java.util.concurrent.TimeUnit
+import kotlin.getValue
 import kotlin.reflect.KClass
 import kotlin.system.measureTimeMillis
 
@@ -303,39 +304,40 @@ public class NettyApplicationEngine(
      * channels reject the option) — an actual NIO `DatagramChannel`'s `supportedOptions()` is
      * probed to confirm real support before the option is used.
      */
-    private val reusePortOption: ChannelOption<Boolean>? get() = reusePortResolution.option
+    private val reusePortOption: ChannelOption<Boolean>? get() = reusePortResolution.getOrNull()
 
     /**
      * Explains why [reusePortOption] is `null`, distinguishing "this JDK doesn't have the
      * `SO_REUSEPORT` constant" (needs Java 9+) from "this JDK has it, but the platform's NIO
      * provider rejects it anyway" (for example, Windows) — the two require different advice.
      */
-    private val reusePortResolution: ReusePortResolution by lazy {
+    private val reusePortResolution: Result<ChannelOption<Boolean>> by lazy {
         if (KQueue.isAvailable() || Epoll.isAvailable()) {
-            return@lazy ReusePortResolution(UnixChannelOption.SO_REUSEPORT, null)
-        }
-        try {
-            @Suppress("UNCHECKED_CAST")
-            val soReusePort = StandardSocketOptions::class.java.getField("SO_REUSEPORT")
-                .get(null) as SocketOption<Boolean>
-            val supported = java.nio.channels.DatagramChannel.open().use { channel ->
-                channel.supportedOptions().contains(soReusePort)
+            Result.success(UnixChannelOption.SO_REUSEPORT)
+        } else {
+            try {
+                @Suppress("UNCHECKED_CAST")
+                val soReusePort = StandardSocketOptions::class.java.getField("SO_REUSEPORT")
+                    .get(null) as SocketOption<Boolean>
+                val supported = java.nio.channels.DatagramChannel.open().use { channel ->
+                    channel.supportedOptions().contains(soReusePort)
+                }
+                if (!supported) {
+                    Result.failure(
+                        IllegalArgumentException(
+                            "the current platform's NIO datagram provider does not support SO_REUSEPORT"
+                        )
+                    )
+                } else {
+                    Result.success(NioChannelOption.of(soReusePort))
+                }
+            } catch (_: ReflectiveOperationException) {
+                Result.failure(IllegalArgumentException("SO_REUSEPORT requires running on Java 9 or newer"))
+            } catch (_: java.io.IOException) {
+                Result.failure(IllegalArgumentException("SO_REUSEPORT support could not be determined"))
             }
-            if (!supported) {
-                return@lazy ReusePortResolution(
-                    null,
-                    "the current platform's NIO datagram provider does not support SO_REUSEPORT"
-                )
-            }
-            ReusePortResolution(NioChannelOption.of(soReusePort), null)
-        } catch (_: ReflectiveOperationException) {
-            ReusePortResolution(null, "SO_REUSEPORT requires running on Java 9 or newer")
-        } catch (_: java.io.IOException) {
-            ReusePortResolution(null, "SO_REUSEPORT support could not be determined")
         }
     }
-
-    private class ReusePortResolution(val option: ChannelOption<Boolean>?, val unsupportedReason: String?)
 
     /**
      * The number of UDP sockets bound per HTTP/3 connector.
@@ -352,7 +354,7 @@ public class NettyApplicationEngine(
             configured != null -> {
                 check(configured == 1 || reusePortOption != null) {
                     "udpSocketCount = $configured requires SO_REUSEPORT support, but " +
-                        "${reusePortResolution.unsupportedReason}. " +
+                        "${reusePortResolution.exceptionOrNull()?.message}. " +
                         "Use a native transport (epoll/kqueue) or set udpSocketCount = 1."
                 }
                 configured
--- a/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationEngine.kt
+++ b/ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationEngine.kt
@@ -25,6 +25,7 @@ import io.netty.channel.kqueue.KQueueDatagramChannel
 import io.netty.channel.kqueue.KQueueServerSocketChannel
 import io.netty.channel.socket.DatagramChannel
 import io.netty.channel.socket.ServerSocketChannel
+import io.netty.channel.socket.nio.NioChannelOption
 import io.netty.channel.socket.nio.NioDatagramChannel
 import io.netty.channel.socket.nio.NioServerSocketChannel
 import io.netty.channel.unix.UnixChannelOption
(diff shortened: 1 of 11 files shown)


ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/http3/NettyHttp3Configuration.kt
```diff
@@ -23,14 +23,20 @@ import kotlin.time.Duration.Companion.seconds
...
+     * round trip and roughly doubles the Initial packet processing per connection. Enable it when
+     * the endpoint is exposed to untrusted networks and needs protection against source-address
+     * spoofing and amplification attacks.
+     *
+     * [HmacQuicTokenHandler] is provided as a secure implementation that signs tokens
+     * with HMAC-SHA256 and rejects forged or expired tokens. Callers may also supply a custom
+     * [QuicTokenHandler] to use a different signing strategy or integrate with external token services.
      */
-    public var quicTokenHandler: QuicTokenHandler = HmacQuicTokenHandler()
+    public var quicTokenHandler: QuicTokenHandler? = null
```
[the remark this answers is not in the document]
bjhham: This feature is not yet released, so breaking changes of this type are acceptable.

ktor-server/ktor-server-netty/jvm/test/io/ktor/tests/server/netty/NettyEngineTest.kt
```diff
@@ -857,6 +862,33 @@ class NettyHttp3Test :
...
+ *
+ * Skipped when `SO_REUSEPORT` is unavailable: the NIO transport needs the Java 9+ socket option,
+ * so on Java 8 the tests only run with a native transport (epoll/kqueue) on the classpath.
+ */
+class NettyHttp3MultiSocketTest : NettyHttp3Test() {
+
+    @BeforeTest
+    fun assumeReusePortSupported() {
+        val reusePortSupported = Epoll.isAvailable() || KQueue.isAvailable() ||
+            runCatching { StandardSocketOptions::class.java.getField("SO_REUSEPORT") }.isSuccess
```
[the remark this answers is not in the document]
bjhham: The other conditions (Epoll / KQueue) account for these environments.

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/http3/HmacQuicTokenHandler.kt
```diff
@@ -52,16 +57,29 @@ private const val TIMESTAMP_LENGTH = 8
...
  * @param keyGen a function for providing the secret key used in HMAC signing and validation.
- *   If not provided, a random 256-bit key is generated.
+ *   If not provided, a random 256-bit key is generated. Provide a shared key when tokens
+ *   must validate across multiple server instances.
  * @param tokenLifetimeMillis maximum age of a valid token in milliseconds.
  */
-internal class HmacQuicTokenHandler(
+public class HmacQuicTokenHandler(
     keyGen: () -> SecretKey = ::generateDefaultKey,
     private val tokenLifetimeMillis: Long = TOKEN_LIFETIME_MS,
```
zibet27: Does it make sense to make a duration?

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/http3/NettyHttp3ChannelInitializer.kt
```diff
@@ -44,23 +88,39 @@ internal class NettyHttp3ChannelInitializer(
...
+                        QuicChannelOption.SEGMENTED_DATAGRAM_PACKET_ALLOCATOR,
+                        EpollQuicUtils.newSegmentedAllocator(10)
+                    )
+                }
+            }
             .apply(http3Configuration.configureQuicServerCodec)
+            .apply {
+                // Applied after user configuration on purpose: with multiple sockets, the dispatcher
+                // owns connection-id generation, and overriding it would break cross-socket routing.
+                if (localConnectionIdLength != null) localConnectionIdLength(localConnectionIdLength)
```
zibet27: `localConnectionIdLength(localConnectionIdLength)` looks a bit confusing

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/http3/NettyHttp3Handler.kt
```diff
@@ -94,7 +94,9 @@ internal class NettyHttp3Handler(
 
     @Suppress("OverridingDeprecatedMember")
```
zibet27: Looks like the suppress is redundant here

ktor-server/ktor-server-netty/jvm/test/io/ktor/tests/server/netty/NettyEngineTest.kt
```diff
@@ -21,29 +21,33 @@ import io.ktor.server.test.base.*
...
 import io.ktor.websocket.*
 import io.netty.bootstrap.Bootstrap
 import io.netty.buffer.ByteBuf
 import io.netty.buffer.Unpooled
-import io.netty.channel.*
+import io.netty.channel.ChannelHandlerContext
+import io.netty.channel.ChannelInboundHandlerAdapter
+import io.netty.channel.epoll.Epoll
+import io.netty.channel.kqueue.KQueue
 import io.netty.channel.nio.NioEventLoopGroup
```
zibet27, line 33: `NioEventLoopGroup` seems to be deprecated

ktor-server/ktor-server-netty/jvm/test/io/ktor/tests/server/netty/NettyEngineTest.kt
zibet27: might want to use withContext(Dispatchers.IO) {} here

zibet27 (approved):

zibet27:
We could also set a custom `sslTaskExecutor` to offload the datagram channel's event loop to improve benchmark results😃

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/http3/HmacQuicTokenHandler.kt
bjhham: Good call, duration is nicer.

bjhham:
> We could also set a custom sslTaskExecutor to offload the datagram channel's event loop to improve benchmark results😃

I think that could mess with the thread pinning and cause some incorrectness in the state.  The extra UDP sockets using the QuicCodecDispatcher here seems to do the trick for parallelism.  I'll try playing around with this setting some more to see if there's a better default.

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/NettyApplicationEngine.kt
```diff
@@ -297,22 +297,46 @@ public class NettyApplicationEngine(
...
+            ReusePortResolution(NioChannelOption.of(soReusePort), null)
         } catch (_: ReflectiveOperationException) {
-            null
+            ReusePortResolution(null, "SO_REUSEPORT requires running on Java 9 or newer")
+        } catch (_: java.io.IOException) {
+            ReusePortResolution(null, "SO_REUSEPORT support could not be determined")
         }
     }
 
+    private class ReusePortResolution(val option: ChannelOption<Boolean>?, val unsupportedReason: String?)
```
zibet27: Is it better than Kotlin's `Result`?

zibet27:
> I think that could mess with the thread pinning and cause some incorrectness in the state

I don't see why, Netty just will delegate work to another executor

> the QuicCodecDispatcher here seems to do the trick for parallelism

Yeah, that's why I would suggest not creating our `sslTaskExecutor` by default; users who need it will use `configureQuicServerCodec`. This might be useful for cryptography-heavy use cases like benchmarks.

Anyway, great work here!
Can't wait for some perf comparison

ktor-server/ktor-server-netty/jvm/src/io/ktor/server/netty/http3/HmacQuicTokenHandler.kt
```diff
@@ -88,7 +108,7 @@ internal class HmacQuicTokenHandler(
         val timestamp = token.getLong(token.readerIndex())
 
         val now = System.currentTimeMillis()
-        if (now - timestamp > tokenLifetimeMillis || timestamp > now) return -1
+        if (now - timestamp > tokenLifetime.inWholeMilliseconds || timestamp > now) return -1
```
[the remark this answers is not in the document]
bjhham: This is too much of an edge case to be relevant.

PR #5824, opened 2026-08-21
Fixes from codex for 3.6

bjhham:
**Subsystem**
Core / JSON schema

**Motivation**
https://github.com/ktorio/ktor/pull/5821

**Solution**
The max arg was inconsistently but always using bytes prior to changes in 3.6, so I updated the KDoc and made this argument apply consistently.

I also noticed that in many places, we're using `buffer.size` as a proxy for `Source.size`, but it is actually `buffer.size <= Source.size` for implementations that are not `Buffer`.  I fixed the spots relevant to the change here, but there could be other problems abound that we've not noticed due to our codebase generally only referencing `Buffer`.

### DIFF
--- a/ktor-io/jvm/src/io/ktor/utils/io/charsets/CharsetJVM.kt
+++ b/ktor-io/jvm/src/io/ktor/utils/io/charsets/CharsetJVM.kt
@@ -83,10 +83,11 @@ public actual fun CharsetDecoder.decode(input: Source, dst: Appendable, max: Int
         return input.readString().also { dst.append(it) }.length
     }
     // ensure buffer is filled
-    input.request(max.toLong())
-    val count = minOf(input.remaining.toInt(), max)
-    dst.append(input.readByteString(count).decodeToString(charset))
-    return count
+    val maxBytes = max.toLong()
+    input.request(maxBytes)
+    val count = minOf(input.remaining, maxBytes)
+    dst.append(input.readString(count, charset))
+    return count.toInt()
 }
 
 // ----------------------------------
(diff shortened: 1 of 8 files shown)


zibet27 (approved):

ktor-io/jvm/src/io/ktor/utils/io/charsets/CharsetJVM.kt
```diff
@@ -82,10 +82,12 @@ public actual fun CharsetDecoder.decode(input: Source, dst: Appendable, max: Int
...
-
-    val result = input.remaining
-    dst.append(input.readByteString().decodeToString(charset))
-    return result.toInt()
+    // ensure buffer is filled
+    val maxBytes = max.toLong()
+    input.request(maxBytes)
+    val count = minOf(input.remaining, maxBytes)
+    dst.append(input.readString(count, charset))
+    return count.toInt()
```
osipxd, line 90: Is it expected that we return the string length when `charset == UTF_8`, and bytes count in other cases?

bjhham: It's returning byte count in all cases here... `input.remaining` is the buffer size.  The return value is ignored everywhere in Ktor and there's no KDoc.  Maybe I'll add some documentation.

osipxd: I mean this piece of code:

```kotlin
if (charset == Charsets.UTF_8) {
    return input.readString().also { dst.append(it) }.length
}
```

bjhham: Ohh I see.  Yeah, that's not right.

PR #5825, opened 2026-08-23
Fix Jetty bodyWriter leaking suspended coroutines on aborted writes

daft-panda:
**Subsystem**

  Server, `ktor-server-jetty-jakarta`

  **Motivation**

  When an asynchronous Jetty response write fails, such as when a client
  disconnects mid-response, `Callback.failed()` cancels the channel but does
  not resume the coroutine suspended around `response.write()`.

  As a result, the body writer job never completes and cannot reach its
  cleanup path. This retains the handler job, application call, Jetty
  exchange, and borrowed buffer for every aborted response.

  **Solution**

  Resume the pending write continuation exceptionally with the same failure
  after cancelling the channel. This lets the writer unwind through its
  `catch` and `finally` blocks and recycle the pooled buffer.

  Add a focused regression test that verifies a failed Jetty write completes
  the body writer job. The test times out without the fix and passes with it.

  **Testing**

  - `./gradlew :ktor-server-jetty-jakarta:jvmTest`
  - `./gradlew :ktor-server-jetty-jakarta:assemble`
  - `./gradlew :ktor-server-jetty-jakarta:formatKotlin`
  - `./gradlew :ktor-server-jetty-jakarta:lintKotlin`

### DIFF
--- a/ktor-server/ktor-server-jetty-jakarta/jvm/src/io/ktor/server/jetty/jakarta/JettyResponseBodyWriter.kt
+++ b/ktor-server/ktor-server-jetty-jakarta/jvm/src/io/ktor/server/jetty/jakarta/JettyResponseBodyWriter.kt
@@ -14,6 +14,7 @@ import org.eclipse.jetty.util.Callback
 import org.eclipse.jetty.util.thread.Invocable
 import kotlin.coroutines.Continuation
 import kotlin.coroutines.resume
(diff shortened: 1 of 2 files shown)


bjhham (commented):
lgtm, thanks! 👍

bjhham (approved):

PR #5826, opened 2026-08-23
KTOR-9827 Authorization header removed from refresh token request

Thomas-Vos:
**Subsystem**
ktor-client-auth

**Motivation**
https://youtrack.jetbrains.com/issue/KTOR-9827

**Solution**
Authorization header is no longer removed for refresh token requests.

### DIFF
--- a/ktor-client/ktor-client-plugins/ktor-client-auth/common/src/io/ktor/client/plugins/auth/providers/BearerAuthProvider.kt
+++ b/ktor-client/ktor-client-plugins/ktor-client-auth/common/src/io/ktor/client/plugins/auth/providers/BearerAuthProvider.kt
@@ -52,6 +52,9 @@ public class RefreshTokensParams(
     /**
      * Marks that this request is for refreshing auth tokens, resulting in a special handling of it.
      *
+     * When a request is marked no additional Authorization headers are included and any custom
+     * Authorization headers are kept.
+     *
      * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.client.plugins.auth.providers.RefreshTokensParams.markAsRefreshTokenRequest)
      */
     public fun HttpRequestBuilder.markAsRefreshTokenRequest() {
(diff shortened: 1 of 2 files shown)


ktor-client/ktor-client-plugins/ktor-client-auth/common/src/io/ktor/client/plugins/auth/providers/BearerAuthProvider.kt
```diff
@@ -231,11 +231,11 @@ public class BearerAuthProvider(
         val token = tokensHolder.loadToken() ?: return
 
         request.headers {
-            val tokenValue = "Bearer ${token.accessToken}"
-            if (contains(HttpHeaders.Authorization)) {
-                remove(HttpHeaders.Authorization)
-            }
             if (request.attributes.contains(AuthCircuitBreaker).not()) {
```
zibet27: This check could be moved before token loading the token

ktor-client/ktor-client-plugins/ktor-client-auth/common/src/io/ktor/client/plugins/auth/providers/BearerAuthProvider.kt
```diff
@@ -231,11 +231,11 @@ public class BearerAuthProvider(
...
+                val tokenValue = "Bearer ${token.accessToken}"
+                if (contains(HttpHeaders.Authorization)) {
```
zibet27: We can now just use `this[HttpHeaders.Authorization] = tokenValue`

zibet27 (changes requested):
@Thomas-Vos, thank you for the contribution!
Code looks good to me.
Just a couple of nitpicks that would be nice to improve before merging.

Also, we should improve KDoc for `markAsRefreshTokenRequest` to mention its behaviour. Would be nice if you could do it in this PR 😅, though it's ok if not.

Thomas-Vos:
@zibet27 Done, see latest commit.

ktor-client/ktor-client-plugins/ktor-client-auth/common/src/io/ktor/client/plugins/auth/providers/BearerAuthProvider.kt
```diff
@@ -52,6 +52,9 @@ public class RefreshTokensParams(
     /**
      * Marks that this request is for refreshing auth tokens, resulting in a special handling of it.
      *
+     * When a request is marked no additional Authorization headers are included and any custom
```
zibet27:
```suggestion
     * When a request is marked, no additional Authorization headers are included, and any custom
```

zibet27 (approved):
@Thomas-Vos Thank you!

PR #5831, opened 2026-08-24
KTOR-5577 Support nullable types in receive

osipxd:
**Subsystem**
Server, Serialization, Utils

**Motivation**
[`KTOR-5577`](https://youtrack.jetbrains.com/issue/KTOR-5577) tracks the confusing nullability contract of `ApplicationCall.receiveNullable<T>()`: it accepts a non-nullable `T`, while serializers use the nullability encoded in the requested type. `TypeInfo` also loses nullability when complete `KType` information is unavailable.

**Solution**
The idea is to allow `ApplicationCall.receive<T>()` to use nullable `T`, and deprecate `receiveNullable<T>()` in favor of `receive<T?>()`:

```diff
 post("/") {
-    val payload = call.receiveNullable<Payload?>()
+    val payload = call.receive<Payload?>()
     // use the payload
 }
```

To make nullability inference reliable I had to add `isNullable` property to `TypeInfo` and preserve it through serialization, content negotiation, parameter conversion, WebSocket serialization, and dependency-injection type transformations. The reason behind this is that `typeInfo<T>()` uses `typeOfOrNull` so the information about nullability is not reliably available.

**Backward compatibility**
- Keep the previous `receive(TypeInfo)` extension and `receiveNullable(TypeInfo)` implementation path for source and binary compatibility.
- Add default implementation for the introduced method

It's better to review by commits.

### DIFF
--- a/ktor-server/ktor-server-core/common/src/io/ktor/server/application/PipelineCall.kt
+++ b/ktor-server/ktor-server-core/common/src/io/ktor/server/application/PipelineCall.kt
@@ -7,14 +7,14 @@ package io.ktor.server.application
 import io.ktor.http.*
 import io.ktor.http.content.*
 import io.ktor.server.plugins.*
+import io.ktor.server.plugins.ContentTransformationException
 import io.ktor.server.request.*
 import io.ktor.server.response.*
 import io.ktor.util.*
 import io.ktor.util.pipeline.*
 import io.ktor.util.reflect.*
 import io.ktor.utils.io.*
-import kotlinx.coroutines.*
-import kotlin.coroutines.*
+import kotlinx.coroutines.CoroutineScope
 
 private val RECEIVE_TYPE_KEY: AttributeKey<TypeInfo> = AttributeKey("ReceiveType")
 
@@ -63,6 +63,27 @@ public interface ApplicationCall : CoroutineScope {
     public val parameters: Parameters
 
     /**
+     * Receives content for this request according to [typeInfo].
+     *
+     * This function returns `null` only when [TypeInfo.isNullable] is `true`.
+     * The caller is responsible for ensuring that [typeInfo] represents the same type as [T].
+     *
+     * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.server.application.ApplicationCall.receive)
+     *
+     * @param typeInfo instance specifying the type to be received.
+     * @return an instance of [T] received from this call.
+     * @throws ContentTransformationException when content cannot be transformed to the requested type.
+     */
+    public suspend fun <T> receive(typeInfo: TypeInfo): T {
+        @Suppress("DEPRECATION")
+        val result = receiveNullable<T>(typeInfo)
+        if (!typeInfo.isNullable && result == null) throw CannotTransformContentToTypeException(typeInfo)
+
+        @Suppress("UNCHECKED_CAST")
+        return result as T
+    }
+
+    /**
      * Receives content for this request.
      *
      * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.server.application.ApplicationCall.receiveNullable)
@@ -71,6 +92,7 @@ public interface ApplicationCall : CoroutineScope {
      * @return instance of [T] received from this call.
      * @throws ContentTransformationException when content cannot be transformed to the requested type.
      */
+    @Deprecated("Use 'receive<T>(typeInfo)' with nullable T instead", ReplaceWith("receive<T?>(typeInfo)"))
     public suspend fun <T> receiveNullable(typeInfo: TypeInfo): T?
 
     /**
--- a/ktor-server/ktor-server-core/common/src/io/ktor/server/application/PipelineCall.kt
+++ b/ktor-server/ktor-server-core/common/src/io/ktor/server/application/PipelineCall.kt
@@ -119,7 +119,7 @@ public interface PipelineCall : ApplicationCall {
         when {
             transformed == NullBody -> return null
             transformed === DoubleReceivePreventionToken -> throw RequestAlreadyConsumedException()
-            !typeInfo.type.isInstance(transformed) -> throw CannotTransformContentToTypeException(typeInfo.kotlinType!!)
+            !typeInfo.type.isInstance(transformed) -> throw CannotTransformContentToTypeException(typeInfo)
         }
 
         @Suppress("UNCHECKED_CAST")
--- a/ktor-server/ktor-server-core/common/src/io/ktor/server/plugins/Errors.kt
+++ b/ktor-server/ktor-server-core/common/src/io/ktor/server/plugins/Errors.kt
@@ -5,11 +5,12 @@
 package io.ktor.server.plugins
 
 import io.ktor.http.*
-import io.ktor.server.application.internal.*
 import io.ktor.util.internal.*
-import kotlinx.coroutines.*
-import kotlinx.io.*
-import kotlin.reflect.*
+import io.ktor.util.reflect.*
+import kotlinx.coroutines.CopyableThrowable
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.io.IOException
+import kotlin.reflect.KType
 
 /**
  * Base exception to indicate that the request is not correct due to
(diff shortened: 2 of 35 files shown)


ktor-server/ktor-server-core/common/src/io/ktor/server/application/PipelineCall.kt
```diff
@@ -62,6 +62,26 @@ public interface ApplicationCall : CoroutineScope {
      */
     public val parameters: Parameters
 
+    /**
+     * Receives content for this request according to [typeInfo].
+     *
+     * This function returns `null` only when [TypeInfo.isNullable] is `true`.
```
osipxd, line 68: Probably we should mention here that it's caller's responsibility to ensure the `TypeInfo` parameter represents the same type as the `T`.

ktor-server/ktor-server-core/common/src/io/ktor/server/request/ApplicationReceiveFunctions.kt
```diff
@@ -122,6 +128,11 @@ public suspend fun <T : Any> ApplicationCall.receive(type: KClass<T>): T {
  * @throws ContentTransformationException when content cannot be transformed to the requested type.
  * @throws NullPointerException when content is `null`.
  */
+@Suppress("EXTENSION_SHADOWED_BY_MEMBER", "DEPRECATION")
+@Deprecated(
+    "Maintained for binary compatibility",
+    level = DeprecationLevel.HIDDEN,
+)
 public suspend fun <T> ApplicationCall.receive(typeInfo: TypeInfo): T = receiveNullable(typeInfo)!!
```
osipxd, line 136: The extension is shadowed by the introduced interface function, so it's safe to mark it HIDDEN

ktor-server/ktor-server-plugins/ktor-server-di/nonJvm/src/io/ktor/server/plugins/di/utils/Types.nonJvm.kt
```diff
@@ -19,14 +19,15 @@ public actual fun TypeInfo.hierarchy(): Sequence<TypeInfo> =
...
- * For non-JVM platforms, we omit support for resolving covariant types of declared dependencies.
+ * Converts raw types without reflection. Converting [kotlinType] is only supported on the JVM platform.
  *
  * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.server.plugins.di.utils.toNullable)
  */
 @InternalAPI
-public actual fun TypeInfo.toNullable(): TypeInfo? = null
+public actual fun TypeInfo.toNullable(): TypeInfo? {
+    if (isNullable || kotlinType != null) return null
+    return TypeInfo(type = type, isNullable = true)
```
osipxd, line 29: Now we can implement this method for nonJvm targets. The question is should we? @bjhham

ktor-utils/common/src/io/ktor/util/reflect/Type.kt
```diff
@@ -28,28 +31,50 @@ public expect val KType.platformType: Type
...
- * @property kotlinType Kotlin reified type with all generic type parameters.
+ * @property isNullable `true` when `null` is a valid value for this type.
+ * @property kotlinType Kotlin reified type with all generic type parameters, or `null` when type information is
+ * unavailable.
  */
 public class TypeInfo(
     public val type: KClass<*>,
-    public val kotlinType: KType? = null
+    public val isNullable: Boolean,
+    public val kotlinType: KType? = null,
```
osipxd: Should we mark this constructor as `@InternalAPI`?
The only use case for it is to create raw types with nullability preserved. We use it only for DI.
@bjhham, wdyt?

ktor-client/ktor-client-plugins/ktor-client-content-negotiation/common/src/io/ktor/client/plugins/contentnegotiation/ContentNegotiation.kt
```diff
@@ -362,6 +362,8 @@ public val ContentNegotiation: ClientPlugin<ContentNegotiationConfig> = createCl
                 return null
             }
 
+        // if (info.isNullable && body.peek(1) == null) return NullBody
```
osipxd: Ooops. I should remove this

ktor-server/ktor-server-core/common/src/io/ktor/server/response/ApplicationResponseFunctions.kt
```diff
@@ -72,9 +76,13 @@ public suspend fun ApplicationCall.respond(
  *
  * @see [io.ktor.server.response.ApplicationResponse]
  */
+@Deprecated(
+    "Use 'respond<T>()' with nullable T instead.",
+    ReplaceWith("this.respond<T>(status, message)", "io.ktor.server.response.respond"),
+)
 public suspend inline fun <reified T> ApplicationCall.respondNullable(status: HttpStatusCode, message: T) {
```
osipxd, line 75: **Up for discussion.** Do we want to make `respond` work consistently with `receive` and deprecate `respondNullable`?
The only downside we see is  that `respondNullable` could be used as defensive programming to prevent accidental responding with `null`.

ktor-server/ktor-server-core/common/src/io/ktor/server/plugins/Errors.kt
```diff
@@ -87,12 +88,21 @@ public abstract class ContentTransformationException(message: String) : IOExcept
 
 @OptIn(ExperimentalCoroutinesApi::class)
 public class CannotTransformContentToTypeException(
-    private val type: KType
-) : ContentTransformationException("Cannot transform this request's content to $type"),
+    private val typeDescription: String
+) : ContentTransformationException("Cannot transform this request's content to $typeDescription"),
```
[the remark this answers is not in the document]
osipxd: Fixed

ktor-server/ktor-server-plugins/ktor-server-di/nonJvm/src/io/ktor/server/plugins/di/utils/Types.nonJvm.kt
bjhham: I see no reason not to 👍

ktor-server/ktor-server-core/common/src/io/ktor/server/response/ApplicationResponseFunctions.kt
bjhham: I'd be in favor of keeping it as is to preserve the defensiveness.  It's more likely that someone would accidentally return null than actually return null intentionally.

ktor-utils/common/src/io/ktor/util/reflect/Type.kt
bjhham: Yeah, makes sense to me.

ktor-utils/common/src/io/ktor/util/reflect/Type.kt
```diff
@@ -28,28 +31,50 @@ public expect val KType.platformType: Type
...
  * @property type Source KClass<*>
- * @property kotlinType Kotlin reified type with all generic type parameters.
+ * @property isNullable `true` when `null` is a valid value for this type.
+ * @property kotlinType Kotlin reified type with all generic type parameters, or `null` when type information is
+ * unavailable.
  */
 public class TypeInfo(
     public val type: KClass<*>,
-    public val kotlinType: KType? = null
+    public val isNullable: Boolean,
```
bjhham, line 41: Wondering why not just delegate to `kotlinType?.isMarkedNullable == true`?  `typeOf` should always work nowadays right?

osipxd: There are still some corner cases when our `typeOfOrNull` wrapper can return `null`:
- [KT-42913](https://youtrack.jetbrains.com/issue/KT-42913) typeOf() fails on anonymous object inside class with generic parameter
- [KTOR-7479](https://youtrack.jetbrains.com/issue/KTOR-7479) KotlinReflectionInternalError (createClientPlugin) when running release APK on Android

So `isNullable = null is T` inside `typeInfo<T>()` gives us more reliable result.

ktor-server/ktor-server-core/common/src/io/ktor/server/response/ApplicationResponseFunctions.kt
osipxd: Agree. It's different from `receive`/`receiveNullable` because there we always specify the desired type explicitly (either as a variable type or as a type parameter). While with `respond` it's very easy to pass a type with unknown nullability. I'll remove this commit

ktor-utils/common/src/io/ktor/util/reflect/Type.kt
bjhham: Ah, thanks, good to know 👍

bjhham: It might be best to include this in the KDoc as well for future reference.

bjhham (approved):

ktor-server/ktor-server-core/common/src/io/ktor/server/request/ApplicationReceiveFunctions.kt
```diff
@@ -96,7 +98,11 @@ public suspend inline fun <reified T : Any> ApplicationCall.receive(): T = recei
  * @return instance of [T] received from this call.
  * @throws ContentTransformationException when content cannot be transformed to the requested type.
  */
-public suspend inline fun <reified T> ApplicationCall.receiveNullable(): T? = receiveNullable(typeInfo<T>())
+@Deprecated(
+    "Use 'receive<T>()' with nullable T instead",
+    ReplaceWith("this.receive<T>()", "io.ktor.server.request.receive"),
```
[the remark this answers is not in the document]
osipxd: Considering `receiveNullable<Foo>()` was an incorrect use of this function, this preserves old semantics.

PR #5833, opened 2026-08-24
KTOR-9796 Add flaky-test infrastructure and stabilize timeout/cache/websocket tests

VGoncharova:
**Subsystem**
Test Infrastructure — `build-logic`, `ktor-test-base` (`common`, `jvm`), `ktor-server-test-base`,                                                            `ktor-server-test-suites`; test-only changes in `ktor-client-tests`, `ktor-client-okhttp`,                                                                        `ktor-server-cio`, `ktor-server-core`, `ktor-server-auth`, `ktor-server-websockets`, `ktor-utils`.                                                                
                                                                                                      
**Motivation**
 A 28-day cross-source analysis found ~840 flaky and 619 failed test-executions across ~4,091 builds, 
 collapsing into a handful of cross-cutting patterns rather than ~50 independent bad tests. 
 Two habits kept them alive:                                                                                       
                                                                                                                                                                    
  - **`@Ignore("flaky")`** — the test disappears from every report, so nobody learns whether it still                                                               
    flips or whether the underlying bug got fixed. Several such tests had been disabled for over a year.                                                            
  - **Blanket `retries = 3..10`** — retries are applied to *any* failure, so a deterministic assertion                                                              
    failure is re-run and eventually reported as "flaky", masking real bugs.                                                                       
                                                                                                                                                                    
  On top of that, the flakiness itself came from repeatable causes: synchronous liveness asserts that                                                               
  check state the instant a call returns, and tight 1–5s timeout ceilings used both as the guard *and*                                                              
  the assertion against a real loopback test server. 

**Solution**

*1. Quarantine instead of `@Ignore`                                                                                                                   
                                                                                                                                                                    
  - New `@Flaky(ticket)` annotation in `ktor-test-base`, requiring a YouTrack issue. Flaky tests are                                                                
    excluded from the default run but still **executed and reported** by a dedicated task, so they stay                                                             
    visible.                                                                                                                                          
  - JVM enforcement via `FlakyTestCondition` (`ExecutionCondition`), auto-registered through JUnit                                                                  
    extension autodetection, so `@Flaky` works on common-source tests without a per-test `@ExtendWith`.                                                                                                                 
  - Native/JS/Wasm have no JUnit Platform, so selection there uses a `_flaky` name token, applied by                                                                
    `FlakyTestsConfig` to every `AbstractTestTask`. `FlakyNameConventionTest` guards the convention.                                                                
                                                                                                                                                                    
  Both selectors — the annotation on JVM and the name token everywhere — are driven by **one property,                                                              
  `ktor.tests.flaky`**, with the same three values as a Gradle property (`-P`) and as the system                                                                    
  property JVM test tasks forward to the condition:                                                                                                                 
                                                                                                                                                                    
  | Invocation                                         | Runs                                     |                                                                 
  |----------------------------------------------------|------------------------------------------|                                                                 
  | `./gradlew :module:jvmTest`                        | everything except flaky (`exclude`)      |                                                                 
  | `./gradlew :module:jsNodeTest -Pktor.tests.flaky=only` | flaky only                           |                                                                 
  | `./gradlew :module:macosArm64Test -Pktor.tests.flaky=all` | everything, flaky included        |                                                                 
  | `./gradlew :module:flakyTest`                      | `@Flaky` only, JVM (pins `only`)         |                                                                 
                                                                                                                                                                    
  An unrecognized value fails at configuration time with the accepted values listed. The JVM-only                                                                   
  `flakyTest` task exists because the annotation is resolved at execution time rather than during                                                                   
  selection, which is what lets it catch JVM tests that carry no `_flaky` token.         

*2. Transient-only retry*                                                                                                                                         
                                                                                                                                                                    
  `BaseTest.runTest` and `runTestWithData` now classify failures through `DeterministicFailureGuard`:                                                               
  an `AssertionError` thrown by the test body is deterministic and is reported by the first attempt                                                                 
  instead of being retried. `runTest`'s own timeout error stays retryable, since it's thrown *around*                                                               
  the body rather than by it. This let the `retries =` overrides be dropped from `HttpTimeoutTest`.                                                                                  
                                                                                                                                                                    
  *3. `assertEventually`*                                                                                                                                           
                                                                                                                                                                    
  Replaces `waitForCondition` (deleted from `ktor-client-tests`) with a real-time-delay helper in                                                                   
  `ktor-test-base`, available to client and server tests alike. Applied to the synchronous                                                                          
  `isCompleted` / `getActiveChildren()` asserts in `HttpTimeoutTest` and `HttpStatementTest`.   

*4. Per-pattern stabilization*                                                                                                                                    
                                                                                                                                                                    
  - **Timeout cluster :** must-timeout tests made structural — a small timeout against a                                                                 
    large delay (`TIMEOUT_TRIGGER_DELAY_MS`) rather than two nearby values; must-succeed guards raised                                                              
    to a single `GUARD_TIMEOUT_MS = 15s`; `testConnectionRefusedException` now asserts `IOException`                                                                
    with a 5s connect timeout instead of swallowing `ConnectTimeoutException`.                                                                                      
  - **Cache :** `testExpiresScenario` waits out the real deadline (`delayUntil`) instead of a                                                            
    separately supplied `expireDelay`, plus a 1s margin for `toHttpDate()`'s second granularity — a 2s                                                              
    offset could leave as little as 1s of freshness, which two round trips can outlive on a loaded agent.                                                           
  - **WebSockets:** `WebSocketBackpressureTest` asserts the close cause on the test coroutine rather                                                                
    than inside `invokeOnClose` (where a failure was swallowed), with a shared `CLOSE_TIMEOUT`.                                                                     
  - **Cleanup:** `OAuth2Test` closes only on successful completion and cancels `testClient` tasks.                                                                  
                                                                                                                                                                    
  *5. Previously-`@Ignore`d tests brought back under quarantine*                                                                                                    
                                                                                                                                                                    
  KTOR-8854, KTOR-7811, KTOR-9179, KTOR-8570, KTOR-9789 (WinHttp `ContentTest.testByteArray`,                                                                       
  `OverridingClassLoaderTest`, `DeflaterReadChannelTest`, ws-route leak test). 
 `RawWebSocketTest.testServerIncomingConnectionLoss` stays `@Ignore`d with a comment: it fails *deterministically*                                                                  
  after ~30s, so quarantining it would just report a standing nightly failure (KTOR-7318).


ktor-shared/ktor-test-base/common/src/io/ktor/test/Eventually.kt
```diff
@@ -0,0 +1,41 @@
...
+suspend fun assertEventually(
+    description: String,
+    timeout: Duration = DEFAULT_EVENTUALLY_TIMEOUT,
+    interval: Duration = (timeout / 20).coerceIn(1.milliseconds, 100.milliseconds),
+    condition: suspend () -> Boolean,
+) {
+    val deadline = TimeSource.Monotonic.markNow() + timeout
+    while (true) {
+        if (condition()) return
+        if (deadline.hasPassedNow()) fail("Timed out after $timeout waiting for $description")
```
osipxd, line 33: Probably we should throw `TimeoutCancellationException` here. But it can be created only by using `withTimeout`.

ktor-shared/ktor-test-base/common/src/io/ktor/test/Flaky.kt
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * On JVM this is enforced by `io.ktor.test.junit.FlakyTestCondition` (auto-registered via JUnit
+ * extension autodetection). There is no JUnit Platform on Native/JS/Wasm, so annotations can't be
+ * used for selection there. **For a test in common code, also add the `_flaky` token to its name**,
+ * which the build filters out on every target:
+ *
+ * ```
+ * @Flaky("KTOR-1234")
+ * @Test
+ * fun testReconnectsAfterClose_flaky() { ... }
```
osipxd, line 22: I think it's a great example of how we can use [Konsist](https://github.com/LemonAppDev/konsist).
```kotlin
@Test
fun `flaky naming and annotation are consistent outside jvm source set`() {
    val tests = Konsist.scopeFromTest()
    val jvmTests = Konsist.scopeFromSourceSet("jvmTest")
    val nonJvmTests = tests - jvmTests

    // Rule 1: @Flaky tests must end with _flaky
    nonJvmTests
        .functions()
        .withAnnotationOf(Flaky::class)
        .assertTrue { function ->
            function.name.endsWith("_flaky")
        }

    // Rule 2: _flaky tests must be annotated with @Flaky
    nonJvmTests
        .functions()
        .filter { function ->
            function.name.endsWith("_flaky")
        }
        .assertTrue { function ->
            function.hasAnnotationOf(Flaky::class)
        }
}
```

ktor-shared/ktor-test-base/jvm/src/io/ktor/test/junit/FlakyTestCondition.kt
```diff
@@ -0,0 +1,109 @@
...
+ *  - `all` — run the full suite with `@Flaky` tests included.
+ *
+ * Mirrors `StressTestCondition`, but is annotation-driven and auto-registered via JUnit
+ * extension autodetection (see `META-INF/services/org.junit.jupiter.api.extension.Extension`
+ * and `junit.jupiter.extensions.autodetection.enabled` in build-logic `JvmConfig`), so
+ * `@Flaky` on a common-source test method is honored without a per-test `@ExtendWith`.
+ */
+class FlakyTestCondition : ExecutionCondition {
+    override fun evaluateExecutionCondition(context: ExtensionContext): ConditionEvaluationResult {
+        val mode = FlakyTestsMode.current()
```
osipxd, line 65: I think we should read the property only once, save as a field in the class and then reuse

ktor-shared/ktor-test-base/jvm/src/io/ktor/test/junit/FlakyTestCondition.kt
```diff
@@ -0,0 +1,109 @@
...
+    /** Finds [Flaky] on the current test method, or on its class or any of its base classes. */
+    private fun ExtensionContext.findFlaky(): Flaky? =
+        AnnotationSupport.findAnnotation(element, Flaky::class.java).orElse(null)
+            ?: testClass.orElse(null)?.findFlakyInHierarchy()
+
+    /**
+     * Searches [Flaky] up the superclass chain.
+     *
+     * JUnit walks superclasses only for annotations meta-annotated with
+     * `java.lang.annotation.Inherited` (see `AnnotationUtils.findAnnotation`), which a Kotlin
```
osipxd, line 96: Actually we can define this annotation as `expect` in the common source set and then make it a bit different for `jvm` and `nonJvm` source sets. Alternatively we could define `@JvmInherited` that would be an alias to `java.lang.annotation.Inherited` for `jvm` source set and an empty annotation for other source sets.

ktor-shared/ktor-test-base/common/test/io/ktor/test/EventuallyTest.kt
```diff
@@ -0,0 +1,75 @@
...
+    fun `evaluates the condition at least once with a zero timeout`(): TestResult = runTest {
+        var evaluations = 0
+
+        assertEventually("an immediately satisfied condition", timeout = Duration.ZERO) { evaluations++ >= 0 }
+
+        assertEquals(1, evaluations)
+    }
+
+    @Test
+    fun `waits in real time so the virtual clock does not skip the poll interval`(): TestResult = runTest {
```
osipxd, line 62: Minor. The explicit `TestResult` can be removed from function here

ktor-shared/ktor-test-base/common/src/io/ktor/test/runTestWithData.kt
```diff
@@ -114,20 +116,27 @@ fun <T> runTestWithData(
...
                 runTest(context, timeout = timeout) {
-                    start = timeSource.markNow()
-                    test(testCase)
-                    afterEach(TestSuccess(testCase, start.elapsedNow()))
+                    val mark = timeSource.markNow()
+                    start = mark
+                    try {
+                        test(testCase)
+                    } catch (cause: Throwable) {
+                        assertionFailure = cause is AssertionError
```
osipxd, line 136: This logic is duplicated in `DeterministicFailureGuard`. Would it be possible to use it here as well?

ktor-server/ktor-server-test-base/jvm/src/io/ktor/server/test/base/BaseTestJvm.kt
```diff
@@ -49,14 +49,21 @@ actual abstract class BaseTest actual constructor() {
...
+                if (retry > 0) println("[Retry $retry/$retries]")
+                beforeTest()
+                try {
+                    block()
+                } catch (cause: Throwable) {
+                    guard.record(cause)
+                    throw cause
+                } finally {
+                    afterTest()
+                }
```
osipxd, line 66: Looks like this implementation might be extracted in a common function an used from there

ktor-server/ktor-server-test-base/jvm/src/io/ktor/server/test/base/BaseTestJvm.kt
```diff
@@ -49,14 +49,21 @@ actual abstract class BaseTest actual constructor() {
...
-            if (retry > 0) println("[Retry $retry/$retries]")
-            beforeTest()
-            try {
-                block()
-            } finally {
-                afterTest()
+    ): TestResult {
+        val guard = DeterministicFailureGuard()
+        return retryTest(retries) { retry ->
+            guard.failFast()
```
osipxd, line 55: Maybe the guard should be applied inside the `retryTest`, so we don't have to call failFast here?

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/HttpStatementTest.kt
```diff
@@ -92,11 +97,15 @@ class HttpStatementTest : ClientLoader() {
         }
     }
 
+    // The Android engine reads the body with a blocking call that doesn't honour cancellation, so
+    // throwing from the block leaves the streaming read hanging until the guard below fires and the
+    // test body never completes.
+    @Flaky("KTOR-8570")
     @Test
-    fun testStreamingResponseExceptionInBodyCancelsImmediately() = clientTests {
+    fun testStreamingResponseExceptionInBodyCancelsImmediately_flaky() = clientTests {
```
osipxd, line 105: The only idea I have for such cases is to provide two tests, like this:

```kotlin
    @Test
    fun testStreamingResponseExceptionInBodyCancelsImmediately() = clientTests(except("Android")) {
        this.testStreamingResponseExceptionInBodyCancelsImmediately()
    }

    @Flaky("KTOR-8570")
    @Test
    fun testStreamingResponseExceptionInBodyCancelsImmediately_flaky() = clientTests(only("Android")) {
        this.testStreamingResponseExceptionInBodyCancelsImmediately()
    }

    private fun TestClientBuilder<*>.testStreamingResponseExceptionInBodyCancelsImmediately() {
        test { client ->
            // implementation
        }
    }
```

This way we won't skip unaffected engines in daily tests

ktor-client/ktor-client-tests/common/test/io/ktor/client/tests/HttpStatementTest.kt
```diff
@@ -81,7 +86,7 @@ class HttpStatementTest : ClientLoader() {
     fun testStreamingResponseExceptionCancelsImmediately() = clientTests {
         test { client ->
             val exception = assertFailsWith<IllegalStateException> {
-                withTimeout(2000) {
+                withTimeout(GUARD_TIMEOUT_MS) {
```
osipxd, line 89: It's better to pass timeout in `clientTests` as it configures `runTest` timeout

build-logic/src/main/kotlin/ktorbuild/targets/FlakyTestsConfig.kt
```diff
@@ -0,0 +1,88 @@
...
+/**
+ * Selects how test tasks treat flaky tests, as a Gradle property (`-P`).
+ *
+ * The same name and the same values are used for the system property that JVM test tasks pass to
+ * `io.ktor.test.junit.FlakyTestCondition`, so there is one spelling to remember for both selectors.
+ */
+internal const val FLAKY_MODE_PROPERTY = "ktor.tests.flaky"
+
+/** How test tasks treat tests marked with [FLAKY_NAME_TOKEN]. Selected by [FLAKY_MODE_PROPERTY]. */
+internal enum class FlakyTestsMode {
```
osipxd, line 32: I think we should share this code between build-logic and the test-base module. It should be achievable by declaring a module both in the root project and in the build logic.

This module could be used also for test server address constants sharing.

```kotlin
// settings.gradle.kts
include("shared/ktor-test-constants")

// build-logic/settings.gradle.kts
include("../shared/ktor-test-constants")
```

build-logic/src/main/kotlin/ktorbuild/targets/JvmConfig.kt
```diff
@@ -60,6 +66,26 @@ private fun Project.configureTests() {
         useJUnitPlatform()
         configureJavaToolchain(java.toolchain.languageVersion, ktorBuild.jvmTestToolchain)
     }
+
+    // Runs ONLY @Flaky-annotated tests (excluded from the default `jvmTest`). Intended for a
+    // nightly/dedicated job that publishes to Develocity so quarantined tests stay tracked.
+    // (Selection is by annotation, resolved at execution time; the `_flaky` name token — same
+    // property, applied by `configureFlakyTests` — is the equivalent for Native/JS/Wasm, which have
+    // no JUnit Platform.)
+    tasks.register<Test>(FLAKY_TEST_TASK) {
```
osipxd, line 75: I think we should also configure to forcefully rerun these tests even if their inputs haven't changed in Gradle

osipxd (commented):
Wow! Huge work!
I've reviewed mostly infrastructure parts. Could you extract them in a separate PR so we could review and merge them before other changes?

PR #5834, opened 2026-08-25
KTOR-9627 Propagate cancellation through blocking bridges

osipxd:
**Subsystem**
I/O, Client/Server

**Motivation**
[`KTOR-9627`](https://youtrack.jetbrains.com/issue/KTOR-9627): `ByteReadChannel.toInputStream` and `ByteWriteChannel.toOutputStream` use `runBlocking` internally. Without a parent job, cancellation of the call or coroutine using these bridges does not interrupt a suspended blocking read or write.

**Solution**
Allow blocking channel bridges to receive a parent `Job` and use it for their internal `runBlocking` calls. Ktor-owned call sites now attach bridges to the appropriate client call, server call, or scoped coroutine job.

Add regression coverage for cancellation of blocked input reads, output writes, and HTTP blocking output bridges. Preserve existing JVM binary signatures and update ABI dumps. Related cleanups remove redundant JVM stream wrappers and avoid dispatching construction of a lazy JSON sequence to `Dispatchers.IO`.


bjhham (approved):
👌

PR #5835, opened 2026-08-25
KTOR-9788 Fix exception in call handler coroutine due to bad state

bjhham:
**Subsystem**
Server, Netty

**Motivation**
[KTOR-9788](https://youtrack.jetbrains.com/issue/KTOR-9788) Netty: engine double-responds after its built-in 400, crashing the call-handler coroutine with "Headers can no longer be set because response was already completed"

**Solution**
After using internal response functions, ensure the base response is marked as sent as well.  I also looked around for any other locations where this state could happen and fixed these spots too.

### DIFF
--- a/ktor-server/ktor-server-core/common/src/io/ktor/server/engine/BaseApplicationResponse.kt
+++ b/ktor-server/ktor-server-core/common/src/io/ktor/server/engine/BaseApplicationResponse.kt
@@ -33,7 +33,7 @@ public abstract class BaseApplicationResponse(
         get() = responded
 
     final override var isSent: Boolean = false
-        private set
+        protected set
 
     override val cookies: ResponseCookies by lazy(LazyThreadSafetyMode.NONE) {
         ResponseCookies(this)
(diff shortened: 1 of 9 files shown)


ktor-server/ktor-server-core/common/src/io/ktor/server/engine/BaseApplicationResponse.kt
zibet27, line 27: Might want to add SubclassOptInRequired at some point here

ktor-server/ktor-server-core/common/src/io/ktor/server/engine/BaseApplicationResponse.kt
```diff
@@ -33,7 +33,7 @@ public abstract class BaseApplicationResponse(
         get() = responded
 
     final override var isSent: Boolean = false
-        private set
+        protected set
```
zibet27, line 37: Can we mark a `set` operation as an internal api?

ktor-server/ktor-server-core/common/src/io/ktor/server/engine/BaseApplicationResponse.kt
bjhham: I'll add this to our 4.0 breaking changes doc.

osipxd (approved):

PR #5837, opened 2026-08-26
Update webrtc-java version and unlock ignored tests

zibet27:
**Subsystem**
WebRTC Client

**Motivation**
`DataChannelEvent.BufferedAmountLow` was not emitted because of a bug in `webrtc-java`, so the corresponding tests were ignored

**Solution**
Updated it to the new version that was released a couple of days ago.
Also, updated `build.gradle.kts` because the IDE was complaining about context parameters in tests.
Removed now redundant IgnoreJvm annotation.


bjhham (approved):
🥳

PR #5838, opened 2026-08-26
KTOR-9796 Add flaky-test quarantine infrastructure and transient-only retry

VGoncharova:
**Subsystem**

 Test Infrastructure — `build-logic`, `ktor-test-base` (`common`, `jvm`), `ktor-server-test-base`                                                                  
  (`common`, `jvm`, `nonJvm`).

**Motivation**

  KTOR-9796, part of the epic KTOR-9789.                                                                                                                            
                                                                                                                                                                    
  A 28-day cross-source analysis found ~840 flaky and 619 failed test-executions across ~4,091 builds. 
  They collapse into ~8 cross-cutting patterns rather than ~50 independent bad tests, and two habits kept them alive:                                                                                            
                                                                                                                                                                    
  - **`@Ignore("flaky")`** — the test disappears from every report, so nobody learns whether it still                                                               
    flips or whether the underlying bug got fixed. Several had been disabled for over a year.                                                                       
  - **Blanket `retries = 3..10`** — retries apply to *any* failure, so a deterministic assertion                                                                    
    failure is re-run and eventually reported as "flaky", masking a real bug.                                                                                       
                                                                                                                                                                    
  There was also no shared primitive for "wait until", so liveness assertions checked state the instant                                                             
  a call returned.                                                                                                                                                  
                                                                                                                                                                    
  This is the infrastructure half of #5833, extracted on review request so it can be reviewed and                                                                   
  merged ahead of the per-test changes. It adds primitives and build wiring only — **no test is                                                                     
  quarantined by this PR**, so `flakyTest` currently selects just the three self-test fixtures.
  
  **Solution**

  Covers the five work items of KTOR-9796.                                                                                                                          
                                                                                                                                                                    
  *1. Transient-only retry*                                                                                                                                         
                                                                                                                                                                    
  `DeterministicFailureGuard` classifies an `AssertionError` thrown by the test body as deterministic:                                                              
  it is reported by the first attempt instead of being retried. `runTest`'s own timeout error stays                                                                 
  retryable, since it is thrown *around* the body rather than by it. Wired into `runTestWithData` and                                                               
  into `BaseTest.runTest` on both JVM and non-JVM. `DEFAULT_RETRIES` is unchanged (`0` on JVM, `1`                                                                  
  elsewhere).                                                                                                                                                       
                                                                                                                                                                    
  *2. `@Flaky(ticket)`*                                                                                                                                             
                                                                                                                                                                    
  New annotation in `ktor-test-base`, requiring a YouTrack issue. Flaky tests are excluded from the                                                                 
  default run but still **executed and reported** by a dedicated task, so they stay visible. Enforced                                                               
  on JVM by `FlakyTestCondition` (an `ExecutionCondition` modelled on the existing                                                                                  
  `StressTestCondition`), auto-registered through JUnit extension autodetection so `@Flaky` works on                                                                
  common-source tests without a per-test `@ExtendWith`. 

*3. Cross-platform selection by name*                                                                                                                             
                                                                                                                                                                    
  Native/JS/Wasm have no JUnit Platform, so selection there uses a `_flaky` name token, applied by                                                                  
  `FlakyTestsConfig` to every `AbstractTestTask`. `FlakyNameConventionTest` guards the convention.                                                                  
                                                                                                                                                                    
  *4. `flakyTest` task*                                                                                                                                             
                                                                                                                                                                    
  JVM-only, pins the mode to `only`, and sets `ignoreFailures = true`: a quarantined test flipping is                                                               
  the expected outcome, not a regression to block on. Failures still land in the reports and in                                                                     
  Develocity, which is where the flip rate is tracked.                                                                                                              
                                                                                                                                                                    
  Both selectors are driven by **one property, `ktor.tests.flaky`**, with the same three values as a                                                                
  Gradle property and as the system property JVM test tasks forward to the condition:

  | Invocation                                                | Runs                                |                                                               
  |-----------------------------------------------------------|-------------------------------------|                                                               
  | `./gradlew :module:jvmTest`                                | everything except flaky            |                                                               
  | `./gradlew :module:jsNodeTest -Pktor.tests.flaky=only`     | flaky only                         |                                                               
  | `./gradlew :module:macosArm64Test -Pktor.tests.flaky=all`  | everything, flaky included         |                                                               
  | `./gradlew :module:flakyTest`                              | `@Flaky` only, JVM (pins `only`)   |                                                               
                                                                                                                                                                    
  An unrecognized value fails at configuration time with the accepted values listed.                                                                                
                                                                                                                                                                    
  *5. `assertEventually`*                                                                                                                                           
                                                                                                                                                                    
  A real-time-delay polling helper in `ktor-test-base`, available to client and server tests alike,                                                                 
  with `DEFAULT_EVENTUALLY_TIMEOUT = 10s`. `waitForCondition` is left in place here; its call sites                                                                 
  migrate in the follow-up PR. 

**Known follow-ups** (kept out to keep this reviewable, all raised in #5833 review) 
                                                                              
  Cache the system property in `FlakyTestCondition`; express `@Flaky` as `expect`/`actual` or add a                                                                 
  `@JvmInherited` alias instead of walking the superclass chain by hand; use Konsist for the naming                                                                 
  rule; share `FLAKY_NAME_TOKEN`/`FlakyTestsMode` between build-logic and `ktor-test-base` via an                                                                   
  included module; `outputs.upToDateWhen { false }` on `flakyTest`; extract the JVM/nonJvm `BaseTest`                                                               
  duplication; and an engine-scoped `clientTests(flaky = ...)` so a flake confined to one engine no                                                                 
  longer needs a duplicated test.

### DIFF
--- a/build-logic/build.gradle.kts
+++ b/build-logic/build.gradle.kts
@@ -22,6 +22,18 @@ dependencies {
     implementation(files(libs.javaClass.superclass.protectionDomain.codeSource.location))
 }
 
+// The flaky-test selectors are driven by one property with one set of values, read here to
+// configure the test tasks and in `ktor-test-base` to decide whether a `@Flaky` test runs. Compiling
+// `ktor-test-constants` into build-logic keeps that a single declaration instead of two that have to
+// be kept in sync by hand.
+//
+// The module is added as a source directory rather than as a project dependency because it is a
+// Kotlin Multiplatform project built by the `ktorbuild.project.internal` convention plugin — which
+// this build produces, so including it here would be circular.
+sourceSets.main {
+    kotlin.srcDir("../ktor-shared/ktor-test-constants/common/src")
+}
+
 // Should be synced with gradle/gradle-daemon-jvm.properties
 kotlin {
     jvmToolchain(21)
--- a/build-logic/src/main/kotlin/ktorbuild/targets/FlakyTestsConfig.kt
+++ b/build-logic/src/main/kotlin/ktorbuild/targets/FlakyTestsConfig.kt
@@ -1,0 +1,93 @@
+/*
+ * Copyright 2014-2026 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package ktorbuild.targets
+
+import org.gradle.api.Project
+import org.gradle.api.tasks.testing.AbstractTestTask
+import org.gradle.api.tasks.testing.Test
+import org.gradle.kotlin.dsl.withType
+
+/**
+ * Name token marking a test as flaky on every target.
+ *
+ * The `@Flaky` annotation is enforced by a JUnit `ExecutionCondition`, and there is no JUnit
+ * Platform on Native/JS/Wasm. A token in the test or class name is the one selector that works
+ * everywhere, because Gradle test filtering applies to every `AbstractTestTask`.
+ */
+const val FLAKY_NAME_TOKEN = "_flaky"
+
+/** Matches the token anywhere in the test name, which also covers the `[target]` name suffixes. */
+private const val FLAKY_TEST_PATTERN = "*$FLAKY_NAME_TOKEN*"
+
+/**
+ * Selects how test tasks treat flaky tests, as a Gradle property (`-P`).
+ *
+ * The same name and the same values are used for the system property that JVM test tasks pass to
+ * `io.ktor.test.junit.FlakyTestCondition`, so there is one spelling to remember for both selectors.
+ */
+internal const val FLAKY_MODE_PROPERTY = "ktor.tests.flaky"
+
+/** How test tasks treat tests marked with [FLAKY_NAME_TOKEN]. Selected by [FLAKY_MODE_PROPERTY]. */
+internal enum class FlakyTestsMode {
+    /** The default: flaky tests don't run. */
+    EXCLUDE,
+
+    /** Nightly: run flaky tests and nothing else, to track whether they still flip. */
+    ONLY,
+
+    /** Run everything, flaky included. */
+    ALL;
+
+    /** The [FLAKY_MODE_PROPERTY] value denoting this mode, as accepted by [of]. */
+    val propertyValue: String get() = name.lowercase()
+
+    companion object {
+        fun of(value: String?): FlakyTestsMode = when (value) {
+            null, "", "exclude" -> EXCLUDE
+            "only" -> ONLY
+            "all" -> ALL
+            else -> error(
+                "Unexpected value '$value' of '$FLAKY_MODE_PROPERTY'. Expected one of: exclude, only, all."
+            )
+        }
+    }
+}
+
+internal fun Project.flakyTestsMode(): FlakyTestsMode =
+    FlakyTestsMode.of(providers.gradleProperty(FLAKY_MODE_PROPERTY).orNull)
+
+/**
+ * Applies the [FLAKY_MODE_PROPERTY] mode to the test tasks that have no other way to select flaky
+ * tests, which is every target except the JVM.
+ *
+ * JVM test tasks are left alone on purpose. They select `@Flaky` by annotation through
+ * `FlakyTestCondition`, which receives the same mode as a system property (see `JvmConfig`), and the
+ * two selectors would *intersect* rather than agree: [FlakyTestsMode.ONLY] would run only the tests
+ * that are both annotated **and** named `*_flaky*`, silently skipping every `@Flaky` test that
+ * carries no name token. The name token exists only because Native/JS/Wasm have no JUnit Platform.
+ */
+internal fun Project.configureFlakyTests() {
+    val mode = flakyTestsMode()
+
+    tasks.withType<AbstractTestTask>().configureEach {
+        if (this !is Test) selectFlakyTests(mode)
+    }
+}
+
+internal const val FLAKY_TEST_TASK = "flakyTest"
+
+private fun AbstractTestTask.selectFlakyTests(mode: FlakyTestsMode) {
+    when (mode) {
+        FlakyTestsMode.EXCLUDE -> filter.excludeTestsMatching(FLAKY_TEST_PATTERN)
+
+        FlakyTestsMode.ONLY -> {
+            filter.includeTestsMatching(FLAKY_TEST_PATTERN)
+            // Most modules have no flaky tests at all, and that isn't a failure.
+            filter.isFailOnNoMatchingTests = false
+        }
+
+        FlakyTestsMode.ALL -> Unit
+    }
+}
--- a/build-logic/src/main/kotlin/ktorbuild/targets/FlakyTestsConfig.kt
+++ b/build-logic/src/main/kotlin/ktorbuild/targets/FlakyTestsConfig.kt
@@ -4,79 +4,47 @@
 
 package ktorbuild.targets
 
+import io.ktor.test.constants.FLAKY_MODE_PROPERTY
+import io.ktor.test.constants.FLAKY_NAME_TOKEN
+import io.ktor.test.constants.FlakyTestsMode
 import org.gradle.api.Project
 import org.gradle.api.tasks.testing.AbstractTestTask
-import org.gradle.api.tasks.testing.Test
 import org.gradle.kotlin.dsl.withType
-
-/**
- * Name token marking a test as flaky on every target.
- *
- * The `@Flaky` annotation is enforced by a JUnit `ExecutionCondition`, and there is no JUnit
- * Platform on Native/JS/Wasm. A token in the test or class name is the one selector that works
- * everywhere, because Gradle test filtering applies to every `AbstractTestTask`.
- */
-const val FLAKY_NAME_TOKEN = "_flaky"
 
 /** Matches the token anywhere in the test name, which also covers the `[target]` name suffixes. */
 private const val FLAKY_TEST_PATTERN = "*$FLAKY_NAME_TOKEN*"
 
-/**
- * Selects how test tasks treat flaky tests, as a Gradle property (`-P`).
- *
- * The same name and the same values are used for the system property that JVM test tasks pass to
- * `io.ktor.test.junit.FlakyTestCondition`, so there is one spelling to remember for both selectors.
- */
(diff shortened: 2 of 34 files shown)


build-logic/src/main/kotlin/ktorbuild/targets/JvmConfig.kt
```diff
@@ -41,25 +51,98 @@ internal fun Project.configureJvm() {
...
         useJUnitPlatform()
         configureJavaToolchain(java.toolchain.languageVersion, ktorBuild.jvmTestToolchain)
     }
+
+    // Runs ONLY @Flaky-annotated tests (excluded from the default `jvmTest`). Intended for a
+    // nightly/dedicated job that publishes to Develocity so quarantined tests stay tracked.
+    // (Selection is by annotation, resolved at execution time; the `_flaky` name token — same
+    // property, applied by `configureFlakyTests` — is the equivalent for Native/JS/Wasm, which have
+    // no JUnit Platform.)
+    tasks.register<Test>(FLAKY_TEST_TASK) {
```
osipxd: Do we need this task? Will it work differently from `test[Something] -Pktor.tests.flaky=only`?

build-logic/src/main/kotlin/ktorbuild/targets/JvmConfig.kt
```diff
@@ -41,25 +51,98 @@ internal fun Project.configureJvm() {
...
+        // Failures still land in the test reports and in Develocity, which is where the flip rate is tracked.
+        ignoreFailures = true
+        // Each run is a fresh sample of whether the test still flips, so an unchanged input tree is
+        // no reason to skip it — up-to-date checking would report the previous run's verdict forever.
+        outputs.upToDateWhen { false }
+        systemProperty(FLAKY_MODE_PROPERTY, FlakyTestsMode.ONLY.propertyValue)
+        systemProperty("junit.jupiter.extensions.autodetection.enabled", "true")
+        exclude("**/*StressTest*")
+        useJUnitPlatform()
+        configureJavaToolchain(java.toolchain.languageVersion, ktorBuild.jvmTestToolchain)
```
osipxd, line 130: I think we should apply common configs in `tasks.withType<Test>().configureEach { ... }` for all JVM test tasks (heap size, properties, JUnit platform, toolchain)

build-logic/src/main/kotlin/ktorbuild/targets/JvmConfig.kt
```diff
@@ -41,25 +51,98 @@ internal fun Project.configureJvm() {
...
+            jvmTestRuntimeClasspath.any { it.invariantSeparatorsPath.contains(TEST_BASE_MODULE) }
+        }
+
+        maxHeapSize = "2g"
+        // A quarantined test flipping is the expected outcome here, not a regression to block on.
+        // Failures still land in the test reports and in Develocity, which is where the flip rate is tracked.
+        ignoreFailures = true
+        // Each run is a fresh sample of whether the test still flips, so an unchanged input tree is
+        // no reason to skip it — up-to-date checking would report the previous run's verdict forever.
+        outputs.upToDateWhen { false }
```
osipxd, line 125: After extracting common configs to another place, these two properties will be visible, so comments won't be necessary

build-logic/src/main/kotlin/ktorbuild/targets/JvmConfig.kt
```diff
@@ -41,25 +51,98 @@ internal fun Project.configureJvm() {
...
+    // Take the test classpath from the JVM test *compilation*, not from the `jvmTest` task. A
+    // provider derived from `tasks.named("jvmTest")` carries that task as its producer, so depending
+    // on it makes Gradle run the whole default suite before `stressTest` or `flakyTest`. The
+    // compilation's own file collections are built by the compile tasks, which is all these need.
+    val jvmTestCompilation = kotlin.jvm().compilations.named("test")
+    val jvmTestRuntimeClasspath = files(
+        jvmTestCompilation.map { it.output.allOutputs },
+        jvmTestCompilation.map { it.runtimeDependencyFiles },
+    )
+    val jvmTestClassesDirs = files(jvmTestCompilation.map { it.output.classesDirs })
```
osipxd, line 65: Won't the same trick as in `inheritExecutionSettingsFrom` work in this case?

build-logic/build.gradle.kts
```diff
@@ -22,6 +22,18 @@ dependencies {
...
 }
 
+// The flaky-test selectors are driven by one property with one set of values, read here to
+// configure the test tasks and in `ktor-test-base` to decide whether a `@Flaky` test runs. Compiling
+// `ktor-test-constants` into build-logic keeps that a single declaration instead of two that have to
+// be kept in sync by hand.
+//
+// The module is added as a source directory rather than as a project dependency because it is a
+// Kotlin Multiplatform project built by the `ktorbuild.project.internal` convention plugin — which
+// this build produces, so including it here would be circular.
```
osipxd, line 32: Could we reduce this comment to something like this?
```
We want to share test constants logic between tests and build-logic.
Adding that module as a project dependency would be circular as it uses plugins from build-logic.
```

ktor-shared/ktor-test-base/common/src/io/ktor/test/Eventually.kt
```diff
@@ -0,0 +1,61 @@
...
+        val remaining = -deadline.elapsedNow()
+        if (remaining <= Duration.ZERO) {
+            throw EventuallyTimeoutException("Timed out after $timeout waiting for $description")
+        }
+        // Never sleep past the deadline: `timeout` is documented as the upper bound on the wait, so
+        // an interval larger than what is left would overshoot it (`interval = 1s` with a 100ms
+        // timeout would fail after a second).
+        realTimeDelay(minOf(interval, remaining))
+    }
+}
```
osipxd, line 56: My idea was to throw the native `CoroutineTimeoutException` so it could be later recognized by `runTest` and it could print coroutines trace. I think the following shape with a separate watcher coroutine should do the trick:

```kotlin
suspend fun assertEventually(
    timeout: Duration = DEFAULT_EVENTUALLY_TIMEOUT,
    interval: Duration = (timeout / 20).coerceIn(1.milliseconds, 100.milliseconds),
    condition: () -> Boolean,
) = coroutineScope {
    val realTimeDispatcher = Dispatchers.Default.limitedParallelism(1)
    val completion = CompletableDeferred<Unit>()

    val watcher = async(realTimeDispatcher + CoroutineName("timeout-watcher")) {
        withTimeout(timeout) { completion.join() }
    }

    while (true) {
        watcher.ensureActive()
        if (condition()) {
            completion.complete(Unit)
            break
        }
        withContext(realTimeDispatcher) { delay(interval) }
    }
    watcher.await()
}
```

One important detail. The `condition` shouldn't be a `suspend` lambda as if it hangs we can't limit its execution time

ktor-shared/ktor-test-base/jvm/test/io/ktor/test/FlakyConventionTest.kt
```diff
@@ -0,0 +1,70 @@
...
+
+    @Test
+    fun `Flaky annotation and name token agree outside jvm-only test sources`() {
+        val functions = nonJvmTestFunctions()
+
+        functions
+            .filter { it.isFlaky() }
+            .assertTrue(
+                additionalMessage = "A @Flaky test compiled for a non-JVM target also needs the " +
+                    "'$FLAKY_NAME_TOKEN' name token, which is the only selector those targets have.",
```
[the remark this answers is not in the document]
osipxd: Looks like a valid point.

ktor-shared/ktor-test-base/jvm/test/io/ktor/test/FlakyConventionTest.kt
```diff
@@ -0,0 +1,70 @@
...
+
+    /**
+     * Test functions from every test source set except the JVM-only one.
+     *
+     * The repository uses a flattened, platform-centric source layout (`<module>/<platform>/test`),
+     * so source sets are selected by path rather than by Konsist's `scopeFromSourceSet`, which
+     * expects the conventional `src/<name>` directories.
+     */
+    private fun nonJvmTestFunctions(): List<KoFunctionDeclaration> = Konsist
+        .scopeFromProject()
```
osipxd, line 58: Does this `Konsist.scopeFromProject()` produces a scope only for this project? Will it cover other modules?

ktor-shared/ktor-test-base/build.gradle.kts
```diff
@@ -14,12 +14,18 @@ kotlin {
...
         }
 
         jvmMain.dependencies {
             api(libs.kotlin.test.junit5)
             api(libs.junit)
             api(libs.kotlinx.coroutines.debug)
         }
+
+        jvmTest.dependencies {
+            // Enforces the `@Flaky` / `_flaky` naming convention across the repository's sources.
```
osipxd, line 27: I think we can drop this comment

build-logic/src/main/kotlin/ktorbuild/targets/FlakyTestsConfig.kt
```diff
@@ -0,0 +1,61 @@
...
+/**
+ * Applies the [FLAKY_MODE_PROPERTY] mode to every test task that cannot select `@Flaky` by
+ * annotation — every target without a JUnit Platform, plus JVM tasks outside
+ * [ANNOTATION_DRIVEN_TEST_TASKS]. The name token exists only for those.
+ */
+internal fun Project.configureFlakyTests() {
+    val mode = flakyTestsMode()
+
+    tasks.withType<AbstractTestTask>().configureEach {
+        if (name !in ANNOTATION_DRIVEN_TEST_TASKS) selectFlakyTests(mode)
```
osipxd, line 45: Btw, what if change this check to `if (this is Test)` and configure the property here, while in the `else` branch call the `selectFlakyTests`? This would remove applying the flaky test mode from `JvmConfig.kt`.

osipxd (changes requested):
Great job! I have some comments on how we could deduplicate test tasks configuration and improve `assertEventually`.

Also I've noticed that the agent (was it Claude?) tends to write too many comments even for cases when code is straightforward. So left comments for some of these places

PR #5842, opened 2026-08-28
Performance improvements for static content handling

bjhham:
**Subsystem**
Server, Static Content

**Motivation**
A 5-10% performance degradation in static files for 3.6.0 [here](https://github.com/MDA2AV/HttpArena/pull/1204)

**Solution**
- Small allocations improvement accept encodings list
- Reduce syscalls in precompressed handling by only checking against the next format listed for size
- Cleaned up some duplicate code


osipxd (approved):
Nice 👍

PR #5849, opened 2026-09-04
KTOR-9870 Optimize OkHttp request/response handling

osipxd:
**Subsystem**
Client, `ktor-client-okhttp`

**Motivation**
[KTOR-9870](https://youtrack.jetbrains.com/issue/KTOR-9870) tracks unnecessary coroutine dispatching and allocations in OkHttp request and response streaming.

For responses, `BufferedSource.toChannel` copied approximately one 4–8 KiB chunk per iteration through a producer coroutine and an intermediate `ByteChannel`. `ByteWriteChannel.write(ByteBuffer)` already flushed on every iteration, and the caller performed a second redundant flush. A 2 MiB response produced about 513 iterations; allocation measurements attributed approximately 24.6 KiB to each flush continuation site.

After optimizing the response path, reading a 2 MiB streaming response allocates **63,605 fewer bytes** than the baseline. Approximately 49.2 KiB corresponds to the two flush continuation sites; the remaining reduction is consistent with removing the suspend `write()` continuation on each iteration. This GET response benchmark does not exercise the request-body change.

The non-duplex request path similarly adapted a `ByteReadChannel` through `InputStream` and Okio `Source` before writing it to the request sink.

**Solution**
- Copy non-duplex request bodies directly from `ByteReadChannel` to the OkHttp sink output stream under the call job, cancelling the channel if the copy fails.
- Create the response `ByteChannel` explicitly and read into it with non-suspending `writeAvailable`.
- Preserve immediate streaming by publishing every response chunk with `flushWriteBuffer()`.
- Invoke suspending `flush()` only when the response channel reaches its backpressure limit.
- Recheck the channel failure after a potentially blocking OkHttp read and before publishing bytes.
- Add a focused test proving that response data is available before source EOF and that the source closes after EOF.

Blocking OkHttp reads remain on the engine dispatcher. Call cancellation, exception mapping, response adapters, source closure, bounded buffering, and single-reader channel semantics are preserved.

### DIFF
--- a/ktor-client/ktor-client-okhttp/jvm/src/io/ktor/client/engine/okhttp/OkHttpEngine.kt
+++ b/ktor-client/ktor-client-okhttp/jvm/src/io/ktor/client/engine/okhttp/OkHttpEngine.kt
@@ -6,7 +6,7 @@ package io.ktor.client.engine.okhttp
 
 import io.ktor.client.call.*
 import io.ktor.client.engine.*
-import io.ktor.client.io.configurePlatform
+import io.ktor.client.io.*
 import io.ktor.client.plugins.*
 import io.ktor.client.plugins.sse.*
 import io.ktor.client.plugins.websocket.*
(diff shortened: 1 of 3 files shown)


ktor-client/ktor-client-okhttp/jvm/src/io/ktor/client/engine/okhttp/OkHttpEngine.kt
```diff
@@ -171,25 +171,31 @@ public class OkHttpEngine(override val config: OkHttpConfig) : HttpClientEngineB
...
+                            runCatching { context.job.getCancellationException() }.getOrDefault(cause)
                         throw mapExceptions(cancelOrCloseCause, requestData)
                     }
                 }
-                channel.flush()
+                responseChannel.rethrowCloseCauseIfNeeded()
+                if (bytesWritten == -1) break
+
+                responseChannel.flushWriteBuffer()
+                if (!responseChannel.hasFreeSpace) responseChannel.flush()
```
osipxd, line 194: I don't like that I kind of reinvent `channel.write { ... }` logic, but I think it's okay for this kind of bridging  APIs.

ktor-client/ktor-client-okhttp/jvm/src/io/ktor/client/engine/okhttp/StreamRequestBody.kt
```diff
@@ -49,8 +51,14 @@ internal class StreamRequestBody(
...
-                block().toInputStream().source().use {
-                    sink.writeAll(it)
+                val channel = block()
+                try {
+                    runBlocking(callContext.job) {
+                        channel.copyTo(sink.outputStream())
+                    }
+                } catch (cause: Throwable) {
+                    channel.cancel(cause)
+                    throw cause
```
bjhham, line 61: Maybe out of scope for this, but it would be cool to try out the interop submodule in kotlinx-io: https://github.com/Kotlin/kotlinx-io/tree/master/integration/kotlinx-io-okio

Could use [asKotlinxIoRawSink](https://kotlinlang.org/api/kotlinx-io/kotlinx-io-okio/kotlinx.io.okio/as-kotlinx-io-raw-sink.html) for the copy for a little more efficiency.

bjhham (approved):

PR #5850, opened 2026-09-05
ktor 7082 resource validation errors

mdedetrich:
**Subsystem:** Server, ktor-server-resources

**Motivation:** Resource instantiation wrapped every exception (including custom validation, e.g. `RequestValidationException`) into a generic `BadRequestException`, so `StatusPages` couldn't handle validation errors by their real type and the reason was hidden one level down in `.cause`.

**Solution:** Decode failures (malformed/missing parameter, a property's own serializer rejecting input) still throw `BadRequestException`/400. Exceptions from the resource class's own constructor/`init` now propagate unchanged, so `StatusPages` can handle them by type — but only when decoded via the compiler-generated serializer (the normal `get<T>`/`post<T>` case); a custom/delegating serializer keeps this behavior only for values it decodes through `decoder.decodeSerializableValue(...)`, not by calling `.deserialize()` directly. `ResourceSerializationException` now extends `SerializationException`, preserving its existing catchability for direct `ResourcesFormat.decodeFromParameters` callers.

**References:** Closes [KTOR-7082](https://github.com/ktorio/ktor-documentation/pull/868). Related documentation PR: ktorio/ktor-documentation#868

### DIFF
--- a/ktor-server/ktor-server-plugins/ktor-server-resources/build.gradle.kts
+++ b/ktor-server/ktor-server-plugins/ktor-server-resources/build.gradle.kts
@@ -15,5 +15,9 @@ kotlin {
             api(projects.ktorResources)
             api(libs.kotlinx.serialization.core)
         }
+
+        commonTest.dependencies {
+            implementation(projects.ktorServerStatusPages)
+        }
     }
 }
--- a/ktor-server/ktor-server-plugins/ktor-server-resources/common/src/io/ktor/server/resources/Routing.kt
+++ b/ktor-server/ktor-server-plugins/ktor-server-resources/common/src/io/ktor/server/resources/Routing.kt
@@ -5,11 +5,14 @@
 package io.ktor.server.resources
 
 import io.ktor.http.*
+import io.ktor.resources.*
+import io.ktor.resources.serialization.*
 import io.ktor.server.application.*
 import io.ktor.server.plugins.*
 import io.ktor.server.request.*
 import io.ktor.server.routing.*
 import io.ktor.util.*
+import io.ktor.utils.io.*
 import kotlinx.serialization.*
 
 /**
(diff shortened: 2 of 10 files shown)


bjhham (changes requested):
It seems like a risky change to pass through all exceptions besides serialization exceptions.  If checking the cause under the `StatusPages` plugin is too cumbersome, it would be better to introduce support there instead of changing the behaviour here.  This also doesn't seem to provide the feature described in the linked ticket.  It would be better to log a new issue in YT.

mdedetrich:
> It seems like a risky change to pass through all exceptions besides serialization exceptions. If checking the cause under the `StatusPages` plugin is too cumbersome, it would be better to introduce support there instead of changing the behaviour here. 

Yeah I  have similar thoughts, initially I thought it would be a simple task but as you can tell with all of the back and forth between the codex AI reviews it ended up being more convoluted.

The core issue here is working with a limitation of the current design, i.e. if you have

```kotlin
@Resource("/viewport")
class Viewport(val west: Double, val south: Double, val east: Double, val north: Double) {
    init {
        val reasons = buildList {
            if (east <= west) add("east ($east) must be greater than west ($west)")
            if (north <= south) add("north ($north) must be greater than south ($south)")
        }
        if (reasons.isNotEmpty()) throw RequestValidationException(this, reasons)
    }
}
```

Due to how the `@Resource` annotation works, where it executes before the route i.e.

```kotlin
get<Viewport> {
  call.respondText("OK ${it.west},${it.east}")
}
```

Since the validation of `ViewPort` happens before that route we have to do all of this convoluted exception chain mechanism.

A better design might be to add a validate argument/parameter to the `@Resource` annotation which points to a method contained within the class.

The nice thing about this solution (i.e. using `init`) is that `Viewport` is validated even if its not constructed in the context of a Ktor route, your basically defining a contract on that class that its only valid with certain parameters and it cannot be constructed otherwise.

Im happy to go in a different direction if given some indication/hint on how we should approach this problem?

> This also doesn't seem to provide the feature described in the linked ticket. It would be better to log a new issue in YT.

https://youtrack.jetbrains.com/issue/KTOR-7082

Unless I am missing something its solving the exact issue, with the only difference being that rather than having a user defined function i.e. `validate` in
```kotlin
@Resource("res")
class Res(val id: Int) {
    fun validate() {
        require(id > 0) { "id must be positive" }
    }
}
```
This solution opts to use `init`. This was an intentional decision, because it makes the implementation far simpler i.e. we don't have to worry about needing to pass in information about which function to call to validate the class. There are other solutions for this, i.e. some kind of `Validation` Interface which defines a `validate` function that `Res` in the above example would extend this `Validation` interface. The disadvantage of this solution is similar to the one mentioned before, i.e. you are now coupling the validation of the class to ktor/routing/Validation interface.

bjhham:
I'm still not quite seeing the benefit here...

Instead of having:

```kotlin
install(StatusPages) {
    exception<BadRequestException> { call, ex ->
         when(ex.cause) {
             is RequestValidationException -> TODO("handle validation")
             /* other cases */
         }
    }
}
```

you have:

```kotlin
install(StatusPages) {
    exception<RequestValidationException> { call, ex ->
         TODO("handle validation")
    }
}
```

But then you lose the standard exception thrown from the resource plugin that you can catch for other exception types.

If you wanted something more ergonomic, I would make the standard exception for the resources init exception extend `BadRequestException` so it's differentiated but backwards compatible, then if you want to skip the cause switch you could add a helper for `StatusPages` like `exceptionCause<RequestValidationException>` that could traverse causes as well as the top-level exception.

PR #5851, opened 2026-09-07
KTOR-9875 Netty performance issues fixes

bjhham:
**Subsystem**
Server, Netty

**Motivation**
[KTOR-9875](https://youtrack.jetbrains.com/issue/KTOR-9875) Netty performance issues

Investigations into Ktor's benchmark statistics have revealed a few problems that appear when a Netty server is under load.

1. Pipelined responses are not flushed while other requests are queued in the same connection.  This leads to a memory leak on the connection which could crash the server.
2. The runningLimit config item is not properly enforced due to a mistake in the logic when checking if reads on pipelined requests should continue.  This can lead to poisoning of connections after the threshold is reached.
3. Call jobs are all direct children of the application job.  When processing connections on a server with high parallelism and many connections, this causes major contention on the coroutine internals.
4. The `shareWorkGroup` flag shares the same overhead as the default configuration due to the thread-pinning logic.  This ought to be circumvented to allow for higher throughput.
5. Write jobs are cancelled instead of completed on happy path requests.  This allocates an exception each time, which is very bad for the allocation rate.
6. Empty HTTP/2 requests cause a leak from not being finished

**Solution**
Each fix required it's own investigation and fix, generally by running specific benchmarks with profiling and letting Claude comb over the results.  With everything combined, we're seeing a pretty significant improvement in Netty's performance https://github.com/MDA2AV/HttpArena/pull/1204#issuecomment-5551945825

Best to review this commit-by-commit.


osipxd (commented):
🔥

osipxd (approved):

bjhham:
Almost below 7kb on Netty Hello now...

```
Request consumes 7.04 KB, expected 7.22 KB.
       Difference: -179 <= 64 (allowed)
       Saved 179 bytes on request
```

PR #5862, opened 2026-09-08
Fix/content negotiation accept header

n0900:
**Subsystem**
Ktor Client ContentNegotiation

**Motivation**
Ktor's duplicate check uses ContentType.parse, which retains only the last media range. If an AcceptHeader is set by the user in the form of a combined header then only the last element is checked. In my example the explicit JSON exclusion (q=0) conflicts with the implicit q=1 added by ContentNegotiation.

**Solution**
Parse all elements of the Header via parseHeaderValue. Since this splits the header into value and params they need to be re-assembled so that it is correctly identified as duplicate.

For both instances test cases were added.
(Closed #5861 to rebase to `release/3.x` after reading #5852)

### DIFF
--- a/ktor-client/ktor-client-plugins/ktor-client-content-negotiation/build.gradle.kts
+++ b/ktor-client/ktor-client-plugins/ktor-client-content-negotiation/build.gradle.kts
@@ -14,5 +14,8 @@ kotlin {
         commonMain.dependencies {
             api(projects.ktorSerialization)
         }
+        commonTest.dependencies {
+            implementation(projects.ktorSerializationKotlinxJson)
+        }
     }
 }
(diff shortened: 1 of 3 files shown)


bjhham (approved):
Looks good, thanks 👍

n0900:
Should this branch be rebased onto the newer `release/3.x` or will you take it from here?
I’m available until Friday but won’t be able for two weeks after that.

Thanks!

bjhham:
Yes, please rebase and resolve the conflict 🙏

n0900:
Heads up: there appear to be unrelated test issues from the point where I branched off of release/3.x. I will force-push another rebase

Update: `UDPSocketTest[jvm] > testBroadcastSuccessful()[jvm] FAILED` out of scope for this PR

bjhham:
The tests should be fixed in the latest release branch now... I'll have to update the allocation benchmarks with a new baseline to get past this one.

n0900:
I will try

PR #5863, opened 2026-09-09
KTOR-9868 Report malformed TLS record headers as TlsException

zibet27:
**Subsystem**
Network — `ktor-network-tls` (`io.ktor.network.tls` record parser, JVM)

**Motivation**
[KTOR-9868](https://youtrack.jetbrains.com/issue/KTOR-9868) — *Ktor Network: TLS record parser crash on malformed input*

`ByteReadChannel.readTLSRecord()` parses the 5-byte TLS record header directly from peer input:

```kotlin
val type = TLSRecordType.byCode(readByte().toInt() and 0xff)   // Parser.kt:19
val version = readTLSVersion()                                  // -> TLSVersion.byCode
```

Both `TLSRecordType.byCode` and `TLSVersion.byCode` throw `IllegalArgumentException` for an unknown code. Because that is not an `IOException`, it slips past every `catch (cause: IOException)` a caller may have around a connection and reaches the application as an uncaught crash. That is exactly what the reporter saw in production:

```
Fatal Exception: java.lang.IllegalArgumentException: Invalid TLS record type code: 75
       at io.ktor.network.tls.TLSRecordType$Companion.byCode(TLSRecordType.java:32)
       at io.ktor.network.tls.ParserKt.readTLSRecord(Parser.kt:19)
```

The inconsistency is visible inside the very same function: the frame-size check four lines below already reports malformed input as a `TLSException`/`TlsException`, which *is* an `IOException` (a `javax.net.ssl.SSLException` on JVM).

**Solution**
`readTLSRecord` now maps both record-header lookups to `TlsException`, so malformed peer input is reported as an IO/TLS failure like every other malformed-input case in that function.

- `ktor-network/ktor-network-tls/jvm/src/io/ktor/network/tls/Parser.kt` — the record type read moves into a private `readTLSRecordType()` helper, and the existing private `readTLSVersion()` gets the same treatment. Both translate `IllegalArgumentException` into `TlsException`, preserving the original message (`Invalid TLS record type code: 75`) and attaching the cause.

Deliberately kept out of scope: `TLSHandshakeType.byCode`, `ServerKeyExchangeType.byCode` and `TLSAlertLevel/Type.byCode` share the same pattern and are reachable from the handshake path, but they are not on the reported stack trace and are not part of the record header. They are worth a separate look.

**Public API is unchanged.** `TLSRecordType.byCode` and `TLSVersion.byCode` keep their documented "or fail" contract and their exception type; only the internal parser is adjusted. `checkKotlinAbi` reports no ABI diff and no `api/*.api` dump was touched.

**Tests**
This branch includes the reproducer from #1 (`KTOR9868ReproTest`), so **PR #1 can be closed** once this is merged. One test was appended for the version half of the record header:

| Test | Before | After |
| --- | --- | --- |
| `unknown record type code is reported as an IO failure` | fails (`IllegalArgumentException`) | passes |
| `unknown record version code is reported as an IO failure` (new) | fails (`IllegalArgumentException`) | passes |
| `well-formed handshake record is parsed` | passes | passes |
| `oversized frame is reported as an IO failure` | passes | passes |

**Validation**
Run against the touched module in the order `AGENTS.md` prescribes (JDK 21):

```bash
./gradlew :ktor-network-tls:jvmTest        # KTOR9868ReproTest 4/4, HostnameUtilsTest 14/14,
                                           # TLSConfigBuilderTest 4/4, ClientCertificateRequestTest 1/1
./gradlew :ktor-network-tls:assemble       # BUILD SUCCESSFUL
./gradlew :ktor-network-tls:formatKotlin   # no changes produced
./gradlew :ktor-network-tls:lintKotlin     # BUILD SUCCESSFUL
./gradlew :ktor-network-tls:checkKotlinAbi # BUILD SUCCESSFUL — no ABI diff
```

Only JVM sources changed, so no other platform test tasks were run.

One pre-existing failure is unrelated to this change: `ConnectionTest.tlsWithoutCloseTest` connects to `www.google.com:443` and fails with `UnresolvedAddressException` in the sandbox used for this run. It was verified to fail identically on the unmodified tree.

Opened by the Air Fix automation; human review required; do not auto-merge.

---
Produced by AIR Automations. Name: Fix Draft / Run: https://air.jetbrains.cloud/org/05cf1a7f-6ab5-713b-abd3-29d0c8a05e2d/automations/78b14ee6-cedb-4284-bfc7-9fa43e5e8378?run=349d949c-52eb-4d0c-bd63-0de9f38c50bd


zibet27:
A PR is copied from https://github.com/zibet27/ktor/pull/2

I was experimenting with Air Automations
It can create a PR for the reproducer, create a PR with a fix, and update the fix after review
Applied limitations: PRs are made to my fork, and updates are made only after my review

osipxd (commented):
We have a flaw in our implementation: on invalid TLS protocol values we currently only throw a ‎`TlsException`, but we should also send a fatal alert and abort the connection, as required by the TLS RFC.

zibet27:
>  but we should also send a fatal alert and abort the connection, as required by the TLS RFC.

@osipxd Should I do it in this PR or open another one?

bjhham:
> [Osip Fatkullin (@osipxd)](https://github.com/osipxd) Should I do it in this PR or open another one?

I think this can wait.  If we have time to implement TLS 1.3 in 3.7.0 we'll probably rewrite quite a bit of the handshake logic.

ktor-network/ktor-network-tls/jvm/src/io/ktor/network/tls/TLSHandshakeType.kt
```diff
@@ -34,10 +34,15 @@ public enum class TLSHandshakeType(public val code: Int) {
          *
          * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.network.tls.TLSHandshakeType.Companion.byCode)
          */
-        public fun byCode(code: Int): TLSHandshakeType = when (code) {
-            in 0..0xff -> byCode[code]
-            else -> null
-        } ?: throw IllegalArgumentException("Invalid TLS handshake type code: $code")
+        public fun byCode(code: Int): TLSHandshakeType =
+            byCodeOrNull(code) ?: throw IllegalArgumentException("Invalid TLS handshake type code: $code")
```
bjhham, line 38: Why not just change the exception type under all these `byCode` functions instead of introducing `byCodeOrNull` everywhere?  To preserve the existing contract for anyone relying on `IllegalArgumentException`?

zibet27:
> I think this can wait

Do you mean this PR as well?

ktor-network/ktor-network-tls/jvm/src/io/ktor/network/tls/TLSHandshakeType.kt
zibet27: Yeah, and feels weird to throw `TlsException` in `byCode` methods

bjhham:
> Do you mean this PR as well?

Nah, this one is fine.  I just wouldn't invest more time until after planning.

ktor-network/ktor-network-tls/jvm/src/io/ktor/network/tls/TLSHandshakeType.kt
bjhham: I think maybe a better approach would be to broaden the exception handling so that everything is caught and wrapped with `TlsException` since it roughly translates to "something failed during the TLS handshake" and we should be sending an alert for any failure, not just a subset of known errors.

PR #5870, opened 2026-09-11
Fix KTOR-9794 Digest Auth: NullPointerException

zibet27:
**Subsystem**
Client Auth

**Motivation**
[KTOR-9794](https://youtrack.jetbrains.com/issue/KTOR-9794)

**Solution**
`parseChallenge` on every `isApplicable` and `addRequestHeaders` which:
- fixes NPE
- allows sending multiple requests safely

### DIFF
--- a/ktor-client/ktor-client-plugins/ktor-client-auth/common/src/io/ktor/client/plugins/auth/providers/DigestAuthProvider.kt
+++ b/ktor-client/ktor-client-plugins/ktor-client-auth/common/src/io/ktor/client/plugins/auth/providers/DigestAuthProvider.kt
@@ -116,10 +116,6 @@ public class DigestAuthProvider(
     override val sendWithoutRequest: Boolean
         get() = error("Deprecated")
 
-    private val serverNonce = atomic<String?>(null)
-
-    private val qop = atomic<String?>(null)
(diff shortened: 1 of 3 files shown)


ktor-client/ktor-client-plugins/ktor-client-auth/common/src/io/ktor/client/plugins/auth/providers/DigestAuthProvider.kt
```diff
@@ -116,10 +116,6 @@ public class DigestAuthProvider(
     override val sendWithoutRequest: Boolean
         get() = error("Deprecated")
 
-    private val serverNonce = atomic<String?>(null)
-
-    private val qop = atomic<String?>(null)
-    private val opaque = atomic<String?>(null)
     private val clientNonce = atomic<String?>(null)
 
     private val requestCounter = atomic(0)
```
osipxd, line 121: It is pre-existing, but I've noticed one problem with `nc`. Not critical.

> nc
>   the number of requests … that the client has sent with **the nonce value in this request** 

It seems we should track different counters per different nonces. But this doesn't block the PR

osipxd (approved):
👍

PR #5872, opened 2026-09-11
KTOR-9889 Fix deadlock in concurrent module loading with Dependency Injection

zackman0010:
## Summary

- Fixes KTOR-9889
- Under `ktor.application.startup = concurrent`, a module that resolves a dependency declared by a later-loaded module suspends on a `DependencyInitializer.Missing` placeholder. When the providing module later calls `provide()`, `Missing.provide()` piped the provided initializer's `Deferred` straight into `completeWith()` without ever starting it.
- `Explicit.resolve()` creates that `Deferred` lazily (`async(start = CoroutineStart.LAZY)`), and `CompletableDeferred.completeWith()` only *observes* completion via `invokeOnCompletion` - it never drives the underlying coroutine. Without an explicit `start()`, the lazy initializer's body never runs, so the `Deferred` never completes, and every resolver suspended on the `Missing` placeholder hangs forever.
- The fix calls `.start()` on the resolved `Deferred` before piping it into `completeWith()`. `start()` is a no-op for an already-started or already-completed `Deferred`, so this is safe for every `DependencyInitializer` variant, not just `Explicit`'s lazy one.
- Confirmed via thread dump that the resolving coroutine's `Deferred` sat in a `New` (never-started) state indefinitely - this is not a timing race, it's a deterministic hang whenever a resolver's module loads before its provider's under concurrent startup, which is exactly the ordering concurrent startup exists to support.
- Also worth noting: this bug defeats `kotlinx-coroutines-test`'s own `runTest` timeout. The `EmbeddedServer` startup path (`avoidingDoubleStartup`) blocks a real thread inside a `runBlocking` call with an unparented root Job (`EmptyCoroutineContext`), so `withTimeout`'s cancellation signal in `runTest` has no connected job tree to propagate into. A test that hits this bug hangs indefinitely rather than failing cleanly at the timeout, in both Gradle and IDE test runs.

Closes KTOR-9889

## Test plan
- Added a failing reproducer test (`KTOR-9889 concurrent startup resolves a dependency declared by a later-loaded module`) that hangs without the fix and passes in ~3ms with it.
- Ran the full `ktor-server-di` JVM test suite (`:ktor-server-di:jvmTest`) - 27 tests, 0 failures.
- Ran `:ktor-server-di:formatKotlin`/`lintKotlin` - clean.
- No public API changed, so no ABI dump update needed.


ktor-server/ktor-server-plugins/ktor-server-di/common/test/io/ktor/server/plugins/di/DependencyInjectionTest.kt
```diff
@@ -197,6 +198,26 @@ class DependencyInjectionTest {
         }
     }
 
+    @Test
+    fun `KTOR-9889 concurrent startup resolves a dependency declared by a later-loaded module`() = runTestWithRealTime {
```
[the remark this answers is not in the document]
zackman0010: Moved the test to JVM.

Side note: If `ktor.application.startup` only affects JVM and the value of `startupMode` is just ignored by all other platforms, shouldn't `startupMode` be an `expect`/`actual` combo that only uses `SEQUENTIAL` in non-JVM platforms? With the current setup, what should be a "Missing dependency" exception turns into a deadlock if `ktor.application.startup` is mistakenly set on a non-JVM platform.

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyInitializer.kt
```diff
@@ -194,7 +194,10 @@ public sealed interface DependencyInitializer {
          */
         public fun provide(other: DependencyInitializer) {
             if (delegate.compareAndSet(null, other)) {
-                deferred.completeWith(other.resolve(resolver))
+                val otherDeferred = other.resolve(resolver)
+                // Start the provided Explicit to avoid a potential deadlock (KTOR-9889)
+                otherDeferred.start()
```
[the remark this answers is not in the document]
zackman0010: Fixed, added a second test to verify

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyInitializer.kt
```diff
@@ -181,21 +181,19 @@ public sealed interface DependencyInitializer {
...
-         *
-         * This allows for suspending consumers to wait for a provider to supply a true function.
-         *
-         * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.server.plugins.di.DependencyInitializer.Missing.provide)
-         */
         public fun provide(other: DependencyInitializer) {
-            if (delegate.compareAndSet(null, other)) {
-                deferred.completeWith(other.resolve(resolver))
+            provideReturning(other).start()
+        }
```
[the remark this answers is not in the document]
zackman0010: Oops, did not mean to delete the documentation. Restoring that now.

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyInitializer.kt
```diff
@@ -193,9 +193,14 @@ public sealed interface DependencyInitializer {
...
+
+        internal fun provideReturning(other: DependencyInitializer): Deferred<Any?> {
+            if (delegate.compareAndSet(null, other.resolve(resolver))) {
+                deferred.completeWith(delegate.value!!)
             }
+            return delegate.value!!
```
bjhham: This change introduces potential NPEs and breaks the current behaviour of deferring loading until application start.

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyInitializer.kt
```diff
@@ -181,7 +181,7 @@ public sealed interface DependencyInitializer {
         private val resolver: DependencyResolver,
     ) : DependencyInitializer {
         private val deferred: CompletableDeferred<Any?> = CompletableDeferred()
-        private val delegate: AtomicRef<DependencyInitializer?> = atomic(null)
+        private val delegate: AtomicRef<Deferred<Any?>?> = atomic(null)
```
bjhham: Replacing the delegate initializer with a `Deferred` makes it another deferred field, not a delegate.

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyProvider.kt
```diff
@@ -82,15 +83,23 @@ internal class MapDependencyProvider(
...
         val create = DependencyInitializer.Explicit(key, value)
         log.debug { "Provided $key ${DependencyReference().externalTraceLine()}" }
-        trySet(key, create)
-        insertCovariantKeys(create, key)
+        listOfNotNull(
+            trySet(key, create),
+            *insertCovariantKeys(create, key).toTypedArray()
+        ).forEach {
+            it.start()
+        }
```
bjhham: This also eagerly starts initialization.

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyProvider.kt
```diff
@@ -82,15 +83,23 @@ internal class MapDependencyProvider(
...
+    private fun trySet(key: DependencyKey, newFunction: DependencyInitializer): Deferred<Any?>? {
+        var deferred: Deferred<Any?>? = null
         map[key] = when (val previous = map[key]) {
             null -> newFunction
 
-            is DependencyInitializer.Missing -> newFunction.also(previous::provide)
+            is DependencyInitializer.Missing -> {
+                deferred = previous.provideReturning(newFunction)
+                newFunction
+            }
```
bjhham: Making the basic functionality here depend on an internal function hurts extensibility.

ktor-server/ktor-server-plugins/ktor-server-di/common/test/io/ktor/server/plugins/di/DependencyInjectionTest.kt
```diff
@@ -197,6 +202,70 @@ class DependencyInjectionTest {
         }
     }
 
+    @Test
+    fun `KTOR-9889 covariant key sees stale Missing placeholder before map replacement is visible`() = runTest {
```
bjhham: We generally keep issue ticket numbers out of the code.

ktor-server/ktor-server-plugins/ktor-server-di/common/test/io/ktor/server/plugins/di/DependencyInjectionTest.kt
```diff
@@ -197,6 +202,70 @@ class DependencyInjectionTest {
...
+            coroutineScope = CoroutineScope(Dispatchers.Unconfined),
+        )
+
+        // Resolving the supertype key before anything provides it creates a Missing placeholder.
+        val greetingDeferred = resolver.getDeferred<GreetingService>(DependencyKey<GreetingService>())
+
+        // A concrete provider whose own initializer resolves that same supertype key (a legitimate
+        // decorator pattern). With Dispatchers.Unconfined, provide()'s start() call runs the
+        // initializer synchronously, deterministically exercising the map-write-visibility race:
+        // the initializer must see itself in the map, not the stale Missing it would await forever.
```
bjhham: Try to avoid the long-winded LLM comments in code.

bjhham (changes requested):
Thanks for the contribution!

I wasn't confident in the approach for the fix here and ended up working through it in another branch https://github.com/ktorio/ktor/tree/bjhham/concurrent-di-deadlock

You can cherry-pick the changes or I could open another PR or you could update with this other branch as reference.  The main thing is to ensure that we don't call the provide functions prematurely.

ktor-server/ktor-server-plugins/ktor-server-di/common/test/io/ktor/server/plugins/di/DependencyInjectionTest.kt
zackman0010: I had removed most of them, but missed this one. I really need to set up my global instructions to tell it to stop leaving these.

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyProvider.kt
zackman0010: I wasn't very happy about this approach either, your commit makes it much cleaner.

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyInitializer.kt
zackman0010: Should we change this variable name then? I changed the type because the `DependencyInitializer` wasn't used anywhere else in the code, so it was effectively only being used as a marker to determine if `provide` had already been called or not.

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyInitializer.kt
zackman0010: It's a moot point with your commit, but isn't `delegate.value` guaranteed to be non-nullable here? If it were null, then `compareAndSet` would return true and enter the `completeWith` branch, and `other.resolve(resolver)` is a non-nullable return type. If it were not null, `compareAndSet` returns false with `.value` already proved to be non-null.

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyProvider.kt
```diff
@@ -90,15 +91,17 @@ internal class MapDependencyProvider(
...
             else -> when (val result = resolveConflict(previous, newFunction)) {
                 Ambiguous ->
                     DependencyInitializer.Ambiguous.of(key, previous, newFunction)
 
                 Conflict -> onConflict(key)
 
-                KeepNew -> newFunction
+                KeepNew -> {
+                   if (previous is DependencyInitializer.Missing)
+                       previous.provide(newFunction)
```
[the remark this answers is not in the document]
zackman0010: @bjhham This is the one change you made that I was going to point out, Codex just beat me to it.

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyInitializer.kt
bjhham: Should be able to revert this class change entirely, since the delegate is only used for thread-safety here.

I also noticed there was a mistake in my branch in the previous linking part.  The fix is here https://github.com/ktorio/ktor/commit/a4bd16d3aa7f0e7cc93602a1c5e4b9ccfae23383

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyProvider.kt
```diff
@@ -82,15 +82,24 @@ internal class MapDependencyProvider(
...
         val create = DependencyInitializer.Explicit(key, value)
         log.debug { "Provided $key ${DependencyReference().externalTraceLine()}" }
-        trySet(key, create)
-        insertCovariantKeys(create, key)
+        val fulfilled = mutableListOf<Pair<DependencyInitializer.Missing, DependencyInitializer>>()
+        try {
+            trySet(key, create, fulfilled)
+            insertCovariantKeys(create, key, fulfilled)
+        } finally {
+            fulfilled.forEach { (missing, newFunction) -> missing.provide(newFunction) }
```
bjhham: I don't think this logic is required here.  If we just link the new initializer to the previous missing where relevant in `trySet`, the tracking is not required.  The https://github.com/ktorio/ktor/commits/bjhham/concurrent-di-deadlock/ branch is updated with the fix here - it had a mistake where it was linking prior to setting the map entry, but it will work so long as it is assigned afterwards.

bjhham (changes requested):

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyProvider.kt
zackman0010: @bjhham This is actually something I tried in one of the earlier commits, Codex marked it as a bug: https://github.com/ktorio/ktor/pull/5872#discussion_r3993795732

If I understand Codex's review correctly, it's a race condition. If the `provide` itself tries to `resolve` one of its own covariant keys, that covariant might not have been inserted yet, which breaks the circular dependency detection and allows a new `Missing` to be created. When the main loop inserts the covariant, it'll see the `Missing` and try to fulfill it, causing a deadlock due to a `Deferred` waiting on itself to complete. 

So either the `provide` can't occur until all the covariants are registered (what I implemented) or we need to add another circular dependency detection into the covariant registration.

bjhham: Oh, my apologies, I didn't see that one.  That's quite a good catch 😅 

Maybe we could clean it up a little by keeping with map standards and returning the previous keys?

Something like:

```kotlin
    override fun <T> set(key: DependencyKey, value: suspend DependencyResolver.() -> T) {
        val create = DependencyInitializer.Explicit(key, value)
        log.debug { "Provided $key ${DependencyReference().externalTraceLine()}" }
        val replaced = mutableListOf<DependencyInitializer>()
        trySet(key, create)?.let(replaced::add)
        replaced += insertCovariantKeys(create, key)
        // all replaced `Missing` initializers must be resolved with new function
        for (waiting in replaced.filterIsInstance<DependencyInitializer.Missing>()) {
            waiting.provide(create)
        }
    }

    private fun trySet(key: DependencyKey, newFunction: DependencyInitializer): DependencyInitializer? {
        val previous = map[key]

        map[key] = when(previous) {
            null -> newFunction

            else -> when (val result = resolveConflict(previous, newFunction)) {
                Ambiguous ->
                    DependencyInitializer.Ambiguous.of(key, previous, newFunction)

                Conflict -> onConflict(key)

                KeepNew -> newFunction

                KeepPrevious -> previous

                is Replace -> result.function
            }
        }

        return previous
    }
    
    private fun insertCovariantKeys(
        createFunction: DependencyInitializer.Explicit,
        key: DependencyKey
    ): List<DependencyInitializer> {
        val covariantKeys = keyMapping.map(key, 0).toList()
        log.trace { "Covariant keys: ${formatKeys(covariantKeys)}" }
        return buildList {
            for ((key, distance) in covariantKeys) {
                trySet(key, createFunction.derived(distance))?.also(::add)
            }
        }
    }
```

zackman0010: Oh, I like that a lot! It'll still need the try-finally part to make sure the `provide` is always called even if a covariant key throws an exception (another Codex callout, your Codex integration is _really_ good), but that's easy enough. I'll get this pushed up when I get into the office.

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyInitializer.kt
zackman0010: Done in the newest commit.

ktor-server/ktor-server-plugins/ktor-server-di/common/src/io/ktor/server/plugins/di/DependencyProvider.kt
zackman0010: Commit made. Took your advice of using map semantics to add the popped key into the MutableList. However, I did have to still pass the MutableList into `insertCovariantKeys` to append to it there instead of using `buildList` like in your suggestion. That was the only way I could see to be able to fulfill any popped `Missing` placeholders if an exception is thrown partway through processing covariants.

bjhham (approved):
Looks good, thanks!

PR #5873, opened 2026-09-13
Reject Accept entries with q=0 in ContentNegotiation

kdelay:
## What
With `checkAcceptHeaderCompliance = true`, the server `ContentNegotiation` plugin served a media type the client had explicitly refused. `checkAcceptHeader` ignored the quality, so `Accept: application/ktor;q=0` returned `200 OK` instead of `406 Not Acceptable`.

## Why
RFC 9110 12.5.1: a media range with a qvalue of 0 means "not acceptable". The server `Compression` plugin already applies this rule to `Accept-Encoding` (#5708).

## How
An entry with quality 0 no longer makes the content type acceptable. Scoped to the compliance check; the default lenient path is unchanged. No public API change.

## Tests
Two cases in `ContentNegotiationTest`. Without the fix, the q=0 case fails with `expected: <406 Not Acceptable> but was: <200 OK>`; the rest of the class passes. `:ktor-server-content-negotiation:jvmTest` is green after it.


bjhham (changes requested):
Thanks for the fix!  Please check the comment from codex with regards to wildcard matches.  We can also log a separate issue to handle this since it's an edge case that was also not handled previously.

bjhham (approved):
Looks good, thanks!

PR #5879, opened 2026-09-14
Fix KTOR-9640 forward autoFlush in reader/writer helpers

eusorov:
Fix [KTOR-9640](https://youtrack.jetbrains.com/issue/KTOR-9640) ktor-io reader/writer ignores autoFlush

**Subsystem**
I/O, `ktor-io`

**Motivation**
`CoroutineScope.reader` and `CoroutineScope.writer` accept an `autoFlush` parameter and drop it, always constructing `ByteChannel()` with the default `autoFlush = false`. The parameter carried `@Suppress("UNUSED_PARAMETER")`, so callers asking for auto-flushing silently got buffered behaviour: written bytes stayed invisible to the consumer until an explicit flush, the `CHANNEL_MAX_SIZE` threshold, or block completion.

**Solution**
Pass the flag into `ByteChannel(autoFlush)` in both helpers and drop the suppressions. The mechanism already worked — `ByteChannel` stores the flag and `ByteWriteChannel.flushIfNeeded()` honours it — only these two entry points discarded it. The overloads taking an explicit `ByteChannel`, and `openWriteChannel(autoFlush)`, were already correct.

- Behaviour change for in-repo callers that already request `autoFlush = true` — deflate/inflate, zstd, file writing, both Jetty engines, partial-content ranges — which now flush per write as intended. Expect movement in throughput and allocation numbers.


bjhham (changes requested):
I can find no reason why the `autoFlush` parameter was being unused here, so it looks like a good fix 👍 

It looks like you pulled in some commits from main.  Could you please fix the history on this branch?

eusorov:
> I can find no reason why the `autoFlush` parameter was being unused here, so it looks like a good fix 👍
> 
> It looks like you pulled in some commits from main. Could you please fix the history on this branch?

Updated the branch, is now based on release branch. Before it was based on main by mistake

bjhham (approved):

PR #5880, opened 2026-09-15
Fix sse heartbeat hang in tests

zibet27:
**Subsystem**
Test Server SSE 

**Motivation**
[KTOR-9790](https://youtrack.jetbrains.com/issue/KTOR-9790) testApplication: `sseSession()` never returns when the SSE route installs a `heartbeat`

**Solution**
- cherry-picked [KTOR-8705 Pass coroutine context to the channel writer](https://github.com/ktorio/ktor/commit/0ad7c7f146eef701b91cdc52a3610b1e189d5411)
- remove the `job` in `respondWriteChannelContent`
- cancel heartbeat right after `handle()`, so `coroutineScope {}` doesn't wait for it
- always close the session
- explicitly state that only one heartbeat is allowed


ktor-server/ktor-server-plugins/ktor-server-sse/common/src/io/ktor/server/sse/ServerSSESession.kt
```diff
@@ -152,25 +152,37 @@ public suspend inline fun <reified T : Any> ServerSSESessionWithSerialization.se
...
  * [Report a problem](https://ktor.io/feedback/?fqname=io.ktor.server.sse.heartbeat)
  *
  * @param heartbeatConfig a lambda that configures the [Heartbeat] object used for the heartbeat.
+ * @throws IllegalStateException if a heartbeat has already been started for this session.
  */
 public fun ServerSSESession.heartbeat(heartbeatConfig: Heartbeat.() -> Unit = {}) {
     val heartbeat = Heartbeat().apply(heartbeatConfig)
-    val heartbeatJob = Job(call.coroutineContext[Job])
+    val heartbeatState = HeartbeatState()
+    val registeredState = call.attributes.computeIfAbsent(heartbeatStateKey) { heartbeatState }
```
bjhham, line 165: Would be nice if we just had `putIfAbsent` in our attributes interface...

ktor-server/ktor-server-plugins/ktor-server-sse/common/src/io/ktor/server/sse/ServerSSESession.kt
```diff
@@ -152,25 +152,37 @@ public suspend inline fun <reified T : Any> ServerSSESessionWithSerialization.se
...
             send(event)
             delay(heartbeat.period)
         }
     }
-    call.attributes.put(heartbeatJobKey, heartbeatJob)
 }
 
-internal val heartbeatJobKey = AttributeKey<Job>("HeartbeatJobAttributeKey")
+internal class HeartbeatState {
+    lateinit var job: Job
```
bjhham: Could just make this nullable instead of lateinit so that we can avoid not initialized exceptions in the case of some random interruptions.

bjhham (approved):

bjhham:
Might make sense to point this to release.

PR #5885, opened 2026-09-17
Fix KTOR-9901 Client tests fail on Java 8

zibet27:
**Subsystem**
Client Tests

**Motivation**
[KTOR-9901](https://youtrack.jetbrains.com/issue/KTOR-9901) Client tests fail on Java 8

**Cause**
- https://github.com/ktorio/ktor/pull/5780 changed discovery from loading the first service to "materialising" all services.
- `ktor-client-java` compiles with Java 11
- tests include it unconditionally as `runtimeOnly`, but compiled with Java 8
- `HttpClient()` tries to load the Java Client and fails

**Solution**
- include `ktor-client-java` only for JVM 11+


ktor-client/ktor-client-tests/build.gradle.kts
```diff
@@ -1,3 +1,5 @@
+import ktorbuild.KtorBuildExtension
+
 /*
  * Copyright 2014-2025 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
  */
```
osipxd: 💅  Could you move the import under the copyright?

ktor-client/ktor-client-tests/build.gradle.kts
```diff
@@ -10,6 +12,8 @@ plugins {
     id("test-server")
 }
 
+val jvmToolchain = extensions.getByType<KtorBuildExtension>().jvmTestToolchain.get()
```
osipxd: The `ktorBuild` accessor should be available here

```suggestion
val jvmToolchain = ktorBuild.jvmTestToolchain.get()
```

ktor-client/ktor-client-tests/build.gradle.kts
zibet27: Good catch

osipxd (approved):

ktor-client/ktor-client-tests/build.gradle.kts
```diff
@@ -10,6 +10,8 @@ plugins {
     id("test-server")
 }
 
+val jvmToolchain = ktorBuild.jvmTestToolchain.get()
```
osipxd: Nit: For consistency with the `ktorbuild.project.client-plugin` this line could be moved closer to its usage

zibet27: yeah, previously it was there to access the correct `extensions`
I'll move it

zibet27:
@osipxd I'll also cherry-pick it onto `main`

PR #5890, opened 2026-09-18
KTOR-9920 Add public-suffix validation to Ktor client cookie storage

zibet27:
**Subsystem**
Client Cookies

**Motivation**
KTOR-9920 Add public-suffix validation to Ktor client cookie storage

**Solution**
- Introduce `CookieAcceptancePolicy` interface
- Introduce `FilteredCookiesStorage` accepting a list of policies, copy `AcceptAllCookiesStorage` logic there
- `AcceptAllCookiesStorage` becomes `FilteredCookiesStorage` with an empty policies list
- Public-suffix validation implemented as a policy, returned by `HttpClient.loadPublicSuffixCookiePolicy`
- `PublicSuffixRules` parsing needs more attention 

**Alternatives**
- enable "Public-suffix validation" directly on `AcceptAllCookiesStorage`(then, it's not "accept all")
- bundle public suffixes directly into Ktor, instead of loading them 

We might want to create a separate KLIP document for the feature

See https://www.rfc-editor.org/info/rfc6265/#section-5.3


ktor-client/ktor-client-core/jvm/src/io/ktor/client/plugins/cookies/PublicSuffixCookiePolicy.kt
```diff
@@ -0,0 +1,152 @@
...
+
+import io.ktor.client.*
+import io.ktor.client.plugins.expectSuccess
+import io.ktor.client.request.*
+import io.ktor.client.statement.*
+import io.ktor.http.*
+import java.text.Normalizer
+import java.text.Normalizer.Form.NFC
+
+private const val PUBLIC_SUFFIX_LIST_URL = "https://publicsuffix.org/list/public_suffix_list.dat"
```
osipxd, line 15: Should we embed it into JAR instead? How frequently it changes?

zibet27: Hmm, https://publicsuffix.org/list/ says almost every day, but it recommends fetching the list at most once a day per client

My initial concern was increasing bundle size, but checked it and it should be ~5%

